PR #1226, opened 2022-08-03
fix: gas price decimals in STC encodedTx OK-11031

taimanhui:
Also use the bigger gas limit between the one provided by dapp and
estimated from RPC.

Fixes: OK-11031

### DIFF
--- a/packages/engine/src/vaults/impl/stc/Vault.ts
+++ b/packages/engine/src/vaults/impl/stc/Vault.ts
@@ -34,6 +34,10 @@ import {
   ITransferInfo,
   IUnsignedTxPro,
 } from '../../types';
+import {
+  convertFeeGweiToValue,
+  convertFeeValueToGwei,
+} from '../../utils/feeInfoUtils';
 import { VaultBase } from '../../VaultBase';
 import { EVMDecodedTxType } from '../evm/decoder/types';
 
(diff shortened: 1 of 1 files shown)


packages/engine/src/vaults/impl/stc/Vault.ts
```diff
@@ -311,7 +314,9 @@ export default class Vault extends VaultBase {
 
     const encodedTxWithFee = {
       ...params.encodedTx,
-      gasPrice: new BigNumber(price || '0.000000001').toFixed(),
+      gasPrice: new BigNumber(price || '0.000000001')
+        .shiftedBy(network.feeDecimals)
```
revan-zhang: 这种计算应该统一封装函数的

revan-zhang (dismissed):

packages/engine/src/vaults/impl/stc/Vault.ts
zuozhuo: 有提供统一的函数 feeInfoUtils.convertFeeNativeToValue

sidmorizon (dismissed):

packages/engine/src/vaults/impl/stc/Vault.ts
sidmorizon: 有封装好的函数 feeInfoUtils.convertFeeGweiToValue

ByteZhang1024 (dismissed):

sunnylqm (approved):

sidmorizon (approved):

originalix (approved):

PR #1636, opened 2022-09-26
fix: always recover backup with cached password OK-12391

taimanhui:
Fixes: OK-12391


revan-zhang (approved):

taimanhui:
my bad, wrong issue ref...

PR #2904, opened 2023-04-26
feat: show error toast when gas fee is not enough

kwoktung:


packages/kit/src/views/Swap/doQuote.ts
```diff
@@ -44,7 +44,35 @@ const findBestResponse = async (
...
   }
-  return responses[0];
+  const notNullResponses = responses.filter(
+    (res) => res.data,
+  ) as Required<FetchQuoteResponse>[];
+  if (notNullResponses.length === 0) {
+    return;
+  }
+  let selectedRes = notNullResponses[0];
+  const currentPriority = Number(selectedRes.data.sellAmount);
```
sunnylqm: 这里会不会有数字溢出的可能？

kwoktung: 改啦。用bignumber

sunnylqm (approved):

originalix (approved):

linleiqin (approved):

PR #3030, opened 2023-05-28
fix:disable brc20 token detail

linleiqin:


originalix (dismissed):

packages/kit/src/views/Wallet/AssetsList/index.tsx
```diff
@@ -130,6 +145,11 @@ function AssetsList({
         onTokenPress({ token: item });
         return;
       }
+      if (
+        !tokenDetailEnable({ networkId, tokenAddress: item.tokenIdOnNetwork })
```
sidmorizon, line 179: 用 vault settings 来判断

kwoktung (dismissed):

PR #3365, opened 2023-08-09
refactor: read react-render-tracker switch value from mmkv store.

huhuanming:


packages/kit/src/hooks/useAppStorageSetting.ts
```diff
@@ -0,0 +1,19 @@
...
+    setData(!!value);
+  }, [key, initialValue]);
+
+  const setNewData = (value: boolean) => {
+    appStorage.setSetting(key, value);
+    setData(value);
+  };
+
+  return [data, setNewData];
+};
```
stackia: Can we just use redux persisted state (backed by appStorage)?

I believe we should restrict direct access to the appStorage (only use it outside the React rendering pipeline).

packages/shared/src/storage/appSetting.ts
```diff
@@ -0,0 +1,31 @@
+import { MMKV } from 'react-native-mmkv';
+
+import type { AsyncStorageStatic } from '@react-native-async-storage/async-storage';
+
+export const appSetting = new MMKV({
+  id: `onekey-app-setting`,
+});
```
stackia, line 7: We need to consider how to migrate existing data to MMKV.
Otherwise, this will break existing installations once got released.

packages/kit/src/hooks/useAppStorageSetting.ts
huhuanming: It's a bit special – the value of this switch needs to be read before React initializes.
This switch is used to start the react-render-tracker, and it must be executed before React initializes.

packages/shared/src/storage/appSetting.ts
huhuanming: No data migration is necessary. In app settings, mainly various scattered toggle values are stored.

stackia (approved):

kwoktung (approved):

ByteZhang1024 (approved):

PR #3369, opened 2023-08-09
fix: fix code scanning alert 754/755 [OK-22218]

stackia:
A small fix for:
- Alert 754: https://github.com/OneKeyHQ/app-monorepo/security/code-scanning/754
- Alert 755: https://github.com/OneKeyHQ/app-monorepo/security/code-scanning/755


packages/kit-bg/src/services/ServiceBootstrap.ts
```diff
@@ -225,7 +225,7 @@ export default class ServiceBootstrap extends ServiceBase {
       const onekeyRpc = Object.values(getPresetNetworks())
         .find((item) => item.id === n.id)
         ?.rpcURLs?.find((rpc) =>
-          rpc.url?.startsWith('https://node.onekey.so'),
```
stackia, line 228: WX20230809-202910

I can confirm this doesn't affect the result. Electron always gives full URL (with trailing slash if no extra path in the URL).

packages/desktop/src-electron/app.ts
```diff
@@ -401,12 +401,12 @@ function createMainWindow() {
...
       if (
-        url.startsWith('http://127.0.0.1:21320') ||
-        url.startsWith('http://localhost:21320')
+        url.startsWith('http://127.0.0.1:21320/') ||
+        url.startsWith('http://localhost:21320/')
       ) {
         // resolve onekey bridge CORS error
         details.requestHeaders.Origin = 'https://jssdk.onekey.so';
-      } else if (url.startsWith('https://mainnet.optimism.io')) {
+      } else if (url.startsWith('https://mainnet.optimism.io/')) {
```
stackia, line 409: WX20230809-202910

I can confirm this doesn't affect the result. Electron always gives full URL (with a trailing slash if no extra path in the URL).

packages/kit-bg/src/services/ServiceBootstrap.ts
```diff
@@ -225,7 +225,7 @@ export default class ServiceBootstrap extends ServiceBase {
...
+          rpc.url?.startsWith('https://node.onekey.so/'),
```
stackia, line 228: https://github.com/OneKeyHQ/app-monorepo/blob/1dee0fea26b01225936e74c9800e6847f196ca2d/packages/shared/src/config/presetNetworks.ts

There is always a trailing slash after `node.onkey.so` in our presets so it's safe to change this check to `.startsWith('https://node.onekey.so/')`

huhuanming (approved):

revan-zhang (approved):

kwoktung (approved):

PR #3634, opened 2023-10-12
OK-23687: Add lottieView to components

hellohublot:


packages/components/src/LottieView/index.tsx
```diff
@@ -0,0 +1,84 @@
...
+        handleStateChange,
+      );
+      return () => {
+        subscription.remove();
+      };
+    }, []);
+
+    useEffect(() => {
+      // animation won't work in navigate(), needs delay here
+      setTimeout(() => {
```
huhuanming: could you please provide an example of using setTimeout for delayed loading?

packages/kit/src/views/Components/stories/LottieView.tsx
```diff
@@ -0,0 +1,105 @@
...
+        '使用 Toast 显示简约明确的信息反馈',
+        '用户点击或触摸 Toast 内容时，浮层将会停留在页面上',
+        'Toast 显示的文本应少于 20 字',
+        '不建议使用 Toast 显示过长的报错信息',
+      ]}
+      boundaryConditions={[
+        'Toast 永远拥有最高层级的浮层',
+        'Toast 组件能显示的最长文本内容为三排，超出三排将会缩略',
+        '界面中只会存在一个 Toast 示例，后触发的 Toast 信息会覆盖前一条 Toast 信息',
+      ]}
```
huhuanming: description & suggestions &  boundaryConditions  can be an empty string or array

huhuanming (changes requested):

packages/components/src/LottieView/index.tsx
hellohublot: @huhuanming 

#### Where did this piece of code originally come from:
- [Diff](https://github.com/OneKeyHQ/app-monorepo/pull/624/commits/6ff731bb1af274827952c98931243f20c43309a9)

#### How to reproduce on iOS:
- You need to place the `LottieView` in the first page of the `TabNavigator`, there is an example:
[Example Link](https://github.com/hellohublot/app-monorepo/tree/lottieViewIOSAutoPlay)

#### Solution:
- In the old version, `autoPlay` was implemented by calling the `play` method in `componentDidUpdate`. Upgrading the `lottie-react-native` version to 6.3.0 can resolve this issue because `autoPlay` has now been moved to the native side.
https://github.com/lottie-react-native/lottie-react-native/pull/986

#### Known issues:
- We can't only pass  `width` to `LottieView`, and we must provide both `width` and `height` now.

Finally, I have already upgraded `lottie-react-native` to the latest version in this PR.

huhuanming (approved):

PR #3722, opened 2023-10-26
chore: ignore build files

huhuanming:


originalix (changes requested):
/packages/desktop/public/static/js-sdk

packages/shared/src/web/index.html

originalix (approved):

PR #3791, opened 2023-11-10
chore: upgrade yarn & set minimum nodejs to v18

stackia:
1. Upgrade yarn to v3.6.4.
2. Set the minimum required nodejs version to v18 (node v18 is now the LTS version). Recent released yarn v4 requires node v18.


huhuanming:
Based Node v18, Have you tried running it on all platforms，such as app, web, desktop, and extensions.😳

PR #3877, opened 2023-11-28
refactor: refactor the rendering layer

huhuanming:
Re-layer the components of the rendering layer.

- Optimization Component: Performance-optimized react native component. For more information, you can read the content [here](https://github.com/marklawlor/react-native-optimizer). optimization components will not be exposed for use by the _**kit**_ layer.
- Primitive Component: primitive components are the basic components of the OneKey component. primitive components have no dependencies on each other, just like basic atoms；
- Layout Component:  layout components is composed of atomic components and is used for the layout of component positions；
- Content Component: content components serves as a container for content display, composed of layout components.
- Action Component: action components is used to handle user interaction responses. It is more like the content components and layout Components with added events.
- Form Component: form Components is used to handle form data. Its composition is the most complex, and the sources of its components are the most diverse.

image

As shown in the figure, modules can only establish references from top to bottom. If the lower layer references the upper layer module, it will directly report a circular reference error.

**Correct Practice 🙆**
```javascript
// packages/components/src/actions/ActionList/index.tsx 

import { Divider } from '../../content';
import { Portal } from '../../hocs';
import { ButtonFrame, Icon, Text, YStack } from '../../primitives';
```

**Incorrect Practice 🙅**
```javascript
// packages/components/src/layouts/ListView/index.tsx 
//
// don't import modules from action components in layout components.
// If you insist on doing this, the linter tool will report a circular reference error.
import { Button } from '../../actions';
```

Based on the above ideas, after reorganizing the directory structure, the directory becomes simple and clean as shown in the figure.

image

So what would the export interface be like?

Don't worry, it's still `@onekeyhq/components`. There will be no changes for external use.
Other export interfaces will be temporarily retained, but in the next refactoring, not only will other export methods be cleaned up, but also obsolete hooks will be cleaned up.


revan-zhang (approved):

packages/components/src/locale/index.ts
```diff
@@ -58,9 +58,8 @@ const LOCALES_OPTION = Object.keys(LOCALES).map((key) => ({
 }));
 
 if (platformEnv.isExtensionBackground) {
-  // debugger;
-  // throw new Error('components/locale is not allowed imported from background');
+  debugger;
```
revan-zhang, line 61: debugger

huhuanming: 😳 故意设置的。
语言包进入 ext 的 kit-bg 时会报错断点。

originalix (approved):

hellohublot (approved):

PR #4437, opened 2024-04-11
fix: swap provider opt ok-26603

ezailWang:


packages/kit/src/views/Swap/hooks/useSwapTokens.ts
```diff
@@ -69,6 +71,14 @@ export function useSwapNetworkList() {
     { watchLoading: true },
   );
 
+  usePromiseResult(async () => {
```
sidmorizon: useEffect

ezailWang: 1

originalix (dismissed):

huhuanming (approved):

kwoktung (approved):

originalix (approved):

PR #6915, opened 2025-03-20
Fix/wallet issues OK-36675 OK-36659 OK-36638

weatherstar:
## Summary by CodeRabbit

- **New Features**
  - Enhanced swap transaction handling with improved processing of swap details, ensuring the correct recipient address and better display of transaction information.
  - Introduced support for a new swap provider option, expanding available swap functionalities.
  
- **Bug Fixes**
  - Improved Bitcoin network selection by using optimized network data retrieval.
  - Adjusted navigation flows during token buying and other wallet actions so users remain on the current screen instead of automatically returning to the previous view.


packages/kit-bg/src/vaults/base/VaultContext.ts
```diff
@@ -65,6 +65,10 @@ export class VaultContext {
     return networkUtils.getNetworkImpl({ networkId: this.networkId });
   }
 
+  async getNetworkChainCode() {
```
originalix: 用 getCoreApiNetworkInfo 里的 networkChainCode 就可以？

huhuanming (dismissed):

packages/kit-bg/src/vaults/base/VaultContext.ts
weatherstar: 确实

originalix (approved):

huhuanming (approved):

revan-zhang (approved):

PR #7069, opened 2025-04-10
fix: do not show toast when message is empty

huhuanming:


packages/shared/src/request/axiosInterceptor.ts
```diff
@@ -99,7 +99,7 @@ axios.interceptors.response.use(
       throw new OneKeyServerApiError({
         autoToast: true,
         disableFallbackMessage: true,
-        message: data?.message || 'OneKeyServer Unknown Error',
+        message: data?.message,
```
originalix, line 102: 主分支不是又加了一个 translationId，准备怎么兼容呢

huhuanming: 那个已经删了，在另外一个分支上，还没提

huhuanming: translationId 讨论过了，不保留，删除掉

originalix (approved):

limichange (approved):

ozie777 (commented):
Too much error margin why not simplify too much traffic slows down response and projects uncertain errors )'(false positives)$$$$

packages/shared/src/request/axiosInterceptor.ts
```diff
@@ -99,7 +99,7 @@ axios.interceptors.response.use(
       throw new OneKeyServerApiError({
         autoToast: true,
         disableFallbackMessage: true,
-        message: data?.message || 'OneKeyServer Unknown Error',
```
ozie777, line 102: Why!

PR #7090, opened 2025-04-12
fix: fixed typo of i18n key

imyuanx:
## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved a typographical error in the messaging for the transaction history view. With this update, users will now see the correct text when certain filter options for scam transactions are not supported.


originalix (approved):

originalix:
Thanks for the careful fix! I’ve also updated the corresponding key on Lokalise. 👍

limichange (approved):

PR #10610, opened 2026-03-11
fix: Dialog focus trap, keyboard avoidance, iOS navigation race condition, and bundle list improvements(OK-51663)

huhuanming:
## Summary
- **Dialog focus trap**: Enable conditional focus trap for Sheet path and fix lock overlay inert issue
- **Dialog keyboard avoidance**: Move keyboard avoidance from Footer to Dialog level so all dialogs (including `showFooter: false`) adapt when keyboard appears. Fixes password setup dialog and market swap dialog
- **iOS navigation race condition**: Add atomic `resetToRoute()` to replace `resetAboveMainRoute()` + `navigate()` pattern. The old pattern caused a race condition where iOS native modal dismiss animation would pop the newly navigated route ~1s after it appeared, leaving users stuck on the home page. Applied across all affected navigation flows (onboarding, swap, perp, firmware update, market, prime)
- **BundleList ciBundleVersion**: Switch from `bundleVersion` to server's new `ciBundleVersion` field, redesign bundle item layout with commitHash as primary identifier, add base36 hash display
- **LocalBundleList**: Align layout with remote bundles style, fix `fallbackBundles.map` crash on desktop

## Test plan
- [ ] Open password setup dialog on iOS/Android, verify keyboard pushes dialog up
- [ ] Open market swap dialog, verify keyboard avoidance works
- [ ] Verify dialogs with footer still work correctly (e.g. confirm dialogs)
- [ ] Verify Dialog bottom has no extra blank space when keyboard is hidden
- [ ] On iOS, create a new unbackedup HD wallet → verify BackupWalletReminder page navigates correctly without freezing on home page
- [ ] On iOS, trigger firmware update flow → verify navigation proceeds without getting stuck
- [ ] On iOS, trigger swap/perp onboarding flow → verify navigation works
- [ ] Check JS Bundle Manager → Remote Bundle Switcher, verify new layout with commitHash + branch + ciBundleVersion
- [ ] Check JS Bundle Manager → Local Bundles page renders without crash
- [ ] Verify Bundle Hash row appears in Runtime Info section


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/34cfeeb3-4366-4c38-9436-e8dbf596ac81?refs=0303b6138809e00ba6d0dcca5e8681ac8b25c972&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/34cfeeb3-4366-4c38-9436-e8dbf596ac81/license?refs=0303b6138809e00ba6d0dcca5e8681ac8b25c972&source=prChecksComment) |

packages/kit/src/components/Password/container/PasswordVerifyPromptMount.tsx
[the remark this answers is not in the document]
huhuanming: Good catch! The `trapFocus` should indeed be added to the verify dialog as well. However, it also needs two additional conditions:

1. Only enable on non-native platforms (`!platformEnv.isNative`) — focus trapping is only needed for desktop/web/extension keyboard navigation
2. Disable during lock screen — when the app is locked, focus trapping can interfere with the lock screen UI

Will update the PR to address both points.

packages/kit/src/provider/Container/AppStateLockContainer/index.tsx
```diff
@@ -91,6 +91,30 @@ export function AppStateLockContainer({
...
+  useEffect(() => {
+    if (platformEnv.isNative || !isLocked) return;
+
+    const lockEl = lockContainerRef.current;
+    if (!lockEl) return;
+
+    const inertElements: HTMLElement[] = [];
+    for (const child of Array.from(document.body.children)) {
+      if (child instanceof HTMLElement && !child.contains(lockEl)) {
+        child.inert = true;
```
originalix, line 106: 🔴 **[P1] Lock-screen inert logic also disables the lock overlay portal**

This loop marks every `document.body` child except the lock screen itself as `inert`, but the lock flow's own `APP_STATE_LOCK_CONTAINER_OVERLAY` is rendered as a sibling under `document.body` via `AppStateContainer`. That means any dialog mounted into that portal while locked, such as the biometric-changed warning in `PasswordVerify`, also becomes inert and can no longer receive focus or interaction.

packages/components/src/composite/Dialog/index.tsx
```diff
@@ -330,6 +332,7 @@ function DialogFrame({
             <TMDialog.Title display="none" />
             <TMDialog.Content
               elevate
+              trapFocus={effectiveTrapFocus}
```
originalix, line 366: 🟡 **[P2] `trapFocus` 在 Sheet 渲染路径上被忽略了**

这里虽然计算了 `effectiveTrapFocus`，但只把它传给了 `TMDialog.Content`。当 `media.md` 为 `true` 时，组件会走上面的 `Sheet` 分支，这个新的公开 `trapFocus` API 就会变成和断点/设备相关的行为，对 sheet-backed dialog 实际上不生效。如果这个能力是想做成跨平台一致的 Dialog 能力，就需要在 Sheet 路径上补齐等价的焦点控制逻辑。

huhuanming: Fixed in bfd452270a. Wrapped `Sheet.Frame` content with `<FocusScope trapped={effectiveTrapFocus} loop>` from `@tamagui/focus-scope` — the same underlying component that `TMDialog.Content` uses for its `trapFocus` prop.

On native, `FocusScope` is a no-op (just returns children), so zero runtime cost. On web it provides real DOM-based focus trapping via `focusin`/`focusout` listeners.

packages/kit/src/provider/Container/AppStateLockContainer/index.tsx
huhuanming: Good catch! Fixed in bfd452270a. 

Root cause: `AppStateContainer` used `createPortal(<>...</>, document.body)` — the Fragment caused the lock screen and `APP_STATE_LOCK_CONTAINER_OVERLAY` portal to become **separate** `document.body` children. The inert loop only skips children that contain `lockEl`, so the overlay portal got marked `inert`.

Fix: replaced `<>` with `<div>` so both the lock screen and the overlay portal share the same `document.body` child element. Now `child.contains(lockEl)` returns true for the wrapper div, keeping both the lock screen and its overlay portal interactive.

ezailWang (approved):

originalix (approved):

PR #11265, opened 2026-04-16
fix(perps): silence perp-config fetch errors and bump fallback builder fee(OK-53395)

Minnzen:
## Summary
- Wrap each `pRetry` attempt of `updatePerpsConfigByServer` in a try/catch that marks the error `autoToast: false` before rethrowing, so retries still happen but the per-attempt timeouts no longer fan out into a stack of user-facing error toasts.
- Remove the stale `// TODO: Change maxAge back to { hour: 1 } before production release` comment on `_updatePerpsConfigByServerWithCache`; per product direction we are intentionally staying on `maxAge: 0` and relying on the persistent SimpleDb layer as the actual cache.
- Bump `FALLBACK_MAX_BUILDER_FEE` from 40 to 50 (i.e. 0.04% → 0.05%) in both `shared/src/consts/perp.ts` and `shared/types/hyperliquid/perp.constants.ts`.

## Intent & Context
QA reported that opening the Perps Portfolio would surface a red-triangle toast "timeout of 30000ms exceeded", sometimes twice in a row, even though the main portfolio UI (chart, balances, stats) rendered fine. The user asked us to trace the source and — once the product direction was clear — silence the retry noise without losing the retry behaviour itself. Separately, product asked to bump the fallback builder fee rate that is used when the server config has not been (or cannot be) fetched.

## Root Cause
1. `Bootstrap.useFetchPerpConfig` schedules `pRetry` with `PERPS_CONFIG_FETCH_MAX_RETRIES = 3` against the OneKey BFF endpoint `/utility/v1/perp-config`. This endpoint goes through `axios`, which the global interceptor caps at 30s (`axiosInterceptor.ts:39`), producing the signature axios error string `timeout of 30000ms exceeded` on timeout.
2. The outer `.catch(noop)` only swallows the final `pRetry` rejection. Each individual attempt that rejects still propagates through the global error pipeline (`BackgroundApiProxyBase` → `globalErrorHandler` → `ErrorToastContainer`) and fires a toast — hence the user seeing 2+ toasts when BFF was slow.
3. Main portfolio data goes through the Hyperliquid SDK (fetch + AbortSignal, default 10s), not axios, which is why the UI data was fine and only this side-effect fetch was noisy.

## Design Decisions
- **Keep the retry loop, just silence the per-attempt toasts.** Product wants retries preserved so a transient slow response still eventually repopulates the cache; they only objected to the toast spam. Wrapping the attempt body and calling `errorToastUtils.toastIfErrorDisable(err)` (the codebase's existing "mark this error as non-toast" utility) is the minimally invasive way — no changes to global interceptor, no changes to shared axios config, no change in retry cadence.
- **Do not introduce a memoizee TTL.** We initially considered restoring `maxAge: { hour: 1 }` per the TODO. After discussing with product, the desired behaviour is a plain stale-while-revalidate: SimpleDb is the persistent cache, consumers read it synchronously, and fire-and-forget background fetches refresh it. A separate in-memory memoizee TTL is unnecessary and could mask real network status, so we explicitly stay at `maxAge: 0` and delete the misleading TODO.
- **Do not push a 3s "network-first with cache fallback" pattern.** Considered but rejected: all current consumers are already fire-and-forget against SimpleDb, so SWR is effectively already in place; switching to network-first would add latency to the common (fast network) case without a concrete consumer that actually needs freshness guarantees for this payload.
- **Builder fee bump is an isolated product-driven constant change.** No logic changes around how the fallback is used; both files that export the constant are updated so the `ServiceHyperliquid.init` fallback path and the `perp.constants.ts` consumer stay in lock-step.

## Changes Detail
- `packages/kit/src/provider/Bootstrap.tsx`: import `errorToastUtils`; convert the `pRetry` attempt callback from sync-return to `async` with a try/catch that calls `errorToastUtils.toastIfErrorDisable(err)` before rethrowing so `pRetry` continues to retry while the global toast pipeline treats each attempt as non-toasting.
- `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts`: remove the outdated TODO comment above `_updatePerpsConfigByServerWithCache`. Behaviour unchanged (still `maxAge: 0`).
- `packages/shared/src/consts/perp.ts` and `packages/shared/types/hyperliquid/perp.constants.ts`: `FALLBACK_MAX_BUILDER_FEE` 40 → 50.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile / Desktop / Web / Extension (all platforms hit the same Bootstrap path and share the constants)
- **Risk Areas**:
  - `pRetry` attempt is now `async` and awaits each underlying promise inside the try/catch. Retry count, delay, and outer `.catch(noop)` are unchanged, so retry semantics are identical apart from the toast suppression.
  - The fallback builder fee bump only affects the in-memory default in `ServiceHyperliquid` and the `FALLBACK_MAX_BUILDER_FEE` import in `perp.constants.ts`. It does not touch any user-specific approval flow (`getUserApprovedMaxBuilderFee` still reads real values from Hyperliquid), and old users already have a `hyperliquidMaxBuilderFee` persisted in SimpleDb that takes precedence over the fallback.

## Test plan
- [ ] With network disabled (simulate BFF timeout via devtools throttle or blocking the domain), open Perps Portfolio: confirm no `timeout of 30000ms exceeded` toast appears, and confirm the UI still renders using cached / fallback data.
- [ ] Restore network, reopen Perps: confirm the config eventually updates (SimpleDb gets refreshed, consumers pick up new values on next mount).
- [ ] Fresh install / clear storage with BFF unreachable: confirm builder fee flow uses the 50 fallback (0.05%) and the app does not crash or show toast spam.
- [ ] Existing user with persisted `hyperliquidMaxBuilderFee`: confirm the SimpleDb-persisted value is still honoured (fallback is only consumed when the persisted value is absent).
- [ ] Regression: other toasts (network 403/429/5xx from other endpoints) still appear as before, since only this specific retry path is silenced.

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11265" target="_blank">
  <picture>
    Open with Devin
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3b15286e-78c1-4078-90b8-9106fa237bbc?refs=26d847aef71358a95b29adb4d2d1afe3e6aa7def&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3b15286e-78c1-4078-90b8-9106fa237bbc/license?refs=26d847aef71358a95b29adb4d2d1afe3e6aa7def&source=prChecksComment) |

originalix (approved):

PR #11357, opened 2026-04-24
perf(defi): force-refresh positions at +40s/+80s after local tx confirms (OK-53474)

weatherstar:
## Summary
- Schedule two force refreshes of the DeFi portfolio for the affected chain at +40s and +80s after a locally-submitted tx transitions pending → confirmed, so Wallet Home DeFi Tab reflects post-tx positions without waiting for the 60s poll.
- Only triggers on DeFi-enabled chains and `EDecodedTxStatus.Confirmed`; multiple txs on the same `(accountId, networkId)` coalesce to the latest schedule.
- Works in both single-network and All Networks modes; background scheduler so the fresh result lands in simpleDb even if the user has already navigated away.

## Intent & Context
Product reported that after users perform DeFi operations (stake / unstake / claim / swap) from inside the OneKey app, the Wallet Home DeFi Tab's Position data stays stale until the next 60s natural polling tick. We need a way to surface post-tx freshness quickly without aggressively shortening the poll interval.

The PM's proposal was to force refresh at +40s and +80s — empirical offsets that cover the common range of DeFi indexer lag across supported protocols (Aave, Lido, Compound, etc.) while strictly bounding the extra server load (at most two additional requests per DeFi tx per chain, coalesced across multiple rapid txs).

## Root Cause (for the stale-data problem)
The existing DeFi polling path in `DeFiListBlock` uses `usePromiseResult` with `pollingInterval: POLLING_INTERVAL_FOR_DEFI` (60s) and debounce 1s. Outside of this cadence, there is no reactive path from tx confirmation → DeFi refresh. The existing `AccountDataUpdate` event fires on pull-to-refresh but not on confirmation.

## Design Decisions
- **Event plumbing**: `ServiceHistory.fetchAccountHistory`'s pending→confirmed detection loop now emits `LocalPendingTxConfirmed` per tx. This loop only iterates locally-pending txs (those persisted by `saveSendConfirmHistoryTxs`), so it naturally filters for "tx submitted via this app" without an explicit origin check.
- **Indexed-account resolution at emit time**: The payload carries `indexedAccountId` resolved via `serviceAccount.getDBAccountSafe`, so All Networks mode subscribers can match reliably rather than parsing a walletId out of the raw accountId.
- **Background scheduling in ServiceDeFi**: The 40s/80s timers live in `ServiceDeFi`, not in `DeFiListBlock`. This means the force refresh writes to simpleDb regardless of whether the DeFi Tab is currently mounted — the user sees fresh data the next time they open the tab.
- **Coalescing key**: `(accountId, networkId)`. New txs reset the schedule to the latest confirmation time. Prevents timer stacking when users chain multiple operations.
- **"DeFi-enabled chain" filter**: Uses `getDeFiEnabledNetworksMap()` (the chain-level capability flag) rather than "has current positions > 0". This covers the first-stake case where the user has no positions on that chain until after the tx.
- **Abort pool separation**: Added `IFetchAccountDeFiPositionsParams.abortable` (default `true`, backward-compatible). The scheduler's force refresh passes `abortable: false` so a UI-initiated `abortFetchAccountDeFiPositions()` on account/tab switch does not cancel the scheduled refresh, which is what delivers the post-tx freshness.
- **All Networks partial merge**: Instead of triggering a full all-networks re-run, the subscriber splices the refreshed network's protocols into the current list and re-aggregates the overview totals from `protocolMap`, keeping the header number consistent.

## Changes Detail
- `packages/shared/src/eventBus/appEventBusNames.ts` — adds `LocalPendingTxConfirmed` and `DeFiPositionRefreshed` enum entries.
- `packages/shared/src/eventBus/appEventBus.ts` — payload types for both events; overview uses `IFetchAccountDeFiPositionsResp['data']['totals']` to stay in sync with the server contract.
- `packages/shared/types/defi.ts` — adds `abortable?: boolean` to `IFetchAccountDeFiPositionsParams`.
- `packages/kit-bg/src/services/ServiceHistory.ts` — emits `LocalPendingTxConfirmed` per confirmed tx with resolved `indexedAccountId`.
- `packages/kit-bg/src/services/ServiceDeFi.ts` — subscribes to `LocalPendingTxConfirmed` in the constructor; schedules timers per `(accountId, networkId)` with coalescing and post-fire Map cleanup; `_runDeFiForceRefresh` calls `fetchAccountDeFiPositions({isForceRefresh: true, saveToLocal: true, abortable: false})` and emits `DeFiPositionRefreshed`.
- `packages/kit/src/views/Home/components/DeFiListBlock/DeFiListBlock.tsx` — subscribes to `DeFiPositionRefreshed`; single-network mode writes atoms from payload; All Networks mode splices the refreshed network's protocols/protocolMap in place and re-aggregates overview via BigNumber. Subscription deps stabilized via refs so it does not rebuild on every atom update.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (all Home Wallet DeFi Tab consumers)
- **Risk Areas**:
  - **Mobile background timers** — `setTimeout` accuracy degrades when iOS/Android backgrounds the app. Acceptable for this feature; if the user locks the phone mid-window, they will see fresh data on next resume via simpleDb + the natural polling cycle. Listed as P2 follow-up in planning doc.
  - **ServiceDeFi lazy instantiation** — the constructor-level subscription only activates after the service is first accessed. In practice, `DeFiListBlock` triggers this within the first render of Home. Edge case: if a tx confirms before ServiceDeFi is ever touched (deeplink flows), the event is missed; 60s natural polling still corrects.
  - **Extra server load** — bounded to 2 extra `/wallet/v1/portfolio/positions?isForceRefresh=true` calls per DeFi tx per chain, coalesced across rapid txs.
  - **Non-DeFi txs still emit** `LocalPendingTxConfirmed` — scheduler filters via `isNetworkDeFiEnabled`; extra events, no wasted network calls.

## Test plan
- [ ] Stake a small amount on Ethereum → verify `isForceRefresh=true` requests hit `/wallet/v1/portfolio/positions` at ~40s and ~80s post-confirm; new position appears in the DeFi Tab within 80s
- [ ] Unstake / claim rewards → same as above
- [ ] Send two stake txs within 30s → verify only one pair of timers fires (i.e. the first pair is cancelled; add temporary log to `_deFiForceRefreshTimers.size` if needed)
- [ ] All Networks mode: stake Ethereum → only the single-chain request flies (no fan-out to 20 enabled networks); header total updates, other chains' positions remain untouched
- [ ] Send DeFi tx → switch away from DeFi Tab immediately → return after 90s → see updated positions (simpleDb write path)
- [ ] Failed tx (insufficient gas) → no force-refresh request fires
- [ ] Non-DeFi chain tx → no force-refresh request fires
- [ ] Rapid account switch during the 40-80s window → verify scheduler's in-flight request is NOT cancelled by the UI's `abortFetchAccountDeFiPositions` (the new `abortable: false` protection)

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11357" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/kit-bg/src/services/ServiceDeFi.ts
+++ b/packages/kit-bg/src/services/ServiceDeFi.ts
@@ -352,9 +352,17 @@ class ServiceDeFi extends ServiceBase {
   }) {
     const { accountId, indexedAccountId, networkId } = params;
     try {
-      const [settings, currencyMap] = await Promise.all([
+      const [settings, currencyMap, accountAddress, xpub] = await Promise.all([
         settingsPersistAtom.get(),
         this.backgroundApi.serviceSetting.getCurrencyMap(),
+        this.backgroundApi.serviceAccount.getAccountAddressForApi({
+          accountId,
+          networkId,
+        }),
+        this.backgroundApi.serviceAccount.getAccountXpub({
+          accountId,
+          networkId,
+        }),
       ]);
       const sourceCurrencyInfo = currencyMap[settings.currencyInfo.id];
       const targetCurrencyInfo = currencyMap.usd;
@@ -362,16 +370,58 @@ class ServiceDeFi extends ServiceBase {
       const resp = await this.fetchAccountDeFiPositions({
         accountId,
         networkId,
+        accountAddress,
+        xpub,
         excludeLowValueProtocols: true,
         sourceCurrencyInfo,
         targetCurrencyInfo,
-        saveToLocal: true,
+        // Do NOT use saveToLocal here. The shared `_localDeFiOverviewCache`
+        // is keyed only by networkId and the debounced flush writes against
+        // the last `accountAddress/xpub` it sees, so concurrent background
+        // refreshes for different accounts on the same network — or a
+        // background refresh racing with a foreground UI fetch — would
+        // overwrite each other's per-account local overview. Write this
+        // single account's overview directly below instead.
+        saveToLocal: false,
         isForceRefresh: true,
         // Do not share the abort pool: a UI-initiated
         // abortFetchAccountDeFiPositions() must not cancel the scheduled
         // force refresh, which is what delivers the post-tx freshness.
         abortable: false,
       });
+
+      if (accountAddress || xpub) {
+        await this.updateAccountsLocalDeFiOverview({
+          accountAddress,
+          xpub,
+          overview: {
+            [networkId]: {
+              totalValue: this._fixCurrencyValue({
+                sourceCurrencyInfo,
+                targetCurrencyInfo,
+                value: resp.overview.totalValue,
+              }).toNumber(),
+              totalDebt: this._fixCurrencyValue({
+                sourceCurrencyInfo,
+                targetCurrencyInfo,
+                value: resp.overview.totalDebt,
+              }).toNumber(),
+              totalReward: this._fixCurrencyValue({
+                sourceCurrencyInfo,
+                targetCurrencyInfo,
+                value: resp.overview.totalReward,
+              }).toNumber(),
+              netWorth: this._fixCurrencyValue({
+                sourceCurrencyInfo,
+                targetCurrencyInfo,
+                value: resp.overview.netWorth,
+              }).toNumber(),
+              currency: targetCurrencyInfo?.id ?? '',
+            },
+          },
+          merge: true,
+        });
+      }
 
       appEventBus.emit(EAppEventBusNames.DeFiPositionRefreshed, {
         accountId,
--- a/packages/kit-bg/src/services/ServiceDeFi.ts
+++ b/packages/kit-bg/src/services/ServiceDeFi.ts
@@ -2,11 +2,16 @@ import BigNumber from 'bignumber.js';
 import { debounce, isEmpty, isUndefined } from 'lodash';
 
 import type { ICurrencyItem } from '@onekeyhq/kit/src/views/Setting/pages/Currency';
+import { settingsPersistAtom } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
 import {
   backgroundClass,
   backgroundMethod,
 } from '@onekeyhq/shared/src/background/backgroundDecorators';
 import { getNetworkIdsMap } from '@onekeyhq/shared/src/config/networkIds';
+import {
+  EAppEventBusNames,
+  appEventBus,
+} from '@onekeyhq/shared/src/eventBus/appEventBus';
 import accountUtils from '@onekeyhq/shared/src/utils/accountUtils';
 import defiUtils from '@onekeyhq/shared/src/utils/defiUtils';
 import networkUtils from '@onekeyhq/shared/src/utils/networkUtils';
@@ -15,6 +20,7 @@ import type {
   IFetchAccountDeFiPositionsResp,
 } from '@onekeyhq/shared/types/defi';
 import { EServiceEndpointEnum } from '@onekeyhq/shared/types/endpoint';
+import { EDecodedTxStatus } from '@onekeyhq/shared/types/tx';
 
 import ServiceBase from './ServiceBase';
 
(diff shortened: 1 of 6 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/498a69c8-7273-4a80-af5a-08f4eb1252fa?refs=e6f5c5904998096cdaa33a6c4aa52fe718478455&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/498a69c8-7273-4a80-af5a-08f4eb1252fa/license?refs=e6f5c5904998096cdaa33a6c4aa52fe718478455&source=prChecksComment) |

packages/kit-bg/src/services/ServiceDeFi.ts
```diff
@@ -266,6 +289,108 @@ class ServiceDeFi extends ServiceBase {
...
+      const sourceCurrencyInfo = currencyMap[settings.currencyInfo.id];
+      const targetCurrencyInfo = currencyMap.usd;
+
+      const resp = await this.fetchAccountDeFiPositions({
+        accountId,
+        networkId,
+        excludeLowValueProtocols: true,
+        sourceCurrencyInfo,
+        targetCurrencyInfo,
+        saveToLocal: true,
```
originalix: 这里的后台 force refresh 不能直接复用 `saveToLocal: true` 的共享 debounced 缓存路径。`_localDeFiOverviewCache` 只按 `networkId` 聚合，flush 时又只使用最后一次 `accountAddress/xpub`，所以多个账户的后台刷新互相重叠，或后台刷新与前台 DeFi 拉取重叠时，可能把 A 账户 overview 写到 B 账户 local key。建议按 account local assets key 隔离累加器，或让 force refresh 使用独立的单账户直接写入路径。

packages/kit/src/views/Home/components/DeFiListBlock/DeFiListBlock.tsx
```diff
@@ -828,6 +836,119 @@ function DeFiListBlock({
...
+      if (currentIndexedId && payloadIndexedId) {
+        if (currentIndexedId !== payloadIndexedId) return;
+      } else {
+        const { walletId: payloadWalletId } = accountUtils.parseAccountId({
+          accountId: payload.accountId,
+        });
+        const { walletId: currentWalletId } = accountUtils.parseAccountId({
+          accountId: account.id,
+        });
+        if (!payloadWalletId || payloadWalletId !== currentWalletId) return;
```
originalix: 这个 fallback 只比较 `walletId` 对 All Networks 不够安全。imported/watching/external 账户会共享 `imported`/`watching`/`external` 这类 walletId，真实账户身份在 accountId 后缀里；当前分支后面没有再做精确 accountId 校验，可能把另一个同类账户的 `DeFiPositionRefreshed` 拼进当前 All Networks 视图。缺少 `indexedAccountId` 时建议至少要求 `payload.accountId === account.id`，或比较明确的 others/account local assets key。

packages/kit-bg/src/services/ServiceDeFi.ts
weatherstar: Good catch — fixed in 74e64225f1. The force-refresh path no longer uses `saveToLocal: true` (which routed through the network-keyed shared `_localDeFiOverviewCache` and the debounced flush that uses the last `accountAddress/xpub`). `_runDeFiForceRefresh` now resolves the account `accountAddress`/`xpub` up front and writes that single account+network overview directly via `updateAccountsLocalDeFiOverview` after the fetch returns, so concurrent background refreshes for different accounts (or a background refresh racing with a foreground fetch) cannot cross-contaminate each other.

packages/kit/src/views/Home/components/DeFiListBlock/DeFiListBlock.tsx
weatherstar: Agreed — fixed in 74e64225f1. The fallback (used when either side lacks an `indexedAccountId`) now requires a strict `payload.accountId === account.id` match instead of comparing wallet buckets. This eliminates the imported/watching/external case where multiple distinct accounts share a single `imported` / `watching` / `external` walletId and could otherwise leak each other's `DeFiPositionRefreshed` into the current All Networks view.

zhaono1 (approved):

originalix (approved):

PR #11368, opened 2026-04-26
fix: normalize fiat send amount precision (OK-53784)

weatherstar:
## Summary
- Floor the fiat→token conversion in `linkedAmount` to the chain's native precision so the displayed and submitted amount is always representable.
- Apply the same flooring in `handleValidateTokenAmount` so validation operates on the exact value that gets submitted to the vault.
- Reject positive fiat inputs that floor to a zero token amount, preventing 0-amount transfers that would waste gas.

## Intent & Context
In fiat send mode the original token amount was computed as `fiat / price` and used both for display and for submission without rounding. For tokens whose `decimals` is smaller than the precision of the division result (e.g. `decimals=2` or `decimals=0` ERC-20s), the value passed to the vault carried more decimal places than the chain could represent, leading to confirm-screen / vault rejection or unexpected dust. Lightning already had a `ROUND_FLOOR` for sats; non-Lightning chains did not.

## Root Cause
`packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx` had three coupled issues in the fiat path:

1. `linkedAmount` did not floor `fiat / price` to `tokenDetails.info.decimals` for non-Lightning chains.
2. `handleValidateTokenAmount` computed `tokenAmountBN = amount / price` *unfloored*, while submit used the floored value from `linkedAmount.originalAmount`. Min/balance checks ran on a different value than what was actually sent.
3. After flooring, a positive fiat input small enough to floor to `0` (sub-sat on Lightning, sub-decimal on a 0-decimal ERC-20) slipped past the min-amount check (which excludes `isZero`) and the native-only zero guard, and the submit-disabled predicate only checks the raw fiat string. The user could send a real on-chain `0`-amount transfer and pay gas for nothing.

## Design Decisions
- **Extract a single helper `floorFiatDerivedTokenAmount`** at module scope rather than duplicating the floor logic in two places. Makes the contract — "this is the precision the chain can actually send" — explicit and prevents the two call sites from drifting again.
- **Lightning unit conversion stays at the call site**, not inside the helper. Validation needs the value in sats; display needs it in BTC when `lnUnit === BTC`. Both share the floor; only the call site decides whether to convert.
- **Conservative zero-amount guard** that only fires in fiat mode when the user provided a positive fiat input but the floored token result is `0`. We deliberately do not broaden the existing native-only zero guard to all tokens, since some chains/contracts have legitimate uses for 0-amount token transfers — the fiat case is the only one regressed by flooring.
- **`send_amount_too_small` reuses an existing translation key** (\"Transfer amount too small.\"), so no i18n churn.
- Cross-validated with Codex on both the validation/submit alignment and the zero-floor edge case.

## Changes Detail
`packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx`:
- Add module-scope `floorFiatDerivedTokenAmount({ amount, isLightningNetwork, decimals })` that returns the input floored to integer sats for Lightning, or to `decimals` places for other chains; returns the input unchanged when `decimals` is missing or invalid.
- `linkedAmount` useMemo: route the fiat→token result through the helper, then convert sats→BTC if `lnUnit === BTC`. Add `tokenDetails?.info.decimals` to the deps array.
- `handleValidateTokenAmount` useCallback: route `tokenAmountBN` through the same helper in fiat mode. Add `tokenDetails?.info.decimals` to the deps array. Insert a `send_amount_too_small` guard for the "positive fiat → 0 token" case.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: All (Send flow, every chain that uses fiat-mode input)
- **Risk Areas**:
  - Behavior change in the fiat send path on every non-Lightning chain. Token-mode input is untouched.
  - Three reviews (primary + Codex twice) confirm validation and submit now compute identical values across all four scenarios (Lightning sats / Lightning BTC / non-Lightning fiat / non-Lightning token).
  - The new "amount too small" check is gated on `isUseFiat && priceBN.isGreaterThan(0) && tokenAmountBN.isZero() && !amountBN.isZero()`, so it cannot fire in token mode and cannot affect the existing "user typed 0" path.

## Test plan
- [ ] Non-Lightning chain (e.g. ERC-20 with `decimals=6`) in fiat mode: enter a fiat amount that produces more than 6 decimal places. Confirm `originalAmount` shown in the input and the amount on the confirm screen are both floored to 6 decimals and identical.
- [ ] Same chain, enter a fiat amount large enough to be representable but small enough that `floor(fiat/price)` is `0` for a `decimals=0` token: expect "Transfer amount too small.", submit blocked.
- [ ] Lightning + fiat + SATS unit, enter sub-sat fiat: expect "Transfer amount too small.", no zero-sat submission.
- [ ] Lightning + fiat + BTC unit, regression check for OK-53396: enter normal fiat, confirm sats are floored and BTC display is consistent.
- [ ] Toggle fiat ↔ token several times with various amounts, confirm no precision drift.
- [ ] Fiat mode with `tokenDetails.price === 0`: existing behavior preserved (early return, no new error path).
- [ ] User types `"0"` in fiat mode for a non-native token: existing behavior preserved (no new error from the zero-floor guard).

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11368" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
+++ b/packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
@@ -352,6 +352,18 @@ function SendAmountInputContainer() {
             chainValueUtils.convertSatsToBtc(originalAmt.toFixed()),
           );
         }
+      } else {
+        const decimals = tokenDetails?.info.decimals;
+        if (
+          typeof decimals === 'number' &&
+          Number.isInteger(decimals) &&
+          decimals >= 0
+        ) {
+          originalAmt = originalAmt.decimalPlaces(
+            decimals,
+            BigNumber.ROUND_FLOOR,
+          );
+        }
       }
       return {
         originalAmount: originalAmt.toFixed(),
@@ -364,7 +376,14 @@ function SendAmountInputContainer() {
       originalAmount: amountBN.toFixed(),
       linkedAmount: linkedAmountValue.toFixed(),
     };
-  }, [amount, isLightningNetwork, isUseFiat, lnUnit, tokenDetails?.price]);
+  }, [
+    amount,
+    isLightningNetwork,
+    isUseFiat,
+    lnUnit,
+    tokenDetails?.info.decimals,
+    tokenDetails?.price,
+  ]);
 
   const handleToggleFiatMode = useCallback(() => {
     // When currently in fiat mode (isUseFiat=true), switching to token mode -> use originalAmount
--- a/packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
+++ b/packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
@@ -515,8 +515,8 @@ function SendAmountInputContainer() {
       const priceBN = new BigNumber(tokenDetails?.price ?? 0);
       // Mirror the flooring applied in `linkedAmount` so the value validated
       // here matches the value submitted to the vault. Without this, fiat
-      // mode could pass min/balance checks against an unfloored amount while
-      // the user actually sends a smaller, floored value.
+      // mode could pass min/balance checks against the raw fiat/price result
+      // while the user actually sends a smaller, floored value.
       const tokenAmountBN =
         isUseFiat && priceBN.isGreaterThan(0)
           ? floorFiatDerivedTokenAmount({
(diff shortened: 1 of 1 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/16a0b94a-0748-4d3d-98e7-f217ee4e4507?refs=be07aea6c71856aefc4e6e0874bfa9c8343b4be6&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/16a0b94a-0748-4d3d-98e7-f217ee4e4507/license?refs=be07aea6c71856aefc4e6e0874bfa9c8343b4be6&source=prChecksComment) |

huhuanming (approved):

originalix (approved):

PR #11375, opened 2026-04-27
feat: fee estimate stop/resume polling on server signal(OK-53769)

weatherstar:
## Summary
- Honor `stopPolling: true` from estimate-fee error responses so failure messages stay visible without spamming retries that can't succeed.
- Resume polling correctly when the user clicks Retry and the response goes back to normal — previously the chain stayed dead.
- Suppress the redundant global toast on stopPolling errors and tighten the fee `errMessage` fallback chain so JSON-RPC inner errors aren't shadowed.

## Intent & Context
The fee estimate poller used to keep hammering the server even after the backend explicitly told us the request can't succeed (e.g. allowance shortage, expired swap quote). The user reported the same error toast/message looping every poll interval, which is noisy and burns API quota.

After landing the initial stop-polling logic, the user found a follow-up bug: clicking Retry after a `stopPolling` error fired one request and stayed silent forever, even when the response had recovered. Polling needed to resume on a successful retry, and stop again only if the next error also carried `stopPolling: true`.

## Root Cause
**Stop-side**: The server already returns `stopPolling: true` on errors that won't recover under the same input, but the client wasn't honoring it. Every tick re-asked and got the same error.

**Resume-side**: `setStopPolling(false)` only flipped the ref flag. The previous polling chain had already exited via the `finally`-block guard once `stopPollingRef` flipped `true`, and clearing the flag alone didn't bring it back. The retry callback's `void run()` didn't help either — `run()` without a `pollingNonce` never schedules the next tick (the finally-block check `pollingNonceRef.current === config?.pollingNonce` is `current === undefined` → false).

## Design Decisions
- **Ref over state for `stopPolling`**: the polling continuation in the `finally`-block reads the flag synchronously. No consumer reacts to it via React state, so a state would only churn renders without changing behavior.
- **Auto-resume inside `setStopPolling(false)`**: instead of adding a separate `resumePolling()` API, fold the resume into the setter. The semantic of `setStopPolling(false)` is "stop pausing polling" — actually resuming is what callers expect. The implementation mirrors the `runnerDeps` effect (bump nonce + fire run with new nonce), so the loop restarts and any stale queued tick from a still-pending old chain is automatically invalidated by the nonce bump.
- **Detect stopPolling AFTER gas-account classification**: the error may match a gas-account fallback strategy first — that path needs to flip the payer and retry on its own, so it must run before we decide to pause.
- **Read `errMessage` from `OneKeyServerApiError` fields**: previously we read from the JSON-RPC path that never matched server API errors, so the inline fee error UI got the wrong text. Reorder the fallback chain so source-specific fields come first; the wrapper Error's generic message is the last resort.
- **Suppress global autoToast on stopPolling errors**: the inline fee error already carries the same message and stays visible after polling stops, so a transient toast on top would be redundant. Mirrors the sponsor-fallback branches.
- **Move `setStopPolling` declaration after `runRef`**: the resume path references `runRef.current`, so the `useCallback` must be declared after `runRef` to keep oxlint happy (and to not rely on TDZ-safe closure resolution).

## Changes Detail
- `packages/shared/types/request.ts` — declare optional `stopPolling` field on `IOneKeyAPIBaseResponse`.
- `packages/kit/src/hooks/usePromiseResult.ts` — expose `setStopPolling` from the hook return; `finally`-block bails before scheduling the next tick when set; `runnerDeps` effect clears the flag automatically so input changes resume polling. Manual `setStopPolling(false)` (true → false transition) bumps the nonce and fires a run with it, so retry resumes the chain.
- `packages/kit/src/hooks/usePromiseResult.test.tsx` — add tests for `setStopPolling(true)` skipping the next tick, deps-change auto-resume, and the manual-resume path.
- `packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx` — detect `error.data.stopPolling` after gas-account fallback classification, suppress autoToast on the stopPolling branch, reorder errMessage fallbacks, and clear stopPolling in the `EstimateTxFeeRetry` callback so manual retry actually resumes the loop.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop, Mobile, Web, Extension (signature-confirm fee panel is shared across all)
- **Risk Areas**:
  - `usePromiseResult` is shared by many features. Behavior change: `setStopPolling(false)` now also fires an immediate run when transitioning from stopped → not stopped. Currently only TxFeeInfo calls it, so the blast radius is contained. The nonce-bump invalidates stale queued ticks, so no double-polling.
  - Gas-account fallback path emits the same `EstimateTxFeeRetry` event with the chain still alive. The retry callback's `void run()` is preserved so that path still gets an immediate retry instead of waiting out the polling interval. abortEstimateFee dedups the resulting overlapping requests.

## Test plan
- [ ] Trigger a fee estimate error that returns `stopPolling: true` (e.g. allowance shortage, expired swap quote). Verify the inline error message stays visible and the request stops repeating.
- [ ] Click Retry while still on the same input. Verify a single retry fires.
- [ ] If the retry succeeds, verify polling resumes on the configured interval.
- [ ] If the retry fails again with `stopPolling: true`, verify polling stops again (not stuck firing).
- [ ] Verify gas-account fallback (e.g. SPONSOR_UNAVAILABLE) still flips the payer and retries immediately — this path should be unaffected.
- [ ] Verify JSON-RPC `execution reverted: ...` inner errors render the inner message in the inline fee error UI, not the wrapper Error's generic text.
- [ ] Verify no double error toast on stopPolling errors (only the inline message should show).

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11375" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/a0febf07-a7f9-4837-9e64-8a327ab61064?refs=ba01fa20179915a7e3cd1d1af6fb14e973ce5ba1&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/a0febf07-a7f9-4837-9e64-8a327ab61064/license?refs=ba01fa20179915a7e3cd1d1af6fb14e973ce5ba1&source=prChecksComment) |

packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx
```diff
@@ -734,13 +734,34 @@ function TxFeeInfo(props: IProps) {
...
+
         updateTxFeeInfoInit(true);
         updateTxAdvancedSettings({ dataChanged: false });
         updateSendFeeStatus({
           status: ESendFeeStatus.Error,
+          // Source-specific fields first so JSON-RPC errors keep their inner
+          // `execution reverted: ...` message instead of being shadowed by
+          // the wrapper Error's generic text.
           errMessage:
+            apiError?.data?.translatedMessage ??
```
originalix: 当 `estimateFee` 抛出 `OneKeyServerApiError`，并且响应里同时有外层 `data.message/translatedMessage` 和内层 `data.data.res.error.message` 时，这里的顺序会先选外层 message。

`OneKeyServerApiError.data` 是完整的 `IOneKeyAPIBaseResponse`，外层 message 往往是服务端包装文案；这样会覆盖内层 JSON-RPC 的 `execution reverted: ...`，用户在 fee error 里看不到真正的 RPC 失败原因。

建议把 `apiError.data?.data?.res?.error?.message` 放到外层 `translatedMessage/message` 前面，再 fallback 到 `apiError.message`，并补一个外层 message 与内层 RPC error 同时存在的用例。

weatherstar: Fixed in ce53bdd. Reordered the priority so `apiError.data?.data?.res?.error?.message` (the inner JSON-RPC error) is checked first, then falls back to `data.translatedMessage` / `data.message` / `apiError.message`. Updated the inline comment to explain why the inner RPC message has to win over the OneKey API wrapper.

No adjacent unit-test scaffolding exists for `TxFeeInfo.tsx`, so I did not add the test case here — happy to spin up a focused test file if you would like that as a follow-up.

packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx
```diff
@@ -1752,12 +1774,21 @@ function TxFeeInfo(props: IProps) {
...
-    const callback = () => run();
+    const callback = () => {
+      // setStopPolling(false) clears any prior server-driven stop AND
+      // resurrects the polling chain (the finally-block guard kills it once
+      // stopPollingRef flips true). The explicit run() covers gas-account
+      // fallback paths that emit the same event without ever stopping the
+      // chain, so the user still gets an immediate retry instead of waiting
+      // out the next tick.
+      setStopPolling(false);
+      void run();
```
originalix: 当后端返回 `stopPolling: true` 后，用户点击 fee error 的 Retry 会触发 `EstimateTxFeeRetry`。这里的 `setStopPolling(false)` 在先前处于 stopped 且存在 `pollingInterval` 时已经会 bump `pollingNonce` 并启动一次带 polling nonce 的 `run`；紧接着再 `void run()` 会用同一组 `unsignedTxs/accountId/networkId` 发起第二次估算。

普通 `ServiceGas.estimateFee` 路径没有把 `AbortController` signal 传进 `vault.estimateFee`，所以第二次调用不能可靠取消第一次，结果是重复打后端，并让 fee status/result 由两个请求竞争更新。

建议让 `setStopPolling(false)` 返回是否已经触发恢复请求，然后这个 callback 只在没有恢复请求时再调用 `run()`，保证每次 Retry 只有一次 estimate。

weatherstar: Fixed in c61fc92. `setStopPolling(false)` now returns whether it actually resurrected a paused polling chain (i.e. bumped `pollingNonce` and triggered a run); the `EstimateTxFeeRetry` callback uses that signal to skip its explicit `void run()` in the resume case, so each Retry now triggers exactly one `estimateFee`. The fallback path (no prior stop) still calls `run()` so gas-account retry events behave as before.

zhaono1 (approved):

originalix (approved):

PR #11383, opened 2026-04-27
fix(home): refresh all-net accounts on HW batch (OK-53863 OK-53864)

weatherstar:
## Summary
- Spot tab in **All Networks** mode loaded only BTC after a hardware-wallet connect; other enabled networks needed a manual refresh.
- DeFi tab in All Networks mode silently sat at the 30s `runAfterTokensDone` fallback — positions only appeared after a manual refresh.
- Both reduce to the same root cause: a 15s in-memory account-list cache captured a half-formed snapshot during HW default-account batch creation.

## Root Cause
After a HW wallet connects, `addDefaultNetworkAccounts → batchBuildAccounts` creates default accounts **in series** (BTC × 4 derive types → EVM → TRON → SOL), emitting `AddDBAccountsToWallet` between each batch. `useAllNetworkRequests` calls `getAllNetworkAccountsBaseCached` to enumerate accounts, and the cache:
- has a 15s TTL keyed only by `walletId / accountId / networkId / networksEnabledOnly / excludeTestNetwork`
- is invalidated **only** on `EnabledNetworksChanged`, not on `AddDBAccountsToWallet`

So the first all-network fetch — which typically lands while only the BTC batch has been written — populates the cache with a BTC-only `accountsInfo`. Spot fetches only BTC. DeFi filters that result through `deFiEnabledNetworksMap`, ends up with an empty network set, never calls `/wallet/v1/portfolio/positions`. The existing per-component `AddDBAccountsToWallet` listener in `TokenListBlock` re-runs, but the re-run still hits the stale cache.

## Fix
Invalidate this wallet's entries on every `AddDBAccountsToWallet`, then re-run with cache bypassed. This is wired at the `useAllNetworkRequests` layer so all consumers (Spot / DeFi / NFT) benefit without per-component duplication.

- `getAllNetworkAccountsBaseCached`: new `skipCache?: boolean` parameter — bypass the lookup but still write the fresh promise back, so concurrent readers see the new result.
- `IAllNetworkRequestsRunConfig`: new `skipAccountsCache?: boolean`. Relayed through `skipAccountsCacheRef` because `usePromiseResult` does not forward runner config into the method body.
- `useAllNetworkRequests`: subscribe to `AddDBAccountsToWallet`; on each batch, drop only this wallet's cache entries (`${walletId}::` prefix) and `runWithQueue({ skipAccountsCache: true, alwaysSetState: true })`. `alwaysSetState` is required so the refresh fires even when the consuming tab is mounted-but-not-active (DeFi while user is on Spot). `allNetworkDataInit` is intentionally **not** reset, so `clearAllNetworkData` does not flicker the visible list between batches.
- `runWithQueue`: OR-merge `skipAccountsCache` into the queued rerun config.
- `TokenListBlock` / `DeFiListBlock`: manual refresh now passes `skipAccountsCache: true` so it always re-derives the account list instead of waiting on the TTL.
- `useAllNetwork.ts` `onFinished`: moved into the runner's `finally` (try/catch guarded). The `TabListStateUpdate isRefreshing:false` signal — consumed by DeFi's `runAfterTokensDone` — is now always emitted, even when the work above threw before reaching the prior call site. DeFi no longer sits on the 30s fallback when the token fetch errors.

## Why per-event invalidation (not "HW first connect only")
`AddDBAccountsToWallet` fires for every account write — HW first connect, HD first creation, manual account add, batch fill — and they all need the same "just-added accounts visible immediately on home" guarantee. `runWithQueue`'s existing coalesce already collapses the HW-connect storm of ~7 emissions into ~2 actual fan-outs (first emit triggers a fetch, mid-batch emits hit `isFetching=true` and merge into a single trailing rerun), so a special-case path for HW first connect would add complexity without meaningful savings.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (all Home tab consumers in All Networks mode)
- **Risk Areas**:
  - `runWithQueue` rerun-config merge now carries the new `skipAccountsCache` flag — verified via OR-merge that newer manual-refresh intent is not lost when coalescing with a queued AddDB rerun
  - Moving `onFinished` to `finally` causes it to fire even on error paths inside the runner. `TokenListBlock` / `DeFiListBlock` `handleAllNetworkRequestsFinished` are pure event emits and idempotent. DeFi additionally clears `isForceRefreshRef.current = false`, which is only consumed by single-network mode (where `finally` already cleared it via the inner `try/finally`); All Networks mode does not consume `isForceRefreshRef`, so no regression
  - `skipAccountsCacheRef` is a single-shot ref; under concurrent deps-triggered re-run a manual-refresh request can lose its skip flag (worst case: falls back to TTL behavior, no data corruption). Documented as a follow-up
- **Out of scope (separate follow-ups)**:
  - DeFi force-refresh scheduler in MV3 extension SW (in-memory `setTimeout` lost on SW teardown)
  - `_runDeFiForceRefresh` `console.error` → `defaultLogger`

## Test plan
- [ ] Connect a fresh HW wallet → land on Wallet Home with All Networks selected → Spot tab shows BTC → EVM → TRON → SOL **progressively** without manual refresh; DeFi tab shows positions immediately after the All Networks fetch completes (no 30s fallback wait)
- [ ] Same flow but stay on Spot tab → switch to DeFi tab after HW batch finishes → DeFi already shows the correct All Networks data (the new `alwaysSetState` path)
- [ ] Manually add a single new account on Solana to an existing HW wallet → home Spot list updates within ~1s, no need to wait 15s TTL
- [ ] Manual refresh button on Spot / DeFi (header refresh) → forces a fresh account-list re-derive (network tab shows new fetch even within 15s of last fetch)
- [ ] Switch wallets / accounts repeatedly → no stuck loading state, no flicker between batches
- [ ] Single network mode (not All Networks): unchanged behavior
- [ ] Inject a throw in `TokenListBlock`'s `handleAllNetworkRequestsStarted` (e.g. mock `simpleDb.aggregateToken.getRawData()` to reject once) → Tokens tab still emits `TabListStateUpdate isRefreshing:false`, DeFi's `runAfterTokensDone` fires immediately rather than after 30s

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11383" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>

### DIFF
--- a/packages/kit/src/hooks/useAllNetwork.ts
+++ b/packages/kit/src/hooks/useAllNetwork.ts
@@ -395,9 +395,9 @@ function useAllNetworkRequests<T>(params: {
     if (!isAllNetworks) return;
     if (!currentWalletId) return;
     const walletIdAtSubscribe = currentWalletId;
-    const onAddDBAccounts = (params?: { walletId: string }) => {
-      if (!params?.walletId) return;
-      if (params.walletId !== walletIdAtSubscribe) return;
+    const onAddDBAccounts = (payload?: { walletId: string }) => {
+      if (!payload?.walletId) return;
+      if (payload.walletId !== walletIdAtSubscribe) return;
       const prefix = `${walletIdAtSubscribe}::`;
       for (const key of Array.from(allNetworkAccountsBaseCache.keys())) {
         if (key.startsWith(prefix)) {
--- a/packages/kit/src/hooks/useAllNetwork.ts
+++ b/packages/kit/src/hooks/useAllNetwork.ts
@@ -404,7 +404,14 @@ function useAllNetworkRequests<T>(params: {
           allNetworkAccountsBaseCache.delete(key);
         }
       }
-      void runWithQueueRef.current?.({ skipAccountsCache: true });
+      // alwaysSetState forces the runner past usePromiseResult's focus
+      // check, otherwise the refresh is dropped when the consuming tab
+      // (e.g. DeFi) is mounted but not the active tab during the HW connect
+      // batch — the cache would be cleared but no fetch would actually run.
+      void runWithQueueRef.current?.({
+        skipAccountsCache: true,
+        alwaysSetState: true,
+      });
     };
     appEventBus.on(EAppEventBusNames.AddDBAccountsToWallet, onAddDBAccounts);
     return () => {
--- a/packages/kit/src/hooks/useAllNetwork.ts
+++ b/packages/kit/src/hooks/useAllNetwork.ts
@@ -471,6 +471,9 @@ function useAllNetworkRequests<T>(params: {
       runCountRef.current += 1;
       isFetching.current = true;
 
+      let onStartedError: unknown;
+      let onStartedTask: Promise<void> | undefined;
+
       try {
         if (!allNetworkDataInit.current) {
           clearAllNetworkData();
@@ -484,8 +487,6 @@ function useAllNetworkRequests<T>(params: {
           allNetworkDataInit: !!allNetworkDataInit.current,
         });
 
-        let onStartedError: unknown;
-        let onStartedTask: Promise<void> | undefined;
         if (onStarted) {
           onStartedTask = onStarted({
             accountId: currentAccountId,
@@ -780,6 +781,18 @@ function useAllNetworkRequests<T>(params: {
         return resp;
       } finally {
         isFetching.current = false;
+        // Wait for onStarted to settle before firing onFinished, so
+        // the started/finished events for this run land in monotonic
+        // order (true -> false). Without this, an early throw above
+        // can fire onFinished while onStarted is still in flight,
+        // letting a stale "isRefreshing: true" arrive after "false".
+        if (onStartedTask) {
+          try {
+            await onStartedTask;
+          } catch (e) {
+            console.error(e);
+          }
+        }
         // Fire onFinished from finally so the "isRefreshing: false" signal
         // (consumed by DeFi tab's runAfterTokensDone) is always emitted —
         // even when the work above threw before reaching the prior call site.
--- a/packages/kit/src/hooks/useAllNetwork.ts
+++ b/packages/kit/src/hooks/useAllNetwork.ts
@@ -106,12 +106,14 @@ function getAllNetworkAccountsBaseCached({
   networkId,
   networksEnabledOnly,
   excludeTestNetwork,
+  skipCache,
 }: {
   walletId: string;
   accountId: string;
   networkId: string;
   networksEnabledOnly: boolean;
   excludeTestNetwork: boolean;
+  skipCache?: boolean;
 }): {
   cacheKey: IAllNetworkAccountsBaseCacheKey;
   reused: boolean;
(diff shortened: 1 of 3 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/4940703e-1bcf-4d1a-a4d9-29e9631295de?refs=8b9548229e168e8b5b4c0278a313ebffea509167&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/4940703e-1bcf-4d1a-a4d9-29e9631295de/license?refs=8b9548229e168e8b5b4c0278a313ebffea509167&source=prChecksComment) |

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1911,7 +1911,10 @@ function TokenListBlock({
...
   const handleRefreshAllNetworkData = useCallback(() => {
     isAllNetworkManualRefresh.current = true;
-    void runAllNetworksRequests({ alwaysSetState: true });
+    void runAllNetworksRequests({
+      alwaysSetState: true,
+      skipAccountsCache: true,
+    });
   }, [runAllNetworksRequests]);
 
   useEffect(() => {
```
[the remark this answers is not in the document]
weatherstar: Fixed in 05a689cdf2 — the duplicate listener at TokenListBlock.tsx:1920-1930 has been removed. The hook-level listener in useAllNetwork.ts:394-420 now exclusively owns AddDBAccountsToWallet with walletId filtering + skipAccountsCache.

packages/kit/src/hooks/useAllNetwork.ts
```diff
@@ -728,14 +776,21 @@ function useAllNetworkRequests<T>(params: {
...
-          networkId: currentNetworkId,
-        });
 
         return resp;
       } finally {
         isFetching.current = false;
+        // Fire onFinished from finally so the "isRefreshing: false" signal
+        // (consumed by DeFi tab's runAfterTokensDone) is always emitted —
+        // even when the work above threw before reaching the prior call site.
+        try {
```
originalix, line 799: 当 `getAllNetworkAccountsBaseCached()` 或后续账号查询在 `await onStartedTask` 之前抛错时，`onStarted` 仍然是并发任务。这里的 `finally` 会先发 `onFinished`，但 `onStarted` 之后还可能继续发 `TabListStateUpdate(isRefreshing: true)`。

结果是刷新状态可能按 `false -> true` 反向落地，Token/DeFi 的 `runAfterTokensDone` 会看到错误的完成时机，最终让 Home 列表卡在刷新中或让 DeFi 过早启动。

建议在 `finally` 里先等待 `onStartedTask` settle，再发 `onFinished`，保证同一次 run 的 started/finished 事件顺序单调。

weatherstar: Fixed in 7126212d39. Hoisted `onStartedTask` above the `try` so the `finally` can `await` it before firing `onFinished`. This ensures every run emits started/finished events monotonically (true → false), even when an early throw (e.g. from `getAllNetworkAccountsBaseCached`) skips the in-block `await onStartedTask` at line 581.

originalix (approved):

PR #11386, opened 2026-04-27
fix: prevent stale data and blank flash on home page network/account switch (OK-53686)

weatherstar:
## Summary
- Bound the per-owner rendered token list cache (`renderedTokenListCacheAtom`) at 50 MRU entries with one-shot legacy-shape migration so it can no longer grow unbounded into MMKV.
- Drop in-flight `initTokenListData` responses on owner change / unmount via a `cancelled` flag so a slow response from a previous owner cannot stomp on the freshly hydrated atoms of the new owner.
- Unify owner IDs written to `allTokenList` to the request-time params (instead of closure-captured React state) and document the Others-wallet `Tabs.Container` key fallback.

## Intent & Context
Follow-up to commits `f569232440` and `b2260ce2ad` (this branch) that introduced the per-owner rendered cache and a stable `Tabs.Container` key to eliminate stale data + blank-flash on home page network/account switches.

Self-review of those commits surfaced three correctness/robustness concerns:

1. **Unbounded memory / cold-start growth** — `byOwner` is a `Record` that gets a new entry every time the user visits a new `(account, network)` pair, and the atom is `coldStartCache: true` so every entry persists into MMKV. Long-running users would steadily slow down their cold start.
2. **Cross-owner async write race** — when the user switches A → B, the in-flight `initTokenListData(A)` finishes after B's `useLayoutEffect` has already eagerly hydrated the singleton atoms from B's cache, and overwrites them with A's data. The `ownerMismatch` UI gate would then briefly trigger a skeleton/empty list for B until B's own fetch completes — partly defeating the eager-hydration optimization the previous commits added.
3. **Inconsistent owner IDs** in `refreshAllTokenList` calls — the empty-cache and non-empty branches both used closure-captured `account?.id` / `network?.id` for the owner fields while the `keys` field used the request-time parameter. This mixes "request-time intent" with "current React state", making the race in #2 harder to reason about and harder to fix cleanly.

## Root Cause
1. The per-owner cache had no eviction policy and no upper bound — straightforward oversight when extending the cache from a single-entry shape to `byOwner`.
2. `initTokenListData` is defined inside a `useEffect` but wraps multiple `await`s without any `cancelled`/`AbortController` gate. React's effect cleanup wasn't doing anything to discard the in-flight result.
3. Reading `account?.id` / `network?.id` inside the async function captures the React state at the useEffect run that started the request — same as the parameter. Mixing the two stylistically makes it look like one is "current" and the other is "snapshot" when they're actually the same closure value, just expressed differently.

## Design Decisions
- **MRU cap of 50 entries** — chosen to comfortably cover a typical HD wallet (~5 active accounts × ~10 networks). Implemented by deleting the current owner key before re-inserting (so the entry lands at the end of `Object.keys` insertion order) and trimming the front when the size exceeds the cap. Considered an LRU on read but rejected — every owner switch triggers a write through `tokens`/`tokenListMap` updates, which is sufficient signal for "recently used".
- **Legacy-shape migration on the first write** — the previous build persisted `{ tokens, initialized, accountId, networkId }` (no `byOwner`). Without migration, the first write on the new build silently discards that entry and the user loses their cold-start cache for one launch. The migration runs only when `legacy.byOwner` is missing, lifts the single legacy entry into `byOwner` (without `tokenListMap`, which downstream guards already treat as a miss), and lets the next fresh write replace it.
- **`cancelled` flag over `AbortController`** — the underlying `getAccountLocalTokens` doesn't accept an abort signal, so we'd still need a check after the await. A simple closure-scoped boolean, flipped in the `useEffect` cleanup, is the minimum viable gate. Placed once just before the empty/non-empty dispatch so all subsequent `refresh*` / `update*` writes are covered.
- **Unify on the request-time params** — replaced `account?.id` / `network?.id` with `accountId` / `networkId` (the function arguments) in both `refreshAllTokenList` calls so "this response belongs to owner X" is expressed consistently.
- **Comment-only Others-wallet caveat in `HomePageView`** — kept the existing Tabs key fallback (`account.id` when `indexedAccountId` is missing) but documented that imported / watching / external wallets still remount on network switch by design, since their account IDs are network-scoped.

## Changes Detail
- `packages/kit/src/states/jotai/contexts/tokenList/atoms.ts` — exports `RENDERED_TOKEN_LIST_CACHE_MAX_OWNERS = 50` and updates the atom doc to reference the cap and writer responsibility.
- `packages/kit/src/components/TokenListView/index.tsx` — rewrites the `setRenderedTokenListCache` callback to handle legacy migration, MRU re-insertion, and LRU eviction.
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx` — adds the `cancelled` flag + cleanup, gates the post-await write path on it, and unifies the owner IDs in both `refreshAllTokenList` calls.
- `packages/kit/src/views/Home/pages/HomePageView.tsx` — comment-only addition documenting the Others-wallet key fallback.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: All (this is shared kit code rendered on Mobile / Desktop / Web / Extension)
- **Risk Areas**:
  - The MMKV-backed atom now has a different shape on disk for users coming from any earlier build of this branch. Migration path is exercised on the first write, but the guard `!legacy.byOwner` assumes the persisted value is either the legacy shape or the new `byOwner` shape — no third shape is expected.
  - The `cancelled` flag changes the semantic of `initTokenListData`'s late writes from "always commit" to "drop on owner change". This is the desired behavior, but any consumer that expected the late side-effects (e.g. `appEventBus.emit(EAppEventBusNames.TabListStateUpdate)`) for a previous owner will no longer see them. That emit only fires in the non-empty success branch and the listener uses the emitted `accountId/networkId` to decide what to do, so dropping the stale emit is safe.
  - LRU eviction relies on `Object.keys` preserving insertion order for non-integer string keys. `\${accountId}__\${networkId}` always contains non-digit chars (the `__` separator and id prefixes like `hd-`), so this is safe per ECMAScript spec.

## Test plan
- [ ] HD wallet, switch networks rapidly across 5+ networks: token list does not flash empty/skeleton on networks visited recently; balance does not show previous network's value.
- [ ] HD wallet, switch between two accounts on the same network repeatedly: same as above.
- [ ] Watching / imported wallet, switch networks: behavior unchanged from current `x` (remount + skeleton acceptable).
- [ ] Cold start after a session that visited many `(account, network)` pairs: verify `byOwner` size in MMKV is capped at 50 (`globalThis.__ONEKEY_CTX_ATOM_SNAPSHOT__` inspection).
- [ ] Upgrade path: build prior commit on this branch, render the home page so the legacy single-entry cache lands in MMKV, then build this commit and verify the migrated entry survives the first launch.
- [ ] Slow network simulation: switch from account A → B while A's `initTokenListData` is still awaiting `getAccountLocalTokens` — verify B's view is not overwritten by A's late response (`tokenListAtom`/`allTokenListAtom` should retain B's hydrated state).
- [ ] Empty-cache target (account with no local tokens cached for the network): verify token list initializes to empty + initialized state instead of staying in skeleton forever.

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11386" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/20566a99-3b93-464b-9612-a687e9927e30?refs=48b5d27fa79bafc0e9ec2278ae187df00c04a6ec&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/20566a99-3b93-464b-9612-a687e9927e30/license?refs=48b5d27fa79bafc0e9ec2278ae187df00c04a6ec&source=prChecksComment) |

packages/kit/src/components/TokenListView/index.tsx
```diff
@@ -313,27 +347,98 @@ function TokenListViewCmp(props: IProps) {
...
+          }
+        }
+
+        // MRU re-insertion: delete first so the spread below puts the
+        // current owner at the end of the key order. Combined with the
+        // size cap below, this keeps the most recently used entries.
+        delete nextByOwner[ownerCacheKey];
+        nextByOwner[ownerCacheKey] = {
+          tokens,
+          tokenListMap,
```
originalix, line 440: 这里的缓存只保存了 `tokenListMap`，但 rendered tokens 里可能包含 aggregate tokens。

触发场景是切换回一个已经缓存过、且列表包含聚合资产的 owner 时，`showSkeleton` 会因为 cache hit 直接返回 false，列表立即渲染 cached tokens；但是 `TokenBalanceView` / `TokenValueView` / `TokenPriceView` 对 aggregate token 会读取 `flattenAggregateTokensMapAtom`，这个 atom 没有随缓存一起恢复，仍可能是上一个账户或网络的 map。

结果就是 token 行已经切到新 owner，但聚合资产的余额/价值仍可能来自旧 owner，资产展示会短暂错配。

建议把 flattened aggregate token map 也纳入 rendered cache，并在 `TokenListBlock` 的 cache hydrate 路径里同步刷新 `aggregateTokensMap`；或者缓存一个覆盖普通 token 与 aggregate token 的统一 map，确保 tokens 与金额来源按同一个 owner 原子切换。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1581,7 +1592,74 @@ function TokenListBlock({
...
 
+  // Eagerly restore the singleton token-list atoms from the per-owner cache
+  // when the user switches to a network/account they've previously rendered.
+  // Runs synchronously before paint so `tokenListMapAtom` is in sync with
+  // the new `tokens` for the same render — without this, the balance and
+  // price components would briefly render against the previous owner's map.
+  // The async `initTokenListData` below still fetches the latest local cache
+  // and overwrites these atoms with fresh data once it returns.
+  useLayoutEffect(() => {
+    const currentAccountId = mergeDeriveAddressData
```
originalix: `mergeDeriveAddressData` 为 true 时，这里用 `indexedAccount?.id` 作为 cache lookup key，但缓存写入方并不是这个 owner。

`TokenListView` 写 `renderedTokenListCache` 时使用传入的 `accountId`，当前调用处传的是 `account?.id`；`initTokenListData` 刷新 `allTokenList` 时也写入 request-time `accountId`。因此合并派生资产开启后，已经写入的 cache key 是 `account.id__network.id`，这里却查 `indexedAccount.id__network.id`，会稳定 cache miss。

影响是 merge derive 账户切换回已缓存网络时，paint 前的 `tokenListMap` hydrate 不会发生，`TokenListView` 仍可能用旧 map 渲染 cached tokens，或退回等待异步 local token fetch。

建议抽出一个 token-list owner id helper，并让 `TokenListView` props、cache 写入、`allTokenList.accountId`、这里的 cache lookup 全部使用同一个 owner 规则。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1581,7 +1592,74 @@ function TokenListBlock({
...
+    const cacheKeys = `${currentAccountId}_${currentNetworkId}_cache`;
+    refreshTokenList({ tokens: cached.tokens, keys: cacheKeys });
+    refreshTokenListMap({ tokens: cached.tokenListMap });
+    refreshAllTokenList({
+      keys: cacheKeys,
+      tokens: cached.tokens,
+      accountId: currentAccountId,
+      networkId: currentNetworkId,
+    });
+    refreshAllTokenListMap({ tokens: cached.tokenListMap });
```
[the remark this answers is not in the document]
weatherstar: Fixed in c8e73074fc. Cache hydration in `TokenListBlock` now also resets `smallBalanceTokenList`/`smallBalanceTokenListMap`/`smallBalanceTokensFiatValue` and `riskyTokenList`/`riskyTokenListMap` (and their refresh fns are added to the dep array), so `TokenListFooter` no longer briefly shows the previous owner's small-balance count/value before `initTokenListData` repopulates them.

packages/kit/src/components/TokenListView/index.tsx
weatherstar: Fixed in c8e73074fc. The per-owner cache entry now also persists the raw nested `aggregateTokensMap` (added to `renderedTokenListCacheAtom` schema), `TokenListView` writes it alongside `tokenListMap`, and `TokenListBlock` 的 useLayoutEffect 在 hydrate 阶段调用 `refreshAggregateTokensMap({ tokens: cached.aggregateTokensMap })` 同步刷新源 atom。这样 `flattenAggregateTokensMapAtom` 的计算结果会和 cached tokens 一同切换，聚合资产的余额/价值不会再短暂错配。Legacy 缓存条目缺这个字段时跳过该 hydrate，让 async fetch 自然填回。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
weatherstar: Fixed in c8e73074fc. 抽出了 `getTokenListOwnerCacheAccountId` helper（位于 `packages/kit/src/components/TokenListView/utils.ts`），写入端（`TokenListView` 的 `ownerCacheKey`）和读取端（`TokenListBlock` useLayoutEffect 的 lookup）现在都通过它统一推导：merge 模式下使用 `indexedAccountId`，否则使用 `accountId`。`TokenListBlock` 已通过新增的 `mergeDeriveAddressData` prop 把规则透传给 `TokenListView`，merge derive 切回已缓存网络时不会再稳定 cache miss。

`allTokenList.accountId` 仍写入 request-time `account.id`（保持 `UniversalSearch.tsx:268` 等消费者的现有语义），cache key 与 `allTokenList` owner 在 merge 模式下分别使用 `indexedAccountId` 和 `account.id` 是有意为之。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1581,7 +1593,99 @@ function TokenListBlock({
...
+      mergeDeriveAddressData: !!mergeDeriveAddressData,
+    });
+    const currentNetworkId = network?.id;
+    if (!currentAccountId || !currentNetworkId) return;
+    if (
+      allTokenListAtomValue.accountId === currentAccountId &&
+      allTokenListAtomValue.networkId === currentNetworkId
+    ) {
+      return;
+    }
```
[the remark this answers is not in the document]
weatherstar: Fixed in 71c7358e8d. Aligned the hydrate path with the writer convention instead of the cache-key convention:

- Guard at lines 1611-1626 now compares `allTokenList.accountId` against `account?.id` (the same axis every `refreshAllTokenList` writer uses — `run`, `initTokenListData`, `updateAllNetworksTokenList`).
- The hydrate `refreshAllTokenList` stamp at lines 1645-1652 now writes `account?.id` (with `currentAccountId` as a defensive fallback for the brief window where `account?.id` is undefined in merge mode).
- The cache lookup key still uses `currentAccountId` via `getTokenListOwnerCacheAccountId`, since the per-owner cache is intentionally keyed by `indexedAccountId` in merge mode.

This converges the guard and writers on `account?.id`, so the hydrate no longer re-fires on every poll in merge mode and `smallBalanceTokenList` / `riskyTokenList` are no longer reset to empty after each fresh fetch.

`ownerMismatch` in `TokenListView/index.tsx:225-231` already compares on the `accountId` prop (= `account.id`), which now matches the hydrate stamp, so it stays correct without further changes.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1735,6 +1858,31 @@ function TokenListBlock({
...
+          // response belongs to) — not closure-captured React state which
+          // can read like "current owner" but is actually frozen at the
+          // useEffect run that fired this request.
+          refreshAllTokenList({
+            keys: emptyKeys,
+            tokens: [],
+            accountId,
+            networkId,
+          });
+          refreshAllTokenListMap({ tokens: {} });
```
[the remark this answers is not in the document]
weatherstar: Fixed in 3ca185100e. The new empty-token-cache branch in `initTokenListData` now also calls `refreshSmallBalanceTokensFiatValue({ value: '0' })` between `refreshSmallBalanceTokenListMap` and `refreshRiskyTokenList`, mirroring the parallel reset in the `useLayoutEffect` hydration. `refreshSmallBalanceTokensFiatValue` was also added to the `useEffect` dependency array. The fiat-value atom no longer retains the previous owner's value when switching to an account/network whose local cache is empty.

zhaono1 (approved):

originalix (approved):

PR #11400, opened 2026-04-28
fix(home): gate app-resume refresh on outer route focus

weatherstar:
## Summary
- Combine outer route focus (`useRouteIsFocused`) with the inner Home-tab `isFocused` signal in `TokenListBlock` and `TxHistoryContainer`, so foreground/visibility resume only refreshes when Home is actually visible.
- Rename the local outer-focus variable to `isRouteFocused` to avoid shadowing the existing `isPageFocused` parameter consumed by nearby `overrideIsFocused` callbacks.

## Intent & Context
The manual `onVisibilityStateChange` listeners on the home Tokens and History lists were firing refreshes on every app resume, even when the user wasn't actually on the Home screen. `useTabIsRefreshingFocused()` only tracks which inner Home tab (Tokens/NFT/DeFi/History) is selected — it stays "focused" while the user is on Market/Swap, on the lock screen, or has a modal stacked above Home. As a result, switching apps and coming back was triggering unnecessary home polling/refresh in those states.

## Root Cause
`isFocused` from `useTabIsRefreshingFocused()` is scoped to the inner Home tab selector and does not reflect outer navigation state. The visibility-resume effect treated this inner-tab focus as a proxy for "user is looking at this list," which is incorrect whenever a higher-level route or modal is on top, or when the app is locked.

## Design Decisions
- Reused the existing `useRouteIsFocused` hook (already used elsewhere in the app) instead of introducing a new focus-tracking mechanism. It already accounts for: bottom-tab focus, root navigator stack depth (modals), and `useAppIsLockedAtom` (lock screen).
- Combined the two signals as `isFocused && isRouteFocused` rather than replacing `isFocused`. The inner-tab signal is still required so we don't refresh the Tokens list while the user is on the History tab and vice versa.
- Renamed the local variable from the initially-considered `isPageFocused` to `isRouteFocused` because both files already pass an `isPageFocused` parameter into `overrideIsFocused` callbacks for nested list components — using the same name would have shadowed that prop and caused subtle bugs.
- Scope kept intentionally narrow: only the two `onVisibilityStateChange` listeners are gated. Other refresh triggers in these files (account switch, network switch, manual pull-to-refresh) are left untouched.

## Changes Detail
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx`: import `useRouteIsFocused`, gate the `onVisibilityStateChange` callback on `isFocused && isRouteFocused`, and add `isRouteFocused` to the effect dependency array.
- `packages/kit/src/views/Home/pages/TxHistoryContainer.tsx`: same change applied to the History list's visibility-resume effect.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile / Desktop / Web / Extension (all platforms — uses cross-platform hooks only)
- **Risk Areas**: The change makes the resume-refresh strictly more conservative (fewer refreshes), so the failure mode is "list shows slightly stale data after backgrounding while on Market/Swap/locked." Pull-to-refresh and the other refresh triggers (account/network switch) remain unchanged and will still re-fetch.

## Test plan
- [ ] Open Home → Tokens, background the app, foreground → list refreshes (baseline behavior preserved).
- [ ] Open Home → History, background and foreground → list refreshes.
- [ ] Switch to Market or Swap tab, background and foreground, then return to Home → no spurious refresh fired while away from Home; manual interaction still works.
- [ ] Lock the app while on Home, foreground → no refresh while locked; refresh resumes once unlocked and Home becomes the active route.
- [ ] Open a modal above Home (e.g. account selector, send flow), background and foreground → no refresh fires while the modal covers Home.
- [ ] Switch between inner Tokens/History tabs and confirm pull-to-refresh, account-switch, and network-switch refresh paths are unaffected.

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11400" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c30c7872-6806-4c00-90a3-46c07dfd4b83?refs=c83cf6c19e3480c3adb0a6a2414a82fcd8fa7e12&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c30c7872-6806-4c00-90a3-46c07dfd4b83/license?refs=c83cf6c19e3480c3adb0a6a2414a82fcd8fa7e12&source=prChecksComment) |

originalix (approved):

limichange (approved):

PR #11406, opened 2026-04-28
feat: align custom User-Agent across non-axios call sites

originalix:
## Summary
- Add `withCustomUAHeaders(url, headers)` (in `@onekeyhq/shared/src/request/customUA.ts`) and wire it into the seven call sites that bypass our axios interceptor: `useNetInfo` health probe, `DesktopApiAppUpdate` (electron-updater feed), `DesktopApiDev` (log upload + curl), `ServicePrimeTransfer.health`, `ServiceSwap` quote SSE, RN `exportLogs.uploadLogBundle` (main + fallback), and CLI `swap-quote` SSE.
- Patch chromium `app.userAgentFallback` synchronously at desktop main-process load so the product token reflects `APP_VERSION` (not the Electron binary version) in dev.
- Single-source the renderer-side app version: `development/scripts/sync-version.js` regenerates `packages/shared/src/runtime/version.ts` from `.env.version` at postinstall, and `platformEnv.version` now reads from that constant.

## Intent & Context
Cloudflare started rejecting requests with the bare runtime UA (`node`, Electron product/binary mismatch in dev, RN raw fetch). Axios already injects a OneKey UA via the request interceptor, but the call sites listed above bypass axios — they use `fetch`, `EventSource`, `electron-updater` requestHeaders, or RN `FileSystem.createUploadTask`. This PR makes those code paths inject the same product token (`OneKeyWallet/<version>`) as axios, so they survive CF's UA filtering.

The dev-only Electron product-token mismatch is fixed at module load (before `ready` fires) so the very first webContents already sees the patched fallback.

## Design Decisions
- **`shared` does not depend on `kit-bg`.** The whitelist (`checkIsOneKeyDomain`) and dev-toggle reads (`getDevSettingsPersistAtom`) are pulled through `requestHelper`'s existing DI surface — same pattern as the existing axios interceptor. Avoids a new circular import.
- **Runtime detection vs. UA payload.** The runtime (`desktop-electron`/`cli-node`/`ios-native`/`android-native`) is conveyed via `X-Onekey-Request-Platform`, not the UA itself. The UA stays a clean `OneKeyWallet/<version>` product token.
- **Web/Extension explicitly return `null`.** The browser-supplied UA already passes CF; injecting our own would just shadow the platform UA without benefit.
- **Refuse to inject when DI is unwired (App).** If `requestHelper.checkIsOneKeyDomain` throws, App returns `false` (no UA). Only CLI falls back to the narrow regex `/\.onekey(cn|test)\.com$/i`, because CLI cannot wire the full kit-bg DI and only talks to those two TLDs.
- **Caller wins.** If a call site already set `User-Agent` (case-insensitive), we keep theirs and emit `defaultLogger.app.customUA.callerConflict` with a sanitized URL (path only, no query) for observability.
- **`disableCustomUA` dev toggle.** Surfaced under DevSettings to reproduce CF UA-block behavior locally without un-installing OneKey UA from the build.

## Changes Detail
- `packages/shared/src/request/customUA.ts` — core module: `buildCustomUA`, `shouldInjectUAForUrl`, `withCustomUAHeaders`, `setCustomUARuntime`. 22 unit tests in `customUA.test.ts` covering platform branches, DI failures, App-vs-CLI fallback policy, and case-insensitive caller-UA detection.
- `apps/desktop/app/app.ts` — sync regex patch on `app.userAgentFallback` to align dev chromium UA with `APP_VERSION`.
- `apps/cli/src/runtime/customUA.ts` + `cli.ts` — side-effect import that calls `setCustomUARuntime('cli-node')` before any swap command runs.
- `packages/shared/src/runtime/version.ts` (auto-generated) + `development/scripts/sync-version.js` — postinstall step writes `APP_VERSION` from `.env.version`.
- `packages/shared/src/logger/scopes/app/scenes/customUA.ts` — `callerConflict` warn-level scene for observability of UA collisions.
- Seven call-site shims (each just wraps existing headers via `withCustomUAHeaders`).

## Known Limitations (follow-up, not blocking)
1. **Dev whitelist leakage** — `checkIsOneKeyDomain` (in `kit-bg/src/endpoints/index.ts`) builds its whitelist from `getEndpointsWithCustomConfig()` when `platformEnv.isDev && devSettings.enabled`. This means dev-only `customApiEndpoints` (e.g. `localhost`, `myproxy.example.com`) will also receive `OneKeyWallet/<version>`. Production builds are unaffected. Plan: tighten with an official-domain regex pre-check in `shouldInjectUAForUrl` in a follow-up PR.
2. **`disableCustomUA` toggle scope** — the toggle only affects `buildCustomUA()` (header injection paths). The chromium webview's UA was patched at module load and is not reverted by the toggle; it requires a restart to come back to the raw Electron UA. Toggle subtitle should be reworded to make this explicit. Will land with the dev-whitelist fix.
3. **`apps/desktop/app/preload.ts:266`** still reads `process.env.VERSION` for `desktopApi.appVersion`. The webpack DefinePlugin keeps it consistent with `APP_VERSION` in production, but the "single source" claim isn't fully true until preload is migrated as well. Tracking separately.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Mobile (iOS+Android) / CLI. Web and Extension explicitly opt out (`buildCustomUA` returns `null`).
- **Risk Areas**:
  - Desktop chromium UA regex patch in `app.ts` runs synchronously at module load with a `try/catch` — if the Electron-shipped fallback UA ever changes shape, the regex silently no-ops (warn-only). Verified on current Electron version locally.
  - `withCustomUAHeaders` is invoked on every request to the seven covered call sites. The whitelist check goes through an existing cached jotai atom read; no new network/disk hits.
  - `runtime/version.ts` is checked into git. If a developer changes `.env.version` and skips `yarn install`, the file is stale until the next install. No CI guard yet.

## Test plan
- [ ] `yarn workspace @onekeyhq/shared test customUA.test.ts` — 22/22 pass (verified locally).
- [ ] Desktop dev: start `yarn app:desktop`, open DevTools, confirm webContents UA shows `OneKeyWallet/6.3.0` (not `OneKeyWallet/<electronVer>`).
- [ ] Desktop prod build: `yarn app:desktop:build`, install, run, confirm electron-updater feed request carries `User-Agent: OneKeyWallet/6.3.0`.
- [ ] iOS / Android: trigger Settings → Export Logs → Upload, capture both `mainPath` (Expo `createUploadTask`) and `fallback` (`fetch`+FormData) requests, confirm `User-Agent` injected.
- [ ] CLI: `yarn workspace @onekeyfe/cli build && node apps/cli/dist/cli.js swap-quote ...`, capture SSE request headers, confirm `User-Agent: OneKeyWallet/<version>`.
- [ ] DevSettings → toggle `disableCustomUA` ON → trigger any of the above, confirm header-injection paths drop the UA. (Note: chromium webview UA is unaffected — see Known Limitations.)
- [ ] Sanity: hit a non-whitelisted host (e.g. `https://example.com`) from any call site, confirm UA is NOT injected.

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11406" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f013d480-9c0c-4880-8cf5-80976b68e878?refs=20f75206791f24b78e1248764cab39db1d128e23&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f013d480-9c0c-4880-8cf5-80976b68e878/license?refs=20f75206791f24b78e1248764cab39db1d128e23&source=prChecksComment) |

apps/desktop/app/app.ts
```diff
@@ -1794,6 +1795,22 @@
...
 /* oxlint-enable typescript/no-unsafe-call */
 
+// In dev, app.getVersion() falls back to the electron binary version because
+// no packaged app/package.json is on disk, leaving chromium UA's product token
+// as `OneKeyWallet/<electronVer>`. Patch only that token so dev/prod agree.
+// Run synchronously at module load (before `ready` fires) so the very first
+// webContents created in the ready handler already sees the patched UA —
+// `app.userAgentFallback` is readable/writable before `ready`.
+try {
+  const electronVer = process.versions.electron.replace(/\./g, '\\.');
```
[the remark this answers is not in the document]
originalix: Fixed in 9a41ea72a9. Replaced the partial regex escape (`/\./g`) with a full meta-character escape (`/[.*+?^${}()|[\]\\]/g`) so backslashes (and any other regex metachar) in `process.versions.electron` are escaped before being interpolated into the RegExp source. In practice the value is always semver digits, but the strict version is a one-liner and clears the alert.

packages/kit-bg/src/services/ServicePrimeTransfer/ServicePrimeTransfer.ts
```diff
@@ -159,8 +161,13 @@ class ServicePrimeTransfer extends ServiceBase {
           const controller = new AbortController();
           const timeoutId = setTimeout(() => controller.abort(), 10_000); // 5 second timeout
 
-          const response = await fetch(`${url}/health`, {
+          const healthUrl = `${url}/health`;
+          const baseHeaders = await getRequestHeaders();
```
originalix: 当用户保存自定义 Prime Transfer server 时，这里会对用户输入的 `endpoint` 请求 `/health`，并且无协议输入还会同时探测 `http://...`。

现在代码在域名确认前无条件调用 `getRequestHeaders()`；即使 `withCustomUAHeaders()` 因非 OneKey 域名不注入 `User-Agent`，这些 `X-Onekey-*` headers 仍会随请求发给自定义服务器。

这会把 `instanceId`、设备/平台、locale、版本等稳定客户端信息暴露给任意自定义 endpoint，HTTP fallback 还会明文发送。修复方向：非 OneKey 自定义 server 的健康检查传空 headers；只有先确认目标属于 OneKey 官方白名单后，再构造并发送 app headers/UA。

packages/shared/src/request/customUA.ts
```diff
@@ -0,0 +1,114 @@
...
+  if (!url || typeof url !== 'string') return false;
+  const parsed = uriUtils.safeParseURL(url);
+  if (!parsed) return false;
+  try {
+    return await requestHelper.checkIsOneKeyDomain(url);
+  } catch {
+    // CLI has no wired DI — fall back to a narrow regex (CLI only talks to
+    // *.onekeycn.com / *.onekeytest.com). For App, refuse rather than
+    // mis-inject under an uncertain whitelist.
+    if (detectRuntime() !== 'cli-node') return false;
```
originalix: 当 `DesktopApiAppUpdate` / `DesktopApiDev` 在 Electron main process 调用 `withCustomUAHeaders()` 时，`requestHelper` 在这个进程里没有被 `updateInterceptorRequestHelper()` 初始化，`checkIsOneKeyDomain()` 会走默认实现抛错。

这里的 catch 对非 `cli-node` runtime 直接返回 `false`，所以 main process 的 update feed 和日志上传永远不会拿到新 `User-Agent`，虽然调用点看起来已经接入了 custom UA。

修复方向：为 `desktop-electron` 的 main process 提供明确的 OneKey endpoint 判定来源，例如初始化 `requestHelper`，或在这个分支只对 Electron main 使用与 endpoint map 一致的官方域名 fallback。

apps/desktop/app/app.ts
```diff
@@ -1794,6 +1794,34 @@ function startWebviewMemoryMonitoring() {
...
+  // well-formed in practice, but CodeQL flags partial escapes and the
+  // strict version is a one-liner.
+  const electronVer = process.versions.electron.replace(
+    /[.*+?^${}()|[\]\\]/g,
+    '\\$&',
+  );
+  // process.env.VERSION is substituted at build time by webpack DefinePlugin
+  // (apps/desktop/scripts/build.js) — the same path every other call site
+  // uses. Falls back to '1' to match buildCustomUA()'s fallback in
+  // packages/shared/src/request/customUA.ts.
```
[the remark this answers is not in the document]
originalix: Fixed in 25e0fabfad. You are right — `app.name = 'OneKey Wallet'` (with the space) at line 137, so Chromium’s product token is `OneKey Wallet/<electronVer>` and the previous regex never matched. Updated the regex to `OneKey ?Wallet/${electronVer}\b` to match both spaced and no-space forms, and normalized the replacement to the canonical no-space `OneKeyWallet/<APP_VERSION>` so the chromium fallback agrees with `buildCustomUA()`.

packages/kit-bg/src/services/ServicePrimeTransfer/ServicePrimeTransfer.ts
originalix: Fixed in e7cd27addd. `verifyWebSocketEndpoint` 现在先 `await checkIsOneKeyDomain(healthUrl)`：只有官方域名（`*.onekeycn.com` / `*.onekeytest.com`）才会附带 `getRequestHeaders()` + `withCustomUAHeaders()`，自定义 server 走空 headers。https fallback 对未知域名也不再发 `X-Onekey-*`，明文 http 探测就更不会泄漏了。

packages/shared/src/request/customUA.ts
originalix: Fixed in 86d27fd17b. 你说的对——`updateInterceptorRequestHelper()` 只在 renderer init，所以 main process 里 `requestHelper.checkIsOneKeyDomain()` 抛错，旧的 catch 对 `desktop-electron` runtime 直接返回 `false`，导致 `DesktopApiAppUpdate` / `DesktopApiDev` 即使调了 `withCustomUAHeaders()` 也拿不到 UA。

现在把已有的 cli-node fallback 扩展到 `desktop-electron`：main process 抛错时走窄正则 `\.onekey(cn|test)\.com$`，正好覆盖 `utility.onekeycn.com` 等真实 endpoint。renderer 仍走 DI 结果（不会落到 catch），native / web / extension 维持原“拒绝 fallback、避免 mis-inject”行为。新加了两个 unit test 覆盖 main fallback 命中 / 拒绝非官方域名。

revan-zhang (approved):

ezailWang (approved):

PR #11433, opened 2026-04-30
fix: send flow account whitelist, hex alert, and narrow-screen font (OK-53899, OK-53898, OK-53519)

weatherstar:
## Summary
- Refresh account address lookup cache after creating new accounts so cross-account transfers no longer flag the recipient as non-whitelisted (OK-53899)
- Restrict the custom hex-data alert to contract recipients only, matching the original spec confirmed with PM (OK-53898)
- Auto-shrink the send amount font size to fit narrow viewports (e.g., extension popup) so token symbols and amounts stay fully visible (OK-53519)

## Intent & Context
Three independent bugs reported against the Send flow:

1. **OK-53899** — After creating a new account in wallet A, initiating a transfer from wallet B to that new account incorrectly displayed a "non-whitelisted address" warning. Users had to switch accounts and re-enter the send flow for the new address to be recognized. The address-lookup memo cache was returning the stale empty result captured before the account existed.
2. **OK-53898** — Any non-empty custom hex-data input was triggering the contract-call risk alert, even when the recipient was a regular EOA. Per PM confirmation, this alert should only fire when the recipient is a contract (the original behavior).
3. **OK-53519** — In the extension popup (narrow viewport) and other compact layouts, the amount input font was too large, causing longer token symbols (6+ chars) to be clipped or pushed off-screen.

## Root Cause
- **OK-53899**: `getAccountNameFromAddress` cached an empty result via the memoized `getAccountNameFromAddressMemo`. When a new account was later created, the cache still returned empty for that address and the address never re-resolved to a known account. Additionally, `saveAccountAddresses` writes were queued in `_saveAccountAddressesBatchByCache` but not flushed promptly after account creation, so the underlying lookup table was stale.
- **OK-53898**: `isCustomHexData` was computed purely from `isCustomTxMessageEnabled && displayTxMessageForm && tokenInfo?.isNative && …`, missing the `submitRecipientIsContract` gate. Any hex input on a native-token transfer would trigger the alert regardless of recipient type.
- **OK-53519**: The auto-size amount input only reduced the base font size when `!md` (i.e., desktop wide layout), which is the inverse of what was intended — the compact-width breakpoint should apply on the `md` (mobile/narrow) layout. Even so, the static base-size approach didn't account for the actual measured content width versus available inline width, so longer values overflowed.

## Design Decisions
- **Account address cache flush** — Two complementary fixes: (1) on cache miss (`isEmpty(result)`), evict the memoized entry so the next read goes back to the source of truth instead of returning a stale empty value; (2) after `saveAccountAddresses` is queued for newly created accounts, explicitly `await _saveAccountAddressesBatchByCache.flush()` so the persisted lookup is consistent before any subsequent send-flow check runs. Errors during flush are swallowed (matches surrounding pattern) since the next read can recover.
- **Hex alert gate** — Smallest possible diff: add `submitRecipientIsContract &&` in front of the existing condition. The flag is already computed upstream from the recipient resolution, so no new state or fetch is needed.
- **Font auto-fit** — Introduced `getFittedInlineFontSize` that iteratively shrinks the font size by 1px until the measured prefix + amount + suffix + gaps fits within `availableInlineWidth` (with an 8px safety margin), bounded by `minFontSize`. All downstream sizing (line height, input height, prefix/suffix offsets, min/max widths) now derives from `effectiveFontSize` instead of the raw `fontSize`, keeping the layout coherent at any size. Also fixed the inverted `!md` → `md` breakpoint check so the compact min-font-size threshold actually applies on narrow layouts.

## Changes Detail
- `packages/kit-bg/src/dbs/local/LocalDbBase.ts` — After saving any new account's address, await `_saveAccountAddressesBatchByCache.flush()` so the lookup table is durable before subsequent reads.
- `packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts` — In `getAccountNameFromAddress`, evict the memo entry when the result is empty so a later create→lookup sequence doesn't return the stale empty cache.
- `packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx` — Add `submitRecipientIsContract` to the `isCustomHexData` predicate so the alert only fires for contract recipients.
- `packages/kit/src/views/Send/components/SendAutoSizeAmountInput/AutoSizeInput.tsx` — Add `getInlineContentWidthPx` + `getFittedInlineFontSize` helpers; route all sizing/measurement through `effectiveFontSize`; accept `minFontSize` prop.
- `packages/kit/src/views/Send/components/SendAutoSizeAmountInput/index.tsx` — Fix inverted breakpoint (`!md` → `md`) so the compact-width min font size applies on narrow viewports.

## Risk Assessment
- **Risk Level**: Low–Medium
- **Affected Platforms**: All (Extension / Mobile / Desktop / Web)
- **Risk Areas**:
  - Address cache flush touches a hot path during account creation — if `flush()` blocks longer than expected it could marginally delay account-creation completion. Errors are swallowed to preserve the prior behavior.
  - Memo eviction on empty result could increase lookup volume for genuinely unknown addresses (each call re-runs the source lookup). Acceptable: the prior behavior of permanently caching empty was the bug.
  - Font auto-fit runs a small iterative shrink loop on every render of the amount input. Bounded by `(fontSize - minFontSize)` iterations with cheap text-measurement, so impact is negligible, but worth eyeballing in the extension popup for jank when typing fast.

## Test plan
- [ ] **OK-53899**: Create a new account in wallet A. Without switching accounts, open Send from wallet B and enter the new account's address as the recipient — verify the address resolves to the known account name and the non-whitelisted warning does NOT appear.
- [ ] **OK-53898**: On a native-token send to a regular EOA, enable custom hex data and enter arbitrary bytes — verify NO contract-call alert appears. Repeat with a contract address — verify the alert DOES appear (regression check).
- [ ] **OK-53519**: Open the extension popup, go to Send, pick a token with a 6+ char symbol — verify the symbol and amount are fully visible and the font shrinks rather than being clipped. Test a long amount (e.g., 12345.6789) and confirm it still fits.
- [ ] Smoke-test Send on desktop, mobile, and web at typical viewport sizes to confirm no layout regression for short symbols / typical amounts.

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/11433" target="_blank">
  <picture>
    Open in Devin Review
  </picture>
</a>


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2d0d690c-41b3-46ca-949b-ac546103f45f?refs=64a16e687a5a79b23ea19f7db01af1c04677e123&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2d0d690c-41b3-46ca-949b-ac546103f45f/license?refs=64a16e687a5a79b23ea19f7db01af1c04677e123&source=prChecksComment) |

originalix (approved):

zhaono1 (approved):

PR #11446, opened 2026-05-04
feat: add deriveContextHash BTC API for deterministic secret derivation

jrwbabylonlab:
Implements `window.$onekey.btc.deriveContextHash(appName, context)`, producing a deterministic 32-byte value from the wallet's key material via HKDF-SHA-256. Useful for HTLC preimages, app-bound identifiers, and similar deterministic-secret use cases.

- Pure HKDF-SHA-256 derivation in `@onekeyhq/core` (salt `"derive-context-hash"`, info `SHA-256(appName) || contextBytes`), with strict input validation
- HD wallets: derive 32-byte IKM at BIP-32 path `m/73681862'` from the stored BIP-39 seed (incorporates BIP-39 passphrase if used at creation)
- Imported wallets: use raw 32-byte private key directly
- Hardware/QR/watching/external accounts: rejected at provider, base keyring, and service layers (positive whitelist on HD/imported only)
- Sensitive payload (`appName`/`context`) kept out of route-query JSON, modal logs, and console output via nonce-keyed in-memory store
- Cross-platform-safe approval modal under `DAppConnection` route
- 29 unit tests including 3 known-answer interop vectors

Marked `@experimental`.

image


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/ffdd8480-adf7-49bc-ab87-a6b04c48b78a?refs=32d6de6d9d3f0dd3cc105c8b49ac55958ba7f111&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/ffdd8480-adf7-49bc-ab87-a6b04c48b78a/license?refs=32d6de6d9d3f0dd3cc105c8b49ac55958ba7f111&source=prChecksComment) |

packages/kit/src/views/DAppConnection/pages/DeriveContextHashModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+// @experimental, so hardcoded English copy is acceptable for the first
+// release behind that flag.
+const COPY = {
+  appNameLabel: 'Application name',
+  contextLabel: 'Context (hex)',
+  warning:
+    'A deterministic value will be derived from your wallet using the application name and context shown above. Anyone with the same key material, application name, and context can produce the same value.',
+  payloadLoadFailed:
+    'Failed to load the request payload. Please cancel and try again from the application.',
+};
```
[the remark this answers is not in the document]
jrwbabylonlab: Acknowledged — the hardcoded English strings are intentional given the constraints:

1. The `ETranslations` enum and JSON files are auto-generated from OneKey's translation pipeline (`yarn i18n:pull` from an external translation management system); a contributor can't add new keys via a code change alone.
2. This API is marked `@experimental` and the surface may change before stabilization, so requesting a designer/translator round-trip for keys that may rename or disappear felt premature.
3. A `TODO(i18n)` comment is in place explicitly tracking the gap so it gets picked up when the API stabilizes.

Happy to follow whichever path the OneKey maintainers prefer:
- (A) leave as-is with the TODO until stabilization,
- (B) add the 4 new keys via the translation pipeline before merge,
- (C) drop the labels and rely on minimal existing keys (`global_*` etc.) as a partial mitigation.

Will defer to maintainer guidance.

packages/kit/src/views/DAppConnection/pages/DeriveContextHashModal.tsx
```diff
@@ -0,0 +1,224 @@
...
+    },
+    [walletId, accountId, networkId, payloadNonce, dappApprove],
+  );
+
+  return (
+    <DappOpenModalPage dappApprove={dappApprove}>
+      <>
+        <Page.Header headerShown={false} />
+        <Page.Body>
+          <DAppRequestLayout
```
originalix, line 148: This approval screen shows the origin, appName, and context, but it does not show the wallet/account/network used for the derivation. Since deriveContextHash returns a stable value derived from wallet key material, the user should be able to verify which account context they are approving.

Could we model this as a BTC-specific signMessage-like request and reuse the existing SignMessageModal account/network presentation pattern? I can help contribute the implementation if you are open to that direction.

packages/core/src/chains/btc/sdkBtc/deriveContextHashFromCredentials.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  // Fail-closed: keyrings populate exactly one of hd/imported. Both being
+  // present would be ambiguous credential state — refuse rather than guess.
+  if (credentials.hd && credentials.imported) {
+    throw new OneKeyLocalError(
+      'deriveContextHash got ambiguous credentials (both hd and imported set)',
+    );
+  }
+
+  if (credentials.hd) {
```
originalix: I want to clarify the intended security model here. The HD branch derives the IKM from the wallet seed at a fixed path, so the output is effectively wallet-level: all BTC accounts / address types under the same seed produce the same result for the same appName and context. The imported branch uses the raw imported private key, so its output is account-key-level instead.

Is this asymmetry intentional? If the API is meant to be account-level, the HD derivation may need to include the selected account path or another account-specific domain separator. If it is meant to be wallet-level, the API contract should make that explicit and confirm that raw imported private keys are the intended equivalent for imported accounts.

packages/kit-bg/src/providers/ProviderApiBtc.ts
```diff
@@ -935,6 +940,91 @@ class ProviderApiBtc extends ProviderApiBase {
...
+  /**
+   * @experimental Derive a deterministic 32-byte value from the wallet's key
+   * material via HKDF-SHA-256. See `core/src/chains/btc/sdkBtc/deriveContextHash.ts`
+   * for the algorithm and validation rules.
+   *
+   * Validation here mirrors the core validators word-for-word so invalid
+   * params fail fast — before showing any approval UI to the user.
+   */
+  @providerApiMethod()
+  public async deriveContextHash(
```
originalix, line 951: TODO: publish cross-inpage-provider sdk & bump version

packages/kit/src/views/DAppConnection/pages/DeriveContextHashModal.tsx
jrwbabylonlab: Good call — accepted. Pushed an update that adds `<DAppAccountListStandAloneItemReadonly accountId={accountId} networkId={networkId} />` at the top of the modal body, mirroring the SignMessageModal pattern. Picked the readonly-with-explicit-IDs variant since this approval has explicit `accountId`/`networkId` from the request, rather than relying on the connected-account selector context.

Not migrating to the `SignatureConfirm` route in this PR — kept under `DAppConnection` to match the existing dApp-approval flows (Nostr, Clipboard, Cosmos, WeblnSendPayment) which all use this same readonly account-list component.

packages/core/src/chains/btc/sdkBtc/deriveContextHashFromCredentials.ts
jrwbabylonlab: updated the code to be leaf based

PR #11474, opened 2026-05-07
fix: tray menubar polish — pending tx refresh, sign prefix, empty watchlist flash (OK-54085 OK-54086 OK-54088)

originalix:
## Summary
- Refresh tray pending tx state when home polling confirms it (OK-54085).
- Restore the sign prefix for multi-asset pending tx amounts in tray (OK-54086).
- Preserve cached watchlist during account switch so the empty-state CTA no longer flashes (OK-54088).

## Intent & Context
Three QA-reported issues in the macOS Menu Bar tray (v6):

- **OK-54085** — A pending tx confirmed by the home page's polling never made it back to the tray, so the tray kept showing it as Pending until the next tray-side gather.
- **OK-54086** — Multi-asset pending txs (e.g. swaps, multi-token sends) lost their `+` / `-` prefix, making the direction ambiguous.
- **OK-54088** — Switching to an account that has pending txs briefly showed the WatchlistTickers empty state (the "Add favorites" CTA) before real data populated.

This PR closes out the v6 tray polish round.

### OK-54088 (this commit)
OK-53623 introduced an optimistic placeholder on account switch that wipes balance, watchlist, and pending txs to clear the previous account's data within one frame. But `watchlist` is a **global** market-favorites list (`SimpleDbEntityMarketWatchListV2`), not account-specific — clearing it forced `WatchlistTickers` into its empty-state branch (`tickers.length === 0` → "Add favorites" CTA). On accounts with pending txs the gather is slower because `refreshTrayPendingTxStatuses` adds a network round-trip, so the empty state was visible long enough for QA to catch.

### OK-54085 / OK-54086
See individual commit messages and prior conversation context.

## Design Decisions

For OK-54088:

- **Preserve watchlist, not balance/pendingTxs.** Watchlist is global; the next gather will return the same items so reusing the cached value is safe. Balance and pending txs are account-specific — preserving them across a switch would briefly show the previous account's data, which is worse than a `0.00` / empty placeholder.
- **Cache the watchlist in the renderer (`lastWatchlistRef`), not in the main process.** The optimistic placeholder is built in the renderer, so the data needs to be reachable there. The main process's `cachedTrayData` is not exposed back over IPC.
- **Only update the cache on a non-throwing fetch.** If `getMarketWatchListV2` succeeds with an empty list, the ref captures the genuinely-empty state. If it throws, keep the previous cached value so transient errors don't blank the list on the next switch.

Alternatives considered:

- Caching everything (balance, pending txs) too — rejected because account-specific stale data would be misleading.
- Letting the main process inject the cached watchlist into the optimistic payload — rejected as more invasive (would need a new IPC channel) for no extra benefit.

## Changes Detail

`packages/kit/src/hooks/useTrayDataProvider.desktop.ts`:

- Add `lastWatchlistRef` to remember the last successful watchlist.
- Update the ref inside the watchlist try block (after `trayData.watchlist` is set), so it captures both populated and genuinely-empty results but skips error paths.
- Use `lastWatchlistRef.current` instead of `[]` in the account-switch optimistic placeholder.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Desktop (macOS Menu Bar tray only — `platformEnv.isDesktopMac` gate, code path is `.desktop.ts`)
- **Risk Areas**: The optimistic-placeholder path is only taken on account switch; the change is additive (new ref, swap one literal `[]` for the ref). pending tx and balance handling are unchanged.

## Test plan
- [ ] On macOS, open the menu bar tray with at least one watchlist item.
- [ ] Switch between accounts; the watchlist rows should remain visible — no "Add favorites" CTA flash.
- [ ] Switch to an account with pending txs; verify the watchlist still doesn't flash and pending txs render after the gather completes.
- [ ] Switch to a new account on a wallet whose watchlist genuinely is empty; verify the CTA still shows correctly (no false-positive caching).
- [ ] Verify OK-54085 and OK-54086 fixes from prior commits still work end to end.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/74fb5da2-d1be-463c-9dfa-5ebbcd6bb842?refs=9689e99fcd4e63e3a763052c062345148580b510&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/74fb5da2-d1be-463c-9dfa-5ebbcd6bb842/license?refs=9689e99fcd4e63e3a763052c062345148580b510&source=prChecksComment) |

huhuanming (approved):

zhaono1 (approved):

PR #11580, opened 2026-05-12
fix: stabilize Android network selector initial scroll (OK-54208)

weatherstar:
## Summary
- Fix Android FlashList re-anchoring the network/chain selector to the top after the initial sections data lands asynchronously, causing the user's selected network to be off-screen.
- Extract the stabilization logic into a shared `useAndroidFlashListInitialScrollFix` hook and apply it to both `EditableChainSelectorContent` and the pure `ChainSelectorSectionList`.

## Intent & Context
On Android, opening the network selector with a non-top selection (e.g. an item far down the list) would briefly land at the correct position and then jump back to the top once the async sections data finished loading. This is a FlashList v2 behavior on Android where `maintainVisibleContentPosition` re-anchors the viewport after the committed dataset replaces the initial one. iOS and web were not affected.

## Root Cause
FlashList v2 on Android re-anchors scroll position after the async data swap that replaces the initial empty/placeholder list. `initialScrollIndex` is honored on first mount, but the later re-anchor pass moves the viewport back to index 0, defeating the intended initial position.

## Design Decisions
- **Disable `maintainVisibleContentPosition` on Android only.** iOS/web rendering is unaffected, so the prop override is gated behind `platformEnv.isNativeAndroid`. Web ignores this prop entirely (it's a SectionList/FlatList there), but we still keep the override Android-only to avoid any cross-platform regressions.
- **Re-issue `scrollToIndex` after the data settles.** A `setTimeout(0)` catches the immediate post-commit frame and a second timer at 120ms covers FlashList's later internal re-anchor. Two-shot is intentional: a single immediate call loses to the later re-anchor, and a single delayed call leaves a visible flash.
- **Bail out once the user scrolls.** `onScrollBeginDrag` flips `hasUserScrolledRef` so the restore timers don't fight a deliberate user gesture if the sections data updates mid-scroll.
- **Extract into a shared hook.** The same fix was needed in two list components (`EditableChainSelectorContent` and `ChainSelectorSectionList`); duplicating the constant, ref, callback, and double-`setTimeout` effect across both was error-prone, so the second commit consolidates them into `useAndroidFlashListInitialScrollFix`.

## Changes Detail
- `packages/kit/src/views/ChainSelector/hooks/useAndroidFlashListInitialScrollFix.ts` — New hook. Returns Android-only `scrollProps` (disabling `maintainVisibleContentPosition`) and an `onScrollBeginDrag` handler; runs a two-shot `scrollToIndex` effect keyed off `contentKey` so each async data swap re-pins to `initialIndex` until the user scrolls.
- `packages/kit/src/views/ChainSelector/components/EditableChainSelector/ChainSelectorContent.tsx` — Wire the hook into the editable selector, computing the flat initial index from the section/item indices and gating on `shouldRenderList && !searchText.trim()` so search results don't get re-pinned.
- `packages/kit/src/views/ChainSelector/components/PureChainSelector/ChainSelectorSectionList.tsx` — Wire the hook into the pure selector, threading `androidScrollProps` and `onScrollBeginDrag` to the inner list content; gated on `!loading && !text.trim() && sections.length > 0`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Android (no behavior change for iOS, Web, Desktop, Extension — the hook returns an empty scroll-props object and a no-op effect off-Android)
- **Risk Areas**: The two-shot timer in `useAndroidFlashListInitialScrollFix`. If `initialIndex` changes mid-effect, the cleanup clears pending timers and the new value re-runs the effect; `hasUserScrolledRef` prevents fighting a user gesture. Worth a manual check that searching/filtering during initial load doesn't cause an unwanted snap-back.

## Test plan
- [ ] Android: open network selector with a network selected that sits far down the list — selected item should remain visible (no top-snap) after async data loads.
- [ ] Android: open network selector, immediately drag-scroll before data finishes loading — list should respect the user drag rather than snapping back to the initial index.
- [ ] Android: type in the search box while the list is settling — search results should appear without the hook trying to re-pin to the original index.
- [ ] Android: open both the editable chain selector and the pure chain selector flows to confirm both code paths apply the fix.
- [ ] iOS / Web: confirm no regression — initial scroll position still respected, no extra jumps.

### DIFF
--- a/packages/kit/src/views/ChainSelector/components/EditableChainSelector/ChainSelectorContent.tsx
+++ b/packages/kit/src/views/ChainSelector/components/EditableChainSelector/ChainSelectorContent.tsx
@@ -2,6 +2,7 @@ import type { Dispatch, SetStateAction } from 'react';
 import {
   useCallback,
   useContext,
+  useEffect,
   useLayoutEffect,
   useMemo,
   useRef,
@@ -44,6 +45,12 @@ import type {
 // below for why we use spread-cast instead of naming the prop directly.
 const flashListOverrideProps = {
   overrideProps: { initialDrawBatchSize: 30 },
+};
+
+const androidFlashListScrollProps = {
+  maintainVisibleContentPosition: {
+    disabled: true,
+  },
 };
 
 const ListEmptyComponent = () => {
@@ -157,6 +164,7 @@ export const EditableChainSelectorContent = ({
   );
   const listRef = useRef<ISectionListRef<any> | null>(null);
   const hasRenderedSectionListRef = useRef(false);
+  const hasUserScrolledRef = useRef(false);
 
   useLayoutEffect(() => {
     setTempFrequentlyUsedItems(frequentlyUsedItems);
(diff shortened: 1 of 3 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0568110a-e151-4776-99dc-965c3b80ac06?refs=d799c11215bc9929bdfaad1e49d5441af1f15ac7&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0568110a-e151-4776-99dc-965c3b80ac06/license?refs=d799c11215bc9929bdfaad1e49d5441af1f15ac7&source=prChecksComment) |

originalix (approved):

PR #11615, opened 2026-05-14
fix: show error view on desktop browser network failures (OK-54656)

limichange:
## Summary
- Restore the error/refresh page on Desktop browser when a network failure happens (e.g. unreachable `127.0.0.1:8888`, offline) instead of showing a blank webview.
- Stabilize desktop tab switching so opening a dApp no longer flashes the previously active tab before showing the new destination.
- Harden URL validation/normalization so localhost & public IP hosts (including Chinese full-width dot variants like `127。0。0。1`) are accepted and routed through `http://` consistently.

## Intent & Context
QA reported `OK-54656`: when the address bar loads an unreachable origin such as `127.0.0.1:8888/` (or the user is offline), the Desktop browser shows an empty page instead of the existing `ErrorView` with a refresh button. Historically QA used `127.0.0.1:8888` and the offline switch as the canonical network-failure scenario, so any regression here blocks several recurring test cases.

While instrumenting the Desktop browser to reproduce this, two adjacent bugs surfaced:
1. Opening a dApp from search on Desktop produced a visible flash of the previously active tab before the new tab rendered, because `MultiTabBrowser` was switched in before the new active tab was persisted.
2. Inputs like `127。0。0。1:8888` (Chinese full-width dot) and bare public IPs (`6.6.6.6:8080`) were not normalized, so the user could not reach internal servers and developer test rigs the same way they can on other platforms.

Both regressions intersect the same fix path (URL normalization + Desktop WebView mount/unmount), so they are addressed together to keep the validation path consistent.

## Root Cause
- `DesktopWebView` cleared `desktopLoadError` from `didFinishLoad`, but Electron emits a finish event right after `did-fail-load` in some failure modes, so the error view was immediately hidden again. There was also no `bg` / `zIndex` on the absolute error layer, so even when it was shown it could be visually covered by the still-mounted webview.
- `InpageProviderWebView.desktop` was registering `onDidStartLoading` via `() => innerOnDidStartLoading` (arrow returning the handler), so the wrapper that prepares load state and timeouts never ran — meaning the load-timeout fallback that should show `ErrorView` after `WEBVIEW_LOAD_TIMEOUT_MS` never triggered.
- `handleOpenWebSite` on Desktop awaited `switchTabAsync(MultiTabBrowser)` before persisting the new active tab. `DesktopCustomTabBar` then rendered using `savedActiveTabId`, which still pointed at the prior tab, producing a flash.
- `uriUtils` only had a localhost short-circuit; bare public IPs went through the HTTPS path (which then failed validation and fell back to Google search), and full-width dot separators were only normalized inside `isLocalhostUrl`, not in `ensureHttpPrefix` / `parseDappRedirect`.

## Design Decisions
- **Track the last main-frame load error in a ref (`lastMainFrameLoadErrorRef`) and gate `setDesktopLoadError(false)` in `didFinishLoad` on that ref.** Rationale: Electron's event order isn't deterministic across navigations; using a ref keeps the source of truth out of React render scheduling so the error view doesn't flicker. Resetting happens explicitly on `did-start-loading` for the main frame and on user-initiated `reload()`.
- **Expose `errorCode` to `ErrorView` (already supported in the component) instead of replacing the error UI.** Keeps existing translations and copy intact for QA's verification flow.
- **Use `onShouldStartLoadWithRequest` in `WebContent.desktop` rather than only reacting to `did-fail-load`.** This blocks navigation to invalid/blocked URLs *before* Electron starts a request, so the user sees the BlockAccessView immediately rather than after a network timeout.
- **Introduce a new `ClearSavedBrowserActiveTab` event** instead of reaching into `DesktopCustomTabBar` from `actions.ts`. The existing pattern already uses `appEventBus` for cross-module browser signaling (`CreateNewBrowserTab`, etc.), so this stays consistent.
- **On Desktop, persist the destination tab → emit `ClearSavedBrowserActiveTab` → `switchTabAsync` (in that order).** Sequencing ensures `DesktopCustomTabBar` reads the new active tab on its first paint of MultiTabBrowser. Native platforms keep the old serialized path because they don't have the same custom tab bar.
- **Add `isIpAddressUrl` as an exported util and reuse `getHostnameFromUrlLikeText`** so the IP detection, separator normalization, and localhost detection all share one parser. Avoids divergent regex / hostname-extraction logic.
- **Treat `http://` public-IP origins as `ALLOW` in `parseDappRedirect`** (separate from the localhost branch) so QA's `6.6.6.6:8080` style test URLs flow through the normal browser path. Domains over plain HTTP still return `DENY`.
- **Restrict separator normalization to the hostname segment only** (`normalizeUrlHostnameSeparators`) — full-width characters in the path/query/fragment are preserved because they may be intentional. Only `localhost` and IP-style hostnames need ASCII dots.

## Changes Detail
- `packages/kit/src/components/WebView/DesktopWebView.tsx` — Track `desktopLoadErrorCode`, retain error state across follow-up finish events, surface `errorCode` to `ErrorView`, wrap output in a positioned `Stack` with `$bgApp` background, reset state on user reload, drop debug `console.log('preloadJsUrl', ...)`.
- `packages/kit/src/components/WebView/InpageProviderWebView.desktop.tsx` — Fix `onDidStartLoading` so it is invoked instead of returned.
- `packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx` — Add `onShouldStartLoadWithRequest`, stop & drop the webview ref when current URL becomes blocked, depend on `[id, url]` for navigation reset, render in a positioned `Stack`, enable `displayProgressBar`.
- `packages/kit/src/states/jotai/contexts/discovery/actions.ts` — Skip `crossWebviewLoadUrl` for blocked localhost URLs, make `openDApp` return `opened`, add Desktop-specific create-tab-before-switch sequencing with `ClearSavedBrowserActiveTab`, remove stale `console.log` statements.
- `packages/kit/src/views/Discovery/pages/DesktopCustomTabBar/index.tsx` — Subscribe to `ClearSavedBrowserActiveTab` to wipe `savedActiveTabId`.
- `packages/kit/src/views/Discovery/components/SearchResultContent.tsx` — On exact-URL hits for localhost/IP, normalize via `ensureHttpPrefix` so the dApp record opened by the browser carries an `http://` scheme.
- `packages/shared/src/eventBus/appEventBus.ts` / `appEventBusNames.ts` — Register `ClearSavedBrowserActiveTab` with `undefined` payload.
- `packages/shared/src/utils/uriUtils.ts` — Add `isIpAddressUrl`; share `getNormalizedParsedHostname`; introduce `normalizeUrlHostnameSeparators` (hostname only); allow `http://` for public IPs in `ensureHttpPrefix` / `validateUrl` / `parseDappRedirect`; deny `http(s)://localhost` when `allowLocalhostUrl` is false.
- `packages/shared/src/utils/uriUtils.test.ts` and `packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx` — New unit tests covering: bare public-IP HTTP, full-width dot variants of localhost, hostname-only separator normalization, dev-flag gating of localhost in `validateWebviewSrc`, and the new Desktop tab-creation-before-switch ordering.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop (Electron) primarily; `uriUtils` and `discovery/actions` also touch Mobile / Web / Extension because they are shared modules.
- **Risk Areas**:
  - `DesktopWebView` error-state lifecycle — incorrect reset could either trap the user in `ErrorView` after a successful retry or hide a real failure. Mitigated by explicit `setDesktopLoadError(false)` on `did-start-loading` and on user-driven `reload()`.
  - `parseDappRedirect` behavior change for `http(s)://localhost` when `allowLocalhostUrl` is `false` (now returns `DENY` instead of falling through). Intended, but worth verifying that no production caller relies on the previous fallthrough path (none observed; only the browser dev setting toggles this flag).
  - Public-IP HTTP allow path — limited to literal IPs (`validator.isIP`), so domain-with-IP-like-text inputs cannot exploit it.
  - Desktop `handleOpenWebSite` reordering — only Desktop branch changes; Native/Web continue to use the serialized switch-then-open path.

## Test plan
- [ ] Desktop: visit `http://127.0.0.1:8888/` while no server is running → `ErrorView` with refresh button appears (no blank page).
- [ ] Desktop: visit any reachable URL, then turn off network and reload → `ErrorView` appears; turning network back on and pressing refresh recovers.
- [ ] Desktop: from Discovery search, click an exact-URL hit for `127。0。0。1:8888` → tab opens with `http://127.0.0.1:8888/` and no previous-tab flash on transition to MultiTabBrowser.
- [ ] Desktop: from Discovery search, click an exact-URL hit for `6.6.6.6:8080` → tab opens with `http://6.6.6.6:8080`.
- [ ] Desktop: open a dApp from search while MultiTabBrowser already has another active tab → the destination tab is active immediately, no flash.
- [ ] Desktop: with `allowLocalhostUrlInDAppBrowser` dev flag OFF, attempt to navigate to `http://localhost:3000` → BlockAccessView appears immediately (no late phishing handler).
- [ ] Desktop: with `allowLocalhostUrlInDAppBrowser` dev flag ON, navigate to `http://localhost:3000` → loads normally.
- [ ] Native / Web smoke: search and open a normal HTTPS dApp → behavior unchanged.
- [ ] Unit tests pass: `yarn jest packages/shared/src/utils/uriUtils.test.ts packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx`.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2842afb4-e22d-4557-9bfc-7977c0fbacfd?refs=6b7e3f851b05dc0525367b82f307438daa8c53a1&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2842afb4-e22d-4557-9bfc-7977c0fbacfd/license?refs=6b7e3f851b05dc0525367b82f307438daa8c53a1&source=prChecksComment) |

originalix (approved):

PR #11624, opened 2026-05-14
fix: prevent stale fiat balance display after currency switch (OK-50362)

weatherstar:
## Summary
- Wipe all fiat-value caches (SimpleDB + in-memory atoms + MMKV cold-start snapshot) inside `ServiceSetting.setCurrency` so the post-restart UI shows `--` until fresh data arrives, instead of the previous currency's amount under the new symbol.
- Add `SimpleDbEntityLocalTokens.clearFiatData()` that only clears `tokenListMap` / `tokenListValue`, preserving token metadata to avoid a redundant refetch.
- Add a generic `clearContextAtomColdStartCacheByAtomKeys()` helper in `kit-bg/states/jotai/utils` that surgically removes targeted atom entries from both the in-memory map and the MMKV snapshot, with a hard-fallback that drops the whole snapshot if MMKV read/parse/write fails.

## Intent & Context
User report: when the user switches the default fiat currency in Settings, the currency symbol updates immediately, but the displayed amount stays the OLD currency's number for a long time after the app restart — until the next token fetch returns with values in the new currency. Net effect: e.g. switching USD → EUR shows `€123.45` where `123.45` is still the USD figure.

Goal: drop every cache that holds a currency-baked fiat number before the existing 100ms-delayed `restartApp()` fires, so the home page and wallet selector fall back to the existing `--` placeholder until fresh data lands.

## Root Cause
Three layers of cache survive the currency change and feed the post-restart UI with old-currency values:

1. **SimpleDB**
   - `SimpleDbEntityAccountValue.data` — per-account fiat total, tagged with the currency at write time (not always USD).
   - `SimpleDbEntityAccountValue.all` — multi-network value (stored as USD).
   - `SimpleDbEntityLocalTokens.tokenListMap` / `tokenListValue` — per-token `fiatValue` and per-account sum, both server-baked under the currency that was active at fetch time.
2. **In-memory jotai atoms**
   - `activeAccountValueAtom`, `accountSelectorValuesMapAtom`, `accountSelectorDeFiMapAtom` — populated from the SimpleDB above.
3. **MMKV cold-start snapshot** (the real culprit on cold restart)
   - `lastConfirmedOverviewBalanceAtom`, `accountWorthAtom`, `renderedTokenListCacheAtom`, `overviewTokenCacheStateAtom`, `overviewDeFiDataStateAtom` — these context atoms persist their last value into MMKV and hydrate it before the first paint on the next launch, so the splash + home overview flash the previous currency's exact balance string.

The `<Currency sourceCurrency>` component does have FX-rate fallback, but for caches where `currency` is missing or assumed-equal-to-current (notably `tokenListMap` and the persisted balance strings), the FX path doesn't kick in — the old number is just rendered under the new symbol.

## Design Decisions
- **Wipe caches before restart, rely on existing `--` fallback** (chosen) over a "USD-canonical + always FX-convert at display" refactor. The refactor is more elegant long-term but touches the server API contract, all `sourceCurrency` annotations across the UI, and BigNumber precision handling — too big a change for this symptom-level fix. The wipe approach is small, localized, and the placeholder UX already exists.
- **No toast / loading hint** — the restart itself is the user-facing feedback; an extra toast would barely render in the 100ms window.
- **No proactive refresh on boot** — the home page and wallet selector already kick off `fetchAccountTokens` on render via existing containers, so no new boot logic is needed.
- **Preserve `localTokens.data / tokenList / smallBalanceTokenList / riskyTokenList`** (token metadata, balances) rather than calling `clearTokens()` — avoids a redundant metadata refetch and the resulting extra empty-state flash on the token list.
- **Parallelize the five cache clears** with `Promise.all` — they're independent and the user is staring at the screen waiting for the restart.
- **MMKV fallback delete** — if reading/parsing/writing the snapshot throws, drop the whole `onekey_jotai_context_atoms_snapshot` key. The cost (other non-currency cold-start values re-fetch on next launch) is much smaller than the cost of silently leaving a corrupted snapshot to hydrate stale balances.
- **`endsWith()` for scoped-key matching** — more robust than `split('::')[1]` if a future scope name ever contains `::`.

## Changes Detail
- `packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityLocalTokens.ts` — add `clearFiatData()` that resets only `tokenListMap` and `tokenListValue`.
- `packages/kit-bg/src/services/ServiceSetting.ts` — `setCurrency` now calls a new private `clearCurrencyDependentCaches()` after `resetEarnCache()`. The new method wipes both SimpleDB entries and the three in-memory atoms in parallel, then invokes the cold-start helper.
- `packages/kit-bg/src/states/jotai/utils/index.ts` — add `clearContextAtomColdStartCacheByAtomKeys(atomKeys)` that:
  - drops matching entries from the in-memory `coldStartValuesMap` and clears their dirty markers so a pending debounced flush won't write them back;
  - rewrites the MMKV snapshot with matching keys removed;
  - on any failure, falls back to dropping the entire snapshot to guarantee no stale entry survives the restart.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: iOS, Android, Desktop, Web, Extension
- **Risk Areas**:
  - The MMKV cold-start cache path is native-only; on web/desktop/ext the storage stub is a no-op so the new helper is effectively a no-op there — verified, but worth eyeballing on Extension.
  - Five caches must stay in sync going forward — if a new currency-derived cache is added later, this list needs to be extended.
  - Race window: if a token fetch with the OLD currency lands in the 100ms between `setCurrency` and `restartApp()`, it could repopulate `activeAccountValueAtom`. The hard restart tears down the JS context immediately after, so this is bounded — the worst case is one extra frame of correct-symbol-stale-amount, which is still better than current behavior.

## Test plan
- [ ] iOS — USD → EUR: after restart, first frame shows `--` on home overview and wallet selector; after token fetch, numbers correctly show EUR amounts.
- [ ] Android — same as above.
- [ ] Desktop — same; verify `desktopApi.system.reload` path.
- [ ] Web — verify the cold-start helper is a no-op (no MMKV) and the SimpleDB + atom wipes alone are sufficient.
- [ ] Extension — verify `chrome.runtime.reload()` path.
- [ ] Back-to-back currency switches (USD → EUR → JPY → USD) — no residual stale numbers, no flicker of intermediate currencies.
- [ ] All Networks vs single-network views — both should show `--` then refresh.
- [ ] Multiple wallets in selector — non-active wallets should also show `--` and independently refresh.
- [ ] Earn / staking values (already cleared via the existing `resetEarnCache()`) — confirm no regression.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e6e0a8d6-3f40-4290-9bd4-1cab42b135ed?refs=09cb605ad3ddb1d73fde2b9f78722d95b4ea4d89&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e6e0a8d6-3f40-4290-9bd4-1cab42b135ed/license?refs=09cb605ad3ddb1d73fde2b9f78722d95b4ea4d89&source=prChecksComment) |

limichange (dismissed):

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityLocalTokens.ts
```diff
@@ -289,4 +289,18 @@ export class SimpleDbEntityLocalTokens extends SimpleDbEntityBase<ISimpleDBLocal
...
     });
   }
+
+  // Wipe only currency-dependent fiat fields, preserve token metadata and
+  // per-account token lists so we don't refetch metadata after a currency switch.
+  @backgroundMethod()
+  async clearFiatData() {
+    await this.setRawData((rawData) => ({
+      data: rawData?.data ?? {},
+      tokenList: rawData?.tokenList ?? {},
```
originalix, line 307: 切换币种后，`clearFiatData()` 会保留 `tokenList`，但清空 `tokenListMap` 和 `tokenListValue`。后续 `getAccountTokenList()` 仍然只看 `tokenList` 是否有该 account/network key 来返回 `hasCache=true`，Home 初始化分支会把 fallback 的 `tokenListValue: '0'` 当成可用余额写进 `accountWorthAtom` 并标记 overview ready。

影响是重启后的首屏可能显示新币种下的 0 余额，而不是等新的 fiat map/value 拉回；这个 0 还可能继续写回 `accountValue` 缓存。

建议新增独立的 fiat readiness 状态，例如 `fiatHasCache`，并让 Home 只在 `fiatHasCache=true` 时提交 `accountWorth`。`clearFiatData()` 清空 fiat 后应返回 `fiatHasCache=false`，同时继续允许 token metadata 复用。

packages/kit-bg/src/services/ServiceSetting.ts
```diff
@@ -262,6 +267,32 @@ class ServiceSetting extends ServiceBase {
...
     await settingsPersistAtom.set((prev) => ({ ...prev, currencyInfo }));
     await this.backgroundApi.serviceStaking.resetEarnCache();
+    await this.clearCurrencyDependentCaches();
+  }
+
+  // Wipe every cache that holds fiat values baked in the previous currency, so
+  // the post-restart UI shows a placeholder until fresh data arrives instead of
+  // displaying stale numbers under the new currency symbol.
+  private async clearCurrencyDependentCaches() {
+    await Promise.all([
```
originalix, line 284: All Networks 拉取会先把旧币种的 `tokenListMap/tokenListValue` 放进 `ServiceToken.localAccountTokensCache`，再由 3 秒 debounce 写入 SimpleDB。这里切币种时只清了 SimpleDB 和几个 atom，没有取消那个 pending debounce，也没有清空内存 buffer。

如果用户在这 3 秒窗口内切换币种，`clearFiatData()` 刚清掉的 fiat 字段可能又被旧 buffer 写回，下一次启动/读取仍会拿到上一币种的余额数据。

建议在 `clearCurrencyDependentCaches()` 里先调用一个 `serviceToken` 的清理方法：cancel `_updateAccountLocalTokensDebounced`，清空 `localAccountTokensCache`，然后再清 SimpleDB 的 fiat 数据。

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityLocalTokens.ts
weatherstar: Fixed in ca18f24dd5.

`SimpleDbEntityLocalTokens.getAccountTokenList` 现在额外返回 `fiatHasCache`（基于 `tokenListValue[key]` 是否存在），区分「token metadata 有缓存」与「fiat 值有缓存」。`TokenListBlock` 在 `fiatHasCache=false` 时不再 commit `updateAccountWorth`，并把 overview 标为 `isRefreshing=true / initialized=false`，等待新币种的真实数据回来再提交 —— 不会再把 fallback 的 `0` 当成余额写入 `accountWorthAtom`。token list metadata 仍然刷新，所以 names/icons 立即可见，只是 worth 处于 loading 状态。

packages/kit-bg/src/services/ServiceSetting.ts
weatherstar: Fixed in ca18f24dd5.

新增 `ServiceToken.cancelPendingLocalTokensWrite()`：调用 `_updateAccountLocalTokensDebounced.cancel()` 取消 pending 的 3s debounce flush，并把 `localAccountTokensCache` 重置为空。`clearCurrencyDependentCaches()` 现在在清 SimpleDB fiat 数据之前先调用它，避免 in-flight 的 All Networks buffer 在 debounce 窗口内把旧币种的 `tokenListMap/tokenListValue` 写回。

weatherstar:
Superseded by a cleaner approach: normalize all fiat caches to USD basis with client-side conversion at render time, instead of clearing caches on currency switch. This avoids the UX regression where the account selector loses values for inactive accounts after switching currency. Tracking in new branch `refactor/fiat-cache-usd-basis`.

PR #11641, opened 2026-05-14
fix(android): patch fast-deep-equal to guard non-callable toString (OK-54522  OK-54696)

huhuanming:
## Summary

Hermes on Android 0.81.5 **lacks native `Array.prototype.toSorted`** (we verified by inspecting the shipped `libhermes.so` builtin string blob), so `polyfillsPlatform.js:24-26`'s `shim()` installs the `array.prototype.tosorted` polyfill — which threads through `es-abstract`'s `IsLessThan` → `ToNumeric` → `ToPrimitive`. When React Navigation's `useLinking` does `isEqual(focusedRoute.params, route.params)` and walks into a nested object whose `toString` is non-callable (e.g. `Object.create(null)` records from i18n/serializer paths), `fast-deep-equal/index.js:26` blindly calls `a.toString()` → `TypeError: undefined is not a function` → `ExceptionsManagerModule.reportException` kills the app.

This PR patches the library at line 26 with a `typeof === 'function'` guard plus a diagnostic logger so:

1. **The app no longer crashes** on that specific code path.
2. **Future occurrences leave a `[fast-deep-equal][BUG]` line** in the file logger with the offending object's `typeof`, `constructor.name`, and key list — so we can chase the dirty data back to its source.

It also bundles in a few orthogonal Android stability/CI fixes that were already in-flight on `fix/android-apk-resume-download`.

## Evidence chain

Sourcemap-resolved stacks from real device logs (build 6.3.0 / buildNumber 2026051318 / bundleVersion 11426912):

| Source | Frame 0 (deepest) | Frame 1 (caller) |
|---|---|---|
| logcat × 1 (Permit / DApp signing) | `array.prototype.tosorted/.../ToNumeric.js:18` | `@react-navigation/core/.../getStateFromPath.js:286` |
| app-latest.log × 4 (Sign & Verify) | `math-intrinsics/maxArrayLength.js:4` (GLB fallback) | `fast-deep-equal/index.js:26` |

Both stacks land inside `@react-navigation/core`'s `useLinking` effect. The 4 file-logger fatals were preceded by `ui => button => click trackId=wallet-action-sign-and-verify` + `app => page => pageView pageName=SignAndVerifyMessage`; the logcat fatal was preceded by an `EarnProtocolDetails → page-footer-confirm` flow that hits `MessageConfirm` (EIP-712 Permit) via `StackActions.replace`.

## What changed direction

The first iteration of this branch (`80a1d651f0`) tried to delete the dead `deriveInfoItems` / `deriveType` route params on the `SignAndVerifyMessage` modal. Verified live build `11540565` no longer crashed on that path, but **the same root cause hits `MessageConfirm` (DApp Permit) through `unsignedMessage` and friends**, which cannot be deleted — receiver needs the data. So that approach was reverted in `eb5be4be54` and replaced with a library-level guard that covers every caller in the bundle, including future ones we haven't catalogued.

Verified on build `11543414` (`25865818927`):
- Click **wallet-action-sign-and-verify** → `SignAndVerifyMessage` opens, 4s on page, no crash (previously: ~1s crash).
- EarnProtocols → page-footer-confirm → `MessageConfirm` opens, 7s on page, no crash (previously: ~1-2s crash).

### Main fix (patch + diagnostic)

| File | What |
|---|---|
| `patches/fast-deep-equal+3.1.3.patch` | New patch. Guard line 26 with `typeof a.toString === 'function' && typeof b.toString === 'function'`; when guard fires, write `[fast-deep-equal][BUG]` to NativeLogger via `globalThis.$$appGlobals.$defaultLogger.app.error.log`. Outer try/catch ensures missing globals never break the equality short-circuit. |
| `packages/shared/src/appGlobals.ts` | Expose `globalThis.$$appGlobals = appGlobals` in production too (was previously dev-only). Required so the patch — bundled into `common.bundle` — can reach `defaultLogger` without a cross-bundle require that the split-bundle DAG forbids. |

The patch was generated automatically (no manual editing) via `npm pack fast-deep-equal@3.1.3` + `diff -u` against the modified file, bypassing `patch-package`'s internal `yarn install` step which can't parse OneKey's `"@ledgerhq/errors@^6.33.0": "6.33.0"`-style resolutions. The output is patch-package-format-compatible (verified by round-trip `restore → patch-package → diff -q`).

### Sentry release unification (bundled)

| File | What |
|---|---|
| `packages/shared/src/modules3rdParty/sentry/index.native.ts` | Pass explicit `release` + `dist` to `Sentry.init()` so runtime events tag with `<version>+<buildNumber>+<bundleVersion>` — the same identifier `apps/mobile/build-bundle-base.js`'s `uploadDirectoryToSentry` uses. Previously: metro plugin auto-tagged events with `so.onekey.app.wallet@<v>+<n>`, but our maps were uploaded under the shared `<v>+<n>+<bundleVersion>` release → maps never matched events → frames stayed unsymbolicated. |

### Android stability (merged from `fix/android-apk-resume-download`)

| File | What |
|---|---|
| `packages/components/src/forms/Input/index.tsx` | Guard `ref.current?.setSelection?.(0, 0)` so a wrapped `TextInput` whose `setSelection` is undefined no-ops instead of throwing. |
| `packages/components/src/forms/TextArea/Input.tsx` | Drop the `useAutoScrollToTop(ref)` call from `BaseTextArea` — the only path that reached a TextArea-wrapped ref. |

Plus the existing in-flight commits on that branch (APK resume after network reconnect, offline-wait retry log surfacing, redundant retry path cleanup).

## Test plan

- [ ] Android release build: tap **Sign and Verify** on Home → `SignAndVerifyMessage` modal opens, ≥3s on page, no crash, Sign flow still completes end-to-end.
- [ ] Android release build: DApp Permit flow (e.g. USDC permit to Morpho via earn) → `MessageConfirm` opens, ≥3s on page, no crash, signing works.
- [ ] Android release build: trigger the actual non-callable toString path → confirm `[fast-deep-equal][BUG]` appears in `file-logger` output with offending `constructor.name` / `keys(a)`.
- [ ] iOS release build: confirm same flows work (Hermes shipped with RN 0.81.5 also lacks native `toSorted` on iOS, so the polyfill is active on both platforms — the patch is platform-agnostic).
- [ ] Sentry: events from this build land under the unified release `6.3.0+<buildNumber>+<bundleVersion>` with auto-symbolicated frames.
- [ ] APK resume-after-reconnect flow (covered by `fix/android-apk-resume-download` portion) still works.
- [ ] `yarn tsc:staged` + `yarn lint:staged` pass.

## Refs

- React Navigation `useLinking.tsx:199` deep-equals `focusedRoute.params` vs `route.params` on every nav state change when `linking.enabled === true` (OneKey config does enable linking).
- `useNavigationBuilder.tsx:645` also calls deepEqual but is gated by `UNSTABLE_routeNamesChangeBehavior === 'lastUnhandled'`, which OneKey does not set.
- `fast-deep-equal@3.1.3` line 26's `Object.create(null)` pitfall is upstream-known but unfixed.
- iOS vs Android: we verified both Hermes binaries lack native `Array.prototype.toSorted` (the builtin string blob is identical and missing the symbol). iOS appearing "unaffected" so far is most likely selection bias on which dApp/account triggered the dirty data — the patch protects both platforms identically.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/33d4c713-5757-4aa7-8a4b-e186f26235a8?refs=9e70e9ae3a16b3cf9eccf652480e15128534f763&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/33d4c713-5757-4aa7-8a4b-e186f26235a8/license?refs=9e70e9ae3a16b3cf9eccf652480e15128534f763&source=prChecksComment) |

packages/kit/src/views/Home/components/WalletActions/WalletActionSignAndVerify.tsx
```diff
@@ -52,8 +45,6 @@ export function WalletActionSignAndVerify({
         accountId: account?.id,
         walletId: wallet?.id,
         indexedAccountId: indexedAccount?.id,
-        deriveInfoItems,
-        deriveType,
```
originalix, line 56: 感觉这样改不对

不应该把派生参数删除

originalix (changes requested):

packages/kit/src/views/Home/components/WalletActions/WalletActionSignAndVerify.tsx
originalix: 会让逻辑发生变化

PR #11478, opened 2026-05-07
build(cli): notarize macOS OneKey CLI standalone package

sidmorizon:
## Summary
- Enable Developer ID Hardened Runtime signing and Apple notarization for the macOS standalone OneKey CLI workflow.
- Sign the embedded @napi-rs/keyring native binding before SEA embedding and add CLI-specific Hardened Runtime entitlements.
- Produce notarizable ZIP and npm tarball artifacts, then verify raw, ZIP-contained, and npm-tarball-contained binaries before upload.

## Intent & Context
The standalone macOS CLI now runs as the OneKey CLI identity for Keychain access, instead of being attributed to node or security. The remaining launch problem was Gatekeeper blocking binaries downloaded from GitHub artifacts, Slack, or browser-based channels. The goal is to make the CI-produced macOS CLI artifact acceptable to Gatekeeper without requiring extra Apple App ID, provisioning profile, CloudKit, sandbox, or keychain access group setup.

## Root Cause
The previous workflow only performed codesign --verify. That validates the code signature structure but does not prove Apple notarization has accepted the artifact or that a quarantined download can execute. The standalone CLI workflow did not submit the signed artifact to Apple notarization and did not run a quarantined first-run check against the packaged artifacts.

## Design Decisions
- Reuse the existing desktop Developer ID and notarization secrets: DESKTOP_KEYS_SECRET, CSC_KEY_PASSWORD, APPLEID, APPLEIDPASS, and ASC_PROVIDER.
- Keep the CLI as a plain Mach-O executable with signing identifier so.onekey.cli. No Apple App ID or provisioning profile is needed because this CLI does not use CloudKit, App Sandbox, app groups, or a Keychain access group.
- Enable Hardened Runtime for notarization and add CLI-specific entitlements for Node/V8 and native addon loading.
- Sign a copied @napi-rs/keyring native .node file before embedding it into the SEA executable so the extracted native binding is not left as an unsigned dependency under the standalone CLI flow.
- Make the packaging script fail if the raw executable, distribution ZIP, or npm tarball contains a missing, non-executable, wrong-architecture, unsigned, or NODE_OPTIONS-unpatched binary.
- Submit the ZIP to notarytool, then verify the online notarization ticket for raw, ZIP-contained, and npm-tarball-contained binaries before upload. A plain binary and ZIP cannot be stapled like an app, pkg, or dmg; a stapled pkg or dmg can be added later if offline first-run support is required.
[...]

PR #11650, opened 2026-05-15
build(cli): publish signed macOS standalone via platform subpackages

originalix:
## Summary
- Wire morizon's signed macOS standalone CLI (#11478) into the npm publish flow as a platform-split package, so `npm i -g @onekeyfe/cli` on macOS pulls a notarized Mach-O binary with its own process identity (`so.onekey.cli`) instead of a Node-subprocess wrapper.
- Re-architect `cli-publish.yml` as the single dispatch entry point: `get-version` → `standalone` matrix (publishes per-arch subpackages) → guarded main-package publish.
- Source repo no longer carries `optionalDependencies`; they're injected into the published tarball at pack time to avoid a yarn 4 chicken-and-egg with not-yet-published subpackages.

## Intent & Context

In Slack thread https://onekeygroup.slack.com/archives/C022268A5EW/p1778146077821629, zuozhuo asked for this work after merging #11478. The thread surfaced a real security issue with the npm-distributed CLI as it shipped: it was a JS bundle launched by `node`, so when it wrote to the macOS keychain the item was attributed to the `security`/`node` namespace rather than a dedicated process identity. Any external process invoking `security find-generic-password` could read OneKey's stored secrets without prompting the user.

The fix already existed in #11478 (signed, notarized macOS standalone Mach-O with bundle id `so.onekey.cli`) but its `cli-macos-standalone.yml` workflow only uploaded build artifacts — it was never wired into the npm publish path. This PR closes that gap.

## Root Cause

Two non-obvious traps were discovered mid-implementation (both cost a full CI round to confirm — visible in the commit history):

1. **Yarn 4 hard-fails on `optionalDependencies` 404s** (unlike npm, which silently skips). If we declared `@onekeyfe/cli-darwin-arm64` in source `apps/cli/package.json`, the `standalone` job itself — which needs to publish that subpackage — could never run `yarn install --immutable` to bootstrap, because the subpackage doesn't exist on npm until that job publishes it.

2. **Guards and injectors sharing a stale source of truth.** The first version of the pre-publish guard read subpackage names from `package.json.optionalDependencies`, but the fix for (1) removed that field from source — the guard then read empty and failed. Plus, retrying after a half-failed pipeline crashed because the standalone publish step wasn't idempotent against npm's no-overwrite rule.

## Design Decisions

1. **Reusable workflow (`workflow_call`) over inlining or cross-workflow scheduling.** `cli-macos-standalone.yml` gained a `workflow_call` entry mirroring its `workflow_dispatch` inputs. `cli-publish.yml` calls it via `uses: ./.github/workflows/cli-macos-standalone.yml`. From the release operator's view this is single-dispatch; from the maintainer's view standalone logic stays a single source of truth that morizon can still dispatch independently. Inlining was considered and rejected (would duplicate 150+ lines and remove the ability to test standalone packaging in isolation).

2. **Platform-split package** (`@onekeyfe/cli-darwin-arm64`, `@onekeyfe/cli-darwin-x64`) using npm's `os`/`cpu` filtering + `optionalDependencies`. This is the industry-standard layout (esbuild, swc, @rollup/rollup-darwin-arm64). The main package's `bin/onekey` is a Node shim that `require.resolve`s the platform subpackage and `spawnSync`s the real binary.

3. **Inject `optionalDependencies` at pack time, not in source.** Source `apps/cli/package.json` stays free of platform deps — yarn installs cleanly everywhere, no lockfile churn. `cli-publish.yml` writes `optionalDependencies` into the package.json right before `npm pack`, so only the published tarball carries the routing metadata. Subpackage names come from a single job-level `PLATFORM_SUBPKGS` env that both the guard and the inject step read from.

4. **Fail-loud on macOS, fallback on other platforms.** macOS arm64/x64: if the platform subpackage is missing (e.g. `--no-optional`, network hiccup, npm cache corruption), the shim exits 1 with an actionable error — never silently downgrade a macOS user to the weak-keychain JS-bundle path. Linux/Windows: fall back to `dist/cli.js` with a stderr warning, because standalone packaging isn't built for those platforms yet.

5. **Idempotent subpackage publish + pre-publish guards.** Guard 1 refuses to re-publish a main version that's already on npm. Guard 2 verifies every platform subpackage is on npm at the same version before publishing the main package. The standalone publish step does an `npm view` check before `npm publish` and skips if the version already exists — so retries after a partial pipeline failure don't require burning a version bump.

6. **`dist-tags.latest` deliberately unchanged.** It still points at `0.1.0-alpha.0`. The alpha line lives under `dist-tags.next`, so default `npm i -g @onekeyfe/cli` (which resolves `latest`) doesn't auto-pull alpha builds. `latest` will flip when the first stable release is cut.

## Changes Detail

- **`apps/cli/bin/onekey`**: rewritten from a 5-line `require('../dist/cli.js')` into a platform router. macOS → resolve `@onekeyfe/cli-darwin-${arch}` and `spawnSync` its bin; other platforms → fall back to JS bundle with stderr warning; macOS missing subpackage → fail-loud with reinstall hint + issue link.
- **`apps/cli/package.json`**: never carries `optionalDependencies` in source. Version bumped through alpha.6 → alpha.9 across iterations.
- **`apps/cli/scripts/bump-version.js`**: unchanged from baseline. An earlier attempt to also sync platform subpackage versions in this script was reverted once `optionalDependencies` moved out of source.
- **`.github/workflows/cli-macos-standalone.yml`**: adds `workflow_call:` entry mirroring `workflow_dispatch:` inputs (callable as reusable workflow without losing independent dispatchability). The npm publish step is now idempotent — `npm view` check before `npm publish`.
- **`.github/workflows/cli-publish.yml`**: orchestrates the full release. `get-version` (with `npm_tag` output) → `standalone` (reusable, matrix-parallel arm64+x64, publishes subpackages) → `publish` (guards + inject `optionalDependencies` + `npm pack` + publish main). Both guard and inject read from a single `PLATFORM_SUBPKGS` env.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: macOS CLI installs (new behavior); Linux/Windows CLI installs (unchanged — JS bundle fallback explicitly)
- **Risk Areas**:
  - macOS users upgrading from `alpha.4` (JS bundle, shared keychain) to `alpha.8+` (standalone, isolated keychain) will see a fresh keychain access prompt the first time the new binary touches keychain items written by the old form. Existing entries created by `alpha.4` carry an access ACL that doesn't include `so.onekey.cli`, so macOS requires explicit user confirmation. This is the intended security behavior, but worth flagging in release notes.
  - Plain Mach-O + zip artifacts cannot be stapled (Apple stapler only works on `.app`/`.pkg`/`.dmg`). First launch on a fresh machine requires online notarization ticket lookup via Gatekeeper. Offline first-run is currently unsupported — to be addressed when we add `.pkg` packaging.
  - Linux/Windows installs still ship the old JS-bundle form. The stderr warning is easy to miss when stdout is piped. When standalone packaging lands for those platforms, the warning will become a hard fail.

## Test plan

- [x] CI: full pipeline green end-to-end at `0.1.0-alpha.8` — https://github.com/OneKeyHQ/app-monorepo/actions/runs/25844130791
- [x] npm registry: all three packages (`@onekeyfe/cli`, `@onekeyfe/cli-darwin-arm64`, `@onekeyfe/cli-darwin-x64`) at `0.1.0-alpha.8`; `dist-tags.next` advanced; `dist-tags.latest` unchanged.
- [x] Local install verification on darwin-arm64: `npm i -g @onekeyfe/cli@next` installs main + arm64 subpackage only; `@onekeyfe/cli-darwin-x64` correctly skipped via `cpu` filtering. `onekey --version` runs without the JS-bundle fallback warning. `codesign -dv` confirms `Identifier=so.onekey.cli`, `TeamIdentifier=BVJ3FU5H2K`, Authority chain to Apple Root CA, notarization valid.
- [x] Functional keychain isolation: `onekey auth login --app-transfer` writes a keychain item named `onekey-cli`. Keychain Access.app shows Access Control set to "Confirm before allowing access" with only the `onekey` binary in the allow list — external processes calling `security` are blocked from silent reads. Confirmed by @zuozhuo in thread.
- [x] Uninstall / reinstall cycle: clean removal via `npm uninstall -g @onekeyfe/cli`; fresh install routes correctly without leftover state.

Refs:
- Slack thread: https://onekeygroup.slack.com/archives/C022268A5EW/p1778146077821629
- Upstream signing/notarization PR: #11478


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c68da5b8-0ddc-4e2b-ba12-67417bfa9279?refs=21e4f5f7ad53ed6e68f49dda0999f0650e83e46a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c68da5b8-0ddc-4e2b-ba12-67417bfa9279/license?refs=21e4f5f7ad53ed6e68f49dda0999f0650e83e46a&source=prChecksComment) |

limichange (approved):

huhuanming (approved):

PR #11668, opened 2026-05-15
fix: scope BTC reward history by instance only (OK-54480)

originalix:
## Summary
- Drop `walletAddress` from BTC reward history query — backend now scopes purely by `X-Onekey-Instance-Id` header.
- Remove `useActiveAccount` / `AccountSelectorProviderMirror` dependency from the page (no longer needed).
- `IBtcRewardHistoryParams.walletAddress` removed from the shared type.

## Intent & Context
Bug OK-54480 reports: after switching accounts in the redemption history page, BTC reward records disappear; another client with the same wallet only sees the currently-selected account's records.

Root issue is twofold:
1. The page was passing the current account's Base address as `walletAddress`, so any account switch changed the query and the list emptied.
2. Even with the existing backend `OR { instanceId, walletAddress }` fallback (added in OK-54113), the wallet-address branch let any caller probe arbitrary addresses they happen to know — a privacy leak (`security: []` on this endpoint, no ownership proof on the param).

After discussion we accepted the boundary: **anonymous users only see records from the current device.** That's the natural scope for a non-authenticated, write-anywhere endpoint, and it cleanly resolves both the "switch account loses records" bug and the address-probe leak.

## Root Cause
`RedemptionHistory.tsx` derived `walletAddress` from the active account's Base derivation and passed it to `btcRewardHistory({ walletAddress, pageSize })`. The backend filter (`server-service-rebate/src/entity/btc-reward/service/btc-redemption.service.ts`) treats a present `walletAddress` as an OR clause against `instanceId`, but the per-address branch is what allowed the cross-account probe. By omitting `walletAddress` entirely the backend falls into its `else` branch — `filter.instanceId = instanceId` — which is what we want.

## Design Decisions
- **Considered scoping by `userId` (OneKeyId)**: cleanest cross-device story, but BTC reward endpoints are explicitly anonymous (`security: []`), and the schema currently stores no `userId` on records. Out of scope.
- **Considered backend-side `instanceId` cluster expansion via `walletAddress` lookup**: rejected — single-hop expansion is incomplete (cannot traverse the full bipartite graph), still leaves fresh accounts empty, and turns the existing per-address leak into a per-device-cluster leak. Real privacy regression.
- **Considered passing `walletAddress[]` of every derived account**: viable but adds 50–300ms first-load BIP32 derivation cost on HD wallets and still has the address-probe leak unless paired with ownership proofs. Defer until a concrete cross-device requirement appears.
- **Chose: drop `walletAddress` from the request entirely.** Smallest diff, no schema change, no privacy regression, immediately fixes the reported account-switch bug.

## Changes Detail
- `packages/shared/src/referralCode/type.ts` — remove `walletAddress` from `IBtcRewardHistoryParams`.
- `packages/kit/src/views/Redemption/pages/RedemptionHistory.tsx` — drop the address-derivation `usePromiseResult` block, the `useActiveAccount` hook, the `AccountSelectorProviderMirror` wrapper, the `baseNetworkId` constant and related imports; BTC reward call now only passes `{ pageSize: 100 }`; `isLoading` no longer waits on address derivation.
- `packages/kit-bg/src/services/ServiceReferralCode.ts` — no edit; signature inherits from the trimmed shared type automatically.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: All (Extension / Mobile / Desktop / Web — single shared page)
- **Risk Areas**:
  - Backend OAS may still mark `walletAddress` as required on `GET /rebate/v1/btc-reward/history`. Schema declares it `required: false`, but if the route validator rejects missing param the request will 400. Needs verification on staging before relying on the auto-merge.
  - Records committed before `instanceId` was stored on `btc_redemption_codes` (field is `required: false` in schema) will not surface for any client. Pre-existing data hole, not introduced by this PR.

## Test plan
- [ ] Open Settings → Redeem → History on a device that has redeemed BTC reward; confirm the records list is non-empty.
- [ ] Switch the active account from the account selector; the BTC reward rows should remain visible (this is the OK-54480 regression).
- [ ] Switch wallets entirely; same — rows persist as long as you stay on the same device/install.
- [ ] On a fresh install (different `instanceId`), import the same wallet; confirm the BTC reward list is empty (expected new boundary; legacy OneKeyId-scoped rows still appear if logged in).
- [ ] Network tab: verify the request to `/rebate/v1/btc-reward/history` no longer includes `walletAddress` in the query string and returns 200.
- [ ] Legacy redemption rows (GiftOutline icon) still load when OneKeyId is logged in and disappear when logged out — unchanged behavior.

### DIFF
--- a/packages/kit/src/views/Redemption/pages/RedemptionHistory.tsx
+++ b/packages/kit/src/views/Redemption/pages/RedemptionHistory.tsx
@@ -4,13 +4,10 @@ import { useIntl } from 'react-intl';
 
 import { Badge, Empty, Page, Spinner, YStack } from '@onekeyhq/components';
 import backgroundApiProxy from '@onekeyhq/kit/src/background/instance/backgroundApiProxy';
-import { AccountSelectorProviderMirror } from '@onekeyhq/kit/src/components/AccountSelector';
 import { ListItem } from '@onekeyhq/kit/src/components/ListItem';
 import { useOneKeyAuth } from '@onekeyhq/kit/src/components/OneKeyAuth/useOneKeyAuth';
 import useAppNavigation from '@onekeyhq/kit/src/hooks/useAppNavigation';
 import { usePromiseResult } from '@onekeyhq/kit/src/hooks/usePromiseResult';
-import { useActiveAccount } from '@onekeyhq/kit/src/states/jotai/contexts/accountSelector';
-import { getNetworkIdsMap } from '@onekeyhq/shared/src/config/networkIds';
 import { ETranslations } from '@onekeyhq/shared/src/locale';
 import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
 import { EBtcRewardStatus } from '@onekeyhq/shared/src/referralCode/type';
@@ -20,15 +17,12 @@ import type {
 } from '@onekeyhq/shared/src/referralCode/type';
 import { EModalReferFriendsRoutes } from '@onekeyhq/shared/src/routes';
 import { formatDate } from '@onekeyhq/shared/src/utils/dateUtils';
-import { EAccountSelectorSceneName } from '@onekeyhq/shared/types';
 
 import {
   formatUsd,
   getBtcRewardStatusConfig,
   isBtcRewardSnapshotStatus,
 } from '../utils';
-
-const baseNetworkId = getNetworkIdsMap().base;
 
 type IUnifiedRecord =
   | {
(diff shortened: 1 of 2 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/52fba251-ea1c-4652-ae52-b83eacadb386?refs=cb98bc618c9aaa8ffe6c9a9b6df6c5c330a8dd4c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/52fba251-ea1c-4652-ae52-b83eacadb386/license?refs=cb98bc618c9aaa8ffe6c9a9b6df6c5c330a8dd4c&source=prChecksComment) |

limichange (approved):

Minnzen (approved):

PR #11684, opened 2026-05-18
fix: render '--' for unavailable token balance/price/value (OK-46491)

weatherstar:
[OK-46491](https://onekeyhq.atlassian.net/browse/OK-46491)

----------

## Summary
- Render `--` (instead of `0` / `$0.00`) when the backend returns null/undefined for a token's `balanceParsed`, `fiatValue`, or `price`, so an upstream indexer/RPC outage no longer looks identical to a real zero balance.
- Propagate the same null/unavailable signal up to the wallet home total: in single-network mode the total renders `--` if any token is unavailable; in All Networks the unavailable network is silently dropped from the partial sum (per product call — no badge or hint).
- Add shared helpers `isValidNumberValue` / `displayOrUnavailable` and a `tokenMapHasUnavailable` signal in `tokenUtils`, with unit tests covering both helpers and the null-skip behavior of `calculateAccountTokensValue`.

## Intent & Context
Users reported (OK-46491) that during an indexer/RPC outage the wallet rendered missing values as `0` / `$0.00`, which is indistinguishable from a real zero balance and made several users believe their funds were lost. Goal: distinguish "unavailable upstream" from "actually zero" everywhere a user reads a balance, price, fiat value, or aggregated total.

## Root Cause
Two layers were collapsing `null`/`undefined` into `0`:
1. The token list views (`TokenBalanceView`, `TokenPriceView`, `TokenValueView`, `TokenPriceChangeView`) passed raw fields straight to `NumberSizeableText`, which falls back to `0` when given a non-numeric value.
2. `TokenDetailsHeader` actively rewrote `data.fiatValue = '0'` at the cache layer, papering over the same class of bug.
At the aggregate layer, `calculateAccountTokensValue` and the worth atom were typed as `string` only — there was no channel to carry "this network total is untrustworthy" up to the home overview.

## Design Decisions
- **Render path**: rely on `NumberSizeableText`'s existing `--` short-circuit rather than touching the formatter — keeps the change localized to view-layer guards via `displayOrUnavailable`.
- **Aggregate channel**: widened `accountWorthAtom.worth` to `Record<string, string | null>` where `null` means "at least one unavailable token in this network". Chose the `accountOverview` Jotai context as the carrier (an earlier plan routed this through `tokenList` context, which was the wrong layer — token list is per-network, total spans networks).
- **All Networks behavior**: confirmed with product — unavailable networks are silently dropped from the partial sum, no badge/hint. Single-network mode does render `--` because there's no partial to fall back to.
- **Backend persistence**: `updateAccountValueForSingleNetwork` / `updateAllNetworkAccountValue` coerce `null` → `'0'` at the boundary, so the persisted sorting/listing surface stays backwards-compatible with old caches and downstream sort code.
- **Cache vs fresh-fetch**: `null` is only injected at fresh-fetch source paths. Existing cached entries stay as pure strings and get overwritten on next fetch — no migration needed.
- **Scope discipline**: this PR covers token list + Token Detail header + home total. Send/Swap/Earn/Perps/Market intentionally keep their zero-fallback semantics and will be migrated separately to keep the blast radius bounded.

## Changes Detail
- `packages/shared/src/utils/tokenValueUtils.ts` (new): `isValidNumberValue`, `displayOrUnavailable` helpers + unit tests.
- `packages/shared/src/utils/tokenUtils.ts`: `tokenMapHasUnavailable` signal used to flag a network's worth as null. `calculateAccountTokensValue` now skips nulls in All Networks / `mergeDeriveAssetsEnabled` sums and returns `null` in the single-network branch.
- `packages/kit/src/components/TokenListView/{TokenBalanceView,TokenPriceView,TokenValueView,TokenPriceChangeView}.tsx`: route raw fields through `displayOrUnavailable` before display.
- `packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsHeader.tsx`: drop the eager `fiatValue = '0'` rewrite.
- `packages/kit-bg/src/states/jotai/contexts/accountOverview/atoms.ts`: widen `worth` to `Record<string, string | null>`.
- `packages/kit-bg/src/states/jotai/contexts/accountOverview/actions.ts`: coerce null → '0' at backend persistence boundary.
- `packages/kit-bg/src/states/jotai/contexts/tokenList/actions.ts`: route sort-partition `findIndex` and `lowValueTokens.reduce` through `isValidNumberValue` so `BigNumber(null)` no longer produces `NaN`.
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx`: inject null at every fresh-fetch source path (single-network, all-networks per-item, all-networks batch loop).
- `packages/kit/src/views/Home/pages/HomeOverviewContainer.tsx`: surface null as `UNAVAILABLE_DISPLAY` before feeding `NumberSizeableText`.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (shared kit + kit-bg)
- **Risk Areas**:
  - Type widening of `accountWorthAtom.worth` (`string` → `string | null`) — all consumers were audited and either guarded or coerced at the persistence boundary, but worth double-checking in review.
  - All Networks partial-sum behavior is a product-defined silent drop; if product changes their mind a badge/hint hook would need to be added.
  - Cache compatibility relies on the persistence-boundary coercion staying in place — losing it would write `null` into storage and break downstream sort.

## Test plan
- [ ] Token list renders `--` when an individual token's balance/price/fiat value is null; renders `0` / `$0.00` for a real zero balance.
- [ ] 24h change column renders `--` when the price-change field is null.
- [ ] Token Detail header renders `--` for unavailable fiat value (no longer silently `$0.00`).
- [ ] Single-network mode: if any token is unavailable, the home total renders `--`.
- [ ] All Networks mode: unavailable network is dropped from the partial sum silently (no badge), sum reflects only available networks.
- [ ] Persisted account value (sorting/listing surface) still works after the change — no `NaN`/null leaks through `BigNumber`.
- [ ] Reload after seeing unavailable values: a successful refetch repopulates the real number.
- [ ] Existing unit tests pass; new tests for `isValidNumberValue`, `displayOrUnavailable`, `tokenMapHasUnavailable`, and `calculateAccountTokensValue` null-skip pass.

[OK-46491]: https://onekeyhq.atlassian.net/browse/OK-46491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3ba1e58e-f509-4c76-95e8-8776261e92e9?refs=094f67b1dba015cc9702f56cd9493f3e7b7c7d76&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3ba1e58e-f509-4c76-95e8-8776261e92e9/license?refs=094f67b1dba015cc9702f56cd9493f3e7b7c7d76&source=prChecksComment) |

packages/shared/src/utils/tokenUtils.ts
```diff
@@ -317,8 +318,12 @@ export function mergeDeriveTokenListMap({
           .plus(value.totalBalanceParsed ?? 0)
           .toFixed();
 
-        mergedToken.fiatValue = new BigNumber(mergedToken.fiatValue)
-          .plus(value.fiatValue)
+        // Drop unavailable participants so the merged value is a partial sum
+        // instead of propagating NaN through toFixed().
+        mergedToken.fiatValue = new BigNumber(
```
originalix: 当 `mergeDeriveAssets` 打开，并且同一个 token 的多个 derive 分支都拿不到有效 `fiatValue` 时，这里会把已合并值和当前值都按 `0` 参与累加，最终写回字符串 `'0'`。

这样后续 `TokenValueView` 会把聚合资产当成真实 0 美元显示并参与零值/小额排序，而不是像单个不可用 token 一样显示 `--`。这会把“价格不可用”误导成“资产无价值”。

建议在合并时记录是否至少有一个有效 `fiatValue`：只有存在有效参与项时才写 partial sum；如果全部参与项都无效，就保留不可用状态，让展示层继续渲染 `--`。

weatherstar: Fixed in 25c3802985. Now tracks whether any participant has a valid `fiatValue` and only writes the partial sum when at least one is valid; otherwise the merged token keeps its unavailable marker so `TokenValueView` continues rendering `--` instead of `$0`. Added a regression test (`mergeDeriveTokenListMap — fiatValue unavailable handling`).

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsHeader.tsx
```diff
@@ -516,15 +515,15 @@ function TokenDetailsHeader(props: IProps) {
                   lineHeight={48}
                   fontWeight={500}
                 >
-                  {tokenDetails?.fiatValue ?? '0'}
+                  {displayOrUnavailable(tokenDetails?.fiatValue)}
```
originalix, line 521: 当 `fetchTokensDetails` 返回了 token detail 但 `fiatValue` 为 `null`/`undefined` 时，这里会让 Header 正确显示 `--`；但同一份原始 `data` 仍会写入 `TokenDetailsContext`，而 `TokenDetailsTabToolbar` 还在用 `isNil(tokenDetail?.fiatValue)` 判断“网络未启用/未创建地址”。

结果是多网络切换菜单会把“价格不可用”的 token 显示成刷新图标和错误 tooltip，而不是 `--`。

建议同步修改 `TokenDetailsTabToolbar`：只有 `tokenDetail` 缺失时显示刷新图标；`tokenDetail` 存在但 `fiatValue` 不可用时，用 `displayOrUnavailable` 展示 `--`。

weatherstar: Fixed. `TokenDetailsTabToolbar` now treats the two states separately: only when `tokenDetail` itself is missing (network not enabled / address not created) does it render the refresh icon + tooltip; when `tokenDetail` is present but `fiatValue` is unavailable, it falls through to `displayOrUnavailable(tokenDetail.fiatValue)` inside the existing `NumberSizeableTextWrapper`, so the multi-network menu shows `--` consistently with the Header.

packages/shared/src/utils/tokenValueUtils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+// Fields typed as `string`/`number` on ITokenFiat may arrive null/undefined
+// when an upstream provider fails. NaN/'NaN' must also be rejected so
+// downstream BigNumber math does not propagate NaN through aggregate sums.
+export function isValidNumberValue(
+  v: string | number | null | undefined,
+): v is string | number {
+  if (v === null || v === undefined || v === '') return false;
+  if (typeof v === 'number') return !Number.isNaN(v);
+  return v !== 'NaN';
```
originalix: 当上游返回非空但不是可解析数字的字符串时，这里会把它当成有效值，因为当前只排除了字面量 `NaN`。

后续 `sumFiatValuesFromTokens` / `sumTokenGroupsFiatValueIgnoringUnavailable` 会把该字符串交给 `BigNumber.plus()`，结果仍会变成 `NaN` 并继续写入 account worth 或展示 formatter，回到本 PR 想避免的“一个坏值污染总额”问题。

建议这里直接用 `BigNumber` 做字符串校验，只有可解析且 finite 的值才返回 true，并补一个非 `NaN` 字面量的非法字符串用例。

weatherstar: Fixed in feacc8473b. `isValidNumberValue` now validates strings via `BigNumber.isFinite()` and numbers via `Number.isFinite()`, so any non-parseable string (`"abc"`, `"1.2.3"`, `"$10"`, `"Infinity"`, etc.) is rejected and never reaches `BigNumber.plus()`. Added regression tests for unparseable strings and `Infinity`, plus a downstream `sumFiatValuesIgnoringUnavailable` case confirming the partial sum stays clean.

originalix (approved):

PR #11702, opened 2026-05-18
fix(discovery): restore browser tabs from storage when mobile tab list opens empty (OK-54651)

limichange:
## Summary
- Restore browser tabs from persistent storage when the mobile tab list modal opens with an empty in-context tab list.
- Guard `initialScrollIndex` / `pinInitialScrollIndex` against empty data so an invalid (`-1`) scroll index is never passed to `ListView`.

## Intent & Context
OK-54651: on mobile the browser tab count was displayed incorrectly and tabs opened via a bookmark could not be shown in the tab list. The root issue is that `MobileTabListModal` rendered against the browser jotai context, which in some entry paths was not yet populated — so the modal showed `0`/incorrect tab count and an empty list even though tabs existed in persistent storage.

## Root Cause
- `MobileTabListModal` reads `tabs` from the discovery jotai context. When the modal is reached through certain paths (e.g. opening a tab via bookmark before the browser context finished initializing), the context `tabs` array is empty while `simpleDb.browserTabs` still holds the persisted tabs. The header count (`explore_tabs_count`) and the list therefore did not match reality.
- `initialScrollIndex` was computed unconditionally: on an empty `data` array it returned `data.length - 1` (`-1`), and `pinInitialScrollIndex` explicitly returned `-1`. Passing a negative `initialScrollIndex` to `ListView` is invalid and contributed to the broken list rendering.

## Design Decisions
- Added a one-shot recovery effect guarded by `hasCheckedEmptyTabsRef` so the storage read runs at most once per mount and only when the context is actually empty (`tabs.length === 0`). When persisted tabs are found it calls `buildWebTabs({ data, options: { isInitFromStorage: true } })` to rehydrate the context, then `setBrowserDataReady()`. This reuses the existing tab-build path rather than duplicating restore logic.
- `setBrowserDataReady()` is always called after the check (even when storage is empty) so downstream consumers are not left waiting on a "data ready" signal.
- For the scroll indices, returning `undefined` (the `ListView` default) instead of a computed/negative index is the minimal, safe fix for the empty-data case.

## Changes Detail
- `packages/kit/src/views/Discovery/pages/MobileTabListModal/index.tsx`
  - Added `useEffect` + `hasCheckedEmptyTabsRef` to restore tabs from `simpleDb.browserTabs.getRawData()` when the context tab list is empty.
  - Pulled `buildWebTabs` and `setBrowserDataReady` from `useBrowserTabActions()`.
  - `initialScrollIndex` returns `undefined` when `data` is empty.
  - `pinInitialScrollIndex` returns `undefined` when `pinnedData` is empty, and returns `undefined` instead of `-1` for the no-scroll case.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile
- **Risk Areas**: The recovery effect runs only when the context tab list is empty and is guarded to run once, so it does not affect the normal flow where tabs are already loaded.

## Test plan
- [ ] Open the browser, create several tabs, then open the mobile tab list — count in the header matches the actual tab count.
- [ ] Open a tab via a bookmark and then open the tab list modal — tabs render and the count is correct.
- [ ] Reach the tab list before the browser context is initialized — tabs are restored from storage instead of showing an empty list.
- [ ] With no persisted tabs, the modal opens cleanly with an empty list and no scroll errors.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0ae45cdc-4e1a-4830-b42e-bf5e07128aed?refs=b5aed6f407db2ee9056bb6f885f89e9aa6618aaa&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0ae45cdc-4e1a-4830-b42e-bf5e07128aed/license?refs=b5aed6f407db2ee9056bb6f885f89e9aa6618aaa&source=prChecksComment) |

PR #11705, opened 2026-05-18
fix(desktop): mirror header no-drag regions to body so clicks land (OK-50357)

huhuanming:
## Background

Empty whitespace in the desktop title bar was sometimes un-draggable while interactive controls (search bar, gift / notification / OneKey ID buttons) were sometimes un-clickable, with the symptom flipping unpredictably between sessions and after route transitions on macOS and Windows.

CDP-driven debugging traced this to a known cluster of Electron / Chromium issues — blink's drag-region collector misses `app-region: no-drag` rects buried inside heavily nested trees, and Electron's downstream `supports_app_region` plumbing only re-runs unreliably. Relevant prior art:

- [electron/electron#41695](https://github.com/electron/electron/issues/41695) — `no-drag` ignored when nested under `drag`
- [electron/electron#27149](https://github.com/electron/electron/issues/27149) — `no-drag` ignored intermittently, hover lost
- [electron/electron#20926](https://github.com/electron/electron/issues/20926) — drag region stops working after BrowserView changes
- [electron/electron#43371](https://github.com/electron/electron/issues/43371) — `app-region: drag` has no effect
- [Chromium embedder-dev: Breaking changes to draggable app-regions](https://groups.google.com/a/chromium.org/g/embedder-dev/c/dNNWXw3Ab08)

Live CDP experiments confirmed: every interactive element has the correct computed `app-region: no-drag`, but Chromium's drag mask cache silently drops them. Adding a fresh fixed-position element with `app-region: no-drag` immediately at the click coordinates makes the click work — i.e. the rect needs to be a body-level node that Chromium does pick up reliably.

### 1. Body-level ghost mirror in `DesktopDragZoneBox`

`DesktopDragZoneBox` now scans its subtree for any descendant that should remain clickable (`.app-region-no-drag`, native form controls, `[role=\"button\"]`, Tamagui `is_GroupFrame`, etc.) and renders a body-level fixed-position no-drag \"ghost\" `<div>` at each one's bounding rect. Ghosts are `pointer-events: none` so real clicks fall through to the underlying control; their only job is to be a stable body-level rect Chromium will reliably include in the OS drag mask.

The mirror is driven by three observers so it stays accurate across React re-renders, layout changes, scrolls and tab transitions:
- `MutationObserver(root, { childList, subtree })` — header content additions / removals
- `ResizeObserver(root)` — header layout changes
- `MutationObserver(document.documentElement, { attributes, attributeFilter: ['aria-hidden'] })` — react-navigation marks inactive tab containers with `aria-hidden=\"true\"`; this is what tells stale background headers to clear their ghosts when the user switches tabs
- `window` `resize` and capturing `scroll` listeners

A pre-flight `isRootShown()` check walks up the ancestor chain for `display:none`, `visibility:hidden` and `aria-hidden=\"true\"` so DesktopDragZoneBox instances in stale background tab stacks never contribute ghosts to the active layer.

All Observer callbacks coalesce through `requestAnimationFrame` to avoid thrashing on bursty mutations.

### 2. React-tree tidy-up so the static state is also correct

These changes are not strictly required once the mirror is in place, but they make the tree read correctly in isolation and remove other footguns:

- `HeaderView` no longer toggles the drag region via `useFocusEffect` / `useIsOverlayPage`. The toggle left a one-frame gap where the title bar was momentarily non-draggable during route or modal transitions. The `disabled={isModelScreen}` prop still disables drag for genuine modal screens.
- `DesktopDragZoneBoxMac` swaps `className` / `style` in place instead of swapping React keys, so Chromium incrementally updates the non-client mask without unmounting the subtree mid-drag.
- `headerRightView` now wraps its content in `<XStack className=\"app-region-no-drag\">` the same way `headerLeftView` already does, so `BaseDesktopTabPageHeader`'s direct injections (Gift / HeaderNotificationIconButton / etc.) inherit no-drag even when they bypass `HeaderRight.tsx`.
- `HeaderIconButton` and `HeaderNotificationButton` carry the `app-region-no-drag` className on the click target itself, so individual controls remain clickable even if a parent wrapper is missed.
- A global CSS fallback marks interactive descendants (button/input/textarea/a/[role=button]) of `.app-region-drag` as no-drag, removing the need for every control in the header to remember the class.

### 3. Dev convenience

`dev:main` gains `--remote-debugging-port=9222 --remote-debugging-address=127.0.0.1` (localhost only, dev script only) so future drag-region investigations can attach a CDP client without rebuilding.

## Test plan

Manual on macOS (primary affected platform):

- [ ] Wallet tab: search bar focusable, gift / notification clickable, empty whitespace between them drags the window
- [ ] Tab through Wallet → Market → Trade → Perp → DeFi → Browser → Wallet; each tab's header is fully clickable AND draggable in whitespace
- [ ] Switch back to Wallet from any tab — old buttons from that tab do NOT leave behind no-drag holes in Wallet's drag area
- [ ] Open a Modal (e.g. settings, send) and close it — header drag still works immediately
- [ ] Window resize — interactive rects stay clickable, no drift
- [ ] DevTools open — drag still works (known to be flakier upstream but should not regress)

Manual on Windows:

- [ ] Same scenarios as macOS; `titleBarOverlay` system buttons (min/max/close) still functional
- [ ] All tab transitions preserve clickability and draggability

Automated:

- [x] `yarn lint:staged` clean
- [x] `yarn tsc:staged` clean

## Notes for reviewers

- The mirror creates O(N) body-level `<div>` ghosts where N is the number of no-drag descendants on the active page (~6 on Wallet, ~10 on Perp in practice). These are `pointer-events: none` and contribute zero work to layout / paint beyond their existence; they are not visible to the user.
- The `isRootShown` ancestor walk happens once per coalesced sync and is bounded by the body. Negligible cost.
- The selector covers Tamagui's `is_GroupFrame` explicitly because Tamagui's Input wraps its `<input>` in a frame that intercepts pointer events; without it the SearchBar would only be no-drag at the inner input rect (~266×28) instead of the full group (~320×32), which we observed leave the icon area un-clickable.
- `apps/desktop/app/app.ts` is intentionally untouched in this PR. The `acceptFirstMouse: true` switch (improves first-click-after-defocus on macOS) ships separately.

### DIFF
--- a/apps/desktop/package.json
+++ b/apps/desktop/package.json
@@ -13,7 +13,7 @@
     "install-app-deps": "electron-builder install-app-deps",
     "dev": "npx concurrently \"yarn build:main:dev\" \"yarn dev:renderer\" \"cross-env LAUNCH_ELECTRON=true node scripts/dev.js\"",
     "dev:rspack": "npx concurrently \"yarn build:main:dev\" \"yarn dev:renderer:rspack\" \"cross-env LAUNCH_ELECTRON=true node scripts/dev.js\"",
-    "dev:main": "electron --inspect=5858 app/dist/app.js",
+    "dev:main": "electron --inspect=5858 --remote-debugging-port=9222 --remote-debugging-address=127.0.0.1 app/dist/app.js",
     "dev:renderer": "TRANSFORM_REGENERATOR_DISABLED=true BROWSER=none WEB_PORT=3001 webpack serve",
     "dev:renderer:rspack": "TRANSFORM_REGENERATOR_DISABLED=true BROWSER=none WEB_PORT=3001 rspack serve",
     "build:main:dev": "rimraf ./app/dist && cross-env node scripts/build.js",
--- a/packages/components/src/layouts/DesktopDragZoneBox/index.desktop.tsx
+++ b/packages/components/src/layouts/DesktopDragZoneBox/index.desktop.tsx
@@ -1,3 +1,5 @@
+import { useEffect, useRef } from 'react';
+
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
 
 import { Stack } from '../../primitives';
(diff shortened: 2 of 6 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/6564cdb5-0a60-4da5-bb2c-961f45b1cd27?refs=2b0a699b86d92f635a27bcdae1617158abdb2fcd&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/6564cdb5-0a60-4da5-bb2c-961f45b1cd27/license?refs=2b0a699b86d92f635a27bcdae1617158abdb2fcd&source=prChecksComment) |

originalix (approved):

originalix (approved):

PR #11710, opened 2026-05-19
fix(home): stabilize history list & overview spinner on network switch(OK-54950)

weatherstar:
## Summary
- Fix a race in the wallet Home **History** tab where switching networks (e.g. BTC → ETH) could clobber the new network's history with a stale empty result and leave the header spinner stuck.
- Track refresh state per Tabs sub-tab in `HomeOverviewContainer` so a TOKENS true-emit cannot block the HISTORY false-emit after a network switch.
- Re-measure virtualized tab inner-container height and re-sync tab visibility on remount, so the new tab paints with the correct height instead of showing blank.
- Light refactor of the history identity tracking so the init-effect short-circuit and the dep-change counter share one source of truth.

## Intent & Context
Reported on Jira **OK-54950**: on the wallet Home History tab, switching networks (most reliably BTC → ETH) could leave the list blank with a stuck "refreshing" spinner in the overview header. Root cause was a stale in-flight fetch from the previous network landing after the identity changed, writing an empty result into the new network's state — and a separate per-tab refresh-state bug that prevented the header spinner from being cleared.

## Root Cause
1. `TxHistoryContainer.run()` was guarded by an account/network dispatch key, but the guard didn't cover:
   - The `usePromiseResult` debounce window (~1s) — a fetch started under the old identity could resolve after the identity flipped.
   - Early-return paths (null account/network) that did not bump the dispatch key, so a superseded in-flight fetch could still write state.
   - Setters outside the `isCurrentRequest` guard (e.g. `setHasMoreOnChainHistory`, `updateAddressesInfo`, `updateAllNetworksState`, the local-cache write after an `await`) — these leaked partial state from the stale identity.
2. `HomeOverviewContainer` tracked a single `isRefreshing` flag across tabs, so a TOKENS `isRefreshing:true` emit could mask the HISTORY `isRefreshing:false` emit on network switch and leave the spinner stuck.
3. `TokenListBlock` fired a bogus `isRefreshing:true` immediately after a successful local-cache load, which the overview interpreted as "still loading".
4. The `composite/Tabs` Container/List did not re-measure the virtualized inner container on remount, so after a sub-tab unmount/remount the new tab could paint with zero height (visible blank area).

## Design Decisions
- **Monotonic request id over dispatch key**: replaced the `account+network` dispatch key with a `fetchRequestIdRef` bumped at `run()` entry *before* any early return AND on identity dep change. This closes the debounce-window and early-return races at the same time, and is cheap to compare.
- **Single try/finally with `isCurrentRequest`-scoped writes**: every shared-state write now lives inside the guard; the `finally` always releases `isManualRefresh` and the header-spinner emit, but the spinner emit is itself restricted to the freshest run so a superseded fetch cannot prematurely dismiss it.
- **Invalidate `lastInitIdentityRef` mid-flight**: when identity goes null between awaits, mark the init owner as null so the awaited local-cache read can't write into a later identity.
- **Per-tab refresh state in `HomeOverviewContainer`**: store `isRefreshing` keyed by sub-tab so HISTORY and TOKENS can't fight over the header spinner.
- **Identity-key memo (refactor commit)**: extracted a single `identityKey` so the init-effect short-circuit and the dep-change counter share one source of truth instead of two hand-aligned dep arrays. Reduces drift risk on future edits.

Alternatives considered:
- Cancelling the in-flight fetch via `AbortController` — rejected: the underlying service call doesn't propagate abort to the backend layer cleanly, so the cheaper "ignore by request id" approach was used instead.
- Resetting `HomeOverviewContainer` refresh state to a single flag on tab-switch — rejected: it would mask, not fix, the underlying cross-tab interference and would regress the legitimate case where both tabs are refreshing concurrently on a fresh mount.

## Changes Detail
- `packages/kit/src/views/Home/pages/TxHistoryContainer.tsx` — request-id-based stale-fetch guard, single try/finally, `lastInitIdentityRef` invalidation, `identityKey` memo refactor, `setHistoryData([])` same-ref short-circuit.
- `packages/kit/src/views/Home/pages/HomeOverviewContainer.tsx` — per-tab `isRefreshing` tracking so HISTORY/TOKENS no longer clobber the header spinner state.
- `packages/kit/src/components/TokenListBlock/TokenListBlock.tsx` — remove the spurious `isRefreshing:true` emit fired right after a local-cache load.
- `packages/components/src/composite/Tabs/{Container,List,context,utils}.tsx` — re-measure virtualized inner-container height on remount; extract `parseCssSize` into `utils`; add deps array to `useImperativeHandle`; replace an unnecessary type assertion with `querySelector<HTMLElement>`.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (Home tab + any caller of `composite/Tabs`)
- **Risk Areas**:
  - `TxHistoryContainer` was substantially restructured (+~200 LOC after the request-id rewrite). Regression surface is "history list correctness on network/account switch" and "header spinner lifecycle".
  - `composite/Tabs` Container/List changes affect every consumer of the shared Tabs component, not just Home. Re-measurement on remount could behave differently on platforms with non-default scroll containers.
  - `HomeOverviewContainer` per-tab refresh state changes the contract slightly for tabs that emit `isRefreshing` — verified TOKENS + HISTORY paths; other tabs use the same hook so should be unaffected.

## Test plan
- [ ] Wallet home → switch BTC → ETH repeatedly; verify the History list renders the new network's data (not blank) and the header spinner clears within ~1s.
- [ ] Wallet home → switch ETH → BTC → ETH rapidly within the debounce window; verify no stale empty list and no stuck spinner.
- [ ] Account switch (HW + software) with the History tab focused; verify list refreshes correctly.
- [ ] Switch to History on a fresh mount (cold start, wallet home) — local cache should render immediately, no flicker of `isRefreshing:true` from the token block.
- [ ] Repeatedly tab Tokens ↔ History; the header spinner should reflect only the active tab's loading state.
- [ ] Any other screen using `composite/Tabs` (Discovery / Earn / Market sub-tabs) — verify tab heights are correct after mount/remount, no blank panels.
- [ ] iOS + Android + Web + Extension + Desktop smoke pass on the Home tab.

### DIFF
--- a/packages/kit/src/views/Home/pages/TxHistoryContainer.tsx
+++ b/packages/kit/src/views/Home/pages/TxHistoryContainer.tsx
@@ -170,6 +170,10 @@ function TxHistoryListContainer(
   );
 
   const isManualRefresh = useRef(false);
+  // Identifies the most recently dispatched fetch. Older fetches that complete
+  // after a newer one started must NOT write their stale result into local
+  // state — they would clobber the new account/network's data.
+  const fetchDispatchKeyRef = useRef('');
   const { run } = usePromiseResult(
     async () => {
       if (!network) return;
@@ -180,12 +184,12 @@ function TxHistoryListContainer(
         accountId = indexedAccount?.id ?? '';
       } else if (!account) return;
 
-      appEventBus.emit(EAppEventBusNames.TabListStateUpdate, {
-        isRefreshing: true,
-        type: EHomeTab.HISTORY,
-        accountId,
-        networkId: network.id,
-      });
+      const refreshNetworkId = network.id;
+      const refreshAccountId = accountId;
+      const dispatchKey = `${refreshAccountId}-${refreshNetworkId}`;
+      fetchDispatchKeyRef.current = dispatchKey;
+      const isCurrentDispatch = () =>
+        fetchDispatchKeyRef.current === dispatchKey;
 
       let r: {
         allAccounts: IAllNetworkAccountInfo[];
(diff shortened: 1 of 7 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f0377d32-a69c-4fd5-97d6-0dc682d51ba4?refs=ef4e030acd554d8d6e1e37f9fe13f82d462b1b41&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f0377d32-a69c-4fd5-97d6-0dc682d51ba4/license?refs=ef4e030acd554d8d6e1e37f9fe13f82d462b1b41&source=prChecksComment) |

packages/kit/src/views/Home/pages/TxHistoryContainer.tsx
```diff
@@ -387,11 +472,24 @@ function TxHistoryListContainer(
       refreshAllNetworksHistory.current = false;
     };
     if ((account?.id || mergeDeriveAddressData) && network?.id && wallet?.id) {
-      void initHistoryState(
-        account?.id ?? '',
-        network.id,
-        indexedAccount?.id ?? '',
-      );
+      if (lastInitIdentityRef.current === identityKey) {
```
originalix: 当用户留在同一个账号/网络下，只切换 history filter、currency 或 `limit` 时，这个 effect 会因为 `lastInitIdentityRef.current === identityKey` 直接返回。

问题是这些值仍然是 `initHistoryState` 的输入依赖，但 `identityKey` 不包含它们，所以本地历史缓存不会按新的过滤/币种重新读取；列表会继续展示旧过滤或旧币种结果，直到后续远端刷新碰巧覆盖，离线或未聚焦场景下尤其明显。

建议把“旧身份写入保护”和“是否重新加载本地缓存”拆开：同一 `identityKey` 下仍允许 `initHistoryState` 因 filter/currency/limit 变化重新执行，只在 await 之后用 `capturedIdentity` 阻止旧账号/网络的数据写回。

weatherstar: Fixed in ef4e030. 拆开了「旧身份写入保护」和「本地缓存重读」两件事：

- 去掉了 `lastInitIdentityRef.current === identityKey` 的同身份提前 return，让 effect 在 filter / currency / limit 变化时也能重跑 `initHistoryState`，把新参数下的本地缓存读出来。
- 保留 await 之后的 `capturedIdentity` 校验，仍然阻止旧账号 / 网络的数据回写到新身份。
- 额外加了 `initRequestIdRef` 单调请求号——同身份下连续触发（比如快速切 filter）时，旧请求即使后完成也不会覆盖新结果。

无本地缓存的分支注释也顺手更新了，覆盖到「filter 切到没缓存」的场景。

originalix (approved):

PR #11735, opened 2026-05-20
fix(account-selector): eliminate avatar/balance/3-dot flicker on open

huhuanming:
## Before / After

<table>
<tr><th>Before</th><th>After</th></tr>
<tr>
<td>

https://github.com/user-attachments/assets/621f0c62-8540-4e35-aed1-59fd0fe3f810

</td>
<td>

https://github.com/user-attachments/assets/405ba531-f32c-4968-9f42-a52fd38c5fb0

</td>
</tr>
</table>

## Summary

Opening the account selector showed several visible jumps in quick
succession — avatar, balance amount, and 3-dot button all appeared to
shift, then balance briefly regressed from a loaded value back to `--`
before settling. Reproduced and root-caused via CDP on the running
Electron build (port 9222). Four independent issues stacked into one
flash; each is fixed at its source.

### 1. Modal enter scale animation (avatar + 3-dot perceived jump)

`createWebModalNavigator` runs every modal through
`scale(0.95) → 1` with an overshoot bezier
`cubic-bezier(0.175, 0.885, 0.32, 1.275)`. Inside the account selector
this made avatars and right-edge buttons appear to fly outward during
the overshoot.

Added a **web-only** opt-out `disableEnterScaleAnimation` to
`IModalNavigationOptions` / `IModalFlowNavigatorConfig`. A parallel
`MODAL_DISABLE_SCALE_LIST` records the choice per modal slot; the
animation listener skips the scale transform for slots that opted out,
keeping only the opacity fade. `AccountSelectorStack` opts in; every
other modal is untouched.

### 2. Balance width change shifts address ~29px right

When `--` (~11px) became `$10.82` (~40px) the address right next to it
moved ~29px. Fixed by wrapping `renderAccountValue` in
`Stack minWidth="$14"` (Tamagui size token = 56px). Stable for typical
fiat balances; rare longer values still `flexShrink`.

### 3. Avatar 200ms fade-in waterfall

`AccountAvatar` ran a `withTiming(opacity 0→1, 200ms)` on first
appearance per address. In the account selector with 8+ simultaneous
mounts these staggered fades read as flicker.

Made the fade **opt-in** via `enableEnterAnimation` (default `false`).
Opted back in at the original sites where the fade was wanted:
`AddressBookListContent`, `SendDataInput/QuickSelectListShared`,
`SendDataInput/RecipientQuickSelect`. The account selector and every
other consumer now renders instantly. Non-opt-in mounts no longer write
to `shownAvatarSourcesCache`, so an opt-in screen later in the session
still gets its first-appearance fade.

### 4. Balance regression `loaded → "--" → loaded`

`useAccountSelectorValuesLoader` did
`next[num] = {}` before each batch fetch, wiping all already-loaded
balances every time `accountsForValuesQuery` changed reference. UI saw
`valuesMap[item.id] === undefined` and re-rendered the `--` placeholder,
then batches re-filled it.

Changed to **prune only entries whose accountId is no longer in the new
query**. Same-wallet effect re-runs preserve every loaded value; switching
wallets still drops stale entries. Same fix applied to
`accountSelectorDeFiMapAtom`.

### Bonus: web blockie cache

`packages/kit/src/components/AccountAvatar/makeBlockieImageUriList/index.ts`
had no main-thread cache, so every fresh `useBlockieImageUri` mount
round-tripped the worker (`postMessage` → worker compute → `setUri` →
re-render) and produced one frame with an empty ``.

Added a `Map<id, dataUri>` keyed cache, seeded `useState`'s initial
value from it, deduped concurrent worker requests for the same id, and
populated the cache from `worker.onmessage`. Repeat opens, re-scrolling,
and other panels using the same address now render the blockie on the
**first frame**.

## Test plan
- [x] CDP frame-by-frame on Electron: row-relative positions stable
  through open (x/y/w/h range 0.00px)
- [x] CDP MutationObserver/rAF over 5s: **0 balance regressions**, **0
  avatar img regressions**
- [x] CDP cache test: cache-cold open shows blockie after worker
  compute (~500ms); cache-warm reopen shows blockie on first frame for
  **28/28** rows
- [x] `yarn tsc:staged` clean
- [x] `yarn lint:staged` clean (0 warnings 0 errors)
- [ ] Manual verification on desktop in `release/v6.3.0`: open / switch
  wallets / reopen — no visible jumps, no `--` regression, avatars
  show immediately on reopen
- [ ] Verify Send recipient quick select and Address Book still show
  the gentle avatar fade-in (intentionally preserved there)
- [ ] **Avatar consumer spot-check** (default went from fade-in to
  instant; verify no jarring "pop" at these entry points):
  - `AccountSelectorTriggerDApp` / `AccountSelectorTriggerBase`
  - `ListItem` (account-flavored avatar)
  - `UniversalSearchAddressItem`
  - `DepositWithdrawModal`
  - Discovery `HeaderRightToolBar`
  - `PortfolioOverview`
  - `ManageAccountActivity`
  - `SelectAddress` (BtcReward)
  - `PagePrimeTransferPreview`

## Review feedback addressed (`9077eca8c3`)

- `mainThreadCache` now has `MAIN_THREAD_CACHE_LIMIT = 500` with the
  same `clear()` strategy as the sibling `shownAvatarSourcesCache` —
  symmetric, bounded across long sessions.
- Balance `<Stack minWidth="$14">` now sets `flexShrink={1}` explicitly
  so the comment's "rare longer values still flexShrink" claim actually
  holds (Tamagui / RN Stack defaults to `flexShrink: 0`).
- Dropped the redundant `descriptor as IModalDescriptor | undefined`
  cast in `createWebModalNavigator` and removed the now-unused
  `IModalDescriptor` import — the descriptor is already typed via
  `useNavigationBuilder<…, IModalNavigationOptions, …>`.

Three lower-priority notes from the review are intentionally left
as-is (sibling atom prune non-atomicity, `MODAL_*_LIST` stale-slot
cleanup) since they match pre-existing code style and have no
observable user impact.

## Round-2 review feedback addressed (`0f42306703`)

- **`disableEnterScaleAnimation` now opts out the whole inner stack**,
  not just the active screen. Previously, navigating from
  `AccountSelectorStack` to a child screen
  (e.g. `ExportPrivateKeysPage`) re-ran the ref callback with the
  child's descriptor and flipped the slot's `noScale` back to `false`;
  a subsequent modal-on-modal push would then shrink the underlying
  screen with `scale(0.95)`. Fixed by checking
  `state.routes.some((r) => descriptors[r.key]?.options?.disableEnterScaleAnimation)`
  so any opt-out screen pins the whole navigator.
- **JSDoc clarified on both option declarations** to make explicit that
  the flag governs enter **and** exit scale, and that opting out on any
  inner-stack screen propagates to the whole navigator instance. (The
  name was kept; renaming to `disableScaleAnimation` would churn four
  files for no behavior gain.)
- **`console.error` replaced with `defaultLogger.app.error.log(...)`**
  in the `useBlockieImageUri` catch — file now complies with the
  project's no-console logging convention.
- **Worker empty-data branch now logs** via
  `defaultLogger.app.error.log(...)` so a regression into a silent
  retry loop (every remount re-posting because we never cache `""`)
  surfaces in error telemetry instead of being swallowed.
- **Cache-hit fast-path is annotated** so the intentional early-return
  (no async work means no cleanup needed) doesn't read like a missing
  cleanup next to the `cancelled`-flag branch below.

### Follow-ups deferred (out of scope, noted for future)
- Two-atom prune (`accountSelectorValuesMapAtom` +
  `accountSelectorDeFiMapAtom`) is two separate `await get → set`
  cycles in `useAccountSelectorValuesLoader`. There's a brief window
  where values is pruned but DeFi isn't. Final state is self-consistent
  (cleanup increments `loadingId`, next loader overwrites both), so no
  user-observable bug — left as-is.
- `MODAL_ANIMATED_VIEW_REF_LIST` / `MODAL_DISABLE_SCALE_LIST` follow
  the existing module-level-permanent pattern; stale slot entries get
  overwritten on next modal open. No leak, no visual artifact.

## Files
- `packages/components/src/layouts/Navigation/Modal/createWebModalNavigator.tsx`
- `packages/components/src/layouts/Navigation/Modal/types.tsx`
- `packages/components/src/layouts/Navigation/Navigator/ModalFlowNavigator.tsx`
- `packages/kit/src/components/AccountAvatar/AccountAvatar.tsx`
- `packages/kit/src/components/AccountAvatar/makeBlockieImageUriList/index.ts`
- `packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx`
- `packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/useAccountSelectorValuesLoader.ts`
- `packages/kit/src/views/AccountManagerStacks/router/index.tsx`
- `packages/kit/src/views/AddressBook/components/AddressBookListContent.tsx`
- `packages/kit/src/views/Send/pages/SendDataInput/QuickSelectListShared.tsx`
- `packages/kit/src/views/Send/pages/SendDataInput/RecipientQuickSelect.tsx`

### DIFF
--- a/packages/components/src/layouts/Navigation/Modal/createWebModalNavigator.tsx
+++ b/packages/components/src/layouts/Navigation/Modal/createWebModalNavigator.tsx
@@ -30,6 +30,7 @@ import {
 import { Stack, XStack, YStack } from '../../../primitives/Stack';
 
 import type {
+  IModalDescriptor,
   IModalNavigationConfig,
   IModalNavigationEventMap,
   IModalNavigationOptions,
@@ -44,6 +45,11 @@ import type {
 import type { GestureResponderEvent } from 'react-native';
 
 const MODAL_ANIMATED_VIEW_REF_LIST: TamaguiElement[] = [];
+// Parallel to MODAL_ANIMATED_VIEW_REF_LIST. When the slot value is true,
+// the corresponding modal opts out of the `scale(0.95) -> scale(1)` enter
+// animation and uses only an opacity fade. Driven by the route option
+// `disableEnterScaleAnimation` (see IModalNavigationOptions).
+const MODAL_DISABLE_SCALE_LIST: boolean[] = [];
 let MODAL_ANIMATED_BACKDROP_VIEW_REF: TamaguiElement | null;
 let ROOT_NAVIGATION_INDEX_LISTENER: (() => void) | undefined;
 
@@ -243,6 +249,7 @@ function WebModalNavigator({
 
         MODAL_ANIMATED_VIEW_REF_LIST.forEach((element, index) => {
           const isHidden = newIndex < index;
+          const noScale = MODAL_DISABLE_SCALE_LIST[index];
           if (media.gtMd) {
             // @ts-expect-error
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
@@ -250,11 +257,20 @@ function WebModalNavigator({
             // @ts-expect-error
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
             element.style.opacity = isHidden ? '0' : '1';
+            let nextTransform: string;
+            if (noScale) {
+              // Skip the bouncy scale; only translateY for stacked modals.
+              nextTransform = isHidden
+                ? ''
+                : `translateY(${-30 * (newIndex - index)}px)`;
+            } else {
+              nextTransform = isHidden
+                ? 'scale(0.95)'
+                : `translateY(${-30 * (newIndex - index)}px) scale(${1 - 0.05 * (newIndex - index)})`;
+            }
             // @ts-expect-error
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-            element.style.transform = isHidden
-              ? 'scale(0.95)'
-              : `translateY(${-30 * (newIndex - index)}px) scale(${1 - 0.05 * (newIndex - index)})`;
+            element.style.transform = nextTransform;
           } else {
             // @ts-expect-error
             // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
@@ -268,13 +284,17 @@ function WebModalNavigator({
 
   const stackChildrenRefList = useRef<TamaguiElement[]>([]);
 
+  const disableEnterScaleAnimation = !!(
+    descriptor as IModalDescriptor | undefined
+  )?.options?.disableEnterScaleAnimation;
+
   useLayoutEffect(() => {
     const element = MODAL_ANIMATED_VIEW_REF_LIST[currentRouteIndex];
     if (element) {
       const el = element as HTMLElement;
       if (media.gtMd) {
         el.style.opacity = '0';
-        el.style.transform = 'scale(0.95)';
+        el.style.transform = disableEnterScaleAnimation ? '' : 'scale(0.95)';
       } else {
         el.style.transform = `translateY(${screenHeight}px)`;
       }
@@ -366,9 +386,11 @@ function WebModalNavigator({
     (ref: TamaguiElement | null) => {
       if (ref) {
         MODAL_ANIMATED_VIEW_REF_LIST[currentRouteIndex] = ref;
+        MODAL_DISABLE_SCALE_LIST[currentRouteIndex] =
+          disableEnterScaleAnimation;
       }
     },
-    [currentRouteIndex],
+    [currentRouteIndex, disableEnterScaleAnimation],
   );
 
   state.routes.forEach((route, routeIndex) => {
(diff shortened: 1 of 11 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/8a1fde94-9a9c-4bf7-8397-1dc04e65bbb6?refs=0f423067030635f6eb61c18814d116384a30e47d&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/8a1fde94-9a9c-4bf7-8397-1dc04e65bbb6/license?refs=0f423067030635f6eb61c18814d116384a30e47d&source=prChecksComment) |

originalix (approved):

PR #11744, opened 2026-05-21
fix: defer Jotai context store cleanup until mirror tracker reaches zero

limichange:
## Summary
- Fix race condition in Jotai context store cleanup where stores were destroyed prematurely while mirror trackers (e.g. account selector, swap providers) were still referencing them.
- Introduce a two-phase teardown: `useJotaiContextRootStore` now *requests* a reset on unmount, and `JotaiContextStoreMirrorTracker` finalizes the removal only when the mirror reference count drops to zero.
- Cancel pending reset requests when the same `storeId` is reused on remount, preventing the new store instance from being silently dropped.

## Intent & Context
During route transitions / remounts in providers such as `AccountSelectorProvider`, `SwapRootProvider`, and `UniversalSearchProvider`, multiple consumers share the same Jotai context store keyed by `storeId`. The previous `useEffect` cleanup unconditionally called `jotaiContextStore.removeStore(...)` on unmount, even when other still-mounted trackers were holding references via `JotaiContextStoreMirrorTracker`. This caused dangling consumers to suddenly read from a fresh store, losing in-memory state (selected account, swap config, etc.) until the page was reopened.

## Root Cause
`useJotaiContextRootStore` removed the store from `storeCache` synchronously on unmount, regardless of `JotaiContextTrackerMap` reference count. When a parent unmounts before its mirrored siblings, `storeCache.delete(id)` ran while `mirrorCount > 0`, so subsequent `getOrCreateStore` calls created a brand-new store with default atom values.

## Design Decisions
- **Reference-counted teardown via mirror tracker**: rather than maintaining a parallel refcount in `useJotaiContextRootStore`, we piggy-back on the existing `JotaiContextTrackerMap` that already counts mirror providers. The root hook only *requests* a reset; the tracker completes it when its count hits zero.
- **Identity-guarded reset (`storeResetRequests` keyed by store instance)**: `requestStoreReset`/`cancelStoreReset`/`completeStoreResetIfRequestedById` all compare against the current `storeCache` instance, so a remount that produces a new store cannot be accidentally finalized by a stale request from the previous instance.
- **Eager completion on unmount when count is already zero**: if the mirror count is already `<= 0` at unmount time (no other mirrors mounted), we immediately call `completeStoreResetIfRequestedById` to preserve the old behavior of synchronous cleanup in the common single-consumer case.
- **Alternative considered**: refcounting inside `useJotaiContextRootStore` itself. Rejected because the mirror tracker already owns the authoritative count and adding a second counter would risk drift.

## Changes Detail
- `packages/kit/src/states/jotai/utils/jotaiContextStore.ts`
  - Add `storeResetRequests: Map<string, IJotaiContextStore>` to track pending teardowns.
  - Add `removeStoreById`, `cancelStoreReset`, `requestStoreReset`, `completeStoreResetIfRequestedById`. Each guards by store identity so stale requests cannot delete a freshly created store.
  - `getOrCreateStore` now reads from `storeCache` directly to avoid going through `getStore` (which currently returns the cached entry but keeps the API stable).
- `packages/kit/src/states/jotai/utils/useJotaiContextRootStore.tsx`
  - On render: cancel any in-flight reset for the resolved store.
  - On unmount: read mirror count via `getJotaiContextTrackerMap()`, call `requestStoreReset`, and complete immediately when `mirrorCount <= 0`.
  - Capture `store` in a ref to ensure the cleanup operates on the exact instance that was mounted.
- `packages/kit/src/states/jotai/utils/JotaiContextStoreMirrorTracker.tsx`
  - After a `remove` action drops the count to zero, call `jotaiContextStore.completeStoreResetIfRequestedById(storeId)` so any pending reset request is finalized at the correct moment.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Mobile / Web / Extension (all platforms share this Jotai infrastructure)
- **Risk Areas**:
  - Provider lifecycle for `AccountSelectorProvider`, `SwapRootProvider`, `UniversalSearchProvider`, and any other consumer using `useJotaiContextRootStore` + `JotaiContextStoreMirrorTracker`.
  - Potential for stores to live slightly longer than before if a mirror provider remains mounted — acceptable, since this is the intended behavior; memory is bounded by the existing tracker map.

## Test plan
- [ ] Open Account Selector across multiple tabs/pages; navigate away and back; confirm selected account / network persists.
- [ ] Open Swap, change tokens, navigate to another page and back; confirm swap state survives.
- [ ] Trigger Universal Search modal, dismiss, reopen; confirm query history / state is preserved as expected.
- [ ] Rapidly mount/unmount a screen using `useJotaiContextRootStore` (e.g. drawer open/close) and verify no console warnings about missing atoms, no broken default values.
- [ ] Run full app on iOS, Android, Desktop and Web — smoke test wallet/account switching flows.

### DIFF
--- a/packages/kit/src/states/jotai/utils/JotaiContextStoreMirrorTracker.tsx
+++ b/packages/kit/src/states/jotai/utils/JotaiContextStoreMirrorTracker.tsx
@@ -32,7 +32,10 @@ import {
 } from '../../../views/Swap/pages/SwapRootProvider';
 import { UniversalSearchProvider } from '../../../views/UniversalSearch/pages/UniversalSearchProvider';
 
-import { buildJotaiContextStoreId } from './jotaiContextStore';
+import {
+  buildJotaiContextStoreId,
+  jotaiContextStore,
+} from './jotaiContextStore';
 
 // AccountSelectorMapTracker
 export function JotaiContextStoreMirrorTracker(data: IJotaiContextStoreData) {
@@ -80,6 +83,10 @@ export function JotaiContextStoreMirrorTracker(data: IJotaiContextStoreData) {
         ...mapCache,
         ...toMergeMap,
       });
+
+      if (action === 'remove' && value.count <= 0) {
+        jotaiContextStore.completeStoreResetIfRequestedById(storeId);
+      }
     };
 
     processMapCount('add');
--- a/packages/kit/src/states/jotai/utils/jotaiContextStore.test.ts
+++ b/packages/kit/src/states/jotai/utils/jotaiContextStore.test.ts
@@ -1,0 +1,259 @@
+/** @jest-environment jsdom */
+
+import type { ReactNode } from 'react';
+import { Component, createElement } from 'react';
+
+import { render } from '@testing-library/react';
+
+import {
+  EJotaiContextStoreNames,
+  getJotaiContextTrackerMap,
+} from '@onekeyhq/kit-bg/src/states/jotai/atoms';
+import type { IJotaiContextStoreData } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
+import { OneKeyLocalError } from '@onekeyhq/shared/src/errors';
+import { EAccountSelectorSceneName } from '@onekeyhq/shared/types';
+
+import {
+  buildJotaiContextStoreId,
+  jotaiContextStore,
+} from './jotaiContextStore';
+import { JotaiContextStoreMirrorTracker } from './JotaiContextStoreMirrorTracker';
+import { useJotaiContextRootStore } from './useJotaiContextRootStore';
+
+jest.mock(
+  '../../../components/AccountSelector/AccountSelectorRootProvider',
+  () => ({
+    AccountSelectorRootProvider: () => null,
+  }),
+);
+jest.mock(
+  '../../../views/Discovery/components/DiscoveryBrowserRootProvider',
+  () => ({
+    DiscoveryBrowserRootProvider: () => null,
+  }),
+);
+jest.mock('../../../views/Earn/EarnProvider', () => ({
+  EarnProvider: () => null,
+}));
+jest.mock(
+  '../../../views/Home/components/HomeTokenListProvider/HomeTokenListRootProvider',
+  () => ({
+    HomeTokenListRootProvider: () => null,
+  }),
+);
+jest.mock(
+  '../../../views/Home/components/HomeTokenListProvider/UrlAccountHomeTokenListProvider',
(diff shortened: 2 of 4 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/09b2ba54-070c-4117-b8fc-7f51b39c37a7?refs=bdc9d2db6dc83afc4f9f0e38b5cc0b851abac3f4&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/09b2ba54-070c-4117-b8fc-7f51b39c37a7/license?refs=bdc9d2db6dc83afc4f9f0e38b5cc0b851abac3f4&source=prChecksComment) |

limichange:
### Jotai context store reset flow

```mermaid
flowchart TD
  A[Root provider commits mount] --> B[getOrCreateStore returns cached or new store]
  B --> C[cancel pending reset for this store instance]

  C --> D[Root provider unmounts]
  D --> E[requestStoreResetById stores storeId plus exact store instance]
  E --> F{Current mirror count for storeId}

  F -- 0 --> G[completeStoreResetIfRequestedById]
  F -- Greater than 0 --> H[Keep cached store alive for active mirrors]

  H --> I[Mirror provider unmounts]
  I --> J[Mirror tracker decrements count]
  J --> K{Mirror count is 0?}
  K -- No --> H
  K -- Yes --> G

  G --> L{Cached store still matches requested instance?}
  L -- Yes --> M[Delete storeCache entry and reset request]
  L -- No --> N[Drop stale reset request only]

  M --> O[Next mount creates a fresh store]
  N --> P[Newer store survives]
```

This is now a generic lifecycle guard for all Jotai context stores, not a `discoveryBrowser`-specific cleanup exception. Root unmount only requests a reset. The store is actually removed when no mirror provider is still tracking the same `storeId`, and only if the cached store is still the exact instance that requested reset.

That gives us two protections:

1. Active mirror providers keep using the existing cached store during startup or transient root unmounts, so their data is not replaced by a fresh empty store.
2. A stale reset request cannot delete a newer store created later with the same `storeId`.

limichange:
https://github.com/OneKeyHQ/app-monorepo/actions/runs/26213583537

weatherstar (approved):

originalix (approved):

PR #11757, opened 2026-05-22
fix: show protocol contract as counterparty for receive-only txs

weatherstar:
## Summary
- Filter EVM zero-address (mint/burn sentinel) out of the source set when computing the history row's counterparty for receive-only transactions.
- Fall back to the interacted-with contract (`to`) instead of the user's own `from` when a receive-only tx has no single resolvable counterparty.
- Mirror the same zero-address filtering on the sends-only branch so burn-style txs also resolve to the contract instead of `0x0`.

## Intent & Context
Borrow-style transactions on protocols like Aave appear in history as receive-only flows with mixed sources: zero-address "mints" for the debt token and aToken, plus a transfer from the protocol's reserve. The history row's counterparty label was rendering as the user's own account name (e.g. "My Wallet") instead of the protocol name, because the previous fallback used the local `from` address when no single source could be resolved.

## Root Cause
In `getTxActionTransferInfo`:
- `uniq(map(receives, 'from'))` produced more than one unique address (the zero address plus the protocol address), so the single-counterparty branch was skipped.
- The fallback then set `transferTarget = from`, but in a receive-only tx `from` is the wallet's own address — i.e. the wrong side of the transfer.
- Result: history showed the user's account as the counterparty for borrow txs.

## Design Decisions
- **Filter zero address before uniquing**, not after — this lets the single-counterparty branch succeed for borrow-style flows where the protocol address is the only "real" source.
- **`to || from` fallback** rather than `to` alone: some decoders may leave `to` empty; preserve the previous behavior as a last resort instead of producing a blank label.
- **Lowercase comparison**: indexers can return zero-address as `0x0000…`, `0X0000…`, or with checksum casing on some chains; normalize before comparing.
- **Apply symmetrically to the sends-only branch**: burn-style txs (target = `0x0`) had the same ambiguity; filtering there too lets them resolve to the interacted contract.
- **EVM-only sentinel is safe on non-EVM**: the literal `0x000…000` is a 20-byte hex form that non-EVM chain addresses cannot collide with, so no platform guard is needed. Documented inline.
- Extracted a small `isZeroAddress` helper to avoid duplicating the literal across both branches.

## Changes Detail
- `packages/kit/src/components/TxAction/TxActionTransfer.tsx`
  - Added local `isZeroAddress` helper with lowercase-safe comparison.
  - Filter zero-address entries from `targets` on both the sends-only and receives-only branches.
  - Change receive-only fallback from `from` to `to || from`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: All (Desktop / Mobile / Web / Extension) — touches shared history rendering logic in `kit`.
- **Risk Areas**:
  - Display-only change in `getTxActionTransferInfo`; no on-chain or signing logic is affected.
  - The new fallback could surface a contract address where it would have previously surfaced the user's own address; this is the intended behavior, but worth spot-checking on non-borrow receive-only flows (airdrops, regular incoming transfers) to make sure the resolved counterparty still looks sensible.

## Test plan
- [ ] Open history for an account that has performed an Aave Borrow — counterparty label shows the Aave Pool contract / protocol name, not the user's own account.
- [ ] Open history for an Aave Repay / Withdraw — counterparty resolves correctly (no regression on the sends side).
- [ ] Open history for a standard incoming transfer (single sender) — counterparty still shows the sender, unchanged.
- [ ] Open history for a burn tx (send to `0x0`) — counterparty falls back to the interacted contract, not `0x0`.
- [ ] Sanity check non-EVM chain history rows (BTC, SOL, etc.) — no regression in counterparty display.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/852e9ff1-865b-4d8a-9643-102604769a1f?refs=1c25230de4d7aa48e902fc0492fdf9acf7e149cf&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/852e9ff1-865b-4d8a-9643-102604769a1f/license?refs=1c25230de4d7aa48e902fc0492fdf9acf7e149cf&source=prChecksComment) |

limichange (approved):

originalix (approved):

PR #11833, opened 2026-05-29
fix: report iOS launch events by fixing native startupTime write

huhuanming:
## Problem

iOS never reported launch events (`appStart` / `jsReadyTime` / `uiVisibleTime`) to analytics, while Android did. `ReactNativeDeviceUtils.getStartupTime()` always returned `0` on iOS, so the durations computed as `<= 0` and were dropped by the `if (... > 0)` gate in `LastActivityTracker`.

## Root cause

`AppDelegate` wrote the startup timestamp into `LaunchOptionsStore.shared` through a KVC bridge:

```swift
cls.value(forKeyPath: "shared")   // -> nil
store?.setValue(..., forKey: "startupTime")  // silently no-ops
```

`LaunchOptionsStore.shared` is a Swift `static let` (stored type property), which is **not exposed to the Objective-C runtime even under `@objcMembers`**. So `value(forKeyPath: "shared")` returned nil, the optional-chained `setValue` silently no-opped, and `startupTime` stayed at its default `0`. `deviceToken` / `launchOptions` writes through the same bridge were also dropped, but those have UserDefaults/JS fallbacks that masked the failure — `startupTime` had none. Android sets the value via a real `@JvmStatic` method, so it was unaffected. Broken since the KVC workaround was introduced.

## Fix

- **Native (`@onekeyfe/react-native-device-utils` 3.0.38):** expose a `sharedInstance()` class method on `LaunchOptionsStore`; `AppDelegate` now resolves the singleton via `perform("sharedInstance")` (the same pattern already used for `BackgroundThreadManager`) instead of the broken KVC path.
- Bump `@onekeyfe` native modules `3.0.37 -> 3.0.38` and drop the temporary react-native-device-utils patch now that the fix is published.
- Remove the redundant iOS `* 1000` in `LaunchOptionsManager.native.ts`; native already returns epoch ms, matching Android. (Without this, the now-real timestamp would be inflated 1000x and the durations would go negative again.)

## Test plan

- iOS device/sim cold start: `getStartupTime()` returns a real epoch-ms timestamp (verified ~1.78e12, was `0`); `jsReadyTime` / `uiVisibleTime` are positive and pass the `> 0` gate.
- Android: unchanged, still works.
- Note: in dev the durations include Metro bundle time and dev analytics is gated off; real magnitudes only meaningful on a release build.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/21e2794d-be0c-42b0-a026-34ca11313ad3?refs=26d901bfb1a003976859b0e8f706a5fb7cc129fb&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/21e2794d-be0c-42b0-a026-34ca11313ad3/license?refs=26d901bfb1a003976859b0e8f706a5fb7cc129fb&source=prChecksComment) |

wabicai (approved):

Minnzen (approved):

originalix (approved):

PR #11897, opened 2026-06-03
feat: merge bridge into swap entry

zhaono1:
## Summary
- Merge the user-visible Bridge entry into Swap while keeping Swap + Limit/Pro as the only header modes.
- Normalize legacy `bridge` / `crosschain` URL params and external Bridge entrypoints to the unified Swap tab.
- Consolidate swap and bridge provider management into one runtime provider manager with legacy config migration.

## Intent & Context
This implements the Figma-driven Swap/Bridge single-entry requirement: users should enter one Swap experience that supports both same-chain and cross-chain pairs. The transaction semantics still need to distinguish cross-chain behavior internally, so historical data, review state, analytics, and backend protocol compatibility remain intact.

Figma MCP access to the referenced nodes was still blocked during implementation, so the first version follows the existing OneKey Swap visual system and focuses on structural/state unification. Pixel-level adjustment can be calibrated later once design access or screenshots are available.

## Design Decisions
- Keep `ESwapTabSwitchType.BRIDGE` as a compatibility and semantic value instead of deleting it, reducing blast radius across review/history/analytics.
- Normalize visible entrypoints to `ESwapTabSwitchType.SWAP`, including old Bridge params from Market, Borrow, Earn/Staking, Swap Pro, and warnings.
- Let Swap support both single-chain and cross-chain token pairs; keep Limit constrained to same-network behavior.
- Use one `swapProviderManager` as the active provider config and read legacy `bridgeProviderManager` only for migration.
- If a provider was disabled in either old swap or old bridge settings, keep it disabled after merge to preserve explicit user choices.

## Changes Detail
- Swap page/header: removes the visible Bridge tab, maps legacy `bridge` / `crosschain` route params to Swap, and deletes the duplicate Bridge mobile container.
- Swap state/token selection: expands Swap network and token-pair logic to include cross-chain pairs while preserving Limit restrictions.
- Provider management: adds `swapProviderManagerUtils` to merge server capabilities, legacy user settings, network disables, and deny-list calculation for same-chain vs cross-chain quotes.
- Entrypoints: updates Market, Borrow, Earn/Staking, Home action, warning, and Swap Pro jump paths to open unified Swap while keeping token prefill behavior.
- Semantics: keeps cross-chain transaction labeling in review/history/analytics based on network mismatch.

## Risk Assessment
- **Risk Level**: High
- **Affected Platforms**: Desktop / Web / Mobile / Extension
- **Risk Areas**: Swap initialization, token selector filtering, provider settings migration, quote provider deny lists, old Bridge entrypoint compatibility.

## Test plan
- [x] `git diff --cached --check`
- [x] `yarn lint:staged`
- [x] Focused Jest: `swapProviderManagerUtils.test.ts`, `SwapPanelContent.test.tsx`, `buildSwapReviewState.test.ts`, `quoteProgress.test.ts`
- [x] Web runtime check: `/swap` and `/swap?tab=bridge` both render unified Swap UI without a Bridge tab
- [x] Web runtime check: settings show one provider management entry
- [ ] `yarn tsc:staged` is currently blocked by unrelated Ledger / `@onekeyfe/hwk-adapter-core` type drift in hardware files outside this diff


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/15a7e0cf-7cbb-4cbf-b7ab-0a4552829018?refs=220147dfdbe06de5b9c2e73a689e5f64b3161376&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/15a7e0cf-7cbb-4cbf-b7ab-0a4552829018/license?refs=220147dfdbe06de5b9c2e73a689e5f64b3161376&source=prChecksComment) |

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
```diff
@@ -308,19 +320,32 @@ export function useSwapInit(params?: ISwapInitParams) {
...
       toToken: toTokenRef.current,
     });
-    if (nextSwapType === swapTypeSwitchRef.current) {
+    const nextVisibleSwapType =
+      getVisibleSwapTabSwitchType(nextSwapType) ?? nextSwapType;
+    const currentVisibleSwapType =
+      getVisibleSwapTabSwitchType(swapTypeSwitchRef.current) ??
+      swapTypeSwitchRef.current;
+    if (nextVisibleSwapType === currentVisibleSwapType) {
       return;
```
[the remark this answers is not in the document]
zhaono1: Fixed in 764381673a. The cold-start tab sync and switch helper now share getVisibleSwapTabSwitchUpdate(), which still writes SWAP when the cached raw tab is BRIDGE even if the visible tab already matches. Added unit coverage for cached BRIDGE -> visible SWAP.

packages/kit/src/views/Swap/utils/swapTypeUtils.ts
```diff
@@ -0,0 +1,81 @@
...
+  toNetworkId?: string;
+}) {
+  if (protocol === EProtocolOfExchange.LIMIT) {
+    return ESwapTabSwitchType.LIMIT;
+  }
+  if (fromNetworkId && toNetworkId && fromNetworkId !== toNetworkId) {
+    return ESwapTabSwitchType.BRIDGE;
+  }
+  return ESwapTabSwitchType.SWAP;
+}
```
sidmorizon, line 71: **BRIDGE enum value still used internally as execution type despite UI normalization**

Type: `Flag`

The `getSwapExecutionTypeFromQuoteResult` and `getSwapExecutionType` functions return `ESwapTabSwitchType.BRIDGE` for cross-chain operations. This value flows into `swapApprovingTransaction.swapType`, history items, logging, and `buildSwapReviewState`. The UI tab atom is normalized via `getVisibleSwapTabSwitchType(BRIDGE) → SWAP`, but the execution type remains BRIDGE in data structures. If any code path sets the `swapTypeSwitchAtom` directly from an execution type (e.g., in the approving success handler in `useSwapQuote.ts`), the atom could end up as BRIDGE, causing no tab to appear selected in the UI. The `swapNetworksIncludeAllNetworkAtom` filter handles BRIDGE correctly (falls through to the SWAP default path), so functionality wouldn't break — but the tab highlight would be wrong.

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
```diff
@@ -654,230 +682,69 @@ export function useSwapInit(params?: ISwapInitParams) {
...
+            bridgeProviderManagers: bridgeProviderManagerSimpleDb,
+          });
+          if (
+            !canUseUnifiedSwapProviderManagers({
+              serverProviders: swapProviderManagerFromServer,
+              unifiedProviderManagers: unifiedProviderManager,
+              bridgeProviderManagers: bridgeProviderManagerSimpleDb,
+            })
+          ) {
+            return;
```
sidmorizon, line 714: **Provider manager migration: canUseUnifiedSwapProviderManagers silently discards valid server data**

Type: `Flag`

In `useSwapGlobal.ts:707-714`, when `canUseUnifiedSwapProviderManagers` returns false, the function returns early without updating anything — the stale cache from `setProviderManagersFromCache()` remains. This is intentional safety behavior to avoid corrupting provider settings with incomplete data. However, this also means stale `serviceDisable` flags (server-side provider toggles) and stale `supportNetworks` lists remain until the next successful sync. In the old code, these fields were always updated from the server even when the overall structure was similar. Worth monitoring whether this causes stale provider disable states in practice.

packages/kit/src/views/Swap/utils/swapTypeUtils.ts
zhaono1: Fixed in 43a7d458ff. The approving-success path already goes through `swapTypeSwitchAction()`, which normalizes `BRIDGE` before writing `swapTypeSwitchAtom`. I did find the same class of direct-write risk in `SwapRootProvider` global-home snapshot hydration, where `defaultTokens.swapType` could still be written directly, and normalized that path as well.

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
zhaono1: Keeping this as-is. `canUseUnifiedSwapProviderManagers()` returns false only when the server/unified provider payload is not safe to migrate, such as missing support networks for enabled capabilities or legacy bridge providers not represented in the unified cross-chain set. In that case we keep the previous cached provider settings instead of mixing partial server fields into a legacy/unified hybrid state. A later successful sync will refresh `serviceDisable` and `supportNetworks` together with a complete unified payload.

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
```diff
@@ -1235,7 +1125,11 @@ export function useSwapInit(params?: ISwapInitParams) {
     params?.importToToken,
     params?.importNetworkId,
     params?.swapTabSwitchType,
+    normalizedSwapTabSwitchType,
+    supportCheckSwapTabSwitchType,
     skipSyncDefaultSelectedToken,
+    fromTokenAmount.isInput,
+    fromTokenAmount.value,
```
sidmorizon, line 1132: **`syncDefaultSelectedToken` gains `fromTokenAmount` as dependency**

Type: `Flag`

The dependency array at `packages/kit/src/views/Swap/hooks/useSwapGlobal.ts:1128-1132` now includes `fromTokenAmount.isInput` and `fromTokenAmount.value`. This means the `syncDefaultSelectedToken` callback identity changes on every keystroke in the amount input. While early-return guards (`shouldSkipSwapDefaultSelectedTokenSync`, `initialSelectedTokensSynced`) prevent actual re-processing after initialization, this could still cause the useEffect that calls it to re-fire on every keystroke, performing the guard checks unnecessarily. The change is needed so the `params?.fromAmount` comparison at lines 809-816 is reactive, but it only matters during initialization when `params?.fromAmount` is set. Consider using a ref for the comparison instead to avoid rebuilding the callback on every amount change.

packages/kit/src/views/Swap/hooks/useSwapQuote.ts
```diff
@@ -78,6 +81,10 @@ export function useSwapQuote() {
   const [swapFromToken, setSwapSelectFromToken] = useSwapSelectFromTokenAtom();
   const { slippageItem } = useSwapSlippagePercentageModeInfo();
   const [swapToToken, setSwapSelectToToken] = useSwapSelectToTokenAtom();
+  const [currentSelectNetwork] = useSwapSelectTokenNetworkAtom();
+  const shouldPauseQuoteForTokenSelector = Boolean(
+    currentSelectNetwork?.networkId,
```
sidmorizon, line 86: **Quote pausing during token selector open is a new safeguard**

Type: `Flag`

In `packages/kit/src/views/Swap/hooks/useSwapQuote.ts:84-86`, a new `shouldPauseQuoteForTokenSelector` boolean is derived from `currentSelectNetwork?.networkId`. This pauses all quoting while the token selector modal is open, preventing unnecessary quote fetches while the user browses tokens. This is added as a guard to 9 separate `useEffect` blocks. The approach is correct but relies on `swapSelectTokenNetworkAtom` being reliably cleared when the modal closes. If the atom gets stuck with a stale network (e.g., due to a navigation edge case), quoting would remain paused until the user opens and closes the selector again.

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
zhaono1: Not changing this one. I traced the caller: the effect that invokes `syncDefaultSelectedToken()` does not include `syncDefaultSelectedToken` in its dependency list and intentionally disables exhaustive-deps. Its dependencies are account/network/import params/runtime support, not `fromTokenAmount`. So typing in the amount input can recreate the callback during render, but it does not make the default-token sync effect re-run on every keystroke. The amount read is only used for the initial `params.fromAmount` sync, and after initialization the existing guards return before default-token processing.

packages/kit/src/views/Swap/hooks/useSwapQuote.ts
zhaono1: Keeping this as-is. The pause flag is bounded by `swapSelectTokenNetworkAtom`, and the only writer in this flow is `SwapTokenSelectModal`: it sets the network while the selector is open and has an unmount cleanup that calls `setCurrentSelectNetwork(undefined)`. Normal selector close/navigation therefore clears the atom before quote effects resume. This is a hypothetical lifecycle edge case, not a current failing path in this PR; if a modal-close path is later found to keep the selector mounted, the fix should be at that modal lifecycle boundary.

originalix (approved):

PR #11913, opened 2026-06-04
fix: universal search account matching, network resolution & tab behavior (OK-42394 OK-39530 OK-46431 OK-54735)

weatherstar:
[OK-42394](https://onekeyhq.atlassian.net/browse/OK-42394) [OK-39530](https://onekeyhq.atlassian.net/browse/OK-39530) [OK-46431](https://onekeyhq.atlassian.net/browse/OK-46431) [OK-54735](https://onekeyhq.atlassian.net/browse/OK-54735)

----------

## Summary
- Match an account whose address belongs to a derive type that has no created address yet (OK-42394).
- Resolve watch-only / imported / external accounts to their own network instead of the default EVM chain when navigating from search results (OK-39530).
- Hide universal-search tabs that have no results for the current query (OK-46431).
- Reset the active tab on every new search so results re-anchor to the "All" tab (OK-54735).

## Intent & Context
QA filed four universal-search issues:
- **OK-42394** — Create a BIP44 address, switch to the LedgerLive derive path without creating its address, then search the BIP44 address: it failed to match the account until the user switched back to BIP44 or created the LedgerLive address.
- **OK-39530** — Searching an imported Polygon watch-only address showed the Polygon balance, but tapping the result navigated to the ETH network.
- **OK-46431** — Tabs with no results for the current query should be hidden; only tabs with results should show.
- **OK-54735** — After re-searching while a single-module tab (e.g. Perps) is active, the result list contained all modules but the highlight did not re-anchor to the "All" tab.

## Root Cause
- **OK-42394:** The `Address` index is populated lazily — on account create, active-account reload, and all-network refresh (the last only writes the current `globalDeriveType` for EVM). A derive type that was never active is absent from the index, so `getAccountNameFromAddress`'s indexed lookup found nothing even though the db account exists.
- **OK-39530:** For watch-only / imported / external accounts the validated `networkId` resolves to the default chain of the impl (eth for EVM), not the account's own `createAtNetwork`, so the value lookup and navigation targeted the wrong chain.
- **OK-46431 / OK-54735:** The tab bar always rendered every tab and the active-tab reset ran on `done` — but the `TabBar` is unmounted while `loading` and seeds its highlight from `focusedTab.value` at mount, so a deferred reset on `done` was unreliable.

## Design Decisions
- **Scan-and-backfill fallback (OK-42394):** When the indexed lookup misses, scan real db accounts by address (covering every derive type) and backfill the index so later lookups hit the fast path. `getAllAccounts()` is backed by an in-memory cache, so the fallback is an in-memory pass, not a per-call DB scan.
- **Single source of truth for the index key:** Extracted `buildAddressRecordId`, shared by the writer (`_saveAccountAddressesBatchByCache`) and the scan fallback, so the reader can never drift from the writer's keying. EVM addresses are lowercased to match how the index keys EVM accounts (no-op in normal operation since EVM db addresses are already lowercased; hardens against a checksum-cased address). Non-EVM SIMPLE chains whose stored `address` differs from their normalizedAddress (e.g. TON friendly form) are not covered by this offline derivation and keep relying on the lazily-built index — documented in-code.
- **Network re-resolution (OK-39530):** Re-resolve the network from the account's own `createAtNetwork` via the existing `serviceAccount.getAccountCreatedNetworkId`, and feed `resolvedNetworkId` into both the value lookup and the DeFi overview, and `resolvedNetwork` into the navigation payload.
- **Derived visible tabs + render-time highlight correction (OK-46431 / OK-54735):** `visibleTabTitles` keeps the leading "All" tab plus only the module tabs that have results; `activeTab` falls back to "All" when the selected tab is hidden. The active tab is reset during `loading` (while the `TabBar` is unmounted, per the known mount-seeding behavior), and `focusedTab` is corrected synchronously at render time to the resolved visible tab before the `TabBar` mounts — avoiding both a stuck highlight and a tab bar with no active tab.

## Changes Detail
- `packages/kit-bg/src/dbs/local/LocalDbBase.ts` — `getAccountNameFromAddress` scan-and-backfill fallback; shared `buildAddressRecordId` used by the writer and the scanner; single-pass backfill + dedup.
- `packages/kit-bg/src/services/ServiceUniversalSearch.ts` — re-resolve watch-only/imported/external accounts to their own network; value lookup and DeFi overview use the resolved network.
- `packages/kit/src/views/UniversalSearch/pages/UniversalSearch.tsx` — `visibleTabTitles` / `activeTab` derived state; hide empty tabs; reset active tab on `loading`; render-time `focusedTab` correction.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**:
  - `getAccountNameFromAddress` is shared by BulkSend, Swap, and SignatureConfirm — the new scan fallback runs for them too (in-memory pass, bounded by the account cache).
  - The `buildAddressRecordId` extraction touches the hot account-save writer path (behavior-preserving refactor; EVM lowercasing is a no-op in normal operation).
  - The `TabBar` `focusedTab` mount-seeding interaction in `UniversalSearch.tsx`.

## Test plan
- [ ] OK-42394: Create a BIP44 address; switch to LedgerLive without creating its address; search the BIP44 address → matches the account. Repeat for BTC derive types.
- [ ] OK-39530: Import a Polygon watch-only account; search it; tap the result → navigates to Polygon (not ETH); balance and DeFi overview match.
- [ ] OK-46431: Search a query that matches only some modules → tabs with no results are hidden; only tabs with results (plus "All") show.
- [ ] OK-54735: Search, switch to a single-module tab, clear and re-search → highlight re-anchors to "All".
- [ ] Edge: enter search from a market/dapp preset whose module returns no results → the tab bar highlights "All" (no blank highlight) while the list shows the "All" results.
- [ ] Regression: searching an already-indexed address still resolves via the fast path; BulkSend/Swap address-name resolution is unaffected.

[OK-42394]: https://onekeyhq.atlassian.net/browse/OK-42394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-39530]: https://onekeyhq.atlassian.net/browse/OK-39530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-46431]: https://onekeyhq.atlassian.net/browse/OK-46431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-54735]: https://onekeyhq.atlassian.net/browse/OK-54735?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e26609b1-935a-46c8-83b9-e78c6734dc02?refs=e960baba99149682ca2292a6370d6a18d9965c69&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e26609b1-935a-46c8-83b9-e78c6734dc02/license?refs=e960baba99149682ca2292a6370d6a18d9965c69&source=prChecksComment) |

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -4783,72 +4783,195 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
+      // The Address index is populated lazily — on account create, active
+      // account reload (AccountSelectorEffects), and all-network refresh, the
+      // last of which only writes the current globalDeriveType for EVM. A
+      // derive type that was never active (e.g. searching a BIP44 address
+      // while LedgerLive is selected without a created address) can be absent
+      // from the index, so the lookup above finds nothing even though the
+      // account exists. Fall back to scanning real db accounts by address and
+      // backfill the index so later lookups hit the fast path.
+      if (isEmpty(items)) {
+        items = await this.scanAccountEntriesByAddress({
```
sidmorizon, line 4798: 🟡 P2 空索引未命中时，每次搜索"未持有地址"都会全量扫描账户

**问题**

`getAccountNameFromAddress` 在 `info` 索引未命中时回退到 `scanAccountEntriesByAddress`，后者遍历 `getAllAccounts()` 的全部账户并对每个账户调用 `buildAddressRecordId` 做匹配。当搜索一个**当前钱包并不持有**的外部地址时（universal search 里最常见的场景），`matched` 为空、不会回填索引；同时 `ServiceAccount.getAccountNameFromAddress` 在 `ServiceAccount.ts:2591` 对空结果执行删除：

```ts
if (isEmpty(result)) {
  void this.getAccountNameFromAddressMemo.delete({ networkId, address });
}
```

空结果因此**不会进入 30s memo**，导致同一个"未持有地址"的每次搜索都会重新触发一次 `O(n_accounts)` 全量扫描。

**影响**

`getAllAccounts()` 命中 `dbAllRecordsCache`，没有额外 DB I/O，但扫描本身（逐账户构造 record id + `Set` 查找）在账户/派生地址较多的钱包上会给每次地址搜索增加可感知的主线程开销，且因空结果被主动从 memo 删除而无法摊销。这是本 PR 新引入、且落在 universal search 热路径上的成本，账户规模越大影响越明显。

**建议**

可在 `scanAccountEntriesByAddress` 层面，对"扫描未命中"的 `(networkId, address)` 组合加一个与 memo 同级（~`30s`）的短期负缓存，避免对同一未持有地址重复全量扫描；或仅在确实存在该 impl 下未激活的 derive 账户时才触发扫描。若实测在目标账户规模下扫描耗时可忽略，也可保留现状，但建议在代码注释中明确这一权衡。

[Claude session ↗](https://claude.ai/code/session_01YWqfrXBCGSpmXyFxH3iEes)

---

sidmorizon (commented):
第 1 轮 review（head `c1c60d8`）：在独立 worktree 中 checkout PR head 后，重点核查了四个高风险点——`buildAddressRecordId` 重构后与原 writer 的 keying 一致性、`ServiceUniversalSearch` 中被移动的 `accountsDeFiOverview` 赋值、`UniversalSearch.tsx` 渲染期对 `focusedTab.value` 的直接修改，以及 `scanAccountEntriesByAddress` 的回填与去重。

**未发现 P0/P1 问题**，PR 的数据流在这几处均正确，部分改动（others 账户按 `createAtNetwork` 重解析网络后 value 与 DeFi overview 一致）实际上是改进。仅提出 1 条 P2（性能，见 inline），且已说明缓解条件，可酌情处理。

CI 13 项全部通过。后续如有新增 commit 会做增量 review。

---

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -4783,72 +4783,195 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
+    }
+
+    const { accounts } = await this.getAllAccounts();
+    const matched: IDBAccount[] = [];
+    for (const account of accounts) {
+      const recordId = this.buildAddressRecordId({ account, networkId });
+      if (recordId && queryIds.has(recordId)) {
+        matched.push(account);
+      }
+    }
```
sidmorizon, line 4956: **Full-table scan fallback in scanAccountEntriesByAddress may be slow for large account sets**

Type: `Flag`

The new `scanAccountEntriesByAddress` at `packages/kit-bg/src/dbs/local/LocalDbBase.ts:4938` calls `this.getAllAccounts()` which loads every account from the database. For users with many accounts (e.g. hundreds of HD-derived accounts across multiple wallets), this linear scan could be noticeably slow. The backfill mechanism (`void this.saveAccountAddresses(...)` at line 4955) mitigates this for subsequent lookups, but the first hit for an uncached derive type will always pay this cost. The `getAllAccounts` method does use an in-memory cache (`dbAllRecordsCache`), so repeated scans within a session are cheaper. Worth monitoring performance in production for power users.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
weatherstar: Fixed in 4afb02ec0e — implemented suggestion (a). Added a `scanAccountMissCache` (LRU, 30s TTL) in `LocalDbBaseContainer`, checked at the top of `scanAccountEntriesByAddress` (keyed by `${networkId}--${address}`) and populated when the scan finds no owner. It is flushed by `clearStoreCachedData()`, which already fires on every Account/IndexedAccount/Wallet/Device write — so a newly created account is found immediately, no staleness vs the memo's delete-on-empty. This removes the repeated `O(n)` scan **and** the `getAllAccounts()` deep-clone (note: `getAllRecordsByCache` returns `cloneDeep`, so the per-call cost was higher than a plain loop) for the same not-held address.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
weatherstar: Addressed together with the P2 above (4afb02ec0e): repeated lookups of a not-held address are now short-circuited by `scanAccountMissCache`, and held-but-unindexed accounts are still backfilled so later lookups hit the fast path. The first-hit scan for an uncached derive type remains by design, but it reads from the in-memory `dbAllRecordsCache` (no DB I/O) and only runs when the indexed lookup misses.

packages/kit/src/views/UniversalSearch/pages/UniversalSearch.tsx
```diff
@@ -240,22 +240,62 @@ export function UniversalSearch({
...
+  // before the TabBar mounts in the `done` branch — to the resolved, always
+  // visible `activeTab`. This is a render-time fix (not a deferred write), so it
+  // is applied at mount-seed time rather than being dropped by it.
+  if (
+    searchStatus === ESearchStatus.done &&
+    focusedTab.value !== activeTab &&
+    !visibleTabTitles.includes(focusedTab.value)
+  ) {
+    focusedTab.value = activeTab;
+  }
```
sidmorizon, line 310: **TabBar highlight desyncs from content filter when deferred results restore a previously-hidden initialTab**

Severity: `non-severe`

When `initialTab='dapp'` is set, the loading-phase `useEffect` resets both `focusedTab` and `filterType` to `initialTabName` ("Dapps"). Since Dapp is a deferred search type (not in `PRIMARY_SEARCH_TYPES`), primary results arrive first without Dapp sections. At that point `visibleTabTitles` excludes "Dapps", so `activeTab` falls back to "All", and the render-time fix at lines 291-297 corrects `focusedTab.value` to "All". However, `filterType` remains "Dapps" (never updated by the render-time fix). When deferred results add a Dapp section, `visibleTabTitles` now includes "Dapps", so `activeTab` becomes "Dapps" (because `filterType` is still "Dapps" and is now in `visibleTabTitles`). But the render-time fix does NOT correct `focusedTab.value` back to "Dapps" because the condition `!visibleTabTitles.includes(focusedTab.value)` is false ("All" is always visible). Result: the TabBar highlights "All" while the content list filters to only Dapp results (`isInAllTab=false`). This happens every time `initialTab='dapp'` is used and both primary and deferred searches return results.

weatherstar: Fixed in e960baba99. The render-time fallback now also pins `filterType` to the resolved fallback, not just `focusedTab.value`. Previously `filterType` kept the hidden preset ("Dapps"), so once its deferred results arrived and the tab became visible again, `activeTab` jumped back to "Dapps" and the content list filtered to it, while the already-mounted TabBar stayed seeded on "All" — exactly the desync you described. Pinning `filterType` keeps the user on the always-visible fallback so highlight and content stay in sync.

The added `if (filterType !== activeTab)` guard is a no-op during swipe gestures (there `filterType` already equals `activeTab`; only `focusedTab` is mid-transition), so it does not fight the gesture — that is also why I corrected `filterType` rather than relaxing the `focusedTab` guard.

originalix (approved):

PR #11915, opened 2026-06-04
fix(perps): don't block trading when rebate wallet bind fails

Minnzen:
## Problem
When the rebate wallet binding (`bind-wallet`) fails to persist server-side, the client treated the wallet as "not bound" on **every** `checkPerpsAccountStatus` call — including background refreshes. Each time it would clear the working agent credential to force a re-bind, which:

- cleared a valid agent on background refreshes → `canTrade` dropped → trade button greyed out
- re-prompted the agent-approval signature in a loop ("sign → cleared → sign → cleared …")
- repeatedly hit `/rebate/v1/wallet/batch-check`

Net effect: affected accounts (non-first EVM address whose binding never lands server-side) **could not trade**.

## Fix
Gate the forced agent-credential clear on `isEnableTradingTrigger`:

```ts
// before
if (!isRebateBound) { ... clearLocalAgentCredentials ... }
// after
if (!isRebateBound && isEnableTradingTrigger) { ... }
```

Rebate binding becomes **best-effort**: it still runs when the user explicitly clicks **Enable Trading** (they're already in a signing flow), but a background status refresh never clears an active agent.

## Behavior

| Scenario | Before | After |
|---|---|---|
| Background refresh, not bound | clears agent → blocks trading, re-prompts signatures | agent untouched, trading continues ✅ |
| User clicks Enable Trading, not bound | clears agent → re-bind | unchanged (still binds) ✅ |
| Already bound | unaffected | unaffected |

Trade-off: while the server-side bind bug persists, affected existing accounts won't be force re-bound (short-term attribution gap), but trading is never blocked. They bind naturally on first agent creation / once the server is fixed.

## Test plan
- Reproduced by forcing the rebate check to always return `false`: confirmed the sign→clear loop and greyed-out trade button.
- With the fix: background refreshes no longer clear the agent; trading stays enabled; clicking Enable Trading still triggers binding.

## Scope
1 file, core change is a single condition. No funds/signing-security/DB logic touched.

### DIFF
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
@@ -1962,8 +1962,10 @@ export default class ServiceHyperliquid extends ServiceBase {
           }),
         ]);
 
-        // Clear local credentials to force new agent creation for rebate binding
-        if (!isRebateBound) {
+        // Clear local credentials to force new agent creation for rebate binding.
+        // Gate on isEnableTradingTrigger: rebate binding is best-effort and must
+        // not clear a working agent on background refreshes (would block trading).
+        if (!isRebateBound && isEnableTradingTrigger) {
           await this.clearLocalAgentCredentials({
             userAddress: accountAddress,
           });


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/1ddfc260-ac6a-4202-87fa-568bd6a7c443?refs=eb2d46c212a0995b9a6a97d8a24fa4eb1a141683&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/1ddfc260-ac6a-4202-87fa-568bd6a7c443/license?refs=eb2d46c212a0995b9a6a97d8a24fa4eb1a141683&source=prChecksComment) |

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1962,8 +1962,10 @@ export default class ServiceHyperliquid extends ServiceBase {
           }),
         ]);
 
-        // Clear local credentials to force new agent creation for rebate binding
-        if (!isRebateBound) {
+        // Clear local credentials to force new agent creation for rebate binding.
+        // Gate on isEnableTradingTrigger: rebate binding is best-effort and must
+        // not clear a working agent on background refreshes (would block trading).
+        if (!isRebateBound && isEnableTradingTrigger) {
```
sidmorizon, line 1968: **Rebate binding is permanently deferred if user never re-triggers Enable Trading**

Type: `Flag`

With this change, if `checkInternalRebateBindingStatusWithCache` returns `false` during a background refresh, no corrective action is taken — credentials are preserved and `internalRebateBoundOk` is marked true. The rebate will only be bound when the user explicitly clicks 'Enable Trading' (or an `INVALID_AGENT` error triggers recovery via `withToast.ts:66`). If the user already has a working agent and never encounters an error, the rebate may remain unbound indefinitely. The 1-minute memoize TTL at `ServiceHyperliquid.ts:2530` means the cache refreshes frequently, but the result is ignored on background paths. This is consistent with the 'best-effort' design documented in the comment, but worth flagging as a product-level consideration — if rebate binding has revenue implications, there's no automatic retry mechanism.

originalix (approved):

PR #11922, opened 2026-06-05
feat(chart): offline TradingView chart + warm/instant chart (native + desktop)

huhuanming:
## Summary

Brings the offline TradingView chart to **desktop** (in addition to the existing native iOS/Android pipeline) and makes the chart open/switch fast across platforms.

The chart bundle (`@onekeyhq/tradingview-charting-library`) is fetched at build time and served locally, so the chart loads with no network round-trip for its assets — and never ships inside the JS hot-update bundle.

### Native offline pipeline (iOS/Android)
- Build-time fetch of the chart bundle (`development/scripts/fetch-tradingview-assets.mjs`), staged into the native apps (iOS Run Script / Android Gradle copy). Gitignored, token-gated, online fallback when absent.
- `CHART_WEBVIEW_MODE` code-level switch; unified pooled chart WebView + prewarm; instant symbol switch via `SYMBOL_CHANGE`.

### Desktop offline chart
- Serves the bundle to the chart `<webview>` under a **custom privileged scheme** `onekey-chart://` (registered `standard`/`secure`/`corsEnabled`), mirroring native's "offline virtual same-origin" — so the chart runs on a real secure origin (direct Hyperliquid `fetch()`/WebSocket work; `file://`'s null origin would not).
- Assets are bundled into the **asar** (outside `build/`), so they ship in the installer but **never enter the hot-update bundle**. `app.ts` registers a path-traversal-guarded protocol handler on the `persist:onekey` session and reports availability via desktop globals (online fallback on no-token builds).

### Desktop warm / instant chart
- The chart renders **in-flow** (a normal kit `<WebView>` in the page) with a **constant per-domain unified source**, so switching tokens never changes the URL (no reload) — the active token rides `SYMBOL_CHANGE`. It scrolls natively with the page (no overlay, no positioning jitter); market and perps stay isolated (separate instances + `storageNamespace`).
- **Instant open:** the tapped token's symbol is predicted at click time (`chartPredictedSymbolAtom`) and the detail layouts mount the chart from it (`useTokenDetail.chartSymbol`) instead of waiting on the token-detail API — removing the desktop white screen and keeping the chart mounted across in-place token switches. Native reuses the same prediction (kline precached via `ServiceMarketV2` memoization so prewarm + detail share one request).

## Packaging boundary
| Artifact | asar / installer | hot-update bundle |
|---|---|---|
| Chart bundle (`tradingview-assets/`) | ✅ | ❌ (outside `build/`) |

## Testing
- `tsc` (root + desktop main-process) and `lint:staged`: clean.
- Desktop verified via CDP: single warm webview, unified constant source, in normal document flow (no fixed ancestor / no jitter), survives token navigation without reloading.
- Offline activates only when assets are bundled (`NODE_AUTH_TOKEN` set at build); otherwise the app falls back to the online chart.

## Notes for reviewers
- Swap K-line modal keeps the legacy per-instance WebView (it remounts per token by design; the warm host gives no benefit there).
- Desktop first cross-screen open still has an irreducible Electron `<webview>` guest spawn + first-kline network fetch; in-place token switching is warm.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c64bdd19-ee72-4b6d-bcb8-a54600e95e01?refs=3ac96acb5f90ff4dd62168f7bfaf22ac9abd3107&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c64bdd19-ee72-4b6d-bcb8-a54600e95e01/license?refs=3ac96acb5f90ff4dd62168f7bfaf22ac9abd3107&source=prChecksComment) |

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -278,48 +278,111 @@ class ServiceMarketV2 extends ServiceBase {
...
+      S: 1,
+      m: 60,
+      h: 3600,
+      H: 3600,
+      d: 86_400,
+      D: 86_400,
+      w: 604_800,
+      W: 604_800,
+    };
+    return n * (unitSec[m[2]] ?? 60);
```
sidmorizon: **`_klineIntervalToSeconds` regex fails to parse bare-number TradingView resolutions, defeating cache bucketing**

Severity: `non-severe`

The new `_klineIntervalToSeconds` method uses regex `/^(\d+)\s*([smhHdDwW])$/` which requires a unit suffix letter. However, TradingView sends resolutions as bare numbers for minutes: `'1'`, `'5'`, `'15'`, `'30'`, `'60'`, `'240'` (visible from `normalizeTradingViewKLineInterval` at `packages/kit/src/components/TradingView/TradingViewV2/messageHandlers/klineDataHandler.ts:26-53`). These bare numbers fail the regex and all fall through to the default of 60 seconds. For a 5-minute interval (`'5'`), `sec` should be 300 but is 60; the bucket function `Math.floor(t / 60) * 60` then floors to 1-minute boundaries instead of 5-minute boundaries, causing cache misses every minute within the same 5-min bar. The stated goal — "repeated requests for the same token+interval within the same bar collapse to one network call" — is defeated for the majority of intervals.

Suggested fix:

```ts
  // Bar length in seconds for a TradingView resolution ('1m','5m','1H','1D'...').
  // TradingView also sends bare numbers ('1','5','15','60','240') representing minutes.
  private _klineIntervalToSeconds(interval?: string): number {
    if (!interval) return 60;
    const trimmed = interval.trim();
    const m = /^(\d+)\s*([smhHdDwW])$/.exec(trimmed);
    if (!m) {
      // Bare number → TradingView minute resolution (e.g. '5' = 5 minutes)
      const asNum = parseInt(trimmed, 10);
      return Number.isFinite(asNum) && asNum > 0 ? asNum * 60 : 60;
    }
    const n = parseInt(m[1], 10);
    const unitSec: Record<string, number> = {
      s: 1,
      S: 1,
      m: 60,
      h: 3600,
      H: 3600,
      d: 86_400,
      D: 86_400,
      w: 604_800,
      W: 604_800,
    };
    return n * (unitSec[m[2]] ?? 60);
  }

```

apps/desktop/app/app.ts
```diff
@@ -1184,6 +1241,53 @@ async function createMainWindow() {
...
+            return new Response('Not Found', { status: 404 });
+          }
+        },
+      );
+    } catch (e) {
+      // protocol.handle throws if already registered (e.g. window recreated on
+      // `activate`). Safe to ignore — the existing handler stays valid.
+      logger.info('[chart] protocol handler already registered:', e);
+    }
+  }
```
sidmorizon, line 1289: **Desktop protocol handler registered only on persist:onekey session — verify kit WebView uses that partition**

Type: `Flag`

The `onekey-chart://` scheme handler is registered exclusively on `webviewSession` (`session.fromPartition('persist:onekey')`) at `apps/desktop/app/app.ts:1253`. The desktop `ChartWebView` at `packages/kit/src/components/TradingView/ChartWebView/index.desktop.tsx` renders a standard kit `<WebView>` component. If that kit WebView resolves to an Electron `<webview>` tag using the default session (not `persist:onekey`), requests to `onekey-chart://local/...` would have no handler and fail. The comment at line 1247 states "the chart <webview> (which runs in this partition)", suggesting the kit WebView does use `persist:onekey`, but this should be confirmed — an incorrect assumption here would completely break the desktop offline chart.

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  const source = useMemo(() => {
+    if (CHART_WEBVIEW_MODE === 'online') {
+      return { uri: onlineUrl };
+    }
+    if (IS_UNIFIED) {
+      // uri:'' so the source keys never flip absent (Nitro rejects that).
+      return {
+        uri: '',
+        localBundle: CHART_WEBVIEW_LOCAL_BUNDLE,
```
originalix, line 218: 这里 native offline 分支没有任何 bundle 可用性检查：`CHART_WEBVIEW_MODE` 当前默认是 `offline`，`TradingViewV2` / Perps 在 native 会直接走 `ChartWebView`，但 `fetch-tradingview-assets.mjs` 在没有 `NODE_AUTH_TOKEN` 时会 `exit 0`，Android/iOS copy 缺少 source 时也只是 no-op。这样开源/no-token 构建，或者 release pipeline 没有显式执行 fetch 时，会加载不存在的 `tradingview-assets/index.html`，native 图表白屏，而且不会使用传入的 `onlineUrl` fallback。建议要么把资源拉取接入 native release build 并在资源缺失时失败，要么在 native 端增加 bundle-ready 检测，缺失时返回 `{ uri: onlineUrl }`。

packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
```diff
@@ -289,8 +295,29 @@ export function TradingViewPerpsV2(
...
+  // as ready (the page's listeners are already up) so lines still sync.
+  const useUnifiedHost =
+    (platformEnv.isNative &&
+      CHART_WEBVIEW_MODE !== 'legacy' &&
+      CHART_WEBVIEW_SCENE === 'unified') ||
+    // Desktop warm overlay is always unified (constant onekey-chart:// source).
+    getDesktopOfflineChartReady();
+  const [unifiedReady, setUnifiedReady] = useState(false);
+  useEffect(() => {
+    if (useUnifiedHost) setUnifiedReady(true);
```
originalix, line 312: 这里在 unified host 挂载时就把 `unifiedReady` 置为 true，但这并不等价于 chart 页面里的 perps/line listener 已经注册完成。下面 `isChartLinesReady` 会立刻变 true，`useChartLines` 可能在 WebView 还没发出 `tradingview_perpsReady` 前发送仓位线/订单线/强平线；如果这次 full sync 被丢掉，后面的 `onLoadEnd` 再 `setUnifiedReady(true)` 也不会触发状态变化重试，冷启动时线会一直缺失直到下一次数据变化。建议 unified 模式也等待真实的 `tradingview_perpsReady` / ACK 后再允许 line sync，或在 ready/onLoadEnd 时强制 resend full lines。

development/scripts/fetch-tradingview-assets.mjs
```diff
@@ -0,0 +1,104 @@
...
+ */
+import { execFileSync } from 'node:child_process';
+import { mkdtempSync, writeFileSync, rmSync, mkdirSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join, dirname } from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+// ── Pinned version (single source of truth) ──────────────────────────────────
+const PKG = '@onekeyhq/tradingview-charting-library';
+const VERSION = '0.1.14-test.1';
```
originalix, line 52: 这里把 TradingView bundle 版本直接 hardcode 在脚本里，后续升级只能改脚本常量，缺少一个可审计、可自动 bump、可校验的版本来源。这个资源最终会进入 native/desktop 安装包，建议不要用 `latest`，但也不要把版本散落在脚本逻辑里。

建议把包信息抽成 manifest，例如 `development/tradingview-assets.manifest.json`：

```json
{
  "package": "@onekeyhq/tradingview-charting-library",
  "version": "0.1.14-test.1",
  "registry": "https://npm.pkg.github.com",
  "integrity": "sha512-...",
  "schemaVersion": 1
}
```

脚本从 manifest 读取 `package/version/registry/integrity`，下载 metadata 后校验 `dist.integrity`，再 stage 到 mobile/desktop。这样私有 chart repo 发布新包后，可以由 bot 给 app-monorepo 开一个只 bump manifest 的 PR，并带上 changelog/hash；release workflow 始终读取 committed manifest，保证构建可复现。内部临时测试可以允许 env override 版本，但正式 release 不应该默认走 override 或 latest。

另外建议 fetch 成功后在 staged assets 里写一份 `onekey-chart-bundle.json`，记录包名、版本、integrity，方便线上排查当前安装包内置的是哪个 TradingView bundle。

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useToMarketDetailPage.ts
```diff
@@ -51,6 +52,20 @@ export function useToDetailPage(options?: IUseToDetailPageOptions) {
...
+      // desktop uses it to mount the detail chart immediately (its gate waits on
+      // tokenDetail.symbol otherwise — the white-screen cause). Read back in
+      // useTokenDetail (matched by address) so a stale prediction never leaks.
+      if (platformEnv.isNative || platformEnv.isDesktop) {
+        void chartPredictedSymbolAtom.set({
+          source: 'market',
+          symbol: item.symbol,
+          networkId: item.networkId,
+          address: item.tokenAddress,
+        });
```
originalix, line 72: 这里预测 market chart symbol 时没有把 `item.decimals` 写进 `chartPredictedSymbolAtom`。但 tradingview bundle 的 unified market symbol 已经把 `decimal` 当作 per-symbol pricescale 输入；这里缺失后，prewarm 和详情页切换期间仍会按 `MarketTradingView` 的默认 `decimal=8` 发 `SYMBOL_CHANGE`，微小价格 token 的 scale 还是会错。建议把 `IMarketToken.decimals` 一起存入 predicted symbol，并在 detail chart props 里优先使用 tokenDetail/predicted decimal。

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/navigateToMarketTokenDetail.ts
```diff
@@ -19,6 +25,18 @@ export function navigateToMarketTokenDetail(
...
+  // Seed the predicted symbol BEFORE changeActiveToken clears tokenDetail, so the
+  // chart gate (useTokenDetail.chartSymbol) stays open through the switch — the
+  // chart stays mounted and rides SYMBOL_CHANGE instead of unmounting/reloading.
+  if ((platformEnv.isNative || platformEnv.isDesktop) && token.symbol) {
+    void chartPredictedSymbolAtom.set({
+      source: 'market',
+      symbol: token.symbol,
+      networkId: token.networkId,
+      address: token.address,
+    });
```
originalix, line 43: TokenSelector 这条入口也只预测 `symbol/networkId/address`，没有 decimal；这会和列表入口修复后形成分支不一致。因为这个 predicted symbol 会在 tokenDetail 回来前驱动 warm chart，建议把 token 类型补上 `decimal/decimals` 并一起写入 `chartPredictedSymbolAtom`，否则从 selector 进入详情仍会用默认精度。

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
```diff
@@ -0,0 +1,260 @@
...
+// the shared WebView never reloads on switch (per-symbol data rides SYMBOL_CHANGE).
+function buildUnifiedParamsJson(params: Record<string, string>): string {
+  const constant: Record<string, string> = {};
+  for (const key of CHART_WEBVIEW_UNIFIED_APP_GLOBAL_KEYS) {
+    if (params[key] !== undefined) {
+      constant[key] = params[key];
+    }
+  }
+  constant.scene = 'unified';
+  constant.storageNamespace = 'unified';
```
originalix, line 59: 这里把 native unified source 的 `storageNamespace` 固定成 `unified` 后，market/perps 在同一个 pooled WebView 下会共享 `tradingview_settings_*` 和 `tradingview_study_template_*`。这改变了原来按 market/perps 隔离设置/指标模板的存储模型。若这是为了单池复用的产品取舍，需要在方案里明确；否则建议仍按业务域拆 namespace，至少避免 settings/template 跨 market/perps 串用。

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
```diff
@@ -0,0 +1,260 @@
...
+  }, [autoDriveSymbol, hybridReady, isFocused, params, sendSymbolChange]);
+
+  const source = useMemo(() => {
+    if (CHART_WEBVIEW_MODE === 'online') {
+      return { uri: onlineUrl };
+    }
+    if (IS_UNIFIED) {
+      // uri:'' so the source keys never flip absent (Nitro rejects that).
+      return {
+        uri: '',
```
originalix, line 216: PR 描述说 no-token / assets absent build 会 fallback online，但 native unified/offline 分支这里传的是 `uri: ''` + `localBundle`，`onlineUrl` 只在 `CHART_WEBVIEW_MODE === 'online'` 时使用。除非 native module 内部保证 localBundle 缺失时自动回退，否则 assets 没被 fetch 到的构建会尝试加载空 URI/local bundle 而不是线上 chart。建议在 JS 层加 assets-ready gate，或把明确的 fallback URL 传给 native module 并补注释/测试。

originalix (commented):
补充几个和 charting-library #167 串联后的架构/正确性问题。

packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
```diff
@@ -289,8 +295,29 @@ export function TradingViewPerpsV2(
...
+  }, [useUnifiedHost]);
```
sidmorizon, line 313: **Unified perps readiness is optimistically set to true on mount**

Type: `Flag`

In `TradingViewPerpsV2.tsx:310-313`, `unifiedReady` is set to `true` immediately via a mount-time `useEffect` when `useUnifiedHost` is true. This assumes the shared pooled WebView is already loaded (since it was prewarmed by `ChartPrewarm` or a previous chart mount). If this assumption is wrong (e.g., first app launch with no prior chart use), `isChartLinesReady` would be `true` prematurely and chart lines (liquidation, position, orders) would try to sync before the chart page is actually ready. The `onLoadEnd` callback (`handleUnifiedLoadEnd` at line 565-568) provides a fallback by setting `unifiedReady` again, but the optimistic `true` means lines could be pushed before the page acknowledges readiness. In practice, the chart page silently ignores messages it's not ready for, so no visible issue.

.github/workflows/release-desktop-snap.yml
```diff
@@ -105,6 +105,12 @@ jobs:
         run: |
           yarn install --immutable
 
+      - name: Stage TradingView offline chart assets
+        env:
+          NPM_GITHUB_READ_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TRADINGVIEW_ASSETS_REQUIRED: '1'
+        run: node development/scripts/fetch-tradingview-assets.mjs
```
sidmorizon, line 116: **Standalone desktop workflows missing `permissions: packages: read` causing TradingView asset fetch to fail**

Severity: `severe`

The standalone desktop workflows (`release-desktop-snap.yml`, `release-desktop-win.yml`, `release-desktop-winms.yml`, `release-desktop-mas.yml`, `release-desktop.yml`) add the "Stage TradingView offline chart assets" step with `TRADINGVIEW_ASSETS_REQUIRED: '1'` and `NPM_GITHUB_READ_TOKEN: ${{ secrets.GITHUB_TOKEN }}`, but unlike `release-android.yml`, `release-ios.yml`, and `release-desktop-all.yml`, they do NOT declare `permissions: packages: read` at the workflow level. Without this permission, the default GITHUB_TOKEN lacks the `packages:read` scope needed to fetch from `npm.pkg.github.com`. Since `TRADINGVIEW_ASSETS_REQUIRED=1` converts a failed fetch into a hard `process.exit(1)` (`development/scripts/fetch-tradingview-assets.mjs:70-71`), these workflows will fail when manually dispatched.

huhuanming: Fixed in 52428df83c — added a workflow-level `permissions:` block granting `contents: read` and `packages: read` to all 5 standalone desktop release workflows (release-desktop, -snap, -win, -winms, -mas), matching the existing pattern in release-desktop-all.yml. The default `GITHUB_TOKEN` now has the `read:packages` scope to fetch the offline TradingView bundle from npm.pkg.github.com, so the `TRADINGVIEW_ASSETS_REQUIRED=1` staging step no longer hard-fails on manual dispatch.

packages/kit-bg/src/services/ServiceMarketV2.ts
huhuanming: Good catch — fixed in 52428df83c. `_klineIntervalToSeconds` now parses TradingView's bare-number resolutions (`'1'`, `'5'`, `'15'`, `'60'`, `'240'`) as minutes (×60) when the unit-suffix regex doesn't match, instead of silently defaulting to 60s. This restores correct cache bucketing per interval (e.g. `'5'` → 300s); the existing `'5m'`/`'1H'`/`'1D'` suffix path is unchanged.

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
huhuanming: 已处理(52428df83c)。先确认了 native module 现状:它**没有任何 fallback、也没有 JS 可见的 asset-presence 信号**——iOS/Android 的 `computeTargetUrl` 都会无条件拼出 local virtual URL(`onekey-chart://` / `appassets`),资源缺失时直接返回 Not Found 白屏,不会用到 `onlineUrl`。

因此 JS 层无法可靠地 gate「bundle 是否就绪」。改法:把 `onlineUrl` 作为显式 `fallbackUri` 传进 offline/unified source,并在代码里写明契约——**native 在 local `entry` 资源缺失时必须加载 `fallbackUri`**。`fallbackUri` 是 app-global,所以驱动加载与复用的 `localBundle`/`entry`/`paramsJson` 跨 market/perps 仍然 byte-identical,unified 不变式不受影响。

⚠️ **限制(已在代码注释和跟进项里标注)**:当前 native `computeTargetUrl` 还没读取 `fallbackUri`,所以这是**前向契约 + 文档**,运行时真正生效需要 `@onekeyfe/react-native-chart-webview` 的后续 PR(读取 `fallbackUri` 并加进 spec)。在那之前,asset-less 构建仍需 release pipeline stage 资源,或设 `CHART_WEBVIEW_MODE='online'`。

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
huhuanming: 已处理(52428df83c)。先确认了 native module 现状:它**没有任何 fallback、也没有 JS 可见的 asset-presence 信号**——iOS/Android 的 `computeTargetUrl` 都会无条件拼出 local virtual URL(`onekey-chart://` / `appassets`),资源缺失时直接返回 Not Found 白屏,不会用到 `onlineUrl`。

因此 JS 层无法可靠地 gate「bundle 是否就绪」。改法:把 `onlineUrl` 作为显式 `fallbackUri` 传进 offline/unified source,并在代码里写明契约——**native 在 local `entry` 资源缺失时必须加载 `fallbackUri`**。`fallbackUri` 是 app-global,所以驱动加载与复用的 `localBundle`/`entry`/`paramsJson` 跨 market/perps 仍然 byte-identical,unified 不变式不受影响。

⚠️ **限制(已在代码注释和跟进项里标注)**:当前 native `computeTargetUrl` 还没读取 `fallbackUri`,所以这是**前向契约 + 文档**,运行时真正生效需要 `@onekeyfe/react-native-chart-webview` 的后续 PR(读取 `fallbackUri` 并加进 spec)。在那之前,asset-less 构建仍需 release pipeline stage 资源,或设 `CHART_WEBVIEW_MODE='online'`。

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
huhuanming: 已处理为「有意取舍 + 文档」(52428df83c)。共享 `'unified'` namespace 是单池设计的必要项:它是 boot param、烘进 source,而 byte-identical-source 不变式要求它跨 host 恒定,这样唯一的 pooled WebView 在 market<->perps 切换时**不 reload**——所以 native 上 market/perps 确实共享 settings/study-template(desktop 因为是挂两个独立 URL 的 WebView 才能按域隔离)。

我在 `storageNamespace = 'unified'` 处加了注释说明这个 why 与取舍。真正的按域隔离只能放到 chart repo 里、按 `SYMBOL_CHANGE` 的 source 来 key,因为 app 侧的 payload 并不携带 `storageNamespace`。

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useToMarketDetailPage.ts
huhuanming: 已处理(52428df83c)。给该入口补了 `decimals` 并把 `decimal: item.decimals` 写入 `chartPredictedSymbolAtom`(atom 本来就有可选的 `decimal?: number`,只是从没被写过)。`useTokenDetail` 现在派生 `chartDecimal`(`tokenDetail.decimals` → predicted decimal → undefined,且 address+networkId 命中当前 token 才用 predicted),两个 layout 都把它传给 `MarketTradingView`,所以 prewarm / SYMBOL_CHANGE 用真实 priceScale 而非默认 `decimal=8`。无单位转换——`IMarketToken.decimals`、predicted、`MarketTradingView` 的 `decimal` 端到端都是同一个 price-decimals 整数。

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/navigateToMarketTokenDetail.ts
huhuanming: 已处理(52428df83c)。给这条入口的 token 类型(以及两个 TokenSelector 回调类型)补了 `decimals?: number`,并写入 `decimal: token.decimals`,与列表入口保持一致;共用的 `useTokenDetail.chartDecimal` 消费端优先用 tokenDetail/predicted decimal,所以从 selector 进详情的 warm chart 不再回退到 `decimal=8`。

packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
huhuanming: 已处理(52428df83c)。保留 warm/prewarmed 的乐观 `unifiedReady` 走快路径,但 chart 页面本来就会发真实的 READY ACK(`onChartLinesReady`)和 `onLoadEnd`;现在在这两个真实信号上 bump 一个 `unifiedReadyConfirmation` 计数器,并对每次 bump 强制一次完整 `sendLinesSync`。这样冷启动若首次乐观同步被丢,页面真正 ready 后线会自愈;因为是完整且幂等的 `PERPS_TV_LINES_SYNC`,warm 场景这次 resend 也无害(用 ref 保证每次 confirmation 只触发一次,不会随每次数据变化重发)。

packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
huhuanming: Fixed in 52428df83c. Kept the optimistic `unifiedReady` for the warm/prewarmed fast path, but the chart page already emits a real READY ACK (`onChartLinesReady`) and `onLoadEnd`; I now bump a `unifiedReadyConfirmation` counter on those genuine signals and force a single full `sendLinesSync` per bump. So if the early optimistic sync is dropped on a cold start, the lines self-heal as soon as the page is actually ready — and since it's a full idempotent `PERPS_TV_LINES_SYNC`, the warm-case resend is harmless (a ref guard fires it once per confirmation, not on every data change).

development/scripts/fetch-tradingview-assets.mjs
huhuanming: 这个建议很好,manifest + `dist.integrity` 校验 + staged 里写一份 `onekey-chart-bundle.json` 确实是更可审计、可复现、可被 bot 自动 bump 的做法。本 PR 先**不在这里改动**,避免把范围扩到资源版本管理;我会另开一个跟进项把版本抽到 `development/tradingview-assets.manifest.json`(package/version/registry/integrity + schemaVersion),脚本从 committed manifest 读取并校验 integrity,release workflow 始终走 manifest,内部测试再用 env override。谢谢 🙏

apps/desktop/app/app.ts
huhuanming: Thanks for the flag — I verified the assumption holds, so no change needed. The desktop `ChartWebView` (`index.desktop.tsx`) renders the kit `<WebView>` without passing a `partition`, and `DesktopWebView` defaults `partition` to `'persist:onekey'` (`DesktopWebView.tsx:528`) — the same partition the `onekey-chart://` handler is registered on (`app.ts:1253`). So the chart `<webview>` does run in `persist:onekey` and requests reach the handler. Appreciate the double-check.

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -278,48 +278,117 @@ class ServiceMarketV2 extends ServiceBase {
...
```
sidmorizon: **Monthly kline interval ('M'/'1M') not handled by _klineIntervalToSeconds, causing wrong cache bucketing**

Severity: `non-severe`

The regex `/^(\d+)\s*([smhHdDwW])$/` in `_klineIntervalToSeconds` does not include uppercase `M` in its character class. When TradingView requests a monthly resolution (e.g. `'1M'`), the regex fails to match. The fallback bare-number parser then runs `parseInt('1M', 10)` which returns `1` (JS parseInt stops at first non-digit), yielding `1 * 60 = 60` seconds — treating a 1-month interval as 1 minute for bucketing purposes.

This means `timeTo`/`timeFrom` are floored to 60-second boundaries instead of ~2.6M-second (monthly) boundaries. Two kline requests 61 seconds apart within the same monthly bar will produce different cache keys and miss the memoizee cache, defeating the deduplication this bucketing was designed to provide for the prewarm→detail page transition.

Suggested fix:

```ts
    const m = /^(\d+)\s*([smhHdDwWM])$/.exec(trimmed);
    if (!m) {
      // Bare number → TradingView minute resolution (e.g. '5' = 5 minutes)
      const asNum = parseInt(trimmed, 10);
      return Number.isFinite(asNum) && asNum > 0 ? asNum * 60 : 60;
    }
    const n = parseInt(m[1], 10);
    const unitSec: Record<string, number> = {
      s: 1,
      S: 1,
      m: 60,
      h: 3600,
      H: 3600,
      d: 86_400,
      D: 86_400,
      w: 604_800,
      W: 604_800,
      M: 2_592_000,
    };
    return n * (unitSec[m[2]] ?? 60);
```

.github/workflows/release-android.yml
```diff
@@ -110,6 +114,19 @@ jobs:
         run: |
           yarn install --immutable
 
+      - name: Stage TradingView offline chart assets for EAS upload
+        env:
+          NPM_GITHUB_READ_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TRADINGVIEW_ASSETS_REQUIRED: '1'
```
sidmorizon, line 120: **CI workflows use GITHUB_TOKEN for cross-repo private package fetch with hard-fail mode**

Type: `Flag`

All release workflows (release-android, release-ios, release-desktop-*) pass `${{ secrets.GITHUB_TOKEN }}` as `NPM_GITHUB_READ_TOKEN` with `TRADINGVIEW_ASSETS_REQUIRED: '1'`. The script's own documentation at `development/scripts/fetch-tradingview-assets.mjs:23-27` explicitly states: "Because the package lives in a *different* repo, the default Actions GITHUB_TOKEN cannot read it — CI must pass a PAT with org-level read:packages." Combined with the hard-fail flag, this would cause all release builds to fail at the fetch step — unless the org's GitHub Packages access settings explicitly grant the app-monorepo repo access to the private package. Worth confirming that the package visibility/access settings are configured to allow `GITHUB_TOKEN` from this repo, or switching to a dedicated PAT secret.

packages/kit-bg/src/services/ServiceMarketV2.ts
huhuanming: Good catch — fixed in 31ac17f9a6. Added uppercase `M` (month = `2_592_000`s) to the regex char class and the unit map, so `'1M'` now buckets as a monthly bar instead of falling through the bare-number path to 1 minute. Lowercase `m` stays minute (case-sensitive). `'1M'` is a real resolution here (`sliceRequest` handles it and has a test), so this matters for the prewarm→detail dedup.

.github/workflows/release-android.yml
huhuanming: Thanks for the flag. This is a deliberate, recent team decision rather than an oversight: commit `07146bef` ("ci: read offline chart package via built-in GITHUB_TOKEN") intentionally switched all release workflows to the built-in `GITHUB_TOKEN`. That relies on the org granting the **app-monorepo** repo read access to the private `@onekeyhq/tradingview-charting-library` package in GitHub Packages (the package's repo-access / "inherit access" setting) — when that's configured, the repo-scoped `GITHUB_TOKEN` can read it cross-repo and a dedicated PAT isn't needed. The script's doc note (lines 23-27) predates that switch.

My change here only brings the 5 standalone desktop workflows in line with that decision by adding the `packages: read` scope they were missing. The cross-repo **package-access setting** is the thing to confirm org-side before relying on it for release; if it isn't (or can't be) granted, the fallback is exactly what you suggest — a dedicated `read:packages` PAT secret. Leaving the code as-is and keeping this thread open until the package-access setting is confirmed.

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
```diff
@@ -0,0 +1,307 @@
...
+    // when present. NOTE/LIMITATION: today's native computeTargetUrl (iOS/Android)
+    // does NOT yet honor `fallbackUri` — a non-empty `uri` simply wins outright and
+    // an empty `uri` falls through to the localBundle regardless of whether its
+    // files exist. So this guard only takes effect once the native module is
+    // updated to read `fallbackUri`; until then asset-less builds still need the
+    // release pipeline to either stage assets or flip CHART_WEBVIEW_MODE to
+    // 'online'. `fallbackUri` is app-global (the same remote chart URL), so it does
+    // NOT taint the byte-identical unified source: localBundle/entry/paramsJson —
+    // the keys that drive the shared WebView's load + reuse — stay identical across
+    // market and perps hosts. Only emitted when an online URL actually exists.
```
sidmorizon, line 211: **Native fallbackUri not yet honored by chart-webview module — offline-only builds without assets show white screen**

Type: `Flag`

The native `ChartWebView` at `index.native.tsx:199-211` documents that `fallbackUri` is passed to the native module but 'today's native computeTargetUrl (iOS/Android) does NOT yet honor fallbackUri'. This means open-source / no-token builds that ship without the offline assets will show a blank/white chart area on native (the localBundle resolves to a missing index.html). The CHART_WEBVIEW_MODE constant is hardcoded to `'offline'`, and there is no runtime JS-side guard for asset presence on native (unlike desktop's `tradingViewOfflineReady`). Until the native module is updated to read `fallbackUri`, such builds must either run `fetch-tradingview-assets.mjs` or change the constant to `'online'`/`'legacy'`.

packages/kit/src/components/TradingView/hooks/useTradingViewUrl.ts
```diff
@@ -32,11 +34,22 @@ export function useTradingViewUrl(options: IUseTradingViewUrlOptions = {}) {
...
+
   const baseUrl = useMemo(() => {
+    // Dev "use local TradingView URL" always wins (chart dev server on :5173).
     if (devSettings.enabled && devSettings.settings?.useLocalTradingViewUrl) {
       return localTradingViewUrl;
     }
 
+    // Local offline chart served from the asar (onekey-chart://local/index.html).
+    if (desktopOfflineChartReady) {
+      return DESKTOP_OFFLINE_CHART_ENTRY_URL;
```
sidmorizon, line 50: **Desktop offline chart URL leaks into legacy WebView path when `preferLegacyChart` is true**

Severity: `severe`

When `desktopOfflineChartReady` is true, `useTradingViewUrl` unconditionally sets `baseUrl` to `onekey-chart://local/index.html` (line 49-50). This URL propagates to `finalUrl` for ALL consumers, including `SwapKLineContent` which sets `preferLegacyChart=true` to force the legacy `<WebView>` path (`packages/kit/src/views/Swap/pages/modal/SwapKLineContent.tsx:921`). In `TradingViewV2`, `preferLegacyChart` disables `useChartWebView` (`packages/kit/src/components/TradingView/TradingViewV2/TradingViewV2.tsx:408-410`), so the legacy `<WebView>` renders with `src={tradingViewUrlWithParams}` containing the `onekey-chart://` scheme. However, the `onekey-chart://` protocol handler is only registered on the `persist:onekey` webview session (`apps/desktop/app/app.ts:1253`), not on the default renderer session. If the legacy kit `<WebView>` component does not use that specific partition, the chart in the Swap K-line modal will fail to load (blank/error page) on desktop builds that ship the offline bundle.

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -278,48 +278,119 @@ class ServiceMarketV2 extends ServiceBase {
...
+      networkId: string;
+      interval?: string;
+      timeFrom?: number;
+      timeTo?: number;
+      autoHandleError?: boolean;
+    }): Promise<IMarketTokenKLineResponse> => {
+      let innerInterval = interval?.toUpperCase();
+      if (innerInterval?.includes('M') || innerInterval?.includes('S')) {
+        innerInterval = innerInterval?.toLowerCase();
+      }
```
sidmorizon: **Pre-existing interval normalization conflates month ('1M') and minute ('1m')**

Type: `Flag`

The interval normalization at line 329-332 first calls `toUpperCase()` (turning `'1m'` → `'1M'`), then checks `includes('M')` and lowercases the whole string. This means both `'1m'` (minute) and `'1M'` (month) become `'1m'` in the API call. The new `_klineIntervalToSeconds` function (line 284-307) correctly distinguishes them for bucketing purposes (`m`=60s, `M`=2,592,000s), but the API request is still wrong for month intervals. This is **pre-existing logic** (identical to the code before this PR), so the PR doesn't introduce a regression. However, the new bucketing makes the discrepancy more visible: month intervals get correct month-level cache bucketing but the API returns minute-level data. In practice, TradingView sends bare numbers for minutes (e.g., `'1'`, `'60'`), so `'1m'` likely never reaches this code path — but `'1M'` for month could.

packages/kit/src/views/Market/MarketDetailV2/hooks/useTokenDetail.ts
```diff
@@ -18,12 +19,47 @@ export function useTokenDetail() {
...
+    if (typeof tokenDetail?.decimals === 'number') return tokenDetail.decimals;
+    if (
+      typeof predictedSymbol?.decimal === 'number' &&
+      predictedSymbol.address === tokenAddress &&
+      predictedSymbol.networkId === networkId
+    ) {
+      return predictedSymbol.decimal;
+    }
+    return undefined;
+  }, [tokenDetail?.decimals, predictedSymbol, tokenAddress, networkId]);
```
sidmorizon, line 61: **chartDecimal reads tokenDetail.decimals — verify this is price precision, not contract decimals**

Type: `Flag`

The new `chartDecimal` in `useTokenDetail` (line 51-61) reads `tokenDetail?.decimals` as the primary decimal source for the chart's `priceScale`. Previously, `MarketTradingView` always used the default `decimal=8`. If `tokenDetail.decimals` represents the token's contract decimals (e.g., 18 for ERC-20), this would break the chart's price formatting (attempting to show 18 decimal places). If it represents the API's price-display decimals, this is correct and an improvement. The `IMarketTokenDetail` type definition (not in the diff) determines which interpretation is correct. The fact that `item.decimals` from the market list is propagated through the predicted-symbol atom suggests the PR author treats this as price decimals.

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -278,48 +278,119 @@ class ServiceMarketV2 extends ServiceBase {
...
+    // so flooring keeps slice boundaries contiguous (no gaps); the latest partial
+    // bar is filled by the realtime update path.
+    const sec = this._klineIntervalToSeconds(params.interval);
+    const bucket = (t?: number) =>
+      t !== undefined && sec > 0 ? Math.floor(t / sec) * sec : t;
+    return this._memoizedFetchMarketTokenKline({
+      ...params,
+      timeFrom: bucket(params.timeFrom),
+      timeTo: bucket(params.timeTo),
+    });
```
sidmorizon, line 411: **Kline time-bucketing floors both timeFrom and timeTo, which may truncate the latest partial bar**

Type: `Flag`

In `ServiceMarketV2.ts:387-393`, both `timeFrom` and `timeTo` are floored to the bar interval boundary. This is intentional for cache deduplication (the comment at line 382 explains: "the latest partial bar is filled by the realtime update path"). However, if the realtime update path is not active (e.g. polling mode, or a token without websocket config), the most recent partial bar's data could be missed because `timeTo` was floored to the previous bar boundary. For example, with a 1-day interval and `timeTo = now`, flooring drops up to 24 hours of data from the API response. The chart would show stale data until the next full bar completes. This is mitigated in practice because TradingView's `getBars` typically requests `timeTo` well past `now`, but worth verifying the partial-bar contract with the API.

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
huhuanming: Confirmed and tracked — same root cause as @originalix's threads above. The native module's `computeTargetUrl` (iOS/Android) unconditionally builds the local virtual URL and exposes no JS-visible asset-presence signal, so JS can't reliably gate bundle-readiness. We pass `onlineUrl` as an explicit `fallbackUri` into the offline/unified source and documented the contract: native must load `fallbackUri` when the local entry is missing. Runtime enforcement needs a follow-up in `@onekeyfe/react-native-chart-webview` to read `fallbackUri`; until then, asset-less builds must run `fetch-tradingview-assets.mjs` or set `CHART_WEBVIEW_MODE='online'`. Tracking as a forward-contract.

packages/kit/src/components/TradingView/hooks/useTradingViewUrl.ts
huhuanming: Verified this doesn't reproduce, and added a belt-and-suspenders guard anyway. The legacy `<WebView>` in `TradingViewV2` passes no `partition` prop, so `DesktopWebView` defaults to `partition='persist:onekey'` (`DesktopWebView.tsx:528`) — exactly the session the `onekey-chart://` handler is registered on (`app.ts:1243`). `onShouldStartLoadWithRequest` → `useNavigationHandler` only blocks `www.tradingview.com`, so `onekey-chart://` (host `local`) is allowed. So the Swap K-line modal chart resolves correctly on offline desktop builds.

To remove the reliance on that default, I pinned `partition="persist:onekey"` explicitly on the legacy WebView path in 3ac96acb5f.

packages/kit-bg/src/services/ServiceMarketV2.ts
huhuanming: Good catch — fixed in 02f0ab0eaf. You're right it was pre-existing (no regression introduced here). Replaced the `toUpperCase()` + `includes('M')` → `toLowerCase()` normalization with a case-sensitive per-unit mapping (`_normalizeKlineApiInterval`): minute/second lowercase, hour/day/week/month uppercase, bare numbers pass through. Verified the output is identical for every currently reachable resolution and only `'1M'` changes (`1m` → `1M`).

packages/kit/src/views/Market/MarketDetailV2/hooks/useTokenDetail.ts
huhuanming: It's price-display decimals, not contract decimals — safe. `useTokenDetail.ts:46-50` documents it as "Price decimals to drive the chart priceScale"; `ChartPrewarm.tsx:24-25` notes the decimal only affects `priceScale` (not the kline request), and the predicted-symbol atom carries the same price decimals. It never feeds amount/balance math. This is actually an improvement over the old hardcoded `decimal=8` for micro-price tokens. No change needed.

packages/kit-bg/src/services/ServiceMarketV2.ts
huhuanming: This doesn't drop the latest partial bar. `bucket(timeTo) = floor(timeTo/sec)*sec` lands on the **current** bar's open timestamp (the bar containing `now`), not the previous bar. `getBars` semantics are `open <= timeTo` (inclusive), so the in-progress bar is still returned, and the realtime/polling path refreshes its evolving values. The only theoretical edge would be a backend treating `timeTo` as strictly exclusive — confirmed inclusive, so no data loss. No change needed.

zhaono1 (approved):

Minnzen (approved):

originalix (approved):

PR #11929, opened 2026-06-05
fix: private send and swap follow-ups(OK-55750 OK-55824 OK-55827 OK-55828 OK-55832 OK-55850 OK-55856 OK-55857)

zhaono1:
[OK-55750](https://onekeyhq.atlassian.net/browse/OK-55750) [OK-55824](https://onekeyhq.atlassian.net/browse/OK-55824) [OK-55827](https://onekeyhq.atlassian.net/browse/OK-55827) [OK-55828](https://onekeyhq.atlassian.net/browse/OK-55828) [OK-55832](https://onekeyhq.atlassian.net/browse/OK-55832) [OK-55850](https://onekeyhq.atlassian.net/browse/OK-55850) [OK-55856](https://onekeyhq.atlassian.net/browse/OK-55856) [OK-55857](https://onekeyhq.atlassian.net/browse/OK-55857)

----------

## Summary
- Fixes Private Send quote/error display, FAQ routing, keyboard dismissal, quote refresh timing, and send completion ordering.
- Updates Swap K-line stable-token handling and hides token switch controls when only one token is available.
- Keeps Swap K-line header price display in USD regardless of app default currency.

## Intent & Context
These changes collect the Private Send and Swap K-line follow-ups handled on `fix/private-send-0605`. The main goal is to make Private Send behave consistently with provider/backend contracts while tightening K-line display behavior after backend API adjustments.

## Root Cause
Private Send had several edge cases where frontend state moved faster than provider/backend state: provider quote errors were not shown in the intended card surface, quote refresh could continue during transaction build, and send completion returned before the Private Send stateTX order creation finished. Swap K-line handling also needed to follow updated backend stable-token metadata and avoid exposing unnecessary token-switch UI.

## Design Decisions
- Reuse provider/server error messages directly instead of reconstructing them on the client.
- Stop the Private Send quote refresh timer while building the transaction instead of introducing broader quote cancellation logic.
- Await stateTX only for Private Send completion, preserving normal Send/Swap behavior.
- Keep K-line token-selection and USD price behavior localized to the Swap K-line modal path.

## Changes Detail
- `packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx`: Private Send quote card/error display, quote refresh/build state handling, keyboard dismissal behavior.
- `packages/kit-bg/src/services/ServiceSwap.ts`: Private Send send-completion sequencing around stateTX order creation.
- `packages/shared/types/swap/SwapProvider.constants.ts`: Private Send FAQ link update.
- `packages/kit/src/views/Swap/pages/modal/swapKLineTokenUtils.ts`: updated stable-token token-selection mapping.
- `packages/kit/src/views/Swap/pages/modal/SwapKLineContent.tsx`: K-line single-token switch hiding and USD price/detail handling.
- `packages/kit-bg/src/services/ServiceMarketV2.ts`: optional MarketV2 USD price-detail fetch behavior for K-line.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop, Mobile, Web, Extension
- **Risk Areas**: Private Send async ordering and Swap K-line token selection/price display.

## Test plan
- [ ] Private Send quote error from provider appears in the quote card and disables preview as expected.
- [ ] Private Send build disables submit and stops auto quote refresh until build flow finishes.
- [ ] Private Send send completion waits for stateTX order creation before order polling starts.
- [ ] iOS Private Send amount keyboard can be dismissed from the amount-entry modal.
- [ ] Swap K-line prioritizes the non-stable token when backend stable-token metadata is returned.
- [ ] Swap K-line hides the switch button when only one token is available.
- [ ] Swap K-line header price is displayed in USD regardless of app default currency.

[OK-55750]: https://onekeyhq.atlassian.net/browse/OK-55750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55824]: https://onekeyhq.atlassian.net/browse/OK-55824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55827]: https://onekeyhq.atlassian.net/browse/OK-55827?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55828]: https://onekeyhq.atlassian.net/browse/OK-55828?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55832]: https://onekeyhq.atlassian.net/browse/OK-55832?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55850]: https://onekeyhq.atlassian.net/browse/OK-55850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55856]: https://onekeyhq.atlassian.net/browse/OK-55856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-55857]: https://onekeyhq.atlassian.net/browse/OK-55857?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/76a1e165-033b-41e6-86c0-41261d2368d2?refs=dce88c797cb6167eef0f2933319c7a68b369d5ae&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/76a1e165-033b-41e6-86c0-41261d2368d2/license?refs=dce88c797cb6167eef0f2933319c7a68b369d5ae&source=prChecksComment) |

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -2146,18 +2159,20 @@ function SendAmountInputContainer() {
...
+              } catch (error) {
+                defaultLogger.app.error.log(
+                  `Fetch private send initial tx state failed: ${
+                    error instanceof Error ? error.message : String(error)
+                  }`,
+                );
+              }
               await backgroundApiProxy.serviceSwap.addSwapHistoryItem(
                 swapHistoryItem,
               );
```
sidmorizon, line 2175: **Reordering `fetchPrivateSendInitialTxState` before `addSwapHistoryItem` can lose transaction history on mobile**

Severity: `non-severe`

The code now `await`s `fetchPrivateSendInitialTxState` (a network request to `/swap/v1/state-tx`) before saving the history item via `addSwapHistoryItem`. Previously, the history was saved first and the state fetch was fire-and-forget. If the network request hangs or the app is killed/backgrounded during the fetch (common on mobile), `addSwapHistoryItem` never executes and the user loses their private send transaction tracking entry. The transaction itself is on-chain but the user cannot see or monitor its status in-app.

<details><summary>Old vs New execution order</summary>

OLD (safe):
1. `await addSwapHistoryItem(...)` — history saved immediately
2. `void fetchPrivateSendInitialTxState(...).catch(...)` — fire-and-forget

NEW (risky):
1. `await fetchPrivateSendInitialTxState(...)` — blocks on network call
2. `await addSwapHistoryItem(...)` — only executes after network completes

</details>

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -1124,6 +1126,10 @@ function SendAmountInputContainer() {
         });
       }
 
+      if (privateSendQuoteError) {
+        return privateSendQuoteError;
+      }
```
sidmorizon, line 1131: **privateSendQuoteError in form validator bypasses balance/min-amount checks for PRIVATE mode**

Type: `Flag`

The new `if (privateSendQuoteError) return privateSendQuoteError;` at line 1129 is placed BEFORE vault-specific validation (balance checks) at line 1237. This means when there's a quote error, the form shows that error instead of balance/min-amount errors. However, `isInsufficientBalance` is independently checked in `isSubmitDisabled` (line 2338), so the submit button still shows 'Insufficient funds' when balance is exceeded. The competing messages (form error = quote error, button = insufficient) could be confusing to users but is likely an intentional prioritization of quote-level errors.

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
zhaono1: This ordering is intentional for Private Send. The history item is what kicks off the later Swap history polling path; if we persist it before the initial stateTX call, order-detail polling can start before the backend has created the Private Send order, which is the race fixed by OK-55850.

The stateTX call is still wrapped in try/catch and we continue to `addSwapHistoryItem` after failure, so a failed stateTX request does not drop the local history row. The intended tradeoff is to wait for the initial stateTX attempt before handing the item to the normal polling path, instead of reintroducing the premature polling failure.

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
zhaono1: This priority is intentional. For Private Send, the quote/provider result is the authoritative source for min/max/provider-specific failure messages, and OK-55828 explicitly moved those messages into this surface instead of showing the client-side composed min hint at the same time.

Balance remains a submit-gating concern and is still reflected by the footer button state/text. The form-level error intentionally prefers the quote/provider message so the user sees the exact provider reason returned by the quote API, without competing local validation text.

sidmorizon (dismissed):

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
sidmorizon, line 2175: **fetchPrivateSendInitialTxState changed from fire-and-forget to sequential await**

Type: `Flag`

Previously, `fetchPrivateSendInitialTxState` was called as a fire-and-forget (`void ... .catch(...)`) before `addSwapHistoryItem`. Now it's `await`ed sequentially before `addSwapHistoryItem` (lines 2162-2175). This means the history item is only persisted after the initial tx state fetch completes (or errors). This introduces additional latency in the `onSuccess` callback — if the initial state fetch is slow, the history entry won't appear until it completes. The error is caught so it won't block the flow, but the ordering change (`fetch state → then add history` vs old `add history || fetch state in parallel`) is a behavioral shift worth noting.

packages/kit-bg/src/vaults/impls/sol/Vault.ts
```diff
@@ -1013,14 +1013,15 @@ export default class Vault extends VaultBase {
...
-      !unsignedTx.swapInfo &&
       !unsignedTx.stakingInfo &&
-      instructions.some(
-        (instruction) =>
-          instruction.programId.toString() ===
-          ASSOCIATED_TOKEN_PROGRAM_ID.toString(),
-      ) &&
+      (!unsignedTx.swapInfo || transferPayload?.isPrivateSend === true) &&
+      hasCreateTokenAccountInstruction &&
       actions[0].assetTransfer
```
sidmorizon, line 1024: **Solana Vault condition expanded to include private send swaps for createTokenAccountFee**

Type: `Flag`

The condition at `Vault.ts:1021-1025` was changed from `!unsignedTx.swapInfo && !unsignedTx.stakingInfo` to `!unsignedTx.stakingInfo && (!unsignedTx.swapInfo || transferPayload?.isPrivateSend === true)`. When `unsignedTx.swapInfo` is truthy, `actions` comes from `buildInternalSwapAction` (line 993-998). The guard `actions[0].assetTransfer` at line 1025 still protects against setting `extraInfo` when the action doesn't have an asset transfer. For this to work as intended for private send, `buildInternalSwapAction` must produce an action with a truthy `assetTransfer` field for same-token private send swaps. If it produces a swap-type action without `assetTransfer`, the `createTokenAccountFee` info won't be set, silently defeating the purpose of this change.

originalix (approved):

PR #11936, opened 2026-06-06
fix: eliminate two desktop renderer freeze render storms

huhuanming:
## Summary

Fixes desktop renderer freezes caused by **two independent React render storms**, both surfacing as `Maximum update depth exceeded`. When either storm ignites, the renderer's main thread pins at 100% (V8 JS + GC), the heap balloons, and the window becomes unresponsive.

The two storms are unrelated in mechanism but produce the same symptom, so they are fixed together here.

---

## Storm A — `@tamagui/focus-scope` ref-detach loop

**Where:** `@tamagui/focus-scope` `useFocusScope` (FocusScope.* / compose-refs).

**Root cause:** the ref callback passed to `useComposedRefs` was an **inline arrow function**, so it had a new identity on every render. `compose-refs` treats an identity change as "the ref changed" and, during commit, detaches the old ref (`safelyDetachRef`) then attaches the new one. The detach path runs `setContainer(null)` inside a `startTransition`, which re-renders, which produces **another new inline callback**, which is again seen as a changed ref → detach → `setContainer(null)` → … a self-feeding render storm.

It ignites in practice when a FocusScope overlay (Tooltip / Dialog / Popover / Sheet) is **open while its subtree gets unmounted** (e.g. navigating away with the overlay still mounted).

**Fix:** wrap the ref callback in `React.useCallback(…, [])` so its identity is stable across renders. `compose-refs` no longer sees a phantom ref change, the detach/attach thrash stops, and the loop is eliminated. Applied via `patch-package` to all build outputs (cjs/esm/jsx) and `src`.

---

## Storm B — `PageHeader` → `navigation.setOptions` loop

**Where:** `packages/components/src/layouts/Page/PageHeader.tsx` → `@react-navigation/core` `useNavigationCache.setOptions`.

**Root cause:** `PageHeader` resolves header options with `reload(props)`, which returns a **brand-new object on every render**. That fresh object is used as a `useLayoutEffect` dependency that calls `navigation.setOptions(...)`:

```
reload(props) → new object every render
  → useLayoutEffect dep changes every render
  → navigation.setOptions(...)            // setState in @react-navigation
  → navigator re-renders → descriptor rebuilt
  → screen (PageHeader) re-renders
  → reload(props) → new object again → …  // Maximum update depth
```

Most pages have stable enough option content that React bails out after a render or two. On heavy, frequently re-rendering pages — **Earn** is the reproducible case (`EarnHome → EarnPageContainer → TabPageHeader → Page.Header`, re-rendering under jotai/refresh churn) — the loop crosses the threshold and runs away.

This is long-standing code, not a recent regression; it only tips over under sufficient re-render pressure.

**Fix:** keep a stable reference for the resolved options and only swap it when the content actually changes, so the layout-effect dependency stops churning. `reload()`'s output is a **flat** object whose values are reference-stable by convention (callers wrap header render functions in `useCallback`), so a cheap **shallow own-key comparison** is sufficient — no recursive deep-equal on this hot, every-page path. The effect now calls `setOptions` only when options genuinely change, breaking the self-sustaining loop.

`reload()`'s generic pass-through design is intentionally left untouched: it forwards arbitrary navigation options across ~240 `Page.Header` call sites, so explicitly enumerating fields would risk silently dropping options. The shallow compare is conservative — it can never skip a real change (it only reuses the old reference when every own key is `Object.is`-identical).

---

## Reproduction & verification

A CDP repro harness is included: `apps/desktop/scripts/cdp-repro-gift-storm.mjs`. It drives the running desktop app — opens the Earn gift overlay (a FocusScope overlay), tab-storms at human cadence while it is open, then lands on Home → opens Settings — and counts `Maximum update depth` / FocusScope console errors and freeze (RTT timeout) hits per round. The `header-gift-action` testID added to `GiftAction` lets the harness locate and open the gift overlay.

```bash
# prereq: yarn app:desktop (CDP on 9222), heavy wallet, DevTools closed
REGRESSION=1 ROUNDS=300 node apps/desktop/scripts/cdp-repro-gift-storm.mjs
# REGRESSION=1 → exits non-zero if any storm reproduces
```

Findings:
- Before the focus-scope patch: Storm A reproduced reliably (~13%/round).
- With the focus-scope patch alone: a 655-round soak fully suppressed Storm A; Storm B still leaked once (round 228) — a transient `Maximum update depth` burst on the Earn header, heap spiking ~1.7GB.
- With both fixes: the Earn `PageHeader` self-loop no longer occurs.

## Scope

- `patches/@tamagui+focus-scope+1.108.0.patch` (new) — Storm A fix
- `packages/components/src/layouts/Page/PageHeader.tsx` — Storm B fix
- `apps/desktop/scripts/cdp-repro-gift-storm.mjs` (new) — repro harness
- `packages/kit/src/components/TabPageHeader/components/GiftAction.tsx` — `header-gift-action` testID for the harness

No existing behavior removed. `tsc` and `eslint` pass on the changed source.

### DIFF
--- a/.oxlintrc.json
+++ b/.oxlintrc.json
@@ -674,7 +674,7 @@
     },
     {
       // JS/JSX files - disable TypeScript-specific rules
-      "files": ["**/*.js", "**/*.jsx"],
+      "files": ["**/*.js", "**/*.jsx", "**/*.mjs", "**/*.cjs"],
       "rules": {
         "typescript/no-floating-promises": "off",
         "typescript/consistent-type-imports": "off",
--- a/apps/desktop/scripts/cdp-repro-gift-storm.mjs
+++ b/apps/desktop/scripts/cdp-repro-gift-storm.mjs
@@ -1,0 +1,246 @@
+/* eslint-disable no-console */
+/**
+ * cdp-repro-gift-storm.mjs — reproduce the FocusScope `Maximum update depth`
+ * freeze by storming away while a FocusScope overlay is OPEN.
+ *
+ * Insight (confirmed): unmounting a page alone is clean. The error stack is
+ * `safelyDetachRef` — a FocusScope overlay must be MOUNTED/OPEN when the storm
+ * unmounts its subtree. The DeFi/Earn header GIFT icon (GiftAction →
+ * HeaderIconButton with a `title` → wrapped in a tamagui <Tooltip>, and onPress
+ * opens the referral <Dialog>) is a FocusScope overlay: hover mounts the
+ * tooltip, click opens the dialog. (Requires testID="header-gift-action" on
+ * GiftAction.)
+ *
+ * Round: go to Earn → open the gift overlay (hover tooltip + click dialog) →
+ *        WITHOUT closing it, tab-storm → the open FocusScope subtree unmounts
+ *        mid-open → setContainer(null) storm → Maximum update depth → freeze.
+ *
+ * Prerequisite: yarn app:desktop (CDP 9222) + heavy wallet + DevTools closed.
+ * RUN: node apps/desktop/scripts/cdp-repro-gift-storm.mjs
+ *      REGRESSION=1 node ...   # after fix: PASS only if 0 hits
+ */
+
+import { chromium } from 'playwright-core';
+
+const CDP = process.env.CDP_URL || 'http://127.0.0.1:9222';
+const REGRESSION = process.env.REGRESSION === '1';
+const ROUNDS = Number(process.env.ROUNDS || 80);
+const FREEZE_RTT = Number(process.env.FREEZE_RTT || 2500);
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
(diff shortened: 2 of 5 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0853a007-829c-4c45-9eaf-03da9c33ddf8?refs=46af57efdbe1242fa5cca0dd8fac5890b75ba960&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0853a007-829c-4c45-9eaf-03da9c33ddf8/license?refs=46af57efdbe1242fa5cca0dd8fac5890b75ba960&source=prChecksComment) |

originalix (approved):

PR #11953, opened 2026-06-08
fix: normalize token price currency for double-convert and aggregated tokens(OK-55943)

weatherstar:
[OK-55943](https://onekeyhq.atlassian.net/browse/OK-55943)

----------

## Summary
- Fix token price/value rendering incorrectly under a non-USD display currency (e.g. CNY).
- Two distinct root causes in the same \"normalize to USD basis on fetch → convert USD→display at render\" pipeline.

## Intent & Context
Token fiat data (`ITokenFiat`) is normalized to a USD basis when fetched and tagged `currency: 'usd'`; the render layer `<Currency sourceCurrency={token.currency}>` converts USD → the user's display currency. Two separate defects broke this pipeline for different token sets, so prices were wrong in non-USD currencies — discovered while comparing USD vs CNY home token list (USD correct, CNY partially wrong).

## Root Cause
**1. Price double-convert (`ServiceToken.ts`).** `convertFiatToCurrency` guarded price normalization with `typeof fiat.price === 'number'`, but the API delivers `price` as a numeric **string** (despite the `number` type). The guard silently skipped price, leaving it in the request currency while every other fiat field was normalized to USD and the entry was tagged `'usd'`. The render layer then converted USD→display again, double-applying the rate (~rate² off).

**2. Aggregated token tag drop (`tokenUtils.ts`).** `flattenAggregateTokensMap` rebuilt the aggregated `ITokenFiat` but never copied `currency`, so multi-chain aggregated tokens (ETH, USDT, USDC) reached `<Currency>` with `currency: undefined`. `Currency` then fell back to the display currency (`sourceCurrency ?? currencyInfo?.id`) and ran a no-op conversion (rate≈1) — rendering raw USD numbers under a non-USD symbol (e.g. ETH shown as ¥1,672 instead of ¥11,335; whole entry, both price and fiatValue). Single-network tokens resolve via `tokenListMap` (which carries `currency: 'usd'`) and converted correctly, which is why only aggregated symbols were affected.

## Design Decisions
- **Price**: use a truthiness guard + `BigNumber` + `isFinite()` (matching the sibling fiat fields) so price normalizes regardless of number-vs-string shape, instead of widening `ITokenFiat.price` to `number | string` — that would require a downstream audit of every numeric consumer and is deferred.
- **Aggregated**: inherit `currency` from `firstEntry` in flatten — the same source `price`/`price24h` already use, and all network entries share the `'usd'` basis after normalization. This is minimal and fixes both the live-fetch and cache-restore paths at once, since every `flattenAggregateTokensMap` caller benefits.

## Changes Detail
- `packages/kit-bg/src/services/ServiceToken.ts` — replace the `typeof === 'number'` price guard with a truthiness + `BigNumber.isFinite()` guard; add an explanatory comment.
- `packages/shared/src/utils/tokenUtils.ts` — propagate `currency: firstEntry.currency` when building the aggregated `ITokenFiat` in `flattenAggregateTokensMap`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Extension / Mobile / Desktop / Web (shared `kit-bg` + `shared` utils)
- **Risk Areas**: currency-conversion correctness across all token lists; main risk is a regression for single-network tokens or USD display — both verified to stay correct by code tracing (tsc + lint pass).

## Test plan
- [ ] CNY display: ETH / USDT / USDC price AND balance convert by rate (ETH ≈ ¥11,335, not ¥1,672); top total corrects upward (no longer undercounts aggregated tokens).
- [ ] Single-network tokens (XRP / BTC / HYPE / RBTC / XLM) remain correct — no regression.
- [ ] Toggle CNY ↔ USD repeatedly; values stay consistent.
- [ ] Cold-start cache-restore shows correct aggregated token prices (cache path also flattens).
- [ ] Other non-USD currencies (EUR / JPY) convert correctly too.

[OK-55943]: https://onekeyhq.atlassian.net/browse/OK-55943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c66b9233-7baf-42eb-b7b8-4bb29310edbf?refs=e0d2fc909d6e4e1a8ea97daff7384cb792a7ccc3&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c66b9233-7baf-42eb-b7b8-4bb29310edbf/license?refs=e0d2fc909d6e4e1a8ea97daff7384cb792a7ccc3&source=prChecksComment) |

limichange (approved):

zhaono1 (approved):

packages/kit-bg/src/services/ServiceToken.ts
```diff
@@ -163,8 +163,16 @@ class ServiceToken extends ServiceBase {
...
+      // string, so a `typeof === 'number'` guard silently skips it — leaving
+      // price in the request currency while the render layer still converts
+      // USD -> display currency, double-applying the rate (CNY price ends up
+      // ~rate^2 off). Guard on truthiness like the fiat fields above.
+      if (fiat.price) {
+        const priceBn = new BigNumber(fiat.price);
+        if (priceBn.isFinite()) {
+          fiat.price = priceBn.div(rate).toNumber();
+        }
       }
```
sidmorizon, line 176: **Type mismatch between ITokenFiat.price (number) and runtime API data (string) is a systemic issue**

Type: `Flag`

The `ITokenFiat` type at `packages/shared/types/token.ts:57` declares `price: number`, but the PR comment confirms the API can deliver it as a string at runtime. This means TypeScript's type system provides a false guarantee here. While this PR correctly works around the mismatch with a BigNumber conversion, a more robust long-term fix would be to either: (a) coerce the API response at the deserialization boundary to ensure `price` is always a number, or (b) update the type to `price: number | string` to reflect reality. The current approach is pragmatic but leaves the type lie in place for other consumers.

originalix (approved):

PR #11981, opened 2026-06-10
fix: restore Perps orderbook cold-start rows

originalix:
## Summary
- Render cached Perps L2 orderbook rows on the first visible mobile frame by preserving the L2 book cold-start snapshot through app bootstrap.
- Add a coin-level latest fallback for Perps L2 book cache keys so a usable cached book can render even when the exact tick-option key is missing.
- Prevent the native mobile orderbook depth view from flashing back to empty rows when the first live WebSocket updates arrive.

## Intent & Context
The Perps orderbook previously benefited from the cold-start cache work and could show cached rows on the first frame. After the orderbook was rewritten with the native depth-bar component, the user observed two regressions during Perps cold start:

1. Once the Perps page became visible, the orderbook stayed blank for a noticeable time before cached rows appeared.
2. After cached rows rendered, the first live WebSocket updates could briefly replace rows with an empty `--` state before real rows returned.

The requested scope explicitly excluded the full-screen Perps loading window. The benchmark measured only the interval after other page elements were visible.

## Root Cause
The first-frame blank state had two causes:

- The Perps L2 book cold-start data was available in the context atom snapshot, but the orderbook could not read it early enough. It depended on Perps provider hydration or SWR storage state, both of which could happen after the page was already visible.
- The native depth-bar view initially received placeholder data. Even after the React side had a cached L2 book, the native view needed a first native prop seed before its imperative `setDepth` / `setText` path could show real rows.

The second empty-state flash was caused by the first native seed optimization itself: after seeding the native view once, later renders in the same epoch submitted empty `percents/prices/sizes` props to keep high-frequency Fabric prop writes low. The native view reacted to that empty props commit by drawing placeholder `--` rows, then recovered when the next imperative `setDepth` / `setText` update arrived.

## Design Decisions
- Keep display acceleration separate from trading interactivity. Cached rows can render immediately, but row selection still requires `isPerpsL2BookInteractive` freshness checks.
- Store only the current coin's exact tick-option cache key plus a coin-level `latest` fallback to keep the cold-start snapshot small.
- Preserve `__ONEKEY_PERPS_L2_BOOK_COLD_CACHE__` during mobile bootstrap so the orderbook can read the L2 book before Perps provider hydration and before the general context snapshot is cleaned up.
- Seed the native depth-bar view with real first-frame props per epoch, then keep that seed stable for the epoch while continuing to push live updates through the imperative native methods. This preserves the original performance optimization while removing the empty props flash.

## Changes Detail
- `apps/mobile/index.ts`: extracts the Perps L2 book cold-start cache from the pre-read Jotai context snapshot and stores it on `globalThis` during bootstrap.
- `packages/shared/src/utils/swrCacheUtils.ts`: adds Perps L2 book `latest` cache key fallback and storage reload support for synchronous cold-start reads.
- `packages/kit/src/views/Perp/utils/l2BookFreshness.ts`: adds reusable L2 book cache freshness helpers and global snapshot access.
- `packages/kit/src/views/Perp/hooks/usePerpMarketData.ts`: falls back from live atom to cold-start atom, global cold cache, and SWR cache while ignoring empty live books.
- `packages/kit/src/views/Perp/components/PerpOrderBook.tsx`: uses the same cold-cache fallback path for the initial visible mobile orderbook.
- `packages/kit/src/states/jotai/contexts/hyperliquid/*`: persists a compact Perps L2 book cold-start atom from live L2 updates.
- `packages/kit-bg/src/states/jotai/utils/index.ts`: preserves hydrated Perps L2 book cold cache on global scope.
- `packages/kit/src/views/Perp/components/OrderBook/AnimatedDepthBlock.native.tsx`: seeds the native depth column with real first-frame props and keeps the seed stable for the epoch to avoid `--` flashes on live updates.
- Tests cover cache key fallback, L2 book freshness/global cache behavior, and cold-start root mounting for the Perps L2 book snapshot.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile, Desktop, Web
- **Risk Areas**: Mobile native Perps orderbook first-frame rendering, Perps L2 book cache freshness, cold-start context hydration.

The native depth-bar change affects native mobile only. The cold-cache and fallback helpers are shared by the Perps orderbook path and are guarded by coin and age checks. Trading interactivity remains gated by live freshness checks, so stale cached rows are display-only.

## Test plan
- [x] `yarn lint --fix`
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn jest packages/kit/src/states/jotai/utils/jotaiContextStore.test.ts packages/kit/src/views/Perp/utils/l2BookFreshness.test.ts packages/shared/src/utils/swrCacheUtils.test.ts packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidCache.test.ts --runInBand`
- [x] iOS simulator frame-by-frame cold-start benchmark: before fix, visible orderbook blank lasted about 19-20 frames / 633-667ms; after fix, cached rows appeared in the first detected orderbook frame across 3 runs.
- [x] iOS simulator frame-by-frame live update benchmark: before fix, cached rows regressed to empty rows for about 21 frames / 700ms when live WebSocket data arrived; after fix, 3 cold-start runs showed no rows-to-empty regression.
- [x] Verified local self-test services stayed running on `8081`, `3001`, and `9222`.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e22ae613-bfa1-4b8a-84c3-3991242d1ab5?refs=4a52f81c71f03371eb2220bb3362beb07e3c5d96&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e22ae613-bfa1-4b8a-84c3-3991242d1ab5/license?refs=4a52f81c71f03371eb2220bb3362beb07e3c5d96&source=prChecksComment) |

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -1339,7 +1346,39 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
         bidLevels: data.levels?.[0]?.length ?? 0,
         askLevels: data.levels?.[1]?.length ?? 0,
       });
-      set(l2BookAtom(), withPerpsL2BookLocalReceivedAt(data));
+      const nextBook = withPerpsL2BookLocalReceivedAt(data);
+      set(l2BookAtom(), nextBook);
+      const now = Date.now();
+      if (
+        hasL2BookCacheableLevels(nextBook) &&
+        now - lastL2BookColdCacheWriteAt >=
```
sidmorizon, line 1354: # 🟡 P2 Cold-cache write throttle is cross-coin, not per-coin

## 问题

`lastL2BookColdCacheWriteAt` is a module-level singleton shared by every coin, and the write is gated by `now - lastL2BookColdCacheWriteAt >= PERPS_L2_BOOK_SNAPSHOT_CACHE_WRITE_INTERVAL_MS` (30s). When the user switches coin, the first L2 frames of the *new* coin are throttled out because the timestamp was just bumped by the *previous* coin. Since `set(perpsL2BookColdCacheAtom(), nextColdCache)` replaces the whole atom with only the current coin's keys, during that 30s window the persisted cold cache still holds the old coin.

## 影响

If the user opens coin `A`, switches to coin `B`, and closes the app within ~30s, coin `B`'s cold-start snapshot is never written — so the next cold start of `B` shows empty rows, which is exactly the regression this PR is meant to fix. Rapid coin browsing (a very common Perps flow) degrades cold-start coverage to whatever coin happened to win the 30s race.

## 建议

Key the throttle by coin so each coin gets its own write cadence:

```ts
const lastL2BookColdCacheWriteAtByCoin = new Map<string, number>();
// ...
const last = lastL2BookColdCacheWriteAtByCoin.get(data.coin) ?? 0;
if (hasL2BookCacheableLevels(nextBook) && now - last >= PERPS_L2_BOOK_SNAPSHOT_CACHE_WRITE_INTERVAL_MS) {
  lastL2BookColdCacheWriteAtByCoin.set(data.coin, now);
  // ...
}
```

[Claude session ↗](https://claude.ai/code/session_013hmWWKzbGKu1b5tusWB4vM)

---

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -243,6 +246,10 @@ function getLedgerUpdateKey(update: HL.IUserNonFundingLedgerUpdate): string {
   );
 }
 
+function hasL2BookCacheableLevels(data: HL.IBook | undefined) {
```
sidmorizon, line 249: # 🟡 P2 Cold cache accepts thin books; rest of L2 cache requires ≥16 levels/side

## 问题

`hasL2BookCacheableLevels` only requires `levels[0]?.length && levels[1]?.length` (i.e. ≥1 level per side) before writing the snapshot into `perpsL2BookColdCacheAtom`. The existing L2-snapshot cache path uses `isL2BookSnapshotCacheEntryComplete` in `ServiceHyperliquidCache.ts`, which requires `getL2BookSnapshotLevelCount(...) >= PERPS_L2_BOOK_SNAPSHOT_CACHE_MIN_LEVELS_PER_SIDE` (16) before treating a snapshot as renderable. The new cold-cache write bypasses that threshold entirely.

## 影响

A partial/thin frame (e.g. 1–2 levels per side, common on the very first WS frames or during a fast market) can be persisted as the cold-start snapshot and later rendered via the fallback chain in `PerpOrderBook`/`useL2Book`. The restored "cold-start rows" can then be a 1–2 row stub instead of the full ladder — a degraded first frame that is also inconsistent with the ≥16 completeness contract used elsewhere.

## 建议

Reuse the same completeness threshold for the cold-cache write so it matches the SWR/simpleDb path:

```ts
function hasL2BookCacheableLevels(data: HL.IBook | undefined) {
  return (
    (data?.levels?.[0]?.length ?? 0) >= PERPS_L2_BOOK_SNAPSHOT_CACHE_MIN_LEVELS_PER_SIDE &&
    (data?.levels?.[1]?.length ?? 0) >= PERPS_L2_BOOK_SNAPSHOT_CACHE_MIN_LEVELS_PER_SIDE
  );
}
```

[Claude session ↗](https://claude.ai/code/session_013hmWWKzbGKu1b5tusWB4vM)

---

huhuanming (approved):

PR #11987, opened 2026-06-10
fix: preserve swap input on account switch

originalix:
## Summary
- Preserve user-entered Swap token pair and amount when switching accounts or wallets.
- Keep existing cold-start cleanup behavior for non-input account owner changes.
- Avoid clearing user input during hidden Swap quote cleanup while still clearing stale quote event state.

## Intent & Context
QA reported that switching accounts or wallets should not wipe the token pair or amount the user already entered in Swap. The requirement was to add only this account-switch preservation behavior while keeping the previous Swap cold-start cases intact and avoiding unnecessary flicker.

## Root Cause
The visible symptom was that tokens could remain selected after switching from Account #1 to Account #2, but the amount still reset to empty. Runtime verification showed the amount was not cleared when leaving Swap for Home; it was cleared after selecting another account. The specific clearing path was `useSwapQuote` hidden-tab quote cleanup: when Swap was hidden and a quote event was still incomplete, it cleared quote event/list state and also called `setFromTokenAmount({ value: '', isInput: true })`.

There was also a second account-update listener in `SwapRootProvider` that could rewrite restored tokens independently from `useSwapGlobal`, so the preservation condition needed to be shared across the Swap initialization and account sync paths.

## Design Decisions
- Added a shared `shouldPreserveSwapUserInputOnAccountSwitch` helper so all account-switch related paths use the same guard.
- Preserve only when there are selected tokens and a user-input from/to amount, and no import params are driving the Swap state.
- Keep hidden quote cleanup behavior for quote event/list state, but skip amount clearing when the user input should be preserved.
- Keep old owner/wallet-change cleanup behavior when there is no user-entered amount, so this change does not broaden cold-start token preservation beyond the QA case.

## Changes Detail
- `useSwapQuote`: avoids clearing `fromTokenAmount` during hidden quote cleanup when user-entered Swap input should be preserved.
- `useSwapGlobal`: uses the shared preservation guard, tracks both from/to amount input refs, and preserves selected token context for account-switch input preservation.
- `useSwapAccount`: lets token-driven account-network sync continue when selected tokens are intentionally preserved for user input.
- `SwapRootProvider`: applies the same preservation guard to the root cold-start account update listener.
- `swapColdStartTokenCacheUtils`: adds the shared preservation helper and targeted preserve parameters for context validation/cleanup helpers.
- `swapColdStartTokenCacheUtils.test`: covers user-input preservation and confirms owner/wallet changes without user input still clear as before.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop, Mobile, Web
- **Risk Areas**: Swap account synchronization, cold-start selected token cleanup, and quote cleanup while Swap is hidden.

## Test plan
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn jest packages/kit/src/views/Swap/utils/swapColdStartTokenCacheUtils.test.ts --runInBand`
- [x] Desktop CDP: Account #1 -> Account #2 keeps `ETH -> USDC / 0.01` on first and settled Swap frames, with Select Token count 0.
- [x] Desktop CDP: Wallet switch to `Seed3 / Account #1` keeps `ETH -> USDC / 0.02` on first and settled Swap frames, with Select Token count 0.
- [ ] iOS simulator manual verification by QA/user.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2d645a4f-4d9c-472b-bbff-08f343b88b6d?refs=46d4ae9c058144ec9d18f5ac6351945d1bdda33b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2d645a4f-4d9c-472b-bbff-08f343b88b6d/license?refs=46d4ae9c058144ec9d18f5ac6351945d1bdda33b&source=prChecksComment) |

packages/kit/src/views/Swap/pages/SwapRootProvider.tsx
```diff
@@ -115,6 +124,18 @@ function SwapColdStartCacheSync() {
...
+          fromTokenAmount: fromTokenAmountRef.current,
+          hasSelectedTokens: Boolean(
+            swapFromTokenRef.current || swapToTokenRef.current,
+          ),
+          toTokenAmount: toTokenAmountRef.current,
+        })
+      ) {
+        markInitialSelectedTokensSynced();
+        return;
+      }
```
sidmorizon, line 155: **Missing scene name filter causes `markInitialSelectedTokensSynced` to fire on non-home account events**

Severity: `non-severe`

In `SwapColdStartCacheSync`, the `handleHomeSelectedAccountUpdate` handler listens to ALL `AccountSelectorSelectedAccountUpdate` events (from any scene: home, swap, discover, etc.). The new `shouldPreserveSwapUserInputOnAccountSwitch` check at lines 127-138 runs **before** any scene name filtering and calls `markInitialSelectedTokensSynced()` for any event when the user has entered a swap amount. This is inconsistent with `useSwapGlobal.ts:500-504` which explicitly filters for home scene events first. Before this PR, non-home events were effectively filtered by `shouldHandleSwapColdStartHomeAccountUpdate` → `shouldClearSwapSelectedTokensOnHomeAccountUpdate` → `isHomeMainAccountUpdate` (`swapColdStartTokenCacheUtils.ts:212`), so no side effects occurred. Now, the early return bypasses that check, prematurely setting `initialSelectedTokensSynced` to `true` — which permanently prevents `shouldSkipSwapDefaultSelectedTokenSync` from allowing re-initialization of default tokens.

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
```diff
@@ -872,6 +908,21 @@ export function useSwapInit(params?: ISwapInitParams) {
...
+        swapType: swapTypeSwitchRef.current,
+        fromToken: fromTokenRef.current,
+        toToken: toTokenRef.current,
+        swapNetworks: swapNetworksRef.current,
+      }) !== false
+    ) {
+      syncSelectedTokensColdStartSwapType();
+      updateSelectedTokensColdStartContext();
+      markInitialSelectedTokensSynced();
+      return;
```
sidmorizon, line 953: **Preserve logic intentionally allows cross-chain token retention on account switch**

Type: `Flag`

In `useSwapGlobal.ts:911-924`, when `shouldPreserveUserInputSelectedTokens()` returns true, the code calls `updateSelectedTokensColdStartContext()` which updates the cold start context to the current (new) account. This means if a user entered an amount for an ETH swap and then switches to a Bitcoin wallet, the ETH tokens are preserved AND the cold start context now associates them with the Bitcoin wallet account. On next cold start, these ETH tokens would be restored for the Bitcoin wallet. This is potentially confusing but appears intentional — the PR's goal is to respect explicit user input. The `getSelectedTokensColdStartLimitSupport` check at line 914 provides a safety valve only for LIMIT tab incompatibility, not for general network mismatch.

packages/kit/src/views/Swap/hooks/useSwapQuote.ts
```diff
@@ -770,12 +780,22 @@ export function useSwapQuote() {
...
+              });
             closeQuoteEvent();
             setSwapQuoteEventTotalCount({
               count: 0,
             });
             setSwapQuoteResultList([]);
-            setFromTokenAmount({ value: '', isInput: true });
+            if (!shouldPreserveUserInputAmount) {
+              setFromTokenAmount({ value: '', isInput: true });
+            }
```
sidmorizon, line 817: # 🟠 P1 保留输入金额但清空报价列表后无自动重新询价，导致 UI 处于残缺状态

## 问题

这段逻辑位于 `isHiddenModel` 分支（即 Swap tab 被弹窗遮挡，而非账户切换路径）。当存在在途询价时，代码先 `closeQuoteEvent()`、`setSwapQuoteEventTotalCount({ count: 0 })`、`setSwapQuoteResultList([])`，随后仅在 `!shouldPreserveUserInputAmount` 时才 `setFromTokenAmount({ value: '', isInput: true })`。

```ts
setSwapQuoteResultList([]);
if (!shouldPreserveUserInputAmount) {
  setFromTokenAmount({ value: '', isInput: true });
}
```

改动前，无条件执行的 `setFromTokenAmount({ value: '', isInput: true })` 会触发第 242-256 行的 effect（`!fromTokenAmount.value && fromTokenAmount.isInput`）重新调用 `quoteAction` 做一次清理/复位。现在保留金额时该 effect 因 `fromTokenAmount` 依赖未变化而不会再触发。

## 影响

在保留路径下，`swapQuoteResultList` 被清空、在途事件被 `closeQuoteEvent()` 关闭，但 `fromTokenAmount` 保持不变。当弹窗关闭、tab 重新聚焦（`isFocus`）时，第 762-774 行只重新注册事件 handler，并不会重新调用 `quoteAction`；其它防抖 effect 也因输入值未变化而不会重跑。结果是：界面仍显示输入金额，但报价列表为空且无任何在途询价，直到用户手动改动输入才会恢复——这是一个可见的残缺/陈旧状态（注意此路径不经由账户地址变化触发的 `quoteAction`，因此不会被那条 effect 兜底）。

## 建议

在保留输入金额的分支里，要么同样保留 `swapQuoteResultList`（不要 `setSwapQuoteResultList([])`），要么在 `shouldPreserveUserInputAmount` 为真时于重新聚焦后主动触发一次 `quoteAction`，确保金额与报价列表状态保持一致。

[Claude session ↗](https://claude.ai/code/session_01R39ZdJfnKtTsujHUz9EMFv)

---

originalix: Fixed

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
originalix: Fixed

packages/kit/src/views/Swap/pages/SwapRootProvider.tsx
originalix: Fixed

zhaono1 (approved):

huhuanming (approved):

packages/kit/src/views/Swap/pages/SwapRootProvider.tsx
```diff
@@ -115,6 +126,33 @@ function SwapColdStartCacheSync() {
...
```
sidmorizon, line 155: **Second preservation check in SwapColdStartCacheSync is unreachable dead code**

Severity: `non-severe`

`shouldPreserveSwapUserInputOnAccountSwitch` (line 145) can never return `true` when reached, because the preceding `shouldPreserveSwapUserInputAmountOnAccountSwitch` check (line 136) already covers all cases and returns early. Tracing the logic: `shouldPreserveSwapUserInputOnAccountSwitch` is defined as `hasSelectedTokens && shouldPreserveSwapUserInputAmountOnAccountSwitch(...)` (`swapColdStartTokenCacheUtils.ts:90-97`). Since neither call passes `hasImportParams`, both evaluate the same `hasSwapUserInputAmount(...)`. If check 1 passes (user has input amount), we return at line 142. If check 1 fails (no user input amount), check 2 also fails because it requires the same amount condition. This means the block at lines 144–155 is dead code that misleadingly suggests a second layer of protection exists.

Suggested fix:

```ts
      if (
        shouldHandleSwapColdStartHomeAccountUpdate({

```

packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
```diff
@@ -872,6 +922,36 @@ export function useSwapInit(params?: ISwapInitParams) {
...
+        swapType: swapTypeSwitchRef.current,
+        fromToken: fromTokenRef.current,
+        toToken: toTokenRef.current,
+        swapNetworks: swapNetworksRef.current,
+      }) !== false
+    ) {
+      syncSelectedTokensColdStartSwapType();
+      markInitialSelectedTokensSynced();
+      return;
+    }
```
sidmorizon, line 954: **Second preservation block in syncDefaultSelectedToken is unreachable dead code**

Severity: `non-severe`

The block at lines 941–954 (`shouldPreserveUserInputSelectedTokens()`) can never be entered. It is only reached when the first block (lines 925–940) did NOT fire. The first block fails in two scenarios: (A) `shouldPreserveUserInputAmount()` is false, or (B) `shouldPreserveUserInputAmount()` is true but `hasSelectedTokens && limitSupport === false`. In scenario A, `shouldPreserveUserInputSelectedTokens()` also returns false because it calls `shouldPreserveSwapUserInputAmountOnAccountSwitch` with the same ref values (`swapColdStartTokenCacheUtils.ts:90-97`). In scenario B, the second block requires `limitSupport !== false`, which contradicts `limitSupport === false`. Therefore the second block is unreachable, and the `shouldPreserveUserInputSelectedTokens` callback at `useSwapGlobal.ts:268-283` is created but never produces a meaningful result.

packages/kit/src/views/Swap/utils/swapColdStartTokenCacheUtils.ts
```diff
@@ -584,19 +653,24 @@ export function getSwapSelectedTokensColdStartContextNetworkId({
...
+  preserveSelectedTokens?: boolean;
   selectedTokensColdStartContext?: ISwapSelectedTokensColdStartContext;
   toToken?: ISwapToken;
 }) {
   if (!fromToken && !toToken) {
     return true;
   }
+  if (preserveSelectedTokens) {
+    return true;
+  }
```
sidmorizon, line 673: **preserveSelectedTokens bypasses cold start context validation entirely**

Type: `Flag`

When `preserveSelectedTokens` is true, `isSwapSelectedTokensColdStartContextValidForAccountNetworkSync` returns `true` immediately at line 671-673, bypassing all context matching logic. This means the debounced `checkTokenForAccountNetworkDebounce` in `useSwapAccount.ts:143` will proceed to sync account networks to match token networks, even if the cold start context is stale or belongs to a completely different account. While this is the intended preservation behavior, it means a user who typed an amount on Account A's ETH tokens and then switched to Account B (which may not support EVM) will have the swap account forcibly synced to ETH. If Account B is a Bitcoin-only wallet, this network sync may fail silently or produce unexpected UI state. This is a deliberate design tradeoff rather than a bug, but worth validating manually on hardware wallets or single-chain accounts.

PR #11996, opened 2026-06-10
fix(app-update): auto-start mid-session OTA downloads and apply silent updates on restart (OK-55397)

huhuanming:
[OK-55397](https://onekeyhq.atlassian.net/browse/OK-55397)

----------

## Summary

This PR fixes two distinct facets of the same "must restart before an update happens" report (OK-55397):

1. **Download won't auto-start mid-session** — a server-pushed silent/seamless update discovered while the app is already running now starts downloading automatically, instead of stalling at `notify` until the next cold start.
2. **Silent install UX** — once a silent update finishes downloading, it is applied on the next restart via a pending install task (and an immediate restart-install button), instead of popping a "ready" dialog.

## Background / Problem

After the client detects a server-pushed update, **the download does not start immediately** — the app must be restarted before it begins downloading the install package.

Desktop log evidence (`desktop-macosStore`, `updateStrategy: 3 = seamless`): the cold-start `/utility/v1/app-update` check returned empty (`missing_release_versions`); the update was only discovered mid-session (after cold start) → status sat at `notify` but the download never started; `builtinBundleVersion` only incremented across restart boundaries.

## Root cause

The automatic background download is only ever triggered by the single cold-start check:

- The only automatic `downloadPackage()` is wired into the first-launch dispatch in `AppUpdateForeground.tsx`, guarded by the module-level `didRunFirstLaunchDispatch` flag, so it runs **exactly once per app process lifetime**.
- Nothing re-triggers it mid-session: the `AppState 'active'` listener only calls `shouldResumeStalledDownload()` (which resumes only `downloadPackageFailed`/`downloadASCFailed`); `fetchAppUpdateInfo` only auto-downloads for `jsBundleRollback`, while a normal `jsBundleUpgrade`/seamless upgrade just sets status to `notify`.
- Desktop works around this only via the manual menu (`onCheckUpdate`, gated by `isDesktop`); **native has no equivalent automatic trigger** → it is affected the same way (confirmed via a parallel investigation).

The `Bootstrap.tsx` comment already acknowledges this: "the auto useEffect only runs once at startup, so mid-session OTA discovery would otherwise stall".

### 1. Auto-start the download mid-session (service layer; covers both desktop & native)
- `ServiceAppUpdate.fetchAppUpdateInfo`: after the status transitions to `notify`, auto-trigger `downloadPackage()` for `isAutoUpdateStrategy` (silent/seamless) `jsBundleUpgrade`/`appShellUpdate` decisions, symmetric to the existing `jsBundleRollback` auto-download (including the frozen/ignored target check and a `setTimeout` re-check of `notify` to avoid duplicate concurrent downloads). This mirrors the cold-start dispatch so mid-session discovery is handled identically.
- Added a shared `isAutoUpdateStrategy` (`shared/appUpdate`) and re-export it from kit's `updateStrategy.ts`, so kit-bg does not need to import from kit. Store-based app-shell updates are shipped as `manual` by the server, so they are naturally excluded.

### 2. Apply silent updates on restart instead of a "ready" dialog
- Silent updates no longer pop a "ready" dialog: `readyToInstall` already queues a pending install task (the `servicePendingInstallTask` strategy gate now allows `silent`), which is applied on the next restart; the header / reminder update button lets the user restart-install immediately.

## Testing / Validation

- ✅ `ServiceAppUpdate.test.ts` +5 cases (silent/seamless auto-download, manual does NOT download, frozen target does NOT download, rollback regression)
- ✅ Updated the affected silent-dialog case (`useAppUpdate.test.ts`) → asserts no dialog / no navigation
- ✅ All affected suites: 503 passed / 4 suites
- ✅ tsgo full-project type check: 0 errors
- ✅ eslint clean on changed files

[OK-55397]: https://onekeyhq.atlassian.net/browse/OK-55397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/servicePendingInstallTask.ts
+++ b/packages/kit-bg/src/services/servicePendingInstallTask.ts
@@ -667,9 +667,16 @@ class ServicePendingInstallTask {
 
     // Rollback always creates a pending task regardless of server strategy —
     // it is a corrective action that should not require user confirmation.
+    // OK-55397: silent now also queues a pending install task instead of
+    // showing a "ready" dialog, so the downloaded update is applied on the
+    // next restart (or immediately when the user taps the update button).
+    // silent + seamless together == isAutoUpdateStrategy, mirroring the
+    // shouldAutoDownload gate in ServiceAppUpdate.fetchAppUpdateInfo (kept
+    // inline here to avoid coupling to the shared helper's import path).
     if (
       decision.decision !== 'jsBundleRollback' &&
-      releaseInfo.updateStrategy !== EUpdateStrategy.seamless
+      releaseInfo.updateStrategy !== EUpdateStrategy.seamless &&
+      releaseInfo.updateStrategy !== EUpdateStrategy.silent
     ) {
       defaultLogger.app.appUpdate.pendingTaskUpsertDecision({
         traceId,
--- a/packages/kit/src/components/AppUpdate/AppUpdateForeground.tsx
+++ b/packages/kit/src/components/AppUpdate/AppUpdateForeground.tsx
@@ -29,7 +29,9 @@ import { whenAppUnlocked } from '../../utils/passwordUtils';
 import { tryShowFeaturedDialog } from '../../views/AppUpdate/dialogs/tryShowFeaturedDialog';
 
 import { buildSoftwareUpdateParams } from './updateAnalytics';
-import { showSilentUpdateDialogUI, showUpdateDialogUI } from './updateDialogs';
+// OK-55397: showSilentUpdateDialogUI no longer used — silent updates are
+// applied via a pending install task on restart instead of a "ready" dialog.
+import { showUpdateDialogUI } from './updateDialogs';
 import { isAutoUpdateStrategy, isForceUpdateStrategy } from './updateStrategy';
 import { useDownloadPackage } from './useDownloadPackage';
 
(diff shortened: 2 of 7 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e20ee41a-8907-48ed-aedd-d7d867c501e6?refs=f642fc51abae0ffa524ade15cbd77b2a67a06fd5&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/e20ee41a-8907-48ed-aedd-d7d867c501e6/license?refs=f642fc51abae0ffa524ade15cbd77b2a67a06fd5&source=prChecksComment) |

limichange (approved):

packages/kit/src/components/AppUpdate/AppUpdateForeground.tsx
```diff
@@ -399,12 +407,13 @@ export function useAppUpdateForegroundEffects(enabled = true) {
...
-          // tick (prevents a duplicate dialog when the persisted atom is
-          // already hydrated at first launch).
+          // OK-55397: silent no longer pops a dialog here. readyToInstall has
+          // already queued a pending install task (silent is allowed past the
+          // strategy gate), applied on the next restart; the update button
+          // offers an immediate restart-install. Keep the guard set so the
+          // (now no-op) silent-ready watcher below stays consistent.
           silentReadyDialogShown = true;
-          showSilentUpdateDialog();
+          // showSilentUpdateDialog();
```
sidmorizon, line 416: **Silent + ready on cold start now does nothing — relies entirely on pending install task**

Type: `Flag`

Previously, cold-starting with `status=ready` + `strategy=silent` would call `showSilentUpdateDialog()`, giving the user an explicit prompt. Now (`packages/kit/src/components/AppUpdate/AppUpdateForeground.tsx:409-416`) it just sets the module guard and returns. The update is expected to be applied by the pending install task system running independently during cold start. If the pending install task was lost (storage corruption, task creation failure in `readyToInstall`), the update would be stuck at `ready` with no automatic recovery. The user would still see the header update button (`isShowAppUpdateUIWhenUpdating` returns `true` for silent+ready), so there is a manual escape hatch, but the previous dialog was a stronger recovery mechanism.

packages/kit-bg/src/services/servicePendingInstallTask.ts
```diff
@@ -667,9 +667,16 @@ class ServicePendingInstallTask {
...
+    // showing a "ready" dialog, so the downloaded update is applied on the
+    // next restart (or immediately when the user taps the update button).
+    // silent + seamless together == isAutoUpdateStrategy, mirroring the
+    // shouldAutoDownload gate in ServiceAppUpdate.fetchAppUpdateInfo (kept
+    // inline here to avoid coupling to the shared helper's import path).
     if (
       decision.decision !== 'jsBundleRollback' &&
-      releaseInfo.updateStrategy !== EUpdateStrategy.seamless
+      releaseInfo.updateStrategy !== EUpdateStrategy.seamless &&
+      releaseInfo.updateStrategy !== EUpdateStrategy.silent
```
sidmorizon, line 679: **Silent strategy gate change in syncPendingInstallTaskWithReleaseInfo has broader impact**

Type: `Flag`

Adding `EUpdateStrategy.silent` to the strategy gate at `servicePendingInstallTask.ts:676-679` means silent updates now create pending install tasks when reaching `ready_to_install` stage. Previously only `seamless` and `jsBundleRollback` passed this gate. This is the core mechanism enabling apply-on-restart for silent updates. The corresponding dialog removal in `AppUpdateForeground.tsx:409-416` (commenting out `showSilentUpdateDialog()`) is consistent — without the dialog, the pending task is the only install trigger. However, this changes the user experience: previously silent updates showed a confirmation dialog before install; now they auto-install on restart. This is intentional per OK-55397 but is a significant behavioral change worth noting for QA.

originalix (approved):

PR #11997, opened 2026-06-10
fix(chart): symbol-aware SYMBOL_CHANGE ack for perps self-heal

huhuanming:
## What
Make the perps chart's `SYMBOL_CHANGE` 3s self-heal **symbol-aware** so the stale boot symbol can no longer false-ack it.

## Why
The page emits `barsState` / `renderReady` / kline messages for whatever symbol it is *currently* showing. The old ack was type-only, so on perps the **boot symbol's** data (HL:BTC) could cancel the `force:true` self-heal before our driven coin took — leaving the chart permanently stuck on the wrong coin (observed: perps `ETHUSDC` rendering BTC candles).

## How
- Track the last-driven raw coin in `drivenSymbolRef`.
- For `type === 'perps'`, only ack when the inbound message both carries `"symbol":"<drivenCoin>"` **and** is a real data request (`getHyperliquidPriceScale` / `getMarks` / `getKLineData`).
- Market path keeps the looser type-only ack (its messages don't reliably echo the symbol param and switching already works).

## Scope
`packages/kit/src/components/TradingView/ChartWebView/index.native.tsx` only (+30/-4). Native only.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/61804150-cc8b-483c-87ad-eeb11e319278?refs=092b6231f80abd2369aa4da3dce4d9f2207402f0&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/61804150-cc8b-483c-87ad-eeb11e319278/license?refs=092b6231f80abd2369aa4da3dce4d9f2207402f0&source=prChecksComment) |

packages/kit/src/components/TradingView/ChartWebView/index.native.tsx
```diff
@@ -406,10 +411,31 @@ export function ChartWebView({
...
+            const sym = drivenSymbolRef.current;
+            if (
+              !!sym &&
+              raw.includes(`"symbol":"${sym}"`) &&
+              (raw.includes('tradingview_getHyperliquidPriceScale') ||
+                raw.includes('tradingview_getMarks') ||
+                raw.includes('tradingview_getKLineData'))
+            ) {
+              switchAckedRef.current = true;
+            }
```
sidmorizon, line 437: **Perps ack relies on tradingview_getMarks carrying symbol, but symbol is optional in the type**

Type: `Flag`

The market message handler (`useTradingViewMessageHandler.ts:336-340`) types the `tradingview_getMarks` data with `symbol?: string` (optional). If a marks request is ever emitted without a `symbol` field, the `raw.includes('"symbol":"${sym}"')` check at line 431 would fail and the ack wouldn't fire for that message. This is the safe direction — it means the self-heal timer fires and resends with `force:true` (a one-time extra re-render) rather than falsely acking. The perps handler types it as `IGetMarksRequest` which likely includes symbol, but worth confirming that the chart bundle always populates it.

limichange (approved):

PR #12027, opened 2026-06-12
fix: build cloud sync items outside IndexedDB tx when creating indexed accounts (OK-56267)

sidmorizon:
[OK-56267](https://onekeyhq.atlassian.net/browse/OK-56267)

----------

## Summary

Fixes [OK-56267](https://onekeyhq.atlassian.net/browse/OK-56267): on the browser extension, batch-creating accounts (账户管理勾选地址后创建) intermittently failed with an error, after which clicking 新建账户 became unresponsive.

### Root cause

`txAddIndexedAccount` called `buildExistingSyncItemsInfo({ tx })` **inside** the IndexedDB transaction. That call awaits non-DB promises — cloud sync credential fetch, keyless credential hydration (storage reads), and async encrypt/decrypt of sync items. IndexedDB auto-commits a transaction as soon as the event loop is reached with no pending IDB requests, so any of those awaits crossing a macrotask boundary killed the transaction. The subsequent sync-item write then threw:

```
Failed to execute 'get' on 'IDBObjectStore': The transaction has finished.
```

Confirmed by the user log attached to the Jira issue — the stack trace matches this exact chain frame by frame (`batchBuildAccounts → addBatchCreatedHdOrHwAccount → withTransaction → txAddIndexedAccount → txWithSyncFlowOfDBRecordCreating → txAddAndUpdateSyncItems → txAddRecords → store.get`). The intermittency comes from cache state: with warm caches the in-tx awaits are microtask-only and survive; on a cold MV3 service worker they hit storage/crypto and the tx dies — which is why this is extension-specific. The codebase already acknowledges this failure mode in a comment at `CloudSyncFlowManagerBase.ts` line 376.

### Fix

Follow the same pattern already used by `addAccountsToWallet` (and validated by the existing `updatePassword` tests): **prepare outside the tx, verify + write inside the tx**.

- New `prepareIndexedAccountsCreationData()` builds indexed-account records (incl. async `idHash`) and cloud sync items with non-tx reads (`buildExistingSyncItemsInfo({ tx: undefined })`) before the transaction opens.
- `addIndexedAccount` (batch create path) pre-builds, then the transaction only re-checks existence and performs pure DB writes.
- `addHDNextIndexedAccount` (create-new-account path) pre-computes the next free index outside the tx; inside the tx the index is re-derived and, on a concurrent conflict, a `LocalDBIndexedAccountIndexConflictError` triggers a re-prepare + retry (max 3 attempts).
- Wallet creation flows (`createHDWallet` / HW / QR / keyless) intentionally keep the legacy in-tx path unchanged to limit blast radius — flagged as a follow-up.
- Hardening for the "unresponsive" symptom: the pre-transaction keyless credential warm-up in `LocalDbBaseContainer.withTransaction` is now wrapped with a 5s timeout + catch so it can never block or fail account-bucket transactions.

### Tests

- New `LocalDbBaseIndexedAccount.test.ts`: asserts sync items are built outside the transaction (`tx: undefined`, never during an open tx), skipIfExists filtering, next-free-index creation, and the conflict-retry path.
- Existing suites pass: `LocalDbBase.test.ts` (createHDWallet legacy path untouched), `LocalDbIndexed.test.ts`.
- `yarn tsc:staged` exit 0, `yarn lint:staged` 0 warnings / 0 errors.

https://claude.ai/code/session_01QiuzjAkvexgiB9nQqjwxMn

---
_Generated by [Claude Code](https://claude.ai/code/session_01QiuzjAkvexgiB9nQqjwxMn)_

[OK-56267]: https://onekeyhq.atlassian.net/browse/OK-56267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/5ca4805f-8d97-4a47-9072-4d5bf989241f?refs=110d3791ed114d7ca9556a1c49099fcf32db8fae&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/5ca4805f-8d97-4a47-9072-4d5bf989241f/license?refs=110d3791ed114d7ca9556a1c49099fcf32db8fae&source=prChecksComment) |

sidmorizon:
Addressed both Codex P2 findings in `14370de3` (rebased onto latest `x`):

**P2 — `addHDNextIndexedAccount` retry budget exhausted under concurrent bursts**
`addHDNextIndexedAccount` is now serialized per wallet with a `Semaphore`, so concurrent calls each prepare against the latest committed state and allocate distinct indexes instead of all racing for the same one. Within the single bg process this makes index conflicts effectively impossible; the in-tx conflict check stays as cross-context defense and the retry cap was raised so it no longer fails while a free index exists.

**P2 — stale prepared sync items written after the in-tx recheck**
When the in-tx existence recheck narrows `indexedAccountsToAdd` (a concurrent flow created some), the pre-built `newSyncItems`/`existingSyncItems` are now filtered down to the surviving accounts before entering `txWithSyncFlowOfDBRecordCreating`, so this tx never writes/uploads a sync row for an account it didn't create. `prepareIndexedAccountsCreationData` returns a deterministic `indexedAccountId → syncItemId` map for this; added a regression test covering the partial-overlap case.

Also folded in two items from an independent review of the same diff:
- `LocalDBIndexedAccountIndexConflictError` now sets `className` + an `EOneKeyErrorClassNames` entry, so it survives serialization boundaries (matches the `LocalDBRecordNotFoundError` pattern).
- The keyless-credential warm-up timeout in `LocalDbBaseContainer.withTransaction` is now cleared when hydration wins, avoiding a dangling 5s timer per account-bucket transaction.

`tsc` and the `LocalDbBaseIndexedAccount` / `LocalDbBase` suites pass on the rebased base.

---

packages/kit-bg/src/dbs/local/LocalDbBaseContainer.ts
```diff
@@ -104,7 +104,29 @@ export abstract class LocalDbBaseContainer implements ILocalDBAgent {
...
+      // timeout per transaction.
+      let warmupTimer: ReturnType<typeof setTimeout> | undefined;
+      try {
+        const warmupTimeout = new Promise<void>((resolve) => {
+          warmupTimer = setTimeout(resolve, 5000);
+        });
+        await Promise.race([
+          appGlobals.$backgroundApiProxy.serviceKeylessCloudSync.hydrateKeylessSyncCredentialFromStorageIfNeeded(),
+          warmupTimeout,
+        ]);
```
sidmorizon: **Unhandled promise rejection when hydration timeout wins the Promise.race**

Severity: `non-severe`

When the 5-second `warmupTimeout` resolves before `hydrateKeylessSyncCredentialFromStorageIfNeeded()`, `Promise.race` settles and execution continues. The hydration promise continues running in the background with no rejection handler. If it later rejects, this becomes an unhandled promise rejection — the outer `catch` block has already exited and won't catch it. This violates the repo's coding pattern rule (`.cursor/rules/coding-patterns.mdc`) to "Avoid floating promises to prevent unhandled rejections". On platforms where unhandled rejections are fatal (or with `--unhandled-rejections=throw`), this could crash the process.

Suggested fix:

```ts
        const hydrationPromise =
          appGlobals.$backgroundApiProxy.serviceKeylessCloudSync.hydrateKeylessSyncCredentialFromStorageIfNeeded();
        // Prevent unhandled rejection if the timeout wins the race
        hydrationPromise.catch(() => {});
        const warmupTimeout = new Promise<void>((resolve) => {
          warmupTimer = setTimeout(resolve, 5000);
        });
        await Promise.race([hydrationPromise, warmupTimeout]);
```

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -2334,31 +2620,110 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
+        break;
+      }
+      if (maxLoop >= 1000) {
+        break;
+      }
+      nextIndex += 1;
+      maxLoop += 1;
+    }
+    return nextIndex;
+  }
```
sidmorizon, line 2660: **OK-56267 fix only covers addIndexedAccount and addHDNextIndexedAccount paths**

Type: `Flag`

The PR's stated goal (OK-56267) is to move async cloud-sync-item building outside the IndexedDB transaction to prevent premature auto-commit. However, the fix only applies to `addIndexedAccount` (line 2106) and `addHDNextIndexedAccount` (line 2603-2641). Other callers of `txAddHDNextIndexedAccount` — specifically `createHDWallet` (`LocalDbBase.ts:3266`), `createKeylessWallet` (`LocalDbBase.ts:3344`), and the two HW/QR wallet creation paths (`LocalDbBase.ts:3992`, `LocalDbBase.ts:4508`) — still invoke `txAddHDNextIndexedAccount` without `prepared`, falling through to the legacy path that calls `buildExistingSyncItemsInfo` with a `tx` parameter inside the transaction. These paths remain susceptible to the same IndexedDB auto-commit issue on web/extension platforms. This may be intentional (incremental fix) or an oversight.

packages/kit-bg/src/dbs/local/LocalDbBaseContainer.ts
sidmorizon: Fixed in `2f9a9837`. Extracted the hydration promise and added `void hydrationPromise.catch(() => undefined)` before the race, so if the 5s timeout wins, a later hydration rejection can't surface as unhandled. Used the `void`-prefixed form so it also stays clean under `no-floating-promises`. The race still resolves/rejects exactly as before.

---

packages/kit-bg/src/dbs/local/LocalDbBase.ts
sidmorizon: Intentional for this PR — keeping `createHDWallet` / `createKeylessWallet` / HW / QR on the legacy in-tx path, tracked as a follow-up.

Reasoning: these paths aren't in the OK-56267 repro/log (which is the batch + 新建账户 flow via `addBatchCreatedHdOrHwAccount → txAddIndexedAccount`), and each builds only a single first-account sync item — far less in-tx async than the batch path that was actually failing. Extending them is also a heavier refactor: `prepareIndexedAccountsCreationData` reads `getWallet()`, but during wallet creation the wallet record doesn't exist yet (it's written inside the same tx), so it needs a separate prepare driven by the in-memory wallet-to-be-created. I'll do that in a dedicated follow-up PR rather than widen the blast radius into wallet + credential + context-counter creation here. Leaving this thread open to track it.

---

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -2124,6 +2170,194 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
+    // derives only from walletXfp + index (not name), so it is stable across the
+    // name fix-up above. Used to drop sync items for accounts removed by the
+    // in-tx recheck, so a concurrent creator's sync row is never overwritten.
+    const syncItemIdByIndexedAccountId: Record<string, string> = {};
+    for (const target of targets) {
+      const keyInfo = await syncManager.buildSyncKeyInfo({ target });
+      if (keyInfo?.key) {
+        syncItemIdByIndexedAccountId[target.targetId] = keyInfo.key;
+      }
+    }
```
sidmorizon, line 2351: **Sync item id assumed to equal buildSyncKeyInfo().key**

Type: `Flag`

The `syncItemIdByIndexedAccountId` map at `LocalDbBase.ts:2345-2351` stores `keyInfo.key` from `buildSyncKeyInfo`, then uses it to filter `newSyncItems` and `existingSyncItems` by their `.id` at lines 2416-2421. This assumes the sync item's `id` field equals the deterministic key from `buildSyncKeyInfo`. The test mock at `LocalDbBaseIndexedAccount.test.ts:77-85` validates this assumption by making `buildSyncKeyInfo.key` match `newSyncItems[].id`. If the production `buildExistingSyncItemsInfo` ever uses a different id scheme, the filtering would silently drop all sync items. Worth confirming this invariant holds in `CloudSyncFlowManagerBase`.

sidmorizon: Confirmed: the invariant `syncItem.id === buildSyncKeyInfo().key` holds structurally in `CloudSyncFlowManagerBase`, not just in the test mock.

The sync item id is *defined* as the sync key everywhere it is built or read:

- **newSyncItems**: `buildSyncItem` → `buildSyncKeyAndPayload` produces `key`, passed to `cloudSyncItemBuilder.buildSyncItemFromRawDataJson({ key })`, which sets `id: key` (`cloudSyncItemBuilder.ts:201`).
- **existingSyncItems**: `txGetSyncItem` derives the same `key` via `buildSyncKeyAndPayload`, then loads the record with `txGetRecordById({ id: key })` (`CloudSyncFlowManagerBase.ts:324-328`) — the row's primary key *is* the sync key.
- **syncItemIdByIndexedAccountId**: stores `buildSyncKeyInfo().key`, and `buildSyncKeyInfo` is just a thin wrapper over the same `buildSyncKeyAndPayload` (`CloudSyncFlowManagerBase.ts:179-182`).

All three resolve to the identical `buildSyncKeyAndPayload(target).key`, so they cannot diverge by construction. The codebase already relies on this same invariant elsewhere — e.g. `_buildInitSyncDBItems` reads `getSyncItemSafe({ id: keyInfo.key })` (`:821-828`) and `syncToScene` gates on `if (keyInfo.key === item.id)` (`:752`). The test mock just mirrors this structural fact rather than papering over a coincidence, so the filter at `LocalDbBase.ts:2410-2421` will not silently drop sync items.

Resolving as confirmed.

---

packages/kit-bg/src/dbs/local/LocalDbBase.ts
sidmorizon: Confirmed as expected design for this PR: `createHDWallet` / `createKeylessWallet` / HW / QR creation paths intentionally stay on the legacy in-tx path and are not in scope here. They each build only a single first-account sync item (far less in-tx async than the batch flow that OK-56267 actually reproduced), and extending them requires a separate prepare driven by the in-memory wallet-to-be-created since the wallet record doesn't exist yet inside the same tx. Tracked as a dedicated follow-up. Resolving.

---

originalix (approved):

PR #12068, opened 2026-06-16
feat: tokenList cells — per-cell render isolation + BG-runtime ViewModel (delete legacy whole-map atoms, slice 1)

huhuanming:
## What

TokenList **cells** refactor (normalized BG view-model) — slice 1. Normalizes the home token list into a `listStructure` (ids + aggregate membership + owner/generation) plus **per-key fiat/meta cells**, moves frame production into a background-runtime `ServiceTokenViewModel`, and **deletes the legacy per-store Jotai atoms** that used to hold whole token lists/maps in the UI runtime.

> **One-line effect:** a price tick now writes only the changed per-key cell → **only that one row re-renders**, instead of the whole-map atom notifying every subscriber and re-rendering the entire list.

## Why — Before vs After

```mermaid
flowchart TB
  subgraph BEFORE["❌ BEFORE — everything in the UI runtime"]
    direction TB
    f1["fetch round"] --> wm["whole-map atoms<br/>tokenListMap · allTokenListMap ·<br/>aggregateTokensMap · riskyTokenListMap · …<br/>(+ 50-owner rendered cache)"]
    wm -->|"merge / filter / sum / flatten<br/>on the UI thread, every refresh"| derive["heavy derivation<br/>blocks UI thread"]
    wm -->|"price tick sets the WHOLE map"| churn["EVERY subscriber re-renders<br/>whole list + every row + footer<br/>O(all tokens) per tick"]
  end

  subgraph AFTER["✅ AFTER — split BG / UI, per-cell granularity"]
    direction TB
    a1["fetch round"] --> bg["BG ServiceTokenViewModel<br/>buildFrames() off the UI thread"]
    bg -->|"structure frame (low-freq)"| struct["listStructure<br/>ids + membership + generation"]
    bg -->|"valuation frame (per round)"| cells["per-key cells<br/>1 atom per key"]
    cells -->|"applyValuationFrame:<br/>fiatEqual → set ONLY changed"| leaf["only the changed leaf<br/>re-renders · O(changed) per tick"]
  end

  BEFORE -.refactor.-> AFTER
```

| Axis | Before | After |
| --- | --- | --- |
| **Re-render / price tick** | `O(all tokens)` — whole-map atom notifies all subscribers | `O(changed)` — only the changed cell's leaf |
| **Derivation thread** | UI main thread, every refresh | BG runtime (off the UI thread) |
| **FlatList `data`** | new array each refresh → churn / re-layout | `orderedIds` ref-stable (`shallowEqual`) → no churn |
| **Cold start** | whole maps persisted per owner; double-authority revival bug; long blank | slim compact bundle, instant T0 paint, single-writer |
| **Write path** | 7 scattered `refresh*` writers, hard to test | 1 pure `apply` contract, injected deps, 115 node-jest tests |

## Data flow (push small, pull large)

```mermaid
sequenceDiagram
  autonumber
  participant Poll as TokenListBlock poller (~15s, focused only)
  participant BG as BG ServiceTokenViewModel
  participant Bus as appEventBus (lossy, fire-and-forget)
  participant UI as UI receive-shell (cells)

  Poll->>BG: ingestRound(settled slices)
  BG->>BG: buildFrames() → structure? + valuation(FULL fiat map)
  BG--)Bus: PUSH valuation (every round)
  BG--)Bus: PUSH structure (only when membership/meta changes)
  Bus--)UI: onValuation(frame)
  UI->>UI: applyValuationFrame — per-cell fiatEqual,<br/>set ONLY changed cells
  Note over UI: only changed leaf re-renders
  UI->>BG: PULL getTokenListFrames (mount / owner-switch backstop)
  Note over BG,UI: largest merged map (raw list + risky + flat-agg)<br/>is PULL-only — never pushed
```

**Wire vs apply — the one thing that confuses everyone:** the valuation frame carries the **FULL** fiat map on the wire every round (named `changedFiatById`, but it is a complete idempotent snapshot — *not* a delta). The “changed-only” behaviour happens at **apply** time: `applyValuationFrame` runs a per-cell `fiatEqual` guard and `set`s only cells whose value actually moved, so only those leaves notify/re-render. Full-on-wire is deliberate: a frame dropped by the lossy bus self-heals on the next round (a delta would leave a permanently stale cell). The stale `"carries only changed values"` docs that implied a wire-level delta are corrected in this PR.

## Performance reasoning

- **Render fan-out collapses from `O(n)` to `O(Δ)` per tick.** Old: one `set` on a whole-map atom re-renders every component reading it — a 200-token list with one moving price reconciles ~200 rows. New: the valuation apply touches only the moved cells, so the same tick reconciles only those rows. On a busy market list this is the dominant win and the reason `useTokenFiat` is a deliberately isolated leaf (do **not** hoist fiat to row level).
- **UI thread freed during refresh.** merge / filter / aggregate-sum / flatten moved into the BG runtime (`buildFrames`), so a heavy all-network refresh no longer competes with rendering on the main thread (directly serves the mandatory native dual-runtime split).
- **FlatList stops thrashing.** `orderedIds` is kept reference-stable, so a price tick never changes the list `data` identity → no full re-layout; structure (membership) and valuation (price) are decoupled channels.
- **New cost, honestly:** the split introduces BG↔UI serialization that the all-in-UI-heap model never paid (≈2 `cloneDeep` + 1 bridge marshal per emit). It is bounded by design — **~15s poll cadence, focused-tab only, and the clone runs on the BG thread**, with payload `O(token count)` of small `ITokenFiat` records (6 fields each). Negligible for typical wallets (sub-ms–few-ms / 15s); a few-ms / 15s on 1000+ token all-network wallets, still off the UI hot path. The largest merged map is kept off the push path entirely (PULL-only, “推小拉大”).

## Memory reasoning

```mermaid
flowchart LR
  subgraph B["BEFORE — UI heap"]
    direction TB
    b1["~8 overlapping whole-map atoms"]
    b2["renderedTokenListCache<br/>up to 50 owners × WHOLE maps<br/>(persisted to cold-start)"]
  end
  subgraph A["AFTER"]
    direction TB
    a1["UI: 1 cell projection (current owner)<br/>+ ids-only listStructure<br/>+ in-memory switch cache (no whole-map persist)"]
    a2["cold-start: SLIM compact bundle<br/>(few fields, not whole maps)"]
    a3["BG: vmByOwner (MRU-capped)<br/>authoritative per-owner copy"]
  end
  B -.->|net: fewer redundant UI copies<br/>+ smaller persisted footprint| A
```

- **UI heap: fewer redundant copies.** The old model kept the same token data duplicated across ~8 overlapping whole-map atoms (list + fiat map + aggregate + risky + small-balance …) plus a 50-owner rendered cache holding *whole* maps. The new model collapses that into one per-key cell projection for the **current owner** (GC’d with the store via `WeakMap`) plus an ids-only `listStructure`.
- **Cold-start persisted footprint shrinks.** The persisted authority moves from per-owner whole maps to a single **slim compact bundle** (only the fields a cold paint needs). Less to write to MMKV/IDB and less to parse at T0 → faster, lighter cold start; also removes the double-authority revival sink (the 50-owner whole-map cache no longer persists).
- **Net trade:** the current owner’s data now lives in **both** runtimes (BG authority + UI cells = ×2 for one owner), and BG holds an MRU-capped `vmByOwner`. This replaces the old UI-side *multi-atom* duplication of the same maps, so total resident footprint is comparable-to-lower while the **persisted** and **per-tick-churned** footprints drop clearly.

## How (commit arc, 23 commits)

- **Phase-1** — normalize home into structure + cells; migrate every home/footer/selector/search reader off the legacy display atoms; delete the 8 legacy home atoms (grep-gated, terminal).
- **Phase-2 BG** — `ServiceTokenViewModel` (`ingestRound` → `buildFrames` → structure/valuation frame PUSH over appEventBus + `getTokenListFrames` PULL backstop); UI `useTokenListCellsProducer` becomes subscribe-then-pull; cold-start slim cache.
- **R0 (`ff5d58b7e1`)** — additive BG infra for the last 3 atoms: a 3rd **risky frame** (idempotent full snapshot, monotonic version, membership+balance sync change-gate, PULL backstop) + `getRawTokenList()` / `getAllTokenListMap()` getters. Deletes nothing.
- **R1 (`4cc80afa07`)** — delete `riskyTokenListAtom` + `riskyTokenListMapAtom`; `TokenListFooter` reads the risky frame.
- **R2+R3 (`729ceea7eb`)** — delete `allTokenListAtom` + `allTokenListMapAtom`; 23 consumers migrate to per-key cells / per-sub-token map / one-shot snapshots, sourced from the BG ViewModel.

## Correctness notes (adversarially reviewed)

Three BLOCKING traps were caught in red-team review and explicitly guarded:
- **Switch skeleton** — `TokenListView` compares the *lagging settled owner* (from `getRawTokenList`), not the scoped current owner, so the account/network switch skeleton still fires.
- **Aggregate badge/auto-select** — icon/name/selector read the **per-network sub-token** fiat (`useAggregateSubTokenFiatMap`), never the summed `aggCell`.
- **Token Manager reactivity** — `useTokenManagement` keeps `listStructure.generation` as a reactive dep + body PULL, so it doesn't freeze.

Two product decisions: search keeps the home cache (no extra round-trip); modal/callback consumers snapshot at open-time (survive >8-owner BG eviction).

BG frame path is fully synchronous (no nextTick — v6.3.0 Android constraint); frames are idempotent full snapshots over the lossy appEventBus.

## Tests

- `ServiceTokenViewModel.test.ts` + `aggregateSubTokenFiat.test.ts` green, covering the risky frame (membership + balance gate, monotonic independent version, synchronous push, PULL/eviction) and the per-network-subcell-vs-summed-aggCell regression guard.
- Cells pipeline node-jest suite green (`115/115`): pure / buildFrames / apply / projection / homeProjection / coldStart / cutover / seamGate.
- `grep=0` for all deleted atoms + their `use*`/`refresh*` symbols in live code.
- `tsc:staged` + `lint:staged` clean per phase.

## Follow-ups (identified)

- **Snapshot PULL dedup** — `useHomeTokenListSnapshot` fans out N identical `getRawTokenList`+`getAllTokenListMap` PULLs across co-mounted home consumers per structure-generation; collapse to one shared inflight result per `ownerKey@generation`.
- **Skip duplicate `useTokenManagement` on home** — gate the inner `TokenListView` call with `!isHomeProjectionPath` (its `customTokens` is unused on the home path).

## Not done yet (why draft)

- Real-device / desktop perf gate (`tokensSpanMs` / `tokensStartMs` before/after) — node layer is green, device not yet verified.
- UI smoke: home list, risky footer count, TokenSelector badge/auto-select, instant search, account-switch skeleton.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0e1ac058-35cb-4c61-991c-da832ff10536?refs=994ab0eb40b6962810a7ace6e6b63c6de53728d2&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/0e1ac058-35cb-4c61-991c-da832ff10536/license?refs=994ab0eb40b6962810a7ace6e6b63c6de53728d2&source=prChecksComment) |

packages/kit-bg/src/services/ServiceTokenViewModel.ts
```diff
@@ -0,0 +1,555 @@
...
+    if (!raw) {
+      return {};
+    }
+    const riskyMap: Record<ITokenKey, ITokenFiat> =
+      this.frames.getFrames(ownerKey).risky.payload?.riskyMap ?? {};
+    return {
+      ...raw.tokenListMap,
+      ...riskyMap,
+      ...flattenAggregateTokensMap(raw.aggregateTokensMap),
+    };
```
sidmorizon: **`getAllTokenListMap` returns stale risky token fiat when only price moves**

Severity: `non-severe`

`getAllTokenListMap` (`ServiceTokenViewModel.ts:545-546`) reads `riskyMap` from the risky frame's last-pushed payload (`this.frames.getFrames(ownerKey).risky.payload?.riskyMap`). However, the risky frame is only pushed when `riskyChanged` returns true (membership or per-key balance change; `ServiceTokenViewModel.ts:413`). On a pure price tick (same risky $keys + same balances, only `fiatValue`/`price` moved), the risky frame is intentionally NOT re-pushed. The current round's fresh `riskyMap` is passed as a parameter to `ingestRound` but is never stored in the raw blob (`IRawTokenListData` at `ServiceTokenViewModel.ts:106-126` has no `riskyMap` field; the `setPullBlob` at line 398-405 omits it). So `getAllTokenListMap` composes stale risky prices until the next membership/balance change triggers a new risky frame push. The legacy `allTokenListMapAtom` always wrote the current round's riskyMap unconditionally, so this is a behavioral regression for the risky-token price slice consumed by `useHomeTokenListSnapshot().map` and downstream (WalletActions gas check, UniversalSearch, FiatCrypto).

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
```diff
@@ -0,0 +1,430 @@
...
+          networkId: result.networkId ?? '',
+        }),
+        { ...result, origin: 'live' },
+        generation,
+      );
+      if (progressiveFlushTimerRef.current === null) {
+        progressiveFlushTimerRef.current = setTimeout(() => {
+          void flushProgressiveView('progPaint');
+        }, PROGRESSIVE_PAINT_THROTTLE_MS);
+      }
```
sidmorizon, line 415: **Progressive paint timer may fire with a stale `flushProgressiveView` closure on rapid owner switches**

Type: `Flag`

In `useTokenListReactivePipeline.ts:384-388`, `ingestLiveRound` schedules a `setTimeout` that captures the current `flushProgressiveView` callback in its closure. If the owner switches rapidly (within the 350ms throttle window), the `reset()` call (which clears the timer via `clearTimeout`) must be called before the new owner starts ingesting. The facade returns `reset` for the consumer to call, but the correctness depends on the consumer (TokenListBlock) calling `reset()` on every owner switch BEFORE the new `ingestLiveRound` fires. If the consumer sequence is: `ingestLiveRound(oldOwner)` → 100ms → owner switch → `reset()` → this is fine (timer cleared). But if `reset()` is somehow delayed past the 350ms window, the old flush would execute with the old owner's closure and the old owner's rounds — which would pass the flush's own owner guard (it compares `rounds[0].ownerAccountId` against the closure-captured old owner). The `ingestMergedSnapshot` would then fire an `ingestRound` for the old owner's key. The BG VM handles this gracefully (it just overwrites the old owner's slot), but it's wasted work. Depends on consumer calling `reset()` promptly.

packages/kit/src/components/TokenListView/useTokenFiatField.ts
```diff
@@ -0,0 +1,168 @@
...
+      m.NativeLogger.write(
+        m.LogLevel.Info,
+        `[TLNATIVE] leaf key=${$key} seam=${useCellSeam ? 1 : 0} agg=${
+          aggregate ? 1 : 0
+        } cellField=${JSON.stringify(cellField)} mapHas=${!!mapToken}`,
+      );
+    } catch {
+      /* noop */
+    }
+  }
```
sidmorizon: **Extensive temporary debug logging left in production code paths**

Type: `Flag`

The `tln()` function (native file logger) is called throughout critical hot paths: `ServiceTokenViewModel.ts:383-391`, `useTokenFiatField.ts:99-113`, `TokenListView/index.tsx:92-100`, `useAllNetwork.ts:46-54`, `useTokenListReactivePipeline.ts:56-62`, and `apply.ts:38-47`. While each call is wrapped in try/catch (so it never crashes), the repeated `require()` call + JSON.stringify inside render paths (e.g. `useTokenFiatField.ts:108` stringifies `cellField` on every native_-key or jupSOL token render) adds measurable overhead on native. The comments mark these as `[TLNATIVE temp]` intended for removal, but they ship in this PR. On non-native platforms the require throws and the catch swallows — wasted exception creation per call.

packages/kit/src/components/TokenListView/TokenListFooter.tsx
```diff
@@ -80,29 +83,62 @@ function TokenListFooter(props: IProps) {
...
+  const { allAggregateTokenMap, ownedAggregateTokenListMap } =
+    useTokenListViewContext();
 
-  const [smallBalanceTokenListMap] = useSmallBalanceTokenListMapAtom();
+  // cells home pipeline: the footer runs under the HOME tokenList provider whose
+  // per-store cells are live + BG-fed after the Phase-2 cutover. The small-
+  // balance rows/total/modal params are rebuilt from `listStructure` + the
+  // per-key cells instead of the legacy whole-map atoms (full-delete PR-4).
+  // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+  const store = useTokenListContextData().store!;
```
sidmorizon, line 95: **TokenListFooter reads store directly via `store.get(meta(...))` — tightly coupled to having a mounted store**

Type: `Flag`

In `TokenListFooter.tsx:93`, the code does `const store = useTokenListContextData().store!` with a non-null assertion, then uses `store.get(meta(store, id))` / `store.get(cell(store, id))` throughout the component (lines 121, 214-215, 229-233). The non-null assertion is documented as safe because 'the footer runs under the HOME tokenList provider', but `TokenListFooter` is rendered as a child of `TokenListView` which is also used by the AssetList modal (`TokenList.tsx`). The AssetList modal DOES mount its own `withTokenListProvider`, so a store exists — but it's a DIFFERENT store than home, and no cells producer writes to that store's cells. The `store.get(cell(...))` calls would return `undefined`, making `smallBalanceTokenListMap` empty and the low-value modal params blank. This might manifest as an empty low-value sheet when opened from the AssetList modal (not the home list).

packages/kit-bg/src/services/ServiceTokenViewModel.ts
huhuanming: Confirmed and fixed in e2d8e32507.

Root cause exactly as described: `getAllTokenListMap` read `riskyMap` from the gated risky-frame payload, which `riskyChanged` suppresses on a pure price tick. Fix decouples the PULL composition from the push gate: `riskyMap` is now stored on the **every-round** raw blob (`IRawTokenListData.riskyMap`, written unconditionally in `setPullBlob`) and `getAllTokenListMap` composes from `raw.riskyMap`. The raw blob is rewritten every round, so the composed map always carries the current round's risky prices — restoring the legacy `allTokenListMapAtom` behavior.

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
huhuanming: Addressed in e2d8e32507. Rather than relying on the consumer's `reset()` timing, the deferred timer now fires through the existing P1-e `flushProgressiveViewRef` indirection instead of capturing `flushProgressiveView` in its closure:

```ts
progressiveFlushTimerRef.current = setTimeout(() => {
  void flushProgressiveViewRef.current?.('progPaint');
}, PROGRESSIVE_PAINT_THROTTLE_MS);
```

So on a rapid owner switch the timer runs the **latest** flush, whose owner guard compares the deferred rounds against the **current** owner (old rounds + new owner → mismatch → SKIP) regardless of when `reset()` lands. `flushProgressiveView` was dropped from `ingestLiveRound`'s deps accordingly. The per-flush capture-once + re-check-after-await semantics (P1-f) are unchanged.

packages/kit/src/components/TokenListView/useTokenFiatField.ts
huhuanming: Fixed in e2d8e32507. Removed all temporary `[TLNATIVE]` instrumentation — the duplicated `tln` logger definitions and every call site across the home token-list hot paths (`ServiceTokenViewModel`, `useTokenFiatField`, `TokenListView/index`, `useAllNetwork`, `useTokenListReactivePipeline`, `apply.ts`, `TokenListBlock`), plus the log-only dead computations they fed (e.g. `liveOwner`, the apply.ts SHRINK detector + `_skipped`/`_nonZeroChanged`). `grep -rn 'TLNATIVE\|tln(' packages/kit packages/kit-bg` is now empty; tsc + eslint clean.

packages/kit/src/components/TokenListView/TokenListFooter.tsx
huhuanming: Not a real issue — `TokenListFooter` is never rendered in the AssetList modal, so its store/cell reads can't go stale there.

The footer is gated behind the `withFooter` prop (`TokenListView/index.tsx`: `{withFooter ? <TokenListFooter … /> : null}`). The AssetList modal mounts `TokenListView` **without** `withFooter` (`AssetList/pages/TokenList.tsx`), and it feeds its own display data via the `host*` props rather than the cells pipeline. So `store.get(cell(...))` in the footer only ever runs under the HOME provider, where the cells producer does write. No empty low-value sheet from this path. Leaving as-is.

packages/kit-bg/src/states/jotai/contexts/tokenList/cellsPure/pure.ts
```diff
@@ -0,0 +1,275 @@
...
+  }
+  if (!a || !b) {
+    return false;
+  }
+  return (
+    a.balance === b.balance &&
+    a.fiatValue === b.fiatValue &&
+    a.price === b.price &&
+    a.currency === b.currency
+  );
```
sidmorizon, line 121: 🟠 P1 `fiatEqual` omits `price24h` → stale 24h price-change badge on the home cell path

## 问题

`fiatEqual` 只比较 `balance / fiatValue / price / currency`，不含 `price24h`：

```ts
return (
  a.balance === b.balance &&
  a.fiatValue === b.fiatValue &&
  a.price === b.price &&
  a.currency === b.currency
);
```

而 `applyValuationFrame` (`cells/apply.ts:354`) 用它作为唯一写门：`if (!deps.fiatEqual(deps.get(cellAtom), next)) deps.set(...)`。同一 owner 的某轮 valuation 若 `balance/fiatValue/price/currency` 全部逐字节相等、仅 `price24h` 变化，`fiatEqual` 返回 `true`，cell atom 不被 `set`，对应 cell 永不更新。

但 home cell-path 的涨跌幅叶子直接读取 `price24h`：`TokenPriceChangeView` → `useTokenPrice24h($key)` → `selectPrice24h(f) => f?.price24h`（`useTokenFiatField.ts:105,138`），数据源就是这个 cell。聚合行经 `aggCell` → `sumAggregateEntry`（取 first entry 的 `price24h`），sub-cell 同样被 `applyValuationFrame` 的 `fiatEqual` 门控，存在相同问题。

旧路径叶子订阅整个 `tokenListMap[$key]` 对象、每轮整体替换，`price24h` 始终刷新；切到 per-cell + `fiatEqual` 门后这是一处行为回归。

## 影响

`price24h` 是 24 小时涨跌百分比，会随服务端 24h 基准滚动而独立于现价变化。最典型的是**稳定币**（USDT/USDC）：现价恒为 `$1`（`price` 逐轮逐字节相等）但 `price24h` 在 ±0.0x% 间波动 → 该行涨跌幅徽标长期停在旧值，直到出现一次结构帧或现价发生可见变化才被动刷新。属于金融数值展示的可见错误。

## 建议

把 `price24h` 纳入 `fiatEqual` 的比较字段：

```ts
return (
  a.balance === b.balance &&
  a.fiatValue === b.fiatValue &&
  a.price === b.price &&
  a.price24h === b.price24h &&
  a.currency === b.currency
);
```

`price24h` 是数字标量，`===` 比较安全；该字段一动即触发 cell `set`，开销可忽略（valuation 每轮已携带全量 map，只是放行更多 changed cell）。

[Claude session ↗](https://claude.ai/code/session_01NPjHUGB3UovRDD9XMJoRVe)

---

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1717,554 +1471,218 @@ function TokenListBlock({
...
+    }
+    const shouldSyncTokenFilterToOverview =
+      allNetworksResult[0].syncTokenFilterToOverview;
+
+    // Build the authoritative snapshot THROUGH the LWW materialized view (facade,
+    // design §2): ∩ enabledKeys so failed-but-still-enabled networks keep their
+    // cache floor (I2) while removed/disabled networks evict; per-round
+    // merge-derive flags resolved inside. P0-b: the snapshot is RETURNED so the
+    // worth write below can read `snapshot.accountsWorth` BEFORE the commit.
+    const snapshot = await buildAuthoritativeSnapshot();
```
originalix, line 1575: 这里一旦跨过 `await buildAuthoritativeSnapshot()`，后面还会继续经历多次 `await`，但在 `updateLocalAggregateToken*`、`updateAccountWorth` 和 `commitAuthoritativeIngest(snapshot)` 之前没有再次确认 owner 仍然没变。

如果用户在这个窗口里切换账户或网络，当前闭包里的 `snapshot` 还是旧 owner 的结果，而 `commitAuthoritativeIngest()` 最终会通过 `cellsIngestInputsRef.current.ownerKey` 读到新 ownerKey。这样旧 owner 的 merged snapshot 会被写进新 owner 的 BG VM / frame cache，随后主线程 UI 可能拿到串号后的资产列表。

建议在这里以及后续 analytics `await` 之后用 `activeOwnerRef` 或本地 generation/cancel token 再做一次 owner 校验，只有 owner 未变化时才继续 commit。

packages/kit/src/states/jotai/contexts/tokenList/cells/useTokenListCellsProducer.ts
```diff
@@ -0,0 +1,297 @@
...
+                valuation: pulled.valuation,
+              } satisfies IValuationPush)
+            : undefined,
+      },
+      {
+        kind: 'risky',
+        eventName: EAppEventBusNames.TokenListRiskyFrame,
+        // risky version is independent + may be -1 for an owner with no risky
+        // set; floor 0 drops the unknown-owner / no-risky PULL.
+        gate: { floorVersion: 0 },
```
originalix, line 283: 当上一个 owner 有 risky token、下一个 owner 没有任何 risky token 时，这里的 `floorVersion: 0` 会把 `getTokenListFrames()` 返回的 `riskyVersion = -1` / 空快照直接丢掉，而 owner 切换路径里又没有先把 `riskyListFrameAtom` 清空。

结果是 Home footer 继续保留上一个 owner 的 `riskyTokens/riskyMap`，风险提示和 `RiskTokenManager` 入口会出现在错误的账户上，属于明确的 owner 数据串用。

建议给 risky 通道补一条“空 owner snapshot 也能清空”的路径，或者在 owner 变化时先显式 reset `riskyListFrameAtom`，这样无 risky owner 不会继承旧状态。

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
```diff
@@ -0,0 +1,417 @@
...
+  ownerKey: string;
+  nonZeroInputs: {
+    keepDefault?: boolean;
+    homeDefaultTokenMap?: Record<string, IHomeDefaultToken>;
+    customTokens?: ICustomTokenItem[];
+  };
+}
+
+/** One per-network LOCAL-cache slice fed to the cache-seed (L1 SWR floor). */
+export interface ICacheSeedItem {
```
originalix, line 78: 当 `onekeyall` 命中本地缓存且 owner 持有聚合资产时，这个 cache-seed contract 只带了 `tokenList / smallBalanceTokenList / riskyTokenList / tokenListMap`，没有 `aggregateTokenListMap` 和 `aggregateTokenMap`。

后面的 `seedAndFlushCache() -> buildMergedAllNetworkSnapshot()` 因此只能还原普通 token/risky slices，aggregate row、badge 和聚合估值要等 live round 才会补齐。这样首屏虽然命中了缓存，但用户先看到的是半成品列表，不是上一次完整的聚合视图。

建议把 aggregate 两个 slice 补进 `ICacheSeedItem`，并沿 cache request 链一路透传或在 seed 前重建，这样 cache-first paint 才能和 live 结果保持一致。

packages/kit/src/views/AssetSelector/pages/AggregateTokenSelector.tsx
```diff
@@ -425,14 +441,14 @@ function AggregateTokenSelector() {
   const sortedAggregateTokens = useMemo(() => {
```
originalix, line 441: 这里的排序和 `hideZero` 依赖 `homeTokenFiatMap`，但这个 map 来自 `useHomeTokenListSnapshot()`，它只会在 home `listStructure.generation` 变化时重新 pull，不会随着 valuation tick 实时刷新。

当 modal 打开后发生纯估值或余额变化时，行内金额会通过 `useAggregateSubTokenFiat` 继续更新，但这里的排序和过滤仍然停留在旧 map 上，最终出现“数值已经变了，列表顺序/显示却没变”的自相矛盾状态。

建议这个 modal 的排序和 `hideZero` 直接基于 reactive sub-cells 推导，不要再依赖 structure-gated snapshot map。

packages/kit-bg/src/states/jotai/contexts/tokenList/cellsPure/pure.ts
huhuanming: Fixed in 70a1cdc57a: `price24h` is now included in `fiatEqual`, with a unit test covering price24h-only valuation changes so the home price-change badge updates on pure 24h ticks.

packages/kit/src/views/AssetSelector/pages/AggregateTokenSelector.tsx
huhuanming: Fixed in 70a1cdc57a: `AggregateTokenSelector` now sorts and applies hide-zero from the reactive aggregate sub-token fiat map instead of the structure-gated home snapshot, so modal ordering/filtering follows valuation ticks while open.

packages/kit/src/states/jotai/contexts/tokenList/cells/useTokenListCellsProducer.ts
huhuanming: Fixed in 70a1cdc57a: the main-side cells producer now clears the risky frame for the current owner before subscribing/pulling frames, so an owner with no risky snapshot cannot inherit the previous owner’s risky footer state.

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
huhuanming: Fixed in 70a1cdc57a: cache seed items now carry aggregate list/map slices. The all-network local-cache path rebuilds aggregate rows from cached tokens before seeding, and the pipeline forwards those slices into `buildMergedAllNetworkSnapshot()` so cache-first paint includes aggregate rows and valuation.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
huhuanming: Fixed in 70a1cdc57a: the authoritative all-network path now re-checks `activeOwnerRef` after `buildAuthoritativeSnapshot()` and after the later analytics awaits before updating local aggregate cache, account worth, VM frames, or token-list state.

packages/kit-bg/src/states/jotai/contexts/tokenList/cellsPure/pure.ts
huhuanming: Fixed in 70a1cdc57a: `fiatEqual` now includes `price24h`, and the pure helper test covers price24h-only ticks so the home cell price-change badge updates when only 24h movement changes.

packages/kit/src/states/jotai/contexts/tokenList/cells/useTokenListCellsProducer.ts
huhuanming: Fixed in 70a1cdc57a and tightened in 71c4c201b8: the main producer clears the risky frame on owner reset, and the footer now ignores risky frames whose owner stamp does not match the current home owner key. A no-risky owner cannot inherit the previous owner’s risky UI.

packages/kit/src/views/AssetSelector/pages/AggregateTokenSelector.tsx
huhuanming: Fixed in 70a1cdc57a: `AggregateTokenSelector` now sorts and applies `hideZero` from the reactive aggregate sub-token fiat map instead of the structure-gated home snapshot, so modal ordering and filtering track valuation ticks while open.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
huhuanming: Fixed in 70a1cdc57a: after `buildAuthoritativeSnapshot()` and the later analytics awaits, the all-network completion path re-checks `activeOwnerRef` before updating aggregate caches, account worth, or committing the authoritative ingest. Stale main-runtime owner work exits without writing.

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
huhuanming: Fixed in 70a1cdc57a: cached all-network token items now rebuild and pass `aggregateTokenListMap` and `aggregateTokenMap` into the cache-seed contract, so `seedAndFlushCache()` restores aggregate rows, badges, and valuation before live results arrive.

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
```diff
@@ -0,0 +1,421 @@
...
+      if (!rounds.length || !enabled) {
+        return;
+      }
+      if (
+        rounds[0].ownerAccountId !== ownerAccountId ||
+        rounds[0].ownerNetworkId !== ownerNetworkId
+      ) {
+        return;
+      }
+      const roundsWithFlag = await resolveRoundsWithMergeFlag(rounds);
```
originalix, line 275: 当这次 progressive flush 在 `resolveRoundsWithMergeFlag(rounds)` 期间发生 owner 切换时，这里的二次 guard 仍然只比较闭包里旧的 `ownerAccountId / ownerNetworkId`，不会看到当前 owner 已经变化。

后面 `ingestMergedSnapshot()` 又会从 `cellsIngestInputsRef.current.ownerKey` 读取最新 ownerKey，所以旧 owner 的 `rounds` 会被重新组装后写进新 owner 的 BG VM / frame cache，形成跨账户资产串用；`reset()` 也不会 bump epoch，挡不住这条已经在飞的 flush。

建议把 owner/generation 做成 live cancel token，并在 `await` 之后、`buildMergedAllNetworkSnapshot()` 之前按当前 token 再校验一次，只有 token 未变化时才允许 ingest。

packages/kit/src/states/jotai/contexts/tokenList/cells/useHomeTokenListSnapshot.ts
```diff
@@ -0,0 +1,139 @@
...
+  // `ingestRound` WRITE side (TokenListBlock) uses, including the merge-derive
+  // `indexedAccountId` axis. Computing it anywhere else risks a write/read drift
+  // that silently misses the BG entry and returns EMPTY.
+  const ownerKey = useHomeTokenListOwnerKey(num);
+
+  // Reactive trigger consumed as a dep below: bumps only on a structure frame
+  // (not price ticks), so the snapshot re-pulls when the home list changes.
+  const structureGeneration = listStructure.generation;
+
+  const { result } = usePromiseResult(
```
originalix, line 129: 当 account / network 切换后，这个 hook 只是用新的 `ownerKey` 重新发起 PULL，但 `usePromiseResult` 在没有 `swrKey` 或显式 reset 的情况下会继续保留上一次 `result`，直到新请求真正返回。

结果是 More / Receive / Buy / ScanQrCode / FiatCrypto 这类通过闭包读取 snapshot 的 action，在切换后的窗口里仍然拿到前一个 owner 的 `tokens / map`，并把旧资产列表继续传给路由、扫码或弹窗逻辑。

建议在 `ownerKey` 变化时先同步回落到 `EMPTY_SNAPSHOT`，或给这里加一个 owner-scoped reset 机制，让旧 owner 的 snapshot 不会穿过切换窗口。

packages/kit/src/views/AssetList/hooks/useTokenManagement.ts
```diff
@@ -33,14 +34,26 @@ export function useTokenManagement({
...
   const isAllNetwork = networkId === getNetworkIdsMap().onekeyall;
-  const [tokenList] = useAllTokenListAtom();
+  // R-#3b: keep a REACTIVE trigger (the home structure generation, which bumps
+  // only when the list membership/structure changes — a cheap pushed signal) as
+  // a `usePromiseResult` dep, then PULL the merged raw list inside the body. A
+  // naive "one-shot PULL + dropped dep" would freeze this Token Manager modal
+  // (newly-discovered tokens would never appear). The displayed `tokenList.tokens`
+  // is replaced by the PULL result.
+  const [listStructure] = useListStructureAtom();
+  const ownerKey = buildOverviewOwnerKey(accountId, networkId);
```
originalix: 这里用 `buildOverviewOwnerKey(accountId, networkId)` 生成 BG VM 的读取 key，但 `serviceTokenViewModel` 在 merge-derive 模式下是按 `indexedAccountId__networkId` 存 owner 的，不是按具体 derive `accountId`。

一旦用户在 BTC / LTC 这类 merge 模式下打开 Token Manager，`getRawTokenList(ownerKey)` 就会直接 miss 到空列表，后面的 `mergeTokenMetadataWithCustomDataBatch()` 只能看到 custom / hidden token，当前 home token 集合会在管理弹窗里消失。

建议不要在这里复用 overview key，而是把 canonical token-list ownerKey 从调用方透传进来，或按 `getTokenListOwnerCacheAccountId` 的同一规则重建。

packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.ts
huhuanming: Fixed in 994ab0eb40: progressive flush now captures the owner token/generation before resolving merge flags, then re-checks the live owner generation and ownerKey after the await before building or ingesting. The BG ingest receives the captured owner token, and a regression test covers owner switching while the flush is in flight.

packages/kit/src/states/jotai/contexts/tokenList/cells/useHomeTokenListSnapshot.ts
huhuanming: Fixed in 994ab0eb40: the home token-list snapshot now carries an ownerKey@generation cache key and returns EMPTY_SNAPSHOT whenever the retained usePromiseResult value does not match the current owner/generation. That prevents action closures from reading the previous owner while the new PULL is still loading.

packages/kit/src/views/AssetList/hooks/useTokenManagement.ts
huhuanming: Fixed in 994ab0eb40: Token Manager now reads the BG raw list with the canonical token-list owner key. The modal passes useHomeTokenListOwnerKey from the home provider, and the shared hook also supports mergeDeriveAddressData/tokenListOwnerKey so merge-derive networks read indexedAccountId__networkId instead of the concrete derive accountId.

originalix (approved):

PR #12075, opened 2026-06-16
fix: hide delisted (TRASH) networks and remove Joystream from preset network list OK-56327

weatherstar:
[OK-56327](https://onekeyhq.atlassian.net/browse/OK-56327)

----------

## Summary
- Hide backend-delisted (`status: TRASH`) networks from the wallet network list — they previously kept showing (with a broken icon) long after being delisted.
- Make the client status-aware end to end: persist server networks with their real status and filter `TRASH` at the `getAllNetworks` merge layer; refresh open selectors only when the set actually changed.
- Remove the Joystream (`dot--joystream`) network from the preset network list.

## Intent & Context
QA reported a network that had been delisted long ago still appearing in the wallet's network selector, only without an icon. Tracing the "available networks" data source: `ServiceNetwork.getAllNetworks()` merges three sources — preset (static built-in), server networks (a simpleDb cache fed by `GET /wallet/v1/network/list?onlyEvmNetworks=true`), and user custom networks.

Inspecting the live API response confirmed the backend **delists a network by setting `status: "TRASH"` while still returning the entry** (e.g. Gnosis `evm--100`, Polygon zkEVM `evm--1101`). The client ignored `status` on the read path, so the delisted entry kept surfacing. Separately, the team asked to delist Joystream on the client.

## Root Cause
- `fetchNetworkFromServer` filtered out `TRASH` **on write**, so once a network was cached while `LISTED` and later flipped to `TRASH`, a stale `LISTED` snapshot lived in the serverNetwork simpleDb until a future fetch happened to reconcile it.
- `getAllNetworks` merged all three sources **without any `status` filter**, so that stale entry surfaced in every network selector. The missing icon is because the CDN logo asset was removed when the network was delisted (the cached `logoURI` now 404s).

## Design Decisions
- **Status-aware end to end (defense-in-depth):** store server networks *with* their real status (including `TRASH`) instead of dropping `TRASH` on write, and filter `status === TRASH` at the `getAllNetworks` merge layer so no source can leak a delisted network.
- **Refresh in place, but only on change:** after a server fetch, reuse the existing `AddedCustomNetwork` event (the de-facto "network list changed, refresh" signal all selectors already listen to) so open selectors update without a reopen — gated on an `id:status` signature diff so an unchanged hourly fetch does not fan out needless refreshes. A dedicated event was considered but would require wiring 6 existing listeners; reused the existing signal and documented it instead.
- **No force-refresh on selector open:** considered bypassing the 1h fetch throttle when the selector opens, but the selector is opened frequently and per-open fetches aren't worth it; the change-gated emit plus the hourly reconcile is sufficient.
- **Joystream removal scope:** removed only from the preset definitions (the `const`, `presetNetworksMap`, and `getPresetNetworks`). The DOT vault impl (`VaultDotSubJoyStream`), DOT vault settings, and the v4->v5 migration reference were intentionally kept — they are plain `'dot--joystream'` string references that don't affect the build, keep existing data resolvable, and allow easy re-listing later.

## Changes Detail
- `packages/kit-bg/src/services/ServiceCustomRpc.ts` — store all non-preset server networks with their status; emit the refresh event only when the server-network set changed.
- `packages/kit-bg/src/services/ServiceNetwork/ServiceNetwork.ts` — filter `status === TRASH` in the `getAllNetworks` merge.
- `packages/shared/src/config/presetNetworks.ts` — remove the Joystream network entry.
- `packages/shared/src/eventBus/appEventBus.ts` — document `AddedCustomNetwork` as the shared "network list changed" refresh signal.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Extension / Mobile / Desktop / Web (kit-bg + shared, all platforms)
- **Risk Areas**: `getAllNetworks` is consumed broadly. A `TRASH`/removed network becomes unresolvable via `getNetwork()` — this matches the delisting intent and the prior remove-by-absence behavior. A Codex cross-review found no consumer that depends on seeing `TRASH` networks and no dangling reference left by the Joystream removal. `tsc` and type-aware lint pass on all changed files.

## Test plan
- [ ] With a serverNetwork entry flipped to `TRASH` (e.g. Gnosis `evm--100`), the network no longer appears in the selector and `getAllNetworks` excludes it.
- [ ] After a server fetch that changes the set, an already-open network selector refreshes in place; no repeated refreshes when the set is unchanged.
- [ ] Joystream no longer appears in the network list / search / pinned; `getNetworkIdsMap()` has no `joy` key.
- [ ] **Regression:** a wallet that previously held a Joystream (or now-`TRASH`) account/balance — account list, assets, and history pages do not crash or error (graceful hide).
- [ ] User-added custom networks (`status: LISTED`) still display normally.

[OK-56327]: https://onekeyhq.atlassian.net/browse/OK-56327?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c9828e9c-fe76-4e98-9d88-48311cd0c6da?refs=603b14d392907694bc462bb55ac144f9733ba81a&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c9828e9c-fe76-4e98-9d88-48311cd0c6da/license?refs=603b14d392907694bc462bb55ac144f9733ba81a&source=prChecksComment) |

packages/kit/src/views/AssetSelector/pages/AggregateTokenSelector.tsx
```diff
@@ -403,6 +403,13 @@ function AggregateTokenSelector() {
       (token) => token.networkId,
     );
 
+    // Drop tokens whose network is no longer listed (delisted/removed). Their
+    // metadata is missing from listedNetworkMap, so they would otherwise render
+    // as blank rows with a broken icon and empty name.
+    result = result.filter((token) =>
+      Boolean(listedNetworkMap[token.networkId ?? '']),
+    );
```
sidmorizon: **`listedNetworkMap` only contains preset networks, incorrectly filtering out tokens on server-fetched networks**

Severity: `severe`

The new filter uses `listedNetworkMap` (line 56, initialized from `getListedNetworkMap()`) to drop tokens whose network is "no longer listed". However, `getListedNetworkMap()` (`packages/shared/src/config/networkIds.ts:104-110`) only includes **hardcoded preset networks** with `status === LISTED` — it does NOT include server-fetched networks (the additional EVM networks fetched via `ServiceCustomRpc.fetchNetworkFromServer` and stored in simpleDb). As a result, any token whose network comes from the server-fetched set (non-preset) will be silently dropped from the aggregate token selector, even though the network is perfectly valid and LISTED. The intent was to filter out TRASH/delisted networks, but `listedNetworkMap` is far too restrictive. The fix should check against the full dynamic network list (e.g. via `getAllNetworks`) or a map that includes server-fetched networks.

packages/kit/src/views/Market/components/MarketDetailPools.tsx
```diff
@@ -179,44 +180,80 @@ export function MarketDetailPools({
...
+      const network = i.onekeyNetworkId
+        ? existingNetworkMap.get(i.onekeyNetworkId)
+        : undefined;
+      return { logoURI: network?.logoURI };
+    });
+    if (tickers?.length) {
+      symbols.push({ networkName: CEX });
+    }
+    return symbols;
+  }, [validPools, existingNetworkMap, tickers?.length]);
```
sidmorizon, line 258: **MarketDetailPools: networkName is undefined for non-CEX pool filter tabs (pre-existing)**

Type: `Flag`

The new `oneKeyNetworkSymbols` memo at lines 236-250 returns `{ logoURI: network?.logoURI }` for non-CEX entries, omitting `networkName`. The `NetworkIdSelect` component passes `oneKeyNetworkSymbols[index]?.networkName` to `NetworksFilterItem`, which conditionally renders the name text only when truthy (`packages/kit/src/components/NetworksFilterItem/index.tsx:82-90`). However, this is NOT a regression — the old code also had this issue. It returned a full `IServerNetwork` object (which has `name`, not `networkName`), so `networkName` was always `undefined` at runtime for non-CEX entries. Both old and new code result in pool filter tabs showing only icons without text labels for DEX networks.

packages/kit/src/views/AssetSelector/pages/AggregateTokenSelector.tsx
weatherstar: Fixed in 18d4111f44. Replaced the preset-only `getListedNetworkMap()` filter with a network map resolved via `serviceNetwork.getNetworksByIds` (backed by `getAllNetworks`, which already excludes delisted/TRASH networks and includes server-fetched ones). The map merges preset LISTED networks (synchronous, so the list does not flash empty during async resolution) with the resolved server-fetched networks, so tokens on valid server-fetched networks are no longer dropped. The resolved network is also passed down to `AggregateTokenListItem` (icon/name) and used by the search filter, both of which previously also relied on the preset-only map.

packages/kit/src/views/Market/components/MarketDetailPools.tsx
```diff
@@ -179,44 +180,88 @@ export function MarketDetailPools({
...
+  const validPools = useMemo(
+    () =>
+      isExistingNetworksLoading
+        ? pools
+        : pools.filter(
+            (i) =>
+              i.onekeyNetworkId && existingNetworkMap.has(i.onekeyNetworkId),
+          ),
+    [pools, isExistingNetworksLoading, existingNetworkMap],
+  );
```
sidmorizon, line 232: **Stale tab index after async pool filtering can show wrong network data**

Severity: `non-severe`

In `MarketDetailPools`, the `index` state (line 259) is never reset when `validPools` shrinks after the `existingNetworks` async resolution. The PR introduces a two-phase transition: initially `validPools = pools` (all pools shown during loading), then `validPools = pools.filter(...)` once `existingNetworkMap` is populated. If the user selects a tab (e.g. index 2) during the brief loading window, and that pool gets filtered out, `validPools[index]` becomes either `undefined` (causing `isCEXSelected = true` at line 260, unexpectedly showing CEX data) or points to a different pool that shifted into that position. The `handleChange` callback (line 291) doesn't clamp or reset the selection when `validPools` changes.

packages/kit/src/views/Market/components/MarketDetailPools.tsx
weatherstar: Acknowledged — this is correct analysis. As noted, the missing `networkName` for non-CEX (DEX) pool filter tabs is pre-existing behavior, not a regression introduced by this PR: both the old code (returning a full `IServerNetwork`, which has `name` not `networkName`) and the new code (returning `{ logoURI }`) leave `networkName` undefined at runtime, so DEX tabs have always rendered icon-only. This PR is scoped to hiding delisted/TRASH networks, so I am leaving the label-rendering behavior unchanged here and tracking it separately rather than expanding scope. Resolving this thread.

packages/kit/src/views/Market/components/MarketDetailPools.tsx
weatherstar: Fixed in 8a5d407f94. The selected pool tab is now tracked by a stable key (pool `localId`, or the `CEX` sentinel) rather than a positional index. The active index is derived from that key via `tabKeys.indexOf(selectedKey)`, so when `validPools` shrinks after `existingNetworks` resolves, the selection stays pinned to the same tab instead of falling out of range (silently switching to CEX) or pointing at a shifted pool. If the previously selected pool was actually delisted, it falls back to the first tab.

originalix (approved):

PR #12091, opened 2026-06-17
fix: make js-conflux-sdk jsbi shim self-contained for SES lockdown(OK-56398 OK-56423)

weatherstar:
[OK-56398](https://onekeyhq.atlassian.net/browse/OK-56398) [OK-56423](https://onekeyhq.atlassian.net/browse/OK-56423)

----------

## Summary
- Patch `js-conflux-sdk`'s `src/util/jsbi.js` so the JSBI shim is self-contained and never mutates the native `BigInt` intrinsic.
- Fixes CFX address creation crashing under the new SES harden runtime.
- Add a regression test that runs under a real `lockdown()`.

## Intent & Context
Reported as two symptoms that turned out to be the same bug:
- Local `yarn app:web`: creating a CFX address throws `TypeError: JSBI.BigInt is not a function` (at `js-conflux-sdk/src/CONST.js`).
- A tester on a packaged/minified build hit `Cannot access 'Vault' before initialization`.

Both block CFX address creation.

## Root Cause
The recently added SES harden runtime (`packages/shared/src/security/sesHarden`, default level `L2`) calls `lockdown()` at startup in every JS runtime. `lockdown()`:
1. Freezes the `BigInt` intrinsic, and
2. **Removes non-standard own properties from intrinsics** ("unpermitted intrinsics").

`js-conflux-sdk`'s jsbi shim works by attaching its operations directly onto native `BigInt`:
```js
module.exports = BigInt;
module.exports.BigInt = BigInt;
module.exports.leftShift = (x, y) => x << y;
// ...
```
After lockdown those additions are stripped/blocked, so `JSBI.BigInt` / `JSBI.leftShift` become `undefined`. `CONST.js` then throws while evaluating `JSBI.leftShift(JSBI.BigInt(1), ...)`.

- In dev (unminified) this surfaces directly as `JSBI.BigInt is not a function`.
- In minified builds the failed module evaluation cascades: `cfx/Vault.ts` reads `sdkCfx` at module top-level (`const { Conflux, address } = sdkCfx`), so the throw leaves the `Vault` class binding in its TDZ → `Cannot access 'Vault' before initialization`.

It broke for everyone at once because SES harden shipped with default `L2` (all runtimes).

## Design Decisions
- **Rejected: warm-up before lockdown.** The existing `defaultWarmUpBeforeLockdown()` pattern (decimal.js, axios) only works for the "override mistake". It does NOT work here: a unit test under a real `lockdown()` proved SES *removes* the non-standard `BigInt.*` properties regardless of load order.
- **Rejected: webpack alias to the real `jsbi` package.** Its `module` field is ESM; a CJS `require` yields a `{ default }` namespace (re-introducing the `JSBI.BigInt` undefined bug), and it would not cover the Metro/native bundler.
- **Chosen: patch the shim to be self-contained.** A plain module-owned object is never an intrinsic, so SES has nothing to freeze/strip. Behavior is otherwise identical to upstream, and the fix is universal across all bundlers/platforms via patch-package.

Verified `js-conflux-sdk` internals only use `JSBI.<method>` (no direct `JSBI()` calls), and `JSBI.prototype.toJSON` (CONST.js) still lands harmlessly on the wrapper's own prototype.

## Changes Detail
- `patches/js-conflux-sdk+2.3.0.patch`: rewrite `src/util/jsbi.js` as a self-contained `JSBI` function object (`JSBI.BigInt = BigInt`, all operations as own properties), `module.exports = JSBI`. Never touches the `BigInt` intrinsic.
- `packages/shared/src/security/sesHarden/sesHardenLibCompat.test.ts`: regression test asserting the shim works after `lockdown()` AND that `BigInt` is not polluted.
- `packages/shared/src/security/sesHarden/runtime.ts`: comment-only — document that intrinsic-mutating libraries can't be fixed by warm-up.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Web / Desktop / Extension / Mobile (the shim is loaded by every bundler; patch applies uniformly via patch-package).
- **Risk Areas**: third-party patch of `js-conflux-sdk`. Scope is limited to the JSBI shim; behavior is byte-for-byte equivalent on non-locked-down runtimes, and the new test locks in post-lockdown behavior.

## Test plan
- [x] `npx jest packages/shared/src/security/sesHarden` → 37/37 pass (incl. new regression test under real `lockdown()`)
- [x] `yarn lint:staged` → 0 warnings / 0 errors
- [x] `yarn tsc:staged` → exit 0
- [x] `npx patch-package` applies cleanly (`js-conflux-sdk@2.3.0 ✔`)
- [x] `yarn app:web` → create a CFX address succeeds (no `JSBI.BigInt`/`Vault` error)
- [x] Verify CFX address creation on a packaged/minified build

[OK-56398]: https://onekeyhq.atlassian.net/browse/OK-56398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-56423]: https://onekeyhq.atlassian.net/browse/OK-56423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2f309e3d-89a0-4297-86cd-7fec319ba1d3?refs=21a458c70aac42b278a6e01c6bc7549cec4d048d&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2f309e3d-89a0-4297-86cd-7fec319ba1d3/license?refs=21a458c70aac42b278a6e01c6bc7549cec4d048d&source=prChecksComment) |

packages/shared/src/security/sesHarden/sesHardenLibCompat.test.ts
```diff
@@ -199,6 +199,40 @@ try {
...
+// --- js-conflux-sdk: the jsbi shim must be self-contained (patch-package) ----
+// Unlike axios/decimal.js (the "override mistake", fixed by 'severe'), the CFX
+// jsbi shim originally added its operations (BigInt, leftShift, ...) as OWN
+// properties directly onto the native BigInt intrinsic, plus
+// BigInt.prototype.toJSON. lockdown() REMOVES those as "unpermitted
+// intrinsics", so warm-up cannot save it (load order is irrelevant) and
+// 'severe' does not apply. patches/js-conflux-sdk+2.3.0.patch makes the shim
+// return a self-contained module object that never touches the BigInt
+// intrinsic, so it survives lockdown. Before the patch, requiring CONST.js
+// after lockdown threw "JSBI.BigInt is not a function".
```
sidmorizon: **Test comment mentions BigInt.prototype.toJSON but patch does not address it**

Type: `Flag`

The test comment at line 205-206 says the original shim added properties 'plus BigInt.prototype.toJSON'. However, the original jsbi.js (visible in the patch's removed lines) never set `BigInt.prototype.toJSON` — that property must be set by a different file within js-conflux-sdk. If that other file still modifies `BigInt.prototype.toJSON` after lockdown, SES will strip it. The test does not verify `toJSON` behavior. This could be a latent issue if any code path depends on `BigInt.prototype.toJSON` existing, but it's outside the scope of this patch (which only targets the jsbi shim). Worth confirming whether `BigInt.prototype.toJSON` is set elsewhere in the SDK and whether it matters for the CFX vault.

weatherstar: Confirmed and addressed.

**Provenance:** `BigInt.prototype.toJSON` was never set by `jsbi.js`. It came from `CONST.js:6` (`JSBI.prototype.toJSON = function () { return this.toString() }`). Before the patch `JSBI === BigInt` (the shim did `module.exports = BigInt`), so `JSBI.prototype` *was* `BigInt.prototype` — which is how `toJSON` ended up on the native intrinsic.

**Does it matter for the CFX vault under lockdown?** No, and the patch does not regress it. Production always runs SES `lockdown()`. Pre-patch under lockdown, `BigInt.prototype.toJSON` was already absent — CONST.js either ran after the freeze (assignment throws on the frozen `BigInt.prototype`) or before it (stripped as an unpermitted intrinsic). So no production path could rely on it. Post-patch, `JSBI` is a self-contained function object, so `CONST.js`'s `JSBI.prototype.toJSON` lands on the shim's OWN (unfrozen) prototype and never touches the native intrinsic.

**Added test coverage** for exactly this (commit e5b028737a). The lockdown test now asserts both directions:
- `typeof JSBI.prototype.toJSON === 'function'` — CONST.js's assignment succeeds post-lockdown (it lands on the shim's unfrozen prototype)
- `typeof BigInt.prototype.toJSON === 'undefined'` — the native BigInt intrinsic keeps no `toJSON` either

huhuanming (approved):

limichange (approved):

PR #11988, opened 2026-06-10
feat: add BTC find-address recovery flow(OK-55653)

weatherstar:
[OK-55653](https://onekeyhq.atlassian.net/browse/OK-55653)

----------

## Summary
- Add a BTC find-address flow for HD, hardware, and QR accounts so users can manually claim off-gap receive addresses by index.
- Persist claimed addresses locally, show them in the BTC receive address list, and expose their UTXOs in coin control only as explicit opt-in selections.
- Fix BTC signing path resolution for non-first-address inputs and add privacy-safe analytics/tests around claimed-address behavior.

## Intent & Context
This change supports `OK-55653`: users may have funds on BTC receive indexes outside the normal gap-scan window and need a controlled way to derive, verify, label, and spend from those addresses.

The feature is intentionally local-first. Claimed find-address records are stored on the local UTXO account only, are excluded from cloud sync payloads, and are not folded into normal send selection. Users must explicitly select claimed UTXOs in coin control before they can be spent.

## Root Cause
The existing BTC fresh-address and coin-control flows only work with addresses already visible through the account's normal discovered address set. Off-gap receive addresses were not represented in local account data, could not be verified through the receive UI, and could not be safely mapped back to derivation paths during hardware PSBT signing.

## Design Decisions
- Store claimed addresses in `IDBUtxoAccount.findAddresses` as a local-only `{ relPath: address }` dictionary and bump `LOCAL_DB_VERSION` to keep Realm schema changes explicit.
- Keep claimed UTXOs out of normal sends, send-max flows, and coin-control select-all operations; they enter a transaction only through an individual coin-control selection.
- Query claimed address details by address only, avoiding xpub-attached requests for the recovered address lookup.
- Reuse existing hardware/QR address verification paths so arbitrary claimed receive paths can be confirmed on-device or through QR verification before use.
- Keep analytics privacy-safe by logging actions and counts only, never claimed addresses, indexes, outpoints, or derivation paths.

## Changes Detail
- `packages/kit-bg/src/dbs/local/*`: add local `findAddresses` account schema support and DB version bump.
- `packages/kit-bg/src/services/ServiceFreshAddress.ts`: add BTC find-address derive, claim, unclaim, discovery dedupe, balance/detail fetch, and cache invalidation methods.
- `packages/kit-bg/src/vaults/impls/btc/*`: merge claimed UTXOs only for explicit selections and resolve per-input derivation path/pubkey for software, QR, and hardware signing.
[...]

PR #12095, opened 2026-06-17
refactor: convert BTC find-address to a modal page and finalize i18n (OK-55653)

franco-chan:
[OK-55653](https://onekeyhq.atlassian.net/browse/OK-55653)

----------

## Summary
Follow-up to #11988 (BTC find-address recovery). Reworks the find-address presentation into a modal page, finalizes localization, and polishes the address-list / Coin Control UI.

## What changed
- **Dialog → modal page.** "Find address" is now a pushed screen that slides in from the address list and pops back after a successful claim. The header back button replaces the dialog's Cancel; the full-path preview sits in the footer above the **Find** button.
- **i18n finalized.** Temporary inline copy (`findAddressCopy.*`) is replaced with `ETranslations.*` and `btcFindAddressCopy.ts` is deleted. Adds 13 `find_address_*` keys (reusing the existing `global_account` / `address_type`); Coin Control reuses the recovered-section label.
- **Address-list UI.** Monospace addresses with an inline copy affordance, heading-style section titles (Used / Next / Manually recovered), the address-type selector moved beside the segment control as a small tertiary button, and the find-address entry as a row-level more-menu.
- **Bug fix.** The "Manually recovered" section now appears after a claim — it revalidates when the list regains focus once the modal pops. The focus-gated app-event listener alone stopped covering this when the flow became a separate screen (previously it was an in-place dialog).
- **Coin Control.** Claimed-address badge restyled (default tone, placed after the address, wrapping below the date on narrow widths).
- **Cleanup.** Shared `BtcAddressText` cell; memoized the per-row claimed label.

## i18n note for reviewers
The locale diff is intentionally scoped to **only the 13 `find_address_*` keys** across all 20 generated files. A full `yarn i18n:pull` currently also drags in unrelated in-flight keys (e.g. `address_risk_check`) for every locale **and** re-surfaces a pre-existing duplicate enum identifier (`referral.web_landing_subtitle_perps`, assigned to two Lokalise platforms) that breaks `tsc`. Both were excluded here; the `referral` collision should be deduped on Lokalise independently.

## Testing
- `tsc:only`, ESLint, and oxlint (`--type-aware --deny-warnings`) pass on the changed files.
- Manual: find an off-gap index → returns to the address list with the new "Manually recovered" entry; funds spendable only via explicit Coin Control selection.

Relates to OK-54992 (product spec).

[OK-55653]: https://onekeyhq.atlassian.net/browse/OK-55653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/76c8325b-1079-4423-aef6-17f5aa43b42c?refs=a2482b545e4916a611a2c10aa6db789ccebe5b55&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/76c8325b-1079-4423-aef6-17f5aa43b42c/license?refs=a2482b545e4916a611a2c10aa6db789ccebe5b55&source=prChecksComment) |

packages/kit/src/views/Receive/pages/BtcFindAddress.tsx
```diff
@@ -0,0 +1,243 @@
...
+            id: ETranslations.find_address_already_exists__msg,
+          }),
+        });
+      } else {
+        defaultLogger.transaction.findAddress.findAddressClaimed({
+          networkId,
+          deriveType,
+        });
+      }
+      navigation.pop();
```
sidmorizon, line 133: **Missing success toast after successfully claiming a new find-address**

Severity: `non-severe`

When a new BTC find-address is successfully claimed (i.e., `alreadyDiscovered` is `false`), the old dialog code showed a `Toast.success` confirming the addition. The new page-based flow at `packages/kit/src/views/Receive/pages/BtcFindAddress.tsx:127-133` only logs to `defaultLogger` and immediately calls `navigation.pop()` without any user-visible success feedback. In contrast, the `alreadyDiscovered === true` path at line 121-126 still shows a `Toast.message`. This creates an asymmetry: when the user enters an index and presses Find, a successful claim silently navigates back with no confirmation, while the "already exists" case properly notifies the user. On slow networks the user may not be sure whether the operation succeeded.

Suggested fix:

```ts
      } else {
        defaultLogger.transaction.findAddress.findAddressClaimed({
          networkId,
          deriveType,
        });
        Toast.success({
          title: intl.formatMessage({
            id: ETranslations.find_address_confirm__action,
          }),
        });
      }
```

packages/kit/src/views/Receive/pages/BtcFindAddress.tsx
```diff
@@ -0,0 +1,243 @@
...
+    } catch (error) {
+      console.error(error);
+      throw error;
```
sidmorizon: **Re-thrown error in onConfirm creates unhandled promise rejection**

Type: `Flag`

In `BtcFindAddress.tsx:134-136`, the catch block does `console.error(error); throw error;`. The caller is `Page.FooterActions` whose `FooterConfirmButton` at `packages/components/src/layouts/Page/PageFooterActions.tsx:136-138` calls `onConfirm?.(pop, popStack)` without awaiting the returned promise. If `claimBtcFindAddress` throws, the re-thrown error becomes an unhandled promise rejection. This is a pre-existing pattern (the old dialog code had the same `throw error`), and in practice the background API typically shows its own error toast. However, the `throw` is unnecessary if the intent is just to show the error — it could be removed or replaced with a `Toast.error`.

packages/kit/src/views/Receive/pages/BtcFindAddress.tsx
sidmorizon, line 133: **Success toast removed for newly claimed addresses**

Type: `Flag`

The old dialog-based `FindAddressDialogContent` showed a `Toast.success` with "Added · #N" when a new address was successfully claimed (non-duplicate). The new `BtcFindAddress.tsx:127-132` only logs the event via `defaultLogger` and then calls `navigation.pop()` without any user-facing success feedback. The `alreadyDiscovered` case still shows a `Toast.message`, so only the success path lost its toast. The user will see the new address appear in the `BtcFindAddressSection` list (which has `revalidateOnFocus: true` at `BtcFindAddress.tsx:262`), so the visual appearance of the address serves as implicit confirmation. This is likely an intentional design choice but worth confirming with the team since it changes the user feedback model.

packages/kit/src/views/Receive/pages/BtcFindAddress.tsx
franco-chan: Fixed in `4c6b70e32f`. Dropped the re-throw: since `Page.FooterActions` does not await `onConfirm`, re-throwing turned a claim failure into an unhandled promise rejection. `claimBtcFindAddress` surfaces its own error toast, so the catch now just logs and keeps the user on the page (input intact) to retry.

packages/kit/src/views/Receive/pages/BtcFindAddress.tsx
franco-chan: Intentional, not a regression. The "Added · #N" success toast was deliberately removed during design review as redundant. On a successful claim the page pops back to the address list, where the new address appears immediately in the "Manually recovered" section (it refetches via `revalidateOnFocus`) — that is the confirmation. The `alreadyDiscovered` branch keeps its toast because nothing new appears in the list in that case, so the explicit message is needed there. Keeping as-is.

franco-chan: Confirming this is intentional. Per the design review the "Added" toast was removed as redundant; the new address showing up in the "Manually recovered" section on return is the success confirmation. The `alreadyDiscovered` path keeps its toast because nothing new appears in the list there. Not changing.

packages/shared/src/locale/enum/translations.ts
```diff
@@ -1396,6 +1396,19 @@ export enum ETranslations {
...
+  find_address_empty_index__msg = 'find_address_empty_index__msg',
+  find_address_full_path__title = 'find_address_full_path__title',
+  find_address_index__title = 'find_address_index__title',
+  find_address_invalid_index__msg = 'find_address_invalid_index__msg',
+  find_address_recovered_section__desc = 'find_address_recovered_section__desc',
+  find_address_recovered_section__title = 'find_address_recovered_section__title',
+  find_address_remove__desc = 'find_address_remove__desc',
+  find_address_remove__title = 'find_address_remove__title',
+  find_address_warning__desc = 'find_address_warning__desc',
+  find_address_warning__title = 'find_address_warning__title',
```
sidmorizon, line 1454: **Locale/translations files modified despite CLAUDE.md 'NEVER modify' rule**

Type: `Flag`

CLAUDE.md line 65 says: "NEVER modify auto-generated files (translations.ts, locale JSON files)". This PR modifies `packages/shared/src/locale/enum/translations.ts` and all 18 locale JSON files. However, the PR's explicit purpose is to migrate hardcoded strings (from the deleted `btcFindAddressCopy.ts`) to proper i18n keys. The `.cursor/rules/locale-handling.mdc` also says "Always use translation keys instead of hardcoding text strings" and "Define new translation keys in the appropriate modules" — which is impossible without updating these files. This appears to be the output of the i18n workflow (all 18 languages have proper translations), not ad-hoc hand-editing that the rule targets. The rule is likely intended to prevent casual manual edits rather than blocking the i18n pipeline itself.

weatherstar (dismissed):

packages/shared/src/locale/enum/translations.ts
franco-chan: fixed in https://github.com/OneKeyHQ/app-monorepo/pull/12095/commits/f0e8823f21a8631991f182664f9cb13f5aa95a96

weatherstar (approved):

originalix (approved):

PR #12121, opened 2026-06-20
feat: OCDS v1.1 conformance — desktop concurrent download + shared update orchestration (OK-56573)

huhuanming:
[OK-56573](https://onekeyhq.atlassian.net/browse/OK-56573)

----------

Brings the desktop (Node) concurrent downloader and the shared AppUpdate orchestration into conformance with the **OneKey Concurrent Download Standard (OCDS v1.1)** (defined in the app-modules repo: `native-modules/react-native-range-downloader/SPEC.md`).

## Desktop concurrent downloader
- Validate each segment 206 `Content-Range` vs the requested window + clamp over-long bodies (no cross-segment overrun); reject `multipart/byteranges`
- Single-stream 200-on-resume restarts cleanly (no append corruption)
- Exhaustive HTTP-status classification (416 transient, 401/501/505 permanent, §4 catch-all default)
- Bytes-stalled timeout + jittered backoff + `Retry-After`
- Per-destination single-flight

## Shared AppUpdate orchestration
- Exhaustive error taxonomy + §4 catch-all default rule
- Persisted cross-restart attempt budget + wall-clock deadline + terminal give-up contract — **opt-in APIs; wiring into the production caller `useDownloadPackage` is a deliberate follow-up** (behavior change)

## Validation
- Tests import & exercise the **real exported** classifiers/validators (not mirror copies)
- `tsc` clean; AppUpdate suites pass
- All four platform classifiers verified consistent against the spec (416 transient everywhere)

## Follow-ups (not in this PR)
- Wire the persisted give-up budget into the production download caller (its own PR + device validation)

Pairs with the app-modules PR (iOS/Android native + the OCDS standard).

> **Standard + conformance harness** live in app-modules [#70](https://github.com/OneKeyHQ/app-modules/pull/70).

---

## ✅ Verification record

How every platform's downloader is verified against the standard, and how each
test was proven **load-bearing** (a green test that survives its own regression
proves nothing). Full methodology + runnable harness:
[`react-native-range-downloader/conformance/`](https://github.com/OneKeyHQ/app-modules/tree/feat/ocds-download-conformance/native-modules/react-native-range-downloader/conformance).

### Approach

```mermaid
flowchart TD
    SPEC["OCDS SPEC v1.2<br/>§6 — 11 conformance scenarios"]
    SPEC --> AND["Android<br/>pure-JVM + OkHttp MockWebServer<br/>real fault injection"]
    SPEC --> NODE["Node / Desktop<br/>jest + real local HTTP/HTTPS server"]
    SPEC --> IOSU["iOS · unit<br/>SwiftPM over the real RangeDownloadLogic"]
    SPEC --> IOSE["iOS · end-to-end<br/>real Release .app on Simulator<br/>+ local fault server (silent auto-download)"]
    AND  --> MUT["Mutation verification<br/>inject the regression each test targets<br/>→ the test MUST go red"]
    NODE --> MUT
    IOSE --> EVID["evidence: server log + app-latest.log<br/>+ on-disk .segN + final SHA"]
    EVID --> ADV["adversarial verify<br/>(default FAIL unless 3 sources agree)"]
    MUT  --> OK(["load-bearing ✔"])
    ADV  --> OK
```

### SPEC §6 coverage matrix

| # | Scenario | iOS (sim) | Android | Node |
|---|---|:--:|:--:|:--:|
| 1 | transient error → backoff retry, no restart | ✅ | ✅ | ✅ |
| 2 | killed mid-run → resume from completed segments | ✅ | ✅¹ | ✅¹ |
| 3 | `200` to a Range → permanent → single-stream | ✅ | ✅ | ✅ |
| 4 | `416` / `5xx` transient (re-evaluate, retry) | ✅ | ✅ | ✅ |
| 5 | short / over-long / mis-aligned / bad-total `206` | ✅ | ✅ | ✅ |
| 6 | corrupted assembly → terminal (v1.2: no retry-once) | ✅ | ✅ | ✅ |
| 7 | flap / background → progress never resets to 0 | ✅ | ◑² | ✅ |
| 8 | user cancel mid-run | ⚪³ | ✅ | ✅ |
| 9 | persistent failure → bounded give-up | ✅ | ✅ | caller⁴ |
| 10 | stalled socket → stall watchdog | ✅ | ✅ | ✅ |
| 11 | two concurrent `download()` → single-flight | ⚪³ | ✅ | ✅ |

¹ resume tested via a persisted-manifest proxy — a true SIGKILL-mid-write isn't reproducible in-process; covered by the manifest-resume + intra-call-resume tests.
² Android progress: the clamp/guard math is unit-tested, but the monotonic CAS primitive stays inline in the JNI adapter (not unit-testable).
³ iOS #8/#11 need a temp shim (the iOS app has **no** cancel trigger at all — #8 would test dead code); both behaviors are fully covered on Android + Node.
⁴ Node #9 give-up budget lives in the shared kit caller (`useAppUpdate.test.ts`), not the downloader.

### Test counts & strength

| Platform | Tests | Mutation-proven | Driver |
|---|--:|--:|---|
| **Android** | 64 | 9 | OkHttp MockWebServer (real sockets, real `.segN`, real SHA) |
| **Node / Desktop** | 18 | 3 | real local HTTP/HTTPS + positioned writes + `.progress.json` |
| **iOS · unit** | 30 | n/a | SwiftPM, source-symlinked to the shipping logic |
| **iOS · e2e** | 9/11 §6 | adversarial | real Release `.app` on a booted Simulator |

### Mutants planted & killed (load-bearing proof)

```mermaid
flowchart LR
    subgraph Android
      a1["503 → permanent"] --> ka(["🔴 caught"])
      a2["416 → permanent"] --> ka
      a3["bypass multipart"] --> ka
      a4["drop total check"] --> ka
      a5["501 → transient"] --> ka
      a6["retry budget +2"] --> ka
      a7["registry identity-drop"] --> ka
      a8["progress clamp removed"] --> ka
      a9["seg-glob narrowed"] --> ka
    end
    subgraph Node
      n1["416 → permanent"] --> kn(["🔴 caught"])
      n2["break single-stream 416"] --> kn
      n3["isConcurrentFallback→false"] --> kn
    end
```

Every mutant produced a real **test** failure (not a compile error) and was reverted.

### Honesty / scope
- **iOS e2e is a fake update task running through real download code.** The app's own native downloader, its silent-auto-download trigger, segmentation, resume, retry, fallback and SHA-256 are all **real**; the update server, manifest, version and the bundle payload are **synthetic** (filler bytes, not a signed zip). It verifies the **download layer only** — the post-download unzip → signature-verify → install chain is out of OCDS scope.
- **SPEC §6 #6** previously mandated a single-stream retry-once on a checksum mismatch. All three reference implementations terminate on mismatch instead (re-fetching the same object would re-corrupt), so **v1.2 removed it** — `#6` is now conformant everywhere.
- The behavior-preserving extractions (Android `RangeDownloadLogic.kt`, iOS `RangeDownloadLogic.swift`) are pure moves — proven by the pre-existing real-path suites staying green.

---

## 🔩 Android APK updater — audit + fixes

The Android APK download path (`react-native-app-update`) shares the OCDS core, but its ~1346-line caller was never audited and had **zero tests**. An adversarial audit found 4 issues.

```mermaid
flowchart LR
  Q["'is the APK update missing changes?'"] --> AUDIT["adversarial OCDS audit<br/>of react-native-app-update"]
  AUDIT --> B1["#1 cancel race (HIGH)"] --> FIX["✅ fixed"]
  AUDIT --> B2["#2a/#2b spurious update/error"] --> FIX
  AUDIT --> B34["#3/#4 single-flight · promote-before-verify"] --> BL["📋 backlog (documented, mitigated)"]
  AUDIT --> COV["AppUpdateLogic extraction<br/>17 JVM tests · 10 mutants killed"]
```

### Issues

| Bug | Severity | Status | What |
|---|---|---|---|
| #1 cancel-then-delete race (OCDS §5.8) | HIGH | ✅ fixed (`213acf0f3`) | concurrent download had NO `cancelHandle` → `clearCache` deleted `.segN` while 8 workers were still writing. Wired `CancelHandle` + cancel-before-delete (mirrors the bundle caller). |
| #2a spurious update/error on cancel | MED | ✅ fixed (`24d5973cb`) | suppress `update`/`error` when the run was intentionally cancelled. |
| #2b spurious update/error when ASC offline | MED | ✅ fixed (`011c24a58`) | a deferred (offline) signature check was mis-reported as a download `error` on every retry; suppressed (the Promise still rejects → JS still retries). |
| #3 process-global single-flight / #4 promote-before-verify | MED / LOW | 📋 backlog | documented; doesn't corrupt, mitigated. |

### Coverage (`b9caf71f9`)

Extracted the dependency-free pure logic into `AppUpdateLogic.kt` (segment names + count, 416/206 `Content-Range` parse, completeness/alignment predicates); the adapter now delegates (pure move). Added junit + **17 JVM unit tests**, with all **10 planted mutants killed** (mutation-proven). Also wired a previously-untested inline 206-parse duplicate through the helper.

### Honesty / scope

Still **device/Robolectric-only**: the JNI download loop, the actual `.segN` deletion + §5.8 cancel-race execution, and the promote/verify/install File-I/O are **not** exercised by these JVM tests.

[OK-56573]: https://onekeyhq.atlassian.net/browse/OK-56573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/8e1c3950-4679-4d8b-8648-f47f32c88688?refs=c17b2adee285b27d2c2adafb31344a97afc6204c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/8e1c3950-4679-4d8b-8648-f47f32c88688/license?refs=c17b2adee285b27d2c2adafb31344a97afc6204c&source=prChecksComment) |

huhuanming:
## OCDS conformance — adversarial review summary (desktop / shared JS)

Independent multi-agent audit against [OCDS SPEC v1.2], each high-severity finding re-checked by a separate skeptic. **No blockers** — the SHA256 verify is a real backstop on every path, so no corrupt bundle is installed or reported complete. Node e2e coverage confirmed real (18/18, mutation-proven: flipping 416→permanent and `isConcurrentFallback`→false both go red against the shipping code).

### Cleanup landed (history rewrite)
This branch's `chore: persist in-progress working-tree changes` commit bundled **10 non-OCDS files** (SegmentSlider, TokenSelector, Prime, Send, JPush, perpsUtils, etc.) into the PR. It has been dropped via rebase + force-push; the PR diff is now OCDS-only. (Unrelated work preserved locally.)

### Confirmed correct
- `classifyHttpStatus` matches SPEC §4 exactly (408/416/429→transient, 501/505→permanent) and is consistent with the Android classifier.
- §5.11 cross-restart give-up budget persists and does **not** suppress a newer-version update notification (verified).
- The `use-after-close` fd-write race that a finder flagged was **refuted** — Node's single-threaded event loop + the `isAborted()` guard running first on every chunk make it unreachable.

### Backlog — real but accepted (not fixed here)
| Item | Why deferred |
|---|---|
| Concurrent `.partial` is full-size pre-allocated (`ftruncateSync`); single-stream resumes by **file size** and shares the same path → a crash-left sparse file can be size-resumed and renamed to final | SHA256 verify rejects the all-zero file (no bad install) and the run self-heals; needs the conjunction of crash-mid-concurrent + probe-fail. Real SPEC §3 violation, user-invisible. Fix: distinct `.cpartial` ext or make single-stream refuse to size-resume a manifest-backed partial. |
| Over-long `206` writes the clamped slice + advances progress before rejecting (§5.5) | Self-correcting (fallback discards); briefly muddies progress. |
| Single-stream 416-with-partial finalizes without a size check before rename | SHA catches; pairs with the pre-alloc item. |
| `?? 'UNKNOWN'` reason not in `UNRECOVERABLE_DOWNLOAD_ERROR_CODES` → would be retried | Happy path stamps `MISMATCH`; edge only. |
| `DesktopApiBundleUpdate.test.ts` re-implements validation/redirect/param logic locally instead of calling the shipped methods | Classifier/backoff/content-range tests *do* hit real exports; the validation tests are coverage theater. |

Full report available on request.

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
```diff
@@ -173,14 +364,57 @@ class DesktopApiAppBundleUpdate {
...
+      );
+      return existing;
+    }
+    const run = this.runDownloadBundle(params).finally(() => {
+      this.inflightDownloads.delete(destKey);
+      this.cancelByDest.delete(destKey);
+    });
+    this.inflightDownloads.set(destKey, run);
+    return run;
+  }
```
sidmorizon, line 408: **Single-flight dedup uses getDestZipPath which requires bundleVersion — app-shell updates may bypass**

Type: `Flag`

The single-flight mechanism in `downloadBundle` (`packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts:385-408`) computes the destination key via `getDestZipPath(params)` which requires both `latestVersion` AND `bundleVersion` (line 373). For app-shell updates where `bundleVersion` might be undefined/empty, `getDestZipPath` returns null and the single-flight is skipped (falling through to `runDownloadBundle` directly). This means app-shell downloads without a bundleVersion cannot benefit from single-flight dedup. This matches the comment at line 389-391 ('No usable key → skip single-flight bookkeeping'), so it appears intentional, but it means concurrent calls for app-shell updates could theoretically race.

huhuanming: Not a real race — this path is safe by construction, so no change needed.

`downloadBundle` is only ever invoked for **JS bundle** downloads. The sole production caller (`useDownloadPackage.tsx:321`) routes by `fileType`:
- `EUpdateFileType.jsBundle` → `BundleUpdate.downloadBundle()`
- everything else (app-shell) → `AppUpdate.downloadPackage()` (electron-updater, a different path entirely)

So app-shell updates never reach `downloadBundle`, and for JS bundles the bundle is *identified by* `bundleVersion`, so it is always present and `getDestZipPath` returns a key.

Even in the theoretical "missing `bundleVersion`" case there is no co-write race: `getDestZipPath` returns null → `runDownloadBundle` → both `downloadBundleConcurrent` (line ~1109) and `downloadBundleSingleStream` (line ~923) `reject('Invalid parameters')` **before any `.partial` file is opened**. Two concurrent calls would just fail validation identically; nothing is written, so there is nothing to co-write. Single-flight only needs to guard runs that actually touch the destination artifacts.

Resolving as intentional.

packages/kit-bg/src/services/ServiceAppUpdate.ts
```diff
@@ -835,6 +1009,58 @@ class ServiceAppUpdate extends ServiceBase {
...
+    downloadUrl?: string;
+    jsBundleDownloadUrl?: string;
+  }> {
+    const before = await appUpdatePersistAtom.get();
+    const beforeUrl = before.downloadUrl;
+    const beforeBundleUrl = before.jsBundle?.downloadUrl;
+    try {
+      // Re-fetch release info; this re-signs the download URL(s) server-side
+      // and writes them back into the persist atom.
+      await this.fetchAppUpdateInfo(false);
```
originalix: 当下载因为 401/403 进入 `refreshDownloadUrlForRetry()` 时，这里仍然走 `fetchAppUpdateInfo(false)`。`false` 会继续受 `isNeedSyncAppUpdateInfo()` 的缓存窗口约束，命中 `skip_sync` 时根本不会向服务端重新取签名 URL，`after.downloadUrl` / `after.jsBundle.downloadUrl` 只能还是旧值，外层重试随后就会把这次失败定格为 `urlDead`。

这个分支需要的是“强制刷新 dead URL”，不是“按常规节流检查更新”。建议这里改成明确 bypass 本地 sync gate 的路径，例如 `fetchAppUpdateInfo(true)` 或专门拆一个只刷新签名 URL 的后台接口。

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
```diff
@@ -962,9 +1315,33 @@ class DesktopApiAppBundleUpdate {
...
                     }
                   }
+                  // OCDS §5.6: only append when the server resumed at exactly
+                  // the offset we asked for. A mis-aligned 206 (start !=
+                  // downloadedBytes) would append at the wrong position, so
+                  // discard the partial and restart cleanly from byte 0.
+                  const resumeStart = contentRangeStart(
+                    response.headers['content-range'],
+                  );
+                  if (
```
originalix, line 1334: 当单流 resume 请求拿到 `206`，但 `Content-Range` 的起点和本地 `downloadedBytes` 不一致时，这里虽然删掉了旧 `.partial` 并把 `downloadedBytes` 置 0，却继续消费当前这个从错误 offset 开始的响应体。结果就是服务端返回的 `resumeStart..end` 字节会被按 `0..` 写入新文件，文件前缀永久缺失，最终只能在 SHA 校验时失败。

这个分支需要真正“重启请求”，而不是只切换成 `w` 模式继续写：发现 misaligned 206 后应立即丢弃当前 response / request，并重新从 byte 0 发起一次新请求后再落盘。

packages/kit/src/components/AppUpdate/useDownloadPackage.tsx
```diff
@@ -328,6 +367,15 @@ export const useDownloadPackage = () => {
...
+          // the single-stream fallback itself failed). Surfaced as a terminal
+          // failure; the persisted record keeps the next relaunch from
+          // re-downloading until a new target version supersedes it.
+          defaultLogger.app.appUpdate.log(
+            `downloadPackage: OCDS §5.11 terminal give-up reason=${e.reason}`,
+          );
+        }
         await backgroundApiProxy.serviceAppUpdate.downloadPackageFailed(
           e as Error,
         );
```
sidmorizon, line 381: **DownloadGaveUpError message surfaces as raw English in the error UI**

Type: `Flag`

When `DownloadGaveUpError` is thrown (`useDownloadPackage.tsx:370`), it's passed to `downloadPackageFailed(e as Error)` at line 379. The `downloadPackageFailed` method in `ServiceAppUpdate.ts:811` uses `e?.message` as the `errorText`. The `DownloadGaveUpError.message` is a technical English string like `"downloadPackage: download gave up (persisted maxAttempts)"` which doesn't match any of the known error patterns, so it's set as-is via `updateErrorText()` (cast to `ETranslations`). This means the user would see the raw technical string in the error UI rather than a localized message. This follows the pre-existing pattern where unrecognized error messages fall through to display as-is, so it's not a regression, but it's a UX quality gap for the new give-up flow.

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
huhuanming: Adversarial re-audit (two independent read-only passes over the branch source) confirms **no co-write race** — but with one correction to my earlier reply.

**Correction:** `useDownloadPackage.tsx` is *not* the sole production caller. `DevBundleSwitcher/BundleList.tsx:110` also calls `BundleUpdate.downloadBundle(...)` and is reachable in shipped builds through the IPC bridge (`desktopApiProxy.bundleUpdate.downloadBundle`). It always passes both `latestVersion` and `bundleVersion`, so it *engages* single-flight rather than bypassing it — the conclusion is unchanged, but "sole caller" was inaccurate.

**Why the null-key path is still race-free (the actual load-bearing invariant):** safety here does not come from single-flight — it comes from the download-path validation predicate being a strict **superset** of the dedup-key predicate:
- `getDestZipPath` returns null iff `!appVersion || !bundleVersion`.
- both download paths reject iff `!appVersion || !bundleVersion || !bundleUrl || !sha256` (single-stream additionally `!fileSize`).

So every param combo that disables single-flight (`key == null`) also fails validation — and that reject runs **before the first filesystem side-effect** in both paths:
- `downloadBundleConcurrent`: reject at L923 precedes `getDownloadDir()`/`mkdir` at L929 and any `.partial` open/rename/rm.
- `downloadBundleSingleStream`: reject at L1109 precedes `getDownloadDir()` at L1149 and every write.

The only pre-validation effects on a null-key call are an idempotent Electron taskbar progress-bar reset and one harmless `bytes=0-0` network probe — no fs mutation. Two concurrent null-key calls therefore cannot co-create the download dir, co-open/truncate a `.partial`, or co-delete artifacts. The single-flight check-then-set itself is also TOCTOU-free (no `await` between `inflightDownloads.get` and `.set`).

**Residual (non-blocking):** this safety rests on the `validation ⊇ key-null` superset, and there's no test asserting it. A future change that relaxes the download-path guard (or derives a dest path from `latestVersion` alone) without tightening `getDestZipPath` would open a real `key=null`-but-writes window. Worth a one-line invariant test later.

Resolving as intentional + verified.

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
```diff
@@ -962,9 +1315,33 @@ class DesktopApiAppBundleUpdate {
...
+                  ) {
+                    logger.warn(
+                      'bundle-download',
+                      `206 resume mis-aligned (server start ${resumeStart} != expected ${downloadedBytes}), restarting`,
+                    );
+                    if (fs.existsSync(partialFilePath)) {
+                      fs.rmSync(partialFilePath, { force: true });
+                    }
+                    downloadedBytes = 0;
+                  }
```
sidmorizon, line 1352: **Misaligned 206 detection in single-stream does not restart the HTTP request, producing wrong bytes in the partial file**

Severity: `non-severe`

When `downloadBundleSingleStream` detects a misaligned 206 (Content-Range start ≠ `downloadedBytes`), it deletes the old partial and resets `downloadedBytes = 0`, but continues consuming the **same** response body — which delivers bytes starting from the server's `resumeStart`, not byte 0. The `flags: 'w'` write stream then writes these wrong-offset bytes starting at file position 0. The resulting partial file starts from `resumeStart`, not from byte 0, so it is silently mis-positioned.

The downstream `downloadedBytes >= totalBytes` check at `DesktopApiBundleUpdate.ts:1398` catches it as "incomplete" (the 206 body is shorter than totalBytes), so the download errors out — but on the **next** retry, the stale misaligned partial still sits on disk. The next attempt resumes from the partial's size, appending more bytes onto already-mispositioned content, wasting a second retry until SHA verification catches the corruption. This wastes two retry attempts and two full downloads worth of bandwidth in this (admittedly rare) edge case. The fix should destroy the current request/response and issue a fresh non-Range request instead of continuing to use the misaligned 206 stream.

packages/kit/src/components/AppUpdate/useDownloadPackage.tsx
huhuanming: Confirmed reachable — thanks. Worth separating the two surfaces:

- **Toast** (`useDownloadPackage.tsx:382-387`) is already localized: it shows `ETranslations.global_update_failed` regardless of the underlying error, so the give-up never surfaces raw English *there*.
- **Persisted `errorText`** is the real gap you flagged. `downloadPackageFailed` (`ServiceAppUpdate.ts:811`) only maps a few known substrings and otherwise stores `e.message` verbatim, which `updateErrorText` then casts to `ETranslations`. `DownloadGaveUpError`'s message (`"...: download gave up (persisted maxAttempts)"`) matches no branch, so it's stored raw and renders as-is on any surface that reads the atom's `errorText`.

One correction worth noting: this give-up is **already reachable on this branch**, not just in the deferred budget follow-up — `useDownloadPackage` wires `getBudget`/`recordAttempt`/`resetBudget` into `runDownloadWithRetry` (lines 324-336), so the persisted-maxAttempts path can fire in production. So the gap is live, not latent.

Keeping the copy change out of this PR by scope, but tracking it as a follow-up: the clean fix is to map `DownloadGaveUpError` by its `reason` (not the raw message string) to localized copy in `downloadPackageFailed` — e.g. a generic network/give-up string for `urlDead` / `deadline`, and a dedicated one for `maxAttempts` — rather than letting it fall through to the verbatim `e.message`.

packages/kit-bg/src/services/ServiceAppUpdate.ts
huhuanming: 确认是真 bug,感谢 review。经源码核对 + 对抗验证:`refreshDownloadUrlForRetry` 调 `fetchAppUpdateInfo(false)`,在 sync 节流窗口内 `isNeedSyncAppUpdateInfo` 会短路返回缓存 atom(`reason: skip_sync`)、根本不回服务端,dead URL 不会被重签 → `after.downloadUrl === beforeUrl` → 外层把 401/403 这次重试定格成 `urlDead`。可达条件:部署的签名 URL TTL 短于节流窗口(1h / 扩展端 1 天),且下载因弱网/暂停/离线等待拖过 TTL。

已按你的建议改成 `fetchAppUpdateInfo(true)`(`22548b4fbb`)。验证过 `true` 会同时绕过 sync 门**和** `getAppLatestInfo` 自带的 5 分钟内存缓存、真正打到 `fetchConfig` 重签,所以是充分修复。另补了 `refreshDownloadUrlForRetry` 的回归测试(此前 `useAppUpdate.test.ts` 把它整个 mock 掉了,所以这个 bug 没被测到),并做了 mutation 验证:改回 `false` 测试即变红。

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
huhuanming: 确认是真 bug,感谢 review。错位 206 分支只删 partial + 重置 `downloadedBytes`,却继续消费同一个从 `resumeStart` 起的响应体 → 这些字节被写到新文件 offset 0,文件变成 `object[resumeStart..end]`,前缀 `[0..resumeStart)` 永久缺失。size+SHA 闸能挡住、不会装上坏包,但会白白耗重试,且残留的短 partial 在后续 resume 上会叠加成「满长度但 SHA 永挂」的文件。

已按你的建议改成真正重启请求(`998f1312df`):错位时 `response.destroy()` + `downloadRequest?.destroy()`、删 partial、`safeReject` 一个可重试错误(已确认它落在可重试分类、不在 unrecoverable 列表),由外层 `runDownloadWithRetry` 下一轮从 byte 0 重发(无 partial → 无 Range → 干净 200/206)。沿用了上面 416/HTTP-error 的 reject-and-return 模式;没有照搬 200 分支的 reset+continue(那个只对从 byte 0 开始的 200 body 成立)。

packages/kit-bg/src/desktopApis/DesktopApiBundleUpdate.ts
huhuanming: Already fixed in `998f1312df` (this finding is against the pre-fix `head_sha=f514018dc3`). The fix is exactly what's recommended here: on a mis-aligned 206 the single-stream path now destroys the current response + request, drops the partial, and rejects with a transient error so the outer `runDownloadWithRetry` re-issues a fresh request from byte 0 (no partial ⇒ no Range ⇒ clean 200/206) — instead of continuing to consume the misaligned stream. Duplicate of the originalix thread on the same branch; resolving.

originalix (approved):

PR #8710, opened 2025-10-20
fix: remove lazy loading for ProviderApiPrivate to fix native webview issues

sidmorizon:
## 
## Summary by CodeRabbit

* **Chores**
  * Updated spell checker configuration.

* **Refactor**
  * Modified background provider initialization handling.

- Removed lazy loading pattern for `ProviderApiPrivate` in background providers
- Changed from `Object.defineProperty` getter pattern to direct import and instantiation
- Added 'getters' to spell checker skip words

This fix resolves issues where `ProviderApiPrivate` was not being properly initialized in native webview contexts due to the lazy loading pattern.

## Test plan
- [ ] Verify native webview functionality works correctly on iOS
- [ ] Verify native webview functionality works correctly on Android  
- [ ] Test that ProviderApiPrivate is properly initialized and accessible
- [ ] Run existing provider-related tests
- [ ] Verify no regression in desktop/web/extension platforms

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #12130, opened 2026-06-22
perf(kit-bg): lazy-load dapp ProviderApis + trim startup bundle (jsrsasign/lottie/lodash)

huhuanming:
## Summary

Reduce the web/native **startup JS bundle** by lazy-loading code that isn't needed on first paint. Four independent cuts, each verified.

| # | Commit | Change |
|---|---|---|
| 1 | `perf(kit-bg)` | **Lazy-load per-chain dapp ProviderApis.** `backgroundProviders.ts` registered each chain via synchronous `require('./ProviderApiXxx')` in a lazy getter — but webpack (web) and Metro's split-bundle serializer (native) treat a static `require()` as a build-time dependency, so every chain's ProviderApi + heavy SDK (Aptos ~5MB, Tron, NEO, Alephium, Cosmos, Sui, …) was anchored into the initial/startup graph despite Home never using them. Converted to an `import()` loader registry + `BackgroundApiBase.getProviderApi(scope)` (mirrors the cross-platform-proven `vaults/factory.ts` pattern). `$private` stays eager (accessed synchronously by name); the one sync `ethereum` consumer now `await`s; broadcasts iterate only already-loaded providers. |
| 2 | `chore(core)` | **Remove dead jsrsasign RSA encryptor (~336KB).** `rsaGenerateKeypair/rsaEncrypt/rsaDecrypt` have zero callers repo-wide; the `export *` barrel kept the unused `jsrsasign` import live. Verified zero references after removal. |
| 3 | `perf(components)` | **Lazy-load lottie-web on web (~600KB).** `LottieView/index.web.tsx` → `React.lazy(() => import('lottie-react'))` + `<Suspense>`. Web only; native unchanged. |
| 4 | `perf(webpack)` | **Scope `vendor-lodash` cacheGroup to `chunks:'initial'`.** With `chunks:'all'` it merged lodash methods used only by async chunks into the named initial chunk (measured ~92% unused on Home). |

## Measured impact (production builds, before → after)

**Web initial JS:**

| Metric | Before | After | Δ |
|---|---|---|---|
| `<script defer>` count | 28 | 22 | −6 |
| uncompressed | 27.3 MB | 18.2 MB | **−9.1 MB** |
| gzip | 7.01 MB | 4.68 MB | **−2.3 MB (−33%)** |
| per-chain SDK on initial | 14.37 MB | 4.30 MB | −10.1 MB → async |
| `vendor-lodash` chunk | 821 KB | 225 KB | −596 KB |

**Native (iOS) `startup.background`** (commit 1 only): 40.77 MB → 27.50 MB (**−13.27 MB / −33%**), −1199 modules; allocation `violations: 0`.

`tsc --noEmit` clean (no new errors vs the pre-existing baseline).

## Runtime verification

Served the production build and loaded it in a real browser (Playwright):

- ✅ App **boots and renders** — Home/Market full UI (exercises `new BackgroundApi()` → `createBackgroundProviders` → `getProviderApi` wiring; no white screen / init crash).
- ✅ **0 chunk-load errors**, 0 page errors, 0 failed `.js` requests — the lazy provider/lottie/swap chunks load on demand correctly.
- (Console errors observed are `net::ERR_CONNECTION_CLOSED` from backend API calls on the local static serve — unrelated to this change.)

**Verification boundary / required QA:** the automated check covers boot + lazy-chunk loading. It does **not** exercise the live dapp provider **dispatch** (`getProviderApi` firing when a dapp invokes a chain method), which needs a real dapp + wallet. Please regression-test on a real device/extension:
- dapp connect across several chains (EVM, Solana, Aptos, Tron, …)
- `accountsChanged` / `chainChanged` reaching connected dapps
- the Hyperliquid perp `ethereum` broadcast path (`notifyHyperliquidPerpConfigChanged`)
- native webview dapp provider init (an earlier async-load attempt #8710 was sensitive here; this change keeps `$private` eager and uses the proven `import()` pattern)

## Follow-ups (intentionally NOT in this PR)

Audited and deferred because they ripple into synchronous-access chains and touch sensitive flows that can't be runtime-verified here:
- **Hyperliquid Perps SDK (~1.5MB):** HL SDK clients are sync getters used throughout the HL services; deferring needs care + perp device testing.
- **Stripe + RevenueCat (~1MB):** stripe import is ordering-coupled to RevenueCat init; RevenueCat is value-imported in 6 places incl. base classes — needs a real Prime-payment-layer lazy-load + payment QA.

https://claude.ai/code/session_01EwxVP1WraZMJfNSaPRmLeV


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/5738435e-3658-4073-af50-1aec0fa09916?refs=d847fdf846bc1ecc7753bf440d12b0de638aa1dc&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/5738435e-3658-4073-af50-1aec0fa09916/license?refs=d847fdf846bc1ecc7753bf440d12b0de638aa1dc&source=prChecksComment) |

packages/kit-bg/src/apis/BackgroundApiBase.ts
```diff
@@ -327,11 +330,46 @@ class BackgroundApiBase implements IBackgroundApiBridge {
...
+          `[${scope as string}] ProviderApi loader is not found.`,
+        );
+      }
+      loading = loader().then((module) => {
+        const ProviderApiClass = module.default;
+        const instance = new ProviderApiClass({ backgroundApi: this });
+        this.providers[scope] = instance;
+        return instance;
+      });
+      this.providerApiLoadingCache[scope] = loading;
```
sidmorizon: **Failed dynamic import permanently poisons the provider loading cache**

Severity: `severe`

If the dynamic `import()` in `getProviderApi` fails (e.g., network error loading a webpack chunk — a common scenario on web, which is why this project even uses `RetryChunkLoadPlugin`), the rejected promise is stored in `providerApiLoadingCache[scope]` and never cleared. On the next call, `this.providers[scope]` is still `undefined` (the success handler never ran), so the code falls through to `this.providerApiLoadingCache[scope]` which returns the stale rejected promise. This makes the provider permanently unavailable for the lifetime of the session — every subsequent dapp request to that chain will fail with the original load error.

<details><summary>Problematic flow in getProviderApi</summary>

1. First call: `loader()` rejects → rejected promise stored in `providerApiLoadingCache[scope]`
2. Second call: `this.providers[scope]` is `undefined` → `providerApiLoadingCache[scope]` is the rejected promise → returned as-is, no retry

The fix should clear the cache entry on rejection, e.g.:
```ts
loading = loader().then(...).catch((err) => {
  delete this.providerApiLoadingCache[scope];
  throw err;
});
```
</details>

Suggested fix:

```ts
      loading = loader().then((module) => {
        const ProviderApiClass = module.default;
        const instance = new ProviderApiClass({ backgroundApi: this });
        this.providers[scope] = instance;
        return instance;
      }).catch((err) => {
        delete this.providerApiLoadingCache[scope];
        throw err;
      });
      this.providerApiLoadingCache[scope] = loading;
```

packages/kit-bg/src/services/ServiceDApp.ts
```diff
@@ -1255,31 +1257,35 @@ class ServiceDApp extends ServiceBase {
...
+    // un-loaded chains have no connected dapp session to update.
+    Object.values(this.backgroundApi.providers).forEach((provider) => {
+      if (!provider) {
+        return;
+      }
+      provider.notifyDappAccountsChanged({
+        send: this.backgroundApi.sendForProvider(provider.providerName),
+        targetOrigin,
+      });
+    });
```
sidmorizon, line 1304: **Behavioral change: notify methods now skip unloaded providers**

Type: `Flag`

In the old code, `Object.values(this.backgroundApi.providers)` on the getter-based object would trigger ALL enumerable getters (they were defined with `enumerable: true`), synchronously instantiating every chain provider. The new code at `packages/kit-bg/src/services/ServiceDApp.ts:1296` and `:1310` only iterates providers already loaded into the `providers` map.

This means that after an ext MV3 service-worker restart, if `warmupConnectedDappProviders` hasn't completed and something triggers `notifyDAppAccountsChanged` (e.g., user switches account very quickly after restart), chain-specific providers won't receive the notification. The dapp will show stale account info until its next RPC call re-triggers provider loading.

The warmup at `packages/kit-bg/src/services/ServiceBootstrap.ts:134` mitigates this, but since `initDeferred()` is invoked with `void` (line 19, not awaited), there's a theoretical timing gap. In practice this is likely sub-second and the dapp self-heals on next interaction.

huhuanming:
### Runtime verification (served prod build, real browser via Playwright)

Loaded the production build of this branch in a headless Chromium and let it settle 14s:

```
#root childElementCount = 1   innerHTML length = 2,289,235
body text: "市场  合约  DeFi  兑换  推荐  连接  …  华尔街精选  Mag 7  AI 科技 …"  ← Home/Market full UI
JS resources loaded = 56
console.error = 75 (all net::ERR_CONNECTION_CLOSED — backend APIs unreachable on local static serve)
pageerror = 0
CHUNK-LOAD errors = 0
failed .js requests = 0
VERDICT: BOOTED + RENDERED ✓
```

| Check | Result |
|---|---|
| App boots (bg init → `createBackgroundProviders` → `getProviderApi` wiring) | ✅ no white screen / init crash |
| Home/Market full UI renders | ✅ |
| Lazy provider/lottie/swap chunks load on demand | ✅ **0 chunk-load errors** |
| JS page errors from the refactor | ✅ 0 |

The 75 console errors are all `net::ERR_CONNECTION_CLOSED` (market/price API calls failing because the local static serve has no backend) — **unrelated to this change**; the harness separately reports JS-failure = 0 and chunk-load failure = 0.

**Boundary / still needs manual QA:** this covers boot + lazy-chunk loading. It does **not** exercise the live dapp provider *dispatch* (`getProviderApi` firing when a dapp invokes a chain method) — that needs a real dapp + wallet. Please regression-test dapp connect (EVM/Solana/Aptos/Tron…), `accountsChanged`/`chainChanged`, the Hyperliquid `ethereum` broadcast, and native-webview dapp provider init before merge.

packages/kit-bg/src/apis/BackgroundApiBase.ts
huhuanming: Fixed in 8de06d92d6. Added a `.catch` on the loader that deletes `providerApiLoadingCache[scope]` on rejection, so a transient chunk-load failure no longer permanently poisons the cache — the next call retries the import instead of returning the stale rejected promise. Good catch, thanks.

packages/kit-bg/src/services/ServiceDApp.ts
huhuanming: Addressed in c8854a2159.

You're right to flag the reconnection assumption — it holds on native/desktop (a restart rebuilds the webviews, so the dapp reloads and reconnects), but **not** on ext MV3: the service worker can be evicted and restarted while the dapp tab stays alive, so `backgroundApi.providers` is reset to just `$private` while the connection in storage is still live. In that window a wallet-initiated `notifyDApp*` would indeed miss a chain whose provider hasn't been lazy-loaded yet.

Fix (ext-only): added `ServiceDApp.warmupConnectedDappProviders()`, run from `ServiceBootstrap`'s deferred init (which every SW cold start goes through). It reads the persisted connections via `getInjectProviderConnectedList()` (IndexedDB, survives SW restart), maps each connected `networkImpl` → scope, and pre-loads the providers that have a loader. Loads are deduped by `providerApiLoadingCache`, so it never double-constructs against concurrent dapp traffic, and it's gated to ext so native/desktop don't pay startup cost they don't need.

One honest caveat: a tiny residual race remains — if a notify fires before warmup finishes loading a chunk, that one push can still be missed; it self-heals on the dapp's next inbound RPC. We deliberately didn't async-load the hot `notifyDApp*` path to close that last sliver, since it's a low-severity, self-healing, no-funds-at-risk UI-staleness window (tx building uses live `getAccountsInfo`, not the cached push value).

packages/kit-bg/src/providers/backgroundProviders.ts
```diff
@@ -8,293 +8,61 @@ import type {
...
+  [IInjectedProviderNames.scdo]: () => import('./ProviderApiScdo'),
+  [IInjectedProviderNames.cardano]: () => import('./ProviderApiCardano'),
+  [IInjectedProviderNames.cosmos]: () => import('./ProviderApiCosmos'),
+  [IInjectedProviderNames.polkadot]: () => import('./ProviderApiPolkadot'),
+  [IInjectedProviderNames.webln]: () => import('./ProviderApiWebln'),
+  [IInjectedProviderNames.nostr]: () => import('./ProviderApiNostr'),
+  [IInjectedProviderNames.btc]: () => import('./ProviderApiBtc'),
+  [IInjectedProviderNames.neo]: () => import('./ProviderApiNeoN3'),
+  [IInjectedProviderNames.stellar]: () => import('./ProviderApiStellar'),
+};
```
sidmorizon, line 50: **Extension chunk-loading considerations for dynamic import()**

Type: `Flag`

The old `require()`-based pattern meant ALL provider code was bundled into the extension's main chunk (only instantiation was deferred). With `import()`, webpack emits separate async chunks for each provider. In Manifest V3 extension service workers, Chrome supports dynamic `import()` of extension-local files, so this should work. However, it's worth verifying that:
1. The extension webpack config doesn't interfere with dynamic chunk loading in the service worker context.
2. The `RetryChunkLoadPlugin` (currently web-only per `isWeb` guard in webpack.prod.config.js) is not needed for extensions since they load from local files, not the network.

If the extension build has any custom chunk-splitting or the service worker has restrictions on dynamic imports for certain manifest configurations, this could surface as a provider loading failure.

huhuanming: Verified — this is safe on ext, and your RetryChunkLoadPlugin point is correct.

The ext webpack config already forces no code-splitting for the MV3 background bundle: `webpack.ext.config.js:200-204` calls `disableCodeSplitChunks()` for the `bg` config (comment: "manifest v3 background can NOT split code"), which sets `output.asyncChunks = false` (`development/webpack/ext/codeSplit.js:54-58`). With `asyncChunks: false`, webpack inlines every `import()` target back into the background bundle instead of emitting a separate async chunk — so the service worker never performs a runtime chunk fetch. That removes the failure mode entirely (there is no chunk whose load could fail), and confirms `RetryChunkLoadPlugin` (web-only via the `isWeb` guard) is correctly not needed for extensions.

Trade-off worth noting: because of that, the startup-bundle trim from this PR does **not** apply to the ext background — providers stay inlined, same bundling as the old `require()` path. On ext, `import()` only defers instantiation (functionally equivalent to the previous sync require), so there's no provider-loading regression. The actual bundle win lands on web (real async chunks) and native (split-bundle segments).

originalix (approved):

PR #12136, opened 2026-06-23
fix: quarantine failed IP Table endpoints(OK-56653)

limichange:
[OK-56653](https://onekeyhq.atlassian.net/browse/OK-56653)

----------

## Summary
- Add an IP Table quarantine table for IPs that repeatedly fail real SNI requests.
- Keep the normal wallet-health speed test flow unchanged, but skip active quarantined IPs during re-selection.
- Preserve axios `validateStatus` behavior for SNI HTTP responses and add focused adapter/service/storage tests.

## Intent & Context
This fixes OK-56653, where Android logs showed `IpTableAdapter` timing out when requesting `utility.onekeycn.com` through a selected IP such as `216.19.4.106`.

The review concern was valid: probing business hosts such as `utility` and `earn` during every normal IP selection would couple service availability into the startup/speed-test path and slow it down. This update keeps the original selection flow and adds a reactive safety mechanism only after a real SNI failure is observed.

## Root Cause
The old IP Table flow selected a root-domain IP using `wallet.${domain}/wallet/v1/health`. The shared request adapter then reused that same root-domain selection for service hosts such as `utility.onekeycn.com`.

If an IP is healthy for the wallet health check but fails SNI for a service host, the old retry path could eventually run another wallet-only speed test and select the same IP again.

## Design Decisions
- Keep normal selection behavior wallet-health based. No extra `utility`/`earn` probes are added to the normal speed-test path.
- When the same selected IP reaches `IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD` consecutive real SNI failures, write it to `runtime.quarantinedIps[domain][ip]`, switch the current selection to domain, and trigger async re-selection.
- During speed tests, skip active quarantined IPs. This prevents the same IP from being selected again while the quarantine TTL is active.
- The adapter also reads the shared quarantine table. If the current selection is quarantined, normal mode falls back to domain; strict mode uses the first non-quarantined configured IP.
- Shared-domain requests such as `*.onekey.so` use the mapped IP Table lookup domain for quarantine writes, so the runtime key matches the key used for selection reads.
- The adapter records a short 10-second in-memory quarantine immediately when SNI throws or returns null. This closes the same-runtime retry loop without turning one transient failure into a 30-minute persisted quarantine.
- The quarantine state is runtime data in SimpleDB, not user data and not a Realm/IndexedDB schema change.
- Runtime model: `ServiceIpTable` runs in `bg` and writes shared persisted/native storage. The request adapter can run in `bg` and, depending on platform wiring, possibly `main`; JS heaps and memo caches are per-runtime, while the persisted quarantine table is shared. The runtime where the SNI failure happens clears its adapter memo immediately; other runtimes are protected once they read the shared quarantine state, with the selected-IP memo TTL shortened to 1 second as the upper bound for already-cached selections.
- Preserve axios semantics in the real SNI adapter path. Transport-level SNI failures still fall back to the original adapter, but HTTP responses now respect `validateStatus`.

## Changes Detail
- `ServiceIpTable`: quarantine failed IPs after consecutive real SNI failures reach the IP quarantine threshold, switch selection to domain, trigger re-selection, and skip active quarantined IPs during speed tests.
- `SimpleDbEntityIpTable`: persist `runtime.quarantinedIps` and clear adapter selection cache after IP Table state mutations.
- `ipTableAdapter`: read quarantine state before using a selected IP or strict-mode fallback; report failed hostname/error with IP failure callbacks; keep non-2xx speed probes unavailable.
- `ipTable` shared types/defaults: add quarantine runtime shape and TTL constant.
- Tests: cover wallet-only normal selection, quarantine skip/retry behavior, adapter quarantine reads, shared-domain failure mapping, same-runtime in-memory quarantine, storage pruning, strict-mode quarantine fallback, and SNI HTTP error semantics.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile/native IP Table request path, especially Android; shared request adapter consumers on platforms where IP Table is enabled.
- **Risk Areas**: Request fallback behavior after SNI failures, persisted runtime quarantine compatibility, and strict-mode fallback selection.

## Test plan
- [x] `npx jest packages/kit-bg/src/services/ServiceIpTable.test.ts packages/shared/src/request/helpers/ipTableAdapter.test.ts packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityIpTable.test.ts --runInBand`
- [x] `npx oxlint --tsconfig ./tsconfig.json --type-aware --deny-warnings packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityIpTable.test.ts packages/kit-bg/src/services/ServiceIpTable.test.ts packages/shared/src/request/constants/ipTableDefaults.ts packages/shared/src/request/helpers/ipTableAdapter.ts packages/shared/src/request/helpers/ipTableAdapter.test.ts packages/kit-bg/src/services/ServiceIpTable.ts packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityIpTable.ts packages/shared/src/request/types/ipTable.ts`
- [x] `git diff --check origin/x...HEAD`
- [x] `npx tsgo -p ./tsconfig.json --noEmit --tsBuildInfoFile /tmp/app-mono-ts-cache-iptable-quarantine-x --pretty false`

[OK-56653]: https://onekeyhq.atlassian.net/browse/OK-56653?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceIpTable.ts
+++ b/packages/kit-bg/src/services/ServiceIpTable.ts
@@ -16,7 +16,9 @@ import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
 import {
   IP_TABLE_INITIAL_SPEED_TEST_DELAY_MS,
   IP_TABLE_PERFORMANCE_IMPROVEMENT_THRESHOLD,
+  IP_TABLE_SNI_FAILURE_QUARANTINE_MS,
   IP_TABLE_SNI_FAILURE_THRESHOLD,
+  IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD,
   IP_TABLE_SPEED_TEST_COOLDOWN_MS,
   IP_TABLE_SPEED_TEST_DELAY_MS,
   IP_TABLE_SPEED_TEST_ITERATIONS,
@@ -33,6 +35,7 @@ import { getRequestHeaders } from '@onekeyhq/shared/src/request/Interceptor';
 import type {
   IIpTableConfigWithRuntime,
   IIpTableRemoteConfig,
+  IIpTableRuntime,
 } from '@onekeyhq/shared/src/request/types/ipTable';
 import {
   isSupportIpTablePlatform,
@@ -441,6 +444,27 @@ class ServiceIpTable extends ServiceBase {
     return avgLatency;
   }
 
+  private getActiveQuarantinedIps(
+    runtime: IIpTableRuntime | undefined,
+    domain: string,
+    now = Date.now(),
+  ): Set<string> {
+    const domainQuarantines = runtime?.quarantinedIps?.[domain];
+    if (!domainQuarantines) {
+      return new Set();
+    }
+
+    return new Set(
+      Object.entries(domainQuarantines)
+        .filter(
+          ([, quarantine]) =>
+            now - quarantine.lastFailureTime <
+            IP_TABLE_SNI_FAILURE_QUARANTINE_MS,
+        )
+        .map(([ip]) => ip),
+    );
+  }
+
   /**
    * Clean up health statistics after speed test
    * Removes health data for IPs that are no longer selected (memory optimization)
@@ -532,29 +556,40 @@ class ServiceIpTable extends ServiceBase {
 
       // 2. Test all IPs with SNI
       const ipResults = new Map<string, number>();
+      const activeQuarantinedIps = this.getActiveQuarantinedIps(
+        configWithRuntime.runtime,
+        domain,
+      );
 
       for (const endpoint of domainConfig.endpoints) {
-        defaultLogger.ipTable.request.info({
-          info: `[IpTable] Testing IP: ${endpoint.ip} for ${domain}`,
-        });
+        if (activeQuarantinedIps.has(endpoint.ip)) {
+          defaultLogger.ipTable.request.warn({
+            info: `[IpTable] Skipping quarantined IP during speed test: ${domain} -> ${endpoint.ip}`,
+          });
+          ipResults.set(endpoint.ip, Infinity);
+        } else {
+          defaultLogger.ipTable.request.info({
+            info: `[IpTable] Testing IP: ${endpoint.ip} for ${domain}`,
+          });
 
-        const ipLatency = await this.testMultipleTimes(() =>
-          testIpSpeed(
-            endpoint.ip,
-            domain,
-            ONEKEY_HEALTH_CHECK_URL,
-            IP_TABLE_SPEED_TEST_TIMEOUT_MS,
-          ),
-        );
+          const ipLatency = await this.testMultipleTimes(() =>
+            testIpSpeed(
+              endpoint.ip,
+              domain,
+              ONEKEY_HEALTH_CHECK_URL,
+              IP_TABLE_SPEED_TEST_TIMEOUT_MS,
+            ),
+          );
 
-        ipResults.set(endpoint.ip, ipLatency);
+          ipResults.set(endpoint.ip, ipLatency);
 
-        defaultLogger.ipTable.request.info({
-          info: `[IpTable] IP test result: ${endpoint.ip} -> ${ipLatency}ms`,
-        });
+          defaultLogger.ipTable.request.info({
+            info: `[IpTable] IP test result: ${endpoint.ip} -> ${ipLatency}ms`,
+          });
+        }
       }
 
-      // 3. Find best IP
+      // 3. Find the fastest non-quarantined IP
       let bestIp = '';
       let bestIpLatency = Infinity;
 
@@ -580,16 +615,17 @@ class ServiceIpTable extends ServiceBase {
         } else {
           // All tests failed
           defaultLogger.ipTable.request.info({
-            info: `[IpTable] All tests failed for ${domain}`,
+            info: `[IpTable] All tests failed for ${domain}, using domain`,
           });
+          await this.backgroundApi.simpleDb.ipTable.updateSelection(domain, '');
         }
         return;
       }
 
       if (bestIpLatency === Infinity) {
-        // All IP tests failed, use domain
+        // All IP tests failed or all candidate IPs are quarantined, use domain
         defaultLogger.ipTable.request.info({
-          info: `[IpTable] All IP tests failed, using domain: ${domain}`,
+          info: `[IpTable] No available IP passed speed test, using domain: ${domain}`,
         });
         await this.backgroundApi.simpleDb.ipTable.updateSelection(domain, '');
         return;
@@ -691,6 +727,10 @@ class ServiceIpTable extends ServiceBase {
     domain: string,
     requestType: 'ip' | 'domain',
     target: string,
+    params?: {
+      hostname?: string;
+      error?: string;
+    },
   ): Promise<void> {
     // Check if IP Table is enabled
     if (!(await this.isIpTableEnabled())) {
@@ -717,6 +757,38 @@ class ServiceIpTable extends ServiceBase {
         endpointHealth.consecutiveFailures
       }) for ${domain} (${target})`,
     });
+
+    if (requestType === 'ip') {
+      if (
+        endpointHealth.consecutiveFailures <
+        IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD
+      ) {
+        defaultLogger.ipTable.request.info({
+          info: `[IpTable] IP quarantine not triggered: ${endpointHealth.consecutiveFailures}/${IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD} consecutive failures for ${domain} (${target})`,
+        });
+        return;
+      }
+
+      await this.backgroundApi.simpleDb.ipTable.markIpQuarantined(
+        domain,
+        target,
+        {
+          hostname: params?.hostname,
+          error: params?.error,
+          timestamp: now,
+        },
+      );
+      await this.backgroundApi.simpleDb.ipTable.updateSelection(domain, '');
+
+      defaultLogger.ipTable.request.warn({
+        info: `[IpTable] Quarantined failed IP and switched to domain: ${domain} -> ${target}`,
+      });
+
+      stats.lastSpeedTestTime = now;
+      endpointHealth.consecutiveFailures = 0;
+      void this.selectBestEndpointForDomain(domain);
+      return;
+    }
 
     // Check if speed test should be triggered
     const shouldTrigger = await this.shouldTriggerSpeedTest(domain, stats);
@@ -767,9 +839,17 @@ class ServiceIpTable extends ServiceBase {
     }, IP_TABLE_INITIAL_SPEED_TEST_DELAY_MS);
   }
 
-  private async hasRuntimeSelections(): Promise<boolean> {
+  private async hasRuntimeSelectionForDomain(domain: string): Promise<boolean> {
     const configWithRuntime = await this.getConfig();
-    return Object.keys(configWithRuntime.runtime?.selections ?? {}).length > 0;
+    return configWithRuntime.runtime?.selections?.[domain] !== undefined;
+  }
+
+  private async getCurrentDomain(): Promise<string> {
+    const { enabled: devSettingEnabled, settings } =
+      await devSettingsPersistAtom.get();
+    return devSettingEnabled && settings?.enableTestEndpoint
+      ? ONEKEY_TEST_API_HOST
+      : ONEKEY_API_HOST;
   }
 
   @backgroundMethod()
@@ -785,9 +865,16 @@ class ServiceIpTable extends ServiceBase {
     });
 
     // Register request failure callback (handles both IP and domain failures)
-    setReportRequestFailureCallback(({ domain, requestType, target }) => {
-      void this.reportRequestFailure(domain, requestType, target);
-    });
+    setReportRequestFailureCallback(
+      ({ domain, requestType, target, hostname, error }) => {
+        void this.reportRequestFailure(domain, requestType, target, {
+          hostname,
+          error,
+        });
+      },
+    );
+
+    const domain = await this.getCurrentDomain();
 
     // Try to refresh CDN config if needed
     const shouldRefresh = await this.shouldRefreshConfig();
@@ -813,7 +900,7 @@ class ServiceIpTable extends ServiceBase {
       defaultLogger.ipTable.request.info({
         info: '[IpTable] CDN config is up to date',
       });
-      needSpeedTest = !(await this.hasRuntimeSelections());
+      needSpeedTest = !(await this.hasRuntimeSelectionForDomain(domain));
     }
 
     // Execute speed test if needed
--- a/packages/shared/src/request/helpers/ipTableAdapter.ts
+++ b/packages/shared/src/request/helpers/ipTableAdapter.ts
@@ -1,13 +1,19 @@
-import axios, { AxiosHeaders } from 'axios';
+import axios, { AxiosError, AxiosHeaders } from 'axios';
 
 import { OneKeyLocalError } from '../../errors';
 import { defaultLogger } from '../../logger/logger';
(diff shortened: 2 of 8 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3adfd429-fcd2-492d-bc87-71108fec772e?refs=a1781946406b3c9dbf25c9a4b78f12a58876e776&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3adfd429-fcd2-492d-bc87-71108fec772e/license?refs=a1781946406b3c9dbf25c9a4b78f12a58876e776&source=prChecksComment) |

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -543,107 +594,117 @@ export function createIpTableAdapter(
...
+        clearSelectedIpForHostCache();
         if (reportRequestFailureCallback) {
           reportRequestFailureCallback({
             domain: rootDomain,
             requestType: 'ip',
             target: selectedIp,
-            error: 'SNI response null',
+            hostname,
+            error: error instanceof Error ? error.message : String(error),
           });
```
sidmorizon, line 670: **Quarantine and failure reporting use rootDomain, but IP selection uses mappedDomain for SHARED_IP_DOMAINS**

Type: `Flag`

For domains in `SHARED_IP_DOMAINS` (currently just `onekey.so`), the adapter maps the domain to the API host (e.g., `onekeycn.com`) when looking up IP selections and quarantine state at `ipTableAdapter.ts:178-179`. However, when the adapter reports a failure, it passes `rootDomain` (e.g., `onekey.so`) at `ipTableAdapter.ts:614`, not the mapped domain. This means quarantine entries are stored under `onekey.so` in the DB (`ServiceIpTable.ts:774`), but the quarantine check in `getSelectedIpForHostInternal` looks under `onekeycn.com` (`ipTableAdapter.ts:196`). As a result, quarantine is ineffective for shared-domain requests — each request will still try the bad IP, fail, fall back to domain, and repeat. The fallback itself works correctly so user impact is limited to unnecessary SNI attempts per request. This is a pre-existing architectural mismatch (the old failure-tracking/speed-test logic had the same problem) but it now also affects the new quarantine feature.

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -718,6 +770,28 @@ class ServiceIpTable extends ServiceBase {
...
+      await this.backgroundApi.simpleDb.ipTable.updateSelection(domain, '');
+
+      defaultLogger.ipTable.request.warn({
+        info: `[IpTable] Quarantined failed IP and switched to domain: ${domain} -> ${target}`,
+      });
+
+      stats.lastSpeedTestTime = now;
+      endpointHealth.consecutiveFailures = 0;
+      void this.selectBestEndpointForDomain(domain);
+      return;
```
sidmorizon, line 790: **Single IP failure now triggers immediate quarantine for 30 minutes — significant behavioral change from 10-failure threshold**

Type: `Flag`

Previously, the system required `IP_TABLE_SNI_FAILURE_THRESHOLD` (10) consecutive failures before triggering a speed test and switching endpoints. Now, in `ServiceIpTable.ts:773-792`, a single IP failure immediately quarantines the IP for 30 minutes (`IP_TABLE_SNI_FAILURE_QUARANTINE_MS`), switches the selection to domain, and fires a background speed test. This means a single transient network hiccup (e.g., one timed-out request) will quarantine a potentially healthy IP for 30 minutes. The old threshold-based logic at lines 795-830 now only applies to `requestType === 'domain'` failures. Whether this aggressiveness is desirable depends on how common transient SNI failures are in practice.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -224,6 +249,10 @@ const getSelectedIpForHost = memoizee(getSelectedIpForHostInternal, {
   primitive: true, // hostname is a string primitive, use simple equality check
 });
 
+export function clearSelectedIpForHostCache(): void {
+  getSelectedIpForHost.clear();
```
sidmorizon: # [🟡 P2] 隔离写入只清除 bg 运行时的 memo 缓存，main 运行时最多 5s 内仍命中被隔离的 IP

## 问题

`clearSelectedIpForHostCache()` 清除的是当前运行时模块级单例 `getSelectedIpForHost`（`memoizee`，`maxAge: 5000`）。在生产分线程平台（iOS/Android），`updateInterceptorRequestHelperWithIpTable` 同时在 `main` 与 `bg` 两个运行时各装一份，`getSelectedIpForHost` 是**每运行时一份**的 JS 堆副本，而底层 MMKV 是共享的同一份原生存储。

`SimpleDbEntityIpTable.markIpQuarantined` / `updateSelection` 跑在 `bg`，写完后调用的 `clearSelectedIpForHostCache()` 只清掉了 **bg** 那份 memo。`main` 运行时的 memo 不会被清，仍返回旧的 `runtime.selections[domain] = 被隔离IP`。

## 影响

`bg` 侧探测失败触发隔离后，`main` 运行时在 `maxAge` 窗口内（最长 `5000` ms）发出的 `appApiClient` 等请求仍会选中刚被隔离的 IP，每次都要走一遍 SNI 失败 → fallback domain，直到某个 main 请求自身失败并清掉 main 的 memo 才自愈。窗口有界且能自愈，但跨运行时存在短暂不一致；隔离是写共享 MMKV 的，缓存失效却是单运行时的，二者不对称。

## 建议

让缓存失效跨运行时生效：例如对 `quarantinedIps` 相关的选择不走 5s memo（或显著缩短其 `maxAge`），或通过现有 bg↔main 通道在写入隔离后广播一次 `clearSelectedIpForHostCache()`，使两个运行时的 memo 同时失效。

[Claude session ↗](https://claude.ai/code/session_01NPjHUGB3UovRDD9XMJoRVe)

---

sidmorizon (commented):
IP Table quarantine review — 1 cross-runtime cache-consistency note.

---

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -718,6 +757,28 @@ class ServiceIpTable extends ServiceBase {
...
```
sidmorizon, line 790: **Every IP failure immediately quarantines and triggers a speed test without deduplication**

Type: `Flag`

In `ServiceIpTable.reportRequestFailure` (line 760-779), every IP failure immediately quarantines the IP, clears the selection, and fires `selectBestEndpointForDomain` via `void`. Unlike the domain failure path which checks `shouldTriggerSpeedTest` (with cooldown and threshold), the IP path has no deduplication. If many concurrent requests fail on the same IP simultaneously, each will call `markIpQuarantined` + `updateSelection` + fire a new speed test. The DB operations are serialized by the SimpleDb mutex, but multiple overlapping speed tests could run. The speed test itself is idempotent, so this causes wasted network traffic rather than incorrect behavior.

packages/shared/src/request/helpers/ipTableAdapter.ts
limichange: Fixed in d6549cb3af.

The adapter now keeps the selected lookupDomain through the SNI request path and reports failures with that lookupDomain instead of rootDomain. For shared domains such as *.onekey.so, quarantine writes now use the same key used by selection/quarantine reads, e.g. onekeycn.com or onekeytest.com.

Covered by the shared-domain SNI failure mapping test.

packages/kit-bg/src/services/ServiceIpTable.ts
limichange: Fixed in d6549cb3af.

Persisted 30-minute quarantine no longer happens on the first IP failure. The IP path now requires IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD consecutive SNI failures for the same IP before writing runtime.quarantinedIps, clearing selection, and triggering re-selection.

The first failure only gets a short 10-second same-runtime in-memory backoff in the adapter, so a transient timeout does not become a 30-minute persisted quarantine. The service test was updated to assert the first failure does not persist quarantine and the second consecutive failure does.

packages/shared/src/request/helpers/ipTableAdapter.ts
limichange: Fixed in d6549cb3af.

Runtime model here is still: bg writes shared persisted/native SimpleDB state, while main and bg each have their own JS heap and their own memoized selected-IP cache. This PR does not add a cross-runtime cache-invalidation event.

To address the bounded stale window, the selected-IP memo maxAge was reduced from 5s to 1s. The runtime that observes the SNI failure clears its own memo immediately and applies a short 10-second in-memory quarantine; the other runtime can still have an independent JS memo entry, but it is now bounded to 1s before it re-reads the shared persisted quarantine state.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -217,13 +278,27 @@ async function getSelectedIpForHostInternal(
...
 }
 
-const getSelectedIpForHost = memoizee(getSelectedIpForHostInternal, {
+const getSelectedIpInfoForHost = memoizee(getSelectedIpInfoForHostInternal, {
   promise: true,
-  maxAge: 5000, // 5 seconds cache
+  maxAge: IP_TABLE_SELECTED_IP_CACHE_MS,
   max: 100, // Max 100 hostname cached
   primitive: true, // hostname is a string primitive, use simple equality check
 });
```
sidmorizon, line 286: **Memoize cache reduced from 5s to 1s increases async overhead**

Type: `Flag`

The `getSelectedIpInfoForHost` memoize `maxAge` was reduced from 5000ms to `IP_TABLE_SELECTED_IP_CACHE_MS` (1000ms) at `packages/shared/src/request/helpers/ipTableAdapter.ts:283`. This means every hostname's IP lookup is re-evaluated every second instead of every 5 seconds. Each evaluation calls `shouldUseIpTable()` → `requestHelper.getDevSettingsPersistAtom()` and `requestHelper.getIpTableConfig()` (DB reads). Under high request throughput, this 5× increase in cache miss rate could add measurable latency, though the explicit `clearSelectedIpForHostCache()` calls on data changes mitigate the staleness concern that the shorter TTL addresses.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -269,6 +344,28 @@ function axiosHeadersToPlainObject(
...
+  }
+
+  throw new AxiosError(
+    `Request failed with status code ${response.status}`,
+    errorCode,
+    response.config,
+    response.request,
+    response,
+  );
+}
```
sidmorizon, line 367: **settleAxiosResponse adds previously-missing validateStatus enforcement for SNI responses**

Type: `Flag`

Before this PR, SNI responses with non-2xx status codes (e.g., 500) were returned as resolved promises — the adapter did not apply `config.validateStatus`. This meant callers expecting axios's standard behavior (rejecting on 4xx/5xx by default) would silently receive error responses as successes. The new `settleAxiosResponse` function at line 347-367 fixes this by applying the same `validateStatus` logic as axios's internal `settle.js`. This is a significant behavioral fix — existing code that relied on the old (incorrect) pass-through of error statuses could see new rejections.

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -718,6 +758,38 @@ class ServiceIpTable extends ServiceBase {
       }) for ${domain} (${target})`,
     });
 
+    if (requestType === 'ip') {
+      if (
+        endpointHealth.consecutiveFailures <
```
sidmorizon, line 763: # [🟡 P2] `consecutiveFailures` 永不在成功时归零，"连续 2 次"实为"累计 2 次"，健康 IP 仍会被持久隔离 30 分钟

## 问题

新的阈值判断用 `endpointHealth.consecutiveFailures`（按 IP 维护于 `domainHealthMap`，进程内存）来决定是否写入 30 分钟持久隔离：

```ts
if (requestType === 'ip') {
  if (endpointHealth.consecutiveFailures < IP_TABLE_SNI_QUARANTINE_FAILURE_THRESHOLD) {
    return; // 第 1 次只记数
  }
  // 第 2 次 -> markIpQuarantined + updateSelection('') + 重置该 IP 计数
}
```

但代码库中**没有任何成功路径会把 `consecutiveFailures` 归零**（全仓搜索仅 L750/L752 在域名 speed-test 触发时、以及 L789 隔离触发后会重置；适配器侧也没有 `reportRequestSuccess` 之类回调）。`lastFailureTime` 虽被写入（L711）却从不被读取来做时间窗口老化。

因此对同一个 IP：第 1 次失败后只做 10s 内存 backoff，selection 不清除，10s 后该 IP 重新可选；其间无论成功多少次都不会复位计数。等到该 IP 在**运行时生命周期内的任意时刻**再失败一次（哪怕距上次失败已数小时），`consecutiveFailures` 就达到 2 → 写入 30 分钟持久隔离。这里的 "consecutive" 名不副实，实际语义是"累计 2 次"。

## 影响

这部分抵消了本次提交的既定目标（resolved thread r3458928249 中说明：让"瞬时超时不再变成 30 分钟持久隔离"）。阈值从 1 提到 2 只把"1 次瞬时失败即隔离"放宽为"任意两次跨时间的瞬时失败即隔离"，对偶发抖动的健康 IP 仍会触发 30 分钟持久隔离（虽有 domain fallback，用户影响有限）。新增测试只覆盖"背靠背两次失败"，未覆盖"中间穿插成功"的场景，掩盖了该缺口。

## 建议

让计数体现真正的"连续"语义，二选一：

- 在 SNI 请求成功路径上回调 service 复位该 IP 的 `consecutiveFailures`（需要新增 success 上报通道）；或
- 在 `reportRequestFailure` 入口用已有的 `lastFailureTime` 做时间窗口：若 `now - endpointHealth.lastFailureTime > 某窗口`（如 quarantine in-memory 窗口或 cooldown），先把 `consecutiveFailures` 归零再自增，使跨越较长时间的孤立失败不被累计。

[Claude session ↗](https://claude.ai/code/session_01NPjHUGB3UovRDD9XMJoRVe)

---

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -44,14 +50,24 @@ interface IRequestFailureParams {
...
+  ip: string;
+  lookupDomain: string;
+  rootDomain: string;
+}
+
 let reportRequestFailureCallback:
   | ((params: IRequestFailureParams) => void)
   | null = null;
 
+const inMemoryQuarantinedIps = new Map<string, Map<string, number>>();
```
sidmorizon, line 69: **In-memory quarantine is per-JS-runtime; bg runtime won't see main runtime quarantine immediately**

Type: `Flag`

Per CLAUDE.md's threading model, `inMemoryQuarantinedIps` (line 69) exists independently in both `main` and `bg` JS runtimes. When an SNI request fails in the main runtime, `markIpQuarantinedInMemory` quarantines the IP in the **main** runtime's map. The **bg** runtime's copy remains empty. The bg runtime relies on persistent DB quarantine (written by `ServiceIpTable.reportRequestFailure`), but the `reportRequestFailureCallback` is only registered in the bg runtime via `ServiceIpTable.init()`. This means: (1) main-runtime failures quarantine the IP in-memory for 10s in main only, (2) the failure callback is `null` in main, so no persistent quarantine or speed test is triggered from main-runtime failures. This is a pre-existing architectural limitation (not introduced by this PR), and the 1-second memoization TTL and the in-memory quarantine provide reasonable short-term protection for the main runtime.

PR #12146, opened 2026-06-24
feat: add desktop native messaging host for browser extension IPC

sidmorizon:
## Summary

Adds an **experimental, dev-only, macOS-only** bridge that lets the browser extension call the OneKey desktop app over **Chrome Native Messaging** to use the desktop's OS-level `safeStorage` for **owner-bound** encrypt/decrypt.

> ⚠️ **Status: EXPERIMENTAL / EXPLORATORY — NOT production-ready.** This lands the plumbing for review; it is gated off in production and has no production consumer yet.

## Gating (off in production)

Three independent switches, all keyed on dev:
1. Extension `nativeMessaging` permission — added only when `NODE_ENV !== 'production'`.
2. Desktop manifest/launcher install — skipped unless dev (`process.defaultApp`).
3. Desktop host serving — refuses unless dev.

- **macOS only**; MAS (App Store) explicitly excluded (sandbox can't manage these files).
- Dev detection uses `process.defaultApp`, **not** `app.isPackaged` (the latter is `true` even in OneKey's unpackaged dev runs).

## What's included

- `appEntry` dispatcher: runs the host on `--onekey-native-messaging-host`, otherwise the normal app.
- Native Messaging host process + per-browser manifest/launcher install.
- **Owner-bound safeStorage protocol**: the extension signs a challenge with a non-extractable P-256 `CryptoKey`; the host binds that owner into the encrypted envelope and verifies it on decrypt.
- `DesktopApiSafeStorageNative` (encrypt/decrypt/verify), shared consts/types, extension client.
- Keychain `setItem` `enableSync` is now **opt-in** (default `false`); iCloud backup opts in explicitly.
- Dev Settings: dedicated **"Desktop Native Messaging"** test page.

## Testing (dev, macOS)

Verified end-to-end from the extension Dev Settings page:
- `ping` → `pong`
- owner-bound safeStorage **round trip** → encrypt + decrypt, `matched: true`
- **failure cases** → tampered blob rejected (`SAFE_STORAGE_DECRYPT_FAILED`), owner mismatch rejected (`OWNER_AUTH_FAILED`)

## Known limitations & before-release checklist

Documented in the `packages/shared/src/consts/desktopNativeMessaging.ts` header:
- **Security**: same-user host impersonation (manifest/launcher in user-writable dirs), encrypt-plaintext exposure to a hijacked host, replay window (no nonce cache). Not a network MITM — it's a stdio pipe.
- **Lifecycle/edge cases not handled**: uninstall cleanup (orphaned files), switch to MAS build, stale launcher path after move/update, multi-channel manifest collisions.
- **Before formal release**: authenticated E2EE (keychain-anchored desktop identity + extension-pinned key + per-session ECDH), replay protection, cover the lifecycle edge cases, re-review before exposing any method beyond owner-bound safeStorage.

`lint:staged` and `tsc:staged` pass.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c46f7846-700e-4462-a221-54ba3b42d7e3?refs=5f5b0663ef7fa4afe3b2fc8c572557bb2f51181d&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/c46f7846-700e-4462-a221-54ba3b42d7e3/license?refs=5f5b0663ef7fa4afe3b2fc8c572557bb2f51181d&source=prChecksComment) |

apps/desktop/app/nativeMessagingHostInstall.ts
```diff
@@ -0,0 +1,262 @@
...
+  args.push(ONEKEY_DESKTOP_NATIVE_MESSAGING_HOST_ARG);
+
+  return args;
+}
+
+function getUnixHostScript() {
+  const commandParts = getHostLauncherArgs().map((arg) => shellQuote(arg));
+
+  return `#!/bin/sh
+exec ${commandParts.join(' ')} "$@"
```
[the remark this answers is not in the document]
sidmorizon: Fixed: the launcher now bakes `ONEKEY_NATIVE_MESSAGING_EXTENSION_IDS` into an `export` line (Unix) / `set` (Windows) at install time via `getHostLauncherEnv()`, so the Chrome-spawned host inherits the custom IDs and no longer returns `NATIVE_HOST_FORBIDDEN` for an allow-listed unpacked extension.

apps/ext/src/manifest/shared.js
```diff
@@ -1,15 +1,41 @@
...
+if (
+  getChromeExtensionIdFromPublicKey(devExtensionIdKey) !==
+  ONEKEY_EXTENSION_ID_DEV
+) {
+  throw new TypeError(
+    'DEV extension ID const does not match manifest public key',
+  );
+}
+
 // TODO:
```
sidmorizon, line 45: **Extension manifest 'key' field now hardcoded instead of env-driven**

Type: `Flag`

In `apps/ext/src/manifest/shared.js`, the dev extension ID key was previously described as "DEV-only and read from env so we don't hardcode a team-wide ID" (old comment). The new code hardcodes a specific public key and adds a build-time assertion that its derived ID matches `ONEKEY_EXTENSION_ID_DEV` from the shared const.

This is intentional — the Native Messaging host needs to know the dev extension ID at manifest install time to populate `allowed_origins`. Hardcoding ensures all developers get the same predictable extension ID for the native messaging allow-list. The `key` field is only relevant for unpacked/side-loaded extensions (Chrome Web Store ignores it).

The tradeoff: developers who previously used a custom key via env (to get a different extension ID) would now always get the hardcoded ID, which may conflict if they need a different ID for other OAuth redirect URL testing.

packages/kit-bg/src/desktopApis/DesktopApiKeychain.ts
```diff
@@ -69,29 +79,25 @@ class DesktopApiKeychain implements IAppleKeyChainStorage {
...
    */
   async setItem(params: IKeychainSetItemParams): Promise<void> {
-    this.ensureMacOS();
-    cloudkitSetContainerID(DESKTOP_ICLOUD_CONTAINER_ID);
-    return keychainSetItem(params);
+    setupDesktopKeychainContainer();
+    return keychainSetItem({
+      ...params,
+      enableSync: params.enableSync ?? false,
+    });
```
sidmorizon, line 100: **Keychain enableSync default changed from implicit-true to explicit-false**

Type: `Flag`

The old `DesktopApiKeychain.setItem` passed `params` straight through to `keychainSetItem`, relying on the native Swift default (documented as `true` for iCloud sync). The new code at `packages/kit-bg/src/desktopApis/DesktopApiKeychain.ts:99` explicitly defaults `enableSync` to `false`. This is a semantic behavioral change: any pre-existing caller that omitted `enableSync` expecting the old native-level default of `true` would now get `false`. I verified all call sites: `ICloudBackupProvider.ts:141` was updated to pass `enableSync: true` explicitly, and the dev test at `DesktopApiProxyTestDevSettings.tsx:524` passes `enableSync: false` explicitly. No call site is left relying on the old implicit default, so this is safe — but it's worth the reviewer confirming the native `keychainSetItem` behavior when `enableSync: false` is passed (i.e., that existing iCloud-synced keychain items remain queryable by `getItem`).

apps/ext/src/manifest/shared.js
sidmorizon: Thanks — clarifying, because the "previously env-driven" premise comes from a stale comment rather than actual behavior:

- On `origin/x`, `devExtensionIdKey` was already the **same hardcoded base64** and the manifest already emitted `key` unconditionally. The "read from env" line was an aspirational comment that was **never implemented** (the only `process.env` reads were `NODE_ENV`/`VERSION`). So there was no custom-key-via-env capability to lose — no regression.
- Hardcoding + the build-time assertion is intentional: the desktop Native Messaging host must know a predictable dev extension ID at manifest-install time to populate `allowed_origins`, and the assertion catches drift between this `key` and `ONEKEY_EXTENSION_ID_DEV` (drift would silently break native messaging).
- For the OAuth/custom-ID case: to let a **different** unpacked extension reach the host, set `ONEKEY_NATIVE_MESSAGING_EXTENSION_IDS=<id>` rather than changing this key. I added an inline comment in `shared.js` documenting exactly this, and restored the derived-`extensionId` annotation next to the key.

apps/desktop/app/nativeMessagingHost.ts
```diff
@@ -0,0 +1,324 @@
+import { app } from 'electron';
+
+import {
+  DesktopSafeStorageNativeError,
+  decryptDesktopSafeStorageString,
+  encryptDesktopSafeStorageString,
+  isDesktopSafeStorageAvailable,
+} from '@onekeyhq/kit-bg/src/desktopApis/DesktopApiSafeStorageNative';
```
originalix: 这里把 host 逻辑直接指到了 `@onekeyhq/kit-bg/src/desktopApis/DesktopApiSafeStorageNative`，但 desktop main 的现有 esbuild 解析链路目前并不能装载这个新子路径。我用同一套构建配置做了聚焦验证：现有 `DesktopApiKeychain` / `DesktopApiWebview` 都能解析，只有这个新 `DesktopApiSafeStorageNative` 直接报 `Could not resolve`。结果就是 `appEntry -> nativeMessagingHost` 这条分支在打包阶段就断掉，Native Messaging host 代码进不了最终的 main bundle。

修复方向建议二选一并补一个最小 smoke check：要么把 safeStorage helper 放到 desktop build 已经能解析的模块里，要么补齐这条新子路径对应的 resolver/package 可见性，确保 host 分支真的能被打进 `appEntry.js`。

sidmorizon: Moved the host-only safeStorage helper out of `@onekeyhq/kit-bg/src/desktopApis` into `apps/desktop/app/nativeMessagingSafeStorage.ts` so the host imports it via a relative path, and added `apps/desktop/scripts/nativeMessagingHostBundle.test.js` — a smoke check that bundles `appEntry.ts` under the real desktop main esbuild config (shared via `desktopMainEsbuildConfig.js`) and asserts the host branch + safeStorage crypto exports are present. FWIW the full `appEntry` bundle already resolved and included the host symbols on this branch (verified locally + the new test passes), so this hardens/guards the chain rather than unblocking a broken build.

apps/desktop/scripts/nativeMessagingHostBundle.test.js
```diff
@@ -0,0 +1,61 @@
...
+      ...getDesktopMainEsbuildResolveOptions(),
+      entryPoints: [path.join(electronSource, 'appEntry.ts')],
+      write: false,
+      metafile: true,
+      logLevel: 'silent',
+    });
+
+    const inputs = Object.keys(result.metafile.inputs);
+    HOST_BRANCH_INPUT_SUFFIXES.forEach((suffix) => {
+      expect(inputs.some((input) => input.endsWith(suffix))).toBe(true);
```
sidmorizon, line 71: # 🟡 P2 `throw new Error` 触发 lint CI 失败

## 问题

本行 `throw new Error(...)` 命中 OneKey 自定义 oxlint 规则 `onekey/no-raw-error`（选择器 `ThrowStatement > NewExpression[callee.name="Error"]`，见 `development/plugins/eslint-plugin-onekey.js`），报错 `Direct use of "throw new Error" is not allowed. Use OneKeyLocalError or OneKeyError instead.`。

该规则在 `.oxlintrc.json` 中全局为 `error`。豁免清单（`development/**`、`apps/*/e2e/**`、`apps/*/harness/**`、`apps/*/plugins/**`、`apps/desktop/native-modules/**`）并不包含 `apps/desktop/scripts/**`；测试豁免项是 `["test/**/*.js", "test/**/*.ts", "**/*.test.ts"]`，仅匹配 `*.test.ts` 而非本文件的 `*.test.js`。因此本文件不被豁免，规则照常生效。

此外，本 `throw` 位于 `catch (error)` 块内却未把 `error` 作为 `cause` 传递，正对应 CI 同时报出的那条 warning（“no cause error attached”）。这就是 `Found 1 warning and 1 error` 的来源。

## 影响

`lint (24.x)` 这个必需检查失败，阻塞合并。注意：`apps/ext/src/manifest/shared.js:40` 的 `throw new TypeError(...)` 不会被该规则命中（`callee.name` 为 `TypeError`，非 `Error`），所以唯一的 error 就是本行。

## 建议

改抛 `OneKeyLocalError` 并携带 `cause`，一并消除 error 与 warning：

```js
const { OneKeyLocalError } = require('@onekeyhq/shared/src/errors');
// ...
throw new OneKeyLocalError(
  `appEntry failed to bundle (Native Messaging host branch likely unresolvable):\n${details}`,
  { cause: error },
);
```

[Claude session ↗](https://claude.ai/code/session_01GBJLHH4ufYv6Gdq7RmfrMy)

---

sidmorizon: ✅ 已在 `57b317f3` 解决：该 `throw new Error` 已移除（去掉多余的 try/catch 重抛，直接 `await build(...)` 让 esbuild 自身的 `Could not resolve ...` 拒绝向上传播）。验证确认 smoke test 仍然有效——host 分支不可解析时 `build()` 仍会 reject 使测试失败，断言保持不变；whole-repo oxlint `onekey/no-raw-error` 不再有命中。本条评论指向已被取代的 `229a1b8`，可视为 outdated。

---

sidmorizon: Fixed in `57b317f3`. Took the simpler route of removing the redundant try/catch re-throw entirely rather than switching to `OneKeyLocalError` + `cause`: `await build(...)` lets esbuild’s own `Could not resolve …` rejection propagate and fail the test, which keeps the smoke check meaningful without importing `@onekeyhq/shared` into a build script. Lint now passes (oxlint `--deny-warnings`, 0 errors).

packages/kit-bg/src/desktopApis/DesktopApiKeychain.ts
sidmorizon: Confirmed the call-site analysis: `DesktopApiKeychain.setItem` now defaults `enableSync` to `false` explicitly (`?? false`), and every caller is explicit — `ICloudBackupProvider.ts:141` passes `enableSync: true` (so iCloud backup still syncs) and the dev test passes `enableSync: false`. No caller relies on the old implicit native `true`, so the wiring is safe. Reads are unaffected: `getItem` queries both local and iCloud-synced items, so keys previously written with sync stay queryable. The only thing left to validate in QA is the native re-write nuance (re-writing a key that was previously synced with `enableSync: false`); not exercised by current call sites. Resolving per PR owner.

apps/desktop/app/app.ts
```diff
@@ -1680,6 +1681,9 @@ if (!singleInstance && !process.mas) {
     logger.info(
       `nativeAppVersion: ${app.getVersion()}, buildNumber: ${process.env.BUILD_NUMBER ?? ''}, builtinBundleVersion: ${process.env.BUNDLE_VERSION ?? ''}`,
     );
+    void ensureDesktopNativeMessagingHostManifest().catch((error) => {
```
originalix: 这里是无条件安装 Native Messaging manifest，但这个进程前面已经在 `PERF_CI_MODE` / `DESKTOP_E2E_MODE` 下把 `userData` 改到了 perf profile 或临时目录。`nativeMessagingHostInstall.ts` 又把 launcher 固定写到 `app.getPath('userData')/native-messaging-hosts/...`，所以测试/性能模式也会去改写真实浏览器的 `NativeMessagingHosts` 注册。

这会导致两类确定性坏状态：E2E 路径里 `DESKTOP_E2E_USER_DATA_DIR` 会在 finally 里被删除，manifest 随后指向一个不存在的 launcher；perf 路径里跑的是 `build:main` 产物，而 `appEntry.ts` 在 production bundle 收到 `--onekey-native-messaging-host` 时会直接 `process.exit(0)`，浏览器只能拿到 no response。之后同一个 `so.onekey.wallet.desktop` host 名会一直坏到下一次正常 dev 启动把它覆盖回来。

建议这里对 `DESKTOP_E2E_MODE` / `PERF_CI_MODE` 做 hard skip，或者把 host launcher 落到独立且稳定的路径，并补对应清理逻辑。

sidmorizon: Added a hard-skip: the `ensureDesktopNativeMessagingHostManifest()` call on app ready now runs only when `!isPerfCiMode && !isDesktopE2EMode`, logging `[NativeMessagingHost] skip manifest install in perf/e2e mode` otherwise. So perf/E2E runs no longer touch the real browser’s NativeMessagingHosts registration — nothing is written to a redirected/temp userData, the E2E finally has nothing to leave dangling, and the perf production bundle never registers a host that would only exit(0). Only a genuine interactive dev run registers `so.onekey.wallet.desktop`, so no cleanup logic is needed.

sidmorizon: # 🟠 P1 确认：E2E 与 PERF 都会真实写入并污染浏览器 host 注册（含对 PERF 机制的修正）

## 问题

独立验证确认了该问题，并修正其中 PERF 的判断。`ensureDesktopNativeMessagingHostManifest()` 在 `app.on('ready')` 首行被无条件调用（`app.ts:1684`），而 `userData` 重定向发生在模块顶层（perf `app.ts:114`、e2e `app.ts:120`），早于 `ready`，故安装时 `userData` 已是 perf/临时目录。安装仅受 `isDesktopDevRuntime()`（`process.defaultApp===true`）与 `darwin && !process.mas` 两道门把关；`allowedExtensionIds.length` 早退不会触发（始终无条件加入两个 prod 扩展 ID）。

关键修正：PERF 并非 packaged 运行。`run-desktop-perf.js` 用 `require('electron')` spawn 未打包的 `appEntry.js`（`PERF_CI_MODE=1`），即 `process.defaultApp===true`，dev 门通过，manifest **确实会被写入**真实浏览器目录，而不是像原评论所设的「packaged → 跳过」。E2E 同理（`electron.launch` 未打包）。

## 影响

两种模式下 launcher 都落在 `app.getPath('userData')/native-messaging-hosts/...`（perf/临时目录），manifest 落在稳定的真实浏览器 `NativeMessagingHosts/so.onekey.wallet.desktop.json`。E2E 的 `finally` 删除 `userData` 后 manifest 指向已删除的 launcher；PERF 的 launcher 指向 production bundle，被 Chrome 拉起时 `process.exit(0)` 无响应。坏状态持续到下次正常 dev 启动覆盖为止。

## 建议

按原评论方向 hard-skip 即可，但务必覆盖 PERF（不要因「packaged 会跳过」而漏掉它）：在 `app.ts:1684` 调用前判断 `isPerfCiMode || isDesktopE2EMode`（两个 flag 已在 `app.ts:89-90` 存在）直接跳过；或将 launcher 落到独立稳定路径并补清理逻辑。

[Claude session ↗](https://claude.ai/code/session_01GBJLHH4ufYv6Gdq7RmfrMy)

---

sidmorizon: ✅ 已在 `babf7389` 验证修复正确。`app.ts:1692` 用 `if (isPerfCiMode || isDesktopE2EMode) { skip } else { install }` 把安装挡在 perf/e2e 之外——所用 flag 与上方 `userData` 重定向（`app.ts:114`/`120`）完全一致，**两种模式都覆盖**（关键的 PERF 也跳过了）；正常交互式 dev 运行仍照常注册（`nativeMessagingHostInstall.ts` 的 `process.defaultApp` + darwin 门未变）。无新增问题。

[Claude session ↗](https://claude.ai/code/session_01GBJLHH4ufYv6Gdq7RmfrMy)

---

packages/kit-bg/src/desktopApis/DesktopApiKeychain.ts
```diff
@@ -69,29 +79,25 @@ class DesktopApiKeychain implements IAppleKeyChainStorage {
...
   }
```
sidmorizon, line 101: **Keychain enableSync default silently changed from true to false**

Type: `Flag`

The `DesktopApiKeychain.setItem` method previously forwarded `enableSync` as-is to the native `keychainSetItem`, which the old JSDoc documented as defaulting to `true`. The new code at `packages/kit-bg/src/desktopApis/DesktopApiKeychain.ts:99` explicitly defaults it to `false` via `params.enableSync ?? false`. The only production caller (`packages/kit-bg/src/services/ServiceCloudBackupV2/backupProviders/ICloudBackupProvider.ts:141`) now explicitly passes `enableSync: true`, so current behavior is preserved. However, the iOS native path through `AppleKeyChainStorage` does NOT apply this default — it passes `enableSync` through as-is, meaning the iOS native module's own default (likely `true`) still applies when callers omit the parameter. This creates a platform asymmetry: omitting `enableSync` yields `false` on desktop but the native default (probably `true`) on iOS. Not a bug today since the only caller specifies it explicitly, but a latent footgun for future callers.

sidmorizon: Fixed the platform asymmetry: applied the same opt-in default (`enableSync ?? false`) in the shared `AppleKeyChainStorage.setItem`, so omitting `enableSync` now yields `false` on both desktop and iOS instead of falling back to the iOS native default. No current caller is affected (`ICloudBackupProvider` passes `true`, the dev test passes `false`); this just removes the footgun for future callers and matches the PR’s opt-in-sync direction. Updated the type JSDoc accordingly.

apps/desktop/app/appEntry.ts
```diff
@@ -0,0 +1,28 @@
...
+        runDesktopNativeMessagingHost(),
+      )
+      .catch(() => {
+        process.exit(1);
+      });
+  } else {
+    process.exit(0);
+  }
+} else {
+  void import('./app').catch((error) => {
```
originalix: 正常桌面启动现在是先进入 `appEntry.ts`，再通过异步 `import('./app')` 去加载真正的 main 逻辑。问题是 `app.ts` 里还有一批必须在主进程同步 top-level 就完成的注册：`requestSingleInstanceLock()`、`app.commandLine.appendSwitch(...)`、`app.userAgentFallback` 修补，以及 `will-finish-launching/open-url` 监听。

把整个普通启动路径放到 Promise microtask 之后，会让这些早期初始化失去 Electron 文档要求的同步保证，冷启动 deep link、协议处理、首个 renderer 的 UA 修补，甚至单实例收敛都可能错过第一次启动窗口。

这里更稳妥的修复方向是：保留普通桌面对 `./app` 的同步加载，只把 `--onekey-native-messaging-host` 这条 dev-only 分支单独做条件化延迟加载。

sidmorizon: # 🟠 P1 确认：appEntry 异步 `import('./app')` 让 app.ts 顶层同步注册晚一个 microtask

## 问题

独立验证确认这是本 PR 引入的回归。entry 从 base 的同步 `app.js` 改成了 `appEntry.js`（`apps/desktop/app/package.json` 的 `main`、`build.js` 的 `entryPoints` 均已改），普通启动经 `void import('./app')`（`appEntry.ts:24`）加载，使 `app.ts` 整个模块体延后一个 microtask 执行。`app.ts` 顶层确有需同步注册项：`will-finish-launching` 内注册 `open-url`（~`2501-2506`）、`requestSingleInstanceLock()`（~`1646`）、`userAgentFallback`（~`2189`）、`commandLine.appendSwitch`（~`2466`，dev-only）。

## 影响

真正有风险的是 **macOS 冷启动 deep link**：`open-url` 仅在 `will-finish-launching` 内注册，而后者触发很早，监听晚一个 microtask 可能错过冷启动 URL（`onekey-wallet://` / WalletConnect）→ 首次链接被丢。其余项风险低：`userAgentFallback` 在首个 webContents（`ready` 之后）才读取；`commandLine.appendSwitch` 在 `ready` 前仍足够早且为 dev-only；`requestSingleInstanceLock` 晚一个 microtask 不构成真实二实例竞态。

## 建议

按原评论方向：普通桌面路径保持**同步** `require('./app')`，只对 dev-only 的 `--onekey-native-messaging-host` 分支做条件化延迟加载。注意：仅把 `open-url` 提到顶层并**不能**修复——整个 `app.ts` 体仍晚一个 microtask；同步加载才是正解。

[Claude session ↗](https://claude.ai/code/session_01GBJLHH4ufYv6Gdq7RmfrMy)

---

sidmorizon: Fixed: normal desktop startup now loads `./app` synchronously via `require('./app')` (esbuild compiles it to a synchronous module initializer), so app.ts top-level Electron registrations — `requestSingleInstanceLock()`, `app.commandLine.appendSwitch(...)`, `app.userAgentFallback`, and the `will-finish-launching`/`open-url` listeners — run inside the synchronous startup window again. Only the dev-only `--onekey-native-messaging-host` branch stays lazily `import()`-ed, so its production DCE is unaffected (verified by the bundling smoke check, both cases green).

sidmorizon: ✅ 已在 `235bf798` 验证修复正确。普通路径改为 `else { require('./app') }`（`appEntry.ts:31-36`，CJS 同步初始化），`app.ts` 顶层注册（`requestSingleInstanceLock`/`userAgentFallback`/`commandLine`/`will-finish-launching`+`open-url`）重新在同步启动 tick 内执行——macOS 冷启动 deep link 风险消除。`require` 仅在 else 分支（host 模式不会 eval `./app`）；dev-only host 分支仍是受 `NODE_ENV` 守卫的 lazy `import('./nativeMessagingHost')`，prod DCE 不受影响（bundling smoke check 仍然有效）。无新增问题。

[Claude session ↗](https://claude.ai/code/session_01GBJLHH4ufYv6Gdq7RmfrMy)

---

apps/desktop/app/nativeMessagingHostInstall.ts
```diff
@@ -0,0 +1,302 @@
...
+  fs.mkdirSync(path.dirname(manifestPath), { recursive: true });
+  fs.writeFileSync(manifestPath, manifestContent);
+}
+
+export async function ensureDesktopNativeMessagingHostManifest() {
+  // Dev-only for now: don't touch user browser Native Messaging configs in
+  // packaged/production builds until a real consumer ships. See the header of
+  // @onekeyhq/shared/src/consts/desktopNativeMessaging for the security model
+  // (same-user host impersonation risk) and the production checklist.
+  if (!isDesktopDevRuntime()) {
```
originalix: 当 production main 是通过 `electron app/dist/appEntry.js` 这类方式启动时，这里只检查 `process.defaultApp`，仍然会把 `so.onekey.wallet.desktop` 注册进真实浏览器的 `NativeMessagingHosts`。

但 `appEntry.ts` 的 production 分支已经把 host 入口裁掉了，浏览器后面再用 `--onekey-native-messaging-host` 拉起它时只会直接退出，结果就是本机留下一个不可用的 host 注册，直到下一次真正的 dev 启动把它覆盖掉。

这里的安装条件需要和 host 入口复用同一套“可服务”判断，至少再加上 `process.env.NODE_ENV !== 'production'`，或者抽成一个 shared helper，避免 production smoke run / bench run 把浏览器注册写坏。

apps/desktop/app/nativeMessagingHost.ts
```diff
@@ -0,0 +1,325 @@
...
+
+  let callerExtensionId: string;
+  try {
+    callerExtensionId = getAllowedNativeMessagingCallerExtensionId();
+  } catch (error) {
+    writeNativeMessage({
+      success: false,
+      error: normalizeError(error),
+    });
+    app.exit(1);
```
originalix, line 285: 当调用方扩展不在 allow-list 时，这里先 `writeNativeMessage()` 再立刻 `app.exit(1)`。Native Messaging 的响应是走 stdout pipe 回传的，进程马上退出会让这帧错误包来不及 flush，扩展侧通常只会收到通用的 `host unavailable/exited`，拿不到本来应该区分出来的 `NATIVE_HOST_FORBIDDEN`。

同样的问题在后面的超长消息分支也存在：先写回错误，再同步退出，错误码就可能在退出过程中被截断。

建议把退出放到 stdout 写完成之后，再 `end/exit`，或者让 `writeNativeMessage` 返回一个 Promise 并在 `await` 完成后结束进程。否则调试页和后续调用方都会把协议错误误判成 host 不可用。

apps/desktop/app/nativeMessagingHostInstall.ts
sidmorizon: Fixed: extracted a shared `isDesktopNativeMessagingHostServiceable()` (`process.env.NODE_ENV !== 'production' && process.defaultApp === true`) and the manifest install now gates on it instead of bare `process.defaultApp`. So an unpackaged production smoke/bench run no longer registers a host that appEntry has already DCE-d. The host runtime guard uses the same helper so the two cannot drift; appEntry keeps the inline NODE_ENV literal because esbuild needs a literal to eliminate the host import in production.

apps/desktop/app/nativeMessagingHost.ts
sidmorizon: Fixed: `writeNativeMessage` now returns a Promise that resolves only after the stdout write completes (via the write callback). The forbidden-caller path `await`s it before `app.exit(1)`, and the oversized-message path (a synchronous stdin handler) chains `.finally(() => app.exit(1))`, so the error frame is flushed before exit and the extension receives the specific code (e.g. NATIVE_HOST_FORBIDDEN) instead of a generic "host exited". The normal reply paths await it too.

PR #12147, opened 2026-06-24
fix: re-validate on-chain nonce before replacing pending txs to avoid 40024 (OK-55650)

weatherstar:
[OK-55650](https://onekeyhq.atlassian.net/browse/OK-55650)

----------

## Summary
- Re-validate a replace (speed up / cancel) transaction's reused nonce against the current on-chain nonce **before signing/broadcasting**, so an already-confirmed/replaced tx is no longer sent to the backend (which rejects it with code `40024` "nonce already used").
- Add a backend `40024` safety net: clean up the stale local pending tx and surface a friendly, localized message instead of the raw backend error.
- Centralize the protection at the single bg submit chokepoint (`batchSignAndSendTransaction`), covering all three replace entry points (SendReplaceTx page, SendConfirm, SignatureConfirm).

## Intent & Context
Users frequently hit a backend error when speeding up / canceling a pending EVM transaction:

> `40024` — Transaction failed: Nonce (0) already used, please reconstruct the transaction. (Next valid nonce is 1)

This is **not** a "use the next nonce" situation — speed up / cancel must reuse the original pending tx's nonce. The error means that nonce slot was already consumed by an on-chain confirmed tx, so the original tx can no longer be replaced. Continuing to sign & broadcast just fails deterministically, and the raw (technical) backend message was shown to the user.

## Root Cause
The whole replace flow used the **locally cached** nonce end-to-end, with **no on-chain re-validation** anywhere from decision to broadcast:
- `useReplaceTx.canReplaceTx` only checks local `status === Pending` + `replaceTxEnabled` + `canAccelerateTx`.
- `evm/Vault.canAccelerateTx` only compares against the **local** min pending nonce (`getLocalHistoryMinPendingNonce`), never the chain.
- `evm/Vault.buildReplaceEncodedTx` freezes `decodedTx.nonce` into the replacement at button-press time.
- `SendReplaceTxContainer.handleOnConfirm` rebuilds the unsignedTx, updates fee and broadcasts — without re-checking the nonce.

Local pending cleanup depends on history refresh matching an on-chain confirmed/same-nonce tx (`ServiceHistory` ~L2087). When the indexer lags, the user replaces via an external wallet, or the UI is faster than the refresh, the nonce is already consumed on-chain but the tx still shows as Pending locally → the button stays active → the broadcast hits `40024`. There is also a race window between opening the fee page (nonce frozen) and confirming (original tx may get mined in between).

## Design Decisions
- **Authoritative check lives in bg, at the last moment before signing** (`batchSignAndSendTransaction`, guarded by `replaceTxInfo && !isMultiTxs`). This closes the race tightest and covers all three replace entry points with one change, instead of patching each UI. In production (native), `main` holds a stale deserialized copy of localHistory, so the on-chain nonce check must run in `bg`.
- **Validate only, never overwrite the nonce.** Overwriting with the on-chain next nonce would turn a replace into a brand-new transaction. The check only aborts when `targetNonce < onChainNextNonce`; equal (still pending) or a nonce gap (`>`) is allowed.
- **Fail-open pre-check.** A flaky `fetchAccountDetails` call never blocks a legitimate replace; the backend `40024` catch is the safety net.
- **Button-display check kept local-only.** `fetchAccountDetails` is uncached, so checking on-chain nonce on every pending-row render would cause request storms / button flicker / rate-limiting. The on-chain check runs only on click and just before broadcast (both low-frequency, user-initiated).
- **i18n reuse instead of a new key.** Surfaces the existing `global_nonce_error_lower` ("Nonce has been used and may cause this transaction to fail"), already translated in all locales. `yarn i18n:add` writes to the shared Lokalise project (external side-effect), so a dedicated key can be added later via the normal i18n flow if a more precise wording is desired.

## Changes Detail
- `packages/kit-bg/src/services/ServiceSend.ts`
  - `precheckReplaceTxNonceConsumed()` — on-chain nonce re-validation (nonceRequired-gated, fail-open).
  - `isReplaceTxNonceAlreadyUsedServerError()` / `buildReplaceTxNonceConsumedError()` — backend `40024` detection + localized business error.
  - `batchSignAndSendTransaction()` — pre-broadcast nonce check + `40024` catch → cleanup + friendly error.
- `packages/kit-bg/src/services/ServiceHistory.ts`
  - `checkReplaceTxAvailability()` — local checks + on-chain nonce, returns `{ available, reason, onChainNextNonce }`.
  - `resolveStalePendingReplaceTx()` — reuses `clearLocalHistoryPendingTxByTxId` to drop the stale pending tx and emit `HistoryTxStatusChanged`.
- `packages/kit/src/hooks/useReplaceTx.ts` — click-time availability check; on `nonceConsumed`, repair local state + toast and do not open the fee page.
- `packages/kit/src/views/Send/pages/SendReplaceTx/SendReplaceTxContainer.tsx` — `handleOnConfirm` detects the business error (className survives the bg→main bridge) and closes the page.
- `packages/shared/src/errors/{types/errorTypes.ts,errors/appErrors.ts}` — new `ReplaceTxNonceConsumed` class name + `ReplaceTxNonceConsumedError`.
- `packages/shared/src/engine/engineConsts.ts` — `SEND_TX_SERVER_ERROR_CODES.NONCE_ALREADY_USED = 40024`.
- `packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts` — unit tests for the nonce comparison + fail-open + `40024` detection.

## Risk Assessment
- **Risk Level**: Low–Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (EVM replace flow)
- **Risk Areas**: The added pre-check runs inside `batchSignAndSendTransaction` for replace txs only (`replaceTxInfo && !isMultiTxs`); normal sends and batch flows are unchanged. The sign/send call is now wrapped in try/catch — behavior is identical except for the replace `40024` branch (re-throws all other errors). Adds one extra `fetchAccountDetails` round-trip per replace submit (user-initiated, low frequency).

## Test plan
- [x] `yarn tsc:staged` passes
- [x] `yarn lint:staged` passes
- [x] Unit tests pass: `yarn test packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts` (consumed / pending / nonce-gap / non-nonce chain / fail-open / 40024 detection)
- [ ] Manual: with a pending EVM tx whose nonce is already confirmed on-chain (e.g. mock `fetchAccountDetails` nonce > local pending nonce), tap Speed up / Cancel → friendly message shown, stale pending cleared, buttons disappear, no doomed broadcast.
- [ ] Manual: a genuinely pending EVM tx (on-chain nonce not advanced) can still be sped up / canceled normally.
- [ ] Manual: verify all three entry points (SendReplaceTx page, SendConfirm, SignatureConfirm).

[OK-55650]: https://onekeyhq.atlassian.net/browse/OK-55650?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3a5d9cbc-389b-4a4b-897e-a64ff3990e06?refs=e3ee2ebd257cae450a945ca68c980b0b5d5410b4&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/3a5d9cbc-389b-4a4b-897e-a64ff3990e06/license?refs=e3ee2ebd257cae450a945ca68c980b0b5d5410b4&source=prChecksComment) |

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -629,6 +636,29 @@ class ServiceSend extends ServiceBase {
...
+        });
+        if (consumed) {
+          await this.cleanupStaleReplaceTxAndThrow({
+            accountId,
+            networkId,
+            replaceHistoryId: replaceTxInfo.replaceHistoryId,
+          });
+        }
+      }
+    }
```
sidmorizon, line 661: **IUnsignedTxPro.nonce may not be populated for EVM replace txs**

Type: `Flag`

The pre-broadcast nonce check at `packages/kit-bg/src/services/ServiceSend.ts:645` reads `unsignedTxs[0]?.nonce` from `IUnsignedTxPro`, which is `number | undefined`. For EVM replace txs, the nonce lives in the `encodedTx` (hex-encoded), and may or may not be mirrored to `IUnsignedTxPro.nonce` depending on vault implementation. If the vault doesn't populate it, the pre-check is silently skipped. This is acceptable due to the fail-open design (the backend 40024 catch at line 701 is the safety net), but worth being aware that the pre-check's effectiveness depends on vault behavior.

weatherstar: Good flag. I traced the actual EVM replace-tx path and the nonce is reliably populated on `IUnsignedTxPro`:

- `buildReplaceEncodedTx` → `buildEstimateTxForSpeedUpOrCancelTx` returns an `IEncodedTxEvm` with `nonce` set as a number (it throws `speedUpOrCancelTx ERROR: nonce is missing!` if absent — `evm/Vault.ts:1430`).
- `SendReplaceTxContainer` then calls `buildUnsignedTx({ encodedTx: replaceEncodedTx })`, and EVM `_buildUnsignedTxFromEncodedTx` mirrors `encodedTx.nonce` → `unsignedTx.nonce` (`evm/Vault.ts:1048`).
- `updateUnSignedTxBeforeSending` only attaches fee info, so it does not clear the nonce.

Since `replaceTxInfo` is only set on this EVM replace flow, `unsignedTxs[0].nonce` is present when the pre-check runs, so the pre-check is effective. And as you noted, even in the theoretical case where a vault did not mirror it, the fail-open design + the backend 40024 catch remains the safety net. Keeping as-is.

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -629,6 +636,29 @@ class ServiceSend extends ServiceBase {
...
+          targetNonce: replaceTargetNonce,
+        });
+        if (consumed) {
+          await this.cleanupStaleReplaceTxAndThrow({
+            accountId,
+            networkId,
+            replaceHistoryId: replaceTxInfo.replaceHistoryId,
+          });
+        }
+      }
```
sidmorizon, line 660: **Three-layer defense-in-depth creates multiple network round-trips for replace flows**

Type: `Flag`

The replace flow now performs up to 3 on-chain nonce fetches: (1) `checkReplaceTxAvailability` in `useReplaceTx.ts:117`, (2) the pre-broadcast check in `batchSignAndSendTransaction` at line 647, and (3) the backend broadcast itself which may return 40024. While this is intentional (defense-in-depth with fail-open), for users on slow connections the latency adds up. The first check (UI-level) and second check (pre-sign) both call `precheckReplaceTxNonceConsumed` → `fetchAccountDetails`. Consider whether the second check could be skipped if the first check was performed recently (within a few seconds), or whether a single cached nonce value could be threaded through.

weatherstar: Good observation on the latency tradeoff. We are intentionally keeping all three checks because each one closes a different failure window:

- **(1) `checkReplaceTxAvailability` (UI, `useReplaceTx.ts:117`)** runs when the user opens the replace flow — it gates the UI early so we never even present a replace form for an already-consumed nonce.
- **(2) Pre-broadcast check in `batchSignAndSendTransaction`** runs immediately before signing/broadcasting. The gap between (1) and the actual broadcast can be arbitrarily long (user reviews fee, edits, gets distracted), and a competing tx can land in that window — so a freshly-fetched nonce right before broadcast is the only thing that catches the late race.
- **(3) Backend 40024** is the ultimate fail-open safety net for the case where even (2) missed the race (network reorg, propagation delay).

On the optimization itself: threading a cached nonce from (1) into (2), or skipping (2) when (1) was recent, would reintroduce exactly the stale-read race (2) exists to catch — the value of (2) is precisely that it is *not* cached. The extra `fetchAccountDetails` is a single lightweight read, and replace/speed-up is a low-frequency, deliberate user action (not a hot path), so we are favoring correctness/safety over shaving one round-trip here. Keeping as-is.

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -806,6 +857,89 @@ class ServiceSend extends ServiceBase {
...
+      if (isNil(targetNonce)) {
+        return { consumed: false };
+      }
+      const { nonceRequired } =
+        await this.backgroundApi.serviceNetwork.getVaultSettings({ networkId });
+      if (!nonceRequired) {
+        return { consumed: false };
+      }
+      const { nonce: onChainNextNonce } =
+        await this.backgroundApi.serviceAccountProfile.fetchAccountDetails({
```
originalix: 这里的 nonce precheck 通过 `serviceAccountProfile.fetchAccountDetails()` 读的是 wallet API 视图，但实际发送时如果用户启用了 custom RPC，会先在节点侧执行广播。

一旦 API 的 nonce 落后于当前 RPC 节点，这里就会放行，而下面的兜底只识别后端 `40024`，接不住 custom RPC 返回的 nonce-too-low / replacement-underpriced 一类错误。结果是用户仍然会看到原始 RPC 失败，本地 stale pending 也不会被清掉。

建议在 custom RPC 启用时改成从同一个 RPC 源读取 nonce，让 precheck 和实际广播共享同一个状态视图。

weatherstar: 已修复（a09f45ba5c）。结论确认：内置网络 + 启用 custom RPC 时，`vault.fetchAccountDetails` 只在 `isCustomNetwork` 时才走 RPC，所以 precheck 读的是 wallet API，而广播走的是 custom RPC 节点，两者 nonce 视图可能不一致；兜底又只识别后端 40024，接不住节点抛的 `nonce too low`，stale pending 不会被清掉。

改动：
- `precheckReplaceTxNonceConsumed` 现在从「与广播相同的来源」读 nonce：当 custom RPC 启用且未选 default RPC 时，走 `vault.fetchAccountDetailsByRpc`（同一节点），否则维持 wallet API（完全自定义网络本就已经走 RPC）。`useDefaultRpc` 已从 `batchSignAndSendTransaction` 透传进来，让 precheck 与广播的 RPC 选择一致。custom RPC 检测做了防御：解析失败时回退到 wallet API，而不是让整个 precheck fail-open。
- 额外加了一层兜底 `isReplaceTxNonceAlreadyUsedRpcError`，覆盖 precheck 与广播之间的残余竞态：custom RPC 广播抛出的原始 `nonce too low` 也会触发 stale pending 清理 + 友好提示。

一个语义上的取舍：**只**把 `nonce too low` 当作「nonce 已消费」处理，故意排除了 `replacement transaction underpriced`——后者表示原 tx 仍 pending、只是替换费率不够，此时绝不能清掉本地 pending。

已补单测覆盖：custom RPC 读 RPC nonce、useDefaultRpc=true 回退 wallet API、custom RPC lookup 抛错回退、以及 RPC 错误匹配器对 `nonce too low` 命中 / 对 `replacement underpriced` 不命中。

packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts
```diff
@@ -0,0 +1,324 @@
...
+  const fetchAccountDetailsByRpc =
+    overrides.fetchAccountDetailsByRpc ??
+    jest
+      .fn()
+      .mockResolvedValue({ data: { data: { nonce: overrides.rpcNextNonce } } });
+  const getCustomRpcForNetwork =
+    overrides.getCustomRpcForNetwork ??
+    jest.fn().mockResolvedValue(overrides.customRpcInfo);
+
+  (vaultFactory.getVault as jest.Mock).mockResolvedValue({
```
sidmorizon: # 🟡 P2 测试中的类型断言导致 tsc 编译失败、阻塞 CI

## 问题

`(vaultFactory.getVault as jest.Mock)` 把 memoized 的 `getVault` 直接断言成 `jest.Mock`。`getVault` 的类型是 `((opts) => Promise<VaultBase>) & Memoized<...>`，与 `jest.Mock` 没有足够重叠，tsc 直接报错：`Conversion of type ... may be a mistake ... If this was intentional, convert the expression to 'unknown' first.`。CI `lint(24.x)` 已因此红灯。

## 影响

这是本 PR 新增测试代码里的真实编译错误，会阻塞 CI 合并门禁；在该断言修复前，`tsc:staged` / lint 流水线无法通过。

## 建议

按 tsc 提示先经由 `unknown` 再断言：

```ts
(vaultFactory.getVault as unknown as jest.Mock).mockResolvedValue({
  fetchAccountDetailsByRpc,
});
```

[Claude session ↗](https://claude.ai/code/session_01SDzYmuMhUPqazdYUDRhUMA)

---

sidmorizon (commented):
Automated review — see inline comment.

---

packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts
weatherstar: 已修复（ed543c9c55）。按提示先经由 `unknown` 再断言，与本文件已有的 `ServiceSend as unknown as ...` 写法保持一致：

```ts
(vaultFactory.getVault as unknown as jest.Mock).mockResolvedValue({
  fetchAccountDetailsByRpc,
});
```

本地已用 CI 同款的 type-aware `oxlint`（`yarn lint:staged`）复测：0 warnings / 0 errors；单测 18/18 仍通过。

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -806,6 +859,178 @@ class ServiceSend extends ServiceBase {
...
+          networkId,
+        });
+      const resp = await vault.fetchAccountDetailsByRpc({
+        accountId,
+        networkId,
+        accountAddress,
+        withNonce: true,
+      });
+      const rpcNonce = resp?.data?.data?.nonce;
+      return isNil(rpcNonce) ? undefined : rpcNonce;
```
sidmorizon, line 967: **Custom RPC response shape for nonce fetch is deeply nested and unverifiable without vault implementation**

Type: `Flag`

At `packages/kit-bg/src/services/ServiceSend.ts:961`, the custom RPC nonce is read via `resp?.data?.data?.nonce` — triple nesting that resembles an axios-style raw HTTP response wrapper. In contrast, the wallet API path at line 965-971 returns `{ nonce }` directly. While the test mock at `packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts:74-76` matches the triple-nested shape, the actual vault `fetchAccountDetailsByRpc` implementation was not inspected. If the real return shape differs, the nonce extraction silently yields `undefined`, triggering the fail-open path (`consumed: false`). This would make the custom-RPC precheck silently ineffective — the backend-40024 safety net still catches it, but it defeats the purpose of the precheck for custom-RPC users.

weatherstar: 已核实，返回结构匹配，无需改动。

`fetchReplaceTxOnChainNextNonce` 的 custom-RPC 分支只会在 replace-tx 流程触发，而 `replaceTxInfo` 仅由 EVM replace 流程设置，所以这里实际命中的 vault 永远是 EVM。追到真实实现：

- EVM `fetchAccountDetailsByRpc` (`evm/Vault.ts:1547`) → `provider.getAccount(params)`
- `BaseApiProvider.getAccount` (`sdkEvm/BaseApiProvider.ts:176`) 末尾固定返回 `{ data: { data: res } }`，其中 `res.nonce` 由 `getNonce` 在 `withNonce: true` 时填充。

因此 `resp?.data?.data?.nonce` 的三层嵌套与真实返回结构逐层对齐，nonce 提取正确，不会静默落到 fail-open 分支。测试 mock 的 `data.data.nonce` 形状也正是照此实现编写。Flag 提出时未检视 vault 实现，现已核实，保持现状。

zhaono1 (approved):

originalix (approved):

PR #12149, opened 2026-06-24
fix: prevent partial Prime Transfer import on duplicate trigger (OK-56787)

sidmorizon:
## Background — OK-56787

Reported: *偶现桌面端传输导入数据不完整* — the desktop (receiver) shows "transfer complete" almost instantly, the page stays stuck on **"传输中"**, and after restart only **2 of 5** wallets (out of 100+ private-key/watch accounts) were imported. Reproduced once; could not be reproduced by hand.

## Root cause (from the attached log `app-latest.log`, session 20:25–20:26)

It is **not** a network drop (no socket error/disconnect during the import). Comparing the day's 5 imports:

| time | `startImport` calls | duration | result |
|---|---|---|---|
| 18:56 / 19:15 / 19:44 | 1 | ~80s | OK |
| **20:26:12** | **2 (same second)** | **~2s** | ❌ data loss |
| 20:29 | 1 | ~82s | OK (recovery) |

The user's local password could not decrypt the payload, so the flow went through the **"对方设备密码" (remote device password) dialog**. That dialog wires the *same* handler to **two** entry points — the confirm button (`onConfirm`) **and** the input's `onSubmitEditing` (Enter) — guarded only by a 300ms `leading` debounce. Submitting it twice (Enter + click) just past the debounce window fired the import flow twice:

1. `startImport` had **no re-entrancy guard**: it just overwrote the shared `currentImportTaskUUID`. The 2nd call took over the UUID, so the 1st (real) import loop hit its `currentImportTaskUUID !== taskUUID` checkpoint and **self-cancelled**.
2. The 1st flow's `completeImportProgress` declared success and its **debounced `finallyImportProgress` set `currentImportTaskUUID = undefined`**, which then aborted the *2nd* loop too — after only ~2 wallets were written.
3. Both flows ended `success:false`, so `exitTransferFlow()` was never reached → page stuck on **"传输中"**.

This is timing-dependent (needs two near-simultaneous submits), which is why it's intermittent and hard to reproduce manually.

## Fix — make the whole import lifecycle task-owned

- **`startImport`** — synchronous re-entrancy guard: if an import already owns `currentImportTaskUUID`, reject the duplicate **before any await/state mutation** and return `{ skipped: true }`. Returns the owning `taskUUID`.
- **`completeImportProgress`** — ownership guard: only finalize when the caller's `taskUUID` matches the active one. A superseded/duplicate flow can **no longer** clear the running task's UUID and abort the real import. *(This is the part that makes the re-entrancy guard actually sufficient — guarding `startImport` alone is not enough, because the duplicate flow's `completeImportProgress`→`finallyImportProgress` would still clear the UUID.)*
- **`PagePrimeTransferPreview` / `ServiceCloudBackupV2`** — thread `taskUUID` from `startImport` into `completeImportProgress`, and add a synchronous `useRef` guard on the page so a fast double-submit can't even start a second flow (React `useState` updates async and can't block it).

Net effect: no matter how many times the UI is double-/mis-triggered, **at most one import runs to completion** and the running import can't be aborted by a duplicate.

## Files
- `packages/kit-bg/src/services/ServicePrimeTransfer/ServicePrimeTransfer.ts`
- `packages/kit-bg/src/services/ServiceCloudBackupV2/ServiceCloudBackupV2.ts`
- `packages/kit/src/views/Prime/pages/PagePrimeTransfer/PagePrimeTransferPreview.tsx`

## Testing notes
- `tsc`/lint could not be run in this environment (no `node_modules`); please run `yarn tsc:staged` & `yarn lint:staged` in CI/local.
- Suggested manual repro/verification: on the receiver, force the data through the **remote-password** path, then submit that dialog twice quickly (press Enter **and** click 确认). Before: ~2s "complete" with partial wallets + stuck UI. After: exactly one import runs to completion; the duplicate submit is a no-op.
- Minor known trade-off: a brand-new import started within ~1.5s of a previous one finishing may be rejected until the debounced finalize clears the task; in the real flow the page navigates away on success, so this isn't user-visible.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vf3uP2PYimnFFAebQa6yPj)_


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2c276478-c565-4d8e-a0a5-240b3c3c68f6?refs=73023c177c3651bc3d30f0c6611601f33acd3de5&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/2c276478-c565-4d8e-a0a5-240b3c3c68f6/license?refs=73023c177c3651bc3d30f0c6611601f33acd3de5&source=prChecksComment) |

originalix (approved):

PR #12117, opened 2026-06-20
fix: prevent Wallet Home cold-start blank state (OK-56734)

originalix:
[OK-56734](https://onekeyhq.atlassian.net/browse/OK-56734)

----------

## Summary
Issue: https://onekeyhq.atlassian.net/browse/OK-56734

- Add real AccountSelector storage/active-account initialization gates before showing the Wallet Home no-wallet empty state.
- Keep the current usable active account when cold-start or sync briefly emits an incomplete selected account without wallet/account identity.
- Add focused unit tests for no-wallet visibility and incomplete selected-account guarding.

## Intent & Context
This hotfix targets the iOS 6.4 cold-start Wallet Home regression. The reported behavior was: after killing and reopening the app with an existing wallet, Wallet Home first flashed the no-wallet empty state, then later showed the normal wallet page. After the first mitigation, iOS still showed a visible loading/blank sequence and the Wallet Home appeared to shrink upward after wallet data became available.

The expected behavior is that existing-wallet users must not see the no-wallet empty state during cold start. A real no-wallet user should still see the no-wallet entry after AccountSelector initialization and wallet-list resolution complete.

## Root Cause
`accountSelectorStorageReadyAtom` uses cold-start cache, so it can be restored as ready before the current launch has actually hydrated selected accounts from storage. During cold start, `buildActiveAccountInfoFromSelectedAccount` can also build a `ready: true` active account from default or incomplete selected-account values.

Frame-by-frame debugging showed an additional transient state after storage init: selected account briefly carried only `networkId` and no wallet/account identity. That overwritten the existing usable active account with an active account that had `network` but no `wallet` or `account`. Home then entered a blank branch, native header layout was remeasured shorter, and the wallet page visually contracted before recovering.

## Design Decisions
- Add a non-cached `accountSelectorStorageInitDoneAtom` so Home does not rely only on the cached ready flag.
- Add `accountSelectorActiveAccountInitDoneAtom` so no-wallet UI is gated until active-account reload has completed for the Home selector.
- Keep Home no-wallet semantics unchanged: `NoWalletContent` only renders when `hasNoUsableWallet` is true and both AccountSelector init plus wallet-list resolution confirm the user has no usable wallet.
- Guard active-account writes at the source: if selected account is default before storage init, or if it is network-only while a usable active account already exists, keep the current active account instead of writing an incomplete one.

## Changes Detail
- `packages/kit/src/states/jotai/contexts/accountSelector/atoms.ts`: adds non-cached storage and active-account init signals.
[...]

PR #12154, opened 2026-06-25
fix: keep restored wallet account on cold start

originalix:
## Summary
- Preserve the restored wallet/indexed account selection during Wallet cold start when `activeAccount` is still temporarily incomplete.
- Validate the restored indexed account before falling back to the first account in the wallet.
- Add regression coverage for restored Account #2 staying selected across incomplete active-account hydration states.

## Intent & Context
PR #12117 fixed a cold-start flash where Wallet could briefly show the empty/no-wallet page. After that fix, users reported a new regression: the Wallet home page no longer showed the no-wallet state, but the selected account could jump on every app open or tab return. A production-package repro loop showed that after selecting `Seed3 / Account #2`, cold start usually displayed Account #2 briefly and then switched to Account #1.

The goal of this change is to keep the user's persisted wallet account selection stable during cold start, without weakening the existing auto-select fallback behavior for invalid or removed accounts.

## Root Cause
During cold start, `selectedAccountsAtom` can already contain the persisted Account #2 selection from storage, while `activeAccountsAtom` may still be partially hydrated and missing `wallet` or `indexedAccount`. `autoSelectNextAccount` previously treated the missing active indexed account as a reason to query all indexed accounts for the wallet and unconditionally pick `accounts[0]`. For HD wallets, that means a valid restored `hd-1--1` selection could be overwritten by `hd-1--0`, causing Account #2 to jump back to Account #1.

This was diagnosed from repeated desktop cold-start reproduction and then reduced into a unit regression test where `selectedAccount.indexedAccountId` is `hd-1--1` but the active account is temporarily incomplete.

## Design Decisions
- Prefer the already restored `selectedAccount.walletId` and `selectedAccount.indexedAccountId` when the active account is not fully hydrated yet.
- If the active wallet object is missing but the restored wallet id exists, load it with `getWalletSafe` so deprecated/mock wallet checks still run against a real wallet object.
- Keep the existing fallback semantics: if the restored indexed account no longer exists in the wallet, fall back to the first indexed account as before.
- Keep the fix scoped to account selection state reconciliation rather than adding UI delays or masking the symptom in Wallet home rendering.

## Changes Detail
- `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx`
  - Uses restored selected-account wallet/indexed-account values as fallbacks while active account hydration is incomplete.
  - Validates the restored indexed account belongs to the current wallet before preserving it.
  - Falls back to the first indexed account only when the restored indexed account is unavailable.
- `packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx`
  - Adds serviceAccount mocks needed by `autoSelectNextAccount`.
  - Adds regression tests for incomplete active account state, missing active wallet state, and removed restored indexed account fallback.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Mobile / Web / Extension
- **Risk Areas**: Account selector auto-selection is shared state logic. The risky path is preserving a restored indexed account that may no longer be valid, so the implementation explicitly checks the indexed account list and keeps the existing fallback to the first available indexed account.

## Test plan
- [x] `yarn lint --fix`
- [x] `yarn tsc:staged`
- [x] `yarn jest packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx --runInBand --modulePathIgnorePatterns='<rootDir>/apps/desktop/app' --testPathIgnorePatterns='<rootDir>/apps/desktop/app'`
- [x] `yarn workspace @onekeyhq/desktop build:JsBundle`
- [x] Verified the new regression test fails before the fix with Account #2 being overwritten by Account #1, then passes after the fix.
- [x] Attempted production app runtime injection for `/Applications/OneKey.app`; macOS denied direct `app.asar` replacement with `Operation not permitted`, and the original installed app remained unchanged with `codesign --verify --deep --strict /Applications/OneKey.app` passing.


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/95e23359-dc63-49ea-b04f-870412816129?refs=8d438a1507a403b013c363d906548de7f8c9c1d3&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/95e23359-dc63-49ea-b04f-870412816129/license?refs=8d438a1507a403b013c363d906548de7f8c9c1d3&source=prChecksComment) |

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -2815,9 +2815,18 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
           let selectedWallet = wallet;
-          let selectedIndexedAccountId = indexedAccount?.id;
+          if (!selectedWallet && selectedWalletId) {
+            selectedWallet = await serviceAccount.getWalletSafe({
+              walletId: selectedWalletId,
+            });
+            if (!selectedWallet) {
+              selectedWalletId = undefined;
+            }
+          }
```
sidmorizon, line 3220: **Behavioral change: wallet search loop may now be skipped when active wallet is transiently missing**

Type: `Flag`

In the old code, when `wallet` from the active account was `undefined`, `selectedWalletId` was also `undefined`, causing the wallet search loop (lines 2844–2908) to always run and scan for a replacement HD/HW/QR wallet. In the new code (`actions.tsx:2818`), `selectedWalletId` falls back to `selectedAccount?.walletId`, and if `getWalletSafe` at `actions.tsx:2821` successfully fetches that wallet, the search loop is skipped entirely. This means the code now preserves the user's wallet selection during transient states instead of potentially switching to a different wallet. This is the intended fix, but it changes observable behavior: previously a missing active-account wallet always triggered a full wallet scan; now it only does so if the fetched wallet is also invalid/deprecated. Reviewers should confirm this is acceptable for all callers of `autoSelectNextAccount`.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -2908,14 +2936,28 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
           // auto select hd or hw index account
           if (selectedWalletId && (isHdWallet || isHwOrQrWallet)) {
             if (
+              !selectedAccountNew.walletId ||
+              !selectedAccountNew.indexedAccountId ||
+              !selectedAccountNew.focusedWallet ||
               !indexedAccount ||
               indexedAccount.walletId !== selectedWalletId
```
sidmorizon, line 3317: **New OR conditions widen the indexed account selection entry point**

Type: `Flag`

The added conditions at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:2939-2941` (`!selectedAccountNew.walletId || !selectedAccountNew.indexedAccountId || !selectedAccountNew.focusedWallet`) significantly widen when the indexed-account selection block executes. Previously it only entered when `!indexedAccount || indexedAccount.walletId !== selectedWalletId`. Now it also enters when the cloned selected account is missing any of those three fields. This means that even when `indexedAccount` is present and matches the wallet, if `selectedAccountNew` (cloned from the atom) lacks `focusedWallet`, the block will re-fetch indexed accounts and overwrite the selection. This is likely intentional for cold-start recovery, but it changes behavior for any scenario where `focusedWallet` is transiently undefined.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
originalix: Confirmed, this behavior change is intentional for the cold-start recovery path.

When the active account wallet is transiently missing, falling back to the restored `selectedAccount.walletId` prevents `autoSelectNextAccount` from scanning to another wallet/account and overwriting the restored selection. The scan path is still preserved when the restored wallet cannot be loaded, is deprecated/mocked, or has no indexed accounts, so invalid/deleted wallet cases still fall back normally.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
originalix: Confirmed, the widened condition is intentional to normalize incomplete cold-start selections.

It does not blindly overwrite with the first indexed account: the code first tries `selectedAccountNew.indexedAccountId`, then the active indexed-account / selected-account indexed id, and only falls back to the first indexed account when that restored id is no longer present. The added tests cover the transiently incomplete active account/wallet cases, network-only cold-start restoration, and the deleted-indexed-account fallback.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -603,13 +1083,49 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
+            ...(before || defaultSelectedAccount()),
+            networkId,
+          };
       await this.updateSelectedAccount.call(set, {
         num,
         builder: (v) => ({
           ...v,
-          networkId,
+          ...resolvedSelectedAccount,
         }),
```
sidmorizon: **Builder pattern change in updateSelectedAccountNetwork may overwrite concurrent field updates**

Type: `Flag`

The old `updateSelectedAccountNetwork` used `builder: (v) => ({ ...v, networkId })`, which only modified `networkId` and preserved the latest atom state for all other fields. The new code at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1114-1119` uses `builder: (v) => ({ ...v, ...resolvedSelectedAccount })` where `resolvedSelectedAccount` (for non-others wallets) is `{ ...before, networkId }` — a full snapshot read OUTSIDE the `mutexUpdateSelectedAccount`. If a concurrent `updateSelectedAccount` completes between reading `before` (line 1087) and acquiring the mutex inside `updateSelectedAccount`, the builder would spread stale field values (e.g. `deriveType`, `focusedWallet`) over the fresh `v`, reverting the concurrent update.

For non-others wallets the read-to-call path is synchronous so the race requires mutex contention, making it unlikely in practice. For others wallets the path is async (DB lookups in `resolveOthersWalletAccountForNetworkSwitch`) making contention more plausible, but the full-snapshot spread is intentional there to apply the resolved `othersWalletAccountId`.

Additionally, this interacts with the derive-type fix logic at line 1333: `if (newDeriveType === oldDeriveType)`. If a stale `deriveType` from `before` differs from the fresh `oldDeriveType` in `v`, the condition fails and the global derive-type fix is skipped, leaving a potentially wrong derive type for the new network.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -593,6 +755,324 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
+      return resolvedSelectedAccount;
+    }
+
+    debugAccountSelectorLog('updateSelectedAccountNetwork:others-unresolved', {
+      source,
+      sceneName,
+      sceneUrl,
+      num,
+      networkId,
+      selectedAccount: pickSelectedAccountLogFields(selectedAccount),
```
sidmorizon: **resolveOthersWalletAccountForNetworkSwitch falls back to keeping the original incompatible account**

Type: `Flag`

When switching networks for an 'others' wallet (imported/watching/external), if no compatible account is found among the wallet's singleton accounts AND the current account has no compatible fallback network, the function at line 936 returns `nextSelectedAccount` — which still contains the *new* networkId but the *old* othersWalletAccountId that is incompatible with it. This means the UI will show a mismatched account/network state until the next reload or auto-select cycle. The `isIncompatibleOthersWalletNetworkPair` guard in `saveToStorage` prevents this from being persisted, which is good defense-in-depth, but the in-memory state will be temporarily inconsistent.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -2371,10 +2988,66 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
+        }
+        const hasAccountIdentityForStorage = Boolean(
+          selectedAccount?.walletId &&
+          (selectedAccount.indexedAccountId ||
+            selectedAccount.othersWalletAccountId),
+        );
+        if (!hasAccountIdentityForStorage) {
+          debugAccountSelectorLog('saveToStorage:skip-no-identity', {
+            sceneName,
+            sceneUrl,
```
sidmorizon: **saveToStorage now blocks persistence of network-only (no account identity) selections**

Type: `Flag`

The new `hasAccountIdentityForStorage` guard at lines 2998-3010 prevents persisting selections that have a `networkId` but no `walletId`/`indexedAccountId`/`othersWalletAccountId`. This is a significant behavioral change — previously such incomplete selections could be saved. The intent appears to be preventing cold-start cache entries from overwriting real saved accounts. This is validated by the test 'does not persist a network-only cold-start selection over a saved account'. However, callers that legitimately save partial selections (e.g., during a network-only update before account resolution completes) will now silently be skipped. Downstream code that relied on saveToStorage always persisting what was passed may behave differently.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
originalix: Fixed in `ed2a50094595c3997c9f13b9f518862a8eac55c4`.

`updateSelectedAccountNetwork` no longer spreads the pre-mutex `before` snapshot back into the current atom state. The non-others path only updates `networkId`; the others-wallet path only applies resolved identity fields when the resolver actually switches to a different others account, otherwise it also preserves the latest fields from `v` and only updates the final network.

Added regression coverage with `preserves concurrent fields while resolving an others wallet network switch`.

Verified:
- `yarn jest packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx --runInBand`
- `npx oxlint --tsconfig ./tsconfig.json --type-aware --deny-warnings packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx`
- `yarn lint:staged`
- `yarn tsc:staged`

zhaono1 (approved):

PR #12170, opened 2026-06-25
fix: cap displayed transfers in tx history to prevent freeze (OK-55756)

franco-chan:
[OK-55756](https://onekeyhq.atlassian.net/browse/OK-55756)

----------

## Summary

A single transaction that moves many assets in one tx — most notably airdrops of **thousands of NFTs** — froze the client. Each NFT carries a unique `tokenIdOnNetwork`, so they don't collapse into one line, and both the desktop history **list row** and the transaction **detail page** rendered one row per transfer (each with a remote image) with no virtualization, mounting thousands of elements at once.

Jira: **OK-55756**

## Changes

- Cap the number of transfer rows rendered for a single tx at `MAX_DISPLAYED_TRANSFERS` (5) for the whole tx (sends + receives combined), across both surfaces, and show a single "+N" overflow row for the rest:
  - History list row — `buildExpandedTransferView` (`TxActionTransfer.tsx`).
  - Detail page — `HistoryDetails.tsx`.
- Keep `getTransferChangeLineCount` (the web row-height fast path in `TxHistoryListView`) in sync with the cap so a many-transfer row can no longer reserve a giant row height.
- The overflow row shows a placeholder chip (sized to the row's token/NFT image, `$bgStrong`, `DotHorOutline` icon, corner radius matching NFT vs fungible) plus a label:
  - "+N NFTs" when the hidden transfers are all NFTs (hardcoded — NFT is not translated).
  - "+N assets" otherwise, reusing the existing `count_assets` string.
- Render-only change: the underlying transfer data is left intact (no DB/backend change), so other consumers (block explorer handoff, KYT scan, UTXO math) still see the full set. The complete list remains available via the block explorer link.

## Notes

- `MAX_DISPLAYED_TRANSFERS` and the overflow-label helper are shared via `components/TxAction/consts.ts`.
- Slicing happens on the data before mapping, so the thousands-of-NFTs case never builds thousands of throwaway rows.

## Test plan

- [x] Receive a tx with thousands of NFTs → list row and detail page show 5 rows + "+N NFTs"; no freeze.
- [x] Tx with more than 5 distinct fungible tokens → "+N assets".
- [x] Normal txs (1–5 transfers, swap, send-to-self, UTXO) render unchanged.

[OK-55756]: https://onekeyhq.atlassian.net/browse/OK-55756?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f0c558e0-e58b-4218-9fe9-5d64af938d41?refs=2b6824b22d3921b6412cb1bc1191376b1bef8198&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/f0c558e0-e58b-4218-9fe9-5d64af938d41/license?refs=2b6824b22d3921b6412cb1bc1191376b1bef8198&source=prChecksComment) |

packages/kit/src/components/TxAction/consts.ts
```diff
@@ -0,0 +1,36 @@
...
+  isNFT: boolean;
+  intl: IntlShape;
+}): string {
+  if (isNFT) {
+    return `+${count} NFT${count > 1 ? 's' : ''}`;
+  }
+  return `+${intl.formatMessage(
+    { id: ETranslations.count_assets },
+    { count },
+  )}`;
```
sidmorizon, line 38: # 🟡 P2 溢出标签同种资产复数 `+1 assets` 语法错误

## 问题

`formatTransferOverflowLabel` 对 NFT 分支正确处理了单复数（`NFT${count > 1 ? 's' : ''}`），但同种资产（fungible）分支直接复用 `count_assets`（其英文值为 `'{count} assets'`，`assets` 为硬编码复数，且无 ICU `plural` 形式）。

`overflowCount` 完全可能等于 `1`：当 tx 的可见行恰好用满 `MAX_DISPLAYED_TRANSFERS`（`5`）、且总转账数为 `6` 时，`combinedTransfers.slice(MAX_DISPLAYED_TRANSFERS)` 长度为 `1`。在详情页（`HistoryDetails.tsx` 未按 token 分组，按原始 transfers 计数）这种 6 笔转账的场景尤其常见。

## 影响

此时溢出行会渲染为 `+1 assets`，英文语法错误（应为 `+1 asset`）。NFT 分支已规避了同样的问题，fungible 分支却遗漏，行为不一致，且是用户可见文案。

## 建议

对 fungible 分支也做单复数处理，例如新增一个单数的 i18n 词条并按 `count` 选择，或临时与 NFT 分支保持一致的内联处理：

```ts
return `+${count} asset${count > 1 ? 's' : ''}`;
```

更优做法是为 `count_assets` 增加 ICU `plural` 形式（`{count, plural, one {# asset} other {# assets}}`），从根上覆盖所有复用点。

[Claude session ↗](https://claude.ai/code/session_017JPWSQ35eCy4iAYp7g6LfE)

---

franco-chan: Good catch — this is a real, reachable bug. `overflowCount` can be `1` (e.g. exactly 6 transfers → 5 shown + 1 hidden, common on the ungrouped detail page), and `count_assets` (`"{count} assets"`) is plural-only, so it would render the grammatically wrong `+1 assets`.

Fixed in 24f8297f4c — but via the call sites rather than the label, so we keep the `count_assets` translation (e.g. zh `种资产`) instead of hardcoding an English `asset`/`assets`:

> We no longer collapse a single trailing transfer into a "+N" row. The overflow row is shown only when it hides 2+ transfers; when exactly one would be hidden we just render that last row. So `overflowCount` is now always `0` or `>= 2`, and `+N assets` / `+N NFTs` are always plural-correct.

This also keeps rendered lines at `min(total, MAX_DISPLAYED_TRANSFERS + 1)`, so `getTransferChangeLineCount` (the web row-height fast path) still matches with no change, and it's slightly better UX (no row wasted to hide a single item). Applied to both the list row (`buildExpandedTransferView`) and the detail page (`cappedAssetChange`); the helper now documents that `count >= 2`.

The deeper root fix you suggested — an ICU `plural` form on `count_assets` (`{count, plural, one {# asset} other {# assets}}`) — is the right long-term answer to make the helper bulletproof for any `count`, but our i18n strings are managed on Lokalise (created there, then `yarn fetch:locale`), so I left that out of this PR. Happy to coordinate that key separately if we'd rather rely on it than on the no-single-collapse rule.

packages/kit/src/views/AssetDetails/pages/HistoryDetails/HistoryDetails.tsx
```diff
@@ -886,6 +891,50 @@ function HistoryDetails() {
...
+    // "+N assets" label stays grammatically correct (never "+1 assets").
+    const visibleCount =
+      transferRows.length > MAX_DISPLAYED_TRANSFERS + 1
+        ? MAX_DISPLAYED_TRANSFERS
+        : transferRows.length;
+    const overflowRows = transferRows.slice(visibleCount);
+    return {
+      visibleRows: transferRows.slice(0, visibleCount),
+      overflowCount: overflowRows.length,
+      // Match the overflow chip's corner to what it summarizes: NFT images use
```
sidmorizon, line 929: **Display order of sends vs receives differs between list view and detail page**

Type: `Flag`

In `buildExpandedTransferView` (`packages/kit/src/components/TxAction/TxActionTransfer.tsx:485`), the combined array is `[...receives, ...sends]` — receives first. In the detail page's `cappedAssetChange` (`packages/kit/src/views/AssetDetails/pages/HistoryDetails/HistoryDetails.tsx:870-877`), the `transfersToRender` array is constructed with sends first (`EDecodedTxDirection.OUT`), then receives (`EDecodedTxDirection.IN`). This means when the cap truncates, different items may be hidden on the list row vs. the detail page. This ordering difference is pre-existing (not introduced by this PR), but the new truncation makes it more visible: a user might see 5 receives in the list row but 5 sends on the detail page for the same transaction.

franco-chan: Good flag — real (if minor) inconsistency. As you note the ordering difference is pre-existing, but the new cap makes it user-visible: for a mixed-direction tx that exceeds the cap, the list row (receives-first) and the detail page (`transfersToRender`, sends-first) could hide different items.

Fixed in 2b6824b22d by aligning the list row to the detail page's order: `buildExpandedTransferView` now iterates `[...sends, ...receives]` (sends first), matching `transfersToRender`. So both surfaces now truncate the same items, and the send→receive flow reads consistently.

Scope notes:
- The target scenario (NFT airdrops) is single-direction (sends empty), so it was never affected — this only matters for the rarer mixed-direction txs.
- This only touches the desktop `tableLayout` expanded column; the mobile list's primary +/- amount comes from `buildTransferChangeInfo`, not this path, so its "received amount" emphasis is unchanged.
- The list row groups by token while the detail page lists raw transfers, so exact counts can still differ — but that's the separate pre-existing grouping behavior, not the ordering concern raised here.

originalix (approved):

PR #12173, opened 2026-06-25
fix: sort perp hot and new tabs(OK-56817)

originalix:
[OK-56817](https://onekeyhq.atlassian.net/browse/OK-56817)

----------

## Summary
- Preserve backend token order for dynamic Perps selector tabs, such as Hot and New List, until the user actively clicks a sortable header.
- Track default vs user-initiated sorting and use live asset values for user sorting so Volume/Open Interest order matches the displayed rows.
- Share dynamic-tab sort behavior across desktop and mobile token selectors and add regression coverage.

## Intent & Context
Related issue: [OK-56817](https://onekeyhq.atlassian.net/browse/OK-56817) - Perps - 新上架/热门 tab 的排序点击无效.

A runtime report on the Perps trading page showed that the token selector dynamic tabs did not reorder correctly when clicking sortable column headers such as Volume. The first symptom was that the newly added Hot list visually kept its original order when the sort arrow changed. After that path was corrected, the list could still be numerically inaccurate because displayed Volume values update live while the sort source could remain based on a previous snapshot.

The requested scope was limited to this Perps token selector sorting bug, without unrelated desktop build or route changes.

## Root Cause
The token selector did not distinguish dynamic tabs' backend/default token order from an explicit user header sort. Dynamic tabs should initially respect the backend-configured token order, but once a user clicks a column header, the rows must be sorted from the same live asset context used to render the column values. Without that separation, the UI could show a sort arrow while rows were still derived from server order or stale snapshot values.

## Design Decisions
- Add `sortSource` to distinguish backend/default tab ordering from explicit user sorting.
- Keep dynamic tabs in backend token order until a header interaction; the first user click starts descending sort instead of toggling from an invisible default.
- Use live asset contexts for user-initiated sorting, while keeping default ordering based on the stable snapshot to avoid unnecessary reorder churn before the user requests sorting.
- Centralize sort transitions and dynamic-tab filtering in shared helpers so desktop and mobile selectors stay aligned.

## Changes Detail
- `tokenSelectorTabs.ts`: adds dynamic-tab ordering helpers, sort source transitions, active-header detection, and live/snapshot sort-source selection.
- `PerpTokenSelector.tsx` and `MoblieTokenSelector.tsx`: use the shared helpers for dynamic tab filtering and user sorting.
- `SortableHeaderCell.tsx`: delegates active state and next-sort calculation to shared helpers.
- `perps.ts`, `types.ts`, and `tokenSelectorPerf.ts`: persist and surface the new `sortSource` field.
- `tokenSelectorTabs.test.ts`: adds regression tests for dynamic default order, user sorting, header state, source changes, and live Volume alignment.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Web / Mobile
- **Risk Areas**: Perps token selector row ordering and persisted selector sort config. This does not affect order placement, signing, wallet balance, or transaction paths.

## Test plan
- [x] `yarn lint --fix`
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn jest packages/kit/src/views/Perp/utils/tokenSelectorTabs.test.ts --runInBand`
- [x] Verify desktop Perps Hot tab Volume descending/ascending order remains monotonic after live updates.

[OK-56817]: https://onekeyhq.atlassian.net/browse/OK-56817?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
+++ b/packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
@@ -103,13 +103,16 @@ import {
   buildPerpTokenSelectorCategoryTabs,
   buildPerpTokenSelectorTabs,
   buildPrimaryTabs,
+  getNextPerpTokenSelectorSortConfig,
+  getPerpTokenSelectorDynamicTabItems,
   getPerpTokenSelectorFallbackTabId,
   getPerpTokenSelectorPrimaryTabId,
+  getPerpTokenSelectorSortAssetCtxsByDex,
   isPerpTokenSelectorFavoritesTab,
   isPerpTokenSelectorPerpsTab,
+  isPerpTokenSelectorSortFieldActive,
   isPerpTokenSelectorSpotTab,
   shouldRefreshPerpTokenSelectorSortSnapshot,
-  sortPerpTokenSelectorItemsByServerOrder,
 } from '../../utils/tokenSelectorTabs';
 
 import { FavoritesEmptyState } from './FavoritesEmptyState';
@@ -540,13 +543,16 @@ function MobileTokenSelectorModal({
   // Freeze sort order; only refresh on sort config change or first data arrival.
   // Does NOT track activeTab — tab switches should not refresh the snapshot.
   const ctxSnapshotRef = useRef(assetCtxsByDex);
+  const [perpSortSnapshot, setPerpSortSnapshot] = useState(assetCtxsByDex);
   const lastSortRef = useRef<{
     field?: string;
     direction?: string;
+    sortSource?: IPerpTokenSelectorConfig['sortSource'];
   } | null>(null);
   useEffect(() => {
     const field = selectorConfig?.field;
     const direction = selectorConfig?.direction;
+    const sortSource = selectorConfig?.sortSource;
     const last = lastSortRef.current;
     // Also refresh when snapshot is empty (first WS data arrival after mount)
     const snapshotEmpty = !ctxSnapshotRef.current?.some(
@@ -556,19 +562,26 @@ function MobileTokenSelectorModal({
       lastSort: last,
       field,
       direction,
+      sortSource,
       snapshotEmpty,
     });
     if (!shouldRefresh) {
       return;
     }
-    lastSortRef.current = { field, direction };
+    lastSortRef.current = { field, direction, sortSource };
     ctxSnapshotRef.current = assetCtxsByDex;
-    if (last?.field !== field || last?.direction !== direction) {
+    setPerpSortSnapshot(assetCtxsByDex);
+    if (
+      last?.field !== field ||
+      last?.direction !== direction ||
+      last?.sortSource !== sortSource
+    ) {
       scrollListToTop();
     }
   }, [
     selectorConfig?.direction,
     selectorConfig?.field,
+    selectorConfig?.sortSource,
     assetCtxsByDex,
     scrollListToTop,
   ]);
@@ -625,6 +638,7 @@ function MobileTokenSelectorModal({
           field: prev?.field ?? DEFAULT_PERP_TOKEN_SORT_FIELD,
           direction: prev?.direction ?? DEFAULT_PERP_TOKEN_SORT_DIRECTION,
           activeTab: tab,
+          sortSource: prev?.sortSource ?? 'default',
         }));
       });
       actions.current.setTradeRouteViewState({
@@ -718,11 +732,16 @@ function MobileTokenSelectorModal({
 
   // Layer 1: sort — only reruns when sort config or underlying assets change.
   // Does NOT depend on activeTab, so tab switches never retrigger the sort.
+  const perpSortAssetCtxsByDex = getPerpTokenSelectorSortAssetCtxsByDex({
+    liveAssetCtxsByDex: assetCtxsByDex,
+    snapshotAssetCtxsByDex: perpSortSnapshot,
+    sortSource: selectorConfig?.sortSource,
+  });
   const perpSortedList = useMemo(() => {
     const perfStartTime = startTokenSelectorPerfMeasure();
     const assetsByDexTyped: IPerpsUniverse[][] = assetsByDex || [];
     const assetCtxsByDexTyped: IPerpsAssetCtx[][] =
-      ctxSnapshotRef.current || [];
+      perpSortAssetCtxsByDex || [];
 
     const combinedEntries = assetsByDexTyped.flatMap(
       (assets: IPerpsUniverse[], dexIndex: number) => {
@@ -751,6 +770,7 @@ function MobileTokenSelectorModal({
 
     const sortField = selectorConfig?.field ?? '';
     const sortDirection = selectorConfig?.direction ?? 'desc';
+    const sortSource = selectorConfig?.sortSource;
     const result = sortField
       ? combinedEntries
           .toSorted((a, b) =>
@@ -768,6 +788,7 @@ function MobileTokenSelectorModal({
         phase: 'perp-sort',
         sortField,
         sortDirection,
+        sortSource,
         perpCount: combinedEntries.length,
         resultCount: result.length,
       });
@@ -777,9 +798,11 @@ function MobileTokenSelectorModal({
   }, [
     assetsByDex,
     computeSortValues,
+    perpSortAssetCtxsByDex,
     sortCompare,
     selectorConfig?.direction,
     selectorConfig?.field,
+    selectorConfig?.sortSource,
     tokenSearchAliases,
   ]);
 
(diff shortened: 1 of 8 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/ae674d50-1a58-4948-882a-6443c80b4083?refs=48906e4a0a0f4cb25a4831a25e9833e94f2d81b8&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/ae674d50-1a58-4948-882a-6443c80b4083/license?refs=48906e4a0a0f4cb25a4831a25e9833e94f2d81b8&source=prChecksComment) |

packages/kit/src/views/Perp/components/TokenSelector/PerpTokenSelector.tsx
```diff
@@ -484,15 +486,12 @@ function BasePerpTokenSelectorContent({
...
-              direction: prev?.direction ?? DEFAULT_PERP_TOKEN_SORT_DIRECTION,
-              activeTab: tab,
-              // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            }) as any,
-        );
+        setSelectorConfig((prev) => ({
+          field: prev?.field ?? DEFAULT_PERP_TOKEN_SORT_FIELD,
+          direction: prev?.direction ?? DEFAULT_PERP_TOKEN_SORT_DIRECTION,
+          activeTab: tab,
+          sortSource: prev?.sortSource ?? 'default',
```
sidmorizon: # 🟠[P1] sortSource 是全局状态，会泄漏到动态 tab 破坏“后端顺序”不变量

## 问题

`sortSource` 存在于全局且持久化的 `perpTokenSelectorConfigPersistAtom` 中，没有按 tab 作用域隔离。一旦用户在 `perps` 这类 primary tab 上点击表头排序，配置变为 `{ field: 'volume24h', direction: 'desc', sortSource: 'user' }`。随后 `updateActiveTab` 用 `sortSource: prev?.sortSource ?? 'default'` 切到 `hot`/`new` 动态 tab 时，`sortSource: 'user'` 被原样保留。

动态 tab 渲染用 `getPerpTokenSelectorDynamicTabItems({ useSortedItemsOrder: selectorConfig?.sortSource === 'user' })`，此时 `useSortedItemsOrder` 为 `true`，于是 hot/new 直接按 volume 排序、表头箭头也判为 active —— 而用户从未在该动态 tab 上点过表头。更严重的是该状态被持久化：用户在任意 tab 排过一次序后，下次冷启动打开 hot/new 立即是排序态，永远看不到后端原始顺序。

## 影响

直接破坏本 PR 声明的核心不变量“dynamic tabs (Hot/New) must keep backend order until a header click”。桌面与移动端 (`MoblieTokenSelector.tsx` 第 641 行同样 `sortSource: prev?.sortSource ?? 'default'`) 均受影响，且经持久化后跨会话长期错误。

## 建议

切换到动态 tab（或从动态↔primary 切换）时，应重置 `sortSource` 为 `'default'`，或把排序状态按 tab 维度作用域化，使非用户主动排序的动态 tab 始终回落到后端顺序。例如在 `updateActiveTab` 中当目标 tab 为动态 tab 时强制 `sortSource: 'default'`。

[Claude session ↗](https://claude.ai/code/session_0137HHahU5dmfyYkVgEvG4P9)

---

packages/kit/src/views/Perp/utils/tokenSelectorTabs.ts
```diff
@@ -231,20 +244,137 @@ function sortPerpTokenSelectorItemsByServerOrder<T>({
...
+  liveAssetCtxsByDex,
+  snapshotAssetCtxsByDex,
+  sortSource,
+}: {
+  liveAssetCtxsByDex: T;
+  snapshotAssetCtxsByDex: T;
+  sortSource?: IPerpTokenSelectorConfig['sortSource'];
+}) {
+  return sortSource === 'user' ? liveAssetCtxsByDex : snapshotAssetCtxsByDex;
+}
```
sidmorizon, line 406: **User sort on primary tabs bypasses frozen snapshot pattern**

Type: `Flag`

When `sortSource` is `'user'` (after clicking a sort header), `getPerpTokenSelectorSortAssetCtxsByDex` at `packages/kit/src/views/Perp/utils/tokenSelectorTabs.ts:358` returns `liveAssetCtxsByDex` regardless of which tab is active. On primary tabs like 'perps', this means `perpSortedList` recomputes on every WebSocket update — breaking the frozen-snapshot pattern that the comments at `MoblieTokenSelector.tsx:543` and `PerpTokenSelector.tsx:585` describe. Previously, sorting ALWAYS used a snapshot so the list order stayed frozen between sort-config changes. Now, on primary tabs, when the user clicks a sort header (`sortSource` becomes `'user'`), `perpSortAssetCtxsByDex` points to live atom data, causing O(n log n) re-sorts on every WS tick. The `isPerpTokenSelectorSortFieldActive` function correctly distinguishes primary vs dynamic tabs for visual indicators, but `getPerpTokenSelectorSortAssetCtxsByDex` does not make this distinction for the data source. This may be intentional (live sort when user explicitly sorts), but it's a behavioral change from the prior frozen-sort design and could cause performance issues on lists with 200+ tokens receiving frequent WS updates.

packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
```diff
@@ -718,11 +736,18 @@ function MobileTokenSelectorModal({
 
   // Layer 1: sort — only reruns when sort config or underlying assets change.
   // Does NOT depend on activeTab, so tab switches never retrigger the sort.
+  const perpSortAssetCtxsByDex = getPerpTokenSelectorSortAssetCtxsByDex({
+    activeTab: displayActiveTab,
+    liveAssetCtxsByDex: assetCtxsByDex,
+    snapshotAssetCtxsByDex: perpSortSnapshot,
+    sortSource: selectorConfig?.sortSource,
+    sortSourceTab: selectorConfig?.sortSourceTab,
+  });
```
sidmorizon, line 741: **Dynamic tab user sort causes full perp list re-sort on every WebSocket frame**

Type: `Flag`

When a user activates sorting on a dynamic tab (e.g. 'hot'), `getPerpTokenSelectorSortAssetCtxsByDex` returns `liveAssetCtxsByDex` (= `assetCtxsByDex`) instead of the frozen snapshot (`MoblieTokenSelector.tsx:739-745`, `PerpTokenSelector.tsx:771-777`). Since `assetCtxsByDex` gets a new reference on every WebSocket update and is included in the `perpSortedList` useMemo dependency array, the entire perp list (potentially hundreds of items) is re-sorted on every WS frame — even though only a small subset of items is shown on the dynamic tab. The old code used `ctxSnapshotRef.current` (a ref read, not a dependency), so the sort only ran when sort config changed.

This appears intentional based on the test at `tokenSelectorTabs.test.ts:458-504`, but it trades sort-order freshness on a small dynamic tab view for O(n log n) work per WS frame over the full list. Consider filtering to the dynamic tab's tokens before sorting, or memoizing the live-data path separately for the small subset.

packages/kit/src/views/Perp/components/TokenSelector/PerpTokenSelector.tsx
originalix: Fixed in 56fc72fca5.

The selector sort state is now scoped by tab via `sortSourceTab`, and tab switches reset `sortSource` back to `default`, so dynamic tabs keep the backend/server token order unless the user explicitly clicks a header in that same dynamic tab. This also handles previously persisted `sortSource: "user"` values without `sortSourceTab` by treating them as non-active for dynamic tabs.

Desktop and mobile/iOS selector paths are both updated, with unit coverage for dynamic-tab scoping and tab-switch reset.

packages/kit/src/views/Perp/utils/tokenSelectorTabs.ts
originalix: Fixed in 56fc72fca5.

`getPerpTokenSelectorSortAssetCtxsByDex` now returns live asset ctxs only when the active tab is a matching dynamic tab user-sort (`sortSource: "user"` and `sortSourceTab === activeTab`). Primary tabs, including `perps`, continue to use the frozen snapshot even after the user clicks a sort header, preserving the existing snapshot/performance behavior.

Added unit coverage that asserts primary user sort uses the snapshot and only matching dynamic tab user sort uses live ctxs.

packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
originalix: Fixed in 48906e4a0a.

The full perp universe sort now always uses the frozen snapshot, so `perpSortedList` no longer depends on live `assetCtxsByDex` for dynamic-tab user sorting.

For dynamic tabs, the user-sort path first filters to the current tab tokens, then sorts only that small subset against live ctx values. This preserves live/accurate Hot/New header sorting while avoiding O(n log n) full-list re-sorts on every WebSocket frame. Desktop and mobile/iOS paths were updated together, and the unit test now asserts the full-list sort source stays snapshot-based.

packages/kit/src/views/Perp/utils/tokenSelectorTabs.ts
```diff
@@ -231,20 +265,166 @@ function sortPerpTokenSelectorItemsByServerOrder<T>({
...
+  tab: string;
+}): IPerpTokenSelectorConfig {
+  return {
+    field: prev?.field ?? DEFAULT_PERP_TOKEN_SORT_FIELD,
+    direction: prev?.direction ?? DEFAULT_PERP_TOKEN_SORT_DIRECTION,
+    activeTab: tab,
+    sortSource: 'default',
+    sortSourceTab: undefined,
+  };
+}
```
sidmorizon, line 354: **Tab switch discards user sort state permanently**

Type: `Flag`

When a user switches tabs via `getNextPerpTokenSelectorActiveTabConfig` (`tokenSelectorTabs.ts:340-354`), `sortSource` is always reset to `'default'` and `sortSourceTab` to `undefined`. This means if a user sorts by 'change24hPercent asc' on dynamic tab 'hot', switches to 'AI', then switches back to 'hot', the sort is lost and 'hot' returns to server order. This is verified by the test 'resets user sort source when switching tabs' (`tokenSelectorTabs.test.ts:361-380`), so it appears intentional. However, this differs from the behavior on primary tabs where the sort field/direction persist across tab switches (only the source metadata is reset). Users who sort a dynamic tab and navigate away may be surprised that their sort preference isn't remembered.

huhuanming (approved):

PR #10819, opened 2026-03-23
perf: lazy tab loading, idle-time preload & device-tier-aware startup for low-end devices

huhuanming:
## Summary

Improves cold-start and tab-navigation performance, with extra care for low-end / low-memory devices. The work grew from "lazy tab loading" into a small startup-performance bundle sharing **one** device-memory signal as the single source of truth.

### 1. Lazy tab loading + progressive idle-time preload
- All tab screens lazy-load by default (`lazy: true`) so the first paint no longer mounts every tab.
- Key tabs are preloaded in the background during **idle time**, driven by the device performance tier (see below). Every preload step now runs inside `requestIdleCallback` — not just the first — so heavy tab mounts never contend with active user interaction.
- Preloaded routes bypass `lazy` / `freezeOnBlur` in `NativeBottomTabView` so they render correctly after preload.
- Preload is dispatched with an explicit `target` and no params, avoiding a StackView crash and unintended remounts.

### 2. Device performance tier (`devicePerformanceTier`)
- New shared module classifying the device into `high` / `medium` / `low` from: cached tier (sync) → runtime UI-visible-time calibration (async, persisted for next launch) → device memory.
- On first launch (fresh install / cleared cache) the tier is derived **synchronously from device memory** instead of a hardcoded `medium`, so low-end devices skip the heavier preload queue from the very first boot.
- Preload queues per platform × tier live in `preloadConfig.ts` (`low` → no preload).

### 3. Low-end iOS cold-start-lock defer
- On low-memory iOS, the app tree render is deferred behind the cold-start lock until the first real unlock, mitigating the jetsam cold-start-lock kill loop. Decision logic is extracted into pure, unit-tested helpers (`lowEndDeviceUtils`).

### 4. Unified low-memory threshold (single source of truth)
- All RAM reads go through one synchronous primitive `getDeviceMemoryGBSync` (platform-split: `expo-device.totalMemory` native / `navigator.deviceMemory` web).
- One shared threshold `LOW_END_MEMORY_THRESHOLD_GB = 3.5` now drives every memory-tiering decision so they agree:
  - `devicePerformanceTier` low tier (`<= 3.5GB`)
  - DApp WebView alive-count cap (`webviewAliveLimit.native.ts`)
  - cold-start-lock defer guard
- Fixes a latent contradiction where a jetsam-risk device (e.g. iPhone 7 Plus ~3.14GB) was tagged `medium` / kept 5 WebViews alive — exactly the devices the guard tries to protect.

[...]

PR #12174, opened 2026-06-25
fix(mobile): defer app tree render behind cold-start lock on low-end iOS

huhuanming:
## Problem

On low-end iOS devices (≤3.5 GiB RAM, e.g. **iPhone 7 Plus**, app 6.4.0), cold-starting **while the app is locked** can make the app effectively **impossible to unlock** — the user enters the password / Face ID spins, but the app keeps restarting before it unlocks.

Diagnosed from a real device log (iPhone 7 Plus, iOS 15.8.8, 3.14 GB): ~50 cold starts in 44 min, the unlock `verifyPassword` dispatched 57× but completed only once.

## Root cause

This is dual: a memory/jetsam loop + bg-thread starvation, both rooted in the same thing.

- The lock screen is an **overlay** — `AppStateLockContainer` renders `{children}` (the entire `DetailRouter` app tree) **unconditionally**, *behind* the lock screen.
- So while locked, the whole app boots behind the lock: the home data cascade (token/defi/staking/account-value), the app-level `Bootstrap` fetches (market/perps/currency), and the `InAppNotification` swap-status loop — **~67 concurrent calls flooding the single background JS runtime thread**, plus the memory of the full tree.
- The only part of unlock that needs the bg is `servicePassword.verifyPassword` — a **600,000-iteration PBKDF2** (native quick-crypto) that needs the bg thread for ~5s. Starved by the flood + the device under memory pressure, the process is **jetsam-killed (~10–19s)** before verify finishes → relaunch → repeat.
- Lock screen + biometric (`isEnable`, `getPassword`/keychain) are main-thread/local — they do **not** need the bg. Only verify does.

## Fix

On **low-end iOS only**, for a **cold-start lock** (process never unlocked yet), do not render `AppStateLockContainer`'s `{children}`. This frees the bg thread and cuts memory so `verifyPassword` can finish before jetsam.

- The latch flips on a real **locked → unlocked transition** (not on any `isLocked === false` render), so after the first unlock children always render again — an **auto-lock-while-using never unmounts the user's current screen** — and a transient default-`unlocked` render during early state hydration (e.g. first cold start after upgrade) cannot permanently defeat the optimization.
- The **lock screen + verify are untouched** (the lock overlay is a sibling of `{children}`, with no dependency on the nav tree) — no path to "can never unlock".
- **Non-low-end and non-iOS behavior is byte-for-byte unchanged** (`IS_LOW_END_DEVICE` is `false` on web/desktop/ext; gated by `platformEnv.isNativeIOS`).
- Threshold `totalMemory ≤ 3.5 GiB` (binary), reusing the `expo-device` tiering precedent in `webviewAliveLimit.native.ts`; covers 2–3 GB-class iOS devices.

## Files

- `packages/shared/src/utils/lowEndDevice.native.ts` / `.ts` — platform-split `IS_LOW_END_DEVICE` constant.
- `packages/shared/src/utils/lowEndDeviceUtils.ts` (+ `.test.ts`) — pure `isUnlockTransition` / `shouldDeferColdStartLockRender` (8 unit tests).
- `packages/kit/src/provider/Container/AppStateLockContainer/index.tsx` — gate `{children}` for cold-start lock on low-end iOS.

## Adversarial review outcome

- **No catastrophic failure mode**: device can always unlock; children always render after unlock in steady state. Deep-link / cold-start-by-notification self-defer via `whenAppUnlocked()` (delayed, not lost); no `rootNavigationRef` NPE.
- The one HIGH finding (a transient default-`unlocked` first render could defeat the gate on the post-upgrade boot) is **fixed** by the transition-based latch.

## Verify on a real 3 GB iOS device before release

- [ ] Cold-start locked: confirm reliable unlock; bg in-flight count at verify drops from ~67 toward ~31.
- [ ] Post-unlock first paint: no lingering blank/white flash while the tree mounts.
- [ ] Auto-lock while using: unlocking returns to the same screen (children not unmounted).
- [ ] `expo-device.totalMemory` reads ~3.14 GB on the 7 Plus → lands in the low-end bucket.

Note: scoped to **iOS** by design; low-RAM Android is intentionally excluded from this hotfix (same architecture, but no evidence yet — can be a follow-up).

https://claude.ai/code/session_01FFUy2BJJ3BAi6nkwqahM45

### DIFF
--- a/packages/kit/src/provider/Container/AppStateLockContainer/index.tsx
+++ b/packages/kit/src/provider/Container/AppStateLockContainer/index.tsx
@@ -6,6 +6,11 @@ import { ANIMATE_ONLY_OPACITY } from '@onekeyhq/components/src/utils/animationCo
 import backgroundApiProxy from '@onekeyhq/kit/src/background/instance/backgroundApiProxy';
 import { useAppIsLockedAtom } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
+import { IS_LOW_END_DEVICE } from '@onekeyhq/shared/src/utils/lowEndDevice';
+import {
+  isUnlockTransition,
+  shouldDeferColdStartLockRender,
+} from '@onekeyhq/shared/src/utils/lowEndDeviceUtils';
 
 import PasswordVerifyContainer from '../../../components/Password/container/PasswordVerifyContainer';
 
@@ -86,6 +91,32 @@ export function AppStateLockContainer({
 }: PropsWithChildren<unknown>) {
   const [isLocked] = useAppIsLockedAtom();
 
+  // Track whether this process has ever been unlocked. On low-end iOS devices we
+  // skip rendering the full app tree behind the lock screen ONLY for a
+  // cold-start lock (never unlocked yet) — so the single background JS thread
+  // is free to finish the 600k-iteration PBKDF2 `verifyPassword` before iOS
+  // jetsam kills the process. Once unlocked, we always render children again so
+  // an auto-lock-while-using never unmounts the user's current screen.
+  //
+  // The latch only flips on a real locked->unlocked transition (not on any
+  // `isLocked === false` render) so a transient/default `false` during early
+  // state hydration (e.g. the first cold start after upgrade) cannot
+  // permanently defeat the optimization on the riskiest boot.
+  const hasUnlockedOnceRef = useRef(false);
+  const prevLockedRef = useRef(isLocked);
+  if (isUnlockTransition(prevLockedRef.current, isLocked)) {
+    hasUnlockedOnceRef.current = true;
+  }
+  prevLockedRef.current = isLocked;
+  const deferColdStartChildren = shouldDeferColdStartLockRender({
+    // RAM-based low-end detection is also true on low-RAM Android, but the
+    // jetsam / dual-runtime PBKDF2-starvation problem this guards against is
+    // iOS-specific, so the defer behavior is gated to iOS.
+    isLowEndDevice: IS_LOW_END_DEVICE && platformEnv.isNativeIOS,
+    isLocked,
+    hasUnlockedOnce: hasUnlockedOnceRef.current,
+  });
+
   const handleUnlock = useCallback(async () => {
     await backgroundApiProxy.servicePassword.unLockApp();
   }, []);
@@ -118,7 +149,7 @@ export function AppStateLockContainer({
 
   return (
     <>
-      {children}
+      {deferColdStartChildren ? null : children}
       {!isLocked ? <AppStateUpdater /> : null}
       <AnimatePresence>
         {isLocked ? (
--- a/packages/shared/src/utils/lowEndDevice.native.ts
+++ b/packages/shared/src/utils/lowEndDevice.native.ts
@@ -1,0 +1,17 @@
+import { totalMemory } from 'expo-device';
+
+// Native low-end-device flag, computed once at module load from device RAM.
+//
+// `expo-device` `totalMemory` is in bytes. We use binary GiB (1024^3) so that
+// 3GB-class devices that report ~3.14e9 bytes (e.g. iPhone 7 Plus) are included.
+// Threshold 3.5 * GiB = 3_758_096_384 bytes.
+//
+// Used to defer rendering the full app tree behind the lock screen during a
+// cold-start lock, so the single background JS thread is free to finish the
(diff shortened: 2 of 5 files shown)


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/02b7e05d-ba01-4213-aacf-8cfba020a9d0?refs=14f1c57a6a2c8b6ebec4895b2212a09adcf32d3c&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/02b7e05d-ba01-4213-aacf-8cfba020a9d0/license?refs=14f1c57a6a2c8b6ebec4895b2212a09adcf32d3c&source=prChecksComment) |

packages/kit/src/provider/Container/AppStateLockContainer/index.tsx
```diff
@@ -86,6 +91,32 @@ export function AppStateLockContainer({
...
+  }
+  prevLockedRef.current = isLocked;
+  const deferColdStartChildren = shouldDeferColdStartLockRender({
+    // RAM-based low-end detection is also true on low-RAM Android, but the
+    // jetsam / dual-runtime PBKDF2-starvation problem this guards against is
+    // iOS-specific, so the defer behavior is gated to iOS.
+    isLowEndDevice: IS_LOW_END_DEVICE && platformEnv.isNativeIOS,
+    isLocked,
+    hasUnlockedOnce: hasUnlockedOnceRef.current,
+  });
```
sidmorizon, line 118: **Transient children mount/unmount during atom hydration on low-end iOS**

Type: `Flag`

If the `useAppIsLockedAtom` initially returns `false` during hydration (before persisted password state loads) and then transitions to `true`, children will briefly render on the first frame then unmount when `deferColdStartChildren` kicks in. This is acknowledged in the comments (lines 100-104) and is unavoidable without risking permanently suppressing children for users with no password set. However, depending on how heavy the child tree is, this could still trigger expensive mount-then-unmount work on the exact devices where the optimization aims to reduce load. If this hydration race is common on low-end iOS devices, it may partially defeat the performance goal.

packages/shared/src/utils/lowEndDevice.native.ts
```diff
@@ -0,0 +1,17 @@
...
+//
+// `expo-device` `totalMemory` is in bytes. We use binary GiB (1024^3) so that
+// 3GB-class devices that report ~3.14e9 bytes (e.g. iPhone 7 Plus) are included.
+// Threshold 3.5 * GiB = 3_758_096_384 bytes.
+//
+// Used to defer rendering the full app tree behind the lock screen during a
+// cold-start lock, so the single background JS thread is free to finish the
+// 600k-iteration PBKDF2 `verifyPassword` before iOS jetsam kills the process.
+const GB = 1024 * 1024 * 1024;
+const LOW_END_MEMORY_THRESHOLD = 3.5 * GB; // 3_758_096_384 bytes
```
sidmorizon, line 13: **RAM threshold may exclude 4GB devices that still exhibit jetsam kills**

Type: `Flag`

The threshold of 3.5 GiB (`3_758_096_384` bytes) targets 3GB-class devices like iPhone 7 Plus. However, 4GB devices (iPhone XS, iPhone 11) can also experience jetsam kills under memory pressure if the app tree is heavy enough. The comment at line 7 says the threshold includes '3GB-class devices that report ~3.14e9 bytes', which is correct, but it's worth considering whether the threshold should be higher (e.g. 4.5 GiB) to cover more devices that experience the PBKDF2 starvation problem in practice.

huhuanming:
Closing in favor of folding this fix into #10819 (feat/lazy-tab-preload).

Reason: this PR re-implemented device-RAM detection (`lowEndDevice.native.ts` reading `expo-device` directly) that already exists in #10819 as the shared `devicePerformanceTier` / `getDeviceMemoryGBSync()` primitive. Rather than landing a third RAM-classification source of truth (alongside that helper and `webviewAliveLimit.native.ts`), the cold-start-lock defer fix now lives on feat/lazy-tab-preload and reuses the shared `getDeviceMemoryGBSync()` primitive.

Note it intentionally does NOT use `getDevicePerformanceTier()` (which returns `medium` until post-launch calibration) because the guard must work on the very first boot after upgrade, so it reads memory synchronously each launch with the lock-appropriate <=3.5GB threshold.

PR #12200, opened 2026-06-29
fix: preserve history pagination on auto-refresh(OK-56924)

weatherstar:
[OK-56924](https://onekeyhq.atlassian.net/browse/OK-56924)

----------

## Summary
- Stop the history list from rolling back to the bottom of page 1 when the periodic first-page refresh (polling / focus / visibility / status-change) lands after the user has loaded more pages.
- Preserve the already-visible loaded range by moving rows displaced from the previous first page into the appended range, instead of clearing it.
- Add two safety guards for edge cases the preserve logic introduces: a mid-list gap on large bursts, and stale local-pending rows lingering after replacement.

## Intent & Context
Reported in **OK-56924**: while scrolling through history that has loaded several pages, the 15s auto-refresh of the list would cause the content to "jump back" to the bottom of the first page, making it impossible to keep viewing older records. Expected behavior (from the ticket): refreshing the first page may add new entries at the top, but must not disrupt viewing records beyond the first page.

The list renders two segments: `historyData` (the first page, refreshed periodically) and `appendedTxs` (pages 2, 3, … loaded via load-more), combined with `unionBy([...historyData, ...appendedTxs])`. The container caps the work to those two segments per consumer (`TxHistoryContainer`, `TokenDetailsHistory`).

## Root Cause
Every first-page response — including periodic polling — unconditionally called `startNewPaginationGeneration()`, which `setAppendedTxs([])` and reset the cursor/page counters. So a single background refresh wiped the entire loaded range, shrinking the rendered list back to page 1 and resetting the cursor. This conflated two distinct semantics on one code path: a *hard reset* (identity/filter switch, pull-to-refresh) vs. a *first-page-only refresh* (polling) that should keep the loaded range intact.

## Design Decisions
- **Displaced-rows bridge (preserve path):** on a first-page refresh while a loaded range exists, recompute `appendedTxs = unionBy(prevFirstPage, prevAppended) − newFirstPageIds`. Rows pushed out of the new first page move into the appended range, so the combined list stays contiguous and stable, and the deep cursor is left untouched so load-more continues from where it was. Hard reset is reserved for `reset()` (identity/filter change) and pull-to-refresh.
- **Gap guard (`overlapsPreviousFirstPage`):** the bridge is only gap-free while the refreshed first page still overlaps the previous one (< `HISTORY_PAGE_SIZE` = 50 new txs since last refresh). If the two first pages are fully disjoint (a burst > one page within the 15s window), an unknown slice of history sits between the new first page and the loaded range with no cursor to backfill it. In that rare case we fall back to a hard reset — a one-time scroll-to-top is strictly safer than silently rendering a hole in the middle of history.
- **Stale local-pending guard (`isLocalPendingTx`):** a local pending tx (RBF / speed-up / cancel, or confirmed under a new id) that drops out of the refreshed first page was being carried forward as a displaced row and lingered forever beside its replacement. `appendedTxs` holds on-chain pages where a local pending never legitimately belongs, so it is dropped when it falls out of the new first page. A still-valid pending keeps its id on the refreshed first page, stays in the overlap, and is filtered as an ordinary first-page row — so it is never wrongly removed.
- **Deliberately out of scope:** (1) re-arming `hasMore` from the first page after pagination was exhausted — kept as-is because it doubles as a self-heal/retry path after defensive stops (reorg / duplicate emit); suppressing it would lose that recovery for one redundant request. (2) Visual scroll-position retention on top insertions (`maintainVisibleContentPosition`) — a cross-platform list-layer change, and the ticket explicitly allows first-page additions; deferred unless QA reports a concrete platform regression.

## Changes Detail
- `useHistoryListLoadMore.ts`
  - Split `onFirstPageResponse` into preserve vs. hard-reset paths; preserve keeps the cursor and migrates displaced rows.
  - Added `firstPageTxsRef` / `firstPageIdsRef` so load-more dedupes against the current first page (`!firstPageIdsRef.has` added to the newRows filter) and the preserve path can compute displaced rows.
  - Added `overlapsPreviousFirstPage` guard → hard reset when first pages are disjoint.
  - Added `isLocalPendingTx` filter so a replaced/confirmed local pending cannot linger in the appended range.
  - `onFirstPageResponse` now takes the first-page `txs` in its meta.
- `TxHistoryContainer.tsx`, `TokenDetailsHistory.tsx`: pass `txs` into `onFirstPageResponse`.
- `useHistoryListLoadMore.test.tsx`: regression tests for in-flight refresh, deep-cursor preservation, disjoint-first-page hard reset, and replaced-local-pending drop.

## Risk Assessment
- **Risk Level**: Medium (touches the history pagination state machine, but logic is isolated to one hook and covered by unit tests).
- **Affected Platforms**: Extension / Mobile / Desktop / Web (shared hook; consumers are Home history and Token details history).
- **Risk Areas**: the preserve vs. hard-reset branch selection and the displaced-rows recomputation; the in-flight load-more × first-page-refresh race (covered by tests).

## Test plan
- [ ] Home history: scroll to load several pages, wait for the 15s auto-refresh, confirm the list does NOT jump back to page 1 and older rows remain viewable.
- [ ] Token details history: same as above.
- [ ] Send a tx (local pending appears at top), load more pages, then speed-up/cancel it; confirm the old pending row disappears (no stale duplicate) after refresh.
- [ ] Pull-to-refresh and account/network/filter switch: confirm the list correctly resets to page 1.
- [ ] Load to the very bottom (no more history), confirm reaching the end still works after a refresh.
- [ ] Verify on web, desktop, iOS, Android, and extension.

[OK-56924]: https://onekeyhq.atlassian.net/browse/OK-56924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


revan-zhang:

|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/850a20c4-3831-4d95-b2db-7d122d74568a?refs=0872b03e085d62e06739188d351dcfcff0efe69b&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/rayston92/pr-checks/850a20c4-3831-4d95-b2db-7d122d74568a/license?refs=0872b03e085d62e06739188d351dcfcff0efe69b&source=prChecksComment) |

originalix (approved):

PR #12278, opened 2026-07-03
fix: stock market-closed refresh, recent pairs cap, and mobile stock header alignment (OK-57346)(OK-57347)(OK-57348)

haicongliao:
[OK-57346](https://onekeyhq.atlassian.net/browse/OK-57346) [OK-57347](https://onekeyhq.atlassian.net/browse/OK-57347) [OK-57348](https://onekeyhq.atlassian.net/browse/OK-57348)

----------

## Summary

Three trade/stock fixes:

### OK-57346 — Stock tab market-closed state never refreshes while staying on the page
- The stock tab's token detail (which carries `stock.isOpen` / next-open text) was fetched once per token selection and never again; the market WebSocket only pushes price. A market that closed/reopened while the user stayed on the tab was never reflected (a stale `MarketClosed` stage could never recover since quoting stops in that stage).
- Added a 10s `pollingInterval` to the stock token-detail fetch in `useSwapStockChannel` (same cadence as Swap Pro's detail polling), so the closed alert, the disabled trade button and the K-line pulse dot stay in sync with the actual market session.
- Hardened the polling path: a transient fetch failure keeps the last successful payload for the same token (instead of degrading the channel to `MissingStock`), a superseded in-flight response can no longer clobber the snapshot after a token switch, and the channel status is driven by first-result arrival rather than per-tick loading so degraded states don't blink every 10s.

### OK-57347 — Recent trade pairs list needs a display cap
- Since Bridge was merged into the Swap entry (#11897), the merged tab renders the whole stored list — up to 20 pairs (the store keeps single-chain and cross-chain buckets of 10 each). The old per-tab type filter that effectively capped the list at 10 was removed in that merge.
- Cap the rendered list at `maxRecentTokenPairs` (10) in `SwapRecentTokenPairsGroup`.
- The store previously re-concatenated single-chain pairs ahead of cross-chain ones on every write, which would have let a full single-chain bucket permanently hide cross-chain recents from the top-10. Both writers (`ServiceSwap.swapRecentTokenPairsUpdate`, `SimpleDbEntitySwapConfigs.addRecentTokenPair`) now share `capRecentTokenPairsPreservingOrder` (unit-tested), which keeps global newest-first order while still retaining up to 10 entries per type in storage — the displayed top-10 mixes types by pure recency (e.g. 8 single + 2 cross), no timestamps or migration needed.

### OK-57348 — Mobile stock header should match Pro mode's token selector
- The stock symbol on the mobile Stocks tab rendered at `$headingSm` (14px) vs Pro mode's `$headingLg` (18px), and sat 7px lower on screen (page top padding + fixed 52px row height + pill padding vs Pro's 8px offset), so the instrument name visibly jumped when switching between the Stocks and Pro tabs with the same instrument.
- `StockMarketTokenHeader` gains a `proAligned` prop used only by the mobile layout: symbol at `$headingLg` with a 160px cap, intrinsic row height, and the page top padding adjusted so the symbol sits at the same 8px offset as Pro. The desktop stock card keeps its compact heading and fixed row height. The loading skeleton mirrors the same layout.

## Jira
- https://onekeyhq.atlassian.net/browse/OK-57346
- https://onekeyhq.atlassian.net/browse/OK-57347
- https://onekeyhq.atlassian.net/browse/OK-57348

## Test
- `yarn tsc:staged` / `yarn lint:staged` clean on all touched files
- Related unit tests pass (`swapProStockMarketClosed`, `SwapStockTradeAlert.utils`, `resolveStockMarketStatusCase`, `swapMarketHistory`, `SwapStockDesktopContainer.utils`)
- Manual QA checklist: ① stay on the stock trade tab across a market open/close boundary and confirm the alert/button/pulse dot flip within ~10s; ② switch Stocks ↔ Pro with the same instrument and confirm the symbol no longer jumps; ③ recent trades expand to at most 10 chips

[OK-57346]: https://onekeyhq.atlassian.net/browse/OK-57346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57347]: https://onekeyhq.atlassian.net/browse/OK-57347?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57348]: https://onekeyhq.atlassian.net/browse/OK-57348?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


haicongliao:
CleanShot 2026-07-03 at 17 16 36@2x

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,59 +160,89 @@ export function useSwapStockChannel() {
...
           perpsInfo: undefined,
-        },
-        watchLoading: !!currentStockTokenKey,
-        swrKey: stockTokenDetailScope
-          ? swrKeys.swapStockTokenDetail({
-              tokenScope: stockTokenDetailScope,
-            })
-          : undefined,
+        };
+      }
```
sidmorizon, line 279: **Polling error resilience keeps stale data visible indefinitely if the endpoint stays broken**

Type: `Flag`

The `lastGoodStockTokenDetailRef` pattern at `packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts:209-211` returns the last successful response when a polling tick fails, preventing the UI from degrading to MissingStock on transient errors. However, if the endpoint returns errors persistently (e.g. the token is delisted or the backend is down for an extended period), the UI will continue showing the last-good data indefinitely — including a potentially stale `isOpen` status. The 10s polling cadence means the error path runs frequently, but since `usePromiseResult` applies the returned `lastGood` object as the new result each time, the downstream `stockTokenDetail` and `stockMarketStatus` remain populated. This is arguably better than the alternative (UI flickers to unavailable), but reviewers should confirm that showing stale market-open/closed state for minutes or hours is acceptable.

haicongliao: Bounded in 41669938eb: the last-good fallback now has a 60s TTL (6 poll ticks) — transient blips still ride on the cached payload, but a persistently broken endpoint degrades the channel to the stable unavailable state instead of showing stale open/closed data indefinitely. FWIW Swap Pro's 10s detail polling keeps its stale atom forever on persistent failure, so this surface is now stricter than the existing precedent.

originalix (dismissed):

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,59 +169,101 @@ export function useSwapStockChannel() {
...
   const stockTokenDetailScope = currentStockTokenKey;
-  const { result: stockTokenDetailState, isLoading: stockTokenDetailLoading } =
-    usePromiseResult(
-      async () => {
-        if (!currentStockToken?.networkId || !currentStockTokenKey) {
-          return {
-            scope: stockTokenDetailScope,
-            token: undefined,
-            perpsInfo: undefined,
+  const lastGoodStockTokenDetailRef = useRef<{
```
originalix: 这里的 `lastGoodStockTokenDetailRef` 只会在本次挂载里的成功请求后才赋值，但 `usePromiseResult` 会先用 `swrKey` 把上一轮成功的 detail 同步读回来。

这样用户重新进入 Stock 页面时，如果首个轮询请求刚好失败，catch 分支拿不到 last-good，就会直接返回 `{ token: undefined }`，`channelStage` 立刻退化成 `MarketUnavailable`，瞬时网络抖动还是会把交易入口清空。

建议在当前 scope 已经命中有效 `stockTokenDetailState`/SWR 缓存时，先把这份结果预热到 `lastGoodStockTokenDetailRef`，再复用同一套 TTL 回退逻辑。

packages/kit/src/views/Swap/components/SwapRecentTokenPairsGroup.tsx
```diff
@@ -42,7 +43,12 @@ const SwapRecentTokenPairsGroup = ({
   const fromTokenAmountBN = new BigNumber(fromTokenAmount ?? 0);
   const tokenPairsInCurrentType = useMemo(() => {
     if (visibleSwapTypes.includes(swapTypeSwitchAtom)) {
-      return tokenPairs;
+      // The store keeps single-chain and cross-chain pairs in separate
+      // buckets (up to maxRecentTokenPairs each), so the merged Swap &
+      // Bridge tab receives up to 2x maxRecentTokenPairs entries. Cap the
+      // rendered list so "show more" never expands beyond one bucket's
+      // worth of pairs (OK-57347).
+      return tokenPairs.slice(0, maxRecentTokenPairs);
```
originalix, line 51: `tokenPairs` 在存储层不是按全局最近时间排序的，而是先拼 `singleChainTokenPairs` 再拼 `crossChainTokenPairs`。

这里直接 `slice(0, maxRecentTokenPairs)`，在单链记录达到 10 条后会把后面的跨链记录整体截掉，即使最新的一笔其实是刚做过的 Bridge 交易，recent list 里也只会留下更旧的单链记录。

建议不要在 UI 层直接截断这个拼接数组；要么在写入端先生成真正按最近时间排序的前 10 条，要么给展示层补上时间信息后再重排。

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
haicongliao: Fixed in ed87722055: when the in-memory snapshot is empty/mismatched for the current scope (fresh mount), the catch path now warms it from the same SWR cache entry via swrCacheUtils.getWithTimestamp — the cache's own timestamp feeds the existing 60s TTL, so a failing first tick after remount keeps the hydrated UI when the cache is fresh, while a genuinely stale cache still degrades to unavailable (consistent with the TTL bound agreed above).

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,59 +178,115 @@ export function useSwapStockChannel() {
...
+            ? swrCacheUtils.getWithTimestamp<IStockTokenDetailFetchState>(
+                swrKeys.swapStockTokenDetail({
+                  tokenScope: stockTokenDetailScope,
+                }),
+              )
+            : undefined;
+          if (cached?.data?.scope === stockTokenDetailScope) {
+            lastGood = { state: cached.data, updatedAt: cached.updatedAt };
+            lastGoodStockTokenDetailRef.current = lastGood;
+          }
```
sidmorizon, line 259: **Stale market status can reappear on page re-entry because the fallback cache timestamp is silently refreshed**

Severity: `non-severe`

The stale-data fallback returned from the error handler is written back to the persistent cache with a fresh timestamp (`swrCacheUtils.set` at `packages/kit/src/hooks/usePromiseResult.ts:303`), so on component remount the cache-warming path (`packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts:247`) reads a refreshed `updatedAt` that passes the 1-minute TTL check even though the original successful fetch was much older.

Impact: After a network outage, re-entering the stock trade page can show an outdated market-open/closed indicator for up to another minute, potentially letting users attempt trades against a closed market.

<details><summary>Mechanism: usePromiseResult refreshes the SWR cache timestamp on every non-undefined return, including stale fallback data</summary>

When the fetch fails, the catch block at `useSwapStockChannel.ts:224-261` returns `lastGood.state` (the last successful payload). `usePromiseResult` treats this as a normal successful result and writes it to the SWR cache via `swrCacheUtils.set(capturedSwrKey, r)` at `usePromiseResult.ts:303`, which stamps it with `Date.now()`.

On remount, `lastGoodStockTokenDetailRef` is null (fresh `useRef`). The catch block at line 233 falls through to the SWR cache warming path at line 239-249. It reads the cached entry whose `updatedAt` was refreshed by the last failed-poll cycle, so the TTL check at line 251-253 (`Date.now() - lastGood.updatedAt <= SWAP_STOCK_DETAIL_LAST_GOOD_TTL_MS`) passes even though the data is arbitrarily old. The warmed ref also inherits the refreshed timestamp, so subsequent failures within the same mount also pass the TTL.

This creates a cycle: each mount gets up to 60s of stale data, the SWR cache timestamp is refreshed during that window, and the next remount reads the refreshed timestamp — effectively making the 1-minute TTL renewable indefinitely across remounts.

</details>

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,59 +178,115 @@ export function useSwapStockChannel() {
...
+      // the disabled trade button and the K-line pulse dot in sync with
+      // the actual market session (OK-57346). 10s matches Swap Pro's
+      // token-detail polling cadence.
+      pollingInterval: timerUtils.getTimeDurationMs({ seconds: 10 }),
+      swrKey: stockTokenDetailScope
+        ? swrKeys.swapStockTokenDetail({
+            tokenScope: stockTokenDetailScope,
+          })
+        : undefined,
+    },
```
sidmorizon, line 305: **10-second polling interval adds continuous network requests while the stock tab is mounted**

Type: `Flag`

The new `pollingInterval: timerUtils.getTimeDurationMs({ seconds: 10 })` at `packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts:282` means `fetchMarketTokenDetailByTokenAddress` fires every 10 seconds while the stock trade tab is mounted and focused. The `usePromiseResult` hook pauses polling when the page loses focus (via the `defer.promise` gate at `packages/kit/src/hooks/usePromiseResult.ts:374`) and when deps change (nonce bump invalidates the old chain). This matches the comment's claim that it aligns with Swap Pro's token-detail polling cadence. Worth noting: on mobile, if the user leaves the stock tab open but switches to another app tab within the same navigation stack, the `useIsFocused` check may still consider it focused depending on the navigation structure, which could keep polling active in the background.

packages/kit/src/views/Swap/components/SwapRecentTokenPairsGroup.tsx
sidmorizon: **Token pairs cap at maxRecentTokenPairs may silently drop cross-chain pairs that were previously visible**

Type: `Flag`

The change at `packages/kit/src/views/Swap/components/SwapRecentTokenPairsGroup.tsx:51` caps the rendered list to `maxRecentTokenPairs` (10). The comment explains that the store keeps single-chain and cross-chain pairs in separate buckets (up to 10 each), so the merged tab could receive up to 20 entries. The `slice(0, 10)` means the first 10 entries are shown regardless of their type. If the store returns single-chain pairs first, all 10 slots could be consumed by single-chain pairs, hiding all cross-chain pairs. Conversely, if cross-chain pairs come first, single-chain pairs could be hidden. The ordering of the merged list determines which pairs are visible, and this ordering is not controlled by this component. This is likely acceptable since the cap is meant to prevent UI overflow, but it's worth verifying that the upstream merge order is intentional.

haicongliao: Fixed at the write side in ea6858783c: extracted capRecentTokenPairsPreservingOrder (packages/shared/src/utils/swapRecentTokenPairsUtils.ts, unit-tested) and replaced the bucket-split-and-concat in both ServiceSwap.swapRecentTokenPairsUpdate and SimpleDbEntitySwapConfigs.addRecentTokenPair — insertion order (newest-first) is now preserved globally while each type still retains up to 10 entries in storage, so the UI's top-10 mixes types by pure recency (a just-completed bridge pair always surfaces; composition floats, e.g. 8 single + 2 cross). No timestamps or migration needed; legacy bucket-ordered data heals as new trades re-insert at the head.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
haicongliao: Fixed in 825d7d71fb: the TTL now trusts a fetchedAt timestamp carried inside the payload itself, stamped only on a real successful fetch. The SWR entry timestamp (re-stamped whenever the fallback result is re-persisted) is no longer consulted, and legacy cache entries without fetchedAt are ineligible for the fallback — so the 60s bound holds across remounts and cannot be renewed by failed polls.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
haicongliao: No change: this is the intended design. The fetch uses usePromiseResult's default checkIsFocused, so poll ticks gate on useRouteIsFocused — react-navigation blurs the swap route when another bottom tab is selected, pausing the poll — and the visibility defer suspends polling when the app/window is backgrounded. Cadence and pause semantics are identical to Swap Pro's existing 10s token-detail polling (useSwapPro.ts).

packages/kit/src/views/Swap/components/SwapRecentTokenPairsGroup.tsx
haicongliao: Addressed in ea6858783c (this comment was generated against the previous head): both store writers now share capRecentTokenPairsPreservingOrder, which keeps global newest-first order (per-type retention caps only drop each bucket's oldest entries), so the slice(0, 10) here shows the true most-recent 10 with the type mix floating by actual usage — a just-completed cross-chain pair always surfaces.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,59 +185,119 @@ export function useSwapStockChannel() {
...
       },
-    );
+      // Market open/closed state (stock.isOpen / description) is only
+      // carried by this endpoint — the market WebSocket pushes price only —
+      // so poll it while the tab stays mounted to keep the closed alert,
+      // the disabled trade button and the K-line pulse dot in sync with
+      // the actual market session (OK-57346). 10s matches Swap Pro's
+      // token-detail polling cadence.
+      pollingInterval: timerUtils.getTimeDurationMs({ seconds: 10 }),
+      swrKey: stockTokenDetailScope
```
originalix, line 300: 当用户重新进入一个之前打开过的 Stock 页面时，`usePromiseResult` 会先根据 `swrKey` 把 MMKV 里的 `swapStockTokenDetail` 同步回 `stockTokenDetailState`。这里新增的 TTL 判断只覆盖 `catch` 分支里的手动 cache warm，没有约束这次首帧水合。

结果是：只要缓存里还留着同 scope 的旧 `token.stock.isOpen/description`，`marketStatusStatus` 就会立刻把它当成当前状态，闭市提示和交易按钮禁用态可能先展示昨天的结果；如果请求挂起或网络很慢，这个错误状态会持续到请求真正返回，而不是被 1 分钟 TTL 截断。

建议在消费 SWR 缓存之前也按 `fetchedAt` 做一次 TTL 过滤；过期缓存应回退到 pending/empty，而不是直接进入 `Ready`。

haicongliao: Fixed in 898170f2d0: the consumer-side derivation now gates every token-carrying state on the in-payload fetchedAt TTL — including the first-frame SWR hydration. A fresh cache (<60s) still renders instantly; an expired or legacy hydration no longer reaches marketStatusStatus and instead keeps stockTokenDetailPending true, so the channel shows Initializing until the first real fetch resolves (and honest empty results still settle to unavailable rather than pending during an extended outage).

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,70 +185,142 @@ export function useSwapStockChannel() {
...
+  // post-TTL fallback settled on unavailable) or a token payload whose
+  // fetchedAt is within the TTL. The freshness gate matters for the SWR
+  // first-frame hydration: usePromiseResult syncs the MMKV cache into the
+  // render state before any request resolves, and without this check a
+  // stale cached isOpen/description would drive the closed alert and the
+  // trade button until a slow/hung request returns. An expired hydration
+  // instead stays pending (Initializing) until the first real result.
+  const stockTokenDetailLanded =
+    stockTokenDetailState?.scope === stockTokenDetailScope &&
+    (!stockTokenDetailState.token ||
```
originalix: 当 detail 接口曾经连续失败超过 TTL，并把 catch 分支产出的 `{ token: undefined }` 写进 SWR 后，用户重新进入这个 stock 页面时会先走到这里。

这里把所有 `token === undefined` 的 state 都当成 landed，没有区分“真实请求返回空结果”和“上一次故障后的 fallback 空结果”。而 `usePromiseResult` 会把这个缓存同步回填到首帧。

结果是服务端已经恢复、或者新请求还没返回时，页面也会先进入 `MarketUnavailable`，错误地展示“不支持交易”并禁用按钮。

建议让空态也绑定来源：至少要求它带有真实响应写入的 `fetchedAt`，或者显式区分 success-empty / fallback-empty，只把真实空结果判定为 landed。

haicongliao: Fixed in 459c7974c7: landed now requires a fetchedAt stamped by a real server response AND within the TTL — uniformly for token payloads and empty answers. The post-TTL fallback empty deliberately carries no fetchedAt (comment added at the return site), so whether it is served in-mount or hydrated back from the SWR cache on re-entry, it keeps the channel pending/Initializing until the first real request resolves; only a genuine success-empty (asset is not a stock) settles to MarketUnavailable. Chosen semantics for an extended outage: a steady Initializing state (retrying every 10s) rather than claiming the market is unavailable.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -155,70 +185,149 @@ export function useSwapStockChannel() {
...
           token: undefined,
           perpsInfo: undefined,
-        },
-        watchLoading: !!currentStockTokenKey,
-        swrKey: stockTokenDetailScope
-          ? swrKeys.swapStockTokenDetail({
-              tokenScope: stockTokenDetailScope,
-            })
-          : undefined,
+        };
```
sidmorizon, line 278: **Stock trade panel shows a perpetual loading spinner instead of settling to 'unavailable' after prolonged backend failures**

Severity: `severe`

The post-TTL error fallback is emitted without a timestamp (`swapRecentTokenPairsUtils.ts` pattern at `packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts:272-276`) so it never counts as "landed", which keeps the detail in a permanent pending state.

Impact: When the market-detail endpoint is down or a token is delisted, the trade UI shows an infinite "checking market status" spinner instead of the intended "market unavailable" message, blocking the user from understanding the situation.

<details>
<summary>Mechanism: post-TTL empty fallback keeps stockTokenDetailPending true forever</summary>

After the 1-minute TTL expires (6+ consecutive poll failures), the catch block at line 261-276 falls through to the final return that deliberately omits `fetchedAt`. This empty result is set as `stockTokenDetailState` by `usePromiseResult`.

On the next render:
1. `stockTokenDetailLanded` at line 318-322 evaluates to `false` because `!!stockTokenDetailState.fetchedAt` is `false`.
2. `stockTokenDetailPending` at line 329-330 becomes `true` (since `currentStockTokenKey` is truthy and `!stockTokenDetailLanded` is true).
3. In the `marketStatusStatus` memo at line 621-644, `activeStockTokenDetail?.stock` is undefined (no landed detail), and `stockTokenDetailPending` is `true`, so it returns `Initializing`.
4. In the `channelStage` memo at line 646-678, `marketStatusStatus === Initializing` maps to `CheckingMarketStatus`.

Every subsequent 10s poll failure returns the same empty fallback without `fetchedAt`, so the channel is stuck in `CheckingMarketStatus` forever — never reaching `MarketUnavailable` as the comments at lines 232-234 and 268-271 claim.

The fix is to include `fetchedAt: Date.now()` on the post-TTL empty fallback so it "lands" as a real (empty) answer. This makes `stockTokenDetailPending` false, `marketStatusStatus` returns `Empty`, and `channelStage` correctly settles to `MarketUnavailable`.
</details>

Suggested fix:

```suggestion
        return {
          scope: stockTokenDetailScope,
          token: undefined,
          perpsInfo: undefined,
          fetchedAt: Date.now(),
        };
```

haicongliao: Fixed in d38f5d2205, but not with the suggested fetchedAt: Date.now() stamp — that would reintroduce the bug fixed in 459c7974c7 (the persisted fallback would hydrate on remount as a fresh landed empty and wrongly claim MarketUnavailable for up to 60s even after the backend recovered). Instead the fallback empty now carries the producing mount's id: within that mount it counts as landed, so an extended outage settles to a stable MarketUnavailable as you require; a persisted copy hydrated on a later mount carries a foreign mount id and stays pending until the first real response. Both this finding and the remount-recovery finding hold simultaneously.

zhaono1 (approved):

originalix (approved):

PR #12279, opened 2026-07-03
chore: add diagnostic logs for wallet backup status migration flag loss

sidmorizon:
## Background

QA reproduced (iOS, fresh install): create a new HD wallet → kill the app → relaunch → the wallet shows as "backed up" with no user action. Log analysis points to `migrateHdWalletsBackedUpStatus` re-running at the second boot because the `hdWalletsBackupMigrated` flag written at the first boot is missing from storage — suspected AsyncStorage key loss across the two native JS runtimes (main + bg each hold their own native instance whose in-memory manifest snapshot rewrites `manifest.json` as a whole on every write on iOS).

This PR adds diagnostic logging (no behavior change to the migration) so the next reproduction proves or disproves that chain from `app-latest.log` alone, plus one runtime guard hardening.

## What's added

All log events are local-only (`@LogToLocal`), no server reporting, and never log values or sensitive data. Marked `TODO(cleanup)` — to be removed once the root cause is fixed.

1. **`backupMigrationStatusCheck`** (bg, every boot, in `migrateHdWalletsBackedUpStatus`): boot-time raw presence of the `simple_db_v5:appStatus` entity, the migrated flag value, and the un-backed-up HD wallet ids the migration is about to touch. The raw sample is a single-key `getItem` on the entity's own backing storage instance, kicked by `ServiceBootstrap.initDeferred()` **before** any deferred task starts, so concurrent migrations that also write `appStatus` cannot pollute the pre-boot on-disk reading.
2. **`backupMigrationMarkedWallets`** (bg): the exact wallet ids the migration itself flipped to backed-up — the smoking-gun line.
3. **`backupMigrationFlagProbe`** (bg): write-then-read-back verification right after the flag write (`stage=afterMigrationFlagWrite`), and a re-check right after every HD wallet creation (`stage=afterCreateHDWallet`). The wallet-creation probe is gated on the migration having settled this session, so an unwritten flag can never be misread as an in-session loss.
4. **Guard hardening** (`simpleDb.ts`): direct use of the real `simpleDb` on the native UI/main runtime (when the split background thread is enabled) now throws lazily on first property access, mirroring `ensureLocalDbNotOnNativeMainThread` for localDb. Previously only extension UI was blocked; on native main the single-writer rule was convention only. The single-runtime native fallback and jest are unaffected; the throw is lazy so merely bundling the module cannot crash startup.

An earlier revision also traced every JS-side AsyncStorage write at the module level; it was removed (too noisy, monkey-patch approach) — identifying the exact clobbering writer will be done via native-level verification (manifest.json inspection / fork source review) if needed.

## How to read the logs after a repro

- First boot: `backupMigrationFlagProbe stage=afterMigrationFlagWrite flagPresent=true` proves the flag write reached the storage layer.
- Same session: `stage=afterCreateHDWallet` shows whether the flag is still present at wallet-creation time.
- Second boot: `backupMigrationStatusCheck` with `appStatusRawExists=false` / `migratedFlag=undefined` while `backupMigrationMarkedWallets` lists the new wallet ⇒ the flag was lost at the persistence layer between launches (AsyncStorage loss confirmed).

## Verification

- `npx tsgo -p ./tsconfig.json --noEmit` passes (including after merging base `x`).
- `oxlint --type-aware --deny-warnings` on the changed files: 0 warnings / 0 errors.
- `jest packages/shared/src/logger/scopes/account/scenes/wallet.test.ts` and `jest packages/kit-bg/src/dbs/simple`: all pass.

## Notes

- No behavior change to the migration logic, ordering, or event emissions; all probes are wrapped in try/catch and must never block business flow.
- Review feedback addressed: probe reads pinned to the entity's backing storage (`$webStorageSimpleDB` on web/desktop/ext), `getAllKeys()` removed from the resident boot path, `afterCreateHDWallet` gated on migration settle, boot raw sample moved ahead of the deferred batch.
- Follow-up (separate PRs): harden `migrateHdWalletsBackedUpStatus` so flag loss can never flip freshly created wallets, and migrate simpleDb off AsyncStorage (e.g. MMKV) on native.


packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -7046,7 +7118,22 @@ class ServiceAccount extends ServiceBase {
         };
       }
     }
+    defaultLogger.account.wallet.backupMigrationStatusCheck({
+      migratedFlag: appStatus?.hdWalletsBackupMigrated,
+      appStatusRawExists,
+      appStatusKeyListed,
+      allKeysCount,
+      hdWalletsCount: wallets.filter((w) => w.type === WALLET_TYPE_HD).length,
+      unbackedUpHdWalletIds: Object.keys(walletsBackedUpStatusMap),
```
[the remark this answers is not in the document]
sidmorizon: Keeping the wallet ids intentionally. Three reasons:

1. HD wallet ids are device-local ordinals (`hd-1`, `hd-2`, built from `nextHD` in `accountUtils.buildHdWalletId`) — they contain no key material, addresses, or user-identifying data, and are meaningless outside the device.
2. This identifier class is already part of the established logging posture: `backupCompleted` in the same scene logs `walletId` with `@LogToServer` (server-side analytics), and import task logs also persist wallet ids. These new events are local-only (`@LogToLocal`), strictly below that bar.
3. Counts would defeat the diagnostic's purpose: the investigation needs to correlate *which* wallet the migration flipped with the wallet the user just created (multi-wallet devices). A count can't distinguish them.

---

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -7026,10 +7036,72 @@ class ServiceAccount extends ServiceBase {
...
   }
 
+  // Diagnostic probe for the wallet backup-status investigation: reads the
+  // persisted appStatus entity raw from appStorage — bypassing the simpleDb
+  // JS-level cache — and logs whether hdWalletsBackupMigrated is present at
+  // the storage layer. Runs in the bg runtime only (this service lives in bg).
+  async logBackupMigrationFlagProbe({ stage }: { stage: string }) {
+    try {
+      const entityKey = simpleDb.appStatus.entityKey;
+      const raw = (await appStorage.getItem(entityKey)) as unknown;
```
sidmorizon: # [🟡 P2] 探针在 web/桌面/插件端读错了存储实例

## 问题

探针用默认的 `appStorage.getItem(entityKey)` 读取原始数据（本行 `7046`，以及迁移函数里的 `7086`）。但 `SimpleDbEntityBase` 实际持久化用的是 `storageHub.$webStorageSimpleDB || storageHub.appStorage`。

在 web / 桌面 / 扩展端，`$webStorageSimpleDB` 是一个独立的（IndexedDB 支撑的）存储实例，与默认 `appStorage` 不是同一个 store：

```ts
// SimpleDbEntityBase.ts
appStorage = storageHub.$webStorageSimpleDB || storageHub.appStorage;
```

只有在原生（iOS/Android）上 `$webStorageSimpleDB` 为 `undefined`，探针才和 simpleDb 落在同一个 store。

## 影响

在 web/桌面/扩展端，即使 `hdWalletsBackupMigrated` 真实存在于 simpleDb，探针也会稳定读到空值，导致 `rawExists` / `flagPresent` / `appStatusRawExists` 恒为 `false`，日志给出与真实持久化相反的结论。虽然本次排查目标是 iOS（此路径正确），但这些误导性日志会污染其它平台的 `app-latest.log`，让读日志的人误判「flag 丢失」。

## 建议

统一从 simpleDb 实际使用的存储实例读取，例如：

```ts
const store = storageHub.$webStorageSimpleDB || storageHub.appStorage;
const raw = (await store.getItem(entityKey)) as unknown;
```

两处读取（`7046` 与 `7086`）都需同步调整。

[Claude session ↗](https://claude.ai/code/session_017s5x6LoFrKWVfEiDqwCSY3)

---

sidmorizon: Fixed in 637be59c. Both raw reads (the probe helper and the boot-time check in `migrateHdWalletsBackedUpStatus`) now go through `simpleDb.appStatus.appStorage` — the entity's own backing storage field — instead of the default `appStorage`. Using the entity field rather than re-deriving `storageHub.$webStorageSimpleDB || storageHub.appStorage` keeps the probe pinned to whatever store the entity actually writes to, even if the storage wiring changes later. `getAllKeys` also moved to the same instance. tsc + oxlint clean.

---

packages/shared/src/logger/scopes/account/scenes/wallet.ts
```diff
@@ -189,6 +189,46 @@ export class WalletScene extends BaseScene {
...
+  }
+
+  @LogToLocal()
+  public backupMigrationFlagProbe(params: {
+    stage: string;
+    rawExists: boolean;
+    flagPresent: boolean;
+  }) {
+    return params;
+  }
```
sidmorizon, line 231: **Diagnostic code is temporary — consider a removal plan**

Type: `Flag`

The three new logger methods (`backupMigrationStatusCheck`, `backupMigrationMarkedWallets`, `backupMigrationFlagProbe`) and their call sites are explicitly described as investigation aids for a suspected iOS AsyncStorage key-loss bug (lines 192-204 of `packages/shared/src/logger/scopes/account/scenes/wallet.ts`). Once the root cause is identified and fixed, this diagnostic scaffolding should be removed to avoid permanent runtime overhead (raw storage reads, `getAllKeys()` enumeration on every boot). Consider adding a TODO or tracking issue for cleanup.

sidmorizon: Done in fde872d5 — added `TODO(cleanup)` markers on both the logger scene block and the ServiceAccount probe, stating that the three events, their call sites, and the boot-time raw check are temporary scaffolding to be removed once the flag-loss root cause is fixed.

---

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -4072,6 +4072,15 @@ class ServiceAccount extends ServiceBase {
...
       appEventBus.emit(EAppEventBusNames.WalletUpdate, undefined);
     }
+
+    // Wallet backup-status diagnostics: check whether the migration flag is
+    // still present at the storage layer right after wallet creation. If it
+    // was verified after the boot-time write but is already gone here, the
+    // loss happened within this session; if it is still present here but
+    // missing at the next boot, the loss happened at the file level between
+    // launches. Fire-and-forget: must never delay wallet creation.
+    void this.logBackupMigrationFlagProbe({ stage: 'afterCreateHDWallet' });
```
originalix: 这里无条件发 `afterCreateHDWallet` probe，会把 bootstrap 并发顺序带进日志语义里。

`migrateHdWalletsBackedUpStatus()` 是 `ServiceBootstrap.initDeferred()` 里的延后任务，不保证在用户创建钱包前已经完成过 `afterMigrationFlagWrite`。如果这次启动里 flag 还没写过，这里打出的 `flagPresent=false` 只是“尚未写入”，但后面读日志时会被误解成“本会话内把 flag 弄丢了”。这样排障链路最关键的因果判断会被污染。

建议只在本会话已经完成过 `afterMigrationFlagWrite` 之后再发这条 `afterCreateHDWallet` probe。

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -7026,10 +7035,86 @@ class ServiceAccount extends ServiceBase {
...
+    let allKeysCount = -1;
+    try {
+      // Read from the entity's own backing storage instance (NOT the default
+      // appStorage): on web/desktop/extension simpleDb persists to a
+      // dedicated store ($webStorageSimpleDB).
+      const entityKey = simpleDb.appStatus.entityKey;
+      const entityStorage = simpleDb.appStatus.appStorage;
+      const raw = (await entityStorage.getItem(entityKey)) as unknown;
+      appStatusRawExists = !isNil(raw);
+      const allKeys = await entityStorage.getAllKeys();
```
originalix: 这里为了诊断额外跑了一次 `entityStorage.getAllKeys()`。

问题是 `hdWalletsBackupMigrated` 是否还在，`getItem(entityKey)` 已经能回答；整库枚举只是在补 `appStatusKeyListed/allKeysCount` 这类旁证。把这一步放在每次 deferred bootstrap 的常驻路径里，会把存储 I/O 放大到 O(N)，在 key 多的设备上增加启动成本，也可能改变你正在排查的 AsyncStorage 时序。

建议把常驻路径里的 `getAllKeys()` 去掉，只保留 `getItem(entityKey)` 这条最小读取。

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
sidmorizon: 已在 64c25bd2 修复。新增会话级门控 `backupMigrationSettledThisSession`:在 `migrateHdWalletsBackedUpStatus` 的两个出口(已迁移早退、完整执行并通过写后读回)都会置位,`afterCreateHDWallet` probe 只在置位后才发。这样该 probe 的 `flagPresent=false` 语义收敛为"本会话内确实丢了",不会再被 bootstrap 并发顺序污染。早退路径也置位是刻意的:那条路径意味着启动时标志在盘上存在,此后建仓探测到 false 同样是有效的会话内丢失信号。

---

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
sidmorizon: 已在 64c25bd2 移除。常驻启动路径只保留 `getItem(entityKey)` 单 key 最小读取,`getAllKeys()` 及 `appStatusKeyListed`/`allKeysCount` 两个旁证字段(含 logger scene 签名)一并删除。

---

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -7026,10 +7041,89 @@ class ServiceAccount extends ServiceBase {
...
+        flagPresent: Boolean(persistedData?.hdWalletsBackupMigrated),
+      });
+    } catch {
+      // diagnostics must never break the business flow
+    }
+  }
+
   @backgroundMethod()
   async migrateHdWalletsBackedUpStatus() {
+    // Boot-time on-disk truth: this runs right after the runtime starts, so
```
originalix: 这里把 `appStatusRawExists` 当成“上一轮启动后真正落盘的原始状态”，但这段代码本身是 `ServiceBootstrap.initDeferred()` 里的一个 `Promise.all` 子任务。像 `serviceHistory.migrateFilterScamHistorySetting()`、`serviceSetting.migrateBTCFreshAddressSetting()`、`serviceHardware.removeDeviceHomeScreen()` 这些并发 deferred migration 也会写 `simpleDb.appStatus`，只要它们先一步 `setRawData()`，这里的 `getItem(entityKey)` 就会把“本次启动刚写出来的 appStatus”读成“上次启动遗留的 appStatus`。

这样一来，`backupMigrationStatusCheck` 里的 `appStatusRawExists=true` / `migratedFlag=undefined` 会把“本会话新建了 appStatus 但还没写 backup flag”误判成“磁盘上原本有 appStatus，只是 `hdWalletsBackupMigrated` 丢了”，会直接污染这次排障最核心的时间线。

更稳妥的修法是把这次原始 `getItem(entityKey)` 采样前移到 `ServiceBootstrap.initDeferred()` 启动 `Promise.all` 之前，并把结果传进 `migrateHdWalletsBackedUpStatus()`，保证它读到的确实是任何并发 migration 触碰前的 boot-time 状态。

sidmorizon: 已在 7d5f448d 修复,按你的方案前移采样:

- `ServiceBootstrap.initDeferred()` 在启动任何 deferred 批次(包括第一段 `Promise.all`)之前,同步调用 `serviceAccount.startBackupMigrationBootRawSample()` 发起单 key `getItem` 采样。async 函数体在首个 await 前同步执行,所以存储读操作在任何并发 migration 的 `setRawData` 入队之前就已入队;同一存储实例按下发顺序串行处理,读到的必然是本次启动前的落盘状态。采样只是同步发起、不 await,不增加 bootstrap 耗时。
- 采样结果缓存在 promise 字段里,`migrateHdWalletsBackedUpStatus` 消费它;函数内保留幂等的 self-start 仅作为非 bootstrap 调用方(如测试)的兜底,注释里标明了该兜底路径可能有并发写者。
- 采样失败返回 `undefined`(区别于"key 不存在"的 `false`),不会把读取异常误报成 key 丢失。

---

packages/kit-bg/src/services/ServiceBootstrap.ts
```diff
@@ -75,6 +75,14 @@ class ServiceBootstrap extends ServiceBase {
   public async initDeferred() {
     const deferredStart = Date.now();
 
+    // Wallet backup-status diagnostics: sample the persisted appStatus raw
+    // BEFORE any deferred task runs — several concurrent migrations below
+    // write simpleDb.appStatus, and a sample taken after their setRawData
+    // would misreport a freshly-created appStatus as pre-boot on-disk state.
+    // Synchronous kick (the read is awaited later inside
+    // migrateHdWalletsBackedUpStatus), so bootstrap is not delayed.
+    this.backgroundApi.serviceAccount.startBackupMigrationBootRawSample();
```
sidmorizon, line 84: **Boot-time sample ordering relies on storage serialization assumption**

Type: `Flag`

The comment at `ServiceAccount.ts:7051-7057` states that the storage instance serializes operations in issue order, so a sample issued first reads the pre-boot on-disk state. The `startBackupMigrationBootRawSample()` is kicked at `ServiceBootstrap.ts:78` before any `timedDeferred` tasks run. However, the deferred tasks include concurrent migrations that also write `simpleDb.appStatus` (e.g. `mergeAllHdDuplicateWallets` at `ServiceAccount.ts:6959`). If the underlying storage does not guarantee strict FIFO ordering of async operations (e.g. if `getItem` and `setRawData` go through different code paths or microtask scheduling), the sample could theoretically read a post-write state. For diagnostics-only code this is acceptable, but the assumption is worth verifying if the investigation results seem inconsistent.

sidmorizon: Acknowledged — keeping as-is (no code change). The ordering holds by construction on both backends, not just by assumption:

1. The sample's `getItem` is dispatched synchronously inside `startBackupMigrationBootRawSample()` (an async function body runs synchronously up to its first `await`), before any `timedDeferred` task is even created. Competing writers additionally go through `getRawData` + a per-entity mutex before their `setItem`, so their native/IDB dispatch happens strictly later in real time.
2. iOS/Android AsyncStorage routes all module methods through one serial native queue — FIFO by arrival. Web/desktop/ext (`$webStorageSimpleDB` IndexedDB): transactions with overlapping scope run in creation order, and a readonly txn created before a readwrite one sees the pre-write state.
3. Residual-risk direction is conservative: if the sample were ever polluted, it could only flip `appStatusRawExists` false→true, which under-claims whole-key loss rather than fabricating it — the core "flag missing at boot" verdict comes from `migratedFlag` independently.

---

originalix (approved):

PR #12287, opened 2026-07-03
fix: forward async storage writes to background(OK-57349)

huhuanming:
[OK-57349](https://onekeyhq.atlassian.net/browse/OK-57349)

----------

## Summary
- Upgrade OneKey app modules to 3.0.78, including the AsyncStorage package with passive write-forwarding hooks, manifest refresh support, web no-op reload support, read-side manifest refresh, and clear-as-multiRemove forwarding.
- Inject an iOS main-runtime AsyncStorage write forwarder that relays writes to the background runtime.
- Add a background RPC method that performs the actual AsyncStorage writes via bg `appStorage`.

## Intent & Context
This fixes an iOS production-only two-JS-runtime storage race. Production native builds run separate main and background JS runtimes in the same native process. Their JS heaps and native module instances are separate, while the underlying iOS AsyncStorage files and manifest are shared native resources. Direct writes from the main runtime can rewrite the shared manifest from a stale runtime-local copy and drop keys written by the background runtime.

## Root Cause
On iOS, AsyncStorage keeps a runtime-local in-memory manifest and rewrites the shared `manifest.json` on writes. In the dual-runtime model, main and bg each have independent JS/native module instances and therefore independent manifest copies. The underlying storage files are shared by both runtimes, so a main-runtime write can overwrite bg-runtime manifest updates with stale state. Android uses a different backend and is not affected by this manifest rewrite path.

## Design Decisions
- Keep the package-level AsyncStorage change passive: without an injected predicate/forwarder, the package falls back to its original path and does not crash on Android, web, or unconfigured runtimes.
- Implement the concrete dependency injection in app-monorepo, not in app-modules, so thread/platform knowledge stays in the app.
- Enable forwarding only for `iOS + main runtime + native background thread`; bg runtime remains the single writer for AsyncStorage writes in this mode.
- Refresh iOS AsyncStorage manifests before reads and around forwarded writes so main reads stay local but can observe bg-origin writes from the shared disk manifest.
- Forward `clear()` as `multiRemove(freshMainKeys)` instead of bg-side `clear()`, avoiding bg enumeration from a stale runtime-local manifest.
- Serialize main-runtime AsyncStorage forwarded writes and retry them through bg after transient `remote-broken` recovery; never fall back to main-local writes because main/bg share disk files but keep separate runtime-local manifest caches.
- Use a shared type-only contract so `@onekeyhq/shared` stays independent and `kit-bg` only imports from allowed lower-level packages.
- Native-vs-JS old-binary compatibility is intentionally out of scope for this PR per release policy; JS and native are shipped together for this change.

## Changes Detail
- `apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts`: installs the global AsyncStorage write predicate and forwarder on the main runtime, routing writes through the existing main-to-bg service transport; comments now spell out the package/app coupling, late-injection boundary, read-side manifest refresh dependency, log-free hot-path predicate, and no-main-local-fallback rule.
- `packages/kit-bg/src/apis/BackgroundApiBase.ts`: adds `writeAsyncStorage` root background method and writes through bg `appStorage`; only error diagnostics are kept on the write hot path.
- `packages/kit-bg/src/apis/BackgroundApiProxyBase.ts` and `IBackgroundApi.ts`: expose the new root bridge method to keep proxy typing complete.
- `packages/shared/src/storage/asyncStorageWriteForwarderTypes.ts`: adds shared type definitions for write methods and the global injection contract.
- `package.json`, `apps/mobile/package.json`, `packages/components/package.json`, `yarn.lock`, `apps/mobile/ios/Podfile.lock`: upgrade app-modules packages to 3.0.78.

## Module Release
- Published app-modules 3.0.78 from `OneKeyHQ/app-modules@codex/fix-async-storage-manifest-refresh`.
- `package-publish` CI run `28689590168` completed successfully.
- Confirmed `@onekeyfe/react-native-async-storage@3.0.78` is visible on npm.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile iOS only for behavioral forwarding; dependency versions are updated for mobile modules. Android and web keep passive/no-op paths for the new manifest reload hook.
- **Risk Areas**: Startup ordering between main and bg transport, AsyncStorage write semantics, and iOS Pod/package version alignment.

## Test plan
- [x] app-modules: `git diff --check`
- [x] app-modules: `yarn --cwd native-modules/react-native-async-storage typecheck`
- [x] app-modules: `yarn prettier --check native-modules/react-native-async-storage/src/NativeAsyncStorage.web.ts native-modules/react-native-async-storage/src/index.tsx native-modules/react-native-async-storage/src/runtimeConfig.ts`
- [x] app-modules: `package-publish` CI for 3.0.78
- [x] app-monorepo: `yarn onekeyhq:modules:upgrade 3.0.78`
- [x] app-monorepo: `yarn lint:staged`
- [x] app-monorepo: `yarn tsc:staged`
- [x] app-monorepo review fix: `yarn prettier --check apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts`
- [x] app-monorepo review fix: `yarn lint:staged`
- [x] app-monorepo review fix: `yarn tsc:staged`
- [x] app-monorepo: `development/scripts/ios-release-build-deploy.sh xcode`
- [x] app-monorepo: `development/scripts/ios-release-build-deploy.sh deploy`
- [x] Confirmed `@react-native-async-storage/async-storage` resolves to `@onekeyfe/react-native-async-storage@3.0.78` in the worktree.
- [x] Confirmed AsyncStorage package remains passive without injection, so unconfigured Android/web paths do not throw.
- [x] iOS release simulator dual-runtime validation: `main` runtime installed the forwarder, bg transport was ready, `AsyncStorageDevSettings` `Save 1M Data` produced `multiSet -> writeAsyncStorage -> bg write-success -> main forward-success`, and `Clear data` produced the same successful path for `multiRemove`.

[OK-57349]: https://onekeyhq.atlassian.net/browse/OK-57349?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +165,48 @@ function createTransportError(message: string) {
...
+          createTransportError(
+            'AsyncStorage write forwarding requires native background thread transport',
+          ),
+        ),
+    );
+  };
+
+  asyncStorageGlobal.__onekeyAsyncStorageShouldForwardWriteGetter =
+    isAsyncStorageWriteForwardingEnabled;
+}
```
sidmorizon, line 641: **AsyncStorage write forwarder globals have no in-repo consumer**

Type: `Flag`

The globals `__onekeyAsyncStorageWriteForwarder` and `__onekeyAsyncStorageShouldForwardWriteGetter` are set in `installAsyncStorageWriteForwarder()` but never read anywhere in this repository. A grep confirms only two files reference them: the type definitions (`asyncStorageWriteForwarderTypes.ts`) and the installer (`setupMainThreadBackgroundRunner.ts:186-207`). The consumer that actually intercepts AsyncStorage writes and calls the forwarder presumably lives inside the `@onekeyfe/react-native-async-storage` package being bumped from 3.0.74 to 3.0.76. This means the feature is only functional when paired with the correct package version — if the package bump were reverted without reverting this code, the forwarder would be installed but never invoked (harmless dead code). Conversely, if the package were bumped without this code, writes would check the getter, find it undefined, and proceed locally (also safe). The coupling is implicit rather than enforced.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +165,48 @@ function createTransportError(message: string) {
...
+        sync: false,
+      },
+      () =>
+        Promise.reject(
+          createTransportError(
+            'AsyncStorage write forwarding requires native background thread transport',
+          ),
+        ),
+    );
+  };
```
sidmorizon, line 629: # 🟠 P1 main 运行时“读己写”不一致：只转发写、读仍走本地

## 问题

`installAsyncStorageWriteForwarder` 把 main 运行时的所有 AsyncStorage 写（`multiSet`/`multiRemove`/`multiMerge`/`clear`）通过 RPC 转发到 bg 执行，但读路径（`getItem`/`multiGet`/`getAllKeys`）仍走 main 本地原生 AsyncStorage。iOS 原生 `RNCAsyncStorage` 在**每个运行时内**维护一份常驻内存 manifest；写发生在 bg 时，main 的这份内存 manifest 不会被更新。

## 影响

仅在 `iOS + main + enableNativeBackgroundThread` 下：main 里 `await AsyncStorage.setItem('k','v')` 后再 `getItem('k')` 会读到旧值或 `null`（新 key 不在 main 的 manifest 里，小值内联在 manifest 中也读不到），直到 App 重启 main 重新加载 manifest 才可见。改动前 main 自己的写会更新自己的内存 manifest，`读己写`是一致的；本 PR 只转发写不转发读，**引入了 main 运行时的读己写不一致**——这正是该 PR 想修复的“stale manifest”问题在读侧的对称面。

## 建议

要么把读路径也统一到 bg（保证读己写一致），要么在转发写成功后让 main 端本地 manifest 失效/重载；若确认 main 侧不再直接读 AsyncStorage，请在此处注释里明确该前提并加断言，防止后续在 main 侧读回时踩坑。

[Claude session ↗](https://claude.ai/code/session_019ecQiAt9FrtwgHTkFFfJCm)

---

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +165,48 @@ function createTransportError(message: string) {
...
+        params: [buildAsyncStorageWriteRequest(method, args)],
+        sync: false,
+      },
+      () =>
+        Promise.reject(
+          createTransportError(
+            'AsyncStorage write forwarding requires native background thread transport',
+          ),
+        ),
+    );
```
sidmorizon: # 🟠 P1 bg 传输不可用时写被 reject 且无本地兜底 → 静默丢写（main）

## 问题

转发器唯一的“兜底”是 `() => Promise.reject(...)`，且它只在传输被 `!isNativeBackgroundThreadTransportEnabled()`（禁用）时才会被调用。而在 `iOS + main` 上传输是启用的，一旦 `transportState === 'remote-broken'`，`callRemoteRequest` 会直接 `return Promise.reject(getRemoteBrokenReason())`（本文件 `line 804`），根本走不到这个 fallback。也就是说 **remote-broken 期间被转发的写没有任何本地持久化兜底，直接 reject**。

## 影响

`remote-broken` 会在 bg 冷启动超过 `READY_TIMEOUT_MS`(`10s`)、`SharedRPC` 不可用、响应非法等情况下触发——慢设备冷启动 bg >10s 是现实场景。在该窗口内 main 的所有 AsyncStorage 写都会 reject：大量调用方是 fire-and-forget（`void AsyncStorage.setItem(...)`），reject 后**数据静默丢失**；`await` 的调用方则拿到抛错。改动前这些写在 main 本地必然落盘，本 PR 在 bg 降级期把它变成了丢写。

## 建议

不要静默丢：在 remote-broken/未 ready 时将写**排队并在 bg 恢复后重放**（transport 本身支持从 remote-broken 恢复到 ready），或把失败明确上抛让调用方决策。注意**不要**简单回退成 main 本地写——那会重新引入本 PR 要修复的 manifest 竞争。

[Claude session ↗](https://claude.ai/code/session_019ecQiAt9FrtwgHTkFFfJCm)

---

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +229,112 @@ function createTransportError(message: string) {
...
+    enabled,
+    isNativeIOS: Boolean(platformEnv.isNativeIOS),
+    isNativeMainThread: Boolean(platformEnv.isNativeMainThread),
+    isNativeBackgroundThread: Boolean(platformEnv.isNativeBackgroundThread),
+    enableNativeBackgroundThread: Boolean(
+      platformEnv.enableNativeBackgroundThread,
+    ),
+  });
+  return enabled;
+}
```
sidmorizon, line 250: **Diagnostic logging on every shouldForward check may be excessively noisy in production**

Type: `Flag`

`isAsyncStorageWriteForwardingEnabled` (`setupMainThreadBackgroundRunner.ts:232-248`) is assigned as the `__onekeyAsyncStorageShouldForwardWriteGetter` global. The consumer (the patched AsyncStorage library in app-modules) will call this getter on every write operation to decide whether to forward. Each invocation logs a full diagnostic payload via `NativeLogger.write` including platform flags. On a busy app with frequent AsyncStorage writes (jotai persistence, settings, etc.), this could produce thousands of log lines per session, bloating `app-latest.log` and adding I/O overhead on the main thread. Consider gating the log behind a `__DEV__` check or logging only once.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +229,112 @@ function createTransportError(message: string) {
...
+          params: [buildAsyncStorageWriteRequest(method, args)],
+          sync: false,
+        },
+        () =>
+          Promise.reject(
+            createTransportError(
+              'AsyncStorage write forwarding requires native background thread transport',
+            ),
+          ),
+      );
```
sidmorizon: **Forwarder's local fallback always rejects — no graceful degradation if transport breaks**

Type: `Flag`

The forwarder's `localFallback` at `setupMainThreadBackgroundRunner.ts:299-304` unconditionally returns `Promise.reject(...)`. In `callRemoteRequest`, if the transport state is `remote-broken`, the call is rejected immediately. But if the transport transitions to broken *after* the call is queued (e.g., background runtime crashes mid-flight), the queued call's `localFallback` will also reject. This means AsyncStorage writes forwarded through this path have no fallback to local execution — a transport failure will surface as a hard error to the caller. This is likely intentional (the comment says forwarding "requires" the transport), but it differs from other `callServiceRequest` callers in `BackgroundApiProxyBase` which provide real local fallbacks. Worth confirming this is the desired behavior for storage writes, since a failed write could leave persisted state inconsistent.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +229,112 @@ function createTransportError(message: string) {
...
+  ) => {
+    const startedAt = Date.now();
+    const summary = getAsyncStorageWriteArgSummary(method, args);
+    asyncStorageForwarderLog('forward-request', {
+      ...summary,
+      transportState,
+      queuedCallCount: queuedCalls.length,
+      pendingRemoteCallCount: pendingRemoteCalls.size,
+      stack: getAsyncStorageForwarderStack(),
+    });
```
sidmorizon: **Heavy diagnostic logging throughout the forwarding path may impact production performance**

Type: `Flag`

Beyond the per-call getter logging (reported as a bug), the forwarding path includes substantial logging at multiple stages: `forward-request` (line 284-290 with stack trace capture via `new Error().stack`), `forward-success` (line 306-312), `forward-error` (line 314-321), `write-start` (BackgroundApiBase.ts:383-391), `write-success` (line 416-418), and `write-error` (line 421-424). The stack trace capture at `apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts:123` (`new Error().stack`) is particularly expensive as it forces the engine to capture and format a stack trace on every forwarded write. While each individual log is wrapped in try/catch for safety, the cumulative cost on a hot path like AsyncStorage writes could be significant in production. This level of instrumentation is appropriate for initial rollout diagnostics but should be reduced once the feature is stable.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +229,112 @@ function createTransportError(message: string) {
...
+    args,
+  ) => {
+    const startedAt = Date.now();
+    const summary = getAsyncStorageWriteArgSummary(method, args);
+    asyncStorageForwarderLog('forward-request', {
+      ...summary,
+      transportState,
+      queuedCallCount: queuedCalls.length,
+      pendingRemoteCallCount: pendingRemoteCalls.size,
+      stack: getAsyncStorageForwarderStack(),
```
originalix: 当 `main` runtime 的 AsyncStorage 写入都走这条 forwarding 路径时，这里会先生成 `Error().stack`，随后 `main`/`bg` 两边还会继续写多条 `NativeLogger`。

这样一次正常写就被放大成额外的 JS 计算和文件日志 I/O，在启动期或 jotai 持久化这类高频场景里，容易把修复存储一致性的路径变成新的卡顿/日志噪声来源。

建议把栈采集和 success 日志降到 debug 或采样开关下，默认只保留 error 日志，至少不要在每次成功写入时抓栈。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +221,168 @@ function createTransportError(message: string) {
...
+      await waitForAsyncStorageForwarderTransportRecovery(summary);
+    }
+  }
+}
+
+function enqueueAsyncStorageWriteToBackground(
+  request: IBackgroundThreadServiceCallRequest,
+  summary: ReturnType<typeof getAsyncStorageWriteArgSummary>,
+) {
+  const forwarded = asyncStorageWriteForwardChain.then(
```
originalix, line 550: 这里把所有 forwarded AsyncStorage 写都串到 `asyncStorageWriteForwardChain` 上了，但下层 `dispatchRemoteRequest()` 对 `service-call` 仍沿用 10 分钟通用超时，而且超时后只 reject 当前 call，不会把 transport 切到 `remote-broken`。

这样一来，只要某一次 `writeAsyncStorage` 在 `background` runtime 卡住或者 response 丢了，后面的所有 `main` runtime 写都会一直堵在这条链上，直到 10 分钟超时；同时 `forwardAsyncStorageWriteToBackground()` 只在 `transportState === 'remote-broken'` 时才重试，这个超时写本身也不会被 replay。

结果就是一次瞬时的 bg transport 抖动会被放大成整条 AsyncStorage 写通道冻结，甚至直接丢掉这次本该落盘的状态。这里需要给 `writeAsyncStorage` 单独的 timeout / recovery 策略，例如把超时视为 transport failure 触发 recovery 后重放，或者至少不要让单个卡死写长期占住整条串行链。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。这里的 consumer 确实不在 app-monorepo 里，而是在 app-modules 发布的 `@onekeyfe/react-native-async-storage@3.0.78`。

当前 app 侧注释和 PR 描述已经明确这个耦合点：app-monorepo 只负责在 iOS `main` runtime 注入 predicate/forwarder；app-modules 的 AsyncStorage wrapper 会 lazy 读取这些 globals。没有 consumer 时 installer 是 inert；没有 installer 时 package 保持原本 local path。这个是 version-locked 的 JS/package contract。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。没有把 read 路径转到 bg，避免 main read 依赖 bg 响应；修复点放在 app-modules `@onekeyfe/react-native-async-storage@3.0.78`：iOS `getItem` / `multiGet` / `getAllKeys` 在 read 前 reload main runtime 的 native manifest。

运行时边界是：`main` 和 `bg` 的 JS heap / native module manifest cache 是各自独立的，但 iOS AsyncStorage 的磁盘文件是共享资源。写入仍由 bg 执行；main 读之前刷新自己的 manifest cache，因此可以从共享磁盘 manifest 观察 bg-origin 写入，保留 main-local read，同时修复 read-after-forwarded-write 的一致性问题。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。现在 main runtime 的 AsyncStorage forwarded writes 会通过 `asyncStorageWriteForwardChain` 串行化；如果 transport 进入 `remote-broken`，不会直接把写丢给调用方，而是等待 bg transport 恢复后重试，直到 recovery timeout。

仍然没有 main-local fallback，因为 iOS 上 main/bg 共享 AsyncStorage 磁盘文件，但各自持有 runtime-local manifest cache；回退到 main 写会重新引入 stale manifest 覆盖风险。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。`isAsyncStorageWriteForwardingEnabled` 现在保持纯 predicate，不写 NativeLogger，也不做任何副作用；代码里补了 hot-path 注释，明确 app-modules 每次 AsyncStorage write 都会调用这个 getter，必须保持 side-effect free / log-free。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。`localFallback` 对 disabled / non-recoverable path 仍然 reject 是有意的，因为不能回退到 main-local write；但 `remote-broken` 这类可恢复状态现在不会直接结束写入，而是由 forwarder 捕获后等待 transport recovery 并通过 bg 重试。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。AsyncStorage forwarding hot path 已降噪：默认不再记录每次成功写入，不再抓 `Error().stack`，bg 侧也只保留 error diagnostics。main 侧保留 install-start/install-done 和最终 forward-error，用于验证注入与失败排查。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。这条路径现在默认不抓 stack，也不在每次成功写入时写 NativeLogger；只保留安装日志和错误日志。正常 `main -> bg` forwarded write 不再被放大成多条成功日志。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。这里现在不再让 `writeAsyncStorage` 继承 10 分钟通用 service-call timeout。

具体改动：
- `main` runtime 的 forwarded AsyncStorage writes 仍通过 `asyncStorageWriteForwardChain` 串行进入 `bg`，保持写入顺序；
- `writeAsyncStorage` 增加专用 15s request timeout，避免单笔丢响应长期占住整条写链；
- 如果这类 request timeout 发生，forwarder 会在 60s 总 recovery 窗口内重试同一笔写；
- 如果 transport 已进入 `remote-broken`，仍等待 bg transport 恢复后重放；
- 仍然不会 fallback 到 main-local write，因为 iOS 上 main/bg 共享 AsyncStorage 磁盘文件，但 native module manifest cache 是 per-runtime 的，main 本地写会重新引入 stale manifest 覆盖风险。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +222,209 @@ function createTransportError(message: string) {
...
+) {
+  const deadline = Date.now() + ASYNC_STORAGE_FORWARDER_RECOVERY_TIMEOUT_MS;
+  let lastError: unknown;
+  while (Date.now() < deadline) {
+    try {
+      await callServiceRequest(request, () =>
+        Promise.reject(createAsyncStorageForwarderTransportRequiredError()),
+      );
+      return;
+    } catch (error) {
```
originalix, line 503: 当 `writeAsyncStorage` 超过 15 秒或者 `main` runtime 丢了 response 时，这里会直接把同一条 request 重发；但下层 timeout 只是把 `main` 侧 pending promise reject 掉，并不会取消 `background` runtime 里已经在执行的原始写请求。

这样 `clear` / `multiRemove` 这类破坏性写如果原始请求晚到成功，就可能和重试请求一起在共享 AsyncStorage 上执行两次，或者在更新后的新写之后才落盘，把 fresh state 又删掉。`main` 侧还会把这类 late response 当成 `no pending call` 丢掉，所以它没有机会阻止这次重复落盘。

这里不要把 timeout 直接当成可安全重放的失败。更稳妥的修复方向是给 `writeAsyncStorage` 带上单调 requestId/sequence，并让 `background` 侧对重复或过期请求去重后再允许重试。

huhuanming: 已处理，这里不再把 15s timeout 后的重试当成“安全的新写入”。

现在每条 `main` runtime forwarded AsyncStorage write 都会带稳定 `requestId`；同一条写在 transport timeout / response lost 后重试时会复用同一个 `requestId`。`background` runtime 的 `writeAsyncStorage` 入口按 `requestId` 做去重：
- 如果同一 request 正在执行，retry 只等待同一个 in-flight promise，不会再次执行 `clear` / `multiRemove` / `multiSet` / `multiMerge`；
- 如果同一 request 已经成功完成但 response 在 `main` 侧丢了，retry 直接返回，不会重复落盘；
- 只有原始 request 失败时，后续 retry 才允许再次执行。

这样保留了 `main` 侧串行写入顺序，同时避免 destructive write 因 timeout retry 在共享 iOS AsyncStorage 文件上重复执行。`main` / `bg` 的 JS heap 和 native module manifest cache 仍然是 per-runtime，底层 iOS AsyncStorage 文件是共享资源，所以仍不做 main-local fallback。

packages/kit-bg/src/apis/BackgroundApiBase.ts
```diff
@@ -120,6 +180,56 @@ class BackgroundApiBase implements IBackgroundApiBridge {
 
   private isFlushingPendingInjectedBridgeMessages = false;
 
+  private asyncStorageWriteRequestsInFlight = new Map<string, Promise<void>>();
```
originalix, line 196: `main` runtime 在 timeout 或 transport 恢复后会拿同一个 `requestId` 重放 `writeAsyncStorage`。

但这里的去重状态只保存在当前 `BackgroundApiBase` 实例的 `asyncStorageWriteRequestsInFlight` / `completedAsyncStorageWriteRequestIdSet` 里，它们属于 `background` JS heap。`main` 和 `background` 共享的是原生 AsyncStorage 文件，不共享这份内存；如果旧的 `background` 已经把写入落盘、却在 response 回给 `main` 之前 crash/restart，新 `background` 会把这个旧请求当成首次请求再执行一次。

这样恢复期里旧的 `clear`、`multiRemove`、`multiSet` 或 `multiMerge` 可能晚到覆盖 restart 后更晚的 bg 写入，把 shared AsyncStorage 回滚到旧状态。这里需要一个跨 `background` 重启仍可见的 durable-commit fence/ack，或者把 restart 后的 replay 改成只有在 bg 能证明旧请求未提交时才允许重放。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +226,214 @@ function createTransportError(message: string) {
...
+  request: IBackgroundThreadServiceCallRequest,
+  summary: ReturnType<typeof getAsyncStorageWriteArgSummary>,
+) {
+  const forwarded = asyncStorageWriteForwardChain.then(
+    () => forwardAsyncStorageWriteToBackground(request, summary),
+    () => forwardAsyncStorageWriteToBackground(request, summary),
+  );
+  asyncStorageWriteForwardChain = forwarded.catch(() => undefined);
+  return forwarded;
+}
```
sidmorizon, line 556: **Serialized write chain blocks all subsequent writes during recovery**

Type: `Flag`

The `enqueueAsyncStorageWriteToBackground` function (`setupMainThreadBackgroundRunner.ts:348-358`) chains all writes sequentially via `asyncStorageWriteForwardChain`. Each write in `forwardAsyncStorageWriteToBackground` can retry for up to 60 seconds (`ASYNC_STORAGE_FORWARDER_RECOVERY_TIMEOUT_MS`). During this window, ALL subsequent AsyncStorage writes from the main thread are queued behind the stuck write. This is intentional (the comments at lines 406-409 explain ordering must be preserved), but it means a single transport hiccup can stall the entire main-thread AsyncStorage write pipeline for up to a minute. If the app performs frequent writes (e.g. during onboarding or bulk token refresh), this could manifest as UI freezes or timeouts in unrelated features that depend on storage writes completing.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +226,214 @@ function createTransportError(message: string) {
...
+    const startedAt = Date.now();
+    const summary = getAsyncStorageWriteArgSummary(method, args);
+    try {
+      await enqueueAsyncStorageWriteToBackground(
+        {
+          type: 'service-call',
+          method: 'writeAsyncStorage',
+          params: [buildAsyncStorageWriteRequest(method, args)],
+          sync: false,
+        },
```
sidmorizon, line 601: **Forwarder bypasses proxy method-name prefix convention**

Type: `Flag`

The `installAsyncStorageWriteForwarder` at line 392-398 constructs a service-call request with `method: 'writeAsyncStorage'` and sends it directly via `callServiceRequest`. However, the normal proxy path (`BackgroundApiProxyBase.writeAsyncStorage` → `callBackground`) prepends `INTERNAL_METHOD_PREFIX` to the method name before sending. This means the forwarder sends a different method name than the proxy would. The bg-side executor in `setupBackgroundThreadRPCHandler.ts:460` must handle both prefixed and unprefixed method names for this to work. If the bg-side only recognizes prefixed names, the write would fail silently or throw on the background thread. This needs verification against the bg-side request executor logic.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -381,7 +734,18 @@ function handleRuntimeSignal() {
...
+    setBackgroundThreadReadyPayload(runtimePayload);
+  }
+
   if (transportState === 'ready') {
+    if (previousBootId && previousBootId !== runtimePayload.bootId) {
+      transportLog(
+        `background runtime bootId changed while transport ready: ${previousBootId} -> ${runtimePayload.bootId}`,
+      );
+    }
     return;
```
sidmorizon, line 900: **Background runtime restart while transport is ready leaves old pending calls dangling**

Type: `Flag`

When `handleRuntimeSignal` detects a bootId change while `transportState === 'ready'` (lines 743-749), it logs the change but returns early without rejecting the pending remote calls from the old bg runtime. Those calls remain in `pendingRemoteCalls` and will only be cleaned up when their individual timeouts fire (up to 10 minutes for service calls, 15 seconds for AsyncStorage writes). During this window, the call slots are occupied but will never receive responses from the now-dead bg runtime. This is unlikely to cause practical issues given the 8192 slot limit, but it means callers of the old bg runtime will experience the full timeout delay rather than an immediate rejection. The AsyncStorage forwarder mitigates this for its own writes via `assertAsyncStorageForwarderReplayStillSafe`, which detects the bootId mismatch and fails fast.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -381,7 +734,18 @@ function handleRuntimeSignal() {
...
   }
 
+  const previousBootId = currentBackgroundRuntimeBootId;
+  currentBackgroundRuntimeBootId = runtimePayload.bootId;
+  if (previousBootId !== runtimePayload.bootId || transportState !== 'ready') {
+    setBackgroundThreadReadyPayload(runtimePayload);
+  }
```
sidmorizon: **setBackgroundThreadReadyPayload call moved earlier, now fires on bootId change even when transport is already ready**

Type: `Flag`

Previously, `setBackgroundThreadReadyPayload` was only called during the `idle/starting/remote-broken → ready` transition. Now (lines 737-741) it is also called when the transport is already `ready` but the bootId changes. This means `runtimeState.ts` listeners registered via `onBackgroundThreadReady` will be re-notified with the new payload. Any listener that assumes it will only be called once (or only during initial ready) could be affected. The `runtimeState.ts:25-27` implementation simply iterates all listeners with the new payload, so idempotent listeners are fine, but stateful listeners that accumulate or toggle on each call should be audited.

packages/kit-bg/src/apis/BackgroundApiBase.ts
huhuanming: 已处理。这里没有再尝试做跨 background 重启的 durable commit fence，而是把 restart 后 replay 直接禁止掉：main runtime 在发送 forwarded AsyncStorage write 前记录当前 bg bootId；timeout / transport recovery 后只有 bootId 仍相同才会复用同一个 requestId retry。若 SharedStore ready payload 显示 bg bootId 已变化，main 会抛出 unsafe replay error，不再把旧 clear / multiRemove / multiSet / multiMerge 发给新的 bg runtime。

运行时边界：request dedupe 仍只在 background JS heap 内用于同一 bg boot 的 in-flight/completed 去重；main / bg JS heap 不共享。共享的是 iOS AsyncStorage native 文件，以及用于观测 bg bootId 的 SharedStore ready payload。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。serialized write chain 保留，因为 main -> bg 的 AsyncStorage writes 需要保持顺序，避免共享 iOS AsyncStorage 文件上的后写被旧写覆盖；但 recovery window 已从 60s 收紧为两个 15s request timeout，也就是最多允许一次 same-boot retry，不再让后续 writes 被旧的 60s 窗口阻塞。

如果 bg runtime restart，main 现在会通过 bootId guard 禁止 replay；如果 restart 发生在 transport 仍 ready 的窗口，后续提交也会立即 reject old-bg pending calls，避免等待通用 10 分钟 service timeout。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已验证并补充注释。这里传未加前缀的 `writeAsyncStorage` 是符合 native bg executor 路径的：main 发 `service-call` 后，apps/mobile/background.ts 会调用 `backgroundApiProxy.callBackgroundMethod(request.sync, request.method, ...request.params)`；`BackgroundApiProxyBase.callBackgroundMethod()` 会在 bg runtime 内应用 `INTERNAL_` decorator prefix，然后调用真实的 `@backgroundMethod`。

所以 forwarder 没有绕过 prefix 约定；它发送的是 proxy method name，prefix 由 bg 侧 proxy 层统一处理。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。现在 `handleRuntimeSignal()` 在 transport 已 ready 且发现 bg bootId 变化时，会立即 reject queued calls 和 pendingRemoteCalls，错误原因包含 old/new bootId，不再让旧 bg runtime 的 pending calls 等到各自 timeout。

运行时边界：pendingRemoteCalls 位于 main JS heap；bg restart 后旧 bg JS heap 已消失，旧 response 不会再可靠回来。这里使用 SharedStore ready payload 的 bootId 作为跨 runtime restart 信号，只清理 main 侧 pending 状态，不依赖共享 JS 内存。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。`setBackgroundThreadReadyPayload()` 已移回只在 `idle/starting/remote-broken -> ready` 转换时调用。ready 状态下看到新的 bg bootId 时，只更新 transport 内部的 currentBackgroundRuntimeBootId 并 reject 旧 pending calls，不再重新通知 `onBackgroundThreadReady` listeners。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +230,277 @@ function createTransportError(message: string) {
...
+      'AsyncStorage write forwarding cannot safely replay after background runtime restart',
+      `requestBootId=${requestBootId}`,
+      `currentBootId=${currentBootId}`,
+      `request=${stringifyAsyncStorageForwarderLogValue(summary)}`,
+      `lastError=${(lastError as Error)?.message || 'unknown'}`,
+    ].join(', '),
+  );
+}
+
+function assertAsyncStorageForwarderReplayStillSafe({
```
originalix: 当这次 main-origin AsyncStorage 写已经发给旧 boot 的 `bg`、但 response 还没回来时，只要 `bg` 在这个窗口里 restart，这里就会因为 `requestBootId !== currentBootId` 直接抛 `UnsafeReplayError`。

这确实挡住了跨 boot 的 blind replay，但也把两种情况合并成同一个 hard fail：一是旧 `bg` 已经落盘、只是 response 丢了；二是旧 `bg` 其实还没执行这次写。对第二种情况来说，这次写会被永久丢掉，因为后面没有 durable ack、提交 fence 或补偿重试去证明“这次请求尚未提交，可以安全重放”。

结果就是 restart 窗口里的 main-origin `setItem` / `multiSet` / `multiRemove` 等写入可能直接失效，调用方只能收到 rejection，恢复后会继续看到旧值。这里需要补一个跨 `bg` 重启仍可见的 commit/ack 机制，或者至少把 restart 后的处理拆成“已提交禁止重放 / 未提交允许安全重放”，而不是一律 reject。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -381,7 +737,21 @@ function handleRuntimeSignal() {
...
   if (transportState === 'ready') {
+    if (bootIdChanged) {
+      const reason = `Background runtime restarted while transport ready: ${previousBootId} -> ${runtimePayload.bootId}`;
+      transportLog(
+        `background runtime bootId changed while transport ready: ${previousBootId} -> ${runtimePayload.bootId}`,
+      );
+      rejectQueuedCalls(reason);
+      rejectPendingRemoteCalls(reason);
+    }
     return;
```
sidmorizon, line 900: **Background runtime restart while transport ready does not update stored ready payload**

Type: `Flag`

When `handleRuntimeSignal` detects `transportState === 'ready'` and `bootIdChanged` is true (lines 746-755), it rejects pending calls and returns early — but does NOT call `setBackgroundThreadReadyPayload(runtimePayload)` with the new runtime's payload. The `currentBackgroundRuntimeBootId` IS updated (line 744), so the forwarder's replay-safety check works correctly. However, any downstream code that reads the stored ready payload (set at `setupMainThreadBackgroundRunner.ts:772` only on the initial ready transition) will see stale data from the previous boot. Whether this matters depends on what `setBackgroundThreadReadyPayload` consumers actually read — if they only use it for one-time init, this is fine; if they re-read it after a bg restart, they'll get stale info.

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。现在不再只靠 bootId mismatch 一律 hard fail，而是在 SharedStore 里为每个 main-origin AsyncStorage write 建一个跨 bg boot 可见的 request status fence：

- main 发送前写 `pending`；
- bg 真正开始执行前把同一个 requestId 标记为 `executing`；
- bg 写入成功后标记为 `committed`；
- main 如果在 timeout / restart recovery 后发现 bootId 变化，会读取这个 status：`committed` 直接视为成功，`pending` 说明旧 bg 尚未开始执行，可以安全重放到新 bg，`executing` 或 status 缺失才继续拒绝 blind replay。

运行时边界：main/bg JS heap 仍不共享；这个 fence 存在 SharedStore 这个 shared native resource 里。底层 iOS AsyncStorage 文件仍是共享 native 文件，main/bg 各自的 manifest cache 仍是 per-runtime。这样避免跨 boot blind replay，同时覆盖“旧 bg 尚未执行”的安全重放场景。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
huhuanming: 已处理。`runtimeState` 现在拆出了 `updateBackgroundThreadReadyPayload()`：ready 状态下检测到 bg bootId 变化时，会 silent update stored ready payload，保证后续 `getBackgroundThreadReadyPayload()` 读到新 boot；但不会调用 `setBackgroundThreadReadyPayload()`，因此不会重新通知 `onBackgroundThreadReady` listeners。

同一分支里仍会立即 reject old-bg 的 queued/pending calls，避免旧 bg runtime 的 pending calls 等到各自 timeout。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -381,7 +873,22 @@ function handleRuntimeSignal() {
...
+    if (bootIdChanged) {
+      const reason = `Background runtime restarted while transport ready: ${previousBootId} -> ${runtimePayload.bootId}`;
+      transportLog(
+        `background runtime bootId changed while transport ready: ${previousBootId} -> ${runtimePayload.bootId}`,
+      );
+      updateBackgroundThreadReadyPayload(runtimePayload);
+      rejectQueuedCalls(reason);
+      rejectPendingRemoteCalls(reason);
+    }
     return;
```
sidmorizon, line 900: **Boot-ID-changed path stays in 'ready' state while rejecting all in-flight calls**

Type: `Flag`

When `handleRuntimeSignal` detects a boot ID change while `transportState === 'ready'` (`setupMainThreadBackgroundRunner.ts:882-892`), it rejects all queued and pending remote calls but does NOT transition the transport state away from `'ready'`. This means subsequent calls immediately dispatch to the (new) background runtime without queueing. This is correct because the new bg runtime already signaled readiness (that's how we got the new bootId), but it's a subtle behavioral choice: all in-flight calls are hard-rejected while the transport appears healthy. Callers of `callServiceRequest` that don't have their own retry logic will see transient failures during a bg restart even though the transport never enters `'remote-broken'`. The AsyncStorage forwarder handles this via its own retry loop, but other service calls do not.

huhuanming: 已处理。这里保持 `ready` 是有意的：新的 bg runtime 已经通过 ready payload 发出了可用信号，所以后续新调用可以直接发给新 bg。旧 in-flight calls 属于旧 background JS heap，已经不可能可靠收到 response，因此会被立即 reject。

没有把这些 generic service calls 自动 replay，是因为它们不一定幂等，跨 bg boot 重放可能造成重复副作用。AsyncStorage writes 是特例：它们有 request-status fence（pending/executing/committed）和专用 retry loop，所以可以区分 committed / pending 并安全处理。代码里已补充注释明确这个选择。

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +237,406 @@ function createTransportError(message: string) {
...
 
+function isAsyncStorageWriteForwardingEnabled() {
+  // Hot path: app-modules calls this getter on every AsyncStorage write.
+  // Keep it side-effect free and log-free.
+  return Boolean(
+    platformEnv.isNativeIOS &&
+    platformEnv.isNativeMainThread &&
+    platformEnv.enableNativeBackgroundThread,
+  );
+}
```
sidmorizon, line 250: **Write forwarding is iOS-only; Android dual-thread mode would skip forwarding**

Type: `Flag`

The `isAsyncStorageWriteForwardingEnabled()` function at line 240-248 gates on `platformEnv.isNativeIOS`. If Android ever enables the dual-thread background runtime (`enableNativeBackgroundThread`), AsyncStorage writes from the main thread would NOT be forwarded and would execute locally, potentially causing the manifest corruption issue this feature is designed to prevent. The comment in the forwarder's install log (line 567) logs `isNativeIOS` but doesn't explain why Android is excluded. This is likely because Android's AsyncStorage implementation doesn't share a manifest file between runtimes, but it's worth documenting explicitly.

huhuanming: 已处理。这里保持 iOS-only gate 不变，并在 predicate 旁补充了原因：iOS AsyncStorage 是共享磁盘文件 + per-runtime native manifest cache，main/bg split runtime 才会触发 stale manifest 风险；Android 当前不使用同一个 shared manifest-file cache 模型，打开 forwarding 只会增加 RPC/retry surface。

同时 install-start 诊断里补了 `isNativeAndroid`，后续如果 Android 引入同类 split-runtime native storage-cache hazard，可以明确看到平台状态并重新评估这个 gate。运行时边界仍是：main/bg JS heap 不共享；iOS 底层 AsyncStorage 文件是 shared native resource，manifest cache 是 per-runtime。

packages/kit-bg/src/apis/BackgroundApiBase.ts
```diff
@@ -105,6 +118,71 @@ function summarizeSetAtomValuePayload(value: unknown) {
...
+
+type IAsyncStorageForwarderSharedStore = {
+  set: (key: string, value: string | number | boolean) => void;
+  get: (key: string) => string | number | boolean | undefined;
+};
+
+function getAsyncStorageForwarderSharedStore() {
+  return (globalThis as { sharedStore?: IAsyncStorageForwarderSharedStore })
+    .sharedStore;
+}
```
sidmorizon, line 179: **Bg-side SharedStore access uses different API than main-thread**

Type: `Flag`

The main thread accesses SharedStore via `getSharedStore()` from `@onekeyfe/react-native-background-thread` (`setupMainThreadBackgroundRunner.ts:4`), while the bg-side `BackgroundApiBase.ts:181-184` accesses it via `globalThis.sharedStore`. These are two different JS APIs to the same underlying native SharedStore. This works because the react-native-background-thread package exposes the store differently in each runtime: as a module export in main, and as a global in bg. However, if the bg-side global name ever changes (e.g., in a future version of the background-thread package), the forwarder status reads/writes on the bg side would silently return undefined, causing the status fence to be ineffective (writes would still execute but without the 'committed' acknowledgment).

huhuanming: 已处理：bg 侧不再自己声明/读取 `globalThis.sharedStore`，而是通过 `@onekeyhq/shared/src/modules3rdParty/react-native-background-thread/sharedStore` 的 wrapper 获取。native 实现内部调用 `@onekeyfe/react-native-background-thread` 的 `getSharedStore()`，因此 main/bg 侧都走同一个 package accessor；非 native 实现返回 `undefined`，避免 `kit-bg` 直接静态依赖 mobile native package。运行时边界保持不变：main/bg 是分离 JS heap，SharedStore 是同一 native shared resource，用于同步 request-status fence。

apps/mobile/src/backgroundThread/runtimeState.ts
```diff
@@ -15,18 +15,30 @@ function getBackgroundThreadStateGlobal() {
...
   listeners.forEach((listener) => {
     listener(payload);
   });
 }
 
+export function updateBackgroundThreadReadyPayload(
+  payload: IBackgroundThreadReadyPayload,
+) {
+  saveBackgroundThreadReadyPayload(payload);
+}
```
sidmorizon: **updateBackgroundThreadReadyPayload intentionally skips listener notification on bg restart**

Type: `Flag`

The new `updateBackgroundThreadReadyPayload` function at `apps/mobile/src/backgroundThread/runtimeState.ts:36-40` saves the payload without notifying listeners, unlike `setBackgroundThreadReadyPayload` which iterates and calls all registered listeners. This is called from `handleRuntimeSignal` when a boot-ID change is detected while transport is already 'ready'. The intent is to update the cached payload (so `getBackgroundThreadReadyPayload()` returns fresh data) without re-triggering ready listeners that may perform one-time initialization. Any code that registered via `onBackgroundThreadReady` and depends on being re-notified when the bg runtime restarts will NOT be called. Verify that no listener needs re-initialization after a bg restart.

huhuanming: 已处理：删除了 silent 的 `updateBackgroundThreadReadyPayload()`，bg runtime bootId 变化时现在直接调用 `setBackgroundThreadReadyPayload(runtimePayload)`，因此会更新 main runtime 的 ready cache，也会通知所有现有 `onBackgroundThreadReady` listener。当前代码里没有实际 listener 调用点，但这样可以保证未来 listener 在 bg 重启后不会漏掉 re-init。运行时边界不变：这是 main JS runtime 内的 listener/cache；bg 重启代表新的 bg JS heap，SharedStore/SharedRPC 仍是跨 main/bg 共享的 native resource。

originalix (approved):

apps/mobile/package.json
```diff
@@ -54,37 +54,37 @@
...
+    "@onekeyfe/react-native-splash-screen": "3.0.78",
+    "@onekeyfe/react-native-split-bundle-loader": "3.0.78",
+    "@onekeyfe/react-native-tab-view": "3.0.78",
     "@onekeyfe/react-native-text-input": "0.3.0",
     "@onekeyhq/components": "*",
     "@onekeyhq/kit": "*",
     "@onekeyhq/shared": "*",
     "@phantom/react-native-juicebox-sdk": "0.3.17",
-    "@react-native-async-storage/async-storage": "npm:@onekeyfe/react-native-async-storage@3.0.74",
+    "@react-native-async-storage/async-storage": "npm:@onekeyfe/react-native-async-storage@3.0.78",
```
sidmorizon, line 87: # 🟡 P2 iOS 每次读操作全量重载并解析 manifest.json,双 runtime 共享串行队列放大开销

## 问题

3.0.78 的 fork 包(本行引入)在 iOS 上把 `getItem` / `multiGet` / `getAllKeys` 全部包进 `runReadWithFreshManifest`,**每次读之前**调 `NativeModule.reloadManifest()`;未注入 forwarder 的本地写路径(含 bg runtime 的全部写)也先 reload。而 native 侧 `reloadManifest` 的实现是 `_haveSetup = NO` → `_ensureSetup` **无条件全量 `RCTReadFile` + `RCTJSONParseMutable` 重新解析整个 manifest.json**——没有利用文件 mtime/size 做"未变化则跳过"的 fast path(`RCTManifestModificationDate` 目前只用于目录迁移检查)。

## 影响(仅 iOS,main / bg 两个 runtime 都受影响)

- 每次读从 1 个 native 任务变成 2 个(reload + read),且所有任务落在 main/bg **共享的进程级串行队列**上(`RCTGetMethodQueue` 是 dispatch_once 单例),两个 runtime 互相排队;
- AsyncStorage 的 manifest 内联全部 ≤1KB 的小值(jotai 持久化、WalletConnect session 等都是小值),manifest 体积随 key 数增长,冷启动阶段 jotai hydration + simpleDb 预热的高频读会放大成几百次全量 JSON parse;
- bg runtime 是 simpleDb 的主要读写方,它的每次读写现在都要多付一次全量 manifest parse。

PR 测试计划验证了写入正确性("Save 1M Data"),但没有冷启动 / `tokensStartMs` 类的性能数据。

## 建议

合并前建议跑一次 iOS release 冷启动性能对比;后续可在 app-modules 侧给 `reloadManifest` 加 mtime/size fast-path(文件未变化时跳过 read+parse,`RCTManifestModificationDate` 已有现成实现),或在 JS 侧合并飞行中的 reload(coalescing),把每读一次全量 parse 降为"变化时才 parse"。

---

apps/mobile/src/backgroundThread/setupMainThreadBackgroundRunner.ts
```diff
@@ -159,6 +234,412 @@ function createTransportError(message: string) {
...
+    () => forwardAsyncStorageWriteToBackground(request, summary),
+  );
+  asyncStorageWriteForwardChain = forwarded.catch(() => undefined);
+  return forwarded;
+}
+
+function installAsyncStorageWriteForwarder() {
+  const asyncStorageGlobal = getAsyncStorageWriteForwarderGlobal();
+  // The matching consumer lives in @react-native-async-storage/async-storage
+  // from app-modules >= 3.0.78. This version-locked JS/package contract is
```
sidmorizon, line 561: # 🟡 P2 OTA 热更新的 native-vs-JS 版本偏斜会导致"每次读都崩",需要发布流程硬性拦截

## 问题

这个 version-locked contract 的注释只覆盖了 JS 侧两个方向的降级(无 consumer / 无 installer 都安全),但 **native-vs-JS 偏斜方向没有安全降级**:3.0.78 的 JS 在每次 `getItem` / `multiGet` / `getAllKeys` / 本地写之前都无条件调用 `NativeModule.reloadManifest()`,而 TurboModule spec 里 `reloadManifest` 是 3.0.78 新增方法。如果 3.0.78 的 JS bundle 通过 bundle hot update 下发到还是 3.0.74 native(Pod)的旧安装包上,`reloadManifest` 不存在 → **所有平台(iOS/Android)每次存储读写都抛 TypeError**,app 启动即不可用,且这是 OTA 下发的,用户无法自救,只能等下一个热更或重装。

CLAUDE.md 明确了 "the only real version skew is native-vs-JS" —— 这正好是该 PR 唯一没有代码级防护的偏斜方向。PR 描述里声明 "JS and native are shipped together per release policy",但这依赖流程(bundle release 的 diff-check)而不是代码。

## 建议

二选一(或都做):
1. **流程侧(最低要求)**:在 bundle 热更发布清单里把本 PR 标记为"含 native module 接口变更,禁止 OTA 到 < 3.0.78 native 的版本",确保 diff-check 能拦截;
2. **代码侧(更稳)**:在 app-modules 的 JS 侧加一行 feature-detect,`typeof NativeModule.reloadManifest === 'function'` 不满足时降级为 no-op —— 旧 native + 新 JS 就回落到 3.0.74 的原行为(读不刷新 manifest),从"全量崩溃"降级成"回到修复前的旧语义",代价只是一次函数存在性判断。

---

sidmorizon (commented):
# PR #12287 代码审查报告(head: 11a2ad7)

## 评分

| 维度 | 得分 | 说明 |
|------|------|------|
| 🔒 安全性 | 9/10 | 无密钥/敏感信息;依赖全部为 @onekeyfe 内部包版本对齐(3.0.74→3.0.78),无新增外部依赖;`writeAsyncStorage` 仅内部 proxy 面可达,dapp 注入侧不可达 |
| 💎 代码质量 | 7/10 | 并发设计(串行链、requestId 去重、bootId guard、SharedStore status fence)非常严谨;扣分项:iOS 每次读全量重载解析 manifest 的性能回归风险未量化(见 inline) |
| 🏛️ 架构合理性 | 9/10 | type-only contract 保持 shared 独立、注入实现放 app 层、sharedStore wrapper 平台分离,均正确;跨仓库 version-locked 全局变量契约有固有脆弱性但已注释清楚 |
| ✅ 完整性 | 7/10 | 双 runtime 手工验证充分;但 forwarder/fence 协议无单测,无冷启动性能数据,OTA 偏斜防护依赖流程而非代码(见 inline) |
| **总分** | **8.1/10** | **✅ 可合入,建议先完成下述两项确认** |

## 本轮新发现(未与既有 20 条已解决线程重复)

1. 🟡 **iOS 每次读全量 reload+parse manifest,main/bg 共享串行队列放大** — 见 `apps/mobile/package.json:87` inline;建议合并前跑一次 iOS release 冷启动对比。
2. 🟡 **OTA 热更 native-vs-JS 偏斜 = 每次读崩溃,无代码级降级** — 见 `setupMainThreadBackgroundRunner.ts:561` inline;建议 bundle diff-check 硬拦截,或包侧加 `reloadManifest` feature-detect。
3. 🟢(不阻塞)SharedStore status key 微量残留:main 在 30s recovery 超时放弃后 `finally` 删除 status key,若 bg 迟到完成会以 `committed` 重建该 key,之后无人清理。会话期内小字符串级泄漏、频率极低;后续可让 bg 提交 status 前检查 key 是否仍存在,或加 TTL 清扫。

## 已确认的既有结论(复核通过)

- 20 条既有 review 线程全部 resolved,逐条复核修复真实落地:读侧 manifest refresh、15s 专用超时、requestId 去重、bootId + status fence 跨 boot 重放栅栏、ready 态 bootId 变化即时 reject 旧 pending calls、热路径零日志 predicate 均已在 head 代码中验证。
- `INTERNAL_` prefix 路径:确认 bg executor 经 `callBackgroundMethod` 统一加前缀,forwarder 未绕过约定。
- `buildAppStorageFactory` 对 `multiSet/multiRemove/multiMerge` 原样透传,bg `writeAsyncStorage` 四个分支均可达。
- Android/web:native `reloadManifest` 均为 no-op(Kotlin `promise.resolve(null)` / web `Promise.resolve()`),被动路径不受影响。
- 第三方库覆盖:yarn.lock 中 `@react-native-async-storage/async-storage` 仅一条 resolution(npm alias 指向 fork),@reown/@walletconnect 等均为 peerDependencies,统一命中 fork 的转发逻辑。
- socket-security bot 评论:32 项全部为内部包版本对齐更新,无新增/移除依赖,无告警,无需行动。

## 已知接受的残余风险(作者已注释,不重复开线程)

- forwarder 安装前(main entry 极早期)完成的写保留旧本地语义,理论上仍可与 bg 并发写竞争,窗口极小;
- `clear()` 转 `multiRemove(freshMainKeys)` 存在 TOCTOU:枚举后、执行前 bg 新写入的 key 会幸存,双 runtime 下 clear 本身即 best-effort 语义。

---

limichange (approved):

PR #12295, opened 2026-07-05
fix: preserve token list cache during refresh

huhuanming:
## Summary
- Preserve Home token rows during all-network and single-network refreshes instead of clearing the whole list before fresh data returns.
- Add TokenSelector view-layer SWR snapshots so Send token selection can paint cached rows immediately and revalidate in the background.
- Add local trace logging, a repeatable home-token stress script, focused tests, and keep the native performance overlay disabled by default unless manually enabled.

## Intent & Context
This fixes regressions observed on the wallet Home and Send token selector flows. The target behavior is stale-while-revalidate at the view layer: once a usable rendered list exists, refreshes should keep readable cached rows on screen and only update changed rows after fresh background data arrives. The request also required repeatable verification across all-network/single-network switching, cold-start with and without cache, account switching, account add/remove, and crash observation.

## Root Cause
Home and TokenSelector refresh paths treated `loading` as a reason to clear visible token state. On all-network refresh this made the Home list shrink or flash skeletons while network rows settled progressively. On Send token selection, the selector did not reuse a ready-to-render view snapshot, so re-opening the page or hitting a revalidation error could show skeletons even when the previous page already had usable token data.

## Design Decisions
- Keep SWR at the view layer for TokenSelector instead of adding a normalized all-network data cache in the main runtime.
- Store ready-to-render selector snapshots keyed by owner/filter/network mode via `swrCacheUtils`.
- Keep Home token reactive pipeline rounds resident after authoritative commits so warm refreshes have a stable floor instead of shrinking to only currently settled rows.
- For single-network cached data, ingest a lightweight cache round into the token list VM so owner/network stamping stays correct even for empty cached lists.
- Use native logger local traces for refresh entry points and cache hydration, while avoiding sensitive wallet data in logs.
- Introduce `showPerformanceMonitorV2` defaulting to false so legacy persisted dev values do not auto-open the overlay.

## Changes Detail
- `TokenSelector.tsx`: reads/writes view-layer SWR snapshots for normal, active-account, and filtered token selector modes; cached rows are rendered before background revalidation.
- `computeShowTokenListSkeleton.ts`: requires `displayCount === 0` for refresh skeletons so cached rows are not hidden.
- `TokenListBlock.tsx` and `useTokenListReactivePipeline.ts`: preserve authoritative rounds as the next refresh floor and ingest single-network cached rounds.
- `buildHomeTokenListCacheIngestRound.ts`: builds cache ingest payloads for single-network cached and empty cached states.
- `allNetworkAccountPerf.ts` and `useAllNetwork.ts`: add local trace points for refresh/cache behavior.
- `home-token-refresh-stress.mjs`: adds repeatable simulator automation for refresh/cache/crash observation.
- `ios-release-build-deploy.sh`: improves local log filtering when no fresh session marker exists.
- `WebPerformanceMonitor` / dev settings / `Bootstrap`: performance overlay no longer starts from legacy default-on state.
- `docs/home-token-list-refresh-cache.md`: records scenarios, expected behavior, current behavior, and validation notes.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile native primarily; shared SWR utility and tests are cross-platform TypeScript.
- **Risk Areas**: Home token list refresh ordering, selector cache key correctness, stale owner/network guards, and cache memory footprint. The selected approach avoids main-thread normalized all-network data parsing and keeps background/main JS heaps isolated.

## Test plan
- [x] `yarn test packages/kit/src/components/TokenListView/computeShowTokenListSkeleton.test.ts packages/kit/src/views/Home/components/TokenListBlock/buildHomeTokenListCacheIngestRound.test.ts packages/kit/src/views/Home/components/TokenListBlock/useTokenListReactivePipeline.test.ts --runInBand`
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn tsc:only`
- [x] `development/scripts/ios-release-build-deploy.sh all`
- [x] iOS simulator: Home has cached token rows, tap Send, TokenSelector shows cached token rows within 1s instead of skeleton.
- [x] iOS simulator: close and reopen Send token selector; cached rows render immediately again.
- [x] iOS simulator: revalidation can fail with a network/server error while cached rows remain visible.

### DIFF
--- a/development/scripts/home-token-refresh-stress.mjs
+++ b/development/scripts/home-token-refresh-stress.mjs
@@ -1,0 +1,660 @@
+#!/usr/bin/env node
+import { spawnSync } from 'node:child_process';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+
+const DEFAULT_UDID = '4837E819-A117-4E08-9936-445785D199E3';
+const DEFAULT_BUNDLE_ID = 'so.onekey.wallet';
+
+const COORDS = {
+  accountButton: [95, 136],
+  account1: [195, 196],
+  account2: [195, 258],
+  addAccount: [190, 310],
+  newAccountMore: [361, 315],
+  passwordClose: [358, 687],
+  removeMenu: [105, 790],
+  removeConfirm: [292, 775],
+
+  networkButton: [340, 136],
+  allNetworksTab: [160, 98],
+  singleNetworkTab: [245, 98],
+  allNetworksDone: [196, 812],
+  singleRows: {
+    solana: [115, 336],
+    bitcoin: [115, 476],
+    polygon: [115, 523],
+  },
+
+  pullToRefreshStart: [200, 360],
+  pullToRefreshEnd: [200, 700],
+};
+
+function parseArgs(argv) {
+  const opts = {
+    iterations: 100,
+    udid: DEFAULT_UDID,
+    bundleId: DEFAULT_BUNDLE_ID,
+    session: 'default',
+    outDir: `.tmp/home-token-stress-${new Date()
+      .toISOString()
+      .replace(/[:.]/g, '-')
+      .replace('T', '-')
+      .slice(0, 19)}`,
+    screenshotEvery: 10,
+    coldStartEvery: 5,
+    accountMutationEvery: 1,
+    newAccountColdStart: true,
+    purgeColdStartCacheEvery: 0,
+    stopOnCrash: true,
+    commandTimeoutMs: 90_000,
+    waitScale: 1,
+  };
+
+  for (const arg of argv) {
+    const [key, rawValue] = arg.split('=');
+    const value = rawValue ?? '';
+    switch (key) {
+      case '--iterations':
+        opts.iterations = Number(value);
+        break;
+      case '--udid':
+        opts.udid = value;
+        break;
+      case '--bundle-id':
+        opts.bundleId = value;
+        break;
+      case '--session':
+        opts.session = value;
+        break;
+      case '--out':
+        opts.outDir = value;
+        break;
+      case '--screenshot-every':
+        opts.screenshotEvery = Number(value);
+        break;
+      case '--cold-start-every':
+        opts.coldStartEvery = Number(value);
+        break;
+      case '--account-mutation-every':
+        opts.accountMutationEvery = Number(value);
+        break;
+      case '--purge-cold-start-cache-every':
+        opts.purgeColdStartCacheEvery = Number(value);
+        break;
+      case '--no-new-account-cold-start':
+        opts.newAccountColdStart = false;
+        break;
+      case '--no-stop-on-crash':
+        opts.stopOnCrash = false;
+        break;
+      case '--wait-scale':
+        opts.waitScale = Number(value);
+        break;
+      default:
+        if (key === '--help' || key === '-h') {
+          printHelpAndExit();
+        }
+        throw new Error(`Unknown arg: ${arg}`);
+    }
+  }
+
+  if (!Number.isFinite(opts.iterations) || opts.iterations <= 0) {
+    throw new Error('--iterations must be a positive number');
+  }
+  if (!Number.isFinite(opts.screenshotEvery) || opts.screenshotEvery < 0) {
+    throw new Error('--screenshot-every must be a non-negative number');
+  }
+  return opts;
+}
+
+function printHelpAndExit() {
+  console.log(`Usage:
+  node development/scripts/home-token-refresh-stress.mjs [options]
+
+Options:
+  --iterations=100
+  --udid=${DEFAULT_UDID}
+  --bundle-id=${DEFAULT_BUNDLE_ID}
+  --session=default
+  --out=.tmp/home-token-stress-...
+  --screenshot-every=10          0 disables routine screenshots
+  --cold-start-every=5
+  --account-mutation-every=1       0 disables add/remove account mutation
+  --no-new-account-cold-start      skip cold start on the newly-added empty account
+  --purge-cold-start-cache-every=0 0 disables MMKV cold-start-cache purge
+  --no-stop-on-crash
+  --wait-scale=1
+
+Scenario per loop:
+  1. optional cold start with existing cache
+  2. All Networks pull-to-refresh
+  3. All <-> Solana <-> All <-> Polygon <-> All <-> Bitcoin <-> All
+  4. Account #1 <-> Account #2
+  5. optional add account, cold start on new no-local-cache account, remove it
+
+Evidence:
+  - screenshots under <out>/screenshots
+  - run log JSONL under <out>/events.jsonl
+  - summary under <out>/summary.md
+  - if a crash appears, crash .ips copy + parsed summary + native log tail
+`);
+  process.exit(0);
+}
+
+const opts = parseArgs(process.argv.slice(2));
+const rootDir = process.cwd();
+const outDir = path.resolve(rootDir, opts.outDir);
+const screenshotDir = path.join(outDir, 'screenshots');
+const crashDir = path.join(outDir, 'crashes');
+fs.mkdirSync(screenshotDir, { recursive: true });
+fs.mkdirSync(crashDir, { recursive: true });
+
+const eventLogPath = path.join(outDir, 'events.jsonl');
+const summaryPath = path.join(outDir, 'summary.md');
+const startMs = Date.now();
+const initialCrashFiles = new Set(listCrashReports().map((f) => f.path));
+let lastFailure;
+let crashDetected;
+let isCheckingCrash = false;
+
+function writeEvent(event) {
+  fs.appendFileSync(
+    eventLogPath,
+    `${JSON.stringify({ ts: new Date().toISOString(), ...event })}\n`,
+  );
+}
+
+function sleep(ms) {
+  const scaled = Math.max(0, Math.round(ms * opts.waitScale));
+  if (scaled === 0) return;
+  spawnSync('sleep', [String(scaled / 1000)], { stdio: 'ignore' });
+}
+
+function runCommand(name, args, options = {}) {
+  const startedAt = Date.now();
+  const result = spawnSync(name, args, {
+    cwd: rootDir,
+    encoding: 'utf8',
+    timeout: options.timeoutMs ?? opts.commandTimeoutMs,
+    maxBuffer: 20 * 1024 * 1024,
+  });
+  const elapsedMs = Date.now() - startedAt;
+  const event = {
+    phase: 'command',
+    cmd: [name, ...args].join(' '),
+    status: result.status,
+    signal: result.signal,
+    elapsedMs,
+    stdout: truncate(result.stdout),
+    stderr: truncate(result.stderr),
+  };
+  writeEvent(event);
+  if (result.error || result.status !== 0) {
+    lastFailure = {
+      ...event,
+      error: result.error ? String(result.error) : undefined,
+    };
+    checkCrash(`command-failed:${name}`);
+    if (options.allowFailure) {
+      return result;
+    }
+    throw new Error(
+      `Command failed (${result.status ?? result.signal}): ${name} ${args.join(
+        ' ',
+      )}`,
+    );
+  }
+  return result;
+}
+
+function truncate(text, max = 4000) {
+  if (!text) return '';
+  return text.length > max ? `${text.slice(0, max)}...<truncated>` : text;
+}
+
+function agent(args, options) {
+  return runCommand(
+    'agent-device',
+    [
+      ...args,
+      '--platform',
+      'ios',
+      '--udid',
+      opts.udid,
+      '--session',
+      opts.session,
+      '--json',
+    ],
+    options,
+  );
+}
+
+function click(x, y, label) {
+  writeEvent({ phase: 'tap', label, x, y });
+  agent(['click', String(x), String(y)]);
+}
+
+function clickSelector(selector, label, fallback) {
+  writeEvent({ phase: 'tap-selector', label, selector });
+  const result = agent(['click', selector], { allowFailure: true });
+  if (result.status !== 0 && fallback) {
+    writeEvent({ phase: 'tap-selector-fallback', label, selector, fallback });
+    click(fallback[0], fallback[1], `${label}:fallback`);
+  }
+}
+
+function swipe(x1, y1, x2, y2, durationMs, label) {
+  writeEvent({ phase: 'swipe', label, x1, y1, x2, y2, durationMs });
+  agent([
+    'swipe',
+    String(x1),
+    String(y1),
+    String(x2),
+    String(y2),
+    String(durationMs),
+  ]);
+}
+
+function screenshot(name, force = false) {
+  if (!force && opts.screenshotEvery <= 0) return;
+  const file = path.join(screenshotDir, `${name}.png`);
+  agent(['screenshot', file], { timeoutMs: 120_000 });
+  writeEvent({ phase: 'screenshot', file });
+}
+
+function shouldTakeIterationScreenshot(iteration) {
+  return opts.screenshotEvery > 0 && iteration % opts.screenshotEvery === 0;
+}
+
+function openApp() {
+  agent(['open', opts.bundleId], { timeoutMs: 120_000 });
+  sleep(2500);
+}
+
+function terminateApp() {
+  runCommand('xcrun', ['simctl', 'terminate', opts.udid, opts.bundleId], {
+    allowFailure: true,
+    timeoutMs: 30_000,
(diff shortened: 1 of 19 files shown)


packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -249,18 +319,172 @@ function TokenSelector() {
...
+      showActiveAccountTokenList,
+      tokenSelectorSearchFilterContext,
+    ],
+  );
+
+  const routeTokenSelectorCache = useMemo<ITokenSelectorNormalViewSnapshot>(
+    () =>
+      buildNormalTokenSelectorViewSnapshot({
+        tokenList: {
+          tokens: routeTokens?.data ?? [],
```
originalix: 这里直接把 `routeTokens.data` 当成了 selector 的首屏缓存，但 Receive 从 Home 打开时，这份 `routeTokens` 来自 `useHomeTokenListSnapshot()`，底层是 bg `getRawTokenList()` 的 merged raw list，里面包含 risky tokens。

现在 live self-fetch / floor snapshot 还没落地，或者请求失败回退到 route cache 时，selector 会短暂甚至持续展示本来不该出现在选择器里的 risky token，用户快点一下就能走进错误的接收/发送路径。

这里的 route cache 需要和 floor/live path 保持同一条过滤规则：先用当前 owner 的 risky set 把 `routeTokens.data` 过滤掉，再写入 snapshot；如果拿不到 risky metadata，就不要把 raw route tokens 直接当成 selector 可选列表。

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -249,18 +319,172 @@ function TokenSelector() {
...
+          smallBalanceTokens: [],
+        },
+        tokenListMap: routeTokens?.map ?? {},
+        aggregateTokenListMap: allAggregateTokenMap ?? {},
```
originalix: 这里把 `allAggregateTokenMap` 直接塞进了 `aggregateTokenListMap`，但后面的 `handleTokenOnPress()` 会把 `selectorAggregateTokenListMap` 当成“当前 owner 实际持有的聚合子 token 列表”来做单子 token 直选和 `AggregateTokenSelector` 入参。

`allAggregateTokenMap` 是全局 aggregate metadata，不是当前账户的 owner-scoped membership，所以在 live self-fetch 还没回来，或者失败后继续回退到 route cache 时，聚合币点击会基于错误成员集合走分支：可能把当前账户并不持有的链带进去，或者错过本来应该直接选中的单子 token。

这里应该改成 owner-scoped membership，例如 bg frame 的 `ownedAggregateTokenListMap` 或本地 `getLocalAggregateTokenListMap()` 的结果，不要再用全局 `allAggregateTokenMap` 充当前账户的子 token 集合。

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -174,6 +193,56 @@ function isSameSelectorTokenListRequestContext(
...
+  key,
+  snapshot,
+}: {
+  key: string | undefined;
+  snapshot: T;
+}) {
+  if (key) {
+    swrCacheUtils.set(key, snapshot);
+  }
+}
```
sidmorizon, line 212: **Token selector view cache is bounded but uses in-memory swrCacheUtils — persistence implications**

Type: `Flag`

The new token selector caching (`TOKEN_SELECTOR_VIEW_CACHE_MAX_TOKEN_ROWS = 300`) stores snapshots via `swrCacheUtils.set()` which persists to MMKV (a shared native resource). Each snapshot includes full `IAccountToken[]` arrays and `Record<string, ITokenFiat>` maps. With the 300-row cap and the SWR cache's own `MAX_ENTRIES` eviction, this is bounded. However, the token data includes fields like `name`, `symbol`, `address`, `logoURI` etc. per token — for 300 tokens across multiple cache keys (normal + filtered + active-account variants), this could add meaningful MMKV write pressure on frequent selector opens. The `readNormalTokenSelectorViewSnapshot` / `readScopedTokenSelectorViewSnapshot` functions correctly evict oversized entries.

development/scripts/home-token-refresh-stress.mjs
```diff
@@ -0,0 +1,660 @@
...
+    runIteration(iteration);
+  }
+  checkCrash('run-complete');
+  writeSummary('passed-no-crash-detected');
+  writeEvent({ phase: 'run:done', status: 'passed-no-crash-detected' });
+  console.log(`[home-token-stress] done: ${summaryPath}`);
+} catch (e) {
+  writeEvent({ phase: 'run:error', error: String(e) });
+  writeSummary(crashDetected ? 'failed-crash-detected' : 'failed-command-error');
+  console.error(`[home-token-stress] failed: ${String(e)}`);
```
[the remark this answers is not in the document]
huhuanming: Fixed: sanitized the stress script error logging so it no longer writes raw error strings to events or stderr, and kept only the error type in logs.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
huhuanming: Fixed: removed allAggregateTokenMap from the initial selector aggregate membership fallback. Owner-scoped aggregate membership now comes from the VM frame, local owner cache, or live selector fetch only.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
huhuanming: Fixed: removed routeTokens.data from the initial selector cache fallback. The selector now only seeds rows from validated SWR/live data or the home VM floor path, where risky tokens are filtered with the owner risky set.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
huhuanming: Fixed: added a per-snapshot token-row cap for token selector SWR entries. Oversized normal/scoped snapshots are not persisted and any oversized existing entry for that key is removed, limiting shared MMKV pressure across main/bg runtimes.

development/scripts/home-token-refresh-stress.mjs
```diff
@@ -0,0 +1,680 @@
...
+  checkCrash('run-complete');
+  writeSummary('passed-no-crash-detected');
+  writeEvent({ phase: 'run:done', status: 'passed-no-crash-detected' });
+  console.log(`[home-token-stress] done: ${summaryPath}`);
+} catch (e) {
+  writeEvent({ phase: 'run:error', error: formatErrorForLog(e) });
+  writeSummary(
+    crashDetected ? 'failed-crash-detected' : 'failed-command-error',
+  );
+  console.error(`[home-token-stress] failed: ${formatErrorForLog(e).name}`);
```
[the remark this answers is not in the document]
huhuanming: Fixed: removed the caught error object from the stress script failure logs entirely. The catch path now emits fixed failure events/messages only, so CodeQL should no longer track passwordClose-derived data into stderr.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1898,6 +1979,9 @@ function TokenListBlock({
         tokenList = localTokens.tokenList;
         smallBalanceTokenList = localTokens.smallBalanceTokenList;
         riskyTokenList = localTokens.riskyTokenList;
+        tokenListMap = localTokens.tokenListMap;
+        smallBalanceTokenListMap = localTokens.tokenListMap;
+        riskyTokenListMap = localTokens.tokenListMap;
```
sidmorizon: **Single-network cache ingest receives the full combined token map as the risky-only map, potentially misidentifying normal tokens as risky**

Severity: `non-severe`

The risky token fiat map is populated with ALL token entries (`localTokens.tokenListMap` at `TokenListBlock.tsx:1984`) instead of only risky-token entries, so the downstream view model receives a risky map that includes normal and small-balance tokens.

Impact: On single-network non-merge-derive owners, the cell view model's risky-token map contains entries for every token, which may cause incorrect risky-token classification or filtering.

<details><summary>Mechanism: non-merge-derive path assigns the combined map to all three map variables</summary>

In the merge-derive path (`TokenListBlock.tsx:1965-1967`), the three maps are correctly separated:
```
tokenListMap = tokenListData.tokenListMap;
smallBalanceTokenListMap = tokenListData.smallBalanceTokenListMap;
riskyTokenListMap = tokenListData.riskyTokenListMap;
```

But in the non-merge-derive path (`TokenListBlock.tsx:1982-1984`), all three are assigned the same combined map:
```
tokenListMap = localTokens.tokenListMap;
smallBalanceTokenListMap = localTokens.tokenListMap;
riskyTokenListMap = localTokens.tokenListMap;
```

`localTokens.tokenListMap` is a single map containing fiat entries for ALL tokens (normal + small balance + risky). When passed to `buildHomeTokenListCacheIngestRound`, the `riskyMap` output (`buildHomeTokenListCacheIngestRound.ts:90-92`) becomes `{ ...riskyTokenListMap }` which is the full combined map rather than just the risky subset.

The unit test at `buildHomeTokenListCacheIngestRound.test.ts:63` explicitly asserts `riskyMap` only contains risky entries, confirming the intended contract.

</details>

Suggested fix:

```ts
        smallBalanceTokenListMap = localTokens.tokenListMap;
        riskyTokenListMap = localTokens.tokenListMap;
```

packages/kit/src/components/TokenListView/computeShowTokenListSkeleton.ts
```diff
@@ -45,7 +45,8 @@ export function computeShowTokenListSkeleton(
   if (
     p.showActiveAccountTokenList &&
     !p.activeAccountTokenListInitialized &&
-    p.activeAccountTokenListIsRefreshing
+    p.activeAccountTokenListIsRefreshing &&
+    p.displayCount === 0
   ) {
```
sidmorizon, line 56: **Skeleton suppression when displayCount > 0 may show stale data during owner mismatch in active-account and selector paths**

Type: `Flag`

The new `&& p.displayCount === 0` guard added to the active-account skeleton condition (`computeShowTokenListSkeleton.ts:48-49`) and the token-selector condition (lines 81-82) means that if cached rows from a PREVIOUS owner are still rendered (displayCount > 0) while the new owner's data is loading, the skeleton will NOT show. This is the intended cold-start behavior (show cached rows instead of skeleton), but it relies on the caller correctly resetting `displayCount` to 0 on owner switch. The `ownerMismatch` guard at line 69 handles the home path, but the active-account and selector paths don't have an equivalent owner-mismatch check — they rely on the consuming component clearing the list state on context change.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
huhuanming: Fixed in the latest push: the single-network local token cache now splits the combined `$key -> fiat` map by the actual normal/small/risky token lists before building the cache ingest payload. I also applied the same split to the merge-derive local-cache path, so `riskyMap` only receives risky entries instead of the full combined map.

packages/kit/src/components/TokenListView/computeShowTokenListSkeleton.ts
huhuanming: Fixed in the latest push: `ownerMismatch` now takes precedence for skeleton decisions across home, active-account, and selector paths. Same-owner cached rows still render immediately, but cached rows from a previous owner no longer suppress the skeleton. Added coverage for active-account and selector cached-row owner switches.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -833,15 +1185,28 @@ function TokenSelector() {
...
+      });
+      setScopedActiveTokenList({
+        tokens: [],
+        keys: '',
+      });
+      setScopedActiveTokenListMap({});
+    }
 
     try {
       const { responses } = await fetchFilteredTokenSelectorTokens({
```
originalix: 这里的 `fetchFilteredTokenSelectorTokens()` 在 all-networks / merge-derive 模式下是 `continueOnError` fan-out，返回值里专门带了 `expectedResponseCount` 来表示本轮是否有子请求被静默丢掉。当前只读取 `responses`，后面就直接 `buildScopedActiveTokenListFromResponses()` 并写入 `writeScopedTokenSelectorViewSnapshot()`，一旦某个子网络或 derive 账户失败，就会把缺了一部分 token 的列表当成完整 snapshot 持久化下来。

这样下次再打开 selector 时，首先恢复出来的就是这份缺口缓存，相关网络/账户的 token 会持续“消失”，直到后续某次完整成功才会补回。

建议这里和 normal selector 分支保持同一套完整性判断：把 `expectedResponseCount` 一起接进来，fan-out 不完整时不要覆盖现有 snapshot，必要时继续保留当前已显示的缓存。

huhuanming: Fixed in the latest push: the filtered scoped selector path now reads `expectedResponseCount` from the fan-out, treats all-networks partial responses as incomplete, keeps any restored cached snapshot visible without overwriting it, and only writes a scoped selector snapshot when the fan-out is complete.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -833,31 +1185,56 @@ function TokenSelector() {
...
+          mergeDeriveAddressData,
+          onlyBackendIndexedNetworks: showLpTokensOnly,
+          tokenSelectorFilterParams,
+        });
 
       if (!isLatestRequest()) {
         return;
       }
 
+      const isIncompleteAllNetworksFanOut =
```
originalix, line 1228: 这里已经拿到了 `expectedResponseCount`，说明这条分支能区分“完整 fan-out”和“有子请求被静默丢掉的 partial round”。但当前只有 `hasRestoredSnapshot` 时才提前返回；如果这是首次打开或缓存刚失效，`responses.length < expectedResponseCount` 仍会继续拼成 `initialized: true` 的列表。

因为 child failure 在 `continueOnError` 下不会抛错，用户看到的是一个没有错误提示的缺币 selector，而且 scoped 分支没有后续 floor/source 会把缺失网络补回来。

建议把 incomplete round 统一当成未完成状态处理：不要直接渲染 ready list，至少继续保留 loading/错误态或显式 retry。

huhuanming: Fixed in `88503ada62`: the scoped filtered selector now treats any incomplete all-networks fan-out as an unfinished round before building a ready snapshot. If a cached snapshot was restored, it keeps that cache visible and stops the refresh indicator. If there is no restored cache, it leaves the selector in the initial loading state and shows the network error toast instead of rendering a partial token list or writing a snapshot.

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -833,31 +1185,64 @@ function TokenSelector() {
...
+          });
+        } else {
+          setScopedActiveTokenListState({
+            initialized: false,
+            isRefreshing: true,
+          });
+          showFetchTokenListErrorToast();
+        }
+        return;
+      }
```
sidmorizon, line 1244: **Incomplete all-networks fan-out in filtered token selector leaves skeleton state without auto-recovery**

Type: `Flag`

When the filtered token selector's all-networks fan-out is incomplete (`responses.length < expectedResponseCount`) and there's no restored snapshot, the state is set to `initialized: false, isRefreshing: true` and an error toast is shown (`TokenSelector.tsx:1237-1244`). However, the function returns early without ever setting `isRefreshing: false`, so the skeleton persists until the next `usePromiseResult` revalidation (focus change, polling, etc.).

This is acceptable because `usePromiseResult` will re-fetch on focus revalidation, and the error toast informs the user. But if the user stays on the page without switching focus, the skeleton persists indefinitely. The old code had no incomplete-fan-out handling at all (it would show partial data), so this is an improvement. Worth noting for QA: a network error during filtered token selector loading on all-networks mode will show a persistent skeleton until the user navigates away and back.

huhuanming: Fixed in `0eb042156c`: when the filtered selector has no restored snapshot and the all-networks fan-out is incomplete, it now ends the loading state with `initialized: true, isRefreshing: false` and shows the existing network error toast. It still does not build or cache a partial token list, but it also no longer leaves the selector on an indefinite skeleton.

weatherstar (approved):

originalix (approved):

PR #12300, opened 2026-07-06
fix: prevent all-network BTC asset hiding(OK-57370)

weatherstar:
[OK-57370](https://onekeyhq.atlassian.net/browse/OK-57370)

----------

## Summary
- Await Token service current-account synchronization before all-network token fan-out starts.
- Use aggregate-aware fiat values when partitioning all-network token rows and calculating the small-balance scalar.
- Add regression coverage for aggregate BTC rows in all-network snapshot merging.

## Intent & Context
The reported issue was that on iOS, after switching accounts in All Networks, an account with BTC could briefly show the correct BTC asset and then hide it after a delay. The issue was hard to reproduce and appeared probabilistic, so this PR applies defensive hardening to the all-network Home token pipeline rather than relying on a narrow reproduction.

## Root Cause
Two risk paths were identified during investigation:

1. The Token all-network fan-out could start before the background Token service current account/network was explicitly awaited. ServiceToken uses its bg runtime current network guard for all-network requests, while the previous Token path only fired updateCurrentAccount from init as a fire-and-forget RPC.
2. buildMergedAllNetworkSnapshot sorted rows with aggregate fiat included, but its zero/unavailable tail partition and small-balance scalar used only the normal token fiat map. Aggregate rows such as BTC can store fiat in the flattened aggregate map, so they could be treated as zero/unavailable during the final authoritative snapshot.

## Design Decisions
- Followed the existing DeFi all-network pattern by awaiting current-account synchronization in onStarted before fan-out work continues.
- Kept the snapshot merge shape unchanged and introduced a local displayTokenFiatMap that combines normal token fiat with flattened aggregate fiat for display-only partitioning and scalar calculation.
- Added pure-function tests instead of UI-level tests because the issue is intermittent and the deterministic risk is in snapshot projection logic.

## Changes Detail
- packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx: await serviceToken.updateCurrentAccount in all-network request startup.
- packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.ts: use aggregate-aware fiat when sorting partition boundaries and calculating small-balance fiat.
- packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts: cover aggregate BTC row partitioning and small-balance fiat scalar behavior.

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: Home Token list in All Networks, especially aggregate assets and account-switch refresh timing. The most visible target is iOS because its token fan-out concurrency is higher, but the merge correction is shared by all platforms.

## Test plan
- [x] yarn jest packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts --runInBand
- [x] yarn tsc:staged
- [x] npx oxlint --tsconfig ./tsconfig.json --type-aware --deny-warnings packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.ts packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts

[OK-57370]: https://onekeyhq.atlassian.net/browse/OK-57370?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
+++ b/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
@@ -1211,12 +1211,13 @@ function TokenListBlock({
       accountId?: string;
       networkId?: string;
     }) => {
-      if (accountId && networkId) {
-        void backgroundApiProxy.serviceToken.updateCurrentAccount({
-          accountId,
-          networkId,
-        });
-      }
+      const updateCurrentAccountTask =
+        accountId && networkId
+          ? backgroundApiProxy.serviceToken.updateCurrentAccount({
+              accountId,
+              networkId,
+            })
+          : Promise.resolve();
 
       perfTokenListView.markStart('allNetworkRequestsStarted_getRawData');
 
@@ -1226,6 +1227,7 @@ function TokenListBlock({
         backgroundApiProxy.simpleDb.riskTokenManagement.getRawData(),
         backgroundApiProxy.simpleDb.localTokens.getRawData(),
         backgroundApiProxy.simpleDb.aggregateToken.getRawData(),
+        updateCurrentAccountTask,
       ]);
 
       perfTokenListView.markEnd('allNetworkRequestsStarted_getRawData');
--- a/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
+++ b/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
@@ -1211,6 +1211,13 @@ function TokenListBlock({
       accountId?: string;
       networkId?: string;
     }) => {
+      if (accountId && networkId) {
+        await backgroundApiProxy.serviceToken.updateCurrentAccount({
+          accountId,
+          networkId,
+        });
+      }
+
       perfTokenListView.markStart('allNetworkRequestsStarted_getRawData');
 
       // eslint-disable-next-line prefer-const
--- a/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
+++ b/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
@@ -1212,7 +1212,7 @@ function TokenListBlock({
       networkId?: string;
     }) => {
       if (accountId && networkId) {
-        await backgroundApiProxy.serviceToken.updateCurrentAccount({
+        void backgroundApiProxy.serviceToken.updateCurrentAccount({
           accountId,
           networkId,
         });
--- a/packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts
+++ b/packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts
@@ -8,6 +8,7 @@
  */
 import BigNumber from 'bignumber.js';
 
+import { TOKEN_LIST_HIGH_VALUE_MAX } from '@onekeyhq/shared/src/consts/walletConsts';
 import accountUtils from '@onekeyhq/shared/src/utils/accountUtils';
 import type { IAccountToken, ITokenFiat } from '@onekeyhq/shared/types/token';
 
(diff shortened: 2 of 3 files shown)


packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1211,6 +1211,13 @@ function TokenListBlock({
       accountId?: string;
       networkId?: string;
     }) => {
+      if (accountId && networkId) {
+        await backgroundApiProxy.serviceToken.updateCurrentAccount({
+          accountId,
+          networkId,
+        });
+      }
```
sidmorizon: **New await in handleAllNetworkRequestsStarted adds sequential latency before data fetch**

Type: `Flag`

The `await backgroundApiProxy.serviceToken.updateCurrentAccount(...)` at `TokenListBlock.tsx:1215` is a cross-runtime IPC call (main→bg) that must complete before the `Promise.all` for raw data fetching begins (line 1224). While `updateCurrentAccount` (`ServiceBase.ts:180-194`) only sets two properties and should resolve quickly, the IPC round-trip adds latency to the critical path of starting all-network requests. Consider whether this could be fired without awaiting (using `void` prefix) since the raw data fetch doesn't depend on the account being set first, or whether it could be included in the existing `Promise.all`.

sidmorizon: **updateCurrentAccount receives the all-networks pseudo-network ID**

Type: `Flag`

The `networkId` passed to `updateCurrentAccount` is `currentNetworkId` from `useAllNetworkRequests`, which equals `network?.id` — the all-networks pseudo-network `'onekeyall--0'` (see `packages/shared/src/config/presetNetworks.ts:38`). This sets `serviceToken._currentNetworkId = 'onekeyall--0'`. Verify that downstream code reading `_currentNetworkId` handles this pseudo-network ID correctly and doesn't treat it as a real chain network.

weatherstar: Verified in e2e6c7847f. For the all-network token path, main intentionally sends the onekeyall networkId to bg; ServiceToken.fetchAccountTokens checks the bg _currentNetworkId against getNetworkIdsMap().onekeyall. These are per-runtime JS fields in the bg service, not shared native state, and main/bg JS heaps remain isolated.

weatherstar: Fixed in e2e6c7847f. The main runtime now dispatches serviceToken.updateCurrentAccount to bg with void, so the raw-data Promise.all starts immediately instead of waiting for main->bg IPC. The bg runtime still receives the current-account update for the all-network guard; no shared native resource is involved, only bg service JS fields.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -1211,6 +1211,13 @@ function TokenListBlock({
       accountId?: string;
       networkId?: string;
     }) => {
+      if (accountId && networkId) {
+        void backgroundApiProxy.serviceToken.updateCurrentAccount({
```
sidmorizon: # 🟡 P2 `void` 使 OK-57370 修复退化为隐式时序保证

## 问题

本次提交把 `await` 改成 `void backgroundApiProxy.serviceToken.updateCurrentAccount({...})`。修复的正确性由此从显式 happens-before 变成隐式约定，仅在两个前提同时成立时才成立：(a) bg 端 `updateCurrentAccount` 在写入 `_currentNetworkId` 之前没有任何 `await`（`ServiceBase.ts` 当前是同步的），(b) `onStarted` 仍会在扇出闸门（`useAllNetwork.ts` 的 `await onStartedTask`）之前 `await` `Promise.all([...getRawData])`，从而让先入队的 update RPC 通过 FIFO 先于活跃 `fetchAccountTokens` 到达 bg guard。

当前两个前提都成立，所以没有实际竞态；但这个正确性依赖没有任何测试或类型层面的保护。

## 影响

后续任意一次重构——在 `updateCurrentAccount` 赋值前加入 `await`、调整/移除 `getRawData` 往返、或把这行移动到扇出之后——都会静默破坏该顺序，`ServiceToken.ts` 的 guard（`currentNetworkId !== getNetworkIdsMap().onekeyall` → `getEmptyTokenData()`）会重新在活跃扇出阶段命中，从而重新出现本 PR 要修复的全网络 BTC 资产被隐藏问题，且不会有任何测试/类型报错。此外此处 `void` 的性能收益有限，因为扇出本身仍阻塞在 `onStartedTask` 上。

## 建议

恢复显式顺序（成本可忽略）：改回 `await`，或把 `updateCurrentAccount` 并入第 `1224` 行现有的 `Promise.all`，让扇出闸门可证明地包含它。若坚持保留 `void`，至少补一行注释，说明其对 FIFO IPC 顺序与 bg 同步赋值的依赖。

[Claude session ↗](https://claude.ai/code/session_01QXXz9ob3udNZnEsmm24McD)

---

weatherstar: Fixed in 7f692da5cb. In the main runtime, handleAllNetworkRequestsStarted now includes serviceToken.updateCurrentAccount in the onStarted Promise.all, so useAllNetwork awaits the bg current-account update before the active token fan-out. In the bg runtime, the guard still reads ServiceToken per-runtime JS fields; no shared native resource is involved, and main/bg JS heaps remain isolated.

originalix (approved):

PR #12301, opened 2026-07-06
fix: reset reward center account on open

weatherstar:
## Summary
- Reset the RewardCenter account selector from the launching home account each time the modal opens.
- Make RewardCenter account selector state non-persistent and scoped per modal instance.
- Send the TRON reward `/api/account` request address in the request body instead of query params.

## Intent & Context
The RewardCenter modal was expected to open with the wallet account currently selected on Home. The reported reproduction was: Home shows account 1, open the free energy modal, switch the modal selector to account 666, close the modal, Home still shows account 1, then reopen the free energy modal and it still shows account 666. The modal account switch should only affect the current RewardCenter session, not the next open from Home.

## Root Cause
RewardCenter used its own stable `rewardCenter` account selector scene. When the user selected account 666 inside the modal, that scene persisted the selection and reused the same in-memory/store state on the next open. Because Home and RewardCenter use separate selector scenes, closing the modal did not update Home, but reopening restored RewardCenter's previous account instead of the Home account that launched it.

The TRON account activation check also sent `fromAddress` as query params for `/api/account`; the endpoint expects it in the request body.

## Design Decisions
RewardCenter now creates a per-open selector scene key and initializes that temporary selector from the route account before rendering the modal content. The `rewardCenter` scene is also marked non-persistent, so account switches inside the modal remain local to that modal instance.

This was chosen over only forcing the account after mount because the old persistent scene could still render or restore stale state before the override. A transient per-open context matches the product behavior: RewardCenter opens from Home, while modal-local account changes are temporary.

## Changes Detail
- `packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx`
  - Adds a RewardCenter selector sync wrapper that initializes from `accountId/networkId/walletId`.
  - Uses `forceSelectToNetworkId` so the modal opens on the launching TRON network.
  - Uses a per-open `sceneUrl` for the RewardCenter selector provider and header trigger.
  - Moves `fromAddress` into the `/api/account` request body.
- `packages/shared/src/utils/accountSelectorUtils.ts`
  - Allows RewardCenter selector stores to be keyed by `sceneUrl`.
  - Prevents RewardCenter selector selections from being persisted to account selector storage.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**: Account selector scene lifecycle and RewardCenter modal initialization. The change is scoped to RewardCenter, but account selector utility behavior now treats `rewardCenter` as non-persistent.

## Test plan
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [ ] On Home with account 1 selected, open the free energy modal and confirm account 1 is selected.
- [ ] Switch the modal account to account 666, close it, and confirm Home still shows account 1.
- [ ] Reopen the free energy modal and confirm it selects account 1 again.
- [ ] Verify claiming/redeeming still uses the currently selected modal account.

### DIFF
--- a/packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
+++ b/packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
@@ -291,10 +291,10 @@ function RewardCenterDetails() {
             body: {
               method: 'get',
               url: '/api/account',
-              data: {},
-              params: {
+              data: {
                 fromAddress: account.address,
               },
+              params: {},
             },
             returnRawData: true,
           });
--- a/packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
+++ b/packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
@@ -1,5 +1,11 @@
 // cspell:ignore Actived
-import { useCallback, useEffect, useRef, useState } from 'react';
+import {
+  type ReactNode,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 
 import { useRoute } from '@react-navigation/core';
 import BigNumber from 'bignumber.js';
@@ -61,7 +67,11 @@ import type { RouteProp } from '@react-navigation/core';
 
 const networkIdsMap = getNetworkIdsMap();
 
-function RewardCenterDetails() {
+function RewardCenterDetails({
+  selectorSceneUrl,
+}: {
+  selectorSceneUrl: string;
+}) {
   const route =
     useRoute<
       RouteProp<
@@ -73,7 +83,6 @@ function RewardCenterDetails() {
   const {
     accountId,
     networkId,
-    walletId,
     onClose,
     showAccountSelector = true,
   } = route?.params ?? {};
@@ -88,7 +97,6 @@ function RewardCenterDetails() {
   });
 
   const { activeAccount } = useActiveAccount({ num: 0 });
-  const { confirmAccountSelect } = useAccountSelectorActions().current;
 
   const navigation = useAppNavigation();
 
@@ -494,58 +502,6 @@ function RewardCenterDetails() {
     }
   }, [account, claimSource, form, intl, network]);
 
-  useEffect(() => {
-    const initActiveAccount = async () => {
-      const [initAccount, initWallet] = await Promise.all([
-        accountId && networkId
-          ? backgroundApiProxy.serviceAccount.getAccount({
-              accountId,
-              networkId,
-            })
-          : undefined,
-        walletId
-          ? backgroundApiProxy.serviceAccount.getWallet({
-              walletId,
-            })
-          : undefined,
-      ]);
-
-      const isOthersAccount = accountUtils.isOthersAccount({
-        accountId,
-      });
-      if (isOthersAccount) {
-        let autoChangeToAccountMatchedNetworkId = networkId;
-        if (
-          networkId &&
-          networkUtils.isAllNetwork({
-            networkId,
-          })
-        ) {
-          autoChangeToAccountMatchedNetworkId = networkId;
-        }
-        await confirmAccountSelect({
-          num: 0,
-          indexedAccount: undefined,
-          othersWalletAccount: initAccount,
-          autoChangeToAccountMatchedNetworkId,
-        });
-      } else if (initWallet) {
-        const indexedAccount =
-          await backgroundApiProxy.serviceAccount.getIndexedAccountByAccount({
-            account: initAccount,
-          });
-        await confirmAccountSelect({
-          num: 0,
-          indexedAccount,
-          othersWalletAccount: undefined,
-          autoChangeToAccountMatchedNetworkId: undefined,
-        });
-      }
-    };
-
-    void initActiveAccount();
-  }, [accountId, confirmAccountSelect, networkId, walletId]);
-
   useEffect(
     () => () => void onClose?.({ isResourceClaimed, isResourceRedeemed }),
     [onClose, isResourceClaimed, isResourceRedeemed],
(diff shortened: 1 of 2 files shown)


originalix (approved):

packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
```diff
@@ -291,10 +299,10 @@ function RewardCenterDetails() {
             body: {
               method: 'get',
               url: '/api/account',
-              data: {},
-              params: {
+              data: {
                 fromAddress: account.address,
               },
+              params: {},
```
sidmorizon, line 305: **Account activation check sends address in request body instead of query string for a GET request, so the server never receives it**

Severity: `severe`

The account address is placed in the request body (`data: { fromAddress: ... }` at `packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx:302-303`) instead of the query string for a GET request to `/api/account`, so the server proxy never receives the address parameter.

Impact: The account activation check always fails silently, so inactive TRON accounts are never detected — users see no "Top up" prompt and the claim button stays enabled for accounts that cannot actually claim energy.

<details><summary>Mechanism: old code used params (query string) but new code uses data (body) for a GET request</summary>

The old code at the base commit had:
```
method: 'get',
url: '/api/account',
data: {},
params: {
  fromAddress: account.address,
},
```

The new code swaps `data` and `params`:
```
method: 'get',
url: '/api/account',
data: {
  fromAddress: account.address,
},
params: {},
```

The `sendProxyRequestWithTrxRes` method (`packages/kit-bg/src/services/ServiceAccountProfile.ts:927`) forwards these fields to the server proxy at `/wallet/v1/proxy/trxres`. The proxy constructs the actual HTTP request using `params` for query string parameters and `data` for the request body. Since HTTP GET requests conventionally have no body, the `fromAddress` in `data` is effectively lost. The proxy sends `GET /api/account` with no query parameters, so the server cannot identify which account to check.

Because the response won't contain valid activation data, `accountActivationResult` stays `undefined`, `isAccountNotActivated` is never `true`, and the inactive-account guard is completely bypassed.
</details>

Suggested fix:

```suggestion
              data: {},
              params: {
                fromAddress: account.address,
              },
```

packages/shared/src/utils/accountSelectorUtils.ts
```diff
@@ -49,6 +49,10 @@ function buildAccountSelectorSceneId({
     return `${sceneName}--${origin}`;
   }
 
+  if (sceneName === EAccountSelectorSceneName.rewardCenter && sceneUrl) {
+    return `${sceneName}--${sceneUrl}`;
+  }
```
sidmorizon, line 54: **Unique sceneUrl per mount prevents cross-instance state leakage but creates unbounded scene IDs**

Type: `Flag`

Each `RewardCenter` mount generates a unique `sceneUrl` via `reward-center-${Date.now()}-${Math.random()}` at `packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx:1070`. Combined with `buildAccountSelectorSceneId` at `packages/shared/src/utils/accountSelectorUtils.ts:52-54` producing `rewardCenter--reward-center-...`, this means every reward center open creates a distinct scene ID. Since `isSceneCanPersist` returns `false` for `rewardCenter` (line 133), these IDs should not accumulate in persistent storage. However, if there's any in-memory map keyed by scene ID that isn't cleaned up on unmount, this could be a slow memory leak over many opens. Worth verifying that the account selector context properly cleans up ephemeral scenes.

huhuanming (approved):

packages/kit/src/views/RewardCenter/pages/RewardCenter.tsx
```diff
@@ -974,15 +931,156 @@ function RewardCenterDetails() {
...
+
+    void syncAccountSelector();
+
+    return () => {
+      cancelled = true;
+    };
+  }, [accountId, actions, networkId, showAccountSelector, targetWalletId]);
+
+  if (!isReady) {
+    return null;
```
sidmorizon, line 1062: # 🟡 P2 同步期间 `return null` 导致整个弹窗空白且无兜底

## 问题

`RewardCenterAccountSelectorSync` 在 `isReady === false` 时 `return null`，而 `Page` / `Page.Header`（含关闭按钮 `renderHeaderLeft`）都定义在其子组件 `RewardCenterDetails` 内。因此在异步 `syncAccountSelector`（`getAccount` / `getWallet` / `waitForAutoSelectUnlock` / `confirmAccountSelect` / `reloadActiveAccountInfo`）完成前，整个弹窗渲染为 `null`——既没有骨架屏，也没有页面头部。

`catch` 只能捕获 reject，无法捕获「挂起」。若链路中任一 `await`（例如 `waitForAutoSelectUnlock` 迟迟不解锁，或 `reloadActiveAccountInfo` 内的网络请求长时间不 resolve）卡住，`isReady` 将永远为 `false`。

## 影响

改动前 `RewardCenterDetails` 会立即渲染并用 `Skeleton` 占位；改动后从打开弹窗到同步完成期间是纯空白。正常情况下是短暂闪烁，但在慢网络/硬件钱包或异常挂起时，用户会长时间停留在没有头部、没有关闭入口的空白弹窗上，属于可感知的体验退化。

## 建议

保留 `Page` 外壳（含 header 与关闭按钮）常驻，仅对内容区做 gating；或在 `isReady === false` 时渲染一个带骨架屏的 `Page` 而非 `null`。同时给同步链路加一个超时兜底，超时后也置 `isReady = true`，避免永久空白：

```tsx
if (!isReady) {
  return <LoadingPageShell />; // 至少保留 header/关闭按钮
}
```

[Claude session ↗](https://claude.ai/code/session_01VNrrQ1R1EgnmiQLvnxhPfM)

---

sidmorizon (commented):
Reviewed the reward-center account-selector reset changes; one non-blocking note on the loading/gating behavior.

---

PR #12308, opened 2026-07-06
fix: prevent dapp browser reload loop(OK-57500)

huhuanming:
[OK-57500](https://onekeyhq.atlassian.net/browse/OK-57500)

----------

## Summary
- Normalize equivalent root URLs before comparing native DApp browser navigation state.
- Add regression coverage for Android WebView reporting a root URL with a trailing slash.
- Extract CSS into standalone files for production web Rspack builds while keeping style-loader in dev and non-web builds.

## Intent & Context
Android could not reliably open `https://app.osmosis.zone` in the native Discovery browser. Logs showed the Osmosis page repeatedly reinitializing and the app repeatedly calling `crossWebviewLoadUrl('https://app.osmosis.zone')`, which made the WebView reload itself.

The Rspack change was also requested to be included in this PR after the WebView fix was committed.

## Root Cause
The native WebView can report the root URL as `https://app.osmosis.zone/`, while OneKey stores the validated tab URL as `https://app.osmosis.zone`. The previous `onNavigation` logic compared raw strings, treated these equivalent URLs as different, and called `gotoSite` during `loading=true`; that path calls `crossWebviewLoadUrl`, causing a reload loop on Android.

## Design Decisions
- Normalize both the WebView-reported URL and current tab URL using the existing URL validation behavior before deciding whether navigation is actually new.
- Keep the comparison scoped to HTTP(S) URLs so deep links and other schemes keep their existing behavior.
- Keep Web Rspack production CSS extraction scoped to non-dev web builds; dev and other platforms continue using injected styles.

## Changes Detail
- `packages/kit/src/states/jotai/contexts/discovery/actions.ts`: compares normalized native WebView navigation URLs before calling `gotoSite`.
- `packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx`: adds regression coverage for `https://app.osmosis.zone` vs `https://app.osmosis.zone/`.
- `development/rspack/rspack.base.config.ts`: adds production web CSS extraction and shared CSS loader construction.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile, Web
- **Risk Areas**: Native DApp browser navigation state handling; production Web Rspack CSS output.

## Test plan
- [x] `yarn jest packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx --runInBand`
- [x] `npx oxlint --tsconfig ./tsconfig.json --type-aware packages/kit/src/states/jotai/contexts/discovery/actions.ts packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx --deny-warnings`
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn tsc:only`
- [x] Android simulator: opened `https://app.osmosis.zone`, confirmed the Osmosis page rendered and `crossWebviewLoadUrl` did not repeat during the observation window.

[OK-57500]: https://onekeyhq.atlassian.net/browse/OK-57500?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -167,6 +167,16 @@ function isLocalhostUrlAllowedInDAppBrowser() {
...
 
+function normalizeWebNavigationUrlForComparison(
+  url: string,
+  allowLocalhostUrl: boolean,
+) {
+  if (!/^https?:\/\//i.test(url)) {
+    return url;
+  }
+  return uriUtils.validateUrl(url, { allowLocalhostUrl });
+}
```
sidmorizon, line 186: **Trailing-slash normalization only applies to root paths on web/desktop**

Type: `Flag`

The `validateUrl` function in `packages/shared/src/utils/uriUtils.ts:617-627` only strips the root pathname `/` universally; non-root trailing slashes (e.g., `/swap/`) are only stripped on native (`platformEnv.isNative`). This means on web/desktop, a webview reporting `https://app.example.com/swap/` when `tab.url` is `https://app.example.com/swap` will still be treated as a new URL and trigger `gotoSite`. The PR's test (`actions.test.tsx:907-953`) only covers the root-path case and runs with `isNative: true`. If the trailing-slash suppression is also needed for non-root paths on web/desktop, additional handling would be required.

huhuanming: Thanks for flagging this. The trailing-slash suppression is intended only for native WebView (iOS/Android). Web/desktop start navigation from different timing paths, so we should not broaden this comparison there. I updated the helper to return the original URL when `!platformEnv.isNative`, added a comment explaining that scope, and added a desktop regression test to lock the behavior. Commit: 68a395ffcd.

packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx
```diff
@@ -904,6 +904,112 @@ describe('useBrowserTabActions', () => {
...
+    );
+  });
+
+  it('does not apply native trailing-slash suppression to desktop navigation', async () => {
+    Object.assign(platformEnv, {
+      isDesktop: true,
+      isNative: false,
+      isNativeAndroid: false,
+      isNativeIOS: false,
+    });
```
sidmorizon: **Desktop test mutates shared platformEnv mock without visible cleanup**

Type: `Flag`

The test at line 956 uses `Object.assign(platformEnv, { isDesktop: true, isNative: false, ... })` to override the module-level mock (`packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx:74-83`). This mutates the shared mock object in place. If no `beforeEach`/`afterEach` restores the original values, every test that runs AFTER this one in the same file will see `isNative: false` and `isDesktop: true`, which could silently change their behavior. The test is placed near the end of the file so current tests may not be affected, but any future test appended below it would inherit the wrong platform flags.

development/rspack/rspack.base.config.ts
```diff
@@ -289,9 +289,31 @@ const buildBasePlugins: (
         '../../packages/kit/src/views/Developer/router.empty.ts',
       ),
     ),
+  !isDev &&
+    platform === 'web' &&
+    new rspack.CssExtractRspackPlugin({
+      filename: '[name].[contenthash:10].css',
+      chunkFilename: 'static/css/[name].[contenthash:10].chunk.css',
+    }),
```
sidmorizon, line 297: **CSS extract plugin and loader are in separate config builders that must be composed correctly**

Type: `Flag`

The `CssExtractRspackPlugin` instance is added in `buildBasePlugins` (line 292-297), while the corresponding `CssExtractRspackPlugin.loader` is used inside `buildCssLoaders` which is called from `createBaseConfig` (line 721). These two config builders are separate exports. If a platform-specific config calls `createBaseConfig` but does not include `buildBasePlugins` in its final plugins array, the extract loader will be present without the plugin, causing a build failure in production (`!isDev && platform === 'web'`). This follows the existing pattern (the `NormalModuleReplacementPlugin` was already split the same way), so it's not a new risk, but it's worth verifying that all web production configs compose both correctly.

huhuanming: Verified. `createBaseConfig()` already includes `buildBasePlugins(platform, basePath)` in its own `plugins` array, so any caller that uses `createBaseConfig` gets the base plugin set together with the module rules. `createWebConfig()` then merges the production config and appends the web-specific plugins on top of that base config. I also instantiated the production web config locally and confirmed `CssExtractRspackPlugin` and `CssExtractRspackPlugin.loader` are both present for `platform === "web"` / `NODE_ENV=production`, so no config change is needed here.

packages/kit/src/states/jotai/contexts/discovery/actions.test.tsx
huhuanming: Good point. There was already a `beforeEach` resetting the shared `platformEnv` mock before every case, so current tests were not leaking state. I made that isolation explicit by extracting a `resetPlatformEnvMock()` helper and calling it from both `beforeEach` and `afterEach`, so future tests appended after the desktop case cannot inherit the desktop/native override. Commit: 36de8f15f3.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -1302,7 +1320,17 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
...
+      const comparableUrl =
+        typeof url === 'string'
+          ? normalizeWebNavigationUrlForComparison(url, allowLocalhostUrl)
+          : url;
+      const comparableTabUrl =
+        typeof tab.url === 'string'
+          ? normalizeWebNavigationUrlForComparison(tab.url, allowLocalhostUrl)
+          : tab.url;
+      const isValidNewUrl =
+        typeof url === 'string' && comparableUrl !== comparableTabUrl;
```
sidmorizon: **Desktop trailing-slash navigation intentionally triggers gotoSite which calls crossWebviewLoadUrl**

Type: `Flag`

On desktop, `normalizeWebNavigationUrlForComparison` returns the raw URL (line 178-179), so a trailing-slash difference like `https://app.osmosis.zone/` vs `https://app.osmosis.zone` will set `isValidNewUrl = true` and trigger `gotoSite`. Inside `gotoSite` (`actions.ts:1056-1060`), `validateUrl` normalizes the URL back to `https://app.osmosis.zone`, and `crossWebviewLoadUrl` is called to reload the webview with the normalized URL. The test at line 1009 explicitly asserts this behavior. This means desktop Electron webviews will get an extra reload when the webview reports a trailing-slash URL, which is the intended behavior per the test name 'does not apply native trailing-slash suppression to desktop navigation'. Worth confirming this doesn't cause a visible flash or navigation loop on desktop Electron webviews in practice.

huhuanming: Fixed. The desktop test no longer locks in the extra reload behavior. I added a narrow non-native no-op guard: we still keep raw URL comparison for non-native navigation, but if the `gotoSite` load path would validate/canonicalize the reported URL back to the current `tab.url`, we skip the reload. This avoids the desktop root trailing-slash reload without broadening non-root trailing-slash suppression on web/desktop. Commit: fb54a02fc5.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -167,6 +167,42 @@ function isLocalhostUrlAllowedInDAppBrowser() {
...
+  // to native WebView (iOS/Android), where the reported normalized URL can race
+  // the tab URL and otherwise trigger a reload loop.
+  if (!platformEnv.isNative) {
+    return url;
+  }
+
+  if (!/^https?:\/\//i.test(url)) {
+    return url;
+  }
+  return uriUtils.validateUrl(url, { allowLocalhostUrl });
```
originalix, line 185: 当 `allowLocalhostUrlInDAppBrowser` 打开时，这里仍会把 `http://127.0.0.1:3000` 和 `http://127.0.0.1:3000/` 当成两个不同地址。

原因是比较逻辑直接复用了 `validateUrl()`，而它在 localhost/IP 白名单分支会先走 `normalizeHttpLocalUrl()`，对根路径不会折叠尾部 `/`，所以 `comparableUrl` 这一层仍然一边带 `/`、一边不带。

结果就是本地 DApp 调试链路里的 `onNavigation()` 还会继续进入 `gotoSite()` / `crossWebviewLoadUrl()`，这次想修掉的重复 reload 在 localhost 场景没有真正消失。

建议把“等价比较”改成独立的 root-only normalizer，对允许的 localhost/IP 根路径也统一折叠尾部 `/`，并补一条 `http://127.0.0.1:3000/` 的回归测试。

huhuanming: 这里不处理。`localhost` / private IP 是 `allowLocalhostUrlInDAppBrowser` 打开后的本地调试路径，不是这次线上 Android Osmosis reload 问题的产品路径。这个 PR 的修复范围会保持在真实 DApp 导航场景，不为本地或测试环境额外扩大 URL comparison/normalization 规则，也不新增 localhost 回归测试。

如果后续要系统性调整 localhost/private IP 的 URL 语义，应单独开任务评估，不能混进这个修复里。

limichange (approved):

originalix (approved):

PR #12320, opened 2026-07-07
fix: preserve bulk send validation during debounce(OK-57475)

weatherstar:
[OK-57475](https://onekeyhq.atlassian.net/browse/OK-57475)

----------

## Summary
- Preserve pending Bulk Send field validation when the validator function identity changes during the debounce window.
- Add regression tests for validator churn, invalid result propagation, and repeated validation calls.
- Include generated locale updates that are already part of this branch.

## Intent & Context
Bulk Send validation could get stuck when the user pasted or entered addresses while async context such as network or vault settings was still settling. If `validateFn` changed before the debounce timer fired, the pending validation was cancelled with `false`, leaving the form invalid without a visible error message and keeping the Next button disabled.

This PR keeps the pending validation alive and re-checks the current value with the latest validator when the debounce fires.

## Root Cause
`useDebouncedValidation` previously treated `validateFn` or `delay` changes as a reason to increment the validation version and cancel the pending promise with `false`. In react-hook-form, a boolean `false` result marks the field invalid but does not provide an error string, so invalid Bulk Send input could show no error text while still blocking progression.

## Design Decisions
- Store the latest `validateFn` in a ref so dependency churn does not cancel the active debounce.
- Keep `delay` as a dependency of the debounced callback, but avoid cancelling just because the validator identity changes.
- Continue using version and current-value guards so stale timers cannot overwrite the latest validation result.

## Changes Detail
- `packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts`: reads the validator from a ref when the debounce fires instead of cancelling on validator identity changes.
- `packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts`: updates the previous cancellation expectation and adds coverage for valid/invalid results and repeated validate calls.
- `packages/shared/src/locale/*`: includes generated locale updates from the branch.

## Risk Assessment
- **Risk Level**: Low
- **Runtime Impact**: Main JS form validation only. No background runtime behavior, shared native resource, or per-runtime JS heap duplication changes.
- **Risk Areas**: Debounced form validation timing, especially superseded values and validator identity churn.

## Test plan
- [x] `npx oxlint --tsconfig ./tsconfig.json --type-aware --deny-warnings packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts`
- [x] `yarn jest packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts --runInBand`
- [x] `yarn tsc:staged`

[OK-57475]: https://onekeyhq.atlassian.net/browse/OK-57475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix (approved):

zhaono1 (approved):

packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
```diff
@@ -58,7 +65,7 @@ export function useDebouncedValidation<T extends string>(
         debounceTimerRef.current = setTimeout(async () => {
           debounceTimerRef.current = null;
           try {
-            const result = await validateFn(value);
+            const result = await validateFnRef.current(value);
```
originalix, line 68: 这里虽然通过 `validateFnRef.current` 解决了“debounce 尚未触发时 validator identity 变化”的场景，但一旦 300ms 已经到期、旧 `validateFn` 已经进入 `await`，后续的 `vaultSettings` / network / token 更新就不会再让这次校验失效了。

触发条件是用户输入后 debounce 已经起跑，而 `useMultiLineAddressValidation` 依赖的异步上下文稍后才收敛。此时旧 validator 会按旧上下文算出结果，并在这里直接 `resolve`；因为 PR 里已经移除了 `validateFn` 变化时的失效逻辑，这个过期结果会覆盖当前表单状态。最直接的影响是最小转账额、allowlist 或网络相关错误可能继续显示旧结果，直到用户再次编辑输入框。

建议把“pre-debounce 读最新 validator”和“post-debounce 丢弃旧 validator 结果”分开处理：保留 ref 方案，但在 `validateFn` 变化时仍递增一个版本号或 request id（不要再强制 `resolve(false)`），让已经启动的旧异步校验在返回时被忽略，并补一条 `debounce fired -> validateFn changes -> old async resolves` 的回归测试。

PR #12326, opened 2026-07-07
fix: ignore stale bulk send validation results(OK-57475)

weatherstar:
[OK-57475](https://onekeyhq.atlassian.net/browse/OK-57475)

----------

## Summary
- Ignore stale Bulk Send validation results when the validator changes after the debounce timer has already fired.
- Re-run the current field value with the latest validator context instead of resolving the form with an old async result.
- Add regression coverage for validator churn during in-flight validation.

## Intent & Context
PR #12320 fixed the pre-debounce validator identity churn case, but one unresolved review comment pointed out a second race: after the 300ms debounce fires, the old validator can already be awaiting background checks while network, token, vault settings, or allowlist context changes. In that case, the old async result could still resolve the React Hook Form validation promise and leave stale errors or stale success state until the user edits the field again.

This change addresses that unresolved review comment without restoring the old behavior that force-resolved pending validation to false.

## Root Cause
useDebouncedValidation stored the latest validateFn in a ref and read it when the timer fired, but it no longer invalidated validator changes after validation had started. The existing guards only checked the validation version, current value, and resolver identity. None of those changed when only validateFn context changed during an awaited async call, so stale results could be accepted.

## Design Decisions
- Keep the validateFn ref so pre-debounce validator churn still uses the latest context.
- Track a validateFn version separately so post-debounce async results can detect that their validator context became stale while awaiting.
- Re-run the same current value with the latest validator before resolving the pending form validation promise, avoiding both stale results and hanging React Hook Form validation.
- Keep cancellation behavior based on the last known result so Android controlled TextInput behavior remains unchanged.

## Changes Detail
- packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts: adds validator version tracking and an active validation run guard, and reruns stale in-flight validations with the latest validator.
- packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts: adds a regression test for debounce fired -> validateFn changes -> old async resolves, proving the stale result is ignored and the latest result resolves the form.

## Risk Assessment
- **Risk Level**: Low
- **Runtime Impact**: Main JS form validation only. Background JS services may still be called by the validator, but this change does not alter bg runtime state, shared native resources, or per-runtime JS heap residency.
- **Risk Areas**: Debounced validation timing and React Hook Form promise settlement. The main risk is repeated validation if validator identity changes frequently while async validation is in flight.

## Test plan
- [x] npx oxlint --tsconfig ./tsconfig.json --type-aware --deny-warnings packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts
- [x] yarn jest packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.test.ts --runInBand
- [x] yarn lint:staged
- [x] yarn tsc:staged

[OK-57475]: https://onekeyhq.atlassian.net/browse/OK-57475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
+++ b/packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
@@ -9,17 +9,12 @@ export function useDebouncedValidation<T extends string>(
   validateFn: (value: T) => Promise<string | boolean>,
   delay = 300,
 ): IDebouncedValidation<T> {
-  // Keep the latest validateFn in a ref so identity churn (e.g. `network` /
-  // `vaultSettings` resolving asynchronously while the user is entering
-  // addresses) never cancels an in-flight validation. Previously, a changed
-  // validateFn force-resolved the pending validation to `false`, which left the
-  // field invalid with an empty error message: react-hook-form renders no error
-  // text for a boolean-`false` result, so the Next button stayed disabled and
-  // invalid addresses produced no visible error. When the debounce fires we
-  // always run the current validator, so the value is re-checked with the
-  // freshest network/token context and resolves with the real result.
   const validateFnRef = useRef(validateFn);
-  validateFnRef.current = validateFn;
+  const validateFnVersionRef = useRef(0);
+  if (validateFnRef.current !== validateFn) {
+    validateFnRef.current = validateFn;
+    validateFnVersionRef.current += 1;
+  }
 
   const currentValueRef = useRef<T>('' as T);
   const debounceTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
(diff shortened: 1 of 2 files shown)


packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
```diff
@@ -41,6 +85,19 @@ export function useDebouncedValidation<T extends string>(
     }
   }, []);
 
+  useEffect(() => {
```
originalix, line 88: 当 `validateFn` 在 debounce 已经触发后发生切换，而旧 Promise 又在 `useEffect` flush 之前很快返回时，这里会出现双重重跑：前面的 in-flight run 会在 `validateFnVersionRef` 检查里再次调用最新 `validateFn`，这个 effect 又会因为 `pendingResolveRef` 仍然存在再起一轮 `startValidation()`。

这样同一个 value 会把最新 validator 执行两次。`handleValidateAddresses` 这类 validator 里不仅有异步地址/allowlist 查询，还有 `setErrors`、duplicate count、toast 等副作用，所以即使最终只有最后一个 promise resolve 给表单，重复调用还是会放大后台工作量，并可能重复写状态。

建议把 in-flight 的 rerun 收敛成单一路径：要么只保留 run 内 version-change 重试，要么给这个 effect 加一个“already rerunning”保护，确保 `validateFn` 切换后最新校验最多启动一次。

originalix (approved):

PR #12330, opened 2026-07-07
fix: stabilize market subtitle tooltip layout ｜OK-56595

zy2000-star:
[OK-56595](https://onekeyhq.atlassian.net/browse/OK-56595)

----------

## Summary

Fixes a layout jitter issue in the extension wallet popup where Market stock subtitles could repeatedly switch between truncated and full text, for example `Silver E...` and `Silver ETF`.

## Changes

- Keep `SubtitleText` render structure stable on web/extension by always rendering through `LazyTooltip`.
- Use `disabled={!isTruncated}` to enable tooltip behavior only when the subtitle is actually truncated.
- Add `triggerAsChild` to avoid an extra tooltip trigger wrapper affecting layout width.
- Avoid redundant `isTruncated` state updates when the measured truncation state has not changed.
- Preserve the existing native behavior by keeping native on the plain text path.

[OK-56595]: https://onekeyhq.atlassian.net/browse/OK-56595?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/views/Market/components/PerpsBadges.tsx
```diff
@@ -134,16 +137,17 @@ const SubtitleText = memo(
...
 
-    // Only show a hover tooltip on desktop when the name is actually cut off.
-    if (platformEnv.isNative || !isTruncated) {
+    if (platformEnv.isNative) {
       return textElement;
     }
 
     return (
       <LazyTooltip
+        disabled={!isTruncated}
```
originalix, line 160: 当用户已经把这个副标题 hover 出 tooltip 后，如果列表宽度变化让 `isTruncated` 从 `true` 变成 `false`，这里现在只是把 `Trigger` 置成 `disabled`。

`Tooltip` 自己的 open state 仍然保存在内部 `isShow` 里，不会因为 `disabled` 变成 `true` 自动收起，所以浮层会继续留在页面上，直到下一次 hover 或 scroll 才关闭。

建议在 `isTruncated` 切回 `false` 时同步把 tooltip close 掉，而不是只禁用 trigger，这样截断状态和浮层状态才能保持一致。

originalix (approved):

PR #12343, opened 2026-07-08
feat: add attempt-level firmware update tracking(OK-57543)

Shoyo1024:
[OK-57543](https://onekeyhq.atlassian.net/browse/OK-57543)

----------

## Summary
- Add new analytics event `firmwareUpdateFailedAttempt` that reports every real firmware-update task failure (fired at the retry funnel, before the user retries)
- Add `retryCount` and `durationMs` optional properties to the existing `firmwareUpdateResult` event (both v1 UI flow and v2 background service flow)
- All properties are sliceable by `deviceType`, `transportType` (USB / BLE / bridge), update flow (v1/v2) and firmware versions

## Intent & Context
We need to measure, per released firmware, how many devices upgraded through the app and what the real failure rate is, split by transport (USB vs Bluetooth) and device type.

The existing `firmwareUpdateStarted` / `firmwareUpdateResult` events only capture the final outcome of a whole update workflow. Because `runUpdateTask` never rejects on task failure (it parks the workflow in a retry-wait state via `firmwareUpdateRetryAtom`), a device that fails 3 times and then succeeds on retry reports a single `success` event and zero failures — attempt-level failure rate was invisible, and retry recovery could not be measured.

## Design Decisions
- **Report failures at the single retry funnel** (`runUpdateTask` catch): v1, v2 and v3 update tasks all flow through `createRunTaskWithRetry` → `runUpdateTask`, so one hook covers all flows without touching each updating function.
- **Per-workflow tracking context on the service** (`updateWorkflowTracking`): only one update workflow runs at a time; the context (flow type, releaseResult, start time, failed-attempt counter) is reset at both workflow entry points (`startUpdateWorkflow` / `startUpdateWorkflowV2`).
- **User exit is not a failure**: `FirmwareUpdateExit` / `FirmwareUpdateTasksClear` are excluded from failed-attempt reporting to avoid polluting the failure rate with user cancellations.
- **Analytics can never break the update flow**: the reporting helper is fully wrapped in try/catch, mirroring the existing `completeUpdateWorkflow` / `failUpdateWorkflow` logging pattern.
- New fields on `firmwareUpdateResult` are optional and additive — existing Mixpanel dashboards and older app versions are unaffected (older clients simply won't send the new properties).

## Changes Detail
- `packages/shared/src/logger/scopes/update/scenes/firmware.ts`: new `firmwareUpdateFailedAttempt` event; `firmwareUpdateResult` gains optional `retryCount` / `durationMs`
[...]

PR #12345, opened 2026-07-08
feat: add attempt-level firmware update tracking(OK-57543)

Shoyo1024:
[OK-57543](https://onekeyhq.atlassian.net/browse/OK-57543)

----------

> Bundle release PR targeting `release/v6.4.0` — this change ships via hot update (JS-only, no native changes). Counterpart PR for `x`: #12343.

## Summary
- Add new analytics event `firmwareUpdateFailedAttempt` that reports every real firmware-update task failure (fired at the retry funnel, before the user retries)
- Add `retryCount` and `durationMs` optional properties to the existing `firmwareUpdateResult` event
- All properties are sliceable by `deviceType`, `transportType` (USB / BLE / bridge), update flow (v1/v2) and firmware versions

## Intent & Context
We need to measure, per released firmware, how many devices upgraded through the app and what the real failure rate is, split by transport (USB vs Bluetooth) and device type.

The existing `firmwareUpdateStarted` / `firmwareUpdateResult` events only capture the final outcome of a whole update workflow. Because `runUpdateTask` never rejects on task failure (it parks the workflow in a retry-wait state via `firmwareUpdateRetryAtom`), a device that fails 3 times and then succeeds on retry reports a single `success` event and zero failures — attempt-level failure rate was invisible, and retry recovery could not be measured.

## Design Decisions
- **Report failures at the single retry funnel** (`runUpdateTask` catch): all update tasks flow through `createRunTaskWithRetry` → `runUpdateTask`, so one hook covers v1 and v2 flows.
- **Per-workflow tracking context on the service** (`updateWorkflowTracking`): only one update workflow runs at a time; the context (flow type, releaseResult, start time, failed-attempt counter) is reset at both workflow entry points (`startUpdateWorkflow` / `startUpdateWorkflowV2`).
- **User exit is not a failure**: `FirmwareUpdateExit` / `FirmwareUpdateTasksClear` are excluded from failed-attempt reporting.
- **Analytics can never break the update flow**: the reporting helper is fully wrapped in try/catch.
- New fields on `firmwareUpdateResult` are optional and additive — existing Mixpanel dashboards and older app versions are unaffected.
- This branch's adaptation differs from the `x` counterpart (#12343): on `release/v6.4.0` both v1 and v2 results are reported from `FirmwareUpdateCheckList` (there is no `completeUpdateWorkflow`/`failUpdateWorkflow` service path here), so the UI fetches `retryCount`/`durationMs` via the new `getUpdateWorkflowTrackingInfo()` background method.

## Changes Detail
- `packages/shared/src/logger/scopes/update/scenes/firmware.ts`: new `firmwareUpdateFailedAttempt` event; `firmwareUpdateResult` gains optional `retryCount` / `durationMs`
- `packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts`: workflow tracking context + `getUpdateWorkflowTrackingInfo()` background method + failed-attempt reporting in the `runUpdateTask` catch
- `packages/kit/src/views/FirmwareUpdate/components/FirmwareUpdateCheckList.tsx`: result events (success/failure) now include the new fields

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: analytics-only, additive change; no control-flow change to the update workflow. The only new code on the failure path is wrapped in try/catch. Hot-update safe: pure JS/TS, no native module changes.

## Test plan
- [ ] `yarn tsc:only` — no new errors vs release/v6.4.0 base (same 6 pre-existing unrelated errors)
- [ ] `oxlint` on the 3 touched files — 0 warnings / 0 errors
- [ ] `jest packages/kit-bg/src/services/ServiceFirmwareUpdate packages/shared/src/logger` — all pass
- [ ] Manual: update firmware over USB and BLE with dev analytics enabled (`enableAnalyticsInDev`), verify `firmwareUpdateFailedAttempt` fires on each induced failure and `firmwareUpdateResult` carries `retryCount` / `durationMs`

[OK-57543]: https://onekeyhq.atlassian.net/browse/OK-57543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
+++ b/packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
@@ -32,6 +32,8 @@ import {
   appEventBus,
 } from '@onekeyhq/shared/src/eventBus/appEventBus';
 import { CoreSDKLoader } from '@onekeyhq/shared/src/hardware/instance';
+import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
+import { parseFirmwareVersions } from '@onekeyhq/shared/src/logger/scopes/update/scenes/firmware';
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
 import deviceUtils from '@onekeyhq/shared/src/utils/deviceUtils';
 import { equalsIgnoreCase } from '@onekeyhq/shared/src/utils/stringUtils';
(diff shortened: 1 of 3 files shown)


Shoyo1024:
Closing: wrong target. v6.5.0 is code-frozen and QA has no time to verify this change, so it will NOT ride the v6.4.0 hot update. Plan: after v6.5.0 ships and `release/v6.5.0` is created, this change will be resubmitted against `release/v6.5.0` for its first bundle (hot) update. The x counterpart #12343 stays open (on hold until the v6.5.0 tag is cut).

PR #12349, opened 2026-07-08
fix: prevent mono font fallback to serif

huhuanming:
## Summary
- Add a system monospace fallback stack for Tamagui mono fonts on web-like targets.
- Change GeistMono web font loading from `optional` to `swap` so loaded mono fonts replace the fallback.

## Intent & Context
The Send confirmation modal showed account addresses with an incorrect serif-looking font on the current branch, while v6.4.0 rendered the same screen correctly. The issue affected address text that uses `$monoMedium`, including the account address shown in the SignatureConfirm Send review UI.

## Root Cause
The SignatureConfirm address component already used `$monoMedium` in both v6.4.0 and the current branch, so the UI component was not the regression. In v6.4.0, the web font hook loaded `GeistMono-Regular` and `GeistMono-Medium` through `expo-font` and gated rendering until fonts were ready. The current branch moved web-like entries to CSS `@font-face` loading for cold-start performance and set GeistMono to `font-display: optional`. When Chromium/Electron rendered the modal before GeistMono was ready, the font could permanently fall back. Because the mono font family did not include a fallback stack, it could fall back to a serif/default font.

## Design Decisions
The fix keeps address text on the mono typography path instead of changing the Send confirmation component to body text. This preserves the intended address rendering while avoiding the v6.4.0 render-blocking font gate. Web-like targets now fall back to system monospace fonts first, then swap to GeistMono after it loads. Native targets keep the explicit React Native font family names.

## Changes Detail
- `packages/components/tamagui.config.ts`: add a web-only monospace fallback stack for `$monoRegular` and `$monoMedium`; keep native font names unchanged.
- `packages/components/src/hocs/Provider/web-fonts.css`: change GeistMono font faces from `font-display: optional` to `font-display: swap`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Web / Extension
- **Risk Areas**: Visual typography only. The fallback stack can slightly alter initial text metrics before GeistMono loads, but it stays monospace and avoids the incorrect serif fallback.

## Test plan
- [x] `yarn lint:staged`
- [x] `yarn tsc:staged`
- [x] `yarn jest --findRelatedTests packages/components/tamagui.config.ts --passWithNoTests`
- [ ] Open the desktop Send confirmation modal and confirm address text renders as monospace, not serif, before and after GeistMono loads.


packages/components/tamagui.config.ts
```diff
@@ -166,6 +166,14 @@ const basicFontVariants = {
 };
 
 const tamaguiWebFontFamily = webFontFamily;
+const monoFontFamilyFallback =
+  'ui-monospace, SFMono-Regular, "SF Mono", Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace';
+const monoRegularFontFamily = isTamaguiNative
+  ? 'GeistMono-Regular'
+  : `"GeistMono-Regular", ${monoFontFamilyFallback}`;
+const monoMediumFontFamily = isTamaguiNative
```
originalix, line 172: 当 `GeistMono-Medium.ttf` 比 regular 更慢返回，或者 Medium 这一个资源单独加载失败时，`$monoMedium` 现在会从 `"GeistMono-Medium"` 直接退到系统 monospace，因为这里的 fallback 栈没有把 `"GeistMono-Regular"` 放进来。

这样同一页里 `$monoRegular` 和 `$monoMedium` 会先显示成两套不同字形，地址、报价这类依赖等宽布局的内容也更容易出现首屏宽度跳动。建议把 web 端的顺序改成 `"GeistMono-Medium", "GeistMono-Regular", ...`，先复用已经可用的 Geist Mono regular，再退到系统等宽字体。

huhuanming: Fixed: `$monoMedium` now falls back to `"GeistMono-Regular"` before generic `monospace`, and the fallback stack is shorter to reduce web startup output size while keeping mono rendering stable.

originalix (approved):

PR #12386, opened 2026-07-09
feat: default tabular numerals app-wide + localized thousands separators

PatrickChoo:
## Summary

Make **tabular (equal-width) figures the app-wide default** for all UI text, and give quoted prices locale-aware thousands separators. This is the standard treatment for trading interfaces — where a common approach is a brand font whose digits are tabular by default, so numbers never reflow as they tick and columns always align. We reach the same result with the OpenType `tnum` feature that our font (Roobert) already carries, applied globally rather than font-frozen (Roobert is a licensed font).

**Hot-update scope**: JS-only — no native code, no dependency, no font-file changes. Branched from `release/v6.4.0` per the bundle release flow.

## How it works (two mechanisms, one per platform)

- **Native**: `SizableText` (the wrapper every text primitive routes through) defaults `fontVariant` to `TABULAR_NUMS`, translated into a real RN `style.fontVariant` (Tamagui silently drops the raw prop). Reference-stable, allocation-free on the hot path.
- **Web**: a single injected `body { font-variant-numeric: tabular-nums }` rule; CSS inheritance covers every text node — including ones that bypass the wrapper (raw `<Text>`, `Badge.Text`, portals). The wrapper therefore emits **no** inline style for the default on web, avoiding a redundant attribute on every node.

Both are opt-out: `fontVariant={PROPORTIONAL_NUMS}` restores proportional figures for a specific text.

Text that bypasses the wrapper on **native** (raw RN `<Text>` in the order book / chart labels, and `Badge.Text` for ticking countdowns) applies the feature explicitly via `getFontVariantStyle(TABULAR_NUMS)` / a `StyleSheet` entry.

## Also in this change

- **Localized thousands separators** on quoted prices (order book, ticker/oracle price, TP-SL, order confirm, positions/orders/history rows) via the existing locale-aware `formatLocalizedNumberString` — European `78.056,1` and Indian grouping work out of the box. Formatting happens at the producing memo, not in JSX.
- **Perps ticker bar (desktop)**: stat values `$headingXs` (600, uppercase label token) → `$bodySmMedium` (500); headline mark price 600 → 500 — the heading weights read too heavy for a dense tabular stat strip.
- The ~150 per-site `fontVariant={TABULAR_NUMS}` props from the earlier draft are **removed** — the global default makes them redundant. Net deletion.

## Verification
- `tsgo --noEmit`: 0 errors; `oxlint --type-aware --deny-warnings`: clean; `numberUtils.locale.test.ts`: 72/72.
- Live web probes on the running app: computed `font-variant-numeric: tabular-nums` confirmed via inheritance across Market home (1208 numeric nodes, 0 non-tabular), /defi (Earn+Borrow), perps ticker/order book; separators rendered.

## Follow-ups (not in this PR)
- Optional: base `Badge.Text` on the wrapped `SizableText` so all badge digits go tabular on native in one place (currently only the ticking-countdown badges are patched).

### DIFF
--- a/development/spellCheckerSkipWords.txt
+++ b/development/spellCheckerSkipWords.txt
@@ -1492,3 +1492,6 @@ retryably
 
 # @mysten/sui 2.x package name
 mysten
+
+# OpenType tabular-figures feature tag
+tnum
--- a/packages/components/src/composite/SegmentSlider/index.tsx
+++ b/packages/components/src/composite/SegmentSlider/index.tsx
@@ -586,6 +586,7 @@ function SegmentSliderComponent({
       fontSize: 12,
       lineHeight: '16px',
       fontWeight: 600,
+      fontVariantNumeric: 'tabular-nums',
       opacity: 0,
       transition: 'opacity 150ms ease',
       pointerEvents: 'none',
--- a/packages/components/src/content/Badge/index.tsx
+++ b/packages/components/src/content/Badge/index.tsx
@@ -8,8 +8,10 @@ import {
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
 
 import { XStack } from '../../primitives/Stack';
+import { TABULAR_NUMS, getFontVariantStyle } from '../../utils/tabularNums';
 
 import type { IXStackProps } from '../../primitives';
+import type { GetProps } from '@onekeyhq/components/src/shared/tamagui';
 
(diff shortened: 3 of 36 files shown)


PatrickChoo:
Superseded by a version targeting `x` (regular monthly release) instead of the hot-update flow.

PR #12400, opened 2026-07-09
fix: reduce Home refresh trigger distance

huhuanming:
## Summary
- Reduce the pull distance required to trigger Home pull-to-refresh on iOS.
- Keep the spinner positioned below the immersive glass header after refresh starts.
- Simplify the Home refresh offset workaround by removing the deferred idle offset state.

## Intent & Context
This is a follow-up to #12399. The spinner became visible below the iOS immersive glass header, but the user still had to pull too far before the spinner started refreshing. The goal is to keep the iOS-only visual placement fix without making the native pull threshold feel heavy.

## Root Cause
On iOS Fabric, `progressViewOffset` is applied by changing the `UIRefreshControl` bounds origin. That moves the spinner down visually, but iOS also effectively counts that offset as part of the pull distance needed to trigger refresh. Keeping Home's header offset active while idle made the user pull roughly the header height more than before.

## Design Decisions
- Keep `progressViewOffset` at the default value while idle so the native trigger distance remains short.
- Apply the Home header offset only while `refreshing` is true, so the active spinner still appears below the glass/search header.
- Keep the behavior gated to iOS through the existing `platformEnv.isNativeIOS` check.
- Avoid a native patch or custom refresh overlay for this follow-up; those would be heavier and riskier than changing when the offset is applied.

## Changes Detail
- `PullToRefresh`: removes the deferred after-mount offset state and applies the Home context offset only during the refreshing state.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile iOS
- **Risk Areas**: The pre-trigger spinner may remain less visible under the glass header, but the active refreshing spinner is still positioned below the header and the pull threshold is restored.

## Test plan
- [x] Ran targeted `oxlint` for the touched Home refresh file
- [x] Ran `git diff --check`
- [ ] Verify on iOS simulator/device that Home refresh triggers with a shorter pull and the active spinner appears below the search header

### DIFF
--- a/packages/kit/src/views/Home/components/PullToRefresh/index.tsx
+++ b/packages/kit/src/views/Home/components/PullToRefresh/index.tsx
@@ -3,7 +3,6 @@ import {
   memo,
   useCallback,
   useContext,
-  useEffect,
   useMemo,
   useState,
 } from 'react';
@@ -60,28 +59,6 @@ function BasePullToRefresh({ onRefresh, ...props }: IPullToRefreshProps) {
     (props.progressViewOffset === undefined ||
       props.progressViewOffset === 0) &&
     progressViewOffsetFromContext !== undefined;
-  const [
-    deferredContextProgressViewOffset,
-    setDeferredContextProgressViewOffset,
-  ] = useState<number | undefined>();
-
-  useEffect(() => {
-    if (!shouldUseContextProgressViewOffset) {
-      setDeferredContextProgressViewOffset(undefined);
-      return;
-    }
-
-    // Fabric iOS ignores the initial progressViewOffset because the deferred
-    // initial props are compared against themselves. Apply Home's offset after
-    // mount so native sees a real 0 -> offset update before refresh starts.
-    setDeferredContextProgressViewOffset(undefined);
-    const frameId = requestAnimationFrame(() => {
-      setDeferredContextProgressViewOffset(progressViewOffsetFromContext);
-    });
-    return () => {
-      cancelAnimationFrame(frameId);
-    };
-  }, [progressViewOffsetFromContext, shouldUseContextProgressViewOffset]);
 
   const handleRefresh = useCallback(() => {
     onRefresh?.();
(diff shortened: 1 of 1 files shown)


packages/kit/src/views/Home/components/PullToRefresh/index.tsx
```diff
@@ -92,10 +69,11 @@ function BasePullToRefresh({ onRefresh, ...props }: IPullToRefreshProps) {
     defaultLogger.account.wallet.walletPullToRefresh();
   }, [onRefresh]);
 
+  // Keep the idle offset at zero because iOS also counts progressViewOffset
+  // toward the native pull distance required to trigger UIRefreshControl.
   const progressViewOffset =
-    shouldUseContextProgressViewOffset &&
-    deferredContextProgressViewOffset !== undefined
-      ? deferredContextProgressViewOffset
+    shouldUseContextProgressViewOffset && refreshing
```
sidmorizon, line 75: # [🟡 P2] offset 与 refreshing 耦合，可能重现 #12399 的 Fabric 定位问题

## 问题

被删除的 `useEffect` 原注释明确指出：Fabric iOS 会忽略初始 `progressViewOffset`，必须在**刷新开始之前**用一次独立的 `0 -> offset` 更新让 native 感知（`requestAnimationFrame` 就是为此隔离一帧）。

本次改为 `shouldUseContextProgressViewOffset && refreshing ? progressViewOffsetFromContext : props.progressViewOffset`。而 `handleRefresh` 里 `setRefreshing(true)` 会在**同一次 React commit** 中把 `progressViewOffset` 从 `0/undefined` 切到 `progressViewOffsetFromContext`。也就是 offset 变化与 `refreshing` 翻转被合并进了一次 batched props 更新——正是 #12399 认定 Fabric 处理不可靠的场景。

结合 `HomePageView.tsx` 的 provider，该 offset 值为 `tabPageHeight`（整段 header 高度，非几像素），下拉阶段 offset=`0`、刷新阶段 offset=`tabPageHeight`，二者跨度很大。

## 影响

在 iOS Fabric 上有两种可观察后果：其一，若 native 未在该 batched 更新中应用新的 `progressViewOffset`，刷新期间转圈会停留在偏移 `0`、被 sticky header 遮住，等于重现 #12399 修复的症状；其二，即便生效，转圈位置也会在刷新触发瞬间从 `0` 跳到 `tabPageHeight`，产生明显跳变/闪动。

## 建议

请在 iOS Fabric 真机上确认刷新期间转圈确实位于 header 下方且无跳变。若出现上述任一现象，建议把 offset 更新与 `refreshing` 翻转解耦——保留一个独立于 `refreshing` 的 offset 状态，在触发刷新前（或用 rAF）先单独提交一次 `0 -> offset`，再置 `refreshing=true`，以维持 #12399 依赖的“刷新前的独立 0 -> offset 更新”。

[Claude session ↗](https://claude.ai/code/session_01MVN1NH6hGCTmjb1mW7kupV)

---

huhuanming:
Closing this follow-up because applying the Home refresh offset only after  shortens the trigger distance but leaves the pre-trigger pull indicator invisible and can cause a visible jump. We need a different approach that keeps the pull indicator visible while preserving a reasonable trigger distance.

PR #12402, opened 2026-07-10
fix: decouple iOS refresh indicator offset

huhuanming:
## Summary
- Add a React Native iOS Fabric patch for `progressViewVisualOffset`, a visual-only refresh indicator offset.
- Keep native `progressViewOffset` semantics unchanged so pull distance is not increased by Home's glass header offset.
- Update Home `PullToRefresh` to use the visual-only offset and remove the JS deferred `progressViewOffset` workaround.

## Intent & Context
This is a follow-up to the Home immersive glass header pull-to-refresh investigation. The previous Home fix made the spinner visible below the glass/search header by setting `progressViewOffset`, but iOS counts that offset toward the native pull distance. The result was a spinner that could appear in the right place only after a very long pull. A later JS-only attempt to apply the offset after `refreshing=true` shortened the threshold, but left the pre-trigger pull indicator invisible and could introduce a visible jump, so that PR was closed.

## Root Cause
In React Native iOS Fabric, `progressViewOffset` is applied by changing `UIRefreshControl.bounds.origin.y`. That moves the spinner visually, but it also changes how far the user must pull before `UIRefreshControl` reaches the refresh threshold. Home needs a visual offset for the glass header while keeping the native trigger distance close to default.

## Design Decisions
- Add a new Fabric prop, `progressViewVisualOffset`, instead of changing existing `progressViewOffset` semantics globally.
- Implement the visual offset with `CGAffineTransformMakeTranslation` on the `UIRefreshControl`, which moves rendering without changing bounds-based pull distance.
- Keep the prop iOS-only in Home usage through the existing `platformEnv.isNativeIOS` guard.
- Fix the Fabric deferred initial props comparison to use the actual `oldProps`, so the visual offset applies on initial layout without a JS `requestAnimationFrame` workaround.

## Changes Detail
- `patches/react-native+0.81.5.patch`: adds `progressViewVisualOffset` to RN RefreshControl Flow/TS/codegen props and iOS Fabric native handling.
- `PullToRefresh`: passes Home's header offset as `progressViewVisualOffset`, removes deferred `progressViewOffset` state, and leaves `progressViewOffset` untouched for native pull distance.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile iOS Fabric
- **Risk Areas**: React Native patch-package maintenance, iOS refresh indicator transform behavior, and interaction with existing RN patches.

## Test plan
- [x] Verified `patches/react-native+0.81.5.patch` applies cleanly to pristine `react-native@0.81.5`
- [x] Verified patch contains no build artifacts
- [x] Ran `git diff --check`
- [x] Ran targeted `oxlint` for `PullToRefresh/index.tsx`
- [ ] Rebuild iOS app and verify Home pull-to-refresh shows the indicator during pull without requiring the longer `progressViewOffset` distance

### DIFF
--- a/patches/react-native+0.81.5.patch
+++ b/patches/react-native+0.81.5.patch
@@ -1,3 +1,37 @@
+diff --git a/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.d.ts b/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.d.ts
+index 81eba74..41d940b 100644
+--- a/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.d.ts
++++ b/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.d.ts
+@@ -69,6 +69,12 @@ export interface RefreshControlProps
+    * Progress view top offset
+    */
+   progressViewOffset?: number | undefined;
++
++  /**
++   * OneKey patch: visual-only progress view offset that does not affect the
++   * native pull distance required to trigger refresh.
++   */
++  progressViewVisualOffset?: number | undefined;
+ }
+
+ /**
+diff --git a/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.js b/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.js
+index 4c6947a..879c0bc 100644
+--- a/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.js
++++ b/node_modules/react-native/Libraries/Components/RefreshControl/RefreshControl.js
+@@ -70,6 +70,12 @@ type RefreshControlBaseProps = $ReadOnly<{
+    * Progress view top offset
+    */
+   progressViewOffset?: ?number,
++
++  /**
++   * OneKey patch: visual-only progress view offset that does not affect the
++   * native pull distance required to trigger refresh.
++   */
++  progressViewVisualOffset?: ?number,
+ }>;
+
+ export type RefreshControlProps = $ReadOnly<{
 diff --git a/node_modules/react-native/Libraries/Components/TextInput/RCTTextInputViewConfig.js b/node_modules/react-native/Libraries/Components/TextInput/RCTTextInputViewConfig.js
 index 05bddcb..5211253 100644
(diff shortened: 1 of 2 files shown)


patches/react-native+0.81.5.patch
```diff
@@ -507,16 +561,55 @@ index bec2ac5..4cab4b5 100644
...
  + (void)supportLegacyViewManagersWithPrefix:(NSString *)prefix
+diff --git a/node_modules/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTPullToRefreshViewComponentView.mm b/node_modules/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTPullToRefreshViewComponentView.mm
+index 0d231bc..3adb26b 100644
+--- a/node_modules/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTPullToRefreshViewComponentView.mm
++++ b/node_modules/react-native/React/Fabric/Mounting/ComponentViews/ScrollView/RCTPullToRefreshViewComponentView.mm
+@@ -78,6 +78,8 @@ - (void)updateProps:(const Props::Shared &)props oldProps:(const Props::Shared &
+
+   const auto &oldConcreteProps = static_cast<const PullToRefreshViewProps &>(*_props);
+   const auto &newConcreteProps = static_cast<const PullToRefreshViewProps &>(*props);
++  const auto &oldConcretePropsForVisualOffset =
```
originalix, line 574: 当 `RefreshControl` 首次挂载且 `progressViewVisualOffset` 初始就是非 0 时，`oldProps` 这里仍然可能是 `nil`。当前 fallback 到 `oldConcreteProps` 后，会沿用之前 `progressViewOffset` 首帧“新旧 props 比较成自己”的那条路径，`_updateProgressViewVisualOffset()` 就不会执行。

JS 侧这次已经删掉了 mount 后补一次 offset 的 workaround，所以这个分支一旦跳过，Home 首次进入时指示器会继续停在默认位置，只有后续 props 再变化时才会下移。

这里建议在 `oldProps == nil` 时直接用 `newConcreteProps.progressViewVisualOffset` 初始化一次 `_refreshControl`，不要再依赖首帧的差异比较。

PR #12407, opened 2026-07-10
fix: perps bug batch (OK-57733, OK-57450, OK-57781, OK-57785, OK-57786)

Minnzen:
[OK-57733](https://onekeyhq.atlassian.net/browse/OK-57733) [OK-57450](https://onekeyhq.atlassian.net/browse/OK-57450) [OK-57781](https://onekeyhq.atlassian.net/browse/OK-57781) [OK-57785](https://onekeyhq.atlassian.net/browse/OK-57785) [OK-57786](https://onekeyhq.atlassian.net/browse/OK-57786)

----------

## Summary

- Batch of five independent Perps bug fixes, one commit per Jira ticket.
- Restores three UI regressions (holdings actions, trade history share action, available balance label), fixes spot favorites search, and removes a blocking network wait from home portfolio loading.
- Adds targeted Jest coverage for the three fixes that introduced extractable logic.

## Intent & Context

These are five separate Perps tickets triaged together rather than one feature. Three of them (OK-57786, OK-57785, OK-57781) are regressions where previously-working UI degraded; one (OK-57733) is a search bug users hit in the token selector; one (OK-57450) is a perceived-performance complaint about the Perps home screen stalling on account switch.

They are batched into a single PR because each diff is small and they touch disjoint files, so reviewing them together is cheaper than five PRs. Each commit stands alone and can be reverted independently.

## Root Cause

- **OK-57786 — holdings actions**: `BalanceRow` routed its copy/open/share buttons through `DesktopActionIconButton`, whose wrapper sizing swallowed the icons in the holdings table.
- **OK-57785 — trade history share action**: an invisible shared wrapper made the Close PnL header align to the action slot instead of the PnL value, so header and values drifted apart.
- **OK-57781 — available balance label**: the trading forms rendered the `Avbl` abbreviation while the full localized available-balance key already existed and carried the correct copy.
- **OK-57733 — spot favorites search**: the favorites tab filtered only by raw symbol, so spot pairs (which display as a base/quote pair name) and alias matches never matched the query.
- **OK-57450 — home portfolio loading**: `_getPortfolioAbstractionMode` ran as a second serial network wave after the clearinghouse fetches, and any non-fresh cached snapshot forced a full blocking refetch on every account switch.

## Design Decisions

- **OK-57450 — stale-while-revalidate over a longer freshness window**: rather than widening `PERPS_HL_PORTFOLIO_ACTIVE_MAX_AGE_MS` (which would serve stale data to pollers too), a stale snapshot up to `PERPS_HL_PORTFOLIO_STALE_SERVE_MAX_AGE_MS` (24h) renders immediately while a background refresh runs. Degraded snapshots are explicitly excluded from stale serving because they hold incomplete dex aggregates. The abstraction-mode request only depends on the address, so it is kicked off alongside the clearinghouse fetches instead of after them.
- **OK-57781 — reuse the existing translation key, do not add one**: the full available-balance key already carries correct localized copy in every locale. Adding or editing `Avbl` translations was rejected; generated locale files remain untouched.
- **OK-57785 — fixed 32px slot**: the share action lives in a stable 32px slot and the header reserves the same slot. Widening the fixed pane was rejected because it leaves the invisible wrapper in place and aligns the header to the action slot rather than the PnL value.
- **OK-57733 — extract the filter into a pure util**: `filterPerpTokenSelectorSpotItemsBySearch` and the display-name helpers live in `tokenSelectorTabs.ts` / `perpsUtils.ts` so both `PerpTokenSelector` and `MoblieTokenSelector` share one implementation and the matching logic is unit-testable without mounting a selector.

## Changes Detail

- `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts` — parallelize abstraction-mode fetch; add stale-while-revalidate path to the portfolio snapshot cache.
- `packages/shared/src/consts/perpCache.ts` — new `PERPS_HL_PORTFOLIO_STALE_SERVE_MAX_AGE_MS` (24h).
- `packages/kit/src/views/Perp/utils/tokenSelectorTabs.ts` (+ test) — spot-aware, alias-aware search filter shared by both token selectors.
- `packages/shared/src/utils/perpsUtils.ts` (+ test) — spot pair display-name helpers used by the filter.
- `.../TokenSelector/PerpTokenSelector.tsx`, `.../TokenSelector/MoblieTokenSelector.tsx` — consume the shared filter.
- `.../OrderInfoPanel/Components/BalanceRow.tsx` — replace `DesktopActionIconButton` with directly-sized `IconButton`s.
- `.../OrderInfoPanel/Components/TradesHistoryShareAction.tsx` (+ test), `TradesHistoryRow.tsx`, `List/CommonTableListView.tsx`, `List/PerpTradesHistoryList.tsx` — stable 32px share-action slot, reserved in the header.
- `.../TradingPanel/panels/LimitOrderForm.tsx`, `PerpTradingForm.tsx` (+ `availableBalanceLabel.test.ts`) — use the full localized available-balance key.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Web / Mobile (Perps surfaces only)
- **Risk Areas**:
  - `ServiceHyperliquid` stale-while-revalidate is the riskiest change: it alters what the Perps home renders on account switch and fires a background refresh whose failure path deletes the memo entry. Runtime scope is background (bg) JS; the snapshot is persisted via `simpleDb.perp` and read back by main-JS through the background API, so a stale snapshot can be observed by main before the refresh lands. This is intentional, but it means a user switching accounts can briefly see the previous snapshot's values.
  - The 24h stale ceiling is a deliberate trade-off: beyond it the UI shows loading rather than very old balances.
  - The three UI fixes are visual and low risk, but `BalanceRow` and `CommonTableListView` are shared by multiple Perps tables, so header/value alignment should be checked on more than the trade history list.

## Test plan

- [ ] Targeted Jest suites pass: `tokenSelectorTabs.test.ts`, `perpsUtils.test.ts`, `TradesHistoryShareAction.test.tsx`, `availableBalanceLabel.test.ts`.
- [ ] Perps home: switch between two accounts with existing snapshots — the portfolio renders immediately from cache and updates within one poller tick, with no loading flash.
- [ ] Perps home: switch to an account with no cached snapshot — loading state shows, then data.
- [ ] Perps home: with a snapshot older than 24h, loading shows instead of stale balances.
- [ ] Token selector: search a spot pair by base symbol, by pair display name, and by a known alias — favorites tab filters correctly on both desktop and mobile selectors.
- [ ] Holdings table: copy / open-explorer / share icon buttons are visible, correctly sized, and hit-testable.
- [ ] Trade history: Close PnL header aligns with the PnL values; share action stays in its slot across row widths.
- [ ] Trading forms (standard + chart limit order): the available-balance label shows the full localized string, verified after a language switch.

[OK-57733]: https://onekeyhq.atlassian.net/browse/OK-57733?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57450]: https://onekeyhq.atlassian.net/browse/OK-57450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57781]: https://onekeyhq.atlassian.net/browse/OK-57781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57785]: https://onekeyhq.atlassian.net/browse/OK-57785?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57786]: https://onekeyhq.atlassian.net/browse/OK-57786?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1157,8 +1163,28 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+      if (
+        !cached.isDegraded &&
+        Date.now() - cached.fetchedAt <=
+          PERPS_HL_PORTFOLIO_STALE_SERVE_MAX_AGE_MS
+      ) {
+        // Stale-while-revalidate: account switches render the last snapshot
+        // immediately instead of blocking on the network; pollers pick up the
+        // refreshed snapshot on their next tick. Degraded snapshots hold
+        // incomplete dex aggregates, so they never serve stale.
+        void this.fetchHyperliquidPortfolioByAddress(normalized)
```
originalix, line 1179: 当本地 portfolio snapshot 已经过了 freshness，但还在 `PERPS_HL_PORTFOLIO_STALE_SERVE_MAX_AGE_MS` 窗口内时，这里会直接起一个后台 refresh，并且无条件把 `fresh` 写回 `simpleDb`。

问题是这条分支没有继承 `skipCacheWriteIfEmpty` 的语义。当前首页充值确认流程会用 `force: true, skipCacheWriteIfEmpty: true` 拉网络快照来避免把暂时为空的结果覆盖缓存；如果随后有一次普通读取走到这里，5 秒 memo 窗口里的空结果仍然可能被这条后台 refresh 路径复用并落库，之前的非空 snapshot 就会被错误冲成空状态。

这样首页/托盘下一轮读取就可能看到错误的 0 余额。这里的 revalidate 写回至少要和主路径一样跳过 `fresh.isEmpty`，或者显式隔离这类 force-only 结果，避免绕过现有的空快照保护。

originalix (approved):

PR #12416, opened 2026-07-10
fix: filter zero-fiat perps deposit tokens(OK-57824)

originalix:
[OK-57824](https://onekeyhq.atlassian.net/browse/OK-57824)

----------

## Summary
- Filter Perps deposit tokens to only include entries with a finite positive fiat value.
- Apply the same filtering to cold-cache data and per-network token maps.
- Add regression coverage for zero, missing, invalid, negative, and cached fiat values.

## Intent & Context
OK-57824 reported that the Perps deposit selector displayed wallet tokens without a fiat amount, including entries rendered as $0.00. The requested behavior is to hide tokens that have no fiat value while keeping funded tokens available for deposit.

## Root Cause
Wallet token responses were mapped into the Perps deposit list before checking fiat metadata. Tokens with missing, zero, invalid, or non-positive fiat values therefore remained visible. The cold-cache path returned previously persisted lists without applying the same validation.

## Design Decisions
- Use BigNumber validation and require a finite value greater than zero, matching the existing fiat-value sorting semantics.
- Filter before sorting so the visible order remains the wallet fiat-value order.
- Reuse the filter for both live responses and cold-cache data to prevent stale $0.00 entries during cache-first rendering.
- Preserve empty network buckets so supported-network UI behavior remains unchanged.

## Changes Detail
- packages/kit-bg/src/services/ServiceWebviewPerp/utils/depositTokenListUtils.ts: add positive-fiat filtering helpers and apply them to wallet response mapping.
- packages/kit-bg/src/services/ServiceWebviewPerp/ServiceWebviewPerp.ts: sanitize cold-cache token data before writing it to the atom.
- packages/kit-bg/src/services/ServiceWebviewPerp/utils/depositTokenListUtils.test.ts: cover live and cached filtering edge cases.
- packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx: add a stable selector test ID for UI verification.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile / Desktop
- **Risk Areas**: Perps deposit-token selection and cache-first rendering; no transaction, signing, or balance calculation logic is changed.

## Test plan
- [x] Targeted Jest suites: 3 suites, 23 tests passed.
- [x] Desktop self-test: zero-fiat account shows no deposit token entries.
- [x] iOS self-test: positive-fiat tokens remain visible and DAI with $0.00 is absent.
- [x] Android self-test: empty wallet deposit form shows no $0.00 token entry.
- [x] yarn agent:check --profile commit passed.

[OK-57824]: https://onekeyhq.atlassian.net/browse/OK-57824?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceWebviewPerp/utils/depositTokenListUtils.ts
+++ b/packages/kit-bg/src/services/ServiceWebviewPerp/utils/depositTokenListUtils.ts
@@ -69,6 +69,28 @@ function toOptionalString(value: string | number | undefined) {
   return String(value);
 }
 
+function hasPositiveFiatValue(token: IPerpsDepositToken) {
+  const fiatValue = new BigNumber(token.fiatValue ?? '');
+  return fiatValue.isFinite() && fiatValue.gt(0);
+}
+
+export function filterPerpsDepositTokensWithPositiveFiatValue(
+  tokens: IPerpsDepositToken[],
+) {
+  return tokens.filter(hasPositiveFiatValue);
+}
+
+export function filterPerpsDepositTokensByNetworkWithPositiveFiatValue(
+  tokensByNetwork: Record<string, IPerpsDepositToken[]>,
+) {
+  return Object.fromEntries(
+    Object.entries(tokensByNetwork).map(([networkId, tokens]) => [
+      networkId,
+      filterPerpsDepositTokensWithPositiveFiatValue(tokens),
+    ]),
+  );
+}
+
 function mapWalletTokenToPerpsDepositToken({
   token,
   fiat,
@@ -121,7 +143,9 @@ export function buildPerpsDepositTokensFromWalletTokenResponses({
     }
   }
 
-  return sortPerpsDepositTokensByFiatValue(tokens);
+  return sortPerpsDepositTokensByFiatValue(
+    filterPerpsDepositTokensWithPositiveFiatValue(tokens),
+  );
 }
 
 export function sortPerpsDepositTokensByFiatValue(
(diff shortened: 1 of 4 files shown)


weatherstar (approved):

packages/kit-bg/src/services/ServiceWebviewPerp/utils/depositTokenListUtils.ts
```diff
@@ -121,7 +143,9 @@ export function buildPerpsDepositTokensFromWalletTokenResponses({
     }
   }
 
-  return sortPerpsDepositTokensByFiatValue(tokens);
+  return sortPerpsDepositTokensByFiatValue(
```
originalix, line 146: 这里把钱包返回的可见 token 直接收紧成 `fiatValue > 0` 之后，只要 quote map 暂时缺项、key 不匹配，或者 fiat 值异常，这个 token 就会从充值列表里消失。

`fetchAccountTokens` 这一层已经用 `hideSmallBalanceTokens` / `hideRiskTokens` 做过可见性筛选了，再把 quote 可用性当成 eligibility gate，会把仍然可充值的资产误判成不可用；而且同一规则还会同步影响 cold cache，用户下次打开弹窗也会继续看不到这个 token。

建议这里保留钱包返回的可见 token，把 `fiatValue` 只用于排序和默认选中，不要因为缺少 fiat quote 就直接过滤掉 token。

PR #12342, opened 2026-07-08
fix: open dapp webview from search(OK-57292)

huhuanming:
[OK-57292](https://onekeyhq.atlassian.net/browse/OK-57292)

----------

## Summary
- Hydrate Discovery browser tab state before opening a DApp from search results on desktop and native.
- Prevent the browser data rebuild mount path from overwriting a freshly opened active DApp tab with an old stored snapshot.
- Wait for the desktop webview preload URL before rendering the initial `<webview>` so the DApp view mounts after async preload resolution.

## Intent & Context
This fixes OK-57292. In Electron, clicking the Uniswap DApp search result could create a browser tab entry without actually focusing and rendering the DApp webview. The requested workflow was to reproduce the issue in Electron, fix it, remove temporary investigation logs, and verify the real webview render instead of only checking that a tab label exists.

The investigation also checked the native path. Native production has two JS runtimes (`main` UI and `background`), with isolated JS heaps. Native storage resources such as SimpleDb/MMKV are shared at the native layer, but the browser tab objects deserialized into JS are per-runtime copies. The bug fixed here is in the UI/browser Jotai hydration ordering and applies to desktop and native browser actions, not to shared native memory.

## Root Cause
Desktop does not have a separate background JS runtime, so the bad state was not caused by desktop reading a bg runtime copy. The failure was caused by browser UI state hydration ordering:

- Search-result DApp open could wait on browser-data readiness while no active hydrator existed in cold or stale waiter cases.
- On desktop, switching to `MultiTabBrowser` could mount `HandleRebuildBrowserData`, which rebuilt tabs from an old SimpleDb snapshot and overwrote the newly opened active tab.
- The saved active tab restore path could therefore restore the old `savedActiveTabId` after the new DApp tab had already been created.
- DesktopWebView used an async preload URL populated outside React state; the first render could return `null` and then never re-render when the preload URL resolved.

## Design Decisions
- Reuse the existing `backgroundApiProxy.simpleDb.browserTabs.getRawData()` and `buildWebTabs(... isInitFromStorage)` hydration path instead of adding a second storage reader.
- Add an explicit browser data readiness waiter that can self-hydrate stale/cold states before `gotoSite` or DApp search result open proceeds.
- Mark browser data ready before switching to the desktop browser and guard `HandleRebuildBrowserData` once ready, so a late mount cannot overwrite the current in-memory tab state.
- Clear the saved active tab before switching to `MultiTabBrowser`, and create/activate the destination tab first on desktop.
[...]

PR #12431, opened 2026-07-10
fix: register desktop webview load listeners after deferred preload(OK-57862)

Minnzen:
[OK-57862](https://onekeyhq.atlassian.net/browse/OK-57862)

----------

## Summary
- Register desktop webview load-event listeners once the `<webview>` node actually mounts, restoring app→webview injected-message delivery (Perps chart `SYMBOL_CHANGE`, chart lines sync, `FORCE_RECOVER_WS`, navigation guard).
- Read dom-ready state through a ref in the imperative-handle wrapper so a parent holding an older wrapper still delivers instead of queueing on a stale snapshot.
- Add jest regression tests covering both failure modes.

## Intent & Context
QA reported on the desktop 6.5.0 test build (OK-57862): on the Perps page, switching to a spot pair did not switch the TradingView chart, and right-click → trade inside the chart opened no order dialog. The production app was unaffected.

Log analysis showed the app-side state switched correctly, but the chart never resolved any new symbol after an in-page switch — perp→perp switches were equally dead, spot just made it visible. Every fresh webview load resolved its URL symbol fine, which is why the page looked normal until the first switch.

## Root Cause
`d7b9d95a6d` (OK-57292, #12342) deferred the preload URL fetch from module-load time to a mount effect. The first `DesktopWebView` of a renderer session now renders `null` first and only re-renders with the `<webview>` once the preload URL resolves. The load-event listener effect ran on the first commit (webview node absent, early return) and never re-ran because its dependency array contained only stable callbacks.

As a result `dom-ready` was never handled, `isDomReady` stayed `false` forever, and every `sendMessageViaInjectedScript` call was pushed into `pendingScriptsRef` and never flushed. The chart→webview receive direction (`ipc-message` effect, which does depend on `isWebviewReady`) kept working, masking the failure: request/response flows like price-scale and marks worked, only app-initiated pushes were dead.

The right-click trade symptom is downstream: the chart was stuck on the old symbol, so the order-intent message carried a stale symbol and failed the `payload.symbol === latestSymbolRef.current` check in `TradingViewPerpsV2` silently.

Also broken for the first webview of each session: the `onShouldStartLoadWithRequest` navigation guard (wired via `did-start-navigation` / `did-redirect-navigation`), load-failure error views and timeouts, and `onLoadEnd`/`onDomReady` callbacks. The dapp browser is equally affected when it is the first webview opened after launch.

The chart repo was ruled out by injecting `SYMBOL_CHANGE` messages directly into the deployed chart bundle (both perp and spot symbols switch correctly).

## Design Decisions
- Minimal-surface fix: add `isWebviewReady` to the listener effect deps instead of restructuring the effect; the effect already early-returns while the node is absent and its cleanup correctly re-registers.
- Introduced `isDomReadyRef` and switched the wrapper's send path to read it, then reduced the imperative-handle deps to `[jsBridgeHost]`. This keeps the wrapper stable across renders and eliminates the whole class of stale-closure delivery bugs (parents such as `InpageProviderWebView.desktop` cache the wrapper and only refresh it on their own re-renders).
- `flushPendingScripts` still keys off `isDomReady` state so queued messages flush on the dom-ready flip.

## Changes Detail
- `DesktopWebView.tsx`: listener effect re-runs on `isWebviewReady`; `updateIsDomReady` keeps a ref in sync with state; wrapper send reads the ref; wrapper handle stable per `jsBridgeHost`.
- `desktopWebViewMessageDelivery.test.tsx`: full-chain tests through `InpageProviderWebView.desktop` with production-equivalent `useWebViewBridge` semantics. Cases: deferred preload (regression), pre-dom-ready queue flush, dom-ready arriving after the last parent render (stale wrapper). Cases 1 and 3 fail without the fix.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**: listener effect now re-registers when `isWebviewReady` flips (cleanup path exercised more often); imperative-handle wrapper identity no longer changes on dom-ready, so consumers relying on wrapper recreation timing would be affected (none found).

## Test plan
- [x] `yarn jest packages/kit/src/components/WebView/desktopWebViewMessageDelivery.test.tsx` — passes with fix, cases 1 & 3 fail without it
- [x] `yarn jest packages/kit/src/components/WebView/DesktopWebView.test.tsx` — existing preload-retry test still passes
- [x] Manual desktop verify: launch app → Perps → switch perp↔spot repeatedly, chart follows; right-click → trade opens the limit-order dialog
- [ ] Dapp browser smoke: first webview after app launch loads normally and blocked URLs are still stopped

[OK-57862]: https://onekeyhq.atlassian.net/browse/OK-57862?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/components/WebView/DesktopWebView.tsx
+++ b/packages/kit/src/components/WebView/DesktopWebView.tsx
@@ -444,7 +444,9 @@ const DesktopWebView = forwardRef(
       ref as Ref<unknown>,
       (): IWebViewWrapperRef => {
         const wrapper = {
-          innerRef: webviewRef.current,
+          // deferred preload mounts the node after the first create; the
+          // isWebviewReady dep re-snapshots innerRef once it exists.
+          innerRef: isWebviewReady ? webviewRef.current : null,
           jsBridge: jsBridgeHost,
           reload: () => {
             webviewRef.current?.reload();
@@ -483,7 +485,7 @@ const DesktopWebView = forwardRef(
       },
       // dom-ready is read via isDomReadyRef so a parent holding an old
       // wrapper still delivers.
-      [jsBridgeHost],
+      [isWebviewReady, jsBridgeHost],
     );
 
     const initWebviewByRef = useCallback(
--- a/packages/kit/src/components/WebView/DesktopWebView.tsx
+++ b/packages/kit/src/components/WebView/DesktopWebView.tsx
@@ -129,6 +129,13 @@ const DesktopWebView = forwardRef(
   ) => {
     const [isWebviewReady, setIsWebviewReady] = useState(false);
     const [isDomReady, setIsDomReady] = useState(false);
+    // Parents hold wrapper closures across renders, so dom-ready must be read
+    // from a ref, not a render snapshot.
+    const isDomReadyRef = useRef(false);
+    const updateIsDomReady = useCallback((value: boolean) => {
+      isDomReadyRef.current = value;
+      setIsDomReady(value);
+    }, []);
     const webviewRef = useRef<IElectronWebView | null>(null);
     const pendingScriptsRef = useRef<string[]>([]);
     const [devToolsAtLeft, setDevToolsAtLeft] = useState(false);
@@ -293,7 +300,7 @@ const DesktopWebView = forwardRef(
             lastMainFrameLoadErrorRef.current = undefined;
             setDesktopLoadError(false);
             setDesktopLoadErrorCode(undefined);
-            setIsDomReady(false);
+            updateIsDomReady(false);
             startLoadTimeout();
           }
           checkGoogleOauth(url);
@@ -351,7 +358,7 @@ const DesktopWebView = forwardRef(
         webview.addEventListener('page-favicon-updated', onPageFaviconUpdated);
         webview.addEventListener('new-window', onNewWindow);
         const handleDomReady = (event: Event) => {
-          setIsDomReady(true);
+          updateIsDomReady(true);
           onDomReady?.(event);
         };
 
@@ -386,8 +393,12 @@ const DesktopWebView = forwardRef(
         console.error(error);
       }
     }, [
+      // the first run can precede <webview> mount when the preload URL
+      // resolves async, leaving every load listener unregistered.
+      isWebviewReady,
       clearLoadTimeout,
       startLoadTimeout,
+      updateIsDomReady,
       onDidFailLoad,
       onDidFinishLoad,
       onDidStartLoading,
@@ -445,7 +456,7 @@ const DesktopWebView = forwardRef(
           },
           sendMessageViaInjectedScript: (message: unknown) => {
             const script = createMessageInjectedScript(message);
-            if (!isDomReady || !webviewRef.current) {
+            if (!isDomReadyRef.current || !webviewRef.current) {
               pendingScriptsRef.current.push(script);
               if (pendingScriptsRef.current.length > 50) {
                 console.warn(
@@ -470,14 +481,19 @@ const DesktopWebView = forwardRef(
         jsBridgeHost.webviewWrapper = wrapper;
         return wrapper as IWebViewRef;
       },
-      [isDomReady, jsBridgeHost],
+      // dom-ready is read via isDomReadyRef so a parent holding an old
+      // wrapper still delivers.
+      [jsBridgeHost],
     );
 
-    const initWebviewByRef = useCallback(($ref: any) => {
-      webviewRef.current = $ref;
-      setIsDomReady(false);
-      setIsWebviewReady(Boolean($ref));
-    }, []);
+    const initWebviewByRef = useCallback(
+      ($ref: any) => {
+        webviewRef.current = $ref;
+        updateIsDomReady(false);
+        setIsWebviewReady(Boolean($ref));
+      },
+      [updateIsDomReady],
+    );
 
     useEffect(() => {
       const webview = webviewRef.current;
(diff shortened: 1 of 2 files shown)


originalix (dismissed):

packages/kit/src/components/WebView/DesktopWebView.tsx
```diff
@@ -470,14 +481,19 @@ const DesktopWebView = forwardRef(
         jsBridgeHost.webviewWrapper = wrapper;
         return wrapper as IWebViewRef;
       },
-      [isDomReady, jsBridgeHost],
+      // dom-ready is read via isDomReadyRef so a parent holding an old
+      // wrapper still delivers.
+      [jsBridgeHost],
```
sidmorizon: # [🟠 P1] 移除 isDomReady 依赖后，首个 webview 的 wrapper.innerRef 永远为 null

## 问题

`useImperativeHandle` 依赖从 `[isDomReady, jsBridgeHost]` 缩减为 `[jsBridgeHost]`。`wrapper.innerRef` 是 `webviewRef.current` 的快照，只有 create 重新执行时才会刷新。而 `jsBridgeHost` 依赖 `[receiveHandler]`，浏览器路径下 `receiveHandler`/`customReceiveHandler` 均为稳定的 `useCallback`，所以 wrapper 只创建一次。

本 PR 修复的正是"会话首个 webview 先渲染 `null`"的场景：此时 `useImperativeHandle` 首次执行时 `webviewRef.current` 仍是 `null`，`innerRef` 被快照为 `null`；随后 webview 真正挂载（`initWebviewByRef` 设置 `webviewRef.current`）时，依赖 `[jsBridgeHost]` 未变，create 不再重跑，`innerRef` 永远停留在 `null`。改动前依赖里的 `isDomReady` 在 dom-ready 后翻转为 `true`，恰好触发 wrapper 重建并把 `innerRef` 刷新为已挂载节点，这条路径被本 PR 删除了。

## 影响

`WebContent.desktop.tsx` 的 `onWebViewRef` 用 `if (ref && ref.innerRef)` 作为守卫才写入 `webviewRefs[id]`。`innerRef` 恒为 `null` 时该守卫永不成立，会话首个标签页的 `webviewRefs[id]` 永不注册、`refReady` 永不置位。所有读取 `webviewRefs[id]?.innerRef` 的桌面逻辑（`loadURL`、`injectJavaScript`、`goBack/goForward`、`getNavStatusInfo`、内存压力、快捷键、DApp 通知、Bitrefill 注入等）对首个标签全部失效。第二个及以后的标签因 `preloadJsUrl` 已被模块级缓存、首帧即渲染 webview 而不受影响，故问题只影响冷启动首个 webview，隐蔽性高。新增测试只覆盖 `sendMessageViaInjectedScript`，未校验 `innerRef`，无法拦截此回归。

## 建议

保留 `isDomReadyRef` 的投递修复，同时让 wrapper 在 webview 节点可用后重建。最小改动是把 `isWebviewReady` 加回依赖：

```ts
[isWebviewReady, jsBridgeHost],
```

这样 `initWebviewByRef` 置位 `isWebviewReady` 后 create 会重跑一次，`innerRef` 刷新为已挂载节点；投递仍走 `isDomReadyRef`，不会退回旧的每次 dom-ready 重建。或将 `innerRef` 改为读取 `webviewRef.current` 的实时 getter。

[Claude session ↗](https://claude.ai/code/session_017TurEdrvVd4rqAPDA4gqxd)

---

Minnzen: Valid catch — fixed in f55daba070.

One nuance from re-testing: strictly speaking `innerRef` did not stay `null` forever. React appends the `ref` argument to `useImperativeHandle`'s effect deps, and `useWebViewBridge`'s `setWebViewRef` is a new function identity on every `InpageProviderWebView` render, so any parent re-render (guaranteed by the progress-state updates during page load) re-ran create and re-snapshotted `innerRef`. The full-chain regression test passed with `[jsBridgeHost]` alone for exactly this reason.

That said, relying on accidental ref-identity churn is the same class of fragility this PR removes, so the hardening is right: `isWebviewReady` is back in the deps and `innerRef` is now explicitly gated on it (`isWebviewReady ? webviewRef.current : null`), which also satisfies exhaustive-deps. Delivery still reads `isDomReadyRef`, so we don't regress to per-dom-ready wrapper rebuilds. Added an `innerRef` assertion to the deferred-mount test to pin the invariant `WebContent.desktop`'s registration guard depends on.

weatherstar (approved):

originalix (approved):

packages/kit/src/components/WebView/DesktopWebView.tsx
```diff
@@ -445,7 +458,7 @@ const DesktopWebView = forwardRef(
           },
           sendMessageViaInjectedScript: (message: unknown) => {
             const script = createMessageInjectedScript(message);
-            if (!isDomReady || !webviewRef.current) {
+            if (!isDomReadyRef.current || !webviewRef.current) {
```
originalix, line 461: 当页面在 `dom-ready` 前已经有脚本排队，而 `onDomReady` 回调里或紧接着又调用一次 `sendMessageViaInjectedScript` 时，这里会因为 `isDomReadyRef.current` 已经是 `true` 而直接执行新脚本，但更早入队的脚本还要等后面的 effect 才 flush。

这样消息顺序会被反转，先发的初始化/同步脚本反而晚于后发的业务消息执行，TradingView 或 bridge 这类依赖 FIFO 的协议链路会出现新的时序回归。

建议在进入直发分支前先同步 drain `pendingScriptsRef`，或者让 `dom-ready` 前后的消息统一走同一个 FIFO flush 路径，保证 injected script 仍然按入队顺序交付。

PR #12432, opened 2026-07-10
fix: restore Perps selector scrolling(OK-57864)

huhuanming:
[OK-57864](https://onekeyhq.atlassian.net/browse/OK-57864)

----------

## Summary
- Disable visible-content anchoring only for the native Perps mobile token selector.
- End the iOS initial-row snapshot phase when search interaction begins so clearing the query cannot restore a stale overlay or cold-start cache.

## Intent & Context
OK-57864 reports an iOS-only regression in the Perps asset selector: start from the full list, filter to a token such as SKHX, then delete all search text. The full dataset returns, but the upper rows appear frozen while only content at the bottom visibly scrolls.

## Root Cause
Two list behaviors overlap:

1. FlashList v2 preserves the currently visible item by default. When the filtered one-row dataset expands, the previous token can remain anchored near the bottom instead of exposing the beginning of the restored results.
2. The iOS cold-start snapshot was reset whenever `searchQuery` changed. Clearing the query made the absolute snapshot overlay visible again, but recycled FlashList cells did not necessarily fire all first-row `onLayout` callbacks again. The overlay could therefore remain fixed while gestures scrolled the live list underneath it.

## Design Decisions
- Keep the anchor override local to `MobileTokenSelector`; shared `ListView` behavior and unrelated lists remain unchanged.
- Treat the iOS initial-row snapshot as a one-shot cold-start optimization. The first search interaction dismisses it for the remainder of the modal lifecycle.
- Gate both the cached initial list and snapshot overlay after search interaction so an empty filtered result cannot re-enter the cold-start path when the query is cleared.
- Do not add `extraData` or remount the list on every query change because neither addresses the stale overlay and remounting adds avoidable render churn.

## Changes Detail
- `MoblieTokenSelector.tsx`: scope native anchor disabling to the Perps selector and add an iOS snapshot-dismissal latch tied to search interaction.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS behavior fixed; existing Android anchor override preserved)
- **Risk Areas**: iOS selector cold-start transition and search-result restoration

## Test plan
- [ ] On iOS, open the Perps selector and verify the initial full list renders normally.
- [ ] Search for `SKHX`, clear all text, and verify the full list returns at the top and the entire list scrolls.
- [ ] Repeat search and clear with a query that has no results.
- [ ] Verify Perps tab/category switching and Android sort/search scrolling remain functional.

[OK-57864]: https://onekeyhq.atlassian.net/browse/OK-57864?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
+++ b/packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
@@ -429,22 +429,23 @@ function MobileTokenSelectorModal({
   const cachedInitialListRef = useRef<ITokenSelectorListItem[]>(
     getCachedPerpsTokenSelectorInitialList(),
   );
-  const initialListRenderedRef = useRef(!platformEnv.isNativeIOS);
-  const initialListRenderedIndexesRef = useRef<Set<number>>(new Set());
   // iOS only (OK-57864): search interaction permanently ends the cold-start
   // snapshot phase so recycled rows cannot leave a static overlay behind.
-  const initialRowsSnapshotDismissedRef = useRef(!platformEnv.isNativeIOS);
+  const [initialRowsSnapshotDismissed, setInitialRowsSnapshotDismissed] =
+    useState(!platformEnv.isNativeIOS);
+  const initialListRenderedRef = useRef(!platformEnv.isNativeIOS);
+  const initialListRenderedIndexesRef = useRef<Set<number>>(new Set());
   const [hasInitialListRendered, setHasInitialListRendered] = useState(
     !platformEnv.isNativeIOS,
   );
   const dismissInitialRowsSnapshot = useCallback(() => {
-    if (!platformEnv.isNativeIOS || initialRowsSnapshotDismissedRef.current) {
+    if (!platformEnv.isNativeIOS) {
       return;
     }
-    initialRowsSnapshotDismissedRef.current = true;
     initialListRenderedRef.current = true;
     initialListRenderedIndexesRef.current.clear();
     setHasInitialListRendered(true);
+    setInitialRowsSnapshotDismissed(true);
   }, []);
   const fixedTabNames = useMemo(
     () => ({
@@ -1099,7 +1100,7 @@ function MobileTokenSelectorModal({
       DEFAULT_PERP_TOKEN_SORT_DIRECTION;
   const shouldUseCachedInitialList =
     platformEnv.isNativeIOS &&
-    !initialRowsSnapshotDismissedRef.current &&
+    !initialRowsSnapshotDismissed &&
     !searchQuery &&
     isDefaultPerpsSelectorView &&
     mockedListData.length === 0 &&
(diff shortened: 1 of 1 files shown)


packages/kit/src/views/Perp/components/TokenSelector/MoblieTokenSelector.tsx
```diff
@@ -430,9 +431,21 @@ function MobileTokenSelectorModal({
...
+  const initialRowsSnapshotDismissedRef = useRef(!platformEnv.isNativeIOS);
   const [hasInitialListRendered, setHasInitialListRendered] = useState(
     !platformEnv.isNativeIOS,
   );
+  const dismissInitialRowsSnapshot = useCallback(() => {
+    if (!platformEnv.isNativeIOS || initialRowsSnapshotDismissedRef.current) {
+      return;
+    }
+    initialRowsSnapshotDismissedRef.current = true;
+    initialListRenderedRef.current = true;
```
sidmorizon, line 445: # 🟡 P2 快照关闭依赖渲染期读取的 ref，纯空格首次输入时可能短暂不生效

## 问题

`initialRowsSnapshotDismissedRef` 是 `useRef`，但它的值在渲染期的 `const` 计算里被消费（`shouldUseCachedInitialList` 第 `1099` 行、`shouldShowInitialRowsSnapshot` 第 `1227` 行）。ref 变更本身不会触发重渲染，因此关闭快照的正确性完全依赖本函数里搭车的 `setHasInitialListRendered(true)`。

在冷启动缓存列表窗口内，rows 完成 layout 后 `markInitialListRowRendered` 已通过 `requestAnimationFrame` 把 `hasInitialListRendered` 置为 `true`；此时若用户首次输入为纯空格，`afterTrim === ''` 使 `setSearchQuery('')` 也成为 no-op（`searchQuery` 仍为 `''`）。两个 setter 同时空转，`initialRowsSnapshotDismissedRef.current` 被置 `true` 但没有任何重渲染被调度。

## 影响

属于隐性健壮性问题。当前实例是自愈的：下一次 `usePerpsAllAssetCtxsAtom` WS tick 或 `mockedListData` 落地即恢复，为亚秒级瞬态，且该窗口下快照本已隐藏，后果很轻。

但渲染输出的正确性被隐式绑定在「ref 变更必须搭配一次 state 变更」上，后续任何改动（去掉这行 `setHasInitialListRendered(true)`，或调整 `searchQuery` 的更新时机）都可能把它退化成真实的 overlay 卡死。

## 建议

把这个一次性开关改成 state 而非 ref，让它的变化必然驱动一次渲染：

```tsx
const [initialRowsSnapshotDismissed, setInitialRowsSnapshotDismissed] =
  useState(!platformEnv.isNativeIOS);
// dismiss 内改为 setInitialRowsSnapshotDismissed(true)，保留一次性 early-return 语义
// 两个 gated const 与 reset effect 改读 initialRowsSnapshotDismissed
```

[Claude session ↗](https://claude.ai/code/session_013e96ERENxbrKT5SF9gSAeP)

---

huhuanming: Fixed in 2690cde102. I replaced the render-affecting dismissal ref with React state, so whitespace-only input now schedules a rerender and removes the iOS cold-start snapshot immediately instead of relying on a later WebSocket update. The commit-profile lint and TypeScript checks pass.

originalix (approved):

PR #12412, opened 2026-07-10
fix: prevent mobile market tabs from reverting(OK-57367)

limichange:
[OK-57367](https://onekeyhq.atlassian.net/browse/OK-57367)

----------

## Summary
- Coordinate native Market Home tab corrections through a single synchronization path.
- Ignore delayed PagerView callbacks after programmatic tab switches while preserving user swipes.
- Add regression coverage for deferred and cancelled synchronization and stale callback guards.

## Intent & Context
Discovery Market Home tabs could jump back to the previously selected tab after switching tabs and opening a list item. This fix is scoped to tab synchronization; list behavior is unchanged.

## Root Cause
PagerView could emit a delayed callback for the previous page after the programmatic target had already been accepted. Multiple imperative correction paths could then compete and write the stale tab back to the selected-tab atom. iOS was especially exposed because its programmatic settle guard was disabled even when PagerView reported `idle`.

## Design Decisions
- Keep corrective `jumpToTab` and `syncCurrentPage` operations in `useSyncedMarketTab` so requests are coalesced and ordered.
- Defer synchronization while PagerView is settling and cancel pending corrections when a user drag starts.
- Apply a 500 ms post-programmatic settle guard on both native platforms; recent user gestures explicitly bypass the stale-event guard.
- Keep the change isolated from Market list item behavior.

## Changes Detail
- Track expected targets, PagerView state, and user-drag ownership in the native Market Home layout.
- Add request and cancel semantics plus deferred retry coordination to the synchronized-tab hook.
- Extract and test the stale programmatic callback predicate.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (iOS and Android)
- **Risk Areas**: Programmatic tab changes, manual tab swipes, nested Discovery pager transitions, and returning from Market detail pages.
- **Runtime Scope**: Main UI runtime only. The native PagerView is involved, but no shared native resource ownership or persistence changes. The bg runtime and its JS heap are unaffected.

## Test plan
- [x] Run targeted Jest tests for `useSyncedMarketTab` and `marketTabChangeGuards`.
- [x] Run `yarn agent:check --profile commit`.
- [x] Verify repeated tab switches and route returns on Android.
- [ ] Recheck tap, swipe, open-item/back, and Discovery re-entry flows on iOS with the final settle guard.
- [ ] Recheck the same interaction flows in the PR build on Android.

[...]

PR #12438, opened 2026-07-10
fix: prevent market tab revert from stale atom echo(OK-57367)

originalix:
## Summary
- Stop the Market home pager from reverting the user's tab selection right after a tap settles (OK-57367 still reproduced 100% on Android release builds after #12412).
- Guard the two places where state derived from the bg-synced \`marketSelectedTabAtom\` could act on a stale UI mirror: the local category restore effect and the tab resync loop.
- Add \`marketTabSelectionGuards.ts\` pure helpers with focused unit tests.

## Intent & Context
QA recordings on Android release packages showed every Trending → US-stocks tab tap sliding back to Trending ~100ms after the pager settled, occasional revert ping-pong (revert → jump forward → revert again), and a double revert (Perps → US-stocks → Trending) after tapping Perps. The tab indicator did not move during the backward slide, proving the motion was programmatic rather than user input. #12412 (OK-57367) added accept-side guards for pager events but did not cover the resync trigger side, so the revert survived.

## Root Cause
Three layers interact:

1. **Structurally async atom writes (main → bg → main).** \`marketSelectedTabAtom\` is a \`persist: true\` cross-runtime globalAtom. On production native builds (\`ENABLE_NATIVE_BACKGROUND_THREAD=true\`) and extension UI, \`wrapAtomPro\` proxies UI writes to the bg runtime and the UI mirror only updates when bg broadcasts the echo back. The mirror stays stale for tens to hundreds of milliseconds after every write. This is also why dev single-runtime builds do not reproduce the bug.
2. **Stale echo restore (the poison).** \`MarketHomeV2\`'s restore effect unconditionally copied the stale \`selectedSpotCategory\` mirror back into the local \`selectedCategory\` state, deterministically undoing the category the user just picked. \`selectedTabName\` then derived to the previous tab.
3. **Resync loop executes the revert.** \`useSyncedMarketTab\` treats \`selectedTabName\` as authoritative; once the Android pager goes idle it issues \`jumpToTab(staleTarget)\`, and \`onBeforeJumpToTab\` marks that jump as the expected target, so the accept-side guards from #12412 approve and persist the wrong state. The later atom echo then flips the target again, producing the ping-pong.

## Design Decisions
- **Pager is authoritative during user interaction; the store is an async follower.** Store-driven pager jumps are only allowed for genuinely external changes (deep links, \`useNavigateToMarketTab\`), which do not go through \`handleTabChange\` and are therefore never flagged as in-flight.
- \`MarketHomeV2\` tracks the last explicitly applied category (\`pendingSpotCategoryRef\` via \`applySelectedCategory\`) and skips the restore effect until the atom echo catches up, which also ignores out-of-order echoes of older writes. Cold-start restore (no pending intent yet) is unchanged.
- \`useMarketTabsLogic\` records the last selection written to the atom and exposes \`isTabSelectionInFlight()\`; \`MobileLayout.native\`'s \`shouldDeferPageSync\` defers page resync while a write is in flight. After the echo lands the loop either converges (no jump) or performs a legitimate external jump.
- Deliberately did NOT change \`wrapAtomPro\` semantics (optimistic local writes would affect every global atom app-wide) nor the react-native-collapsible-tab-view patch (Android \`onPageSelected\` phantom events; candidate for separate hardening).
- Web/desktop apply atom writes synchronously in-process, so both guards are structural no-ops there.

## Changes Detail
- \`layouts/marketTabSelectionGuards.ts\` (new): pure decision helpers \`shouldRestoreSpotCategoryFromAtom\` and \`getIsMarketTabSelectionInFlight\`.
- \`layouts/marketTabSelectionGuards.test.ts\` (new): 9 unit tests covering cold start, lagging echo, out-of-order echoes, category-less (watchlist/perps) writes.
- \`MarketHomeV2.tsx\`: route all intent-driven category updates through \`applySelectedCategory\`; guard the restore effect with \`shouldRestoreSpotCategoryFromAtom\`.
- \`layouts/hooks/useMarketTabsLogic.ts\`: track \`lastWrittenSelectionRef\` on atom writes; expose \`isTabSelectionInFlight()\`.
- \`layouts/MobileLayout.native.tsx\`: defer page resync while a tab selection is in flight.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (primary), Extension UI (same proxied-write path); Web/Desktop structurally unaffected
- **Risk Areas**: cold-start restore of the persisted spot category, deep-link \`spotCategoryToSelect\` flow, external tab navigation via \`useNavigateToMarketTab\`, and rapid tap/swipe sequences while a previous write is still in flight

## Test plan
- [x] \`npx jest packages/kit/src/views/Market/MarketHomeV2/layouts/\` — 30/30 passed (9 new)
- [x] \`yarn agent:check --profile commit\` — lint/tsc all green
- [ ] QA on \`ENABLE_NATIVE_BACKGROUND_THREAD=true\` package ([release-app-bundles run](https://github.com/OneKeyHQ/app-monorepo/actions/runs/29107075023)): repeated Trending ↔ US-stocks taps (×10, no revert/ping-pong), Watchlist → US-stocks cross-page tap, Perps round trips, horizontal swipe gestures, deep link into a specific category, kill & relaunch restores last category
- [ ] iOS same matrix; verify web/desktop market category switching unchanged

### DIFF
--- a/packages/kit/src/views/Market/MarketHomeV2/MarketHomeV2.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/MarketHomeV2.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { useIntl } from 'react-intl';
 
@@ -33,6 +33,7 @@ import { useMarketRenderCommitProbe } from '../utils/marketReactPerf';
 
 import { useNetworkAnalytics, useTabAnalytics } from './hooks';
 import { DesktopLayout } from './layouts/DesktopLayout';
+import { shouldRestoreSpotCategoryFromAtom } from './layouts/marketTabSelectionGuards';
 import { MobileLayout } from './layouts/MobileLayout';
 import { isMarketStockCategory } from './utils';
 
@@ -109,6 +110,14 @@ const useMarketHomeLayoutProps = () => {
   const [selectedCategory, setSelectedCategory] = useState(
     selectedSpotCategory || 'trending',
   );
+  // The category most recently applied by an explicit intent (user tap or
+  // deep link). The bg-synced atom echoes writes back asynchronously, so this
+  // ref is the freshest truth; the atom may lag behind it.
+  const pendingSpotCategoryRef = useRef<string | undefined>(undefined);
+  const applySelectedCategory = useCallback((categoryId: string) => {
+    pendingSpotCategoryRef.current = categoryId;
+    setSelectedCategory(categoryId);
+  }, []);
 
   const categories: IMarketCategoryItem[] = useMemo(() => {
     if (apiSpotCategories.length > 0) {
@@ -157,12 +166,24 @@ const useMarketHomeLayoutProps = () => {
       return;
     }
 
+    // The atom echoes UI writes back after a bg round-trip; until it carries
+    // the latest locally-applied category, restoring from it would revert the
+    // user's selection and make the pager jump back (OK-57367).
+    if (
+      !shouldRestoreSpotCategoryFromAtom({
+        pendingSpotCategoryId: pendingSpotCategoryRef.current,
+        atomSpotCategoryId: selectedSpotCategory,
+      })
+    ) {
+      return;
+    }
+
     const hasSelectedCategory = categories.some(
       (item) => item.id === selectedSpotCategory,
     );
     if (hasSelectedCategory) {
       if (selectedCategory !== selectedSpotCategory) {
-        setSelectedCategory(selectedSpotCategory);
+        applySelectedCategory(selectedSpotCategory);
       }
       return;
     }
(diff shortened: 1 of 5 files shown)


packages/kit/src/views/Market/MarketHomeV2/MarketHomeV2.tsx
```diff
@@ -157,20 +166,32 @@ const useMarketHomeLayoutProps = () => {
       return;
     }
 
+    // The atom echoes UI writes back after a bg round-trip; until it carries
+    // the latest locally-applied category, restoring from it would revert the
+    // user's selection and make the pager jump back (OK-57367).
+    if (
+      !shouldRestoreSpotCategoryFromAtom({
```
originalix, line 173: 这里用 `pendingSpotCategoryRef` 去挡旧 atom 回声的思路没问题，但它只会被写入，不会在 atom 已经追平后释放。

这样一来，只要本地曾经选中过某个 spot 分类，后面如果 `bg` 或另一个 UI runtime 再把 `selectedSpotCategory` 同步成新的值，`shouldRestoreSpotCategoryFromAtom()` 仍会拿旧 pending 值比较并直接 `return`，当前页面就会一直停在旧分类，新的真实同步进不来。

建议在确认 atom 已经追平当前 pending 后立刻清空 pending，或者把它改成一次性的 sequence/epoch 标记，只屏蔽这一轮回声，不要长期拦截后续同步。

Minnzen (approved):
代码审查通过：未发现阻塞性问题。已核对异步 atom 回传的 main/bg 时序、导入层级与 CI 状态；建议后续按 PR 测试计划完成原生 release 真机矩阵验证。

PR #12442, opened 2026-07-11
fix: restore Perps actions and empty-wallet deposits(OK-57890)

Minnzen:
## Summary
- Restore visible Perps order-list actions with the intended icon interaction states.
- Keep server-configured Perps deposit defaults selectable when a new wallet has no qualifying assets.

## Intent & Context
Users reported that isolated-margin editing disappeared from the Perps position panel, following the same class of regression as a previously missing position-share action. A second report showed an empty token selector when a new account opened the Perps deposit modal.

## Root Cause
- The desktop action icon relied on group hover/press styles that did not apply in the rendered list context, leaving the action visually unavailable.
- A successful empty wallet-token response replaced the configured deposit-token list; the selector therefore lost its server default token after zero-balance filtering.

## Design Decisions
- Use direct icon hover and press properties, preserving the requested Kahn-style visual behavior with a one-file action fix.
- Apply the default-token fallback only when the wallet token list is empty, across initial load and refresh paths. Funded wallets retain the existing positive-fiat filtering and order.

## Changes Detail
- Update the shared desktop Perps action icon to use direct icon interaction styles.
- Add a small deposit-token fallback helper, wire it into deposit token synchronization, and cover an empty-wallet default-token case.

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: Perps order-list icon interactions and deposit-token refresh behavior.

## Test plan
- [x] Verify isolated-margin edit and related Perps actions are visible and usable.
- [x] Run targeted Jest tests for the deposit modal and deposit token list utilities.
- [x] Run ESLint for the three deposit-fix files.
- [ ] Verify a new account can select the configured default token in the Perps deposit modal.

### DIFF
--- a/packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
+++ b/packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
@@ -95,6 +95,7 @@ import {
   getPerpsDepositTokenDisplayList,
   getPerpsDepositTokensWithDefaultFallback,
   mergePerpsDepositTokensPreservingOrder,
+  shouldPreservePerpsDepositSelectedToken,
   shouldShowPerpsDepositTokenSkeleton,
 } from './depositTokenDisplayUtils';
 import { DepositTokenSelectionContent } from './DepositTokenSelectionContent';
@@ -496,7 +497,7 @@ function DepositWithdrawContent({
     [],
   );
 
-  const { result, isLoading: balanceLoading } = usePromiseResult(
+  const { isLoading: balanceLoading } = usePromiseResult(
     async () => {
       const requestKey = depositTokenRequestKey;
       if (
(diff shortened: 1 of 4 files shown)


originalix (dismissed):

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -522,15 +523,19 @@ function DepositWithdrawContent({
           return [];
         }
         depositTokenListOwnerKeyRef.current = ownerKey;
+        const displayDepositTokens = getPerpsDepositTokensWithDefaultFallback({
```
originalix, line 527: 当钱包初始为空时，这里会把 `defaultTokens` 直接当成 `result` 返回，后面的 `resolvePerpsDepositSelectedToken(... preserveCurrentToken: depositTokenListSource === 'walletBalance')` 就会把这个 server fallback token 记成当前 wallet 选中项。

一旦用户随后到账，或者静默刷新拿到了真实 wallet token，`preserveCurrentToken` 仍然会优先保留这个旧 fallback token，即使它已经不在新的 `depositTokensWithPrice` 列表里。结果是表单继续显示 0 余额 / 无 price 的旧 token，入金最小额校验和 quote 都可能起不来，必须手动重选一次。

建议把 `defaultTokens` 只作为渲染层的空钱包占位，不要把它喂进 `result` 和选中态同步链路；真实 wallet tokens 到达时应重新按真实列表解析 selected token。

huhuanming (dismissed):

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed in e12402de38.

The selected-token synchronization now follows `depositTokensWithPrice`, so it also runs after silent and atom-driven wallet refreshes. A token is preserved only when it is still present in the latest displayed list; otherwise the selection is resolved from the refreshed wallet tokens. The server default remains an empty-wallet display fallback, and the regression is covered by a focused test.

originalix (approved):

PR #12445, opened 2026-07-11
fix: merge Perps deposit server tokens(OK-57887)

originalix:
## Summary
- Merge wallet home tokens with the complete Perps deposit token list from `perp-config.depositTokensByNetwork`.
- Preserve wallet balance, fiat value, and price metadata for matching tokens while keeping server-configured tokens available to empty accounts.
- Add owner-scoped caching, selection resolution, and regression coverage for empty and funded accounts.

## Intent & Context
OK-57887 reports that a newly created or empty wallet account can open the Perps deposit flow with an incomplete or empty token selector. The expected list is the union of the wallet home token list and the complete server-configured deposit token list. Empty accounts must still see every token configured by `depositTokensByNetwork`, while funded accounts should prioritize their available assets and retain live wallet metadata.

## Root Cause
The wallet-balance refresh replaced the server-configured token map. Wallet tokens without a positive fiat value are intentionally filtered out, so an empty account produces no wallet-derived tokens. The only server data retained across the account-owner refresh was `defaultTokens`, which represents selection hints rather than the complete `depositTokensByNetwork` list. As a result, empty accounts displayed only one default token per network instead of the full server list, and server-only tokens had no explicit zero-balance state.

## Design Decisions
- Store the complete parsed server token list separately as `serverTokens`; keep `defaultTokens` dedicated to default selection behavior.
- Merge by normalized `networkId + contractAddress`, with wallet fields taking precedence so live balances, fiat values, prices, and token metadata are preserved.
- Append missing server tokens with explicit zero balance and fiat value after wallet balance loading completes.
- Keep wallet tokens first so funded assets remain prioritized, while preserving all server-configured tokens for empty accounts.
- Keep the merge in the background service, with account-owner and write-generation guards, instead of adding a UI-only fallback.

## Changes Detail
- `ServiceHyperliquid.ts`: retain the complete parsed `depositTokensByNetwork` token snapshot.
- `ServiceWebviewPerp.ts`: fetch account-scoped wallet tokens, use memory/cold caches safely, and merge them with the complete server list.
- `depositTokenListUtils.ts`: add deterministic mapping, filtering, sorting, selection, grouping, and normalized merge helpers.
- `perps.ts`: add `serverTokens` to the Perps deposit token atom contract.
- Tests cover empty accounts, funded accounts, deduplication, selection priority, stale owners, cache behavior, and partial network-account availability.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop, Mobile, Web, Extension
- **Risk Areas**: Perps deposit token ordering and default selection, account switching, cached wallet balance freshness, and server-config updates.

## Test plan
- [x] Run targeted Perps deposit token tests: 21 tests passed.
- [x] Verify an empty account on Desktop: the selector contains all 22 tokens from the current seven-network `depositTokensByNetwork` response, including non-default USDT/USDC variants.
- [x] Verify matching wallet tokens retain wallet balance, fiat value, and price metadata through unit tests.


packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -332,6 +333,12 @@ function DepositWithdrawContent({
     () => getPerpsDepositTokenDisplayList(tokens),
     [tokens],
   );
+  const defaultTokenIdentitiesKey = useMemo(
```
originalix, line 336: 这里把重新同步 token 列表的触发条件收窄成了 `defaultTokenIdentitiesKey`。当服务端只增删非默认的 deposit token，或者只更新 default 之外的 server token 集合时，这个 key 不会变化；同时 `parseDepositConfig()` 更新 `serverTokens/tokens` 时也没有递增 `depositTokenListRevision`。

结果是 modal 已经打开的情况下，后台 atom 虽然拿到了新的 server token 集合，这里不会重新 fetch / sync，用户会继续看到旧 token 列表，直到重新打开弹窗或触发一次额外 refresh。建议把触发条件扩展到完整的 server token identity 集合，或者在服务端 token 配置变化时 bump 一个 revision，让这个同步路径能跟随非 default token 的变化重新收敛。

zhaono1 (approved):

Minnzen (approved):

PR #12443, opened 2026-07-11
fix: render name labels with proportional figures, not tabular (OK-57876)

PatrickChoo:
## Summary

Fixes [OK-57876](https://onekeyhq.atlassian.net/browse/OK-57876). Since #12391 made tabular figures (`tnum`) the app-wide default, **user-facing names render with tabular digits** — which is wrong for names.

Roobert's tabular `1` is a different glyph: to fill the wider tabular advance it nearly doubles in width (23.3px → 45.4px at 72px) **and gains a foot serif**. In an account name like `A1787` or a wallet name like `OneKey Pro2-1-6`, that reads as a typo rather than a name.

Tabular figures exist to align columns of ticking numeric data. A name is an identifier, not data — so names now opt out via `fontVariant={PROPORTIONAL_NUMS}` (the escape hatch shipped with the default).

## The rule

- **Numeric data** (balances, prices, amounts, percentages, counts) → stays **tabular**. Unchanged.
- **Names / identifiers / prose** (account, wallet, device, address-book, market category, marketing copy) → **proportional**.

Tabular is for digits that sit in a column and tick. Everything a human reads as words gets the font's natural figures.

## Changes

**Component layer (the choke-points):**
- `ListItem`: the `primary`/`title` text now defaults to proportional. A list item's title is a name; its `secondary` is where numeric data lives, and that stays tabular. This covers many lists in one place. Callers can still override via `primaryTextProps`.
- `Badge.Text`: previously hard-coded tabular **and silently dropped the `fontVariant` prop** (it is styled from raw tamagui text). It now honors `fontVariant` (default stays tabular), so a badge holding a name can opt out.

**Name render sites** (account + wallet names, incl. the ones that pass an element-form `primary` and therefore bypass the `ListItem` default):
- Account: `AccountSelectorAccountListItem` (the reported `A1787`), `AccountRenameButton`, `AddressListItem`, `UniversalSearchAddressItem`
- Wallet: `WalletListItem` (the wallet rail, incl. `Hidden #1`), `WalletRenameButton`, `DeviceInfoCard`, `DeviceManagementListModal`
- Triggers: `AccountSelectorTriggerBase`, `AccountSelectorTriggerDApp`
- Badges: `AddressInfo`, `AddressInput` (wallet/account name + address-book name)

**Content names & prose:**
- `CategoryFilterItem` — market category chips (`Test 0702`, `Mag 7`, `Crude Oil`). Shared, so it covers the market / perps / stock category selectors at once.
[...]

PR #12391, opened 2026-07-09
feat: default tabular numerals app-wide + localized thousands separators

PatrickChoo:
## Summary

Make **tabular (equal-width) figures the app-wide default** for all UI text, and give quoted prices locale-aware thousands separators. Tabular digits keep numbers from reflowing as they tick and keep columns aligned — the standard treatment for trading interfaces. We get it from the OpenType `tnum` feature our font (Roobert) already ships, applied globally.

## How it works (two mechanisms, one per platform)

- **Native**: `SizableText` (the wrapper every text primitive routes through) defaults `fontVariant` to `TABULAR_NUMS`, translated into a real RN `style.fontVariant` (Tamagui silently drops the raw prop). Reference-stable and allocation-free on the hot path.
- **Web**: a single `body { font-variant-numeric: tabular-nums }` rule in `web-fonts.css` (already imported by every web-stack app). CSS inheritance covers every text node — including ones that bypass the wrapper (raw `<Text>`, `Badge.Text`, portals) — so the wrapper emits **no** inline style for the default on web.

Both are opt-out: `fontVariant={PROPORTIONAL_NUMS}` restores proportional figures for a specific text.

Text that bypasses the wrapper on **native** applies the feature at the source: `Badge.Text` is centralized in the `Badge` component (all badge digits go tabular in one place), and raw RN `<Text>` in the order book / chart labels carries it via a `StyleSheet` entry.

**Large hero balances stay proportional** (the escape hatch): the home total balance, account-overview balance, token-detail fiat value, and balance-details dialog use `PROPORTIONAL_NUMS`. A single large display number has no column to align with, and the font's natural proportional figures read better at hero sizes — tabular is reserved for tables and ticking data.

## Also in this change

- **Localized thousands separators** on quoted prices (order book `displayPrice`, ticker/oracle price, TP-SL, order confirm, positions/history rows) via the existing locale-aware `formatLocalizedNumberString` — European `78.056,1` and Indian grouping work out of the box. Formatting happens at the producing memo, not in JSX.
- **Perps ticker bar (desktop)**: headline mark price weight → 500 (the heading token's 600 reads too heavy for a dense tabular stat strip).

## Verification
- `tsgo --noEmit`: 0 errors in changed files; `oxlint --type-aware --deny-warnings`: clean; `numberUtils.locale.test.ts`: 72/72.
[...]

PR #12452, opened 2026-07-11
refactor: tabular figures only on dense numeric data; proportional by default (OK-57876)

PatrickChoo:
[OK-57876](https://onekeyhq.atlassian.net/browse/OK-57876)

----------

## Summary

Inverts the tabular-numerals policy shipped in #12391.

**Before:** `SizableText` defaulted `fontVariant = TABULAR_NUMS` and `web-fonts.css` carried a `body { font-variant-numeric: tabular-nums }` rule — so *every* string in the app rendered with equal-width digits.

**After:** text is **proportional by default**. Tabular figures are an explicit **opt-in**, applied only to numbers that are **small, sit in a column, and tick** — market/perps list columns, order books, trading tables, tickers, fee rows.

Supersedes #12443 (which was patching the symptoms) and fixes [OK-57876](https://onekeyhq.atlassian.net/browse/OK-57876).

## Why invert

Tabular digits exist to keep a *column* aligned. A global default applies them to everything a human reads as words, where they are actively wrong:

- **Names** — `A1787`, `OneKey Pro2-1-6`. Roobert's tabular `1` is ~2× wider and grows a foot serif; a name reads like a typo. (This is OK-57876.)
- **Addresses** — `0x7a84...4444`. Hex mixes digits and letters, so tabular equalizes only *half* the characters and the string comes out ragged — worse than either pure option.
- **Prose / marketing copy**, **amount input fields**, and **large hero balances** (a single big number has no column to align with).

Under the old default each of these needed its own opt-out, and the list only ever grew. The set of surfaces that genuinely *want* tabular, by contrast, is bounded and enumerable: tables and tickers. So the default was on the wrong side.

Net effect: names, addresses, prose, inputs and hero numbers are now correct **with zero annotations** — the escape hatches for them are all deleted.

## The rule

| | |
|---|---|
| **Tabular (opt-in)** | Small numbers in a column that tick — list/table cells, order book, tickers, fee rows |
| **Proportional (default)** | Everything else — names, addresses, labels, headers, prose, inputs, hero numbers |

Documented on `TABULAR_NUMS` in `packages/components/src/utils/tabularNums.ts`.

## Changes

**Core (`@onekeyhq/components`)**
- `SizableText` — dropped the app-wide `fontVariant` default (it still *translates* an explicit `fontVariant` into a real style, since Tamagui drops the raw prop).
- `web-fonts.css` — dropped the global `body` rule.
- `tabularNums.ts` — `TABULAR_NUMS_STYLE` now always builds a real style. It previously resolved to `undefined` on web because the body rule covered it; without that rule, an explicit opt-in would have been a **no-op on web**. `PROPORTIONAL_NUMS` is gone — it has nothing left to escape from.
- `Badge.Text` — no longer hard-codes tabular; it now honors a `fontVariant` prop (it is styled from raw tamagui text, which silently drops the prop).

**Opt-ins** — market/perps list columns, order book (incl. the raw-RN `<Text>` depth rows), ticker bars, perps trading tables, token/home lists, DeFi & Earn/Borrow/Staking tables, transaction history, Swap Pro, provider/quote lists, referral tables, gas-fee editors.

New shared choke-point `PerpTableCellText` collapses ~89 perps table cells into one component, so a new column can't silently forget.

## Verification

- `tsgo --noEmit`: no new errors. `oxlint --type-aware --deny-warnings` on all 120 files: clean.
- Roobert's glyphs were confirmed by rendering the shipped font directly: the tabular `1` nearly doubles in width and gains a foot serif; the proportional `1` does not.
- Live web probes assert the split **in both directions** per screen:
  - Market — 446 numeric cells tabular; category names, badges and **all 35 addresses** proportional.
  - Perps — 113 tabular (mark price, order book, funding, countdown); labels and controls proportional.
  - DeFi — APY columns tabular; the `$0.00` hero proportional.

## Review notes

A cross-review (local + Codex) caught real gaps in the first pass, all fixed here:
- **`TxActionCommon`'s opt-in was dead.** Real transfers pass the amount as a `ReactNode`, which bypassed the `typeof change === 'string'` branch the prop was added to — the wallet's whole transaction-history amount column was silently untouched. Fixed at the `TxActionTransfer` / `TxActionTokenApprove` sources.
- **`MarketHomeList` was half-converted**, which is worse than untouched — its volume / market-cap / rank columns jittered against already-tabular price cells in the same table.
- **Whole modules had been missed**: Staking, Referral, the two gas-fee editors, Home's perps tab, PopularTrading.

[OK-57876]: https://onekeyhq.atlassian.net/browse/OK-57876?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Home/pages/PerpsContainer.tsx
+++ b/packages/kit/src/views/Home/pages/PerpsContainer.tsx
@@ -13,6 +13,7 @@ import {
   SizableText,
   Skeleton,
   Stack,
+  TABULAR_NUMS,
   Tabs,
   XStack,
   YStack,
@@ -202,17 +203,26 @@ function useOpenPerpAsset() {
   );
 }
 
+// These wrappers exist to render live USD values that tick off the websocket and
+// sit in columns, so they opt into tabular figures by default; callers can still
+// override with an explicit `fontVariant`.
 function PerpsUsd({
   value,
+  fontVariant = TABULAR_NUMS,
   ...rest
 }: { value: number | undefined } & Omit<ISizableTextProps, 'children'>) {
   if (value === undefined) {
-    return <SizableText {...rest}>--</SizableText>;
+    return (
+      <SizableText fontVariant={fontVariant} {...rest}>
+        --
+      </SizableText>
+    );
   }
   return (
     <NumberSizeableText
       formatter="value"
       formatterOptions={{ currency: '$' }}
+      fontVariant={fontVariant}
       {...rest}
     >
       {value}
@@ -223,6 +233,7 @@ function PerpsUsd({
 function PerpsTotalUsd({
   value,
   isDegraded,
+  fontVariant = TABULAR_NUMS,
   ...rest
 }: {
   value: number | undefined;
@@ -244,6 +255,7 @@ function PerpsTotalUsd({
       <NumberSizeableText
         formatter="value"
         formatterOptions={{ currency: settings.currencyInfo.symbol }}
+        fontVariant={fontVariant}
         {...rest}
       >
         {displayValue}
@@ -258,6 +270,7 @@ function PerpsTotalUsd({
       <NumberSizeableText
         formatter="value"
         formatterOptions={{ currency: settings.currencyInfo.symbol }}
+        fontVariant={fontVariant}
         {...rest}
       >
         {displayValue}
@@ -270,6 +283,7 @@ function PerpsTotalUsd({
 // carry the sign via a currency prefix + color (mirrors PositionsRow's pnl).
 function PerpsSignedUsd({
   value,
+  fontVariant = TABULAR_NUMS,
   ...rest
 }: { value: number | undefined } & Omit<
   ISizableTextProps,
@@ -277,7 +291,7 @@ function PerpsSignedUsd({
 >) {
   if (value === undefined) {
     return (
-      <SizableText color="$textSubdued" {...rest}>
+      <SizableText color="$textSubdued" fontVariant={fontVariant} {...rest}>
         --
       </SizableText>
     );
@@ -288,6 +302,7 @@ function PerpsSignedUsd({
       formatter="value"
       formatterOptions={{ currency: negative ? '-$' : '+$' }}
       color={negative ? '$red11' : '$green11'}
+      fontVariant={fontVariant}
       {...rest}
     >
       {new BigNumber(value).abs().toFixed()}
@@ -728,6 +743,7 @@ function PerpsEmptyRecommendSection() {
                   textAlign="right"
                   formatter="price"
                   formatterOptions={{ currency: '$' }}
+                  fontVariant={TABULAR_NUMS}
                 >
                   {token.markPrice}
                 </NumberSizeableText>
@@ -760,6 +776,7 @@ function PerpsEmptyRecommendSection() {
                   }
                   formatter="priceChange"
                   formatterOptions={{ showPlusMinusSigns: true }}
+                  fontVariant={TABULAR_NUMS}
                 >
                   {token.change24hPercent}
                 </NumberSizeableText>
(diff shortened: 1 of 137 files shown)


packages/kit/src/views/Home/pages/PerpsContainer.tsx
```diff
@@ -223,6 +233,7 @@ function PerpsUsd({
 function PerpsTotalUsd({
   value,
   isDegraded,
+  fontVariant = TABULAR_NUMS,
```
originalix, line 236: 这里给 `PerpsTotalUsd` 设了 `fontVariant = TABULAR_NUMS`，但这个 wrapper 不只用于持仓列表，也直接被首页和桌面 Perps 头部的 `$headingXl` 账户总额复用。触发条件就是 Perps 头部展示总资产时；调用方没有显式覆盖的话，hero 数字会被强制切到等宽字形，`1` 会明显变宽，和这次把 tabular 收窄到 dense data 的目标相反。

建议把这里的默认值去掉或改回 proportional，只在真正的列式/列表场景显式传 `fontVariant={TABULAR_NUMS}`。

packages/kit/src/views/Staking/components/ProtocolDetails/GridItemV2.tsx
```diff
@@ -75,13 +81,21 @@ export function GridItem({
             cursor="pointer"
             onPress={() => openUrlExternal(actionIcon?.data?.link)}
           >
-            <EarnText text={description} size="$bodyLgMedium" />
+            <EarnText
+              text={description}
+              size="$bodyLgMedium"
+              fontVariant={TABULAR_NUMS}
```
originalix, line 87: 这里把 `GridItemV2` 的 `description` 槽位统一传成了 `fontVariant={TABULAR_NUMS}`。触发 `provider` / `profit` 这类后端驱动卡片时，只要 `description.text` 不是纯数字而是普通 `IEarnText`，这个通用组件就会替调用方强行开启等宽数字。

结果是字体策略重新从业务调用点回流到了基础卡片层，后续任何带数字的说明文本都会被同一层默认值误伤。建议让 `GridItemV2` 默认保持 proportional，只在确认是数值字段的具体 item 或调用点显式传 `fontVariant={TABULAR_NUMS}`。

packages/kit/src/views/Staking/components/ProtocolDetails/EarnAmountText.tsx
```diff
@@ -30,9 +34,13 @@ const LEADING_NUMBER_PATTERN = /^(\d[\d,]*(?:\.\d+)?)([\s\S]*)/;
  *      while any trailing text (token symbol) is rendered as-is.
  *
  * Falls back to plain SizableText when the string does not start with a number.
+ *
+ * Every branch renders numeric data, so this component opts into tabular
+ * figures by default; callers can still override `fontVariant`.
  */
 export function EarnAmountText({
   children: text,
+  fontVariant = TABULAR_NUMS,
```
originalix, line 43: 这里把 `EarnAmountText` 默认成了 `TABULAR_NUMS`，但这个 wrapper 不只服务于密集数列，也被 `UniversalWithdraw` 这类单值大号金额直接复用，还会继续走下面的 `passthrough` / 混合文本分支。触发这些 one-off amount 或普通文本路径时，调用方没有显式覆盖的话，本来应该恢复 proportional 的金额也会被强制切成等宽数字。

这会把这次想收窄到业务调用点的字形决策又重新推回通用 wrapper，连组件注释里“suffix rendered as-is”的边界也一起变宽了。建议去掉这里的默认值，只让明确需要列对齐的调用点显式传 `fontVariant={TABULAR_NUMS}`。

PR #12467, opened 2026-07-13
fix: resolve Jira UI issues(OK-57565)

zhaono1:
## Summary
- Preserve structured Pendle transaction tooltips instead of flattening text tooltip payloads into strings.
- Keep remote typography fields such as description.size available to the shared Earn tooltip renderer.
- Add focused regression coverage for the structured tooltip handoff.

## Intent & Context
This is a Draft aggregation PR for selected Jira UI fixes. The initial scope fixes OK-57565, where the Pendle fee tooltip text rendered larger than other tooltips and ignored typography supplied by the transaction-confirmation payload.

## Root Cause
Pendle transaction details treated text tooltips without comparison items as plain strings. The App extracted only description.text and passed it to CalculationListItem.Label, which rendered Popover.Tooltip with a hard-coded bodyLg size. The structured IEarnTooltip payload, including description.size, was discarded before rendering.

## Design Decisions
- Route every structured Pendle transaction detail tooltip through EarnTooltip, including simple text tooltips.
- Keep CalculationListItem.Label for rows without tooltip data.
- Avoid changing the shared Popover.Tooltip default because that would affect unrelated product surfaces.
- Remove the temporary background-response mock used during diagnosis.

## Changes Detail
- Update PendleSharedComponents to preserve and render the complete tooltip DTO.
- Add a focused hook test proving the same structured tooltip object reaches EarnTooltip.

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: Pendle transaction-detail tooltip presentation only; trigger and popover behavior now use the existing Earn tooltip implementation.

## Test plan
- [x] Focused Jest test for structured Pendle tooltip handoff
- [x] yarn agent:check --profile commit
- [ ] Runtime verify the Pendle fee tooltip against a transaction-confirmation response containing description.size
- [ ] Add subsequent selected Jira UI fixes as separate commits while this PR remains Draft

### DIFF
--- a/packages/kit/src/views/Staking/components/ProtocolDetails/PendleSharedComponents.test.tsx
+++ b/packages/kit/src/views/Staking/components/ProtocolDetails/PendleSharedComponents.test.tsx
@@ -1,0 +1,91 @@
+import { Children } from 'react';
+
+import type { ReactElement, ReactNode } from 'react';
+import { renderHook } from '@testing-library/react-native';
+
+import type {
+  IEarnTooltip,
+  IStakeTransactionConfirmation,
+} from '@onekeyhq/shared/types/staking';
+
+import { EarnTooltip } from './EarnTooltip';
+import { usePendleTransactionDetails } from './PendleSharedComponents';
+
+
+jest.mock('@onekeyhq/components', () => ({
+  Divider: 'Divider',
+  Icon: 'Icon',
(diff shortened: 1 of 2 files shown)


zhaono1:
Closing this draft because the fix is being rebuilt from the latest x branch with a narrower regression-safe implementation.

PR #12436, opened 2026-07-10
feat: align swap k-line and same-chain token selection (OK-57650, OK-57751)

zhaono1:
[OK-57650](https://onekeyhq.atlassian.net/browse/OK-57650) [OK-57751](https://onekeyhq.atlassian.net/browse/OK-57751)

----------

## Summary

- Enable the existing native chart-controls mode for the Swap K-line view.
- Add a `Same chain` network shortcut to the Swap token selector based on the opposite token's network.
- Keep the shortcut first in compact and full selectors, preserve disabled bridge-only states, recalculate `N+`, and align the new UI with the Figma dimensions.

## Intent & Context

OK-57650 aligns the Swap K-line controls with the Market native control layer without changing chart data or trading behavior.

OK-57751 makes same-chain selection immediately visible in the combined Swap & Bridge token selector. When selecting the From token, the To token network is featured; when selecting the To token, the From token network is featured. This behavior is limited to the regular Swap selector.

## Root Cause

- The Swap `TradingViewV2` adapter did not opt in to `enableNativeChartControls`, so the hosted chart rendered its legacy control UI.
- The token selector rendered fixed slices of the network list and had no presentation model for a featured network. It could not pin the opposite token network without duplication, keep a disabled featured network visible, or calculate `N+` from the rendered list.

## Design Decisions

- Enable the existing chart-control opt-in only at the ordinary Swap K-line adapter and preserve its data source, storage namespace, callbacks, and disabled features.
- Derive `sameChainNetwork` as selector presentation data instead of mutating the underlying network order.
- Keep Limit and Stock selectors unchanged.
- Pass optional featured-network metadata through the generic chain selector without adding Swap dependencies to shared components.
- Keep disabled non-featured networks filtered while allowing the disabled featured network to remain visible.
- Use the exact Figma logical dimensions for the 36px compact controls, 1x24 divider, 48px featured row, and 9px full-list divider region.

## Changes Detail

- `SwapKLineContent.tsx`: enable native chart controls on the existing `TradingViewV2` instance and hide the duplicate selector.
- Extend the network filter item with optional badge, responsive image-size, and text props.
- Add and test a Swap-only network view model that extracts the opposite token network and removes its duplicate.
- Update the compact network row and full chain selector with featured-network ordering, badge, divider, disabled state, and responsive `N+` behavior.
- Update generated translations in a dedicated commit.

## Issues

- Fixes OK-57650
- Fixes OK-57751

## Risk Assessment

- **Risk Level**: Medium
[...]

PR #12485, opened 2026-07-14
feat: add same-chain network shortcut(OK-57751)

zhaono1:
[OK-57751](https://onekeyhq.atlassian.net/browse/OK-57751)

----------

## Summary

- Add a `Same chain` shortcut to the Swap token selector based on the opposite token's network.
- Keep the shortcut first in both compact and full network selectors, preserve its disabled state, remove duplicate network entries, and recalculate the `N+` count.
- Align the new compact and full-list UI with the Figma dimensions and include the separately committed translation update.

## Intent & Context

OK-57751 requires the combined Swap & Bridge token selector to make same-chain selection immediately visible. When selecting the From token, the To token network is featured; when selecting the To token, the From token network is featured. The behavior is limited to the Swap selector and follows the mobile and desktop Figma variants.

## Root Cause

The selector previously rendered fixed slices of the network list and had no presentation model for a featured network. This made it impossible to pin the opposite token network without duplicating it, keep a disabled bridge-only network visible, or calculate the remaining-network count from the rendered list. The compact `N+` item also used excessive horizontal padding and an implicit desktop height, which could truncate the count.

## Design Decisions

- Derive `sameChainNetwork` as selector presentation data instead of mutating the underlying network order.
- Enable the shortcut only for regular Swap token selection; Limit and Stock selectors retain their existing behavior.
- Pass featured-network metadata through the generic chain selector so the full list can render the badge, ordering, divider, and disabled state without introducing Swap dependencies into shared components.
- Keep disabled non-featured networks filtered from the full selector while allowing the disabled featured network to remain visible.
- Use the exact Figma logical dimensions for the 36px compact controls, 1x24 divider, 48px featured row, and 9px full-list divider region.

## Changes Detail

- Extend the network filter item with optional badge, responsive image-size, and text props.
- Build and test a Swap-only network view model that extracts the opposite token network and removes its duplicate.
- Update the compact network row and `N+` behavior for responsive mobile and desktop layouts.
- Extend the chain selector route and list rendering with an optional featured network.
- Update generated translations in a dedicated commit.

## Risk Assessment

- **Risk Level**: Medium
- **Risk Areas**: Swap From/To direction mapping, bridge-only disabled networks, responsive compact-row counts, and generic chain-selector rendering.

## Test plan

- [x] `yarn jest packages/kit/src/views/Swap/pages/modal/SwapTokenSelectModal.utils.test.ts --runInBand` (13 tests)
- [x] Verify the mobile compact row, complete `28+` label, featured full-list row, and divider in the iOS simulator.
- [ ] Verify the 640px desktop compact row and `N+` navigation at runtime.
- [ ] Verify Swap From/To direction changes and bridge-only disabled states on mobile and desktop.

[OK-57751]: https://onekeyhq.atlassian.net/browse/OK-57751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


zhaono1:
Superseded by #12436.

PR #12487, opened 2026-07-14
fix: swap history order id, network logo and no-provider button(OK-57251 OK-57730 OK-57545)

haicongliao:
[OK-57251](https://onekeyhq.atlassian.net/browse/OK-57251) [OK-57730](https://onekeyhq.atlassian.net/browse/OK-57730) [OK-57545](https://onekeyhq.atlassian.net/browse/OK-57545)

----------

## 概述 / Summary

Three swap fixes:

### OK-57251 — Remove Order ID from swap history detail
- Remove the "Order ID" row (and its `orderSupportUrl` deep link) from the generic swap history detail modal (`SwapHistoryDetailModal`). It affected SWFT / Cow Swap / 1inch Fusion / ChangeHero / Stock entries.
- Intentional product decision: the order id is not needed in this surface. Limit orders keep their Order ID in the separate `LimitOrderDetailModal` (untouched). Underlying `txInfo.orderId` data is retained (still used for history identity/dedup and status polling).

### OK-57730 — Unify swap network logo to circular
- The From/To network row in the swap panel rendered the network logo with a raw `<Image>`, so backend-configured square logos (e.g. robinhood) rendered square.
- Replaced with the shared `NetworkAvatarBase` (canonical circular network logo, same as other 16px call sites), using the `$4` size token so the load-failure fallback icon also sizes correctly.

### OK-57545 — Disable swap action button when no provider supports trade
- The action-state if-ladder let the later "Refresh quotes" override re-enable the button after the quote interval expired, even when the completed quote round had zero supporting providers. The button now stays disabled with "No provider supports trade".
- Extracted the predicate as `isSwapNoProviderSupportsTrade` (in `quoteProgress.ts`, unit-tested):
  - It only applies to the **current pair** (`!quoteResultNoMatch` gate) so a stale mismatched quote cannot lock the button out of its refresh recovery state.
  - `stepState.isRefreshQuote` excludes this state as well, so the rate-line refresh (auto interval + manual tap in `SwapRefreshButton`) stays alive as the recovery path — the state can self-heal if providers come back.

## 验证 / Verification

- `quoteProgress.test.ts`: 23 passed (5 new assertions for `isSwapNoProviderSupportsTrade`).
- Full-project type check (tsgo): 0 errors; oxlint type-aware: clean on all touched files.
- Desktop (CDP-driven): with a pair no provider supports, the action button renders disabled ("没有提供商支持此交易", `aria-disabled=true`) and stays disabled past the quote-interval expiry window (checked at T0 and T+70s); network logo renders circular in the From/To row.

[OK-57251]: https://onekeyhq.atlassian.net/browse/OK-57251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57730]: https://onekeyhq.atlassian.net/browse/OK-57730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-57545]: https://onekeyhq.atlassian.net/browse/OK-57545?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


haicongliao:
CleanShot 2026-07-14 at 16 25 13@2x 
CleanShot 2026-07-14 at 16 25 32@2x

zhaono1 (approved):

originalix (approved):

PR #12284, opened 2026-07-03
feat: unify OneKey ID and Keyless OAuth

sidmorizon:
## Summary
- Unify OneKey ID and Keyless Wallet OAuth sessions around the Keyless Supabase OAuth flow.
- Add OneKey ID OAuth login, legacy email fallback entry, legacy-to-OAuth binding prompts, profile identity display, and local Keyless upgrade handling.
- Harden logout, invalid-token cleanup, account mismatch handling, and migration behavior across OneKey ID and Keyless Wallet flows.

## Intent & Context
This change implements the Keyless OneKeyID merge flow: new OneKey ID login should use Google/Apple OAuth, while legacy email OTP remains available only as a low-priority fallback for historical accounts. Keyless Wallet and OneKey ID now share the OAuth session model, so upgrade scenarios must work for users who already have Keyless, OneKey ID, both, or neither.

The implementation also addresses migration edge cases discussed during review:
- avoid persisting OAuth access tokens in simpleDb and rely on Supabase session refresh instead;
- keep legacy email and Keyless OAuth sessions source-separated to avoid token/source mixups;
- support local Keyless OAuth reuse when logging into or binding OneKey ID;
- prevent Keyless wallet removal unless the user explicitly confirms the existing logout/remove-wallet flow.

## Root Cause
The old OneKey ID flow relied on legacy email OTP/Supabase behavior, while Keyless already had its own OAuth token/session storage. When OneKey ID moved to Google/Apple OAuth, treating these sessions as interchangeable caused several risks:
- stale access tokens could bypass Supabase refresh if copied into simpleDb;
- a standalone Keyless OAuth session could be misread as an active OneKey ID login;
- invalid-token cleanup could clear the wrong session after late responses;
- old-version users with both Keyless and OneKey ID needed explicit binding and mismatch recovery paths.

## Design Decisions
- Use `EPrimeAuthSessionSource` to separate `LegacyEmailSupabase` from `KeylessOAuth`.
- Read active Prime/OneKey ID tokens from the appropriate Supabase session instead of storing OAuth access tokens directly.
- Make OneKey ID OAuth login and legacy OAuth binding share local Keyless OAuth reuse logic.
- Keep the existing Google/Apple button layout, but when local Keyless exists only show the matching provider and explain that the linked Keyless account is used.
- Keep mismatch errors visible as toast and add a narrow inline path to the existing Keyless logout dialog for users who need to switch accounts.
[...]

PR #12493, opened 2026-07-14
fix: run OneKey ID OAuth login in expand tab from extension popup

sidmorizon:
Follow-up fix for a high-priority finding on #12284 (extension MV3 popup OAuth login).

## Problem

The unified OneKey ID login now shows `PrimeLoginOAuthDialog`, whose Google/Apple buttons call `chrome.identity.launchWebAuthFlow` in the **current UI runtime**:

```
loginOneKeyId (useOneKeyAuth)
  → Dialog.show(PrimeLoginOAuthDialog)
     → handleSocialLogin → getOAuthAccessToken → signInWithSocialLogin
        → performOAuthSignIn → OAuthPopup.open (ext) → chrome.identity.launchWebAuthFlow
```

The **browser-action popup** (the default extension entry, `chrome_v3.js: default_popup: 'ui-popup.html'`) is the one surface Chrome tears down on focus loss: the moment the auth window steals focus the popup document is destroyed, so the pending PKCE exchange / session persistence / `apiOAuthLogin` continuation dies with it. The user completes Google/Apple sign-in and **nothing happens, with no error**.

Every entry point that opens login from the popup is affected: user center (`MoreActionButton`), `OneKeyIdTabItem`, `OneKeyIdSettingsPage`, refer-a-friend, `PrimeDashboard`, `usePrimeRequirements`. The pre-existing keyless OAuth path avoided this because onboarding forced an expand tab via `useToOnBoardingPage`; the old `loginOneKeyId` used the popup-safe **email OTP** dialog. Switching to OAuth-in-popup is the regression.

**Scope — action popup only.** The side panel (docked, persistent) and the standalone window (a real browser window) are persistent contexts that are *not* destroyed when the auth window steals focus, so `launchWebAuthFlow` completes in place on those surfaces. The expand tab and native/web/desktop runtimes likewise survive. Only the action popup needs the redirect.

## Fix

Redirect OneKey ID login **from the action popup** to the persistent expand tab:

- `ServiceApp` gets a one-shot **in-memory** flag (never persisted) plus `openExtensionExpandTabForOneKeyIdLogin()` / `consumePendingOneKeyIdLoginExpandTabRedirect()`. The flag lives on the single shared MV3 bg service worker, so it is visible across every extension UI runtime — immune to URL/hash rewrite timing.
- `loginOneKeyId` (`useOneKeyAuth`) guards at the top: on `isExtensionUiPopup` it sets the flag + opens the expand tab, then returns. The popup auto-closes once the tab takes focus (no explicit `window.close()`). Because the guard is in `loginOneKeyId`, every entry point is covered centrally. The side panel and standalone window are deliberately left alone — they finish OAuth in place.
- `PrimeGlobalEffect` (mounted app-wide, including the expand tab) consumes the flag once on mount and re-invokes `loginOneKeyId` in the expand-tab runtime, where `launchWebAuthFlow` survives. Consumption is gated to `isExtensionUiExpandTab`, so the popup that set the flag never consumes its own flag; the read-then-clear in bg is atomic, so a double-mount can't double-trigger.

The email OTP fallback stays popup-safe and is untouched.

## Notes

- No DB schema / persisted fields touched (the flag is a plain in-memory service field), so no `LOCAL_DB_VERSION` bump or persistence audit impact.
- The same `launchWebAuthFlow`-in-popup limitation also applies to the legacy OAuth **bind** flow (`OneKeyIdLegacyOAuthBind`); that path is tracked as a separate finding and is not addressed here.

## Test plan

- [ ] Ext **action popup** → OneKey ID sign in → expand tab opens and the OAuth dialog resumes; Google/Apple login completes and updates logged-in state.
- [ ] **Side panel** → OneKey ID sign in → OAuth completes in place, no redirect (panel stays open).
- [ ] **Standalone window** → OneKey ID sign in → OAuth completes in place, no redirect.
- [ ] Expand tab directly → login shows in place, no redirect loop.
- [ ] Email OTP fallback still works from the popup.
- [ ] Native / web / desktop login unchanged.
- [ ] `yarn tsc:staged`, `yarn lint:staged`.

### DIFF
--- a/packages/kit-bg/src/services/ServiceApp.ts
+++ b/packages/kit-bg/src/services/ServiceApp.ts
@@ -363,6 +363,26 @@ class ServiceApp extends ServiceBase {
     return extUtils.openExpandTab(routeInfo);
   }
 
+  // One-shot, in-memory only (never persisted). Set by an ephemeral extension
+  // UI surface (action popup / side panel / standalone window) that cannot host
+  // the OneKey ID OAuth flow, and consumed once by the persistent expand tab so
+  // it can resume the login there. Lives on the single shared bg service
+  // worker, so it is visible across every extension UI runtime.
+  private pendingOneKeyIdLoginExpandTabRedirect = false;
+
+  @backgroundMethod()
+  async openExtensionExpandTabForOneKeyIdLogin() {
+    this.pendingOneKeyIdLoginExpandTabRedirect = true;
+    return extUtils.openExpandTab({ routes: '' });
+  }
+
+  @backgroundMethod()
+  async consumePendingOneKeyIdLoginExpandTabRedirect(): Promise<boolean> {
+    const pending = this.pendingOneKeyIdLoginExpandTabRedirect;
+    this.pendingOneKeyIdLoginExpandTabRedirect = false;
+    return pending;
+  }
+
   @backgroundMethod()
   async openExtensionMarketTokenDetail(params: {
     tokenAddress: string;
--- a/packages/kit/src/components/OneKeyAuth/useOneKeyAuth.tsx
+++ b/packages/kit/src/components/OneKeyAuth/useOneKeyAuth.tsx
@@ -16,6 +16,7 @@ import {
 } from '@onekeyhq/shared/src/keylessWallet/keylessWalletTypes';
 import { ETranslations } from '@onekeyhq/shared/src/locale';
 import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
+import platformEnv from '@onekeyhq/shared/src/platformEnv';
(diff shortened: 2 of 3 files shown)


sidmorizon:
Closing as redundant. The base branch already fixes this in `91ae5267a` ("fix: run ext popup OneKey ID OAuth flows in expand tab") via a dedicated `extOneKeyIdAuthExpandTab.ts` helper + `ExtOneKeyIdAuthOnMount` consumer. That implementation uses the same action-popup-only scoping (side panel / standalone window run in place) and additionally covers the legacy OAuth **bind** flow, which this PR had left out. No action needed here.

---

PR #12512, opened 2026-07-15
fix: gate keyless teardown and invalid-token events on auth-state generation

sidmorizon:
Follow-up to #12511, targeting #12507's head branch. Addresses the two remaining race findings plus the write-path half of the sealed-codec comment.

## Auth-state commit generation

Both remaining findings share one root cause — a destructive cleanup decides on state snapshotted before an `await`, and a **fresh KeylessOAuth login that commits during that await is indistinguishable from the state being torn down** (the source re-reads as `KeylessOAuth` either way):

- **Codex P2 (`ServiceKeylessWallet.ts:3642`)**: `clearKeylessAuthSessionAndLoginState` awaits `clearKeylessAuthSession()` (signOut can hang on the network); a login committing in that window is then wiped by the source-only in-lock recheck, and the unversioned `KeylessAuthSessionCleared` broadcast makes main sign out the fresh client copy — main's clients persist through the SHARED session storage, so the fresh login's persisted session is deleted too.
- **originalix (`PrimeGlobalEffect.tsx:308`)**: a stale `PrimeLoginInvalidToken` event still propagating bg → main after a fresh login unconditionally signs out the current keyless client, deleting the just-persisted session; the trailing guarded atom reset cannot restore deleted credentials.

Mechanism (as suggested in the review):
- `simpleDb.prime.authStateGeneration` — monotonic counter bumped by every `setAuthSessionSource` commit (login, bind switch, legacy self-heal). Clears intentionally do NOT bump: gating only needs to detect commits, and a redundant clear is idempotent.
- `clearOneKeyIdAuthStateIfSourceStillKeylessOAuth` accepts `expectedAuthStateGeneration` (snapshotted before the session clear) and skips in-lock with `generationChanged: true` when a commit landed in between; `clearKeylessAuthSessionAndLoginState` then also suppresses the `KeylessAuthSessionCleared` broadcast.
- `handlePrimeLoginInvalidToken` returns the in-lock generation; both emit sites (ServiceBase interceptor, `throwIfAllPrimeUserInfoRequestsFailedByInvalidTokenError`) carry it in the `PrimeLoginInvalidToken` payload; the `PrimeGlobalEffect` handler compares it against the current generation and skips the whole stale event on mismatch. Payload change is safe on split-runtime targets: `main`/`bg` bundles ship version-locked.
- Generation reads/writes stay inside `authStateWriteMutex` sections that only touch simpleDb/atom — no network I/O added under the lock, per the documented lock policy (the Codex-suggested alternative, serializing the signOut under the mutex, would violate it).

## Write-path sealed-value downgrade (originalix, `sealedValueCodec.ts` comment)

Kept the plaintext fallback on transient seal failure **deliberately**, now documented in code: rejecting the write would lose a just-rotated refresh token that auth-js persists through this exact `setItem` (the old token is already consumed server-side; past GoTrue's ~10s reuse window the family is revoked → guaranteed forced re-login). Mitigation added instead: after a plaintext fallback, `setItem` immediately triggers the opportunistic reseal (same self-guarded `resealLegacyPlainValue` used by the read path — no-op in Main runtimes, re-reads before writing so it can never clobber a newer rotated session), shrinking the plaintext window from "next read" to roughly one retry round-trip.

## Test plan
- [x] `yarn agent:check --profile commit` — lint-worktree-ts / lint-staged / tsc-staged all PASS
- [x] `jest ServicePrime.test.ts SimpleDbEntityPrime.test.ts SupabaseStorage.test.ts` — 40/40, new coverage:
  - generation bumps on every `setAuthSessionSource`, not on `clearAuthTokens`; defaults to 0 for pre-upgrade data
  - keyless-source clear skips with `generationChanged` when a login committed after the caller snapshot; source-only behavior unchanged without a snapshot
  - `PrimeLoginInvalidToken` emit payloads carry the in-lock generation (interceptor + HTTP-200 body-code paths)
  - transient seal failure on write falls back to plaintext (write never rejected) and reseals immediately after device-key recovery

---
_Generated by [Claude Code](https://claude.ai/code/session_01EJTPZLp5xj28qEQqbzJQeB)_

### DIFF
--- a/packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityPrime.test.ts
+++ b/packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityPrime.test.ts
@@ -74,6 +74,54 @@ describe('SimpleDbEntityPrime.getEffectiveAuthSessionSource', () => {
   });
 });
 
+describe('SimpleDbEntityPrime.authStateGeneration', () => {
+  test('defaults to 0 for pre-upgrade data', async () => {
+    const entity = new SimpleDbEntityPrime();
+    jest.spyOn(entity, 'getRawData').mockResolvedValue({});
+
+    await expect(entity.getAuthStateGeneration()).resolves.toBe(0);
+  });
+
+  test('setAuthSessionSource bumps the generation on every commit', async () => {
+    const entity = new SimpleDbEntityPrime();
+    let persisted: Record<string, unknown> = { authStateGeneration: 2 };
+    jest.spyOn(entity, 'setRawData').mockImplementation((async (
+      updater: (rawData: Record<string, unknown>) => Record<string, unknown>,
+    ) => {
+      persisted = updater(persisted);
+      return persisted;
+    }) as never);
+
+    await entity.setAuthSessionSource(EPrimeAuthSessionSource.KeylessOAuth);
+    expect(persisted.authSessionSource).toBe(
+      EPrimeAuthSessionSource.KeylessOAuth,
+    );
+    expect(persisted.authStateGeneration).toBe(3);
+
+    // A bind switch (KeylessOAuth while already logged in) is also a commit
+    // and must advance the epoch again.
+    await entity.setAuthSessionSource(
+      EPrimeAuthSessionSource.LegacyEmailSupabase,
+    );
+    expect(persisted.authStateGeneration).toBe(4);
+  });
+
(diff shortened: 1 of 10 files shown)


sidmorizon:
Commits `5bfcc41e96` + `c9d21b63b9` were pushed directly to `prague13/merge-onekeyid-keyless-accounts` (#12507) per the author's request, so this PR's diff is now empty — closing.

---

PR #12521, opened 2026-07-16
fix: surface keyless OAuth session persist failures in OneKey ID login

sidmorizon:
## Background

On iOS, OneKey ID Google/Apple OAuth login fails with a generic "unknown error" toast even though the OAuth round-trip itself succeeds. User log analysis showed the actual failure chain (100% reproduction):

1. `POST https://auth.onekey.so/auth/v1/token?grant_type=id_token` → **200** (native sign-in + token exchange OK)
2. `GET https://auth.onekey.so/auth/v1/user` → **403** (issued internally by `auth.setSession()` while persisting the keyless session)
3. `setSession()` returns the error instead of throwing; `persistKeylessOAuthSession` discarded it → keyless session slot left empty, flow continues
4. `POST /prime/v1/account/oauth/login` → **200** (server now considers the user logged in)
5. Local commit fails: `getActiveAuthToken()` reads an empty slot → `apiOAuthLogin ERROR: Active auth token not found` → generic "unknown error" toast; server logged in, client rolled back

Field verification with this branch confirmed the 403 body is a **non-JSON interception page** (surfaced as `Supabase non-JSON HTTP 403 error response treated as network failure...`), i.e. an intermediary in front of GoTrue — not a GoTrue verdict. The server/network-side root cause is under separate investigation; this PR fixes the client-side amplifiers and adds the diagnostics to pinpoint which intermediary it is.

## Changes

**Commit 1 — fail fast + real error surfacing**
- `useSupabaseAuth.tsx` — `persistKeylessOAuthSession`: check the `{ data, error }` result of `auth.setSession()` and throw `OneKeyLocalError` with the underlying message plus HTTP `status`/`code`, instead of silently leaving the session slot empty.
- `ServicePrime.tsx` — `apiOAuthLogin`: before the server login POST, verify a keyless session is actually persisted in the shared storage slot (pure local read via `readPersistedAccessTokenBySessionSource`). If empty, log and throw before any server-side state changes — eliminating the "server logged in / client logged out" inconsistency for every caller.
- `oneKeyIdLoginToastUtils.ts` — `showOneKeyIdLoginFailedToast` keeps the generic title (the dedicated `id_login_failed` i18n key is still missing) but surfaces the underlying error message as the toast body.

**Commit 2 — make the failure diagnosable from exported logs alone**
- New prime logger scene methods (`LogToLocal` + `LogToServer`): `onekeyIdSessionPersistFailed` (logged at the persist-failure source) and `onekeyIdLoginFailedToast` (mirrors the fallback toast body), so exported `app-latest.log` captures the real reason without needing a toast screenshot.
- `supabaseClientUtils.ts` — `sessionPreservingSupabaseFetch` now appends an interception fingerprint to the non-JSON error it throws: `server` / `content-type` / `cf-ray` / `cf-mitigated` response headers plus the first 160 chars of the error-page body (headers and error-page prefix only — no request data, no tokens). Since the auth host sits behind Cloudflare, `cf-ray=none` proves a local proxy/VPN interception, `cf-mitigated`/CF-branded body proves a Cloudflare block, and `cf-ray` + non-Cloudflare `server` proves an origin gateway rejection.

All existing call sites of `persistKeylessOAuthSession` (Prime login dialog, legacy OAuth bind, keyless wallet create/verify flows) already depend on the session actually being persisted for their next step, and each host either wraps with `withErrorAutoToast` or has an explicit catch — so the new throw surfaces through existing error paths.

## Verification

- `yarn agent:check --profile commit` passes on both commits (lint + tsc).
- Field-tested on the affected iOS device: the failing login now stops before `POST /prime/v1/account/oauth/login` and the toast/body shows the real reason (`Failed to persist Keyless OAuth session: Supabase non-JSON HTTP 403 ...`), confirming the fix chain works end to end.

https://claude.ai/code/session_01LAPUaTZurR9GccUg8o2Ni2


originalix (approved):

zhaono1 (approved):

packages/kit/src/views/Prime/components/oneKeyIdLoginToastUtils.ts
```diff
@@ -44,5 +60,6 @@ export function showOneKeyIdLoginFailedToast({
     title: intl.formatMessage({
       id: ETranslations.global_unknown_error_retry_message,
     }),
+    message: errorMessage,
```
sidmorizon, line 63: # [🟡 P2] 登录失败 toast 直接把内部拦截指纹展示给终端用户

## 问题

`message: errorMessage` 把底层错误原文原样作为 toast 正文展示。对于本 PR 主目标的 403 场景，`errorMessage` 来自 `sessionPreservingSupabaseFetch` 抛出的字符串，内部拼接了 `buildInterceptedResponseFingerprint` 的结果，例如 `Failed to persist Keyless OAuth session: Supabase non-JSON HTTP 403 ... [server=cloudflare content-type=text/html cf-ray=<id> cf-mitigated=none body=cloudflare-page]`。也就是说 `cf-ray`（Cloudflare 请求追踪 ID）、`server` 头、`body=<label>` 这些纯内部诊断标识会作为一段未翻译的原始英文串直接弹给所有终端用户。

而这份完全相同的诊断信息，本 PR 已经通过新增的 `onekeyIdSessionPersistFailed` / `onekeyIdLoginFailedToast`（`@LogToLocal()` + `@LogToServer()`）写入了导出日志与服务端分析，用户侧 toast 再镜像一份并不带来额外的可诊断价值。

## 影响

面向全球终端用户的失败提示里出现 `cf-ray=<id>`、`server=...` 这类基础设施标识和 `body=cloudflare-page` 这类内部分类标签，既是可读性/体验问题，也是不必要的内部实现细节外泄；且该正文无法走 i18n，非英语用户只能看到一段技术噪音。诊断价值已被新日志覆盖，用户侧无需承担这段原文。

## 建议

给用户展示友好的、可翻译的通用失败原因，把带指纹的原始串留在日志侧即可；如需现场排障，可仅在 `enableKeylessDebugInfo` 打开时才把原始 `errorMessage` 落到 toast 正文，例如：

```ts
Toast.error({
  title: intl.formatMessage({ id: ETranslations.global_unknown_error_retry_message }),
  message: enableKeylessDebugInfo ? errorMessage : undefined,
});
```

（`enableKeylessDebugInfo` 已在 keyless 相关代码中使用，可透传进来。）

[Claude session ↗](https://claude.ai/code/session_01MvBWrJhz6ebpu8ZrsFZ5iP)

---

PR #12526, opened 2026-07-16
fix: harden OneKey ID OAuth login guard, surfacing and intermediary diagnostics

sidmorizon:
## Background

Follow-up to #12521 (merged), which surfaced the iOS OneKey ID OAuth "unknown error" (a keyless-session persist failure caused by a 403 on Supabase's internal `GET /auth/v1/user`). A deep review of that merged change found one CI-blocker plus several correctness/robustness gaps; field analysis additionally confirmed the web OAuth popup flow loses its authorization code to an SPA URL rewrite. This PR addresses all of them.

> Note: #12521 was squash-merged from a snapshot that predated its own lint fix, so `x` currently fails the full-eslint CI job (`no-use-before-define` in `supabaseClientUtils.ts`). This PR restores that fix as part of the restructure below.

### Commit 1 — harden guard, surfacing, and diagnostics

**Guard false-negative (highest severity) — `ServicePrime.tsx`, `primeAuthSessionAccess.ts`**
The `apiOAuthLogin` pre-POST keyless-slot guard could abort a *successful* login and make callers tear down (sign out + server-revoke) the just-persisted session. Two causes, both fixed:
- It read through `readPersistedAccessTokenBySessionSource`, whose contract degrades a *transiently-unreadable* slot to `''`. Added `readPersistedAccessTokenBySessionSourceStrict`, which distinguishes a definitely-empty slot (`''`) from a transient/unparseable read (rethrows). The guard now aborts only on a definitely-empty slot; on uncertainty it proceeds (recoverable) rather than destroying a possibly-valid session.
- It ran before `clearCachedAuthToken` and could read a ≤30s-stale memoized empty slot on split runtimes. The guard now runs after the cache clear and forces a fresh local read via `clearSupabaseStorageLocalCache`.
- The guard/logging is extracted into a shared `assertKeylessSessionPersistedBeforeLogin` helper and now logs the dedicated `onekeyIdSessionPersistFailed` event (was the catch-all `onekeyIdAtomNotLoggedIn`).

**Irreversible bind unguarded — `ServicePrime.tsx`**
`apiBindLegacyOneKeyIdOAuth` performs an irreversible server-side identity bind then commits KeylessOAuth with no pre-POST slot check. It now calls the shared guard, so an empty slot fails before the bind instead of after.

**Silent login failure — `useSupabaseAuth.tsx`**
`persistKeylessOAuthSession` threw an error with `autoToast` unset, so flows without their own catch (e.g. the verify-PIN recovery flow) died with no toast. The error now carries `autoToast: true` (surfaced by the global handler) and `httpStatusCode`; the fallback toast marks it triggered so no duplicate toast appears on the dialog rethrow path.

**Intermediary-diagnostics accuracy — `supabaseClientUtils.ts`**
- The fingerprint is now attached to the transient-status branch too (Cloudflare 503/429 pages were previously unfingerprinted).
- Body is read once (single clone) and classified from the already-read text.
- Cloudflare body detection is anchored to CF-specific markers instead of the bare word `cloudflare` (a page merely loading a `cdnjs.cloudflare.com` asset no longer mislabels as `cloudflare-page`).
- A `platform` field is emitted so log triage knows that on browser-CORS runtimes `cf-ray`/`server` absence means "CORS-hidden", not "local proxy".
- A non-OK response with a parseable JSON body still passes through unchanged (definitive GoTrue verdicts reach callers); session-preservation behavior for transient/non-JSON responses is unchanged.
- Helper declaration order fixed (the `no-use-before-define` CI blocker).

**Duplicate server log & reason extraction — `oneKeyIdLoginToastUtils.ts`**
The fallback toast skips its `@LogToServer` event when the source already logged the reason (persist failures were double-counted), and extracts the reason via the shared `toPlainErrorObject` so unstructured throws no longer collapse to `unknown`.

**Web popup swallow — `OAuthPopup.tsx`**
The cross-origin `popup.location.href` read is narrowed so a real persist failure rejects the flow immediately instead of being swallowed as "continue polling" (which hung one poll interval and retried persistence).

### Commit 2 — register a real web route for the OAuth popup callback

Supabase redirects the web OAuth popup to `${origin}/oauth_callback_web/?code=...`, but no route existed for that path: it fell into the `NotFound: '*'` wildcard and `getPathFromState` rewrote the URL to `/`, stripping `?code=` before the opener's 500ms poll could read it — the login then failed as a cancel. (A static callback page is not viable here: the service worker serves the SPA app shell for every navigation, and the build/hosting layers have no static-page pipeline for this path.)

- New web-only root screen `ERootRoutes.OAuthCallbackWeb` (`views/OAuthCallbackWeb`, a bare spinner — the opener owns the PKCE exchange and closes the popup).
- Registered via `buildOAuthCallbackWebRouter()` (mirrors `buildPermissionRouter`), gated on `platformEnv.isWeb` so the extension/desktop/native routing surface is unchanged; `rewrite: OAUTH_CALLBACK_WEB_PATH, exact: true` makes it match before the `NotFound` wildcard.
- allowList rule (`showUrl` + `showParams`) so `getPathFromState` keeps the path and query; `onGetStateFromPath` resolves the callback path before the `?r=` referral / WebDappMode rewrites; a defensive guard in `getPathFromState` keeps the URL frozen while the popup sits on the callback path with a `code=` query, so no unrelated navigation-state change can race the opener's poll.

## Verification

- `yarn agent:check --profile commit` passes on both commits (lint-worktree-ts + tsc).
- Direct eslint on all changed files passes (incl. the previously-failing `supabaseClientUtils.ts`).
- Targeted tests pass: `SupabaseStorage`, `SimpleDbEntityPrime`, `ServiceKeylessWallet`, `customUA` (108 tests).
- The fetch-wrapper restructure is behavior-equivalent for session preservation (transient → masked; non-transient JSON → pass-through; non-transient non-JSON → masked); only diagnostics and body-read count change.
- Web callback route verified with an executable sanity check against the real `@react-navigation/core` `getStateFromPath`/`getPathFromState`: `/oauth_callback_web/?code=abc&onekey_oauth_state=x` (with and without trailing slash) resolves to the new screen with both query params, unknown paths still resolve to `NotFound`, and the re-synced browser path keeps `code=abc`.

https://claude.ai/code/session_01LAPUaTZurR9GccUg8o2Ni2


packages/kit/src/views/Prime/components/oneKeyIdLoginToastUtils.ts
```diff
@@ -62,4 +101,13 @@ export function showOneKeyIdLoginFailedToast({
     }),
     message: errorMessage,
   });
+
+  // This fallback toast has now surfaced the error. Mark it so the global
+  // unhandledrejection handler (autoToast errors flow there when a caller
+  // rethrows into a `void`ed promise) does not show a second toast for it.
+  if (err) {
```
sidmorizon: # [🟠 P1] 字符串类型的 error 会让 `showOneKeyIdLoginFailedToast` 在末尾抛出 `TypeError`

## 问题

本 PR 新增的收尾标记块只用 `if (err)` 做真值判断，随后直接在 `err` 上写属性：

```ts
if (err) {
  (err as IOneKeyError & { $$autoToastErrorTriggered?: boolean })
    .$$autoToastErrorTriggered = true;
}
```

ES module 恒为严格模式，向字符串原始值写属性会抛 `Cannot create property '$$autoToastErrorTriggered' on string`。而本 PR 同时新增的 `getLoginFailureReason` 明确处理了 `typeof error === 'string'` 分支，说明字符串 error 是被预期会进入本函数的输入。此外同文件的 `markOneKeyIdFailureServerLogged` 已经用 `error && typeof error === 'object'` 做了防护 —— 这里漏掉了同样的判断。

## 影响

当 `error` 为非空字符串时：`shouldSkipOneKeyIdLoginFailedToast` 不会跳过（字符串上取 `.autoToast`/`$$autoToastErrorTriggered` 均为 `undefined`），`Toast.error` 先弹出，随后这一行抛 `TypeError`。该异常会从 `showOneKeyIdLoginFailedToast` 冒泡，多数调用点是在 `catch` 里调用它做兜底提示，于是原本已被处理的登录失败又变成一个新的抛出，可能中断后续清理、把 UI 卡在 loading，或触发 `unhandledrejection`。触发条件较窄（需字符串 error），但一旦命中即为确定性崩溃。

## 建议

与 `markOneKeyIdFailureServerLogged` 保持一致，收窄到对象再写标记：

```ts
if (err && typeof err === 'object') {
  (err as IOneKeyError & { $$autoToastErrorTriggered?: boolean })
    .$$autoToastErrorTriggered = true;
}
```

[Claude session ↗](https://claude.ai/code/session_01FnPdCTyrPgL3Emrxq2QWgt)

---

sidmorizon: Fixed in 0bb27157 — narrowed to `err && typeof err === 'object'` before writing the marker, matching the `markOneKeyIdFailureServerLogged` guard.

---

packages/kit-bg/src/services/ServicePrime/ServicePrime.tsx
```diff
@@ -722,6 +724,42 @@ class ServicePrime extends ServiceBase {
...
+    // invalidation after the UI-side persist is best-effort.
+    clearSupabaseStorageLocalCache();
+    let isSlotDefinitelyEmpty = false;
+    try {
+      const persistedKeylessAccessToken =
+        await readPersistedAccessTokenBySessionSourceStrict(
+          EPrimeAuthSessionSource.KeylessOAuth,
+        );
+      isSlotDefinitelyEmpty = !persistedKeylessAccessToken;
+    } catch {
```
originalix: 当 Keyless session key 存在但内容无法 JSON 解析时，`readPersistedAccessTokenBySessionSourceStrict()` 会抛错，而这里的无差别 `catch` 会把它当作“可能仍有效”并继续执行。随后 `/oauth/login` 或不可逆的 `/oauth/bind` 可能已经在服务端成功，但本地 commit 会再次读取同一损坏会话并失败，造成服务端身份状态与本地登录状态分叉。

请区分错误类型：JSON 解析失败应在 POST 前直接阻断；仅对 `SupabaseStorageTransientError` 保留可重试语义并原样抛给调用方，同时让调用方保留临时 session，不要继续执行服务端 mutation。

packages/shared/src/utils/supabaseClientUtils.ts
```diff
@@ -45,89 +45,138 @@ const storage = supabaseStorageInstance;
...
+  const response = await fetch(input, init);
+  if (response.ok) {
+    return response;
+  }
+  // Read the error body ONCE from a clone; the original stays unread so
+  // auth-js can still consume it on the JSON pass-through path below.
+  let bodyText: string | undefined;
   try {
-    text = (await response.clone().text()).slice(0, 2048);
+    bodyText = await response.clone().text();
```
originalix: 当代理或 WAF 返回 408、429、5xx，且 body 很大、持续流式或迟迟不结束时，这里会在判断 status 前先用 `text()` 读取完整 clone；后面的 `slice(0, 2048)` 只限制分类输入，并不限制实际下载与内存占用。结果是本应立即转成 retryable fetch error 的请求会长时间挂起，甚至在认证路径造成额外内存压力。

请先按 status 处理 transient 响应；如果必须采集 body 指纹，使用带字节上限和超时的读取，只消费前 2KB。

packages/kit-bg/src/services/ServicePrime/ServicePrime.tsx
sidmorizon: Fixed in 236c4629. `readPersistedAccessTokenBySessionSourceStrict` now returns a discriminated result (`empty` / `corrupt` / `ok`) instead of throwing on unparseable JSON: both `empty` and `corrupt` deterministically block BEFORE the login/bind POST (with distinct log reasons), while `SupabaseStorageTransientError` propagates as-is without running the server mutation — and `isTransientNetworkLikeError` now recognizes it by name, so callers (PrimeLoginOAuthDialog / useKeylessWallet) keep the temp session instead of tearing it down.

---

packages/shared/src/utils/supabaseClientUtils.ts
sidmorizon: Fixed in 236c4629. Transient statuses (408/429/5xx) are now classified by status FIRST, before any body read; the fingerprint on that branch uses `readErrorBodyPrefixBestEffort` — a stream-reader read capped at 2KB with a 1.5s deadline (then cancels the reader), falling back to a timeout-raced `text()` on React Native where fetch exposes no ReadableStream. The non-transient branch keeps the full single read on purpose: the JSON-verdict check needs body fidelity (a truncated read could misclassify a real GoTrue verdict as an intermediary page), matching the pre-existing behavior of that branch.

---

packages/kit-bg/src/services/ServicePrime/ServicePrime.tsx
```diff
@@ -1008,6 +1030,14 @@ class ServicePrime extends ServiceBase {
         );
       }
 
+      // Same fail-fast as apiOAuthLogin — and it matters MORE here: the bind
+      // POST below is an irreversible server-side identity bind, so a
+      // subsequent KeylessOAuth commit failing on an empty slot would leave the
+      // account bound on the server while the client rolls its login state back.
+      await this.assertKeylessSessionPersistedBeforeLogin(
```
originalix, line 1051: 当这个 guard 因设备密钥或存储暂时不可用而抛出 `SupabaseStorageTransientError` 时，错误会被正确分类为 transient；但 `OneKeyIdLegacyOAuthBind` 的绑定和切换账户 catch 目前仍会对所有失败调用 `clearOAuthSignInTempSession()`。这样刚成功持久化、仍可恢复的 Keyless session 会被删除，切换账户路径还可能在旧账号已退出后把用户留在登出状态。

请让这两个调用方与普通登录流程保持一致，仅在 `!isTransientNetworkLikeError(error)` 时清理临时 session，transient storage error 应保留 session 供重试。

sidmorizon: Fixed in fe850b1d. Both catch blocks in `OneKeyIdLegacyOAuthBind` (bind and switch-account) now gate the cleanup with `didUseOAuthSignIn && !isTransientNetworkLikeError(error)`, matching PrimeLoginOAuthDialog — transient failures (including the guard's `SupabaseStorageTransientError`) keep the just-persisted session for retry. The user-declined-switch path (dialog cancel) still clears unconditionally, since that's an explicit user decision rather than an error.

---

originalix (approved):

PR #12529, opened 2026-07-16
fix: harden ip table failover and config verification under extreme networks

originalix:
## Background

This PR addresses request reliability issues observed under some network conditions. Log analysis confirmed two independent root causes:

1. **The selection algorithm kept traffic on an unavailable domain path.** After 23 consecutive real-request failures on the direct domain, the triggered speed test still judged the domain "competitive" — the health probe succeeded (813ms) and the IPs were only ~25% faster (below the 30% threshold) — so it re-selected the failing domain, forming a fail → probe → re-pin loop.
2. **Config signature verification failed for 18h+ with no way to diagnose.** The background runtime was refused when loading the `@ethersproject` segment (`seg:nm.@ethersproject` is owned by `main` in the split-bundle manifest); a catch-all swallowed the error and reported it as "signature verification failed", misleading the investigation toward the CDN. (The segment-ownership fix lands in a separate branch; this PR covers the observability and resilience side.)

## Changes (by commit)

| Commit | Content |
|---|---|
| Types & constants | Structured decision/verify types; failover threshold (3) and TTL (5min) constants |
| Pure endpoint decision | `decideEndpoint`: reachability beats latency (real-traffic failures override the 30% threshold), symmetric 30% hysteresis against flapping; includes an incident-scenario regression test |
| Selection rewiring + fast failover | Service adopts the decision function; after 3 consecutive domain failures it switches to the last-best IP immediately (no waiting for a speed test round); fixes the `currentSelection === undefined` deadlock that could block speed tests forever; persists lastBestIp on every speed test |
| Speed-test single-flight | Per-domain coalescing + timer dedup, eliminating the ~20 concurrent speed-test storm seen in the incident log |
| Adapter fail-open | After 3 consecutive transport-level network errors, hosts of that root domain resolve to a builtin/last-best IP for a 5-minute window — **covers the main runtime and the cold-start `no_config` gap** (60 occurrences in the incident log); HTTP error responses don't count; **failing requests are never replayed** (wallet POST double-send risk); recovery on first domain success |
| Structured verification | Staged failure reasons (`verifier_load_failed` / `signer_mismatch` / `malformed_signature` / `missing_signature`) — this log line alone would have pinpointed the incident on day one; canonicalization now uses an allowlist of signed fields (extra server-side metadata no longer breaks verification — validated against the real signer); shape guard on the CDN response |
| LKG + rollback protection | Persists version/generated_at/payloadHash of the last verified config; rejects legitimately-signed but older configs |
| CDN fetch via IP-capable client | `config.onekeycn.com` is fetched through the adapter-backed client instead of plain axios, giving the config-delivery path the same network resilience as business traffic |
| Server metrics | `endpointSwitched` / `adapterFailover` / `configVerifyFailed` LogToServer events (no PII), so regional failover and verify-failure rates are visible without asking users to upload logs |
| Kill switch | `disableIpTableFailover` dev setting + Dev Settings UI entry — the new behaviors can be turned off in production without a rollback |

## Testing

- 82 tests / 7 suites green (29 new: 11 decision function, 6 fail-open, 12 verification/guard/hash)
- Incident regression cases included: `incident regression: domain failing real traffic beats 30% threshold`, `verifier import failure -> verifier_load_failed (split-bundle incident)`
- `yarn agent:check --profile pr` local checks all pass

## Out of scope (tracked separately)

- Split-bundle segment runtime ownership fix (`@ethersproject` + the mirrored sentry issue) — separate task/PR
- Native log-upload channel (expo-file-system UploadTask bypasses the adapter; the SNI module doesn't support file streams) — separate task
- 30s API timeout tuning — server-side concern, intentionally untouched

## Risk & verification suggestions

The largest risk surface is the adapter fail-open (shared by appApiClient / analytics / hardware configFetcher). Constraints: only transport-level errors count, threshold 3 + 5min TTL, exits on first success, kill switch available. Suggested QA: regress the core request paths both with the direct domain blackholed (hosts entry for wallet.onekeycn.com) and on a normal network to confirm no behavior change in the healthy case.


packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +154,165 @@ async function getMappedDomainForIpLookup(
...
+
+const adapterFailoverStates = new Map<string, IAdapterFailoverState>();
+
+/** Test-only helper: clears fail-open state and the selection memo cache. */
+export function resetAdapterFailoverStatesForTesting(): void {
+  adapterFailoverStates.clear();
+  getSelectedIpForHost.clear();
+}
+
+function isNetworkLevelError(error: unknown): boolean {
```
originalix: 当请求被 AbortController 或 Axios 主动取消时，错误通常没有 `response`，这里会把 `ERR_CANCELED` 也累计为传输层失败。连续三次页面卸载或请求去重就会打开五分钟 fail-open，导致网络正常时后续请求仍被改走 builtin IP。

请先排除 `axios.isCancel(error)` / `ERR_CANCELED`，并仅对白名单中的超时、DNS、连接重置等明确网络错误累计计数。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -371,6 +399,28 @@ class ServiceIpTable extends ServiceBase {
...
+      // Rollback protection: a valid signature is not enough — refuse
+      // configs older than what we already verified, so a stale (but
+      // legitimately signed) file cannot displace a newer one.
+      const remoteGeneratedAtMs = Date.parse(remoteConfig.generated_at);
+      const lastVerifiedGeneratedAtMs = lastVerified
+        ? Date.parse(lastVerified.generatedAt)
+        : Number.NaN;
+      const isVersionRegression =
+        remoteConfig.version < localConfig.version ||
+        (lastVerified !== undefined &&
```
originalix: 升级前已经持久化的 IP Table 配置没有 `runtime.lastVerified`。在这种状态下，如果 CDN 返回同一 `version` 但更早 `generated_at` 的已签名配置，这个条件不会进行时间比较，随后还会把更早时间保存成新的 `lastVerified`，使回滚保护的单调锚点后移。

请在 `lastVerified` 缺失时回退比较 `localConfig.generated_at`，并对不可解析的时间拒绝更新。

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +154,165 @@ async function getMappedDomainForIpLookup(
...
+  if ('response' in error && (error as { response?: unknown }).response) {
+    return false;
+  }
+  return true;
+}
+
+async function isFailoverDisabledByDevSettings(): Promise<boolean> {
+  try {
+    const devSettings = await requestHelper.getDevSettingsPersistAtom();
+    return !!devSettings.settings?.disableIpTableFailover;
```
originalix, line 316: 这个开关只在累计新失败时读取；fail-open 一旦激活，后续选择仍只检查 `isFailoverActive()`，所以运行中打开 kill switch 仍会继续走 IP 直到 TTL 到期。另外这里忽略了 `devSettings.enabled`，而 `ServiceIpTable` 会检查它，split-runtime 的 main 与 bg 可能得到相反结论。

请让端点选择统一使用包含 `devSettings.enabled` 的开关判断，并在禁用时清理 failover state 和 memoized selection。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -484,13 +536,43 @@ class ServiceIpTable extends ServiceBase {
...
+  async selectBestEndpointForDomain(
+    domain: string,
+    opts?: { trigger?: 'domain_failure' | 'ip_failure' | 'periodic' },
+  ): Promise<void> {
+    const inflight = this.speedTestInFlight.get(domain);
+    if (inflight) {
+      defaultLogger.ipTable.request.info({
+        info: `[IpTable] Speed test already in flight for ${domain}, coalescing`,
+      });
+      return inflight;
```
sidmorizon: # 🟠 P1 单飞合并丢弃更强的 `domain_failure` 触发信号

## 问题

`selectBestEndpointForDomain` 的 per-domain 单飞逻辑在已有 in-flight 时直接 `return inflight`，完全忽略新调用携带的 `opts.trigger`。当一个 `periodic` 或 `ip_failure` 触发的测速正在进行时，随后到达的 `domain_failure` 调用被合并进去，其 `domainFailingRealTraffic: true` 语义被丢弃。被合并的那一轮 `decideEndpoint` 收到的是 `domainFailingRealTraffic=false`，于是走普通 30% 阈值分支。

这恰好命中 PR 注释里引用的 2026-07-16 事故形态：域名健康探测成功（`domainLatency` 有限），但真实流量在直连域名上持续失败。此时只有 `domain_failure` 触发的"任何可达 IP 都胜出"覆盖逻辑能救场，而它被合并稀释掉了。

## 影响

冷启动 + 墙内网络是最坏也是最常见的场景：此时 `lastBestIp` 尚未被任何成功测速填充，`reportRequestFailure` 里的快速故障切换（依赖 `lastBestIp`）无法触发，唯一的升级路径就是 `domain_failure` 测速的 reachability-first 覆盖。若首轮 `runFullSpeedTest`（`trigger` 为 `periodic`）此刻在飞，逃逸恰好被吞掉，`decideEndpoint` 因健康探测成功而保留失败中的域名，用户会被继续钉在坏域名上，直到下一轮累计约 `IP_TABLE_SNI_FAILURE_THRESHOLD`（`10`）次失败才有机会自愈。这与本 PR "极端网络下快速故障切换" 的核心目标直接冲突。

## 建议

合并时不要静默丢弃更强的触发。可记录 in-flight run 的触发等级，`domain_failure` 到达时对其升级；或对 `domain_failure` 绕过合并、单独排队一次。示意：

```ts
const inflight = this.speedTestInFlight.get(domain);
if (inflight) {
  if (opts?.trigger === 'domain_failure') {
    this.pendingDomainFailure.add(domain); // 当前 run 结束后补跑一轮 domain_failure
  }
  return inflight;
}
```

[Claude session ↗](https://claude.ai/code/session_01CpVJ4RZGJ2UwWYFZVLZYsp)

---

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +154,165 @@ async function getMappedDomainForIpLookup(
...
+// ========== Fail-open on domain network failures ==========
+//
+// When real requests on the direct domain keep failing at the transport
+// level (timeouts, resets — NOT HTTP error responses), the adapter opens a
+// time-boxed window during which hosts of that root domain resolve to a
+// fallback IP (runtime last-best IP when available, builtin config
+// otherwise). This is the only protection available in runtimes where the
+// simpleDb-backed config is absent: the main runtime on split-runtime
+// targets and the cold-start window before background init completes.
+// Failing requests are never replayed — a wallet must not double-send
```
originalix: 这个约束目前只对触发 fail-open 的 direct-domain 失败成立。fail-open 激活后，后续所有 method（包括非幂等 POST）都会走 SNI；而现有 SNI catch 在 `SNI_TIMEOUT` / `SNI_REQUEST_TIMEOUT` 等非 fail-closed 错误或空响应时，会使用同一个 Axios config 再走一次 domain fallback。第一条请求可能已经被服务端处理，只是响应丢失，因此这里仍可能造成修改型请求双发。

请在 SNI 已经开始后按 method/idempotency 限制 fallback：只允许 GET/HEAD 或明确带幂等键的请求重试；其余请求应返回原错误，除非 native 能证明请求尚未写出。

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +154,165 @@ async function getMappedDomainForIpLookup(
...
+async function recordDomainRequestOutcome(options: {
+  rootDomain: string;
+  ok: boolean;
+  error?: unknown;
+}): Promise<void> {
+  const { rootDomain, ok, error } = options;
+  const lookupDomain = await resolveFailoverLookupDomain(rootDomain);
+  const state = getFailoverState(lookupDomain);
+
+  if (ok) {
```
originalix, line 394: 这里的 state 以 `lookupDomain`（root domain）为 key，而调用时已经丢失 hostname。因此 `utility.onekeycn.com` 的一次成功会清掉 `wallet.onekeycn.com` 的失败计数和 active fail-open；同 hostname 的晚到成功也会立即关闭刚打开的 circuit。部分 hostname 或部分连接成功、wallet 大量黑洞时，后续请求会被重新送回坏域名。

建议按 hostname 记录被动健康。root domain 可以共享最终路由，但只能由同 hostname 的连续成功或受控 half-open probe 关闭 fail-open，不能由任意一次 root-domain success 清空。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -718,6 +786,34 @@ class ServiceIpTable extends ServiceBase {
       }) for ${domain} (${target})`,
     });
 
+    // Fast failover: after a few consecutive real-request failures on the
+    // direct domain, switch to the last known-good IP immediately instead of
+    // waiting for the slow failure threshold + a full speed test round.
+    if (
+      requestType === 'domain' &&
+      endpointHealth.consecutiveFailures === IP_TABLE_DOMAIN_FAILOVER_THRESHOLD
```
originalix, line 1049: 这里把 `consecutiveFailures === 3` 当成连续的传输层失败，但 `callOriginalAdapter()` 对所有 reject 都会调用 `reportRequestFailureCallback`，Service 注册 callback 时又丢掉了 error；同时成功请求没有回调重置这个计数。于是 401/429/5xx、取消请求，甚至中间夹着大量成功的三次偶发错误，都可能触发 fast failover；累计 10 次还会以 `domain_failure` 强制选 IP。

请改为传递结构化 request outcome，只对白名单 transport error 计数，并在同 endpoint 真实成功时重置；HTTP response 应证明网络可达，cancel 不应计入。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -554,88 +636,74 @@ class ServiceIpTable extends ServiceBase {
...
+          defaultLogger.ipTable.metrics.endpointSwitched({
+            domain,
+            from: 'domain',
+            to: 'ip',
+            trigger: 'speed_test',
+            reason: decision.reason,
+          });
+        }
+      } else if (decision.action === 'select_domain') {
         await this.backgroundApi.simpleDb.ipTable.updateSelection(domain, '');
```
originalix: `configWithRuntime/currentSelection` 在测速开始前读取，而一轮串行测速可能持续几十秒。这期间真实流量达到三次失败会在 `reportRequestFailure()` 中先写入 last-best IP；旧 periodic run 完成后仍可能在这里基于启动时的 `currentSelection` 写回空字符串，把刚完成的 fast failover 撤销。随后更强的 `domain_failure` 还可能被当前 single-flight 合并掉。

请给 selection 写入增加 generation/priority，提交前重新读取并 compare-and-set；低优先级旧测速不能覆盖其启动后产生的 fast-failover 状态。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -295,7 +298,13 @@ class ServiceIpTable extends ServiceBase {
         info: `[IpTable] Fetching remote config from: ${IP_TABLE_CDN_URL}`,
       });
 
-      const plainAxios = axios.create();
+      // Fetch the config through the IP-Table-capable client: the config CDN
+      // (config.onekeycn.com) shares the walled root domain, so this fetch
+      // needs the same fail-open protection as business traffic. No cycle:
+      // endpoint selection only reads the local simpleDb/builtin table.
+      const plainAxios = createAxiosWithIpTable({
```
originalix, line 335: 改用 `createAxiosWithIpTable()` 仍不能保护 fresh install 的首次配置下载：adapter 需要在同一 runtime 累计三次 direct-domain transport failure 才激活 fail-open，而 `fetchRemoteConfig()` 只发送一次 GET，失败后本轮初始化直接返回。main 和 bg 的 failover Map 又属于独立 JS heap，main 的业务失败不会替 bg 累计阈值。

配置 GET 本身是幂等请求，建议为它提供专用的 direct → current LKG/builtin IP fallback，而不是依赖普通业务请求的三次失败状态。

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 8ba9fda75f. `isNetworkLevelError` is replaced with an explicit transport allowlist (`ERR_NETWORK`, `ECONNABORTED`, `ETIMEDOUT`, `ECONNRESET`, `ECONNREFUSED`, `ENOTFOUND`, `ENETUNREACH`, `EHOSTUNREACH`, `EAI_AGAIN`); `axios.isCancel()` / `ERR_CANCELED` and any error without an allowlisted code never count toward the threshold. Covered by `cancellations do not count as transport failures` and `errors without an allowlisted transport code do not count`.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 6aaf609dc9. The check is extracted into `isIpTableConfigRegression()` (shared/utils/ipTableUtils.ts): when `lastVerified` is absent it anchors on the active config's own version/generated_at, so a same-version-but-older signed config is rejected before it can move the anchor backwards, and an unparseable remote `generated_at` is rejected outright. The pre-upgrade (no `lastVerified`) case you described is pinned by a unit test.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 8ba9fda75f + f5544cf98f. Selection now goes through `getActiveFailoverIp()`, which re-checks the kill switch on every lookup while a circuit is open and immediately clears the failover state plus the memoized selection when the switch is flipped (test: `flipping the kill switch while fail-open is active takes effect immediately`). The service-side check now also ignores `devSettings.enabled` to match the adapter and the existing `disableIpTableInProd` convention, so main and bg runtimes reach the same conclusion for the same settings.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 459abea292 using the follow-up-queue approach: the in-flight entry records its trigger; when a `domain_failure` call arrives during a weaker run, the domain goes into `pendingDomainFailureRerun` and exactly one `domain_failure` round runs after the current one settles (no re-queue when the in-flight run is already `domain_failure`, so no loop). Together with the stale-write guard in the same commit, the cold-start scenario now escalates instead of being coalesced away.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 8ba9fda75f. Once a request has been handed to the SNI transport, domain fallback is gated by `canFallbackAfterSniStarted()`: idempotent methods (GET/HEAD/OPTIONS) may retry; non-idempotent methods only when the error code proves the connection was never established (`SNI_CONNECTION_REFUSED`, `SNI_DNS_FAILED`, `SNI_NETWORK_UNREACHABLE`, `SNI_INVALID_URL`); ambiguous outcomes (`SNI_TIMEOUT`, `SNI_REQUEST_TIMEOUT`, null response) surface the original error. Note this intentionally tightens the pre-existing `selected_ip` path too — it had the same double-send exposure. Blocked fallbacks are logged as `sni_fallback_blocked` so the UX impact is observable. Tests cover GET/POST x timeout/refused/null.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 8ba9fda75f. Failure counters are per hostname (`hostFailures: Map<hostname, {consecutive, lastFailureAt}>`); routing stays per root domain, but early close now requires a success that (a) comes from a hostname in `activatedHostnames` and (b) belongs to a request started after activation — so neither a `utility.*` success nor a late in-flight success closes a circuit opened by `wallet.*`. The TTL remains the backstop, which also serves as the half-open probe window. Tests: `a domain success on another hostname does not close the circuit`, `a late success from a request started before activation does not close the circuit`.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 8ba9fda75f + 459abea292. Two changes: (a) the adapter only reports allowlisted transport failures to the service — HTTP responses and cancellations never reach the counters; (b) a new `setReportRequestSuccessCallback` reports real successes (both domain and SNI paths) and the service resets `domainDirect` / per-IP `consecutiveFailures` on them. "Consecutive" is now literal for both the fast-failover threshold (3) and the speed-test threshold (10).

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 459abea292. Every selection write bumps a per-domain generation token; a speed-test round captures it before probing and re-checks before applying. A changed generation discards the result unless the round itself was `domain_failure`-triggered — live-traffic evidence stays authoritative over stale periodic measurements. Fast failover writes go through the same `applySelection()`, and the coalescing escalation in the same commit closes the merge-away path this raced with.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 41105111d7. `fetchRemoteConfig()` no longer depends on accumulated fail-open state: on failure it replays the (idempotent) config GET directly over SNI against up to 3 candidates — current selection, then last-best IP, then builtin endpoints — behind a proxy-preflight guard, and shape-validates the response before returning it. First-run devices on walled networks get a working config path with zero prior failure history, and the main/bg heap separation stops mattering because the fallback runs inside the same call.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -521,14 +860,50 @@ export function createIpTableAdapter(
...
+      }
+
       return response;
     } catch (error) {
-      // Only report domain failures if this is NOT a fallback request
+      if (rootDomain && hostname) {
+        // Track transport-level failures for adapter fail-open. The failing
+        // request itself is never replayed (no double-send of POSTs); only
+        // subsequent requests take the fail-open route.
+        await recordDomainRequestOutcome({
```
originalix: 当原始 Axios adapter 因 `validateStatus` 拒绝 4xx/5xx 响应时，请求会直接进入下面的 `catch`，这里的 success reset 不会执行。虽然 catch 不会把带 `response` 的错误计为新网络失败，但之前的失败计数也不会清除，因此被正常 HTTP 响应隔开的传输错误仍可能累计到 fail-open 阈值。

请在 catch 中检测 `error.response`，按 domain success 重置 adapter 和 service 计数后再原样抛出该 HTTP 错误。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -789,6 +1069,24 @@ class ServiceIpTable extends ServiceBase {
       void this.reportRequestFailure(domain, requestType, target);
     });
 
+    // Successes reset the consecutive-failure counters, so "consecutive"
+    // means what it says: interleaved successes prevent sporadic errors
+    // accumulated over hours from ever tripping the failover threshold.
+    setReportRequestSuccessCallback(({ domain, requestType, target }) => {
```
originalix: 当多个请求并发完成时，上方 failure callback 会通过 `void reportRequestFailure()` 异步落账，而这里的 success callback 会立即清零，两个结果没有共同的时序标识。旧失败可能在新成功之后增加计数，早先启动的慢成功也可能清除较新的失败，导致快速切换误触发或无法触发。

请把 `startedAtMs` 传入 success/failure callback，并在 ServiceIpTable 按 endpoint 保存最后处理的请求时间，只接受不早于当前状态的 outcome。

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -806,6 +1181,21 @@ export function createIpTableAdapter(
...
+        // methods; anything else must surface the failure to the caller.
+        const method = (config.method || 'GET').toUpperCase();
+        if (!canFallbackAfterSniStarted(method)) {
+          logIpTableEvent('warn', 'sni_fallback_blocked', {
+            hostname,
+            rootDomain,
+            method,
+            reason: 'non_idempotent_after_sni_started',
+          });
+          throw new OneKeyLocalError(
```
originalix, line 1287: 当非幂等请求收到 null SNI response 时，上面已经调用了一次 `reportRequestFailureCallback`，这里抛出的 `OneKeyLocalError` 随后又会进入同一层 `catch`，并对相同 IP 再上报一次。因此一次 POST/PATCH 失败会增加两次 `consecutiveFailures`，提前触发测速和端点切换。

请让 null 分支和 catch 共用一个 request-local 的 `reportIpFailureOnce` 去重入口，并补充 callback 只调用一次的测试。

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in ea33f9881e. The catch now detects `error.response` and treats a rejected 4xx/5xx exactly like a success for counter purposes — `recordDomainRequestOutcome(ok: true)` plus the service success callback run before the HTTP error is rethrown — so transport failures separated by healthy responses can no longer accumulate to the threshold. Pinned by `an http error response resets the consecutive failure counter` (2 transport failures + 500 + 2 transport failures must not open the circuit).

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in ea33f9881e. Both callbacks now carry `startedAtMs` (transport hand-off time), and `IEndpointHealth` tracks `lastOutcomeAt`: `reportRequestFailure` ignores a failure whose request started before the newest applied outcome, and the success handler ignores a slow success that started before it. Ordering by request start time makes the two interleavings you described converge to the same state — an old failure can no longer re-increment after a newer success, and an early-started slow success can no longer clear newer failures.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in ea33f9881e. The null-response branch and the outer catch now share a request-local `reportIpFailureOnce()`, so one SNI attempt produces at most one ip-failure report regardless of which paths it traverses (the fail-closed early-throw still reports nothing, unchanged). Test `one sni attempt reports at most one ip failure (null response + blocked fallback)` asserts the callback fires exactly once for the POST + null-response case.

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -785,9 +1087,42 @@ class ServiceIpTable extends ServiceBase {
...
+    );
+
+    // Successes reset the consecutive-failure counters, so "consecutive"
+    // means what it says: interleaved successes prevent sporadic errors
+    // accumulated over hours from ever tripping the failover threshold.
+    // Ordered by request start time: a slow success that started before the
+    // newest applied outcome must not clear failures that came after it.
+    setReportRequestSuccessCallback(
+      ({ domain, requestType, target, startedAtMs }) => {
+        const stats = this.domainHealthMap.get(domain);
```
originalix: 当较早开始的失败仍在 `reportRequestFailure()` 的异步配置读取中等待，而较晚开始的成功先完成时，`domainHealthMap` 或对应 IP endpoint 可能尚不存在，这里会直接丢弃成功。随后旧失败创建状态并累加，`lastOutcomeAt` 没有机会阻止它，恢复后的链路仍会被记为连续失败并可能误触发切换。

请让 success 路径也通过 `getDomainHealth()` 和 `getEndpointHealth()` 创建或取得 outcome ledger，先写入 `startedAtMs` 再清零计数，并补充新成功先落账、旧失败后返回的回归测试。

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +208,244 @@ async function getMappedDomainForIpLookup(
...
+    return;
+  }
+
+  const now = Date.now();
+  const hostFailure = state.hostFailures.get(hostname) ?? {
+    consecutive: 0,
+    lastFailureAt: 0,
+  };
+  hostFailure.consecutive += 1;
+  hostFailure.lastFailureAt = now;
```
originalix: 当较早请求慢失败、较晚请求先成功时，success 分支在没有 `hostFailure` 时不会保存任何顺序；随后这里又用失败处理时的 `Date.now()` 写入 `lastFailureAt`，旧失败仍会成为最新 outcome 并增加计数。并发请求因此可能把已经恢复的 hostname 误判为连续失败并打开 fail-open。

请为每个 hostname 持久记录 `lastOutcomeAt`，让 success 和 failure 都按 `startedAtMs` 比较并拒绝旧 outcome，不要用完成时刻代替请求开始顺序。

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 28a5f93a86. The success handler now goes through `getDomainHealth()` + `getEndpointHealth()` (get-or-create) and applies the outcome via the shared start-time ledger, so a later-started success completing first always leaves its ordering mark — the early-started failure that lands afterwards is rejected as stale instead of creating fresh state and counting. The ordering logic itself is extracted into `ipTableOutcomeLedger.applyOutcome()` with the regression you asked for pinned as `a newer success applied first blocks an older failure arriving late`, plus an interleaving-convergence test.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 28a5f93a86. The adapter's per-hostname record is now the same shared ledger: entries are get-or-create for BOTH success and failure (a success on a fresh entry records `lastOutcomeAt`), entries persist across successes instead of being deleted (deleting lost the ordering mark), and failures are ordered by `startedAtMs` — the completion-time `Date.now()` write is gone. Adapter-level regression: `a late failure from a request started before a newer success does not count` (slow-fail started first, success lands first, late failure must be stale and two more fresh failures must NOT open the circuit).

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -100,12 +107,66 @@ interface IRequestFailureParams {
...
+ * the server even when we got no usable response back. Re-sending it over
+ * the domain is only safe for idempotent methods, or when the error code
+ * proves the connection was never established (nothing was written).
+ */
+const IDEMPOTENT_HTTP_METHODS = new Set(['GET', 'HEAD', 'OPTIONS']);
+
+const SNI_PRE_WRITE_ERROR_CODES = new Set([
+  'SNI_CONNECTION_REFUSED',
+  'SNI_DNS_FAILED',
+  'SNI_NETWORK_UNREACHABLE',
```
originalix: `SNI_NETWORK_UNREACHABLE` 不能证明请求尚未写出。当前 iOS `@onekeyfe/react-native-sni-connect@3.0.78` 会把 `NSURLErrorNetworkConnectionLost` 也映射成这个 code；该错误可发生在 body 已发送、服务端已执行、仅响应途中断链之后。这里因此会允许 POST/PATCH 再走 domain fallback，造成修改请求双发。请先从 non-idempotent 的 pre-write allowlist 移除该 code；若需要保留能力，应由 native 拆分明确的 connect-before-send 与 connection-lost 阶段，只有前者才能重试。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -753,16 +1032,41 @@ class ServiceIpTable extends ServiceBase {
     });
 
     // Trigger speed test to find and switch to better endpoint
-    void this.selectBestEndpointForDomain(domain);
+    void this.selectBestEndpointForDomain(domain, {
+      trigger: requestType === 'domain' ? 'domain_failure' : 'ip_failure',
```
originalix, line 1109: `disableIpTableFailover` 只在三次失败的 fast switch 分支中检查。开关开启后，累计到测速阈值仍会从这里传入 `domain_failure`，而 `decideEndpoint()` 会把 `domainFailingRealTraffic=true` 并无视 30% 门槛继续切 IP，所以这个 kill switch 不能真正关闭 failover。请在触发测速或构造 decision 前统一读取开关：禁用时不要传递 `domain_failure` override，并同步清理 adapter circuit/memo 及由 failover 写入的 selection。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -380,7 +533,9 @@ class ServiceIpTable extends ServiceBase {
         } domains`,
       });
 
-      await this.saveConfig(mergedConfig);
+      await this.backgroundApi.simpleDb.ipTable.saveConfig(mergedConfig, {
+        payloadHash: computeIpTableConfigHash(remoteConfig),
```
originalix, line 556: 这里保存的是 `mergedConfig`，但 `payloadHash` 计算自原始 `remoteConfig`。`mergeIpTableConfigs()` 又会永久保留旧 local endpoints，因此持久化的 effective config 并不是该 hash/signature 覆盖的对象；远端即使删除已下线或失陷 IP，旧 endpoint、selection 和 `lastBestIp` 仍可继续被使用。请把 signed remote envelope 原样持久化，并与 builtin/effective config 分离；每次从当前签名配置重新派生 effective endpoints，同时清理已撤销的 runtime selection/LKG。

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +210,256 @@ async function getMappedDomainForIpLookup(
...
+    logIpTableEvent('info', 'adapter_failover_deactivated', {
+      lookupDomain,
+      cause,
+    });
+    defaultLogger.ipTable.metrics.adapterFailover({
+      lookupDomain,
+      action: 'deactivated',
+    });
+  }
+  state.hostFailures.clear();
```
originalix: 恢复 success 刚通过 `applyOutcome()` 写入最新 watermark，随后这里却把整个 `hostFailures` ledger 清空。如果还有更早启动的慢请求未完成，它们之后失败时会重建空 entry 并重新计数，三个旧 failure 就能在链路已恢复后再次打开 circuit。请保留每个 hostname 的最新 outcome watermark，只重置 `consecutiveFailures` 和当前 circuit generation；并补一个“恢复关闭 circuit 后，旧失败再落地仍被判 stale”的测试。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -785,9 +1089,34 @@ class ServiceIpTable extends ServiceBase {
...
+    // Successes reset the consecutive-failure counters, so "consecutive"
+    // means what it says. The success path MUST create the ledger entry when
+    // none exists yet: a later-started success completing before an
+    // early-started slow failure has to leave its ordering mark, otherwise
+    // the old failure would create the entry afterwards and count against a
+    // link that already proved healthy.
+    setReportRequestSuccessCallback(
+      ({ domain, requestType, target, startedAtMs }) => {
+        const stats = this.getDomainHealth(domain);
+        const endpointHealth = this.getEndpointHealth(
```
originalix, line 1185: 这个 success 现在会进入 `getEndpointHealth()`，但该 helper 对所有 `requestType === 'domain'` 都返回同一个 `stats.domainDirect`，完全忽略 `target`。因此 bg runtime 中 `utility.onekeycn.com` 的成功仍会清零 `wallet.onekeycn.com` 的失败，不同 hostname 的失败也能共同凑到阈值；而 main adapter 已按 hostname 计数，两个独立 JS heap 会得到不一致的切换结果。请让 Service 的 direct health 同样按 hostname 建 ledger，再显式聚合 root-domain 路由状态。

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 26e1f8753b. Confirmed in the native source: SniConnectClient.swift maps `NSURLErrorNotConnectedToInternet` and `NSURLErrorNetworkConnectionLost` to the same `.networkUnreachable` case, and connection-lost can indeed fire after the body was written. `SNI_NETWORK_UNREACHABLE` is removed from the pre-write allowlist (a code comment documents why), leaving `SNI_CONNECTION_REFUSED` / `SNI_DNS_FAILED` / `SNI_INVALID_URL`, which map only to pre-connect phases. Negative test added: POST + `SNI_NETWORK_UNREACHABLE` surfaces the error without fallback. Splitting connect-before-send from connection-lost in the native module is the right long-term shape — that belongs to the sni-connect repo and I'll leave it as a follow-up there.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 153274f971. The trigger is now normalized at the `selectBestEndpointForDomain` entry: when `disableIpTableFailover` is on, `domain_failure` downgrades to `periodic` before anything reads it — that single point disables the reachability-first override inside `decideEndpoint`, the stale-write authority, and the coalescing escalation. The adapter circuit/memo already self-clear when the switch is read while a circuit is open (`getActiveFailoverIp` → `deactivateFailover('kill_switch')`). On clearing failover-written selections: selections carry no provenance, so wiping IP selections wholesale would also destroy legitimate speed-test choices that predate this PR's behaviors; with the override gone, the next probing round re-evaluates under the plain 30% policy and writes the domain back if it is competitive, and the existing "Reset IP Table Cache" dev action covers immediate manual cleanup. If you want provenance-tagged selections so the switch can surgically revert only failover writes, I'd rather do that as a follow-up than widen this PR further.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 153274f971. `fetchAndApplyRemoteConfig` (renamed from `fetchAndMergeRemoteConfig`) now persists the signed remote envelope verbatim — `mergeIpTableConfigs` is removed entirely — so the stored config is exactly what the signature and the persisted `payloadHash` cover, and endpoints removed from the CDN config actually disappear. `saveConfig` prunes `runtime.selections` and `lastBestIp` entries pointing at endpoints the new config no longer endorses (`pruneIpTableRuntimeSelections`, unit-tested including the revoked-IP and dropped-domain cases). The builtin config remains only as the pre-first-verify default and the adapter's circuit-open last resort — binary-shipped, not persisted state — and existing installs with a merged superset self-heal on their next successful verify.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: Fixed in 26e1f8753b. `deactivateFailover` no longer clears the ledger: it resets each hostname's `consecutiveFailures` but preserves `lastOutcomeAt` watermarks (plus circuit state: `failOpenUntil`/`activatedAt`/`activatedHostnames`). Regression test added exactly as you described: three slow requests started before the incident, circuit opens and then recovers, the three old failures land afterwards — all stale, circuit stays closed.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 153274f971. `domainDirect` is replaced by `directHosts: Map<hostname, IEndpointHealth>` — `getEndpointHealth` buckets domain outcomes per hostname (target is the hostname), so a `utility.*` success no longer clears `wallet.*` failures and different hostnames can't jointly reach a threshold. Root-domain decisions aggregate explicitly via `getMaxDirectConsecutiveFailures` (speed-test trigger fires when any single hostname reaches the threshold), and the pre-probe reset iterates all host entries. This matches the adapter's per-hostname ledger, so main and bg runtimes now reason identically.

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -554,89 +825,89 @@ class ServiceIpTable extends ServiceBase {
...
-        });
-        await this.backgroundApi.simpleDb.ipTable.updateSelection(
+      // Persist the best measured IP for fast failover, regardless of the
+      // final decision — it is the candidate we switch to when the direct
+      // domain starts failing real traffic.
+      const bestMeasuredIp = Object.entries(ipLatencies)
+        .filter(([, latency]) => latency !== Infinity)
+        .toSorted((a, b) => a[1] - b[1])[0]?.[0];
+      if (bestMeasuredIp) {
+        await this.backgroundApi.simpleDb.ipTable.updateLastBestIp(
```
originalix: 当测速基于旧 `domainConfig` 运行期间，新的已验签配置可能已经保存并移除了某个 endpoint。这里仍会把旧测速得到的 IP 写回 `lastBestIp`，后续还可能写成 selection；`selectionGeneration` 不会因 `saveConfig()` 和 pruning 而变化，所以无法拦截这次陈旧写入。

请在测速开始时记录配置 hash/version，并在写入前重新读取当前配置；配置已变化或候选不再属于当前 endpoint 集合时，丢弃本轮结果并重新测速。

packages/shared/src/request/helpers/ipTableOutcomeLedger.ts
```diff
@@ -0,0 +1,38 @@
...
+export function createOutcomeLedgerEntry(): IIpTableOutcomeLedgerEntry {
+  return { consecutiveFailures: 0, lastOutcomeAt: 0 };
+}
+
+export function applyOutcome(
+  entry: IIpTableOutcomeLedgerEntry,
+  outcome: { ok: boolean; startedAtMs: number },
+): 'applied' | 'stale' {
+  if (outcome.startedAtMs < entry.lastOutcomeAt) {
+    return 'stale';
```
originalix: 当两个并发请求在同一毫秒启动时，它们会得到相同的 `startedAtMs`。如果较晚请求先成功、较早请求随后慢失败，这里的严格 `<` 判断仍会应用旧失败并重新增加计数，导致 outcome ledger 不再与请求顺序收敛，可能误触发 fail-open 或测速。

请在 transport 启动时分配 runtime 内单调递增的请求序号，并使用该序号排序 outcome，不要只依赖毫秒时间戳。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -316,6 +353,16 @@ class ServiceIpTable extends ServiceBase {
         return null;
       }
 
+      // A misconfigured CDN edge can return HTML or a truncated body that
+      // still parses; reject it here so the log points at the delivery
+      // layer instead of a bogus "signature verification failed".
+      if (!isValidIpTableRemoteConfigShape(remoteConfig)) {
```
originalix, line 361: 当 captive portal 或异常 CDN edge 以 HTTP 200 返回 HTML、空内容或结构不完整的数据时，请求不会抛异常，而是进入这个分支后直接返回 null。这样 `fetchRemoteConfigViaSniFallback()` 完全不会执行，即使 builtin SNI 候选能够取得正确配置，首次安装设备仍无法恢复。

请将空响应和 shape validation 失败统一转入 `fetchRemoteConfigViaSniFallback()`；代理是否允许绕行继续由 fallback 方法里的 preflight 决定。

packages/kit-bg/src/services/ServiceIpTable.ts
```diff
@@ -554,89 +845,121 @@ class ServiceIpTable extends ServiceBase {
...
-        await devSettingsPersistAtom.get();
-      const forceIpTableStrict =
-        devSettingEnabled && settings?.forceIpTableStrict;
-      // In strict mode, always use IP if available (regardless of domain speed)
-      if (forceIpTableStrict) {
-        defaultLogger.ipTable.request.info({
-          info: `[IpTable] [STRICT MODE] Using best IP: ${domain} -> ${bestIp} (${bestIpLatency}ms)`,
-        });
-        await this.backgroundApi.simpleDb.ipTable.updateSelection(
+      if (bestMeasuredIp) {
```
originalix: 当本轮测速的配置复核通过后，新的已验签配置仍可能在这里写入前完成保存和 pruning。当前检查与 `updateLastBestIp()`、后续 selection 更新是独立的 SimpleDB 操作，因此旧测速会在新配置移除 endpoint 后再次把该 IP 写回，导致撤销失效。

请将 expected config hash 和 endpoint membership 校验移入与 runtime 更新相同的 `setRawData` 临界区；校验失败时拒绝写入并排队重测。

packages/shared/src/request/helpers/ipTableOutcomeLedger.ts
```diff
@@ -0,0 +1,46 @@
...
+export function nextIpTableRequestSequence(): number {
+  ipTableRequestSequence += 1;
+  return ipTableRequestSequence;
+}
+
+export function applyOutcome(
+  entry: IIpTableOutcomeLedgerEntry,
+  outcome: { ok: boolean; requestSequence: number },
+): 'applied' | 'stale' {
+  if (outcome.requestSequence < entry.lastOutcomeSequence) {
```
originalix: 当请求顺序为失败 1、成功 2、失败 3，而完成顺序为失败 1、失败 3、成功 2 时，这里的 watermark 会把成功 2 判为 stale，使失败计数错误地保持为 2。按请求顺序，成功 2 应截断失败 1，只保留失败 3；并发 API 流量因此可能误触发 fail-open 或测速。

请按 hostname/endpoint 保存最新成功序号和其后的失败序号集合；成功到达时移除不晚于它的失败，再由剩余集合计算 `consecutiveFailures`。

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed across d1ce18dc90 and the follow-up 9a253c86e1. A speed-test round captures the signed payload hash at start, and final persistence now goes through `commitSpeedTestResult()`, which validates the current hash plus every measured/selected endpoint inside the same SimpleDB `setRawData` mutex used by `saveConfig()`. A stale result performs no runtime write and queues one fresh round. Covered by `SimpleDbEntityIpTable.test.ts` and `ServiceIpTable.review.test.ts`.

packages/shared/src/request/helpers/ipTableOutcomeLedger.ts
originalix: Fixed in d1ce18dc90 and hardened in 9a253c86e1. Every transport receives a runtime-monotonic `requestSequence` before network work starts. The ledger now tracks the latest success sequence plus the set of later failure sequences, so same-millisecond starts and out-of-order completions converge by request order. The focused IP Table regression suite passes 81/81.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in d1ce18dc90. Empty HTTP 200 responses and responses that fail the config shape guard now enter `fetchRemoteConfigViaSniFallback()` instead of returning `null`; proxy bypass policy remains centralized in the fallback preflight. Both empty and captive-portal/invalid-shape cases are covered in `ServiceIpTable.review.test.ts`.

packages/kit-bg/src/services/ServiceIpTable.ts
originalix: Fixed in 9a253c86e1. The pre-write check and the runtime mutation are now one `SimpleDbEntityIpTable.commitSpeedTestResult()` transaction under the same `setRawData` mutex as config replacement/pruning. It checks the expected signed-config hash and endpoint membership inside that critical section, atomically writes last-best plus selection only on success, and returns `stale_config` without a write so the service queues a fresh speed test. Dedicated entity and service regressions cover both the race and membership rejection.

packages/shared/src/request/helpers/ipTableOutcomeLedger.ts
originalix: Fixed in 9a253c86e1. Each hostname/endpoint ledger now stores `latestSuccessSequence` and the set of failure sequences after it. A success removes failures with sequence less than or equal to its own, then recomputes `consecutiveFailures` from the remaining set. The exact failure-1/success-2/failure-3 completion reordering is covered, and an adapter-level test also verifies that a provisional fail-open is retracted when a late middle success shows the failures were not consecutive. Focused IP Table tests pass 81/81.

packages/shared/src/request/helpers/ipTableOutcomeLedger.ts
```diff
@@ -0,0 +1,66 @@
...
+    return 'applied';
+  }
+
+  if (
+    outcome.requestSequence <= entry.latestSuccessSequence ||
+    entry.failureSequences.has(outcome.requestSequence)
+  ) {
+    return 'stale';
+  }
+  entry.failureSequences.add(outcome.requestSequence);
```
sidmorizon: # 🟡 P2 failureSequences 集合在长期失败端点上无界增长

## 问题

`failureSequences` 只有两种收缩途径：一次被判定为 `applied` 的成功（`forEach` 删除 `<= requestSequence` 的项），或 `deactivateFailover` 里的 `entry.failureSequences.clear()`。失败路径这里 `entry.failureSequences.add(outcome.requestSequence)` 没有任何上限，永远只增不减。

服务端（`ServiceIpTable`）的同构 `IEndpointHealth` 在每次触发 speed test 时会 `failureSequences.clear()`，所以有界；但适配器侧 `hostFailures` 的清理**只**发生在 `deactivateFailover`（成功恢复 / kill switch / reset）。当某 root domain 的 `selections[domain]` 停留在 `''`（选域名）、且没有 `lastBestIp` 可供 fast failover 切到 IP 时，熔断窗口（`IP_TABLE_ADAPTER_FAILOVER_TTL_MS` = `5min`）到期后会再发一次域名请求并再次失败，`add` 一个新 sequence 又重开熔断，如此循环，且期间从无成功来做 prune。

## 影响

在被墙/持续不可达网络上的全新安装（无 `lastBestIp`、选择恒为 `''`）里，`failureSequences` 会以约 `1 条 / 5min`（TTL 边界并发时更多）的速度对该 hostname 单调增长，进程生命周期内无任何复位——多天会话可累积数千条 number。改造前 `consecutiveFailures` 是 `O(1)` 整数，切换到 `Set` 后变成 `O(n)` 内存，这是本次引入的慢性泄漏。虽非崩溃级，但发生在整个 app 生命周期常驻的 `@onekeyhq/shared` 传输层模块里。

## 建议

由于判定只关心 `consecutiveFailures >= IP_TABLE_DOMAIN_FAILOVER_THRESHOLD`（=`3`），无需保留超过阈值数量的失败 sequence。可在 `add` 后对集合做上限裁剪，仅保留最新的若干条（`>= THRESHOLD` 即可），成功仍按 `<=` 语义 prune：

```ts
entry.failureSequences.add(outcome.requestSequence);
const CAP = IP_TABLE_DOMAIN_FAILOVER_THRESHOLD * 4; // 远大于阈值即可
if (entry.failureSequences.size > CAP) {
  // Set 迭代按插入顺序，删最旧的
  const oldest = entry.failureSequences.values().next().value;
  entry.failureSequences.delete(oldest);
}
entry.consecutiveFailures = entry.failureSequences.size;
```

[Claude session ↗](https://claude.ai/code/session_01CpVJ4RZGJ2UwWYFZVLZYsp)

---

originalix: Fixed in ab6620dc7b. `failureSequences` is now capped at `max(IP_TABLE_DOMAIN_FAILOVER_THRESHOLD, IP_TABLE_SNI_FAILURE_THRESHOLD) * 4` (currently 40). Since completion order is not request order, eviction deliberately removes the numerically smallest request sequence rather than the first Set insertion, retaining the greatest sequences. This keeps the set bounded while preserving every threshold decision under a later success: once pruning leaves a below-threshold count, all discarded sequences are necessarily older than that success. Added regressions insert the greatest sequence first to reject insertion-order eviction and verify that a late success still restores the exact below-threshold count. Focused IP Table tests pass 83/83; typecheck, type-aware oxlint, oxfmt, and diff-check pass.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -149,6 +215,282 @@ async function getMappedDomainForIpLookup(
...
+    }
+    // A success only proves the path that was exercised: it may close the
+    // circuit only when it comes from a hostname that opened it AND started
+    // after activation (a long-in-flight request resolving late says
+    // nothing about recovery).
+    if (
+      applied === 'applied' &&
+      state.failOpenUntil > Date.now() &&
+      state.activatedHostnames.has(hostname) &&
+      requestSequence > state.activatedRequestSequence
```
originalix: 当同一 root domain 下有多个 hostname 达到失败阈值时，它们都会进入 `activatedHostnames`，但这里仍只比较首次打开 circuit 的全局 `activatedRequestSequence`，随后直接清空整个状态。因此，任一 activated hostname 的较新成功都会关闭 circuit，即使其他 hostname 仍保留达到阈值的失败；其后续请求会重新走仍然故障的域名路径。

请为每个 activated hostname 保存独立激活序号，恢复时只移除已被较新成功证明恢复的 hostname，并仅在激活集合为空时关闭 circuit。

originalix: Fixed in b38ebb26e2. The root-domain circuit now stores an independent activation sequence in `activatedHostSequences` for each hostname. A successful outcome retires only that hostname after its unresolved failures fall below the threshold, and the circuit is deactivated only when no activated hostname remains. Added a regression where `wallet.*` and `utility.*` both activate; after only `wallet.*` recovers, `utility.*` continues through SNI. Focused IP Table tests pass 84/84.

packages/shared/src/request/helpers/ipTableAdapter.test.ts
```diff
@@ -228,3 +239,935 @@
...
+      markAllRequestsStarted = resolve;
+    });
+    let startedRequests = 0;
+
+    fallbackAdapter.mockImplementation(async (config) => {
+      startedRequests += 1;
+      if (startedRequests === 6) {
+        markAllRequestsStarted?.();
+      }
+      if (config.url?.includes('wallet.onekeycn.com')) {
```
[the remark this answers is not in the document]
originalix: Fixed in f89034e3ef. The test fixture now parses `config.url` with `URL` and compares the resulting `hostname` against the exact allowlisted host, so host-like text in a path, query, or attacker-controlled hostname cannot satisfy the branch. Focused IP Table tests pass 84/84; typecheck, type-aware oxlint, oxfmt, and diff-check pass.

zhaono1 (approved):

wabicai (approved):

PR #12550, opened 2026-07-19
[OK-33404] fix: read Android bundled assets as strings

huhuanming:
[OK-33404](https://onekeyhq.atlassian.net/browse/OK-33404)

----------

## Issue

[OK-33404](https://onekeyhq.atlassian.net/browse/OK-33404) reports that bundled images loaded with React Native `require()` cannot be read as Base64 in Android production builds.

The affected flow is equivalent to:

```ts
const source = require('./blank.png');
const uri = Image.resolveAssetSource(source).uri;

await FileSystem.readAsStringAsync(uri, {
  encoding: 'base64',
});
```

Observed behavior:

- Android development builds work.
- iOS, web, and desktop builds work in both development and production.
- Android production resolves the image, but the resulting value is not a complete file URI and Expo FileSystem rejects it when reading as Base64.

Expected behavior: callers should be able to pass either a normal URI or the result of `require()` to one string-reading API and receive the requested content.

## Root cause

React Native `require('./image.png')` does not return a filesystem path. On native platforms it returns a numeric Metro asset ID, which is resolved by `Image.resolveAssetSource()`.

The resolved Android value differs by build mode:

- In development, Metro generally provides an HTTP URI.
- In production, Android packages the image into the APK resource table. React Native resolves the asset ID to a scheme-less drawable resource name such as `packages_shared_src_assets_blank`.

That drawable name is intentional and cannot be converted into a normal absolute path because the image is an APK resource rather than a standalone file.

The legacy Android implementation in `expo-file-system` then takes different native paths depending on the requested encoding:

- Non-Base64 `readAsStringAsync` reads a null-scheme resource through `openResourceInputStream`.
- Base64 `readAsStringAsync` uses `getInputStream`, which supports URI-backed inputs but rejects the scheme-less drawable name.
- `copyAsync` has a resource-aware null-scheme branch and can copy the packaged drawable to a regular file URI.

As a result, the failure is specific to the combination of Android production, a numeric `require()` asset, a scheme-less resolved resource name, and Base64 encoding.

## Solution

This PR adds a shared `readAsStringAsync(source, options)` wrapper that accepts both normal URI strings and React Native image sources.

The wrapper resolves the source first and performs the Android resource-copy path only when all four known conditions are true:

```ts
platformEnv.isNativeAndroid &&
isNumber(source) &&
!uri.includes(':') &&
options?.encoding === 'base64'
```

Behavior by path:

1. Normal URIs and all non-Base64 reads are passed directly to Expo FileSystem.
2. An Android production `require()` drawable requested as Base64 is copied to the native cache directory with `copyAsync`.
3. The copied `file://` URI is read with Expo's Base64 reader.
4. The temporary file is deleted with an idempotent, best-effort cleanup.

The temporary filename contains the Metro asset ID, timestamp, and random suffix. This prevents the main and background JS runtimes, which have separate JS heaps but share the native filesystem, from overwriting or deleting each other's in-flight files.

The existing required-image Base64 conversion now uses this wrapper, so callers receive the same raw Base64 content before the image helper adds the `data:image/...;base64,` prefix.

## Scope and compatibility

- Android URI-backed reads are unchanged.
- Android non-Base64 bundled-resource reads continue using Expo's direct resource-aware path.
- iOS, web, and desktop behavior is unchanged.
- Non-Android errors are propagated directly rather than entering the Android fallback.
- A repository-wide audit found no other production `readAsStringAsync` callers to migrate. The other Expo FileSystem integrations only write CSV files or create log upload tasks.

## Validation

  - worktree TypeScript lint
  - formatting
  - agent context
  - staged lint
  - staged TypeScript checks
- Type-aware Oxlint and Oxfmt checks passed.
- Built and installed an Android API 35 `prodRelease` APK through ADB.
- Called the new API directly with `require('./blank.png')` and `encoding: 'base64'`.
- Compared the entire returned Base64 string against the source PNG:
  - exact Base64 equality: `true`
  - source size: 74 bytes
  - Base64 length: 100
  - SHA-256: `2117bbb5576526b3ebfb7f54cfe783c30388558ae32116a282c75f96256ebf86`

## Related issue

- [OK-33404](https://onekeyhq.atlassian.net/browse/OK-33404)

[OK-33404]: https://onekeyhq.atlassian.net/browse/OK-33404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-33404]: https://onekeyhq.atlassian.net/browse/OK-33404?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/shared/src/utils/imageUtils.ts
```diff
@@ -911,6 +880,67 @@ async function getUriFromRequiredImageSource(
...
+    !uri.includes(':') &&
+    options?.encoding === 'base64';
+
+  if (!shouldCopyAndroidBundledResource) {
+    return ExpoFSReadAsStringAsync(uri, options);
+  }
+
+  const cacheDir = await getNativeCacheDirectory();
+  const timestamp = Date.now();
+  const random = Math.floor(Math.random() * 10_000);
```
sidmorizon: # [🟡 P2] 临时文件名唯一性不足，存在删除竞态窗口

## 问题

临时文件名由 `${source}-${timestamp}-${random}` 构成，其中 `timestamp = Date.now()`（毫秒）、`random = Math.floor(Math.random() * 10_000)`（仅 4 位随机）。同一 `source`（同一 drawable 资源 ID）在同一毫秒内被并发读取时，只需再命中相同的 random 即会得到完全相同的 `copiedUri`。此时任一路径 `finally` 中的 `ExpoFSDeleteAsync(copiedUri)` 会在另一路径仍在 `ExpoFSReadAsStringAsync(copiedUri)` 期间删除该文件，导致后者读取失败。

紧邻的注释声称 “per-call name avoids one runtime deleting or overwriting another one's temporary resource” —— 但 `Date.now() + Math.random()` 并不保证唯一，该断言在同毫秒同随机数时并不成立。

## 影响

启动阶段 `main` 与 `bg` 两个 JS 运行时可能同时为同一张内置壁纸/图片计算 base64（正是本函数的典型调用场景），构成同一 `source` 的并发调用；即使在单一运行时内，同一资源的并发读取也同样存在该竞态。命中后表现为 `getRNLocalImageBase64` 抛出 `failed`、图片 base64 缺失。概率虽低，但注释给出的“唯一性保证”并不可靠。

## 建议

引入模块级单调自增计数器保证同一运行时内绝对唯一，并保留 timestamp/random 以降低跨运行时碰撞概率，例如：

```ts
// module scope
let tempResourceSeq = 0;
// ...
const copiedUri = `${cacheDir}bundled-resource-${source}-${timestamp}-${random}-${tempResourceSeq++}`;
```

这样可将同运行时内碰撞降为零，跨运行时碰撞降至可忽略，也使注释中的唯一性描述成立。

[Claude session ↗](https://claude.ai/code/session_01MsegPMd5ockY5rVVh3pGcp)

---

huhuanming: 已在 e57185e75c 修复：增加模块级单调序号并将其写入临时文件名，保证同一 JS runtime 内并发调用不会碰撞；同时保留时间戳，并把随机空间扩大到 `Number.MAX_SAFE_INTEGER`，将 Android main/bg 两个隔离 runtime 之间的碰撞概率降到可忽略。注释也已补充说明 counter 不跨 runtime 共享，以及碰撞会导致覆盖/提前删除的原因。目标文件的格式检查与 type-aware lint 均已通过。

sidmorizon (approved):

originalix (approved):

PR #12555, opened 2026-07-20
feat: add global availability client metrics

originalix:
## Summary
- Client **API** availability metrics, sent as one aggregated `availabilitySnapshot` event at most once every 2 hours per JS runtime (so at most 12 per runtime per day). Event volume never scales with traffic or failure storms.
- Per-service counters, plus dedicated counters for the endpoints behind the blocking criteria in the availability requirements doc (broadcast, fee estimate, fiat entry, swap quote stream and build, earn, perps config, transaction security check, login, IP Table config).
- Breakdowns by network type, IP Table route, proxy and IP Table state.
- Fixes: extension double counting through the XHR shim; Android background runtime never receiving "app hidden".
- Fixes a build-time bug that silently disabled parts of this feature in every native bundle shipped so far: `platformEnv.isJest` is a compile-time define resolved from `JEST_WORKER_ID`, which Metro's worker-thread transformer sets, so `if (platformEnv.isJest) return;` compiled to `if (true) return;`. The network-type tracker never ran, and the aggregator's 60s tick and visibility flush were never wired. This PR removes its own dependence on that flag; the build-wide cause is tracked separately (see Risk).
- No server changes, no remote switch. Hyperliquid is measured for reachability only.

## Event
`app.network.availabilitySnapshot` (filter `schemaVersion >= 4`):

| Properties | Meaning |
|---|---|
| `api_<service>_<ok\|api_error\|http_error\|network_error\|timeout>`, `api_<service>_slow_outcome` | All counted requests per service. Services: `wallet swap utility lightning earn notification prime rebate transfer` on OneKey API hosts, `onekeyApiOther` (other OneKey API hosts, incl. the IP Table config CDN), `hyperliquid`, `supabase`, `onekeyWeb` (`data.onekey.so` firmware manifest, `auth.onekey.so` keyless auth) |
| `api_endpoint_<name>_<ok\|error\|failed\|slow_outcome>` | Critical endpoints, see below |
| `api_net_<wifi\|cellular\|ethernet\|vpn\|offline\|other\|unknown\|unsupported\|unread>_<ok\|error\|failed>` | Network type at completion. `unknown` = the platform reported an unidentified transport, `unsupported` = a browser realm with no transport API (desktop and web Chromium expose none), `unread` = the provider produced no reading at all |
| `api_route_<domain\|sni\|fallback\|none>_<ok\|error\|failed\|slow_outcome>`, `api_proxy_<on\|off\|unknown>_*`, `api_ip_table_<enabled\|disabled\|noConfig\|unknown>_*` | Only requests the IP Table adapter handled, excluding the native main runtime (it has no IP Table config) |
| `failures_1..8`, `failuresOmitted`, `countersOmitted` | Top failure keys `api\|service\|status\|route:<endpoint or 3-segment route group>\|errorCode=count`. The two `*Omitted` counts report what did not fit; read them beside the details |
| `schemaVersion`, `snapshotId`, `runtimeScope`, `endpointEnv`, `windowStartTs`, `windowEndTs`, `windowCount` | Meta; every analytics event also carries platform, appVersion, os, tier, deviceTimeZone and GeoIP country |
| `bundleVersion`, `windowBundleVersion` | `bundleVersion` is the JS bundle that **sent** the snapshot. `windowBundleVersion` is the bundle that **recorded** the window; filter on it to ask what a bundle measured. It is absent for windows recorded before it existed |
| `lastFlushIssue` | Why the previous flush did not send (`cannotSend`, `canSendStalled`, `readyStalled`, `claimFailed`, `sendStalled`, `tickArmFailed`, …). Absent when the previous flush was fine. The device-side diagnostic is a debug line that the native logger drops under load, so this carries it into the data |

Critical endpoints:

| Name | Path | Requirement |
|---|---|---|
| `tokenList` | `/wallet/v1/account/token/list` | 资产首页 refresh |
| `estimateFee` | `/wallet/v1/account/estimate-fee(-batch)` | 转账 fee estimate ≤ 3s |
| `sendTransaction` | `/wallet/v1/account/send-transaction` | 转账 broadcast success (lower bound: `error` includes chain rejections) |
| `fiatPay` | `/wallet/v1/fiat-pay/list`, `/url` | 法币 entry API reachability and latency |
| `swapQuote` | `/swap/v1/quote/events` (SSE) | first quote result ≤ 3s; streams abandoned after 3s count as `timeout` |
| `swapBuildTx` | `/swap/v1/build-tx` | Swap build |
| `earnStake`, `earnWithdraw` | `/earn/v2/stake`, `/earn/v2/unstake` + `/claim` | Earn build (withdraw includes reward claims) |
| `perpConfig` | `/utility/v1/perp-config` | Perps config fetch |
| `txSecurityCheck` | `/utility/v1/transaction/check` | risk checks must not block the confirm page |
| `primeLogin` | `/prime/v1/user/login`, `/prime/v1/account/oauth/login` | OneKey ID login endpoints (not the full login flow) |
| `ipTableConfig` | `config.onekeycn.com/data.json` | disaster-recovery config delivery (first attempt) |

## Reading rules
- Resent windows are deduplicated by PostHog (`$insertId` -> `uuid`); `snapshotId` remains as a readable property.
- Transport availability = 1 − (network_error + timeout) / total. `api_error` includes business rejections; `http_error` merges 4xx and 5xx.
- In breakdowns and endpoints, `error` = api_error + http_error (server answered), `failed` = network_error + timeout.
- `slow_outcome` counts outcomes ≥ 3s of **any** status, failures included, so its denominator is the sum of that prefix's status counters, never the `_ok` one. Slow share ≤ 5% ⇔ P95 ≤ 3s. No P95 values.
- Bucket by `windowEndTs`; windows can span restarts. No hourly charts.
- Key parse rule (longest source first): `^(api_endpoint|api_ip_table|api_proxy|api_route|api_net|api)_([A-Za-z]+)_(api_error|http_error|network_error|timeout|error|failed|ok|slow_outcome)$`.
- **Evaluate a bundle by `windowBundleVersion`, not `bundleVersion`.** After an OTA update the first snapshot a device sends is usually the window recorded by the previous bundle: it is sealed and sent on its own, without `windowBundleVersion` (or with the older value), and must not be read as the new bundle's result.
- `endpointEnv = test` means at least one counted request in the window went to a test host, so it marks the whole window as QA traffic.
- `$geoip_country_code` is the **egress** country of the path the runtime used, so segment it by `api_proxy_on_*` / `api_proxy_off_*`: proxied rows are not geographically attributable.
- `api_route_*`, `api_proxy_*` and `api_ip_table_*` have a smaller denominator than `api_*` and `api_net_*` (IP Table adapter traffic only, and never the native main runtime); web and extension emit none of them.

Not counted: cancellations, analytics delivery, `/health` and `/cdn-cgi/` probes, WebSockets, WebViews, CDN images, BLE/USB, native downloads, web embed, extension offscreen. Hyperliquid order rejections arrive as HTTP 200 and count as ok.

## Design
- `availabilityAggregator.ts`: counters, one storage key per runtime. Three triggers: the 60s tick, the send on app hidden (relayed from main to the native background runtime), and a recorded outcome. The last one keeps delivery going where a host timer or a visibility signal is missing (an MV3 service worker has none), is throttled to the tick cadence, and only ships a window at least 60s old — otherwise a fresh runtime, which has never sent, would ship its first outcome alone, before the asynchronous network-type read lands. It is limited to single-instance runtimes; tabs and extension pages share one send time and flush on visibility instead, with the claim serialised by Web Locks and only the wait for the lock bounded. Every wait in a flush is bounded (readiness, analytics, the shared re-read, the claiming write, the send), so a stalled dependency costs one attempt rather than the runtime, and a trigger arriving mid-flush re-runs it instead of being dropped. The send is claimed by storing its send time first; a claim whose write fails is rolled back and retried, because that stored time is the only thing pacing a runtime across a restart. A window restored from another bundle is sealed and sent on its own rather than continued. Nothing is sent or written until stored state is read.
- `availabilityMetrics.ts`: host allowlist, digit-free route groups, critical endpoint map, error classification (timeout signals count as timeouts), breakdowns. Transport failures are classified into fixed cause tokens (`sslhandshake`, `unknownhost`, `connectionrefused`, `netunreach`, `econnreset`, `certificate`, `sockettimeout`, `failedtofetch`) read from the message, the cause and React Native's `request._response`; only fixed tokens are emitted, never matched text.
- Instrumentation: axios interceptor, patched fetch (timed to the end of the body; event streams left to their owner), IP Table adapter, Swap quote stream in `ServiceSwap`.
- Network type reads RNCNetInfo directly on native (explicit `getCurrentState(undefined)`: a zero-argument TurboModule call crashes on iOS).
- The RN `AbortSignal.timeout` polyfill now sets `signal.reason` to a `TimeoutError`, like the web API.

## Deduplication
The snapshot carries its id and window end time as transport fields (`$insertId`, `$timestamp`), which the utility service already maps to PostHog's dedupe `uuid` and event time (`insertId` / `timestamp` in the track DTO). A snapshot whose acknowledgement was lost is therefore counted once, and events are dated at the window end rather than on arrival. No server change is needed. The analytics client strips both fields from the event properties; events that do not carry them are unchanged.

## Server-side follow-up (not in this PR)
- Verify `$geoip_country_code` is populated on this event after release, and confirm on the first resent snapshot that PostHog dedupes it.

## Risk
- **Level**: Medium. Shared request interceptors, the IP Table adapter and the swap quote stream are instrumented; recording never changes request results.
- **Platforms**: Mobile, Desktop, Web, Extension.
- **Device-verified** (iOS 26 and Android 17, test endpoints, PostHog project 207895): delivery of every snapshot seen in device logs; dedupe (`uuid` is a UUIDv5 of `snapshotId`); event time equals the window end; 2h pacing; windows merged across restarts; the 60s tick on both platforms; cause classification (`econnreset`, `sslhandshake`, `sni_request_timeout`); route/proxy/IP Table breakdowns including SNI fallback; no crash with the network-type tracker actually executing.
- **Not yet verified on device**: real `api_net` values from the fixed tracker, `windowBundleVersion` and sealing, `unsupported` on desktop, and the extension. These affect data quality only — a failure leaves a dimension empty, it does not affect requests or cost.
- **Out of scope, more urgent than this PR**: the same `isJest` build-time bug applies bundle-wide. About 134 non-test files read `platformEnv.isJest`, including crypto paths (`core/src/secret/curves/elliptic.ts` native EC acceleration, `encryptors/aes256.ts` raw-password guard, pbkdf2 and credential cache clearing in `secret/index.ts`) and `babel-plugin-transform-remove-console` in `development/babelTools.js`. Tracked as a separate task; it needs its own security review.

## Test plan
- [x] `yarn tsc:only`, oxlint and oxfmt; `yarn agent:check --profile commit`
- [x] Aggregator, metrics, network type, interceptor wiring (XHR shim single count, event-stream skip, timeout signal; mutation-checked) and IP Table adapter suites; related request, polyfill, travel mode, ServiceSwap, ServiceApp, ServiceIpTable and provider container suites
- [x] Local iOS union build with the strict module-id registry, startup graph budgets and bundle architecture check
- [x] iOS and Android OTA bundles on device: no crash; snapshots delivered and deduplicated; 60s tick running on both platforms
- [x] Reproduced the `isJest` compile-time bug with `@babel/core` and the mobile babel config under `JEST_WORKER_ID`, and confirmed the fixed tracker, tick and visibility wiring compile intact under the same condition
- [ ] `api_net` resolves to wifi/cellular in snapshots with `windowBundleVersion` set (QA data, next 2 days)
- [x] Background the app after 2 hours: one snapshot per runtime, `runtimeScope=bg` on Android
- [ ] Extension: one API request and one swap quote produce one count each


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-59c3c071-fc63-5389-8e45-e8d89e8b1266)

sidmorizon:
Claude review session: https://claude.ai/code/session_01EdL2KFhiMJE57STzAdXEd4

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+        this.record({
+          source: 'flow',
+          target: flow,
+          status: result.status,
+          durationMs: this.deps.now() - startedAt,
+          detail: result.detail ?? options?.detail,
+          errorCode: result.errorCode,
+        });
+        if (trackUnfinished) {
+          this.changeInflight(flow, -1);
```
sidmorizon, line 855: 🟡 P2: Completed flows disappear after crashes

Severity: `non-severe`

When a tracked flow finishes, `finish` persists the cleared in-flight count before persisting its outcome. A crash during the debounce loses both completion and `unfinished`.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+            trackedSetInterval('availabilityAggregator', fn, ms),
+          setTimeout: (fn, ms) => setTimeout(fn, ms),
+        },
+    subscribeVisibility: onVisibilityStateChange,
+    canSend: waitUntilAnalyticsCanSend,
+    send: (params) =>
+      defaultLogger.app.network.reportAvailabilitySnapshot(params),
+    withSharedBudgetLock: lockManager
+      ? (task) => lockManager.request(budgetKey, task)
+      : undefined,
```
sidmorizon: 🟡 P2: Concurrent tabs bypass send pacing

Severity: `non-severe`

Concurrent tabs can read the same due `lastSendTs` before either writes. Both then send snapshots within the two-hour gap.

packages/kit/src/components/WebView/DesktopWebView.tsx
```diff
@@ -265,6 +278,20 @@ const DesktopWebView = forwardRef(
           const failedUrl = event?.validatedURL ?? event?.url;
           if (event.isMainFrame) {
             clearLoadTimeout();
+            // ERR_ABORTED (-3) means the load was superseded or stopped, and a
+            // failure for an older URL must not settle the current attempt.
+            if (
+              event.errorCode !== -3 &&
+              failedUrl === availabilityTimingRef.current?.url
+            ) {
```
sidmorizon, line 286: 🟡 P2: Stale failures disable current timeout

Severity: `non-severe`

A stale main-frame failure clears the current load timeout but leaves `availabilityTimingRef` unsettled. A hanging current navigation then never times out.

packages/shared/src/request/axiosInterceptor.ts
```diff
@@ -280,6 +286,72 @@ function logNetworkThrottleRequestTiming({
...
+    // Third-party hosts do not follow the OneKey `code` envelope.
+    reportApiAvailabilityResult({ status: 'ok', timing });
+    return;
+  }
+  const responseCode = (
+    response.data as IOneKeyAPIBaseResponse | null | undefined
+  )?.code;
+  reportApiAvailabilityResult({
+    responseCode,
+    status: responseCode === 0 ? 'ok' : 'api_error',
```
originalix, line 350: P2: [Raw successful OneKey responses are reported as API errors]

When a successful OneKey endpoint deliberately returns a raw body, `response.data?.code` is `undefined`. For example, `/rebate/v1/invite/export` uses `responseType: 'text'` with `autoHandleError: false`. This branch therefore records an accepted 2xx response as `api_error`, which pollutes availability snapshots and makes a healthy service look unavailable.

Treat a response as an envelope error only when it actually carries a numeric `code`; classify accepted non-envelope bodies as `ok`.

originalix:

Review of commit `054a96bbe753` finished, but some comments could not be published automatically.

packages/shared/src/request/helpers/ipTableAdapter.ts
```diff
@@ -1028,7 +1053,16 @@ export function createIpTableAdapter(
...
-  return async (config: InternalAxiosRequestConfig) => {
+  const ipTableAdapter = async (config: InternalAxiosRequestConfig) => {
+    // Every route taken below overwrites this.
+    markApiAvailabilityRoute(config.$oneKeyAvailabilityTiming, 'none');
+    if (config.$oneKeyAvailabilityTiming) {
+      // Read on every route: proxy-on requests never reach IP selection.
+      void readIpTableAvailabilityState().then(
+        setAvailabilityIpTableState,
+        () => undefined,
+      );
```
sidmorizon, line 1064: 🟡 P2: IP Table state crosses requests

Severity: `non-severe`

Concurrent requests update shared `ipTableState` asynchronously. A response can therefore report another request's IP Table state.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,448 @@
...
+
+  flush(reason: 'hidden' | 'tick'): Promise<void> {
+    this.start();
+    // Before coalescing: the process may be killed soon after it is hidden.
+    if (reason === 'hidden' || this.dirty) this.persist();
+    this.flushing ??= this.flushInternal()
+      .catch(() => undefined)
+      .finally(() => {
+        this.flushing = undefined;
+      });
```
sidmorizon: 🟡 P2: Early backgrounding loses counters

Severity: `non-severe`

Hidden during hydration makes `persist` return while `canWrite` is false. If sending is not due, termination discards the unstored window.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,457 @@
...
+    const state: IStoredState = { version: 2, lastSendTs: this.lastSendTs };
+    if (this.deps.persistWindows) {
+      state.current = this.current;
+      state.pending = this.pending;
+    }
+    // Serialized now, so later changes to the windows are not written early.
+    const text = JSON.stringify(state);
+    this.writeQueue = this.writeQueue
+      .then(() => this.deps.storage.save(text))
+      .catch(() => undefined);
```
sidmorizon: 🟡 P2: Failed persistence permits untracked sends

Severity: `non-severe`

When `storage.save()` fails, `writeQueue` swallows the error and the snapshot still sends. Process death can then lose or reload stale counters.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,694 @@
...
+  /**
+   * Multi-instance runtimes store only the send time, when sending. Resolves
+   * false when the write did not land, so the caller can decide: the stored
+   * send time is the only thing that paces this runtime across a restart.
+   */
+  private persist(sending = false): Promise<boolean> {
+    if (!this.canWrite || (!this.deps.persistWindows && !sending)) {
+      return Promise.resolve(false);
+    }
+    this.dirty = false;
```
sidmorizon, line 693: 🟡 P2: Failed writes strand current counters

Severity: `non-severe`

When storage rejects a normal write, `persist` leaves `dirty` false. Later ticks skip persistence, so process termination before sending loses those counters.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,694 @@
...
+      }
+      window = this.pending;
+    };
+    const lock = this.deps.withSendLock;
+    const held = lock
+      ? await withTimeout<boolean | typeof STALLED>(
+          lock(claim).then(() => true),
+          FLUSH_STEP_TIMEOUT_MS,
+          STALLED,
+        )
```
sidmorizon: 🟡 P2: Late locks claim unsent windows

Severity: `non-severe`

When `withSendLock` exceeds ten seconds, its queued task keeps running after `flushInternal` returns. It can claim the slot without sending, blocking snapshots for two hours.

packages/shared/src/request/fetchInterceptor.ts
```diff
@@ -100,9 +133,32 @@ const newFetch = async function (
...
+              reportApiAvailabilityResponse({
+                timing: availabilityTiming,
+                httpStatus: res.status,
+              }),
+            (e: unknown) =>
+              reportApiAvailabilityError(
+                availabilityTiming,
+                e,
+                options?.signal,
+              ),
```
sidmorizon: 🟡 P2: Request timeouts escape availability counts

Severity: `non-severe`

When `AbortSignal.timeout` belongs to the `Request`, reporting reads only `options.signal`. The resulting `AbortError` becomes a cancellation, so the timeout is omitted.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Obsolete** — this describes the pre-reduction tree.

At `054a96bb`, `finish` called `changeInflight` → `persistBudget()` (an immediate write) while `record` only armed the 5-minute `AVAILABILITY_WINDOWS_PERSIST_DEBOUNCE_MS` timer, so the cleared count really could be durable ahead of the outcome. All of it — flow tracking, the in-flight count, `unfinished`, the debounce — was removed in `82a1f69713` when the scope was cut to API requests only. `changeInflight` and `trackUnfinished` have zero hits in `packages/` today, and `IAvailabilitySource` has no `flow` member.

There is also no structural successor: `createApiAvailabilityTiming` builds an in-memory object and records nothing at request start, and every `recordAvailabilityOutcome` call sits inside `reportApiOutcome` behind a once-guard, i.e. after the request settles. Nothing in-flight is counted or stored.

Persistence is now one atomic value — `persist()` stringifies `{version, lastSendTs, current, pending}` before queueing it — so a cleared counter cannot be durable ahead of the outcome that cleared it.

packages/kit/src/components/WebView/DesktopWebView.tsx
originalix: **Obsolete** — WebView instrumentation was removed entirely in the scope reduction (`82a1f697`).

`git diff origin/x...HEAD -- packages/kit/src/components/WebView/` is empty: both WebView files are byte-identical to `x` on this branch. `availabilityTimingRef` has zero hits in `packages/`.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+      addMetricProp(`${group.prefix}_n`, group.total);
+    }
+    const partitionSeries = sortedSeries.filter(
+      ({ source }) => isBreakdownSource(source) === breakdown,
+    );
+    for (const { prefix, status, series } of partitionSeries) {
+      // A breakdown series without its total is not interpretable.
+      if (
+        (breakdown && !selectedPrefixes.has(prefix)) ||
+        !addMetricProp(`${prefix}_${status}`, series.count)
```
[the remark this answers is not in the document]
originalix: **Obsolete** — group selection was removed in the scope reduction (`82a1f697`). `AVAILABILITY_MAX_GROUPS`, `selectedPrefixes`, `p95_ms` and the `_n` totals all have zero hits in `packages/` today, so orphan statuses cannot occur: the snapshot now emits a flat, closed counter vocabulary.

The underlying worry — silent budget exhaustion corrupting the data — was real, and it is now addressed rather than merely deleted. `addCount` still refuses a new key past `MAX_COUNTERS = 200`, but that overflow is no longer invisible: `countersOmitted` is reported alongside `failuresOmitted` as of `fab16648e8`, so a truncated window is readable as truncated. The comment on the constant now carries the measured worst case (193 keys per runtime, 34 other properties) instead of an estimate, with a note that two more service or endpoint labels would overflow it.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+  const budgetKey = `onekey_availability_metrics_budget_${suffix}`;
+  const windowsKey = `onekey_availability_metrics_windows_${suffix}`;
+  const lockManager = getLockManager();
+  return new AvailabilityAggregator({
+    now: () => Date.now(),
+    createId: () => generateUUID(),
+    persistWindows: Boolean(
+      platformEnv.isNative ||
+      platformEnv.isDesktop ||
+      platformEnv.isExtensionBackground,
```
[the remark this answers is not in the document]
originalix: **Not an issue** — the chain breaks at its first step: travel-mode counters can never be persisted, so there is nothing to upload after switching back.

Every `appStorage` operation is wrapped by `appStorageBuildFactory.ts`, which routes it through `travelModeManager.getRuntimeEnvironmentSync().persistence.run({ operation, onBlocked })`. In travel mode `setItem` takes `onBlocked` and the write is dropped; `getItem` returns `null`. The aggregator's storage is exactly that `appStorage`, so its writes are already no-ops one layer down — no aggregator-level check is needed, and adding one would duplicate an existing guard and pull a travel-mode dependency into the request path.

A correction to my own reasoning while checking this, since it matters for the runtime topology: the main-runtime analytics init is *not* blocked via `BackgroundApiBase.handleInternalMethods`, which is extension-only (`isExtensionInternalCall` requires `platformEnv.isExtension`). On native the gate is `apps/mobile/background.ts:192`, where every `service-call` from main is wrapped in `travelModeCommandDispatcher.runTransportServiceCall`. `getInstanceId` is absent from `TRAVEL_MODE_ALLOWED_METHODS.serviceSetting` (`travelModeCommandPolicy.ts:116-123` lists only `refreshLastActivity`, `setCurrency`, `setHapticFeedbackEnabled`, `setLocale`, `setSelectedBrowserTab`, `setTheme`), so `LastActivityTracker` throws before reaching `analytics.init`.

So both halves hold independently: nothing is sent, and nothing is stored to be sent later.

The comment that led here was misleading and has been fixed in `6a2b53e2b4`. The header of `availabilityMetrics.ts` said only *"Travel mode is not a dimension because analytics is not initialized while it is on"* — naming the send-side guard alone. It now names the storage-side one too.

packages/shared/src/request/axiosInterceptor.ts
originalix: **Already fixed**, in `c8298e12da` — thanks, this was a real bug when you filed it.

`reportApiAvailabilityResponse` now treats a 2xx as an envelope error only when the body actually carries a numeric `code`. A raw body — `/rebate/v1/invite/export` with `responseType: 'text'` and `autoHandleError: false` — has `apiCode === undefined` and is counted `ok`, exactly as you asked.

Covered by `availabilityMetrics.test.ts` → *"counts bodies without a numeric OneKey code as ok"*.

packages/shared/src/request/helpers/ipTableAdapter.ts
originalix: **Not an issue in practice**, though the mechanism you describe is real: `ipTableState` is one module-level value read at report time, so concurrent requests do read whatever is current rather than a per-request snapshot.

It is a *per-runtime configuration* state — `enabled` / `disabled` / `noConfig` / `unknown` — not per-request state. It reflects whether the IP Table config was fetched and whether the feature is on, which is a property of the runtime, and `readIpTableAvailabilityState` is memoized for ~5s on top of that. Two requests in flight at the same moment resolve to the same value except in the one transition tick where the config first arrives, and on that tick either value is a truthful description of the runtime.

Threading it per request would mean carrying it on every timing object to fix a counter that cannot be materially wrong, so I have left it. Happy to revisit if you have a case where the transition window is wide enough to matter.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Confirmed and fixed** in `6a2b53e2b4`. You were right, and the re-read I had added did not close it.

`flushInternal` re-reads the shared `lastSendTs` before sending on `persistWindows: false` runtimes, but there are `await`s between that read and the write that claims the slot, and no cross-instance exclusion in between — so it only closed the sequential case. I reproduced it against the real module: two instances sharing one store with `Promise.all([a.flush('hidden'), b.flush('hidden')])` → 2 sends, three instances → 3 sends, sequential control → 1 send. The existing test only covered the sequential ordering.

This also turned out to be the one piece of the deleted budget machinery with no replacement: `navigator.locks` was used at `054a96bb` and removed in `82a1f69713` with nothing put back.

The fix restores exclusion around the claim only:

- a new optional `withSendLock` dep, wired in `createRuntimeAggregator` to `navigator.locks.request(key, task)` where Web Locks exist. Per-origin is exactly the sharing scope — one namespace for the app's tabs, one for the extension's pages. It is `undefined` on React Native, so native and desktop, which own their storage key, are unchanged.
- the shared re-read, the due check, taking the window and storing the new send time all moved into one `claim()` closure that runs under the lock. **`deps.send()` stays outside it**, so a slow network call never blocks another instance.
- lock acquisition is bounded by the same `withTimeout` the other waits use and fails closed (`lockStalled`).

Covered by `availabilityAggregator.test.ts` → *"lets only one of two instances claim the same shared send slot"*, which asserts `tabA.sent.length + tabB.sent.length === 1` under concurrent flushes.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+    const { scheduler } = this.deps;
+    if (!scheduler) return;
+    scheduler.setInterval(() => {
+      void this.flush('interval');
+    }, AVAILABILITY_FLUSH_CHECK_INTERVAL_MS);
+    try {
+      this.deps.subscribeVisibility?.((visible) => {
+        if (!visible) {
+          void this.flush('hidden');
+        }
```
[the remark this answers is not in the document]
originalix: **Confirmed and fixed** in `6a2b53e2b4` — your conclusion was right, though via a guard neither of us named.

The mechanism you described (the write is only appended to `writeQueue`, so the callback returns before it lands) is accurate but benign once the aggregator has hydrated: `persist()` serializes the state *before* queueing it, so no already-counted outcome can be lost to a late write.

The real loss is earlier. `persist()` begins `if (!this.canWrite || ...) return;`, and `canWrite` stays false until the initial `storage.load()` resolves. So when the app is hidden before that read lands, the synchronous `persist()` at the top of `flush()` returns at the guard — nothing serialized, nothing queued, `dirty` never cleared — and `flushInternal` then bails at `notDue` in the normal case, so the send-path write never runs either. **The hidden trigger wrote zero bytes**, which is exactly your "does not achieve its purpose", and the loss is not one debounce window but everything counted since process start.

Fix: `start()` now chains the hydrate so a dirty window is persisted as soon as the read resolves.

```ts
this.ready = this.hydrate().then((ok) => {
  if (ok && this.dirty) void this.persist();
  return ok;
});
```

Costs one extra write per process, and it is automatically a no-op on multi-instance runtimes, i.e. it writes only for the runtimes that can lose counters. I did not weaken the `canWrite` gate — it is what stops a write landing between "read resolved" and "stored windows merged in", and what keeps a runtime whose read failed from inventing `lastSendTs = 0` and sending past the 2h gap.

The remedy as literally written is not available, for the record: `appVisibility.ts` uses `AppState.addEventListener('change', cb)` and React Native discards the listener's return value, so awaiting inside the callback delays nothing; the bg side is reached through an async proxy RPC that has already returned.

Covered by *"writes counters recorded before stored state was read"*.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Confirmed and fixed** in `6a2b53e2b4`. Same root cause as the Codex finding on the hidden path — one fix closes both.

A window recorded before `storage.load()` resolves was never written: the hidden-path `persist()` is refused by `canWrite`, and nothing re-persisted once hydrate completed, so if the send was not due the runtime could be suspended holding the only copy. `start()` now re-persists a dirty window as soon as hydrate resolves.

Two bounds worth stating precisely: the loss needed *both* a flush trigger inside the read window *and* the send not being due when hydrate resolved — if it is due, the send path writes before sending. And the native bg runtime does receive a hidden signal (the main runtime relays it via `flushAvailabilitySnapshotOnHidden`), so this was never about bg missing the event.

Covered by *"writes counters recorded before stored state was read"*.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Confirmed and fixed** in `6a2b53e2b4`.

`persist()` swallowed a rejected `storage.save()` while `flushInternal` treated the write as landed and sent anyway. That stored write is not bookkeeping — it is the *only* thing carrying the 2h gap across a process boundary, so on a storage-unhealthy device a restart would send again inside the gap, and that resend is a new window with a new id, so `$insertId` does not deduplicate it. The daily cap offered no protection there.

Fix: the write now claims the send.

- `persist()` returns whether the write actually landed.
- a send that is not claimed does not go out: `lastSendTs` is rolled back to its previous value so the window is still due, and the next trigger retries. It reports `claimFailed` / `claimStalled` through the new diagnostic line rather than failing silently.

While here I also bounded the two `await`s in `flushInternal` that were still unbounded — `await this.writeQueue` and the shared-key `storage.load()`. That is the same failure class the previous commit claimed to have closed (an await that never settles latches `this.flushing` and silences the runtime for its whole life), and I had missed these two. They now time out and report `loadStalled` / `claimStalled`.

Covered by *"does not send when the write that claims the send fails"*, which also asserts the window ships once storage recovers.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,1403 @@
...
+      return;
+    }
+    this.windowsPersistTimer = scheduler.setTimeout(() => {
+      this.windowsPersistTimer = undefined;
+      this.persistWindows();
+    }, AVAILABILITY_WINDOWS_PERSIST_DEBOUNCE_MS);
+  }
+
+  private enqueueWrite(write: () => Promise<void>) {
+    this.writeQueue = this.writeQueue.then(write).catch(() => undefined);
```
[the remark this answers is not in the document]
originalix: **Split verdict.** The budget machinery is gone; the underlying concern was real and is now fixed.

`saveBudget`, `reserveSend`, `budgetUsed` and `lastAttemptTs` were all removed in `82a1f69713` when the scope was cut to API requests — zero hits in `packages/` today — so that specific code path is **obsolete**.

The structural concern survived it, and it is the same one raised separately on the current code (see the thread on the `writeQueue` catch): a rejected `storage.save()` was swallowed while the snapshot went out regardless. Your named harm was half right on HEAD. A resent *pending* window carries the same `$insertId` and is deduplicated server-side, so that half does not apply. But the broken cap does: the stored send time is the only thing pacing a runtime across a restart, so a failed write plus a restart produces a second billable event inside the same 2h gap, under a fresh id the server cannot collapse.

Fixed in `6a2b53e2b4`: `persist()` now reports whether the write landed, and a send that is not claimed by a successful write is rolled back and retried on the next trigger instead of going out unpaced. I also bounded the two remaining unbounded `await`s in the flush (`writeQueue`, and the shared-key re-read), which could otherwise park a runtime permanently.

packages/shared/src/request/availabilityAggregator.ts
```diff
@@ -0,0 +1,794 @@
...
+    // visibility flush. Tests construct the aggregator with their own deps.
+    setInterval: (fn, ms) =>
+      trackedSetInterval('availabilityAggregator', fn, ms),
+    subscribeVisibility: onVisibilityStateChange,
+    // Web Locks are per origin, which is exactly the sharing scope: one
+    // namespace for the app's tabs, one for the extension's pages. Absent on
+    // React Native, where each runtime owns its key and needs no exclusion.
+    withSendLock: globalThis.navigator?.locks
+      ? (task) => globalThis.navigator.locks.request(key, task)
+      : undefined,
```
sidmorizon, line 774: 🟡 P2: Unlocked tabs duplicate snapshots

Severity: `non-severe`

Without `navigator.locks`, concurrent tabs run `claim` without exclusion and read the same due timestamp. Both send distinct snapshots, exceeding the shared two-hour pacing limit.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Confirmed and fixed** in `a858b0e4ab`.

`persist()` cleared `dirty` before queueing the write, and a rejected `storage.save()` left it cleared, so the next tick saw nothing to write and a process killed before sending lost those counters. The rejection handler now sets `dirty` back to `true`, so a later tick retries the write.

Covered by *"retries a failed write on a later tick instead of keeping counters in memory"*, which fails with the fix reverted.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Confirmed and fixed** in `a858b0e4ab`.

`withTimeout` wrapped the whole `lock(claim)` promise, so when it gave up the queued task was still waiting for the lock, and on a late grant it ran `claim()` anyway — storing a send time with nobody left to send and blocking every instance for the gap.

Now only the *wait* is bounded. The locked task marks itself `granted` as its first synchronous step and returns immediately if the attempt was already `abandoned`. Because the check and the flag are set in the same synchronous run, there is no interleaving: either the grant came in time, in which case the claim runs to completion (each of its steps — the shared re-read and the claiming write — is already bounded) and the send proceeds; or it came late, in which case it claims nothing.

Covered by *"claims nothing when the send lock is granted after the attempt gave up"*, which fails with the guard removed.

packages/shared/src/request/fetchInterceptor.ts
originalix: **Confirmed and fixed** in `a858b0e4ab`.

The patched `fetch` passed only `options?.signal` to `reportApiAvailabilityError`, so for `fetch(new Request(url, { signal: AbortSignal.timeout(n) }))` the signal was `undefined`, the resulting `AbortError` matched the cancellation branch, and the timeout was never counted. The wrapper now uses `options.signal ?? request.signal`, so an explicit `options.signal` still takes precedence as it does in the platform.

Covered by *"counts a timeout carried by a Request as a timeout, not a cancellation"* in `axiosInterceptor.availability.test.ts`, which fails with the fallback removed.

packages/shared/src/request/availabilityAggregator.ts
originalix: **Acknowledged, leaving as is.**

The race is real where `navigator.locks` is missing: tabs fall back to re-reading the shared send time, which narrows it but does not close it. I'm not adding a second exclusion mechanism for that case because the Web Locks API is available in every runtime this code targets — Chromium 69+ (so Electron and Chrome extension pages and MV3 service workers, where it is exposed to workers too), Firefox 96+ and Safari 15.4+. React Native has no `navigator.locks`, but native runtimes own their storage key and never take this path.

What remains is an occasional extra snapshot in a browser older than those versions, when two tabs flush in the same instant. It stays bounded by the per-tab gap, so it cannot turn into a send storm. If we see web traffic from such browsers in PostHog, a `localStorage`-based lease would be the fallback.

zhaono1 (approved):

PR #12473, opened 2026-07-13
feat: support Permit2 approvals and revocation

yikZero:
## Summary
- Fetch and display Permit2 approval records alongside existing ERC20 approvals.
- Revoke Permit2 allowances through `AllowanceTransfer.approve` while keeping the ERC20 revoke path unchanged.
- Preserve independent selection identities, expiration display, bulk revoke, and confirmation semantics for both approval types.

## Intent & Context
Arbitrum users reported Permit2 allowances visible on revoke.cash but missing from OneKey. The server index is ready and returns Permit2 metadata when `withPermit2` is enabled.

## Root Cause
The client only requested direct token approvals and assumed every revoke transaction called `approve` on the token contract. Permit2 allowances live inside the Permit2 contract and require token, DApp spender, amount, and expiration arguments.

## Design Decisions
- The main runtime owns list rendering and selection; the background runtime owns API access, trusted-target validation, calldata construction, strict calldata decoding, confirmation parsing, and signing.
- Permit2 is allowed only on the existing seven approval networks. The background runtime resolves the target from a local trusted deployment allowlist, requires the API address to match it, and never uses the API value as the signing target.
- Permit2 classification relies only on `permit2Address`, so a direct ERC20 approval granted to the Permit2 contract remains a normal ERC20 approval.
- Expiration milliseconds are normalized to safe `uint48` seconds; invalid or partial metadata fails closed.
- Local confirmation parsing verifies selector, sender, trusted target, value, token, spender, amount, and expiration against the encoded transaction.
- Permit2 confirmations always use locally verified title, approval, spender, and interact-contract semantics; only server simulation components and alerts are retained.

## Changes Detail
- Enable `withPermit2` for approval queries and extend approval data models.
- Add outlined Permit2 badges and finite, never-expiring, and invalid expiration states.
- Add stable selection keys for ERC20 and Permit2 records across single and bulk revoke flows.
- Build Permit2 `approve(address,address,uint160,uint48)` calldata with selector `0x87517c45` and amount zero.
- Validate the Permit2 transaction target against a background-owned deployment allowlist for the existing seven EVM approval networks.
- Add focused regression coverage for expiration normalization, selection identity, calldata, trusted-target validation, and verified local confirmation semantics.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile, Desktop, Web, Extension
- **Risk Areas**: EVM approval transaction construction, trusted signing targets, confirmation parsing, bulk selection identity

## Test plan
[...]

PR #12595, opened 2026-07-22
fix: dedupe all-network worth summing and trust PancakeSwap Permit2

weatherstar:
## Summary
- Fix Home total assets showing exactly double on desktop: dedupe token `$key`s when summing per-network account worth, so cache-floor rounds (which reuse one cached tokenListMap for both token groups) are no longer counted twice.
- Fix PancakeSwap Permit2 revocation being blocked: allow multiple trusted Permit2 addresses per network and add the PancakeSwap deployment to the client whitelist.

## Intent & Context
Two hot-update-targeted production issues on v6.5.0:

1. A user reported (support ticket 906970) that ALL assets were double-counted on desktop (¥78,857 shown vs ¥39,388 real, ratio ≈ 2.0), while the same account on mobile was correct. Clearing app cache did not help. Individual token rows and amounts displayed correctly — only the header totals doubled.
2. After the latest hot update, users reported that revoking a PancakeSwap Permit2 approval was rejected by the client. The server-side approvals whitelist returns both Uniswap Permit2 (`0x000000000022d473030f116ddee9f6b43ac78ba3`) and PancakeSwap Permit2 (`0x31c2f6fcff4f8759b3bd5bf0e1084a055615c768`) (together >99.9% of Permit2 approval records), but the client trusted-address list from #12473 only contained Uniswap.

## Root Cause
**Worth doubling** (diagnosed from the user's desktop logs + code trace):
- `useTokenListReactivePipeline.seedAndFlushCache` (introduced in #12068) builds all-network cache-floor rounds passing the ONE cached full `tokenListMap` as BOTH `tokens.map` and `smallBalanceTokens.map` (the local cache stores a single merged map).
- `buildMergedAllNetworkSnapshot` computes each round's worth via `sumTokenGroupsFiatValueIgnoringUnavailable` = sum(tokens.map) + sum(smallBalanceTokens.map), so every cache-floor network's worth is exactly doubled. List rows stay correct because display maps merge by `$key`.
- The user's logs showed why floors never got replaced by live data: the desktop all-network fan-out issues full-concurrency token-list requests, while the desktop SNI direct-IP channel caps 16 active requests per destination and fails closed (`SNI_RESOURCE_LIMIT`, durationMs=0). ~83 of 98 network requests died every refresh round, so high-value networks (tron/evm/sol) stayed on the doubled cache floor permanently. Mobile uses a capped fan-out and does not hit this, which is why only desktop doubled. Clearing cache cannot help because any successful fetch immediately rebuilds the cache and the doubling is in the floor-round construction, not the cached data.
- Numeric cross-check against the user's screenshots: live-settled low-value networks (~¥1,020) single + floored networks (~¥38,412) doubled ≈ ¥77,844, matching the reported "代币: ¥77,849.43" header within rate-refresh noise.

**Permit2 blocking**: `_assertPermit2ApproveInfo` → `resolveTrustedPermit2Address` returned `undefined` for the PancakeSwap contract, flipping `isPermit2InfoValid` to false and throwing before the revoke tx could be built.

## Design Decisions
- **Worth fix at the summing helper, not the floor-round construction**: making `seedAndFlushCache` pick per-group maps would silently drop aggregate-member token values (aggregate members are not in `tokenList` data) and risky-row display depends on the full map — both display regressions. Deduping by `$key` in `sumTokenGroupsFiatValueIgnoringUnavailable` is a strict no-op for live rounds (their maps are disjoint) and exactly corrects shared-map floor rounds. Minimal, release-safe diff.
- **Keep the client Permit2 whitelist instead of trusting the server**: the `permit2Address` becomes the `to` address of the transaction the user signs; dropping the client check would let a compromised server direct users to sign calls to arbitrary contracts presented as a benign "revoke". Both deployments are deterministic CREATE2 addresses shared across chains, and new mainstream Permit2 deployments are rare (two in four years), so a static two-entry whitelist per network is the right cost/risk trade-off. Structure changed from single address to address list per network to make future additions one-line.
- The desktop SNI 16-per-destination fail-closed limit vs full fan-out (which keeps floors alive and data stale) is a separate issue and intentionally NOT addressed here; it should be tracked independently.

## Changes Detail
- `packages/shared/src/utils/tokenValueUtils.ts`: `sumTokenGroupsFiatValueIgnoringUnavailable` now tracks seen `$key`s across both maps and counts each once.
- `packages/shared/src/utils/tokenValueUtils.test.ts`: unit tests — shared-map shape must not double (5150, not 10300); overlapping keys counted once while distinct keys still sum.
- `packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts`: integration test — a cache-floor-shaped round (one map reused for all three groups) yields single-counted per-network worth.
- `packages/kit-bg/src/vaults/impls/evm/approveTransactionUtils.ts`: `TRUSTED_PERMIT2_ADDRESSES` is now per-network address lists containing Uniswap + PancakeSwap Permit2; `resolveTrustedPermit2Address` matches against the list. PancakeSwap address EIP-55 checksum verified with ethers `getAddress`.
- `packages/kit-bg/src/vaults/impls/evm/approveTransactionUtils.test.ts`: PancakeSwap Permit2 resolves on all whitelisted networks, still undefined on non-whitelisted networks.
- `development/spellCheckerSkipWords.txt`: add `pancakeswap`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Mobile / Web / Extension
- **Risk Areas**: The worth-summing helper is shared by the Home overview, single-network path, and WebAccountPanel — all live callers pass disjoint maps, so behavior only changes for shared-map (cache-floor) inputs, covered by tests. Permit2 change only widens the trusted set; validation logic (amount=0, value=0, selector/decode checks) is untouched.

## Test plan
- [x] Unit + integration tests added (red → green): 8/8 approveTransactionUtils, 34/34 tokenValueUtils + buildMergedAllNetworkSnapshot
- [x] Regression: TokenListBlock suite + tokenList cells + tokenUtils — 16 suites / 209 tests pass
- [x] `yarn agent:check --profile commit` passes (lint, lint-staged, tsc)
- [ ] Manual: offline cold start (cache floor only) — Home total must equal the real portfolio value, not double
- [ ] Manual: revoke a PancakeSwap Permit2 approval on BSC/ETH — transaction builds and reaches the confirmation screen
- [ ] Manual: revoke a Uniswap Permit2 approval — unchanged behavior

### DIFF
--- a/packages/shared/src/utils/tokenValueUtils.ts
+++ b/packages/shared/src/utils/tokenValueUtils.ts
@@ -83,6 +83,12 @@ export function sumFiatValuesIgnoringUnavailable(
 // collapses the duplicated `tokens.map + smallBalanceTokens.map` sum at every
 // accountWorth write site. Single pass over both maps to avoid the
 // toFixed → reparse round trip.
+//
+// Each $key is counted ONCE across both maps: cache-floor rounds
+// (useTokenListReactivePipeline.seedAndFlushCache) reuse the ONE cached full
+// tokenListMap as both tokens.map and smallBalanceTokens.map — summing them
+// verbatim doubled every cache-floor network's worth on the Home overview.
+// Live rounds carry disjoint maps, so the dedup is a no-op for them.
 export function sumTokenGroupsFiatValueIgnoringUnavailable(r: {
   tokens?: { map?: Record<string, ITokenFiatValueShape | undefined> };
   smallBalanceTokens?: {
@@ -90,13 +96,17 @@ export function sumTokenGroupsFiatValueIgnoringUnavailable(r: {
   };
 }): string {
   let acc = new BigNumber(0);
+  const seenKeys = new Set<string>();
   const addAll = (
     map: Record<string, ITokenFiatValueShape | undefined> | undefined,
   ) => {
     if (!map) return;
-    for (const entry of Object.values(map)) {
-      if (entry && isValidNumberValue(entry.fiatValue)) {
-        acc = acc.plus(entry.fiatValue);
+    for (const [key, entry] of Object.entries(map)) {
+      if (!seenKeys.has(key)) {
+        seenKeys.add(key);
+        if (entry && isValidNumberValue(entry.fiatValue)) {
+          acc = acc.plus(entry.fiatValue);
+        }
       }
     }
   };
--- a/development/spellCheckerSkipWords.txt
+++ b/development/spellCheckerSkipWords.txt
@@ -640,6 +640,7 @@ Outout
 p2pkh
 p2wpkh
 pageview
+pancakeswap
 passcode
 passpharse
 passphrase
--- a/packages/kit-bg/src/vaults/impls/evm/approveTransactionUtils.test.ts
+++ b/packages/kit-bg/src/vaults/impls/evm/approveTransactionUtils.test.ts
@@ -82,6 +82,39 @@ describe('approval transaction utils', () => {
     ).toBeUndefined();
   });
 
+  test('resolves the PancakeSwap Permit2 deployment on trusted networks', () => {
+    // The server-side approvals whitelist returns Uniswap AND PancakeSwap
+    // Permit2 contracts; only trusting Uniswap blocked PancakeSwap Permit2
+    // revocations at _assertPermit2ApproveInfo.
+    const pancakePermit2Address = '0x31c2f6fcff4f8759b3bd5bf0e1084a055615c768';
+    const networkIdsMap = getNetworkIdsMap();
+    const supportedNetworkIds = [
+      networkIdsMap.eth,
+      networkIdsMap.bsc,
+      networkIdsMap.polygon,
+      networkIdsMap.arbitrum,
+      networkIdsMap.avalanche,
+      networkIdsMap.optimism,
+      networkIdsMap.base,
+    ];
+
+    for (const networkId of supportedNetworkIds) {
+      expect(
+        resolveTrustedPermit2Address({
+          networkId,
+          permit2Address: pancakePermit2Address,
+        }),
+      ).toBe('0x31c2F6fcFf4F8759b3Bd5Bf0e1084A055615c768');
+    }
+
+    expect(
+      resolveTrustedPermit2Address({
+        networkId: networkIdsMap.zksyncera,
+        permit2Address: pancakePermit2Address,
+      }),
+    ).toBeUndefined();
+  });
+
   test('encodes the on-chain Permit2 revoke example', () => {
     expect(buildPermit2Tx()).toEqual({
       from: owner,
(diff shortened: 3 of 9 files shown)


packages/shared/src/utils/tokenValueUtils.ts
```diff
@@ -83,20 +83,30 @@ export function sumFiatValuesIgnoringUnavailable(
...
   let acc = new BigNumber(0);
+  const seenKeys = new Set<string>();
   const addAll = (
     map: Record<string, ITokenFiatValueShape | undefined> | undefined,
   ) => {
     if (!map) return;
-    for (const entry of Object.values(map)) {
-      if (entry && isValidNumberValue(entry.fiatValue)) {
-        acc = acc.plus(entry.fiatValue);
+    for (const [key, entry] of Object.entries(map)) {
```
originalix, line 104: 当 cache-floor 包含风险代币时，`seedAndFlushCache` 会把同时含普通、小额和风险项的完整 `tokenListMap` 复用给 `tokens.map` 与 `smallBalanceTokens.map`。这里按 `$key` 去重只消除了两次遍历造成的翻倍，但仍会把仅属于 `riskTokens` 的 key 加进 `accountsWorth`；而缓存生成时的 `tokenListValue` 明确只统计 `tokens + smallBalanceTokens`，因此失败网络保留 floor 后，authoritative snapshot 仍会把总资产写高。

请把缓存保存的 `tokenListValue` 作为 cache-floor round 的显式 `accountWorth` 传入，并在 `buildMergedAllNetworkSnapshot` 计算缓存 round 时优先使用它；同时补一个完整 map 含 risk-only key 的回归测试，确保该 key 不进入账户净值。

weatherstar: 已按建议修复(2a7e0e4383):

1. `IAllNetworkSnapshotRound` 新增可选 `accountWorth`,`buildMergedAllNetworkSnapshot` 计算每轮 worth 时优先使用它(`isValidNumberValue` 校验通过才生效,否则回退到 map 求和),`createAtNetworkWorth` 累加同样走该值;
2. `seedAndFlushCache` 把缓存的 `tokenListValue`(BG 侧只统计 tokens + smallBalanceTokens,已做货币换算)作为 cache-floor round 的显式 `accountWorth` 传入;`ICacheSeedItem.tokenListValue` 设为可选以兼容 legacy 缓存(缺失时回退 map 求和);
3. 回归测试:`buildMergedAllNetworkSnapshot.test.ts` 新增完整 map 含 risk-only key(scam=999)+ 显式 `accountWorth='5150'` 用例,断言 `accountsWorth`/`createAtNetworkWorth` 均为 5150;`useTokenListReactivePipeline.test.ts` 新增端到端用例(seed → authoritative snapshot),断言 risk-only key 不进入账户净值。

Live 轮次不受影响:`IFetchAccountTokensResp` 无 `accountWorth` 字段,live round 走原 map 求和路径。

originalix (approved):

PR #12554, opened 2026-07-19
feat(perps): optimize quick trading flows (OK-57989)

Minnzen:
[OK-57989](https://onekeyhq.atlassian.net/browse/OK-57989)

----------

## Summary

- Add Perps BBO 1/5 pricing using five valid Hyperliquid price ticks, including confirmation display.
- Add one-shot Chase for eligible Perps GTC open orders using each order’s own market BBO.
- Add a coin-scoped Market/Limit position-increase flow without switching the active asset, reusing the TradingForm size-unit selector.
- Recognize existing position TP/SL from subscribed Open Orders and seed missing legs with editable ±10% ROE defaults.
- Keep new action copy hardcoded in English: `Chase` and `Add Position`; no generated locale files changed.

## Product boundaries

- Mobile position actions are exactly Add Position | TP/SL | Close Position; no market-close-all action was added.
- Chase and Add Position never switch the active asset.
- BBO 5 means a five-valid-tick offset, not the fifth order-book level.
- Position TP/SL opens immediately from account/coin-scoped subscribed Open Orders; there is no loading Spinner, Refresh action, or `frontendOpenOrders` info request.
- TP/SL submission does not perform a read-before-write preflight. After the required trading guard and account/position scope check, it submits directly to the exchange path.

## Implementation notes

- BBO pricing follows Hyperliquid’s five-significant-figure and decimal-limit price grid and fails closed for stale or incomplete BBO data.
- Chase is restricted to ordinary Perps GTC orders, preserves remaining size/TIF/cloid, fetches the order coin’s raw L2 BBO, and sends executable modify semantics with `a: true` only for Chase.
- `@nktkas/hyperliquid@0.32.2` does not expose the Chase modify field, so a minimal patch-package patch extends only the modify schema/runtime declarations.
- Add Position validates the original account, coin, direction, target precision, latest max size, and $10 minimum before calling the coin-scoped order-open path with `reduceOnly=false`.
- Position TP/SL identifies bare `Trigger` position orders independently of serialized size, hides existing legs, and strips legs already present in the subscribed snapshot from the outgoing payload.
- Missing TP/SL legs default to 10% ROE from entry/effective order price and leverage while preserving existing orders, presets, user values, and user-cleared fields.

## Commit boundaries

Primary product boundaries:

1. `acf9348f72` — five-tick BBO pricing
2. `32bb944425` — one-shot Chase
3. `91f6404c40` — coin-scoped Add Position
[...]

PR #12621, opened 2026-07-24
fix(perps): follow up on add position and TP/SL (OK-58389, OK-58391, OK-58396)

Minnzen:
[OK-58389](https://onekeyhq.atlassian.net/browse/OK-58389) [OK-58391](https://onekeyhq.atlassian.net/browse/OK-58391) [OK-58396](https://onekeyhq.atlassian.net/browse/OK-58396)

----------

Follow-ups to #12554.

## OK-58396 — TP/SL seeded at a strict 10% ROE

Percent-derived TP/SL triggers were truncated to five significant figures. Truncation always shifts the trigger toward zero, and leverage amplifies the remainder:

| Entry | Leverage | TP ROE | SL ROE |
| --- | --- | --- | --- |
| 3.4567 | 20x | 9.95% | -10.01% |
| 3.4567 | 40x | 9.95% | -10.07% |
| 0.0012345 | 25x | 9.11% | -11.14% |

Percent-derived triggers now snap to the *nearest* valid price instead (`snapHlPriceToGrid` gained a `nearest` mode). User-entered trigger prices are untouched, and spot paths do not pass `szDecimals`, so their behaviour is unchanged.

Separately, the position TP/SL dialog seeded a price and then reverse-computed the percent field, so a seeded 10% ROE displayed as `9.98`, while typing `10` manually kept `10`. Both paths compute the identical price; only the display differed. `TpslInput` now accepts an optional `seedPercent` and shows the seeded percent verbatim while the leg still sits on its seeded price, falling back to the reverse-computed value once the user moves the price. Only `SetTpslModal` passes it, so every other caller is unaffected.

Where the price grid cannot express 10% exactly, the nearest tick is used — e.g. AAPL at 324.6/20x wants 326.223, the grid allows 326.22 (9.98%) or 326.23 (10.04%), so 326.22 wins. The percent field still reads `10` because that is the requested intent, matching manual entry.

## OK-58391 — Explain unavailable Add Position amounts

The confirm button was disabled whenever the size failed validation, so a below-minimum or over-balance amount gave no reason at all, and the max-addable row beside it only restated the balance.

The button now stays pressable and toasts on press, matching the main trading panel: below-minimum and insufficient-margin are separate messages, and the minimum is expressed in the size unit currently selected (`$10`, `0.10 ETH`, or the margin equivalent). The max-addable row is gone.

## OK-58389 — Size slider in Add Position

Add Position only accepted a typed size. It now reuses the trading panel's slider (4 segments, tap-to-snap), mapping 0-100% onto the max tradable size; typing or focusing the size field returns to manual entry. On submit the slider size is re-resolved against the refreshed max, so a 100% drag cannot fall out of range while the dialog is open.

## Validation

- 42 suites / 396 tests pass across `views/Perp` and the shared perps utils, including new coverage for nearest-tick snapping, realized-ROE drift bounds, the minimum-amount label per size unit, and an invariant test asserting the seeded price equals the manual percent→price conversion (the `seedPercent` display depends on that equality, and would silently regress to `9.98` if the two ever diverged).
- `yarn agent:check --profile commit` passes; it was run per commit, and each commit is independently coherent.

## Not verified

Runtime behaviour of both dialogs was not exercised here — they require a connected Hyperliquid account with an open position, which this environment cannot drive. Correctness rests on the unit tests and type checks above. The author confirmed both dialogs behave correctly in manual testing.

Still worth a look on device:
- Slider tap-snap on iOS/Android, and the manual/slider hand-off.
- Toast copy for each size unit.
- Seeded TP/SL reading `10`, reverting to the real percent after a manual price edit, and chart-originated preset prices staying untouched.

[OK-58389]: https://onekeyhq.atlassian.net/browse/OK-58389?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-58391]: https://onekeyhq.atlassian.net/browse/OK-58391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-58396]: https://onekeyhq.atlassian.net/browse/OK-58396?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/AddPositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/AddPositionModal.tsx
@@ -30,6 +30,7 @@ import { ETranslations } from '@onekeyhq/shared/src/locale';
 import {
   formatPriceToSignificantDigits,
   parseDexCoin,
+  resolveTradingSize,
 } from '@onekeyhq/shared/src/utils/perpsUtils';
 import { EPerpsSizeInputMode } from '@onekeyhq/shared/types/hyperliquid';
 import type {
@@ -47,6 +48,7 @@ import {
   PERP_DIALOG_BUTTON_SIZE,
   PERP_MOBILE_DIALOG_CONTENT_CONTAINER_PROPS,
 } from '../PerpDialogLayout';
+import { PerpsSlider } from '../PerpsSlider';
 import { TradingGuardWrapper } from '../TradingGuardWrapper';
 import { PriceInput } from '../TradingPanel/inputs/PriceInput';
 import { SizeInput } from '../TradingPanel/inputs/SizeInput';
@@ -94,6 +96,10 @@ const AddPositionForm = memo(
 
     const [orderType, setOrderType] = useState<IAddPositionOrderType>('market');
     const [amount, setAmount] = useState('');
+    const [sizeInputMode, setSizeInputMode] = useState<EPerpsSizeInputMode>(
+      EPerpsSizeInputMode.MANUAL,
+    );
+    const [sizePercent, setSizePercent] = useState(0);
     const [limitPrice, setLimitPrice] = useState('');
     const [hasTpsl, setHasTpsl] = useState(false);
     const [tpType, setTpType] = useState<'price' | 'percentage'>('price');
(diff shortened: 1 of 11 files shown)


packages/kit/src/views/Perp/components/OrderInfoPanel/AddPositionModal.tsx
```diff
@@ -418,13 +542,25 @@ const AddPositionForm = memo(
...
-          sliderPercent={0}
+          onChange={handleManualSizeChange}
+          sizeInputMode={sizeInputMode}
+          sliderPercent={sizePercent}
+          onRequestManualMode={switchToManual}
           leverage={leverage}
           allowMarginInput
           ifOnDialog
         />
+        <PerpsSlider
```
sidmorizon, line 553: # [🟡 P2] 滑块模式下缺少下单数量/名义金额的数字反馈

## 问题

新增滑块的同时删除了原来的最大可开数量提示行（`Max: {maxSize} {displayName}`）。而在滑块模式下 `SizeInput` 的输入框只显示百分比（`sliderDisplay` 返回 `` `${sliderPercent}%` ``，见 `SizeInput.tsx` 的 `displayValue`），并不会显示换算出的 token / USD / 保证金数量；`PerpsSlider` 又设置了 `showBubble={false}`。

Add Position 对话框在 `handleSubmit` 里直接调用 `placeOrderByCoin` 下单，没有二次确认弹窗（不同于主面板走 `useOrderConfirm` 的确认流程）。

## 影响

用户用滑块开仓时，整个对话框里唯一的数量反馈就是一个 `50%`，既看不到实际会成交多少 token、约合多少 `USD`，也失去了原有的最大值参照。`50%` 对应的是 `maxTradeSzs` 的一半（可能是很大的杠杆仓位），用户容易误判规模并一键下单，属于交易对话框里实实在在的可用性/安全隐患。

## 建议

在滑块下方（或复用被删除的那一行）展示由 `resolvedSize` 换算出的实际数量与名义金额，例如 `≈ {resolvedSize} {displayName}` / `≈ ${resolvedSize * effectivePrice}`，或保留一行 `Max` 参照，让用户在下单前能看到绝对数量。

[Claude session ↗](https://claude.ai/code/session_01XhBysncA56cCnmFenD8VCr)

---

sidmorizon (commented):
1 finding (P2). See inline comment.

---

originalix (approved):

PR #12637, opened 2026-07-27
feat: rebuild KeyTag backup and import with forced verification (OK-58354)

franco-chan:
[OK-58354](https://onekeyhq.atlassian.net/browse/OK-58354)

----------

## What

Rebuilds the KeyTag backup and import flows around one shared interactive plate, and stops the backup from certifying a wallet on the user's word alone.

**Backup** becomes review → verify → success. Review shows the read-only dot map to copy onto the plate; verify swaps in a blank interactive plate, so the answer is off-screen and the only faithful copy on hand is the punched plate. Each row is compared to the wallet's own per-row values, and `wallet.backuped` flips only when every row matches. Leaving at any point is allowed and leaves the wallet not-backed-up, so an abandoned verify is honest rather than a silent false positive. Verification also runs on the enter-phrase branch, where it checks the plate without marking any wallet.

A valid-but-wrong word was previously indistinguishable from a correct one — the exact error that loses funds — so the plate gained a mismatch mask that flags those rows red, and failures name the offending row numbers instead of relying on colour alone.

**Import** drops the word-by-word editor and the password gate, keeps dots-only entry on every platform, and a failed Confirm now flips to the face holding the first unresolved row instead of dead-tapping.

Both flows hold the screen awake for the minutes-long hands-on task, tick haptically per toggle, and share `KeyTagInputBoard` for the plate, flip card and placeholder padding.

## Copy corrections

Audited against OneKey's own product page and help centre:

- The KeyTag is **aerospace-grade TC4 titanium alloy**, but the new copy called it steel, and that had already shipped into Chinese as 钢板. The app's own settings string had it right, so the feature contradicted itself one screen apart.
- "Place your KeyTag over the pattern shown" asked for a physical overlay that cannot work — the on-screen plate is elastically sized and capped at 400px, never 1:1 with the 36x33.5mm tag. It now states the per-row correspondence, matching OneKey's own guide.
- Import said the phrase is "verified" on confirm. Confirm only runs the BIP39 checksum: that proves the dots decode to a valid phrase, not that they match the plate. The word collided with the meaning the verify step teaches one flow away.
- The mispunch help called a mispunched plate unusable. A missing dot can still be punched; only an extra one is unrecoverable.

19 KeyTag keys were missing all 17 non-Chinese translations and had been falling back to English (`i18n:add` writes only `en_US` and `zh_CN`, and these keys were created after the last batch translation run). Now complete.

## Test plan

QA cases live in OK-58354. Verified in the web preview:

- 12-word and 18-word backup → verify → success, including the two-face flip and 15/18/21-word placeholder rows
- mismatch detection catches a valid-but-wrong word (a mis-tapped row 12 was flagged red and named in the toast)
- import golden path, incomplete-Confirm auto-flip, word-count shrink guard
- already-backed-up wallet still shows a plain Done; a wallet awaiting backup shows Verify

Not yet verified on a real device: flip animation, haptics, keep-awake, screenshot protection.

## Notes for review

- Non-CJK translations (de/es/fr/it/pt/ru/uk/vi/th/id/hi/bn) are machine-produced and want a native review pass — particularly the line distinguishing a missing dot (fixable) from an extra one (not).
- Follow-ups deliberately left out: hoisting `GridCell` into `plate.tsx` (needs the two border models reconciled first) and cell-level accessibility on the 144-cell grid.

Fixes OK-58354

[OK-58354]: https://onekeyhq.atlassian.net/browse/OK-58354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/views/KeyTag/pages/BackupRecoveryPhrase/index.tsx
```diff
@@ -1,70 +1,144 @@
...
+      const { mnemonic } = await phaseInputAreaRef.current.submit();
+      navigation.push(EOnboardingPagesV2.KeyTagBackupDotMap, {
+        encodedText: mnemonic,
         title: '',
       });
-    },
-    [navigation],
-  );
+    } catch (error) {
+      console.error(error);
```
sidmorizon: # 🟡 P2 助记词提交路径上残留 `console.error(error)`

## 问题

`handleConfirm` 的 `catch` 直接 `console.error(error)`，而它包裹的 `phaseInputAreaRef.current.submit()` 会执行 `encodeSensitiveText` 与 `validateMnemonic`，属于恢复助记词代码路径。本 PR 自己新增的同类文件 `ImportKeyTag.tsx` 在相同位置刻意用 `catch (_error)` 且注释说明 `validateMnemonic` 是 `@toastIfError()`、toast 已自动弹出，无需再记录。

```ts
} catch (error) {
  console.error(error);
} finally {
```

## 影响

`CLAUDE.md` 明确要求不得在助记词/种子路径上 log。虽然 `InvalidMnemonic` 的默认 message 不含短语且 `defaultAutoToast: true`，但该 `catch` 是宽泛捕获，会同时吞下 `encodeSensitiveText`/`validateMnemonic` 抛出的任意其他错误对象并原样打印，形成潜在敏感信息落盘点，同时与同 PR 的姊妹实现不一致。

## 建议

对齐 `ImportKeyTag.tsx:236` 的写法，改为 `catch (_error)` 不记录（toast 已由 `@toastIfError` 负责），或改走 `defaultLogger` 且不带 error payload。

[Claude session ↗](https://claude.ai/code/session_01FRyA6k69Gsewc53qRKYHpF)

---

franco-chan: 已修正（6c3a5ccb65）：改为 `catch (_error)` 不记录，并补注释说明 `validateMnemonic` 是 `@toastIfError()` 已自动弹 toast — 与姊妹实现 `ImportKeyTag.tsx` 对齐；另确认 KeyTag 与 DotMap 路径下已无其他 `console.*`。这行是本 PR 自己引入的（3a99171d12 重写该 handler 时加的），确实是宽泛 catch 会原样打印任意错误对象，感谢抓到。

packages/kit/src/views/KeyTag/pages/BackupRecoveryPhrase/index.tsx
```diff
@@ -1,70 +1,147 @@
...
-      navigation.push(EModalKeyTagRoutes.BackupDotMap, {
-        encodedText: params.mnemonic,
+  const handleConfirm = useCallback(async () => {
+    if (!phaseInputAreaRef.current) {
+      return;
+    }
+    setIsConfirming(true);
+    try {
+      const { mnemonic } = await phaseInputAreaRef.current.submit();
+      navigation.push(EOnboardingPagesV2.KeyTagBackupDotMap, {
```
originalix, line 55: P1: [嵌套页面返回后会解除助记词截屏保护]

在 iOS/Android 上，这里 `push` 后 `PhaseInputArea` 仍保持挂载，新的 `KeyTagBackupDotMap` 也会调用 `useRecoveryPhraseProtected`。返回时顶部页面的 native cleanup 会无条件执行共享 native 单例的 `CaptureProtection.allow()`，导致下层仍显示助记词时截图/录屏保护已被解除。

请先把 `useRecoveryPhraseProtected.native.tsx` 改成与 desktop 版本相同的引用计数：只有最后一个保护消费者卸载时才调用 `allow()`，再保留这条嵌套路由。

packages/kit/src/views/Onboardingv2/pages/ImportKeyTag.tsx
```diff
@@ -1,107 +1,446 @@
...
+
+  // Leaving with punched holes discards them (input is intentionally not
+  // persisted); ask first. Successful submits disable the guard.
+  useEffect(() => {
+    const unsubscribe = reactNavigation.addListener('beforeRemove', (e) => {
+      if (allowLeaveRef.current) {
+        return;
+      }
+      let isDirty = false;
+      setRows((currentRows) => {
```
originalix: P2: [离开确认依赖 setState 同步读值，可能直接丢失输入]

用户带着孔位输入返回时，这里通过 `setRows` updater 修改局部 `isDirty`，然后立即读取它。React 不保证 updater 在这次读取前执行；存在 pending update 或未走 eager state 时，代码会跳过 `preventDefault()`，而本页又不持久化输入，结果是孔位被无提示丢弃。

请改用仓库已有的 `usePreventRemove(dirty, callback)`，直接以渲染态决定是否拦截，并只在用户确认后 dispatch 原 action。

packages/shared/src/routes/onboardingv2.ts
```diff
@@ -1,4 +1,7 @@
-import type { IKeylessWalletDetailsInfo } from '@onekeyhq/kit-bg/src/dbs/local/types';
+import type {
+  IDBWallet,
```
originalix: P2: [shared 路由契约继续依赖 kit-bg 数据库模型]

此次路由迁移把 `IDBWallet` 放进 `IOnboardingParamListV2`，使 `@onekeyhq/shared` 继续反向依赖 `@onekeyhq/kit-bg`，同时让导航状态绑定完整本地数据库模型。DB 类型变化会因此扩散到 shared route contract 和各平台 bundle。

请把该参数收窄为 shared 自有的 `walletId`、`isWalletBackedUp` 等字段，由 kit 页面消费这些 primitives。

packages/kit/src/views/KeyTag/pages/BackupRecoveryPhrase/index.tsx
franco-chan: Fixed in `ab11c3a149` — `useRecoveryPhraseProtected.native.tsx` now ref-counts like the desktop version: `prevent()` on the first consumer, `allow()` only when the last one unmounts, so the nested `KeyTagBackupDotMap` → `PhaseInputArea` pop no longer drops protection.

packages/kit/src/views/Onboardingv2/pages/ImportKeyTag.tsx
franco-chan: Fixed in `ab11c3a149` — replaced the `beforeRemove` listener + `setRows` read with `usePreventRemove(dirty && !hasSubmitted, ...)`, so the guard is armed from render state instead of a synchronous updater. The dialog dispatches `data.action` on confirm (react-navigation stamps the route key on the action before emitting, so the re-dispatch walks past the guard); `allowLeaveRef` became the `hasSubmitted` state that disarms it after a successful submit.

packages/shared/src/routes/onboardingv2.ts
franco-chan: Fixed in `ab11c3a149` — `KeyTagBackupDotMap` params are now `walletId?: string` + `isWalletBackedUp?: boolean`, matching `IVerifyRecoveryPhraseParams`. The `IDBWallet` import is gone from `shared/src/routes/onboardingv2.ts` and from `usePageNavigation.ts`; `useBackUpWallet` and `BackupWallet` pass the primitives, and `BackupDotMap` consumes them directly.

packages/kit/src/components/DotMap/KeyTagInput.tsx
```diff
@@ -0,0 +1,360 @@
...
+  line: IKeyTagLine;
+  onPress?: () => void;
+  onHoverIn?: () => void;
+  onHoverOut?: () => void;
+}) {
+  const dotSize = Math.max(6, Math.round(size * 0.42));
+  const guideSize = Math.max(2.5, size * 0.1);
+  return (
+    <Stack
+      width={size}
```
originalix, line 75: P1: [移动端孔位触控区过小，可能静默导入错误助记词]

在 iOS/Android 上，每个孔位的可点击宽高直接等于 `cellSize`。当前 400px plate 上限扣除 chrome 后最大只有 26pt，即使移动端完成 1.27 倍缩放也约 33pt，而且相邻格没有间距；用户照抄 KeyTag 时一次偏移会直接切换邻格。

错误 bit 通常仍会解码为另一个合法单词，而 12 词助记词只有 4 bit checksum，错误组合约有 1/16 可以通过 `validateMnemonic`，随后被 `FinalizeWalletSetup` 当作已备份钱包创建。

请把移动端交互改成逐行放大编辑器，保证每个孔位的实际触控区至少 44pt，再允许提交。

packages/kit/src/hooks/useRecoveryPhraseProtected/useRecoveryPhraseProtected.native.tsx
```diff
@@ -121,7 +129,10 @@ export const useRecoveryPhraseProtected = ({
     );
     let showTimer: ReturnType<typeof setTimeout> | undefined;
 
-    void CaptureProtection.prevent();
+    activeConsumerCount += 1;
+    if (activeConsumerCount === 1) {
+      void CaptureProtection.prevent();
+    }
     const listener = CaptureProtection.addListener(
```
originalix: P2: [引用计数未合并 capture listener，嵌套页会重复弹警告]

在 iOS/Android 上，从 `KeyTagEnterPhrase` push 到 `KeyTagBackupDotMap` 后，下层 `PhaseInputArea` 仍保持挂载，因此两个 `useRecoveryPhraseProtected` 消费者会同时存在。

`activeConsumerCount` 目前只合并了共享 native singleton 的 `prevent/allow`，这里仍为每个消费者注册一个 listener；一次截图事件会同时进入两个 timer/debounce，最终堆叠两个 `Dialog.confirm`。

请把 listener、timer 和 debounce 也集中到模块级 manager：首个消费者挂载时注册，最后一个卸载时移除。

franco-chan: Fixed in `0e9f8013a9` — the listener, timer and debounce now live in a module-level registry alongside the ref count: first consumer registers `prevent()` + one listener, last one removes both. The dialog resolves its copy from the topmost (last-registered) consumer, so `ExportPrivateKeys`' `sensitiveInformation` variant still wins when it is the visible screen; each hook reads its own `intl`/`dialogType` through a ref so a changed identity never churns the registration.

packages/kit/src/components/DotMap/KeyTagInput.tsx
franco-chan: Verified — and the numbers are worse than stated. `cellSize = floor((min(hostWidth, 400) - 84) / 12)` with the page at `px="$5"`, so on real phones: iPhone SE 20pt, iPhone 15 22pt, 15 Pro Max 25pt. The 26pt you cite is the ceiling (only reached once the host hits the 400 cap). After the 1.27 zoom that is 25–33pt, and the zoom has a 700ms delay + 900ms duration, so the first ~1.6s on screen is at the un-zoomed size. `GridCell` is `width/height = size` with no gap and no `hitSlop`, so the touch target really is exactly the cell.

The severity chain also holds. Flipping any bit except col 0 (2048) yields another in-range index, so ~11/12 of mistaps produce a valid-but-wrong word; a single flipped entropy bit leaves the stored 4-bit checksum unchanged while the computed one re-randomises, giving the 1/16 pass rate. The aggravating factor is real too: the wrongly-decoded wallet is created with `isWalletBackedUp: true`, but no KeyTag anywhere holds it — the plate holds the wallet the user *meant* to import. So the app asserts a backup that does not exist.

**Not fixing it in this PR.** 12 columns x 44pt = 528pt exceeds every phone width, so there is no way to reach the HIG minimum inside the current one-plate model — `hitSlop` cannot help either, since adjacent expanded targets overlap. Your proposal (a row-by-row enlarged editor on mobile) is the only geometry that works, and it is a mobile interaction redesign on a surface whose presentation decisions — the 1.27 zoom factor, the left drift, backup deliberately not zooming — were made deliberately by the design owner. Silently replacing that model from a review thread is the wrong call. Handing this to @Franco with the measurements above; leaving the thread open so it is not lost.

originalix, line 75: P1: [移动端孔位触控区过小，可能静默导入错误助记词]

在 iOS/Android 上照抄 KeyTag 时，每个孔位的可点击宽高直接等于 `cellSize`。当前 400pt plate 上限扣除 chrome 后单格最大只有 26pt，即使完成 1.27 倍缩放也约 33pt，窄屏还会更小，而且相邻格没有独立间距；手指稍有偏移就会切换邻格的 bit。

错误 bit 通常仍会解码为另一个合法 BIP39 单词。如果最终 checksum 也通过，`FinalizeWalletSetup` 会创建一个与实体 KeyTag 不同、却带有 `isWalletBackedUp: true` 的钱包。

请将移动端交互改为逐行放大的编辑器，保证每个孔位至少有 44pt 且不与相邻孔位重叠的实际触控区，再允许提交。

franco-chan: This restates #discussion_r3654877439, where I posted the full verification — measurements confirmed and worse than quoted (20/22/25pt on SE / 15 / Pro Max before the 1.27 zoom; 26pt is the ceiling, not the typical case).

Still not implementing it here, for the same reason: 12 columns x 44pt = 528pt exceeds every phone width, so the proposal is a mobile interaction redesign, not a bug fix — and the sizing decisions on this surface (the 1.27 factor, the left drift, backup deliberately not zooming) belong to the design owner. @Franco has the measurements and is weighing it as we speak; a narrower fix aimed at the actual silent-failure path (the import page has no ground truth to check against, unlike backup verify, which already flags mismatched rows against the wallet) is on the table. Leaving both threads open until he calls it.

franco-chan: Implemented in `20409521ab`. On small screens the plate is now a read-only map (tapping anywhere in a row selects it) and a docked `KeyTagRowPad` carries the input: one row of 48pt-tall cells with 4pt gutters, in `Page.Footer` so it stays in the viewport on short screens. Its chevrons step rows and flip the face; a failed Confirm lands it on the first offending row. gtMd keeps direct cell toggling.

Being straight about the geometry, since your ask was ">= 44pt and non-overlapping": the gutters give the non-overlapping half outright, and cells are ~48pt tall, but 12 columns cannot exceed ~30pt wide on a phone at any zoom — 12 x 44 = 528pt. The pad targets what actually causes the silent failure instead: a miss now lands on dead space (no toggle, no haptic) rather than flipping the neighbouring bit, which is the same reason an iOS keyboard works at ~26pt key width.

Two things your comment surfaced that are tracked separately rather than fixed here: `gtMd` is standing in for "precise pointer", so an iPad still gets 26pt direct-tap targets — the real predicate is the measured `cellSize` plus `pointerCoarse`. And `mismatchMask` reaches the plate but not the pad, so during backup verify the map can show a valid-but-wrong row and the pad cannot. Both are with @Franco.

franco-chan: Implemented in `20409521ab` — see the full reply on #discussion_r3654877439, which this thread restates. Short version: plate becomes a read-only row-selector map on small screens, and a docked `KeyTagRowPad` (48pt cells, gutters between them, pinned in `Page.Footer`) is the input surface; gtMd is unchanged.

packages/kit/src/components/DotMap/KeyTagRowPad.tsx
```diff
@@ -0,0 +1,146 @@
...
+          onPress={() => onStep(1)}
+          testID="keytag-row-pad-next"
+        />
+      </XStack>
+      <YStack gap="$1">
+        {PAD_SCALE_ROW}
+        <XStack gap={PAD_CELL_GAP}>
+          {keyTagRowValueToBits(value).map((on, col) => (
+            <Stack
+              key={KEY_TAG_ROW_WEIGHTS[col]}
```
originalix, line 122: P1: [移动端孔位命中区仍不足以防止误导入]

在触屏设备从 KeyTag 导入时，这 12 个 `flexBasis={0}` 单元仍被压进同一行。即使把完整 390pt 宽度全部交给它们，扣除 11 个 4pt 间隔后每格也不足 29pt，实际经过容器 padding 后还会更窄；而 `gtMd` 设备会绕过 RowPad，继续直接点击约 26pt 的网格。

用户横向稍有偏移就可能切换相邻 bit。错误值通常仍对应合法 BIP39 单词；如果整组恰好通过 checksum，`FinalizeWalletSetup` 会创建一个与实体 KeyTag 不同、却带有 `isWalletBackedUp: true` 的钱包。

请让所有原生触控设备使用横向可滚动的逐行编辑器，并为每个孔位设置至少 44pt 的固定宽高，避免 flex 压缩后再开放提交。

franco-chan: Measured rather than estimated, on a 375pt viewport: the pad's inner width is 301pt, so each cell renders at **21.4pt x 48pt** with 4pt gutters — worse than the <29pt you cite, because you assumed the full 390pt with no page or card padding. You are right about the mechanism, and right that `gtMd` still bypasses the pad entirely.

**Not implementing the horizontal scroller**, because for this specific task it trades one silent-failure mode for a worse one. 12 x 44pt + 11 x 4pt = 572pt of content in a 301pt viewport — **1.9 screens**. Users transcribe a KeyTag row by *pattern-matching*: they look at the physical row, see holes at (say) columns 2, 7, 11 as one shape, and reproduce that shape. Showing half a row at a time makes pattern-matching impossible and forces column-counting, which is the error-prone operation the 2^n scale exists to eliminate. It also adds a genuinely new failure: a user can submit a row without realising columns existed off-screen to the right, and on a horizontally scrollable strip of tappable cells a slightly-draggy tap becomes a scroll instead of a toggle.

The alternative that actually reaches the target without scrolling: **wrap the row into two lines of 6**. Same 301pt gives **46.8pt x 48pt per cell** — over 44pt in *both* dimensions, all 12 bits visible at once, no new gestures. That is a layout decision on a surface whose spacing, radius and fill were all set deliberately by @Franco, and it does change how the pad mirrors the physical one-line row, so it is his call, not something to land off a review thread. Escalating it to him with the measurement above.

Leaving this open until he rules. The `gtMd` bypass is tracked with it — the real predicate is measured `cellSize` plus `pointerCoarse`, not viewport width, and both are recorded as known limitations on OK-58354.

franco-chan: Closing this out: the current solution stands as-is for this PR. The row pad (48pt-tall cells, dead-space gutters between them, docked so it cannot scroll out of view) is what ships; the remaining width gap and the `gtMd` bypass are accepted, not fixed.

The reasoning, from @Franco: further tuning here would be guessing. The pad already removes the failure mode that mattered — a miss now lands on dead space instead of silently flipping a neighbouring bit — and whether the residual 21.4pt width actually produces mistaps in the field is an empirical question. We would rather see real usage data or user reports than optimise blind, especially when the two candidate fixes (horizontal scrolling, or wrapping to two rows of 6) each trade away the one-line correspondence with the physical plate row.

Both limitations are documented on OK-58354 as known-and-expected so QA does not file them as bugs.

packages/kit/src/components/DotMap/KeyTagRowPad.tsx
```diff
@@ -0,0 +1,146 @@
...
+              flexGrow={1}
```
originalix, line 123: P1: [触控孔位命中区仍可能生成未受实体 KeyTag 保护的钱包]

在 iOS/Android 小屏上从 KeyTag 导入时，这 12 个 `flexBasis={0}` 单元会被压进同一行。扣除 11 个间隔、面板 padding 和 Footer 留白后，每个孔位只有约二十余 pt；触屏设备进入 `gtMd` 时还会绕过 RowPad，直接操作约 26pt 的网格。

用户横向稍有偏移就可能切换相邻 bit。错误值通常仍对应合法 BIP39 单词，完整短语也可能通过 checksum；随后 `FinalizeWalletSetup` 会创建一个与实体 KeyTag 不同、却带有 `isWalletBackedUp: true` 的钱包，之后存入该钱包的资产将无法从这块 KeyTag 恢复。

请让所有原生触控设备使用横向可滚动的逐行编辑器，并为孔位设置固定最低宽高（iOS 44pt、Android 48dp），不要让 `gtMd` 触控设备回退到小网格直接输入。

franco-chan: This is the fourth posting of this finding (#discussion_r3654877439, #discussion_r3655032274, #discussion_r3656294320), and it has been decided rather than left open.

**Decision by @Franco, the design owner: keep the current solution for this PR; revisit on real usage data or user reports.** Not implementing the horizontal scroller or the `gtMd` touch-device change.

The measurement is not in dispute — I measured the rendered cell at **21.4pt x 48pt** on a 375pt viewport, which is worse than the "二十余 pt" here. What the pad already does is remove the specific silent-failure path: the 4pt gutters mean a horizontal miss lands on dead space and produces no toggle and no haptic, rather than flipping the neighbouring bit unnoticed. Reaching a fixed 44pt width would need either 1.9 screens of horizontal scrolling — which breaks the pattern-matching users actually rely on to transcribe a row, and adds a new failure where a row is submitted with columns never seen — or wrapping to two rows of 6. Both trade away the one-line correspondence with the physical plate, and neither is worth committing to before there is evidence the residual width causes mistaps in the field.

Both limitations (residual cell width; tablets at >=768pt using the direct grid) are recorded on OK-58354 as known-and-expected, with QA cases written so they are not filed as bugs. Resolving; please reopen with field evidence rather than a re-post if this proves out in practice.

originalix, line 123: P1: [过窄的孔位触控区可把错误钱包标成已有 KeyTag 备份]

在 iOS/Android 的 KeyTag 导入页使用这个 RowPad 时，`FooterActions` 的水平 padding、RowPad 自身 padding 和 11 个间隔会把这 12 个 `flexBasis={0}` 单元压到约 17–23pt 宽；触控设备进入 `gtMd` 时还会绕过 RowPad，直接操作最多约 26pt 的网格。

用户横向稍有偏移就可能切换相邻 bit。错误值仍可能对应合法 BIP39 单词，完整短语也可能通过 checksum；随后 `FinalizeWalletSetup` 会创建一个与实体 KeyTag 不同、却带有 `isWalletBackedUp: true` 的钱包，之后存入该钱包的资产将无法从这块 KeyTag 恢复。

请让所有触控或 coarse-pointer 设备使用横向可滚动的逐行编辑器，并为孔位设置固定最低宽高（iOS 44pt、Android 48dp），不要让触控 `gtMd` 设备回退到紧凑网格直接输入。

franco-chan: Fifth posting of this same finding (#discussion_r3654877439, #discussion_r3655032274, #discussion_r3656294320, #discussion_r3656877396). Nothing new here — the 17–23pt range agrees with the 21.4pt I measured on a 375pt viewport, and the prescription is unchanged.

**Already decided, not outstanding.** @Franco, the design owner, ruled: keep the current solution for this PR and revisit on real usage data or user reports. The full reasoning — why the gutters already remove the silent neighbour-toggle path, and why 1.9 screens of horizontal scrolling would break the pattern-matching users rely on to transcribe a row — is on the threads above. Both limitations are recorded on OK-58354 as known-and-expected with QA cases written so they are not filed as bugs.

Please reopen with field evidence rather than a re-post.

originalix (approved):

PR #12649, opened 2026-07-27
fix: restore perps orderbook and dead banner taps on notification entry

Minnzen:
## Summary

Entering Perps from the **notification center** or a **home wallet banner** left the order book empty for 20-30 seconds. The 19-second recording from the original report shows an empty book end to end — the data was never missing, it just arrived long after the user gave up.

Two independent defects are fixed here, both on the notification/banner entry path.

## Root cause — order book

These two entries are the only ones that switch the active coin **while the Perp route is blurred**; every other entry calls `switchTab` first. They also navigate through `popToMainRoute()` plus a 350ms wait before the tab even changes.

Because the switch lands in that blurred window, the `routeFocused` gate in `syncSubscriptionsAfterInstrumentChange` swallowed **both** the subscription-target publish and the resubscribe. `perpsActiveOrderBookOptionsAtom` kept naming the previous coin, `isOrderBookOptionsTargetReady` then aborted the **entire** background reconcile — allMids, L2, BBO and user channels, not just the book — until the next focus transition.

Instrumented run, at the moment it strands:

```
probe_ui_update_subscriptions_blocked {"backed":true,"ctxCoin":"BTC","bgPerpCoin":"BTC","orderBookCoin":"ETH","routeFocused":false}
probe_instrument_sync_skipped        {"shouldSync":false,"routeFocused":false}
```

Context and background already agree on BTC; only `orderBookCoin` is stuck on ETH.

Leaving Perps for **1-5 minutes** is the worst window: the socket is already CLOSED, but `pauseSubscriptions()` (5m30s) has not run its full teardown yet, so recovery serialises a 3s wait, a 5s connection timeout and exponential backoff — roughly 20 extra seconds. That is exactly how long a user spends on the home screen before tapping a notification.

## Changes

**1. Publish the subscription target unconditionally** (`actions.ts`)
Publishing the target is idempotent state sync and is the key that unblocks the background reconcile, so it should not be focus-gated. Only `updateSubscriptions()` stays behind the gate.

**2. Narrow the initial-symbol latch skip** (`PerpsGlobalEffects.tsx`)
`tryClaimInitialSymbolSelect` is a process-wide latch (introduced by OK-53208), which makes this skip the *only* remount-time resync. It now bails out only when the context instrument still agrees with the background target, so a dropped event-bus message can self-heal.

**3. Set `tradingModeAtom` and prewarm the socket on the notification path** (`NotificationHandlerContainer`)
The payload targets a perp market but `changeActiveAsset` only writes `perpsActiveAssetAtom`; a user left in spot mode would have the resync read `spotActiveAssetAtom` and drop the coin entirely. The reconnect is also kicked off early so it overlaps navigation.

**4. Fix the never-true `dexUniverses?.length === 0` test** (`ServiceHyperliquid.ts`)
`universesByDex` is `[]` when nothing is cached, so `[dexIndex]` is `undefined` and the test never matched — execution fell through and committed `assetId: -1` as a resolved asset. Both `useOrderConfirm.ts:97` and `ServiceHyperliquidExchange.ts:1012` only guard against `undefined` / non-number, so **`-1` slipped straight into the order path**. The requested coin is now kept with unresolved metadata, and re-selecting the active coin no longer downgrades metadata that was already resolved.

**5. Stop dead banner and notification taps** (`notificationsUtils.ts`, `useWalletBanner.ts`)
`JSON.parse('{}')` succeeds, so a payload without a `screen` field reached the navigator as `StackActions.push(undefined)` — logged as `The action 'PUSH' with payload {} was not handled by any navigator` in development, silently dropped in production, i.e. a tap that does nothing. A missing target is now treated as a parse failure so the caller's fallback runs, and `useWalletBanner` no longer swallows that failure with an empty handler — an unusable payload falls through to the banner's own `href`.

## Review follow-up (3rd commit)

- **Focus-transition dedupe on `selectInitialSymbol`.** It runs off `useOnRouterChange`, so any in-tab push/pop re-entered it. The token selector calls `popStack()` before its own `switchTradeInstrument()` resolves, and the context instrument is written optimistically ahead of the background atoms — that window read as divergence and would have forced the previous coin back over the user's selection.
- **Clear `perpsActiveAssetCtxAtom` on the empty-universe early return.** That branch used to fall through to the shared tail which does this on a coin switch. `perpsActiveAssetCtxMidPriceAtom` reads the ctx with no coin guard, so `useTradingPrice` was seeding the form with the previous coin's mid price.
- **Corrected the comment on the resubscribe gate.** The background atom watcher subscribes to `perpsActiveOrderBookOptionsAtom` on extension and native, so the gate only holds while the handler is disabled (`AutoPauseSubscriptions` does that on blur) and is the sole gate on desktop/web.

## Measured (iOS Simulator, iPhone 16 Pro)

| Scenario | Before | After |
|---|---|---|
| Cold start, first entry (ETH) | 8.7s | **0.3s** |
| Coin switch after 2 min away (ETH→BTC) | **28.9s** | **0.5s** |
| Real home banner (SpaceX, `xyz:` dex coin) | — | book renders immediately, no errors |
| Banner with an invalid payload | dead tap + PUSH warning | opens its `href` in the in-app browser |

Reproduced through Dev Settings → Notification Payload Test, which goes through the same `parseNotificationPayload` entry as real notifications and banners, then cross-checked with a real home banner.

## Verification and risk

- Type check: the only remaining error repo-wide is `TS2688: Cannot find type definition file for 'web'`, a missing `@types/web` in the local workspace, unrelated to this change.
- ⚠️ `yarn agent:check --profile commit` **did not run**: `typescript`, `tsgo` and `eslint-plugin-prettier` are absent from the local `node_modules` and fail identically on a clean branch. **Needs a rerun in CI or a fully installed environment.**
- Pure TypeScript, no native / Podfile / DB schema changes, so it satisfies the bundle-release constraints.
- Only exercised on the iOS Simulator; **Android runs the same code and still needs a regression pass.**

## QA focus

1. Fresh install, never open Perps, then enter from a notification or banner — the coin from the payload should show, and should not flip back to the default once metadata resolves.
2. Open Perps on coin A, leave for 1-5 minutes, then enter coin B from a notification or banner — the book should appear immediately. **This is the originally reported scenario.**
3. Leave Perps while in spot mode, then enter a perp coin from a notification — the page should switch back to perp mode.
4. Repeat with an `xyz:`-prefixed coin such as SPCX.
5. Pick a coin from the mobile token selector — the selection must not roll back to the previous coin.
6. Switch coins while the target dex has no cached universe — the order form must not prefill the previous coin's mid price.
7. A banner with an invalid payload but a valid `href` should open the href instead of doing nothing.
8. Full Android regression.


packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -952,7 +952,24 @@ function useHyperliquidSymbolSelect() {
...
+          await buildActiveInstrumentSwitchParamsFromGlobal();
+        const ctxInstrument = activeTradeInstrumentRef.current;
+        const diverged =
+          !bgSwitchParams ||
+          bgSwitchParams.coin !== ctxInstrument?.coin ||
+          bgSwitchParams.mode !== ctxInstrument?.mode;
+        markPerpsColdStartPerf('initial_symbol_latch_skip', {
+          diverged,
+        });
+        if (!diverged) {
```
sidmorizon, line 968: # 🟠 P1 latch-skip 分支会把用户正在进行的换币回滚

## 问题

`selectInitialSymbol` 的触发器没有做焦点跳变去重：

```ts
useListenTabFocusState(ETabRoutes.Perp, (isFocus: boolean) => {
  if (!resolvePerpRouteFocused(isFocus)) return;
  void selectInitialSymbol();
});
```

`useListenTabFocusState` 底层是 `useOnRouterChange`，只要 Perp 是当前 tab，**任何一次路由状态变化**（弹/关任意 modal、栈内 push/pop）都会再跑一次 `selectInitialSymbol`。改动前 `claimed === false` 会立刻 `return`，代价为零；改动后它变成一次「BG 与 context 对账 + 分歧则强制切换」。

而 context 的 instrument 是**乐观先写**、BG atom 后写的。以移动端选币为例（`MoblieTokenSelector.tsx` `handleSelectToken`）：

```ts
navigation.popStack();                              // 立刻触发 useOnRouterChange → selectInitialSymbol()
await actions.current.switchTradeInstrument({ mode, coin: symbol });
```

`changeActiveAsset` 先同步 `set(activeTradeInstrumentAtom(), optimisticInstrument)`，随后要经过 `cancelPendingActiveAssetChange()`、`perpsActiveAssetAtom.get()` 等多次 BG 往返才写 `tradingModeAtom` / `perpsActiveAssetAtom`。在这个窗口里 `ctxInstrument.coin` 已是新币、`bgSwitchParams.coin` 仍是旧币，`diverged` 为 `true`，于是继续往下走并执行 `switchTradeInstrument({ coin: 旧币, force: true })`。

## 影响

`switchTradeInstrument` 会调用 `beginActiveInstrumentChange()` 抬高 `activeInstrumentChangeRequestId`，使用户那次仍在飞行的切换在下一个 `isLatestActiveInstrumentChange` 检查点返回 `false` 被丢弃，最终**用户选中的币被静默回滚到上一个币**。`changeActiveSpotAsset` 同理：`ctx.mode` 已乐观置为 `'spot'` 而 BG `tradingModeAtom` 仍是 `'perp'`，同样判为分歧并回滚到 perp。

因为走的是 `force: true`，回滚路径还会执行 `clearActiveAssetData` 与 `updateTradingForm`（`triggerPrice: ''`、`executionPrice: ''`、重算 limit `price`），**用户已经填写的触发价/委托价会被清空**。这是一个与 BG 往返耗时赛跑的竞态，在原生桥接下窗口可达 `100ms` 量级，表现为「偶发换币不生效 / 表单被清空」。

## 建议

不要在 latch-skip 分支里直接对账并强制切换。可选方向：

1. 让触发器只在焦点真正发生 `false → true` 跳变时才跑（对齐 `AutoPauseSubscriptions` 里 `isFocusPrev !== nextFocused` 的写法），避免每次路由变化都重入；
2. 对账前先确认当前没有在飞行的 instrument 切换，例如在 actions 上暴露一个 `isActiveInstrumentChangeInFlight()`，或让 `selectInitialSymbol` 在读取前后各取一次 `activeInstrumentChangeRequestId`，两次不一致就放弃本轮对账：

```ts
const seq = actions.current.getActiveInstrumentChangeRequestId();
const bgSwitchParams = await buildActiveInstrumentSwitchParamsFromGlobal();
if (seq !== actions.current.getActiveInstrumentChangeRequestId()) return; // 有人正在切，交给它
```

3. 或者把分歧兜底从「强制切换」降级为「仅当 context 长时间未收敛才切」，并复用非 `force` 路径以免清空交易表单。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -2661,22 +2661,32 @@ export default class ServiceHyperliquid extends ServiceBase {
...
-          assetId: rollbackActiveAsset?.assetId,
-          universe: rollbackActiveAsset?.universe,
-          margin: rollbackActiveAsset?.margin,
+          coin: newCoin || rollbackActiveAsset?.coin || '',
+          assetId: undefined,
+          universe: undefined,
+          margin: undefined,
         };
+        if (requestId === this.activeAssetChangeRequestId) {
+          await perpsActiveAssetAtom.set(result);
```
sidmorizon, line 2690: # 🟡 P2 空 universe 提前返回漏掉了换币时的 activeAssetCtx 清理

## 问题

修复 `dexUniverses?.length === 0` 恒假是对的，但这条分支以前是**穿透**下去的：`selectedUniverse` 为 `undefined`、`assetId` 落到 `-1`，然后仍然会走到函数末尾的

```ts
if (oldCoin !== newCoin) {
  await perpsActiveAssetCtxAtom.set(undefined);
  schedulePerpsActiveAssetCtxDisplayUpdate({ nextValue: undefined, immediate: true });
}
```

现在改成在 `!dexUniverses?.length` 时直接 `return result`，这段「换币即清空上一个币的 ctx」的清理被跳过了，`perpsActiveAssetCtxAtom` / `perpsActiveAssetCtxDisplayAtom` 仍然保留着**旧币**的 mark/mid/funding。

## 影响

`usePerpsActiveAssetCtxDisplay` 有 `assetCtx?.coin === activeCoin` 的守卫所以行情区不受影响，但 `perpsActiveAssetCtxMidPriceAtom` 这条链路没有按 coin 校验，`useTradingPrice` 直接把它当首选价格用：

```ts
const midPrice = activeTradeInstrument.mode === 'spot'
  ? ...
  : activeAssetCtxMidPrice || activeMidPrice;
```

也就是说，在目标 dex 的 universe 缓存为空时切币（`detectDexIndexByCoin` 对 `xyz:` 前缀返回 `1`，该 dex 的 meta 未落盘时即命中），下单表单会拿**上一个币**的中间价来预填 limit 价 / 估算，直到第一条新币的 `activeAssetCtx` 推送到达才纠正。这是交易路径上的错价，比原先 `assetId: -1` 的问题面更隐蔽。

## 建议

在提前返回前补上同样的换币清理：

```ts
if (requestId === this.activeAssetChangeRequestId) {
  await perpsActiveAssetAtom.set(result);
  if (oldCoin !== newCoin) {
    await perpsActiveAssetCtxAtom.set(undefined);
    schedulePerpsActiveAssetCtxDisplayUpdate({ nextValue: undefined, immediate: true });
  }
}
```

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -544,6 +546,12 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
       return;
     }
 
+    // Resubscribing stays focus-gated so a blurred route does not rebuild
+    // streams it cannot render.
+    if (!this.shouldSyncSubscriptionsAfterInstrumentChange(params.viewState)) {
```
sidmorizon, line 550: # 🟡 P2 这道 focus 门在 iOS/Android/扩展上不生效

## 问题

注释说「Resubscribing stays focus-gated so a blurred route does not rebuild streams it cannot render」，但上面刚刚无条件写入的 `perpsActiveOrderBookOptionsAtom` 本身就是 BG 的重建触发器。`ServiceHyperliquidSubscription._watchSubscriptionAtoms()` 里：

```ts
this._subscriptionAtomsUnsubs = [
  perpsActiveAccountAtom.sub(handler),
  perpsActiveAssetAtom.sub(handler),
  spotActiveAssetAtom.sub(handler),
  tradingModeAtom.sub(handler),
  perpsActiveOrderBookOptionsAtom.sub(handler),   // handler → void this.updateSubscriptions()
];
```

该 watcher 仅在 `platformEnv.isExtension || platformEnv.isNativeBackgroundThread` 时安装，且只有 `pauseSubscriptions()` 会 `_unwatchSubscriptionAtoms()`——而 `AutoPauseSubscriptions` 的 pause 要在失焦 `5m30s` 之后才触发。

## 影响

在 iOS / Android / 扩展（`main` 与 `bg` 分离运行时）上，失焦后的 `5m30s` 窗口内 watcher 仍然在位，`publishLatestOrderBookOptions` 一写入 BG 就会立刻 `updateSubscriptions()`，这道新增的 `shouldSyncSubscriptionsAfterInstrumentChange` 门形同虚设。只有 desktop/web（单运行时、无 watcher）才真正被它挡住。结果是同一份代码在两类目标上语义相反，而注释描述的恰好是不成立的那一边。

## 建议

二选一，并让注释与实际行为一致：

- 如果确实要保留「失焦不重建」语义，就把这个判断下沉到 BG（例如在 `updateSubscriptions` / atom watcher 的 handler 里读 `setRouteSubscriptionState` 已经维护的路由状态再决定是否重建），让两类目标行为统一；
- 如果无条件重建是可接受甚至期望的（毕竟目标就是让 BG 尽快跟上），那就删掉这道门并把注释改成说明 BG watcher 会自行接管，避免后续读者误以为存在 focus 保护。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -1098,8 +1115,18 @@ function useHyperliquidSymbolSelect() {
...
+  // selectInitialSymbol mid-switch. The context instrument is written
+  // optimistically ahead of the BG atoms, so that window reads as divergence
+  // and would force the previous coin back over the user's selection.
+  const isRouteFocusedRef = useRef(shouldTreatPerpAsFocusedOnMount);
   useListenTabFocusState(ETabRoutes.Perp, (isFocus: boolean) => {
-    if (!resolvePerpRouteFocused(isFocus)) return;
+    const nextFocused = resolvePerpRouteFocused(isFocus);
+    const wasFocused = isRouteFocusedRef.current;
+    isRouteFocusedRef.current = nextFocused;
+    if (!nextFocused || wasFocused) return;
```
sidmorizon, line 1124: # 🟠 P1 焦点跳变去重挡不住 remount，选币竞态在原生端仍然成立

## 问题

去重只在组件**持续挂载**时有效。`useOnRouterChange` 在订阅时会立刻用当前路由状态回调一次：

```ts
useEffect(() => {
  routerRef.current.push(callback);
  if (rootNavigationRef.current) {
    callback?.(rootNavigationRef.current?.getState());   // 挂载即触发
  }
  ...
}, []);
```

而 `isRouteFocusedRef` 是 `useRef`，**每次 remount 都会重新初始化**为 `shouldTreatPerpAsFocusedOnMount`（原生端为 `false`）。于是每次 remount 时首个回调必然是 `wasFocused === false && nextFocused === true`，判定为跳变并再次执行 `selectInitialSymbol()`。

这正好命中注释里点名的那条路径。本文件自己已三处记录了原生端的 detach 行为：

- L538：`OK-53208: perps atoms/WS must survive unmount — mobile modal push detaches this tab`
- L946：`latch lives in ServiceHyperliquid (singleton) so that Perp tab detach/remount does not re-trigger this init`
- L1320：`unmount ≠ user left Perp. Mobile modal push detaches this tab`

也就是说 `MoblieTokenSelector` 的选币流程在原生端是「push modal → Perp tab detach/unmount」，随后 `handleSelectToken` 里 `navigation.popStack()` 触发 **remount**，紧接着才 `await switchTradeInstrument(...)`——去重被 remount 重置，`selectInitialSymbol` 照旧在切换中途重入。

## 影响

时序与上一轮完全一致：`popStack()` 之后 `changeActiveAsset` 的同步前缀已把 context instrument 乐观写成新币，而 BG `perpsActiveAssetAtom` 要再经 `cancelPendingActiveAssetChange()`、`perpsActiveAssetAtom.get()`、`tradingModeAtom.set()` 等多次桥接往返才更新。重入的 `selectInitialSymbol` 只需 `tryClaimInitialSymbolSelect()` + 两次 atom 读即可完成对账，通常先读到「ctx=新币 / BG=旧币」，`diverged` 为 `true`，随后 `switchTradeInstrument({ coin: 旧币, force: true })` 抬高 `activeInstrumentChangeRequestId`，把用户那次切换作废并回滚。

即使 remount 后 context 尚未恢复、`activeTradeInstrumentRef.current?.coin` 为空而跳过 `diverged` 分支，代码同样会直接走到 `buildActiveInstrumentSwitchParamsFromGlobal({ force: true })` 并强制切回 BG 里的旧币——**两条分支殊途同归**。`force: true` 路径还会执行 `clearActiveAssetData` 与 `updateTradingForm`，清空已填写的 `triggerPrice` / `executionPrice`。

## 建议

去重是有价值的（挡住了持续挂载期间的重复重入），但不足以覆盖 detach/remount，需要再加一道与挂载生命周期无关的护栏，即对「有在飞行的 instrument 切换」本身做判定。例如在 actions 上暴露序号并做前后比对：

```ts
const seq = actions.current.getActiveInstrumentChangeRequestId();
const bgSwitchParams = await buildActiveInstrumentSwitchParamsFromGlobal();
if (seq !== actions.current.getActiveInstrumentChangeRequestId()) return;
```

这道检查放在 `selectInitialSymbol` 里对账与 `switchTradeInstrument` 之间即可，同时覆盖 `diverged` 分支和 `coin` 为空的分支。另一种做法是把强制切换降级为非 `force`，至少避免连带清空用户已填写的交易表单。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/shared/src/utils/notificationsUtils.ts
```diff
@@ -103,6 +104,14 @@ export async function navigateToNotificationDetailByLocalParams({
   getEarnAccount: IGetEarnAccountFunc;
 }) {
   const { screen, params: navigationParams } = payload;
+  // Defence in depth for callers that bypass parseNotificationPayload (the
```
sidmorizon: # 🟡 P2 新增注释里的 Defence 拼写导致 CI lint 失败

## 问题

本行使用了英式拼写 `Defence`，不在仓库 cspell 词典中：

```ts
// Defence in depth for callers that bypass parseNotificationPayload (the
```

这是全仓唯一一处 `Defence`，由本 PR 的 commit `b773341` 新增。oxlint 的 spellcheck 规则据此报出：

```
##[warning]Unknown word: "Defence"
Found 1 warning and 0 errors.
```

## 影响

oxlint 在 CI 模式下 warning 即以 exit code `1` 结束，因此 `lint (24.x)` job 在 head `3d2ca287` 上直接失败：

```
[2:03:15 PM] Oxlint check failed. (183.59s)
```

这是**合并阻塞项**——在修掉之前 PR 无法通过 required check。

## 建议

改成美式拼写 `Defense`，与仓库其余用词保持一致：

```ts
// Defense in depth for callers that bypass parseNotificationPayload (the
```

若确实希望保留英式拼写，则需按仓库既有约定把该词加入 cspell 词典文件，但为单处注释扩充词典不划算，建议直接改词。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Confirmed and fixed in 3d2ca287cf.

Verified the race exists as described: `MoblieTokenSelector.tsx:384` calls `navigation.popStack()` before awaiting `switchTradeInstrument()` at `:389`/`:395`, and the trigger had no dedupe, so that pop re-entered `selectInitialSymbol` while the context instrument was already optimistically ahead of the BG atoms.

Fix is a focus-transition dedupe on the trigger, matching the `isFocusedRef` pattern the other three `useListenTabFocusState` callers in this file already use:

```ts
const isRouteFocusedRef = useRef(shouldTreatPerpAsFocusedOnMount);
useListenTabFocusState(ETabRoutes.Perp, (isFocus: boolean) => {
  const nextFocused = resolvePerpRouteFocused(isFocus);
  const wasFocused = isRouteFocusedRef.current;
  isRouteFocusedRef.current = nextFocused;
  if (!nextFocused || wasFocused) return;
  void selectInitialSymbol();
});
```

In-tab router churn no longer re-enters it; the notification/banner path still does, because leaving and re-entering the tab is a genuine `false -> true` transition.

Verified on the iOS Simulator: picked XMR/USDC from the mobile token selector while CXMT perp was active — the selection holds, no rollback after 60s, and no `initial_symbol_latch_skip` in the log. Re-checked that the banner path still works (CXMT banner renders its book immediately).

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Confirmed and fixed in 3d2ca287cf, implemented as suggested.

Verified the exposure: `perpsActiveAssetCtxMidPriceAtom` is a `globalAtomComputedR` reading `perpsActiveAssetCtxAtom.atom()?.ctx?.midPrice` with no coin guard (`atoms/perps.ts:730-734`), and `useTradingPrice.ts:84` takes it as the preferred price via `activeAssetCtxMidPrice || activeMidPrice`. So the skipped cleanup did leak the previous coin's mid into the order form.

```ts
if (requestId === this.activeAssetChangeRequestId) {
  await perpsActiveAssetAtom.set(result);
  if (oldCoin !== newCoin) {
    await perpsActiveAssetCtxAtom.set(undefined);
    schedulePerpsActiveAssetCtxDisplayUpdate({
      nextValue: undefined,
      immediate: true,
    });
  }
}
```

Kept it inside the `requestId` check so a superseded change cannot clear a newer coin's ctx.

Added to the QA list: switch coins while the target dex has no cached universe (the `xyz:` prefix case you called out) and confirm the form does not prefill the previous mid.

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
Minnzen: Partly confirmed — the comment was over-promising and I rewrote it in 3d2ca287cf. One correction on the scope though.

The watcher does fire, but `updateSubscriptions()` short-circuits before doing anything:

```ts
// ServiceHyperliquidSubscription.ts:633
if (this.subscriptionsHandlerDisabled) {
  markPerpsColdStartPerf("service_update_subscriptions_skipped_disabled");
  return;
}
```

and `AutoPauseSubscriptions` disables the handler on blur immediately (`PerpsGlobalEffects.tsx:1250`), not at the 5m30s mark — the timer only governs the later `pauseSubscriptions()` teardown. So for the whole blurred window the gate does hold on native and extension.

Where you are right is the modal case: `AutoPauseSubscriptions` resolves focus with `resolvePerpRouteFocused(isFocus)`, which ignores `isHiddenByModal`, while `routeFocused` does not. With the Perp tab current but covered by a modal, the handler stays enabled and the publish does trigger a rebuild — so the gate is bypassed there. Same for the async gap around the blur transition.

I opted for the second of your two options, minus the removal: kept the gate (it is the only one desktop/web have) and made the comment describe the real behaviour instead of an invariant that does not hold:

```ts
// Skip the explicit resubscribe when the route cannot render the result.
// This is a best-effort saving, not a guarantee: on extension and native
// the BG atom watcher subscribes to perpsActiveOrderBookOptionsAtom and
// reconciles on its own, so the publish above can already trigger a
// rebuild there. It only stays suppressed while the handler is disabled
// (AutoPauseSubscriptions does that on blur) — notably not while the Perp
// tab is merely covered by a modal. Desktop and web have no watcher, so
// this branch is the only gate for them.
```

Pushing the decision into BG so both targets behave identically is the right end state, but it changes subscription lifecycle semantics for every entry point, which I would rather not land in a bundle-release hotfix. Happy to open a follow-up issue for it if you agree.

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
sidmorizon: 你是对的，我第二轮的 remount 结论不成立，撤回。逐条核实如下（均在 head `3d2ca287`）：

**B. modal push 不会 unmount `PerpsGlobalEffects`。** 这是我错的根源——我把 L538/L946/L1320 那三处 "detaches this tab" 读成了 React 卸载，实际证据指向原生 view 层/freeze：

- `Perp.tsx` 的 `Perp()` 主路径直接渲染 `<PerpView/>`（内含 `<PerpsGlobalEffects/>`），**没有** `useIsFocused()` 门控；仅 SUB+landscape 分支（L255）和 `ExtPerpNull` 用到 `isFocused`。
- `GlobalScreenOptions.ts` 对 root/modal stack 统一设置 `detachPreviousScreen: false`。
- `TabStackNavigator.native.tsx` 的 `nativeTabScreenOptions` 是 `freezeOnBlur: !platformEnv.isNativeIOS`——你测试的 iOS 上连 freeze 都没有；Android 走 react-freeze 也只是挂起渲染，保留 state/ref。
- 全仓无 `unmountOnBlur` / `detachInactiveScreens`。

所以 `isRouteFocusedRef` 不会被重置，我关于 `useOnRouterChange` 挂载即回调导致伪跳变的推论失去前提。

**C. 也不存在伪跳变。** `useListenTabFocusState` 的 `currentTabName` 取自 `ERootRoutes.Main` 的 tab state，而 pushModal 推的是独立 root route，只体现在 `isHideByModal`。整个 open → select → `popStack()` 周期里 `isFocus` 恒为 `true`，`nextFocused === wasFocused === true`，`selectInitialSymbol` 不会重入。这与你观察到的「选择保持住、无 `initial_symbol_latch_skip`」一致。

**A. 你的日志证据是成立的**，我顺带核实了这一点：`markPerpsColdStartPerf` 没有冷启动窗口门控（`isPerpsColdStartPerfEnabled()` 恒返回 `true`，函数无条件调用 `coldStartBenchmark`，该方法带 `@LogToLocal({ level: 'info' })`），且 `initial_symbol_latch_skip` 用的是 `markPerpsColdStartPerf` 而非 `...Once`，每次进入都会记录。所以「日志里没有」确实能说明该分支未被触达。

**D.** `isInitializingRef` 只拦截并发重入（会打 `initial_symbol_skip_initializing`），此场景下没有在飞行的调用，不构成对日志缺失的替代解释。

焦点跳变去重对 in-tab modal churn 是有效的，第一轮 P1 视为已修复。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
sidmorizon: 你的更正成立，我原评论里「失焦后 `5m30s` 窗口内 watcher 仍在位、gate 形同虚设」这一论断是错的，撤回。

核实结果（head `3d2ca287`）：

**1 + 2. 短路与禁用时机。** `updateSubscriptions()` 在 L633 就 `if (this.subscriptionsHandlerDisabled) return;`（`_updateSubscriptionsCore` 在 L565 还有一道），而 watcher 的 handler 正是 `void this.updateSubscriptions()`，所以确实被挡住。`AutoPauseSubscriptions.onFocusHandler` 的 blur 分支：

```ts
} else {
  void backgroundApiProxy.serviceHyperliquidSubscription.disableSubscriptionsHandler();  // L1260，同步调用
  clearTimeout(pauseSubscriptionsTimerRef.current);
  pauseDelay = pauseDelay ?? timerUtils.getTimeDurationMs({ minute: 5, seconds: 30 });
  pauseSubscriptionsTimerRef.current = setTimeout(() => {
    void backgroundApiProxy.serviceHyperliquidSubscription.pauseSubscriptions();          // L1269，仅 teardown
  }, pauseDelay);
}
```

禁用在 blur 回调里即刻发生，`5m30s` 只管后续 `pauseSubscriptions()` 的 teardown。我把这两件事混为一谈了。

**4. modal 遮挡的缺口确认仍然存在**，两处用法差异如你所述：`AutoPauseSubscriptions` 的监听（L1278）只取 `resolvePerpRouteFocused(isFocus)`，忽略 `isHiddenByModal`；而 `tradeRouteViewStateAtom` 的 `routeFocused`（L233）是 `resolvePerpRouteFocused(isFocus) && !isHiddenByModal`。于是「Perp 是当前 tab 但被 modal 遮挡」时，`shouldSyncSubscriptionsAfterInstrumentChange` 返回 `false`（gate 关闭）而 handler 仍是 enabled，上面无条件 publish 的 `perpsActiveOrderBookOptionsAtom` 照样经 watcher 触发 rebuild。触发条件就是「在 Perp tab 上打开任意 modal 期间发生 instrument 切换」，例如 token selector 里选币、或 modal 打开时收到 `PerpSwitchActiveInstrument`。blur 过渡的异步间隙同理——`disableSubscriptionsHandler()` 是 `void` 的跨桥调用，置位前有一小段窗口。

**关于 follow-up。** 同意留作 follow-up issue，不必在本 PR 处理。理由：现在注释已如实描述行为（best-effort 而非保证），残留缺口的范围是有界的——只在 modal 遮挡与 blur 过渡间隙触发，后果是多做一次本可省略的 rebuild，不产生错误数据；而把判断下沉到 BG 会改变所有入口的订阅生命周期语义，风险面远大于它在 bundle-release hotfix 里能带来的收益。排期由你们定。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Agreed on the conclusion, and it matches what I found independently before your retraction landed.

The one piece I can add as positive evidence rather than absence-of-evidence: I re-ran the exact flow on the iOS Simulator against head `3d2ca287` — CXMT perp active, opened the mobile token selector, picked XMR/USDC spot. The selection held, still XMR/USDC spot 60s later with a live book, and no `initial_symbol_latch_skip` in the log. If a remount had reset `isRouteFocusedRef`, that spot selection is the case that would have rolled back hardest, since both `coin` and `mode` diverge.

Worth recording for whoever reads those three `detaches this tab` comments next: they describe native view-layer detach, not React unmount. `GlobalScreenOptions.ts` sets `detachPreviousScreen: false`, `nativeTabScreenOptions` uses `freezeOnBlur: !platformEnv.isNativeIOS`, and there is no `unmountOnBlur` / `detachInactiveScreens` anywhere in the repo — so `PerpsGlobalEffects` stays mounted and its refs survive.

Resolving as fixed by the round-1 dedupe.

packages/shared/src/utils/notificationsUtils.ts
Minnzen: Fixed in 0a9483c53d — changed to `Defense`.

Confirmed the failure was real and blocking: `gh pr checks 12649` showed `lint (24.x) fail` on head `3d2ca287`, and `Defence` was the only occurrence in the repo, introduced by this PR.

Also swept the rest of the added lines for words that could hit the same rule. Everything else I introduced already has precedent on the base branch, so the dictionary accepts it: `prewarm` (125 occurrences), `dedupe` (178), `reconcile` (94), `idempotent` (30), `resubscribe` (6). `Defence` was the only new word.

Note the local `yarn agent:check --profile commit` cannot catch this class of issue in this workspace — `eslint-plugin-prettier`, `tsgo` and `typescript` are missing from `node_modules` and both lint and tsc steps fail identically on a clean branch, so oxlint never runs. CI is currently the only gate for it here.

packages/kit/src/hooks/useWalletBanner.ts
```diff
@@ -73,21 +73,30 @@ function useWalletBanner({
...
-              setPerpPageEnterSource(EPerpPageEnterSource.WalletBanner);
+        // A banner whose payload cannot be turned into a navigation target
+        // used to be a dead tap: the empty fallback swallowed the failure and
+        // the early return skipped the href handling below. Fall through
+        // instead so a banner that also carries an href still goes somewhere.
+        let payloadHandled = true;
+        parseNotificationPayload(item.mode, item.payload, () => {
+          payloadHandled = false;
+        });
+        if (payloadHandled) {
```
originalix, line 83: P2: [无效 WebView 页面仍会吞掉 banner 的 href 回退]

当 banner 同时带有 `mode=page`、指向 `ERootRoutes.WebView` 的无效或被 URL policy 拦截的 URL，以及可用 `href` 时，`parseNotificationPayload()` 会在 URL 校验处直接 `break`，不会 dispatch 导航，也不会调用这里传入的 fallback。因此 `payloadHandled` 仍为 `true`，随后提前 `return`，点击仍然没有结果且不会处理 `href`。

请让 `parseNotificationPayload()` 返回是否实际完成处理，并让 WebView URL 拒绝等 no-op 分支返回 `false`；这里仅在返回 `true` 时提前结束。

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Correction to my previous reply in this thread: one of the two pieces of evidence I gave does not hold, and I should not have stated it.

I wrote that there was no `initial_symbol_latch_skip` in the log during the simulator run. That log line could not have appeared regardless of whether the branch executed. `markPerpsColdStartPerf` reaches the local log only through `defaultLogger.perp.hyperliquid.coldStartBenchmark`, and `LoggerConfigManager.shouldLog` gates every `@LogToLocal` scene behind explicit opt-in on dev builds:

```ts
// loggerConfigManager.ts:86-91
shouldLog(scopeName: string, sceneName: string): boolean {
  if (!this._env.isDev) {
    return true;
  }
  return !!this._config?.enabled?.[scopeName]?.[sceneName];
}
```

with `createDefaultLoggerConfig` shipping `enabled: {}`. During the earlier diagnostic phase I had a temporary dev-only mirror writing these marks straight to the native log file, which is why marks were visible then; that mirror was removed during cleanup, before this verification run. So the absence of the line is not evidence of anything.

What does still stand:

- The behavioural result, which is what the bug is about: with CXMT perp active I picked XMR/USDC spot from the mobile token selector, and the selection held — still XMR/USDC spot with a live book 60s later, no rollback. Directly observed, twice, via screenshots.
- The static facts, which I have now checked myself rather than restating from your comment: `detachPreviousScreen: false` appears 9 times in `GlobalScreenOptions.ts`; `unmountOnBlur` and `detachInactiveScreens` have no source-level occurrence (the only hits are inside the built `apps/mobile/out-dir-bundle/ios/main.jsbundle`, i.e. react-native-screens own default); `freezeOnBlur: !platformEnv.isNativeIOS` in `nativeTabScreenOptions`; `Perp.tsx` renders `<PerpView/>` unconditionally on the main path.

What is not proven either way is the mechanism — whether `selectInitialSymbol` re-entered at all during that flow. The behavioural test shows no rollback occurred; it does not show the branch was never reached. If you want that closed properly I can re-run with the `perp/hyperliquid` scene enabled in the logger config so the marks actually land.

sidmorizon: 感谢主动更正——你是对的，我那条 A 项结论有误，撤回。

我当时只追到 `isPerpsColdStartPerfEnabled()` 恒为 `true` 和 `@LogToLocal` 装饰器存在就收手了，没有继续追到装饰器背后的落盘条件。完整链路是 `markPerpsColdStartPerf` → `coldStartBenchmark`（`@LogToLocal`）→ `_emitLog` → `logFn.processEntry`，而 `processEntry` 在 L54 取 `const shouldLog = loggerConfig.shouldLog(...)`，L75 的 `case 'local'` 到 L80 才是真正的门：`if (!platformEnv.isWebEmbed && shouldLog)`。再往下就是你引用的 `loggerConfigManager.ts:86-91`——dev build 上要求 `enabled[scopeName][sceneName]` 显式 opt-in，而 `createDefaultLoggerConfig`（`loggerConfigShared.ts:21`）出厂即 `enabled: {}`。所以 iOS 模拟器这种 dev build 上，未 opt-in 时 `perp`/`hyperliquid` scene 的 mark 根本不落盘，「日志里没有」确实不能作为证据。（注：`shouldLog` 在 `!isDev` 时直接 `return true`，所以 release build 不受影响，问题只在 dev。）

**不需要你再跑一次带 opt-in 的验证**——B/C 两项静态事实已经能单独闭合机制问题，不依赖任何日志。进入 `selectInitialSymbol` 只有两条途径，逐条排除：

1. **focus 监听（L1125-1131）**：需要 `resolvePerpRouteFocused(isFocus)` 发生 `false → true`。`useListenTabFocusState` 的 `isFocus` 只由 `ERootRoutes.Main` 那条 root route 的 tab state 推导，而 `pushModal` 推的是独立的 `ERootRoutes.Modal` root route，只体现在 `isHideByModal`，Main 的 tab index/routeNames 不变。因此 open → select → `popStack()` 全程 `isFocus === true`；配合组件不 unmount（`Perp()` 主路径无 `useIsFocused()` 门控、root/modal stack 统一 `detachPreviousScreen: false`、iOS 侧 `freezeOnBlur: !platformEnv.isNativeIOS` 连 freeze 都没有、全仓无 `unmountOnBlur`/`detachInactiveScreens`）——`isRouteFocusedRef` 不被重置，`wasFocused` 也恒为 `true`，`!nextFocused || wasFocused` 命中，早退。

2. **mount `useEffect`（L1133-1138）**：门是 `shouldTreatPerpAsFocusedOnMount`（L99-102），在 native 上恒为 `false`，effect 直接 return；且由第 1 点的「不 unmount」，该 effect 在此 flow 中本来也不会重跑。双重排除。

两条途径都被静态封死，与日志无关。

补一句以防将来用得上：若哪天确实要用日志复核这类问题，该盯的是 `initial_symbol_start` / `initial_symbol_claimed`（任何一次进入都会打），而不是 `initial_symbol_latch_skip`——后者只在 `!claimed && activeTradeInstrumentRef.current?.coin` 块内触发，即使函数被重入、只要 ctx coin 为空也不会出现。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

packages/kit/src/hooks/useWalletBanner.ts
Minnzen: Confirmed and implemented as you suggested, in 627199e32b.

Verified the gap: `notificationsUtils.ts:409` was a bare `break` with no `fallbackHandler()` call, so `payloadHandled` stayed `true` and the banner returned before reaching its `href`.

`parseNotificationPayload` now returns whether the payload was actually dispatched, and every no-op branch reports `false` — not just the WebView rejection you flagged, but the others in the same class:

- `page`: invalid payload, WebView URL rejected by policy, JSON parse failure
- `dialog`: JSON parse failure
- `openInBrowser` / `openInApp` / `openInDapp`: empty payload
- `openInApp`: URL rejected by `isAllowedWebViewUrl`
- `command`: missing `action`, JSON parse failure
- `default`: unrecognised mode (including anything `normalizeNotificationMode` cannot map)

`useWalletBanner` now branches on the return value instead of the mutable flag set by the fallback.

Two things I deliberately did not change:

- The `fallbackHandler()` call sites are untouched, so the notification path behaves exactly as before — only the banner reads the new signal.
- The URL policy itself is unchanged. The banner falling through to `item.href` is not a bypass: `href` is a separate server-provided field that goes through `parseQRCode.parse` with its own handlers, not the rejected WebView URL.

Also guarded `openInDapp` against an empty payload, which previously emitted `undefined as string` to the dapp subscriber.

packages/shared/src/utils/notificationsUtils.ts
```diff
@@ -366,12 +375,30 @@ export function parseNotificationPayload(
     };
     [key: string]: any;
   },
-) {
+): boolean {
+  // Returns whether the payload was actually dispatched. Several branches are
+  // deliberate no-ops (blocked WebView URL, empty payload, unknown mode) and
+  // do not run `fallbackHandler`, so callers that want to degrade gracefully
+  // — the wallet banner falling back to its own `href` — cannot rely on the
+  // handler alone to tell them nothing happened.
```
sidmorizon: # 🟡 P2 另一个带 href 兜底的调用方没有接上新返回值

## 问题

返回值语义本身实现得很干净，我逐条核对过：9 条 no-op 路径全部返回 `false`，成功 dispatch 全部返回 `true`，5 处 `fallbackHandler()` 调用点数量与分支完全未变，`openInApp` 的裸块 `{}` 内每条路径都有 `return`，不存在 fall-through 到 `openInDapp`。

但 `parseNotificationPayload` 有两个「拿不到结果就退回自己的 href」的消费方，这次只接上了一个。`showFeaturedChangelogDialog.tsx:42-58` 的 `dispatchFeatureCta` 仍是修复前 `useWalletBanner` 的形状：

```ts
if (activeFeature.mode !== undefined) {
  parseNotificationPayload(
    activeFeature.mode,
    activeFeature.payload ?? activeFeature.href,
    () => {
      if (isAllowedFeaturedHref(activeFeature.href)) {
        handleDeepLinkUrl({ url: activeFeature.href });
      }
    },
  );
  return;   // 无条件 return，href 只能经 fallbackHandler 到达
}
```

## 影响

对于**不调用** `fallbackHandler()` 的那几条分支——`default`（后端下发一个前端还不认识的 `mode`）、`page` 且 WebView URL 被策略拒绝、空 payload——fallback 不会执行，`return` 又跳过了后面的 href 处理，CTA 点了没有任何反应。这正是 originalix 报的那个缺陷，只是换了个调用方；该函数自己的注释写着「fall back to href so backends that only set href still work」，说明 href 可用本就是预期。

`default` 这条尤其现实：`mode` 来自后端 featured-changelog 配置，一旦下发了新增的 mode 值，老客户端就是死链接。改动后两个同类消费方行为不一致——banner 会退回 href，changelog CTA 不会。

## 建议

同样改成用返回值判定即可，一行：

```ts
const dispatched = parseNotificationPayload(
  activeFeature.mode,
  activeFeature.payload ?? activeFeature.href,
  () => {},
);
if (dispatched) return;
if (isAllowedFeaturedHref(activeFeature.href)) {
  handleDeepLinkUrl({ url: activeFeature.href });
}
return;
```

`isAllowedFeaturedHref` 仍在原位把关，安全边界不变。若认为超出本 PR 范围，留 follow-up 也可以，只是建议别让这个不一致长期存在。

[Claude session ↗](https://claude.ai/code/session_01XmwTQ6jZfgLx1FCGATK6ik)

---

Minnzen: Confirmed and fixed in 395d120e3a, implemented as you wrote it.

I should have swept the consumers when I introduced the return value rather than fixing only the one that was reported. Did that sweep now — every `parseNotificationPayload` call site:

| Call site | Has an href fallback | Status |
| --- | --- | --- |
| `useWalletBanner.ts:82` | yes (`item.href`) | uses the return value |
| `showFeaturedChangelogDialog.tsx:52` | yes (`activeFeature.href`) | fixed here |
| `DeviceGetStarted.tsx:39`, `:140` | no | nothing to hook up — `fetchGetStartedLinks` returns `{ linkId, title, mode, payload, image, description }` (`ServiceSetting.ts:558-566`), there is no href field and no alternative path after the call |
| `NotificationPayloadTest.tsx:108` | n/a | dev tool, uses the handler to show a toast |

So the two you identified are the complete set.

The `default` case you called out is the realistic one — `mode` comes from backend featured-changelog config, and `normalizeNotificationMode` maps anything unknown to `undefined`, so a newly introduced mode value turns older clients' CTA into a dead tap. That now degrades to the href.

`isAllowedFeaturedHref` still gates the href, and the fallback path keeps using `handleDeepLinkUrl` exactly as before rather than the `hrefType`-aware branch below it, so behaviour for the cases that already worked is unchanged.

Also swept the diff for British spellings after the earlier `Defence` CI failure — `unrecognised` had slipped into this comment (0 precedent on base, `unrecognized` has 9), corrected before pushing.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -2661,22 +2661,44 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+      // `universesByDex` is `[]` when nothing is cached yet, so `[dexIndex]`
+      // is `undefined` and the old `?.length === 0` test was never true —
+      // execution fell through and committed `assetId: -1` as if it were a
+      // resolved asset. Keep the seeded coin (l2Book subscribes by coin, so
+      // the target stays correct) but leave the metadata unresolved.
+      if (!dexUniverses?.length) {
+        // Re-selecting the coin that is already active must not downgrade
+        // metadata we already resolved; only an actual switch falls back to
+        // the unresolved shape.
+        if (!shouldSeedSubscriptionTarget && oldActiveAsset?.coin === newCoin) {
```
originalix: P2: [同币重选会保留已持久化的无效 assetId]

当旧版本在 universe 尚未缓存时已经把 `assetId: -1` 写入持久化的 `perpsActiveAssetAtom`，升级后离线或冷启动并重选同一 coin 会进入这个分支。这里仅按 coin 相等直接返回旧记录，正好保留了本次修复要清除的 `-1`，导致下游继续把无效 metadata 当成已解析值，目标 orderbook 与 subscription 无法可靠收敛。

请只在旧记录确实完整有效（例如 `universe` 存在且 `assetId >= 0`）时原样返回；否则也写入下面的 unresolved shape，让后续 metadata refresh 重新解析。

Minnzen: 已修复,感谢指出 — 这个分支确实会把持久化的 `assetId: -1` 原样带回来。

`perpsActiveAssetAtom` 是 `persist: true` 的 globalAtom,旧构建写入的 `-1` 会跨升级存活;原来只按 coin 相等就短路返回,正好保住了这个分支本来要清掉的值。

现在要求旧记录**确实已解析**才复用,否则照常落到 unresolved shape,交给后续 metadata refresh 重新解析:

```ts
if (
  !shouldSeedSubscriptionTarget &&
  oldActiveAsset &&
  oldActiveAsset.coin === newCoin &&
  oldActiveAsset.universe &&
  typeof oldActiveAsset.assetId === 'number' &&
  oldActiveAsset.assetId >= 0
) {
  return oldActiveAsset;
}
```

没有把 `margin` 纳入判定 — 部分标的本来就没有 margin table(日志里可见 `hasMargin=false` 的正常标的),要求它存在会把有效记录误判成未解析。

commit: 9b233b26cb

originalix (approved):

PR #12647, opened 2026-07-27
Add terminal payment session discard for closed unpaid invoices

sidmorizon:
## Summary

Adds support for releasing claimed invoices that the server closed with nothing collected, allowing users to change their asset selection instead of being stranded on a dead payment. This introduces a new `discardTerminalPaymentSession` callback that performs an atomic check-and-delete operation with optimistic concurrency control.

## Key Changes

- **primeInfiniPaymentReplacement.ts**: Added `discardTerminalPaymentSession` parameter to `resolvePrimeInfiniPaymentReplacement()`. When a payment is closed unpaid and `sendStarted` is latched (preventing normal discard), attempts terminal discard. Only reports replacement on successful atomic delete.

- **SimpleDbEntityPrime.ts**: Enhanced `discardTerminalInfiniPendingPaymentSession()` with optimistic concurrency checks (`expectedUpdatedAt`, `expectedSendStarted`). Prevents race conditions where another window claims the same invoice while the terminal snapshot is in flight, which would incorrectly delete a fresh claim.

- **PrimeInfiniWalletPayment.tsx**: Implemented `discardTerminalPaymentSessionForSelectionChange()` callback that:
  - Fetches the current persisted session state
  - Calls the atomic discard with expected revision values
  - Returns success/failure to the replacement resolver

- **primeInfiniExternalCheckoutGuard.ts**: Updated to pass `expectedUpdatedAt` and `expectedSendStarted` when calling terminal discard.

- **Tests**: Added comprehensive test coverage for:
  - Successful terminal discard of closed unpaid invoices
  - Refusal to discard when terminal release is rejected
  - Race condition handling when session is claimed during snapshot flight
  - Updated existing tests to provide required `discardTerminalPaymentSession` parameter

## Implementation Details

The terminal discard uses optimistic concurrency control by comparing the stored session's `updatedAt` timestamp and `sendStarted` flag against the values observed when the caller fetched the session. This ensures the atomic delete only succeeds if the session hasn't been modified by another window, preventing data loss in concurrent scenarios.

https://claude.ai/code/session_012kdtSAriNZaXGu2W8v9bxE

PR #12671, opened 2026-07-29
fix: tier the Infini session TTL and keep the new-payment choice reachable everywhere

sidmorizon:
## What

Follow-up to #12647, with one goal — a pending Infini session must never strand the user without a manual way to start a new payment, and never silently drop the fence around funds that may still be in flight:

1. Tiered session TTL: unsent sessions expire after **1 day** (down from 7); sent sessions (broadcast claimed or durable progress recorded) keep the **7-day** bound.
2. Every entry path with a pending session now funnels through the existing-payment choice screen, where the user can force a new payment.

## Why

The choice screen shipped in #12647 was only reachable on some paths:

- **Entry gate**: with a pending session, `getPrimeInfiniPaymentEntryGuard` fetched the invoice unconditionally; a persistent fetch failure threw, and `PrimePurchaseDialog` blocked **every** purchase channel behind a toast. The one screen able to release the session sat unreachable behind the gate, leaving the TTL as the only exit — and a 7-day lockout.
- **Resume-intent entries**: both the fresh choice screen and the stale fallback screen required the payment-creation intent, so an entry that came in to resume (exactly what the dialog's auto-resume produces) landed on a plain retry screen with no release option when the load failed.

A uniform 1-day TTL was the first cut, but Codex review correctly flagged the double-payment risk (P1): a broadcast transfer can settle after 24h (low fee, congestion, server outage), and silently expiring its session would let the entry gate admit an IAP/Stripe/crypto purchase while the first transfer can still arrive. Hence the tiered bound: the short TTL only applies to sessions that commit no funds.

## Changes

- `SimpleDbEntityPrime.ts`: `INFINI_UNSENT_PAYMENT_SESSION_MAX_AGE_MS` (1d) for never-sent sessions; `INFINI_SENT_PAYMENT_SESSION_MAX_AGE_MS` (7d) for sent sessions, payment-cache tombstones, and the superseded archive — they fence the same in-flight risk. `isValidInfiniPendingPaymentSession` picks the bound from `sendStarted` / recorded progress.
- `primeInfiniExternalCheckoutGuard.ts`: on invoice-fetch failure the entry guard no longer throws. It degrades to the local snapshot: a locally replaceable session (never sent, no recorded progress) keeps the method picker open; anything else reports a pending payment so the dialog resumes the crypto flow, whose stale fallback screen can force a replacement. The double-charge invariant is preserved — unknown state never opens a second channel and never mints a new invoice by itself.
- `PrimeInfiniWalletPayment.tsx`: the fresh choice screen and the stale fallback screen are no longer gated on the creation intent. Every entry with a non-replaceable pending session pauses at the choice once (continue → resumes exactly as before; start new → forced replacement, then back to the method picker).
- Tests: unsent 2-day-old session purged with tombstone written; sent 2-day-old session retained; guard degraded-mode tests (claimed session → pending reported, no discard/latch; replaceable session → picker stays open).

## What still can't be skipped

`resolvePrimeInfiniPaymentForcedReplacement` keeps its hard dependency on the purchase-status endpoint: "was the subscription already granted" cannot be degraded without reintroducing a double charge. If that endpoint is down, forcing a new payment fails with a retryable toast while the user stays on the choice screen.

## Notes

- With the choice screen reachable everywhere, the 7-day sent bound no longer locks anyone out by itself — sent sessions release earlier via the server-verified terminal path or the user's explicit forced replacement; the only scenario waiting on a TTL is a full backend outage, where no purchase channel works anyway.
- Worst-case unsent lockout with the unchanged 24h clock-skew allowance is ~48h for a device whose clock was a full day ahead at write time.
- `yarn jest` — entity: 84/84; Prime hooks: 191/191. `yarn agent:check --profile commit` passes.


packages/kit/src/views/Prime/components/PrimeInfiniWalletPayment.tsx
```diff
@@ -3694,8 +3694,11 @@ function PrimeInfiniWalletPaymentRoot({
       if (!hasError) {
         paymentCreationIntentRef.current = false;
       }
+      // Deliberately not gated on the payment creation intent: every entry
+      // with a non-replaceable pending session pauses here once, resume
+      // included, so the user always passes the one screen that can force a
+      // new payment instead of being funneled straight into polling.
       const shouldShowExistingPaymentChoice = Boolean(
```
sidmorizon, line 3734: # [🟠 P1] 去掉 intent 门控后，同一 mount 内的 mid-flow reload 会把刚广播的付款打回 choice 屏

## 问题

移除 `shouldCreatePayment &&` 后，`shouldShowExistingPaymentChoice` 在 loader 的**每次** re-run 都会对非 replaceable session 为 true，而不仅仅是"入口路径"。`onReloadPaymentSession`（即 `run`）在 content 内部大量恢复路径中会在 `sendStarted === true` 之后触发（如 `result.type === 'reload'`（约 L2900、L3060）、`'track' && !isAttemptOwned()`（约 L1508、L1756）、attempt-stale 分支等）。此时 `continuedExistingPaymentBindingId` 仍是 `''` —— 本 mount 内新建并广播的 session 从未经过 choice 屏，`freshExistingPaymentChoiceSession` 的抑制条件不成立，`paymentContent` 中 choice 屏优先于 `PrimeInfiniWalletPaymentContent` 渲染。

## 影响

用户刚在本屏广播交易，一次内部触发的 reload 就会把正在 polling 的流程打断，弹出"继续/发起新付款"的选择屏。此时点 `Start new payment` 走 `resolvePrimeInfiniPaymentForcedReplacement`，而刚广播、服务端尚未观测到 `amountConfirming` 的窗口正是 progress 检测拦不住的时段，supersede 后再开新 invoice 即重复付款。注释里 "pauses here once" 的假设对本 mount 内创建的 binding 不成立：每次 reload 都会重新暂停。

## 建议

对本 mount 内已进入 send/polling 的 binding 做抑制：在 send 启动（或 `persistPaymentSession` 标记 `nextSendStarted: true`）时把该 `bindingId` 上报给 Root 并计入 `continuedExistingPaymentBindingId`（或一个 `Set`），使 `freshExistingPaymentChoiceSession` 只在真正的入口加载（首个 loadAttempt）或用户尚未经过 choice 屏的 resume 入口出现。

[Claude session ↗](https://claude.ai/code/session_018a37LEvoAAmSWesqCuUktC)

---

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityPrime.ts
```diff
@@ -37,7 +37,21 @@ import {
 } from '../../../services/ServicePrime/primeAuthSessionAccess';
 import { SimpleDbEntityBase } from '../base/SimpleDbEntityBase';
 
-const INFINI_PENDING_PAYMENT_SESSION_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000;
+// Upper bound on how long an unsent pending session may block the purchase
+// entry when every server-side release path stays unreachable (for example the
+// invoice endpoint failing on that paymentId). It is the worst-case lockout, so
+// it must stay short; sessions with reachable escapes are released long before
+// this, and an unsent invoice commits no funds.
```
sidmorizon, line 44: # [🟡 P2] 缩短该常量同时把 tombstone / superseded 归档窗口从 7 天砍到 1 天

## 问题

与已有评论（session 过早被清除）不同的另一个副作用：`INFINI_PENDING_PAYMENT_SESSION_MAX_AGE_MS` 还被 `getActiveInfiniPaymentCacheTombstones`（L157）和 `getActiveInfiniSupersededPaymentSessions`（L224）复用为存活窗口。本次改动把 tombstone 的 stale-writer 防护和 forced-replacement 归档保留期也一并从 `7d` 缩到 `24h`，而 L40-43 的新注释只论证了 session lockout 需要变短，未覆盖这两个安全记录。

## 影响

tombstone 的作用是阻止滞留窗口把已 retire 的 binding 通过 `setInfiniPendingPaymentSession`（L1155 的 `isRetiredPayment` 检查）重新持久化。现在只要另一个窗口/被挂起的 App 持有旧 session 超过 `24h`（此前需要 `7d`），其恢复写入就不再被拦截，可能复活一个已 discard/已完成的 invoice 并重新占住 entry gate。本 PR 新增测试还专门断言 "tombstone survives"，但其实际保护窗口被隐式缩短了 7 倍。

## 建议

把三个用途解耦，session lockout 用新的 1 天值，安全记录保持原值：

```ts
const INFINI_PENDING_PAYMENT_SESSION_MAX_AGE_MS = 24 * 60 * 60 * 1000;
// Stale-writer protection and replacement audit records keep the longer window.
const INFINI_PAYMENT_CACHE_TOMBSTONE_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000;
```

并在 L157 / L224 改用后者。

[Claude session ↗](https://claude.ai/code/session_018a37LEvoAAmSWesqCuUktC)

---

sidmorizon:
Re: Codex P1 (保留已发送会话直到服务端确认终态, https://github.com/OneKeyHQ/app-monorepo/pull/12671#discussion_r3670614254) — fixed in 5aa40fdfce. Replying here because a pending review currently blocks thread replies.

The TTL is now tiered instead of uniform:

- `INFINI_UNSENT_PAYMENT_SESSION_MAX_AGE_MS` (1 day): only sessions never broadcast and with no recorded progress may expire on the short bound — they commit no funds.
- `INFINI_SENT_PAYMENT_SESSION_MAX_AGE_MS` (7 days): a session with `sendStarted` latched or durable progress keeps fencing the purchase entry, so `getInfiniPendingPaymentSession` no longer purges it before the entry gate can consult the server. Tombstones and the superseded archive stay on this bound, since they fence the same in-flight risk.

With the choice screen reachable on every entry path (764d076a), the long bound doesn't lock users out on its own: a sent session is released earlier either by the server-verified terminal path or by the user's explicit forced replacement; the only scenario still waiting on the TTL is a full backend outage, where no purchase channel works anyway.

Regression tests added: a two-day-old sent session survives the read path; a two-day-old unsent session is purged with its tombstone written.

---

packages/kit/src/views/Prime/components/PrimeInfiniWalletPayment.tsx
sidmorizon: Fixed in 4c7725c4ae. The root now keeps a per-mount `handledPaymentBindingIds` set; the content reports every binding it persists via a new `onPaymentSessionPersisted` callback (funneled through `persistPaymentSession`, which every in-flow write path goes through — pre-send refresh, progress latch, error re-persist). Both `freshExistingPaymentChoiceSession` and `staleExistingPaymentChoiceSession` filter those bindings out, so a mid-flow `run()` from the reload/track/attempt-stale branches can no longer bounce a just-broadcast payment back to the choice screen — exactly the window where "start new payment" would slip past the progress checks. Entry behavior is unchanged: a binding first seen by this mount's loader still pauses at the choice once, and a remount starts clean.

---

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityPrime.ts
sidmorizon: Addressed in 5aa40fdfce (one commit before this thread landed). The constants are now tiered: `INFINI_UNSENT_PAYMENT_SESSION_MAX_AGE_MS` (1d) applies only to never-sent sessions, while `INFINI_SENT_PAYMENT_SESSION_MAX_AGE_MS` (7d) covers sent sessions **and** is the bound used by `getActiveInfiniPaymentCacheTombstones` and `getActiveInfiniSupersededPaymentSessions` — so the stale-writer fence and the replacement audit records keep their full 7-day window, matching what this thread proposes (same split, different constant name). The "tombstone survives" test now runs against the 7-day bound.

---

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityPrime.ts
```diff
@@ -37,7 +37,11 @@ import {
 } from '../../../services/ServicePrime/primeAuthSessionAccess';
 import { SimpleDbEntityBase } from '../base/SimpleDbEntityBase';
 
-const INFINI_PENDING_PAYMENT_SESSION_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000;
+// Upper bound on how long a pending session may block the purchase entry when
+// every server-side release path stays unreachable (for example the invoice
+// endpoint failing on that paymentId). It is the worst-case lockout, so it must
+// stay short; sessions with reachable escapes are released long before this.
+const INFINI_PENDING_PAYMENT_SESSION_MAX_AGE_MS = 24 * 60 * 60 * 1000;
```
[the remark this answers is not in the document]
sidmorizon: Fixed in 5aa40fdfce (see also https://github.com/OneKeyHQ/app-monorepo/pull/12671#issuecomment-5112300283, posted while a pending review blocked thread replies). Sent sessions — `sendStarted` latched or durable progress recorded — now stay on the 7-day bound and are no longer purged by `getInfiniPendingPaymentSession` before the entry gate can consult the server; only never-sent sessions, which commit no funds, expire on the 1-day bound. Regression test added: a two-day-old sent session survives the read path.

---

packages/kit/src/views/Prime/hooks/primeInfiniExternalCheckoutGuard.ts
```diff
@@ -62,11 +65,35 @@ export async function getPrimeInfiniPaymentEntryGuard() {
...
+    // gate. Fall back to the local snapshot instead: an invoice that never
+    // left this device commits no funds and keeps the picker open, anything
+    // else resumes the crypto flow, whose stale fallback screen still lets
+    // the user force a replacement.
+    const localSendStarted =
+      pendingPaymentSession.sendStarted ||
+      hasPrimeInfiniPaymentProgress(pendingPaymentSession.payment);
+    return {
+      isLoggedIn: true,
+      hasPendingPayment: !isPrimeInfiniPaymentReplaceable({
```
originalix: P1: [网络降级使用旧会话快照，可能放行重复购买]

当这里等待 `apiGetInfiniPayment` 时，另一窗口可能已经把同一会话原子标记为 `sendStarted`。请求失败后仍使用等待前的 `pendingPaymentSession`，会把该会话判断为可替换并返回 `hasPendingPayment: false`，随后 IAP/Stripe 可以继续，而 crypto 转账已经获准广播，存在重复扣费风险。

请在本地快照可替换时先用 `discardUnsentInfiniPendingPaymentSession` 和预期 cache key 原子退休该会话；仅在退休成功时开放第二渠道，若操作因发送状态已变化而失败则继续按 pending payment 处理。

sidmorizon: Fixed in fd0309b90f, exactly as proposed. The degraded branch no longer trusts the pre-fetch snapshot: when the local snapshot looks replaceable, it calls `discardUnsentInfiniPendingPaymentSession` with the expected cache key and only reports `hasPendingPayment: false` when that atomic retire succeeds. The discard refuses once `sendStarted` is latched (and a broadcast cannot be marked on a session that no longer exists, so a successful retire proves no other window holds an authorized broadcast). Refusal or a thrown discard both fall back to `hasPendingPayment: true`, resuming the crypto flow. Tests: the replaceable-degrade case now asserts the retire call, and a new case pins the race — discard refused → picker stays closed.

---

originalix (approved):

zhaono1 (approved):

PR #12675, opened 2026-07-29
fix: stop Discovery tab render storm and make webview keep-alive eviction work

originalix:
## Background

The desktop window renderer can grow past 1 GB and peg a full CPU core after the DApp browser has been used for a while. Restarting the app clears it; it returns after further browser use. The growth is triggered by DApp browser usage, not by uptime — a desktop session that never opens a browser tab stays flat.

Reproduced locally: 26 tabs open (including several trading DApp pages), one tab activation every 20 s. Within two hours the window renderer went from 475 MB to 1043 MB, and a heap snapshot diff showed `FiberNode` growing from 29,603 to 147,821 (5x) while DOM nodes only went 6,291 → 24,653 — a 6:1 fiber-to-DOM ratio against a typical 2–3:1.

## Root cause

Three defects compound; none is fatal on its own.

**1. Title storm.** Trading DApps stream the live price through `document.title`, so `page-title-updated` fires roughly once per second per tab. The 500 ms debounce in `onNavigation` sits inside `if (isValidNewUrl)`, so title-only events fall straight through to the `setWebTabData` call at the bottom. `setWebTabData` then mutated the array held by `webTabsAtom` in place and called `buildWebTabs` **unconditionally**, and `buildWebTabs`' `isEqual` only compared the tab id list — the `webTabsMapAtom` replacement and the full-array persist to SimpleDb were both unconditional.

Measured locally: **40–55 writes/min with the app idle and no user interaction**, dropping to zero the moment the app navigates off the Discovery route. The existing comment notes the debounce was added because of "~65 writes in 4 minutes" on iPad (16/min) — this is 55/min *with* that debounce in place.

**2. Every write re-renders every tab.** `useWebTabDataById` subscribes to the whole map and memoises on `map`, so any tab's change hands a fresh object to all mounted tab shells; desktop `WebContent` had no memo boundary, so the entire webview subtree was rebuilt with it. Allocation sampling put **39% of all allocations in Tamagui's `getStylesAtomic`/`createAtomicRules`** — recomputing styles over an oversized tree is exactly the shape of the pegged core. Notably the JS main thread stays responsive throughout (long tasks are rare), because the cost is spread across high-frequency small renders and GC rather than any single blocking loop.

**3. LRU eviction never actually unmounted anything.** An evicted tab is by definition inactive and therefore frozen, and react-freeze suspends by throwing from a Suspender — so the render that sets `body` to null **could never commit**. React kept the previous children mounted and the `<webview>`, along with its guest renderer process, survived eviction. Measured: **91% of samples had more live webviews than the limit of 8, sitting steadily at 10 and 16–17**. Knock-on effect: `releaseDesktopWebviewResources` drops the ref while the guest process is still running, so it is never reclaimed.

Separately, the navigation bar cached the `<webview>` element in React state, refreshed by an effect keyed on `tab.refReady`. `refReady` is set to true exactly once per tab and never reset, so after an eviction replaced the webview the effect never fired again — every nav bar kept pointing at a destroyed element and pinned the old WebContent subtree through its `__reactFiber$` expando.

## Changes

Split into three independently revertable commits:

**`fix: stop per-second browser tab persist and full re-render storm`**
- Throttle title-only navigation events to one accepted update per 5 s per tab, and drop the title when it matches the stored one
- `setWebTabData` copies the tab instead of mutating the array held by `webTabsAtom`, and bails out when the payload changes nothing
- `buildWebTabs` compares tab payloads, not just the id list, before replacing the map and scheduling a persist
- `useWebTabDataById` memoises on the tab rather than the map; desktop `WebContent` gets a memo boundary

**`fix: stop desktop browser nav bar retaining destroyed webviews`**
- Resolve `innerRef` on demand, bypassing the never-reset `refReady`
- Drop the url from the Freeze key (it previously discarded and remounted the whole InfoBar subtree on every in-page navigation)

**`fix: make desktop webview keep-alive eviction actually unmount`**
- Return null for a cold tab **above** the Freeze boundary so the unmount commits

## Behaviour change (please focus review here)

Commit 3 changes behaviour: cold tabs beyond `MAX_ALIVE_WEBVIEW_COUNT` (8 on desktop) now **genuinely unmount and reload their page when reactivated**. This was always the intent of the LRU — `DesktopBrowserContent`'s own comment says "remounts and reloads when activated again" — it simply never happened. If product finds the reload disruptive, the right response is to raise `MAX_ALIVE_WEBVIEW_COUNT`, not to restore the old no-op eviction.

## Verification

Typecheck is clean (0 errors) against the `origin/x` baseline, and `lint-staged` passes. `tsc:staged` could not run locally because `@typescript/native-preview` is not installed here (`npx tsgo` 404s), so a real `tsc` run against the root config was used instead — **this one still needs confirming on CI**.

Once a bundle is built, suggest regressing with the same protocol (26 tabs, several trading DApp pages, one activation every 20 s) against the pre-fix baseline:

| Metric | Before fix (measured) | Expected |
|---|---|---|
| `setRawData` rate | 40–55/min | < 5/min |
| Live webviews | 10–17 (limit 8) | ≤ 8 |
| Window renderer memory (2 h churn) | 475 → 1043 MB | well under 1 GB, plateauing |
| FiberNode | 29,603 → 147,821 | growth largely flattened |

## Not included

Tamagui 1.108's module-level `_idToUID` leak (`node_modules/@tamagui/web/dist/cjs/hooks/useTheme.js` retains every ThemeManager forever, with no deletion path anywhere in dist) is a separate issue requiring patch-package, and is left out of this PR.

### DIFF
--- a/packages/kit/src/states/jotai/contexts/discovery/actions.ts
+++ b/packages/kit/src/states/jotai/contexts/discovery/actions.ts
@@ -1601,18 +1601,21 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
           if (title && title !== tab.title) {
             pendingTitleUpdates[tabId] = {
               title,
-              timer: setTimeout(() => {
-                const pending = pendingTitleUpdates[tabId];
-                delete pendingTitleUpdates[tabId];
-                if (!pending) {
-                  return;
-                }
-                lastTitleUpdateFlags[tabId] = Date.now();
-                this.setWebTabData.call(set, {
-                  id: tabId,
-                  title: pending.title,
-                });
-              }, TITLE_UPDATE_THROTTLE_MS - (now - lastTitleAt)),
+              timer: setTimeout(
+                () => {
+                  const pending = pendingTitleUpdates[tabId];
+                  delete pendingTitleUpdates[tabId];
+                  if (!pending) {
+                    return;
+                  }
+                  lastTitleUpdateFlags[tabId] = Date.now();
+                  this.setWebTabData.call(set, {
+                    id: tabId,
+                    title: pending.title,
+                  });
+                },
+                TITLE_UPDATE_THROTTLE_MS - (now - lastTitleAt),
+              ),
             };
           }
           return;
--- a/packages/kit/src/states/jotai/contexts/discovery/actions.ts
+++ b/packages/kit/src/states/jotai/contexts/discovery/actions.ts
@@ -930,6 +930,11 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
           delete webviewRefs[id];
         }
       }
+      // Same as closeWebTab: drop any title update still waiting on its
+      // throttle window for a tab that is going away.
+      for (const tab of tabsToClose) {
+        clearPendingTitleUpdate(tab.id);
+      }
 
       loggerForEmptyData(pinnedTabs, 'closeAllWebTabs');
       // Same rationale as closeWebTab: persist the final (pinned-only) array
--- a/packages/kit/src/states/jotai/contexts/discovery/actions.ts
+++ b/packages/kit/src/states/jotai/contexts/discovery/actions.ts
@@ -218,6 +218,23 @@ export const lastNavigationFlags: Record<string, number> = {};
 export const lastTitleUpdateFlags: Record<string, number> = {};
 const TITLE_UPDATE_THROTTLE_MS = 5000;
 
+// Titles suppressed inside the throttle window, kept so the window can close
+// with a trailing update. Without this a page that sets a placeholder title
+// and then its final one within the window would keep the placeholder forever,
+// because nothing else would ever fire page-title-updated again.
+const pendingTitleUpdates: Record<
+  string,
+  { title: string; timer: ReturnType<typeof setTimeout> }
+> = {};
+
+export function clearPendingTitleUpdate(tabId: string) {
+  const pending = pendingTitleUpdates[tabId];
+  if (pending) {
+    clearTimeout(pending.timer);
+    delete pendingTitleUpdates[tabId];
+  }
+}
+
 let discoveryHomeBookmarksPrefetchGeneration = 0;
 let isDiscoveryHomeBookmarksPrefetchListenerReady = false;
 
@@ -430,23 +447,35 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
         newTabs = [];
       }
       const result = buildWebTabData(newTabs);
-      // Should update tabs
-      if (!isEqual(result.keys, webTabs.keys) || options?.forceUpdate) {
-        set(webTabsAtom(), { keys: result.keys, tabs: result.data });
-      }
 
       // `keys` only covers the tab id list, so a change to a tab's own fields
-      // (title/url/loading/favicon) never showed up here. The map replacement
-      // and the persist below were therefore unconditional: every no-op write
-      // handed all mounted tab shells a fresh map object and re-serialised the
-      // whole tab array to SimpleDb. Compare the tab payloads too, and skip
-      // both when nothing actually moved.
-      const tabsUnchanged =
-        !options?.forceUpdate && isEqual(result.data, webTabs.tabs);
-      if (tabsUnchanged) {
+      // (title/url/loading/favicon) never showed up here: the map replacement
+      // and the persist below used to be unconditional, so every no-op write
+      // handed all mounted tab shells a fresh map object and wrote the whole
+      // tab array back to SimpleDb. Compare the tab payloads too.
+      const keysChanged = !isEqual(result.keys, webTabs.keys);
+      const tabsChanged = !isEqual(result.data, webTabs.tabs);
+      // An Immediate persist is an authoritative snapshot from a discrete
+      // action (closing one tab / all tabs) and must never be skipped: it also
+      // cancels a pending trailing write, and its caller may have already
+      // mutated the array this comparison reads as the baseline.
+      const isImmediatePersist =
+        options?.persist === EBrowserTabPersistMode.Immediate;
+      if (
+        !keysChanged &&
+        !tabsChanged &&
+        !options?.forceUpdate &&
+        !isImmediatePersist
+      ) {
         return;
       }
 
+      // Both atoms must advance together. Writing webTabsAtom only on a key
+      // change left field-only updates (title/loading/favicon/isPinned) with a
+      // stale webTabsAtom while webTabsMapAtom and the persisted snapshot moved
+      // on, so the next action read the old array and clobbered the field that
+      // had just been written.
+      set(webTabsAtom(), { keys: result.keys, tabs: result.data });
       set(webTabsMapAtom(), () => result.map);
       loggerForEmptyData(result.data, 'buildWebTabs->saveToSimpleDB');
       if (options?.persist === false) {
@@ -755,6 +784,9 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
     ) => {
       const { tabId, entry, navigation } = payload;
       delete webviewRefs[tabId];
+      // Drop any title update still waiting on its throttle window, so it
+      // cannot resurrect an entry for a tab that no longer exists.
+      clearPendingTitleUpdate(tabId);
       const mountOrder = get(webTabMountOrderAtom());
       if (mountOrder.includes(tabId)) {
         set(
@@ -762,14 +794,18 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
           mountOrder.filter((id) => id !== tabId),
         );
       }
-      const { tabs } = get(webTabsAtom());
-      const targetIndex = tabs.findIndex((t) => t.id === tabId);
+      const { tabs: previousTabs } = get(webTabsAtom());
+      const targetIndex = previousTabs.findIndex((t) => t.id === tabId);
+      // Copy-on-write. Splicing the array held by webTabsAtom in place made it
+      // its own "previous state", so any equality check downstream compared the
+      // post-close array against itself and concluded nothing had changed.
+      let tabs = previousTabs;
       if (targetIndex !== -1) {
-        const closedTab = tabs[targetIndex];
+        const closedTab = previousTabs[targetIndex];
         const activeTabId = get(activeTabIdAtom());
         const isClosingCurrentTab =
           closedTab.isActive || activeTabId === closedTab.id;
-        tabs.splice(targetIndex, 1);
+        tabs = previousTabs.filter((_, index) => index !== targetIndex);
 
         // Add to browser history when tab is closed
         if (closedTab.url && closedTab.url !== homeTab.url) {
@@ -805,7 +841,9 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
           const newActiveTab = tabs[newActiveTabIndex];
 
           if (newActiveTab) {
-            newActiveTab.isActive = true;
+            // Replace the entry rather than flipping isActive on the shared
+            // tab object — it is still referenced by the pre-close array.
+            tabs[newActiveTabIndex] = { ...newActiveTab, isActive: true };
             this.setCurrentWebTab.call(set, newActiveTab.id);
           } else if (platformEnv.isDesktop) {
             // if the new active tab is not in tabs, switch to Discovery (Desktop only)
@@ -1553,14 +1591,35 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
           canGoForward === undefined &&
           loading === undefined;
         const lastTitleAt = lastTitleUpdateFlags[tab.id];
-        if (
-          isTitleOnlyUpdate &&
-          lastTitleAt &&
-          now - lastTitleAt < TITLE_UPDATE_THROTTLE_MS
-        ) {
+        const withinThrottleWindow =
+          Boolean(lastTitleAt) && now - lastTitleAt < TITLE_UPDATE_THROTTLE_MS;
+        if (isTitleOnlyUpdate && withinThrottleWindow) {
+          // Hold the newest title and flush it when the window closes, so the
+          // last value inside a burst is never lost.
+          const tabId = tab.id;
+          clearPendingTitleUpdate(tabId);
+          if (title && title !== tab.title) {
+            pendingTitleUpdates[tabId] = {
+              title,
+              timer: setTimeout(() => {
+                const pending = pendingTitleUpdates[tabId];
+                delete pendingTitleUpdates[tabId];
+                if (!pending) {
+                  return;
+                }
+                lastTitleUpdateFlags[tabId] = Date.now();
+                this.setWebTabData.call(set, {
+                  id: tabId,
+                  title: pending.title,
+                });
+              }, TITLE_UPDATE_THROTTLE_MS - (now - lastTitleAt)),
+            };
+          }
           return;
         }
         if (title !== tab.title) {
+          // A value is going through now, so any held one is stale.
+          clearPendingTitleUpdate(tab.id);
           lastTitleUpdateFlags[tab.id] = now;
         } else {
           nextTitle = undefined;
--- a/packages/kit/src/states/jotai/contexts/discovery/actions.ts
+++ b/packages/kit/src/states/jotai/contexts/discovery/actions.ts
@@ -212,6 +212,12 @@ export const homeResettingFlags: Record<string, number> = {};
 // latter also drives sidebar sort order (`top` mode freezes it on creation).
 export const lastNavigationFlags: Record<string, number> = {};
 
+// Tracks last accepted title update per tab id. Trading DApps push a live
+// price through document.title, so title-only navigation events arrive at
+// ~1Hz per tab and would otherwise persist the whole tab array that often.
+export const lastTitleUpdateFlags: Record<string, number> = {};
+const TITLE_UPDATE_THROTTLE_MS = 5000;
+
 let discoveryHomeBookmarksPrefetchGeneration = 0;
 let isDiscoveryHomeBookmarksPrefetchListenerReady = false;
 
@@ -429,6 +435,18 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
         set(webTabsAtom(), { keys: result.keys, tabs: result.data });
       }
 
+      // `keys` only covers the tab id list, so a change to a tab's own fields
+      // (title/url/loading/favicon) never showed up here. The map replacement
+      // and the persist below were therefore unconditional: every no-op write
+      // handed all mounted tab shells a fresh map object and re-serialised the
+      // whole tab array to SimpleDb. Compare the tab payloads too, and skip
+      // both when nothing actually moved.
+      const tabsUnchanged =
+        !options?.forceUpdate && isEqual(result.data, webTabs.tabs);
+      if (tabsUnchanged) {
+        return;
+      }
+
       set(webTabsMapAtom(), () => result.map);
       loggerForEmptyData(result.data, 'buildWebTabs->saveToSimpleDB');
       if (options?.persist === false) {
@@ -597,42 +615,56 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
 
   setWebTabData = contextAtomMethod((get, set, payload: Partial<IWebTab>) => {
     const { tabs: previousTabs } = get(webTabsAtom());
-    const tabs = previousTabs;
-    const tabIndex = tabs.findIndex((t) => t.id === payload.id);
-    if (tabIndex > -1) {
-      const tabToModify = tabs[tabIndex];
-      Object.keys(payload).forEach((k) => {
-        const key = k as keyof IWebTab;
-        const value = payload[key];
-        if (value !== undefined && value !== tabToModify[key]) {
-          if (key === 'title') {
-            if (!value) {
-              return;
-            }
-          }
-          // @ts-expect-error
-          tabToModify[key] = value;
-          if (key === 'url') {
-            // Navigation normally bumps timestamp to Date.now(), which
-            // re-sorts the tab to the bottom. Skip when the user chose
-            // 'top' so the tab stays where it was created. Record the
-            // navigation time separately for the onNavigation debounce.
-            if (!isNewTabPositionTop()) {
-              tabToModify.timestamp = Date.now();
-            }
-            if (payload.id) {
-              lastNavigationFlags[payload.id] = Date.now();
-            }
-            if (value === 'about:blank' && payload.id) {
-              homeResettingFlags[payload.id] = Date.now();
-            }
+    const tabIndex = previousTabs.findIndex((t) => t.id === payload.id);
+    if (tabIndex === -1) {
+      return;
+    }
+    // Copy instead of mutating the tab held by webTabsAtom. The old in-place
+    // write made every downstream equality check useless — by the time
+    // buildWebTabs compared "previous" against "next" they were the same
+    // object — and it left `changed` undetectable, so a payload identical to
+    // the current state still replaced webTabsMapAtom and scheduled a full
+    // tab-array persist. A DApp that puts a live price in document.title
+    // (Hyperliquid) drove that path once per second, per tab.
+    const previousTab = previousTabs[tabIndex];
+    const tabToModify: IWebTab = { ...previousTab };
+    let changed = false;
+    Object.keys(payload).forEach((k) => {
+      const key = k as keyof IWebTab;
+      const value = payload[key];
+      if (value !== undefined && value !== tabToModify[key]) {
+        if (key === 'title') {
+          if (!value) {
+            return;
           }
         }
-      });
-      tabs[tabIndex] = tabToModify;
-      loggerForEmptyData(tabs, 'setWebTabData');
-      this.buildWebTabs.call(set, { data: tabs });
+        // @ts-expect-error
+        tabToModify[key] = value;
+        changed = true;
+        if (key === 'url') {
+          // Navigation normally bumps timestamp to Date.now(), which
+          // re-sorts the tab to the bottom. Skip when the user chose
+          // 'top' so the tab stays where it was created. Record the
+          // navigation time separately for the onNavigation debounce.
+          if (!isNewTabPositionTop()) {
+            tabToModify.timestamp = Date.now();
+          }
+          if (payload.id) {
+            lastNavigationFlags[payload.id] = Date.now();
+          }
+          if (value === 'about:blank' && payload.id) {
+            homeResettingFlags[payload.id] = Date.now();
+          }
+        }
+      }
+    });
+    if (!changed) {
+      return;
     }
+    const tabs = [...previousTabs];
+    tabs[tabIndex] = tabToModify;
+    loggerForEmptyData(tabs, 'setWebTabData');
+    this.buildWebTabs.call(set, { data: tabs });
   });
 
   openUrlInHomeTab = contextAtomMethod(
@@ -1504,10 +1536,41 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
         }
       }
 
+      // Trading DApps stream the live price through document.title, so
+      // `page-title-updated` fires about once a second forever. Only the URL
+      // path above is debounced, so those title-only events used to reach
+      // setWebTabData at full rate and, through buildWebTabs, replace the tab
+      // map and persist the whole tab array every second. Drop title updates
+      // that arrive inside the throttle window unless something else in this
+      // event actually changed; the tab title is refreshed again on the next
+      // event that passes, and on navigation.
+      let nextTitle = title;
+      if (title !== undefined) {
+        const isTitleOnlyUpdate =
+          !isValidNewUrl &&
+          favicon === undefined &&
+          canGoBack === undefined &&
+          canGoForward === undefined &&
+          loading === undefined;
+        const lastTitleAt = lastTitleUpdateFlags[tab.id];
+        if (
+          isTitleOnlyUpdate &&
+          lastTitleAt &&
+          now - lastTitleAt < TITLE_UPDATE_THROTTLE_MS
+        ) {
+          return;
+        }
+        if (title !== tab.title) {
+          lastTitleUpdateFlags[tab.id] = now;
+        } else {
+          nextTitle = undefined;
+        }
+      }
+
       this.setWebTabData.call(set, {
         displayUrl: url,
         id: tab.id,
-        title,
+        title: nextTitle,
         favicon,
         canGoBack,
         canGoForward,
--- a/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
+++ b/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
@@ -22,6 +22,7 @@ import {
   BITREFILL_BRIDGE_SCRIPT,
   isBitrefillEmbedUrl,
 } from '../../utils/bitrefillUtils';
+import { applyCurrentThrottleStateToWebview } from '../../utils/browserOffRouteThrottle';
 import { webviewRefs } from '../../utils/explorerUtils';
 import BlockAccessView from '../BlockAccessView';
 
@@ -242,6 +243,9 @@ function WebContent({ id, url, customReceiveHandler }: IWebContentProps) {
                 });
               }
               webviewRefs[id] = ref;
+              // A WebView mounted while the user is away from the browser
+              // route would otherwise start out un-throttled.
+              applyCurrentThrottleStateToWebview(id);
             }
           }}
           allowpopups
--- a/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
+++ b/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
@@ -205,6 +205,11 @@ function WebContent({ id, url, customReceiveHandler }: IWebContentProps) {
       // @ts-expect-error
       ref.__domReady = true;
     }
+    // The ref-publish call site cannot cover every case: a tab frozen by
+    // react-freeze never runs the imperative handle, and a guest that reloads
+    // or navigates while off-route comes back with a fresh, unpatched document.
+    // dom-ready is the signal that is guaranteed for both.
+    applyCurrentThrottleStateToWebview(id);
     // Inject the Bitrefill bridge on every dom-ready so raw window.postMessage
     // events from embed.bitrefill.com are re-emitted as $private JSBridge
     // requests reaching useDiscoveryMessageHandler.
@@ -244,7 +249,11 @@ function WebContent({ id, url, customReceiveHandler }: IWebContentProps) {
               }
               webviewRefs[id] = ref;
               // A WebView mounted while the user is away from the browser
-              // route would otherwise start out un-throttled.
+              // route would otherwise start out un-throttled. This publish can
+              // land after the guest is already dom-ready, in which case the
+              // dom-ready callback has run with no ref to work with — so this
+              // is the only chance for that ordering. The reverse ordering is
+              // covered from onDomReady.
               applyCurrentThrottleStateToWebview(id);
             }
           }}
--- a/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
+++ b/packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
+import { memo, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import { Stack } from '@onekeyhq/components';
 import WebView from '@onekeyhq/kit/src/components/WebView';
@@ -317,4 +317,9 @@ function WebContent({ id, url, customReceiveHandler }: IWebContentProps) {
   );
 }
 
-export default WebContent;
+// The parent shell re-renders whenever the shared tab map changes. Without a
+// memo boundary here that rebuilt the whole webview subtree — including the
+// useMemo'd <WebView> element — on every tab-state write. All props are stable
+// (`customReceiveHandler` is a no-dependency useCallback), so this bails out
+// unless the tab's own url/id or its active state actually changes.
+export default memo(WebContent);
--- a/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
+++ b/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
@@ -23,12 +23,13 @@ export const useWebTabs = () => {
 
 export const useWebTabDataById = (id?: string) => {
   const [map] = useWebTabsMapAtom();
-  return useMemo(
-    () => ({
-      tab: map[id ?? ''] as IWebTab | undefined,
-    }),
-    [map, id],
-  );
+  const tab = map[id ?? ''] as IWebTab | undefined;
+  // Memoise on the tab, not on the map. buildWebTabs hands out a new map
+  // object whenever any tab changes, so keying the result on `map` produced a
+  // fresh `{ tab }` for every mounted tab shell on every write — with 26 open
+  // tabs one title change re-rendered all 26 subtrees. Tabs whose own entry is
+  // untouched now keep a stable reference and their children bail out.
+  return useMemo(() => ({ tab }), [tab]);
 };
 
 export const useActiveTabId = () => {
--- a/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
+++ b/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
@@ -39,7 +39,11 @@ export const useWebTabs = () => {
 const selectWebTabIds = (value: IWebTabsAtom) => value.tabs.map((t) => t.id);
 const areIdListsEqual = (a: string[], b: string[]) =>
   a === b || (a.length === b.length && a.every((id, i) => id === b[i]));
-const webTabIdsAtom = selectAtom(webTabsAtom(), selectWebTabIds, areIdListsEqual);
+const webTabIdsAtom = selectAtom(
+  webTabsAtom(),
+  selectWebTabIds,
+  areIdListsEqual,
+);
 
 export const useWebTabIds = () => {
   const { store } = useDiscoveryContextStoreData();
--- a/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
+++ b/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
@@ -1,41 +1,70 @@
-import { useMemo, useRef } from 'react';
+import { useMemo } from 'react';
+
+import { useAtomValue } from 'jotai';
+import { selectAtom } from 'jotai/utils';
 
 import {
   useActiveTabIdAtom,
   useAliveWebViewIdsAtom,
   useDisabledAddedNewTabAtom,
+  useDiscoveryContextStoreData,
   useDisplayHomePageAtom,
   useWebTabsAtom,
-  useWebTabsMapAtom,
+  webTabsAtom,
+  webTabsMapAtom,
 } from '@onekeyhq/kit/src/states/jotai/contexts/discovery';
 
-import type { IWebTab } from '../types';
+import type { IWebTab, IWebTabsAtom } from '../types';
 
+// Fresh snapshot of the full tab array. Always tracks the atom, so consumers
+// that read fields off the array (isPinned grouping in the tab bars, the
+// pinned-tab guard in useShortcuts.desktop, MobileTabListModal) can never see
+// stale values. Shell containers that only need to key child components by id
+// should use useWebTabIds instead, which stays referentially stable across
+// field-only writes.
 export const useWebTabs = () => {
   const [webTabs] = useWebTabsAtom();
-  // Key on the id list, not on the atom value. webTabsAtom is now written on
-  // any tab field change (title/loading/favicon) so later reads cannot go
-  // stale; keying on the whole value would hand every consumer a fresh `tabs`
-  // array on each of those writes. Consumers here only iterate ids to render
-  // per-tab shells — they read fields through useWebTabDataById — so the array
-  // only needs a new identity when the set or order of tabs changes.
-  const idsKey = webTabs.tabs.map((t) => t.id).join(',');
-  const cache = useRef({ idsKey, tabs: webTabs.tabs });
-  if (cache.current.idsKey !== idsKey) {
-    cache.current = { idsKey, tabs: webTabs.tabs };
-  }
-  const { tabs } = cache.current;
-  return useMemo(() => ({ tabs }), [tabs]);
+  return useMemo(
+    () => ({
+      tabs: webTabs.tabs,
+    }),
+    [webTabs],
+  );
 };
 
+// Module-level select/equality so the derived atom is created once (the same
+// discipline as useTokenFiatField). Emits a new value only when the id list
+// (set or order) changes — field-only writes keep the previous array identity,
+// so shell containers mapping ids to memoized children skip re-rendering.
+const selectWebTabIds = (value: IWebTabsAtom) => value.tabs.map((t) => t.id);
+const areIdListsEqual = (a: string[], b: string[]) =>
+  a === b || (a.length === b.length && a.every((id, i) => id === b[i]));
+const webTabIdsAtom = selectAtom(webTabsAtom(), selectWebTabIds, areIdListsEqual);
+
+export const useWebTabIds = () => {
+  const { store } = useDiscoveryContextStoreData();
+  const ids = useAtomValue(webTabIdsAtom, { store });
+  return useMemo(() => ({ ids }), [ids]);
+};
+
+// Per-tab subscription. Subscribing to the whole map atom meant every tab
+// shell and navigation bar re-rendered on every map replacement no matter
+// whose tab changed — memoizing the returned object cannot stop a re-render
+// that the component's own atom subscription triggers. selectAtom narrows the
+// subscription to this tab's entry: buildWebTabData reuses untouched tab
+// objects (setWebTabData is copy-on-write for the changed one only), so the
+// Object.is equality holds and only the changed tab's subscribers run.
 export const useWebTabDataById = (id?: string) => {
-  const [map] = useWebTabsMapAtom();
-  const tab = map[id ?? ''] as IWebTab | undefined;
-  // Memoize on the tab, not on the map. buildWebTabs hands out a new map
-  // object whenever any tab changes, so keying the result on `map` produced a
-  // fresh `{ tab }` for every mounted tab shell on every write — with 26 open
-  // tabs one title change re-rendered all 26 subtrees. Tabs whose own entry is
-  // untouched now keep a stable reference and their children bail out.
+  const { store } = useDiscoveryContextStoreData();
+  const tabAtom = useMemo(
+    () =>
+      selectAtom(
+        webTabsMapAtom(),
+        (map) => map[id ?? ''] as IWebTab | undefined,
+      ),
+    [id],
+  );
+  const tab = useAtomValue(tabAtom, { store });
   return useMemo(() => ({ tab }), [tab]);
 };
 
--- a/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
+++ b/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
@@ -24,7 +24,7 @@ export const useWebTabs = () => {
 export const useWebTabDataById = (id?: string) => {
   const [map] = useWebTabsMapAtom();
   const tab = map[id ?? ''] as IWebTab | undefined;
-  // Memoise on the tab, not on the map. buildWebTabs hands out a new map
+  // Memoize on the tab, not on the map. buildWebTabs hands out a new map
   // object whenever any tab changes, so keying the result on `map` produced a
   // fresh `{ tab }` for every mounted tab shell on every write — with 26 open
   // tabs one title change re-rendered all 26 subtrees. Tabs whose own entry is
--- a/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
+++ b/packages/kit/src/views/Discovery/hooks/useWebTabs.ts
@@ -1,4 +1,4 @@
-import { useMemo } from 'react';
+import { useMemo, useRef } from 'react';
 
 import {
   useActiveTabIdAtom,
@@ -13,12 +13,19 @@ import type { IWebTab } from '../types';
 
 export const useWebTabs = () => {
   const [webTabs] = useWebTabsAtom();
-  return useMemo(
-    () => ({
-      tabs: webTabs.tabs,
-    }),
-    [webTabs],
-  );
+  // Key on the id list, not on the atom value. webTabsAtom is now written on
+  // any tab field change (title/loading/favicon) so later reads cannot go
+  // stale; keying on the whole value would hand every consumer a fresh `tabs`
+  // array on each of those writes. Consumers here only iterate ids to render
+  // per-tab shells — they read fields through useWebTabDataById — so the array
+  // only needs a new identity when the set or order of tabs changes.
+  const idsKey = webTabs.tabs.map((t) => t.id).join(',');
+  const cache = useRef({ idsKey, tabs: webTabs.tabs });
+  if (cache.current.idsKey !== idsKey) {
+    cache.current = { idsKey, tabs: webTabs.tabs };
+  }
+  const { tabs } = cache.current;
+  return useMemo(() => ({ tabs }), [tabs]);
 };
 
 export const useWebTabDataById = (id?: string) => {
--- a/packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
+++ b/packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
@@ -33,16 +33,15 @@ function injectIntoAllWebviews(code: string, label: string): number {
   let applied = 0;
   for (const id of Object.keys(webviewRefs)) {
     const innerRef = webviewRefs[id]?.innerRef as IElectronWebView | undefined;
-    if (!innerRef) {
-      continue;
-    }
-    try {
-      innerRef.executeJavaScript(code);
-      applied += 1;
-    } catch {
-      // A WebView that is not dom-ready yet cannot be reached, and it does not
-      // need to be: it has nothing running to throttle, and the next
-      // transition re-applies the current state.
+    if (innerRef) {
+      try {
+        innerRef.executeJavaScript(code);
+        applied += 1;
+      } catch {
+        // A WebView that is not dom-ready yet cannot be reached, and it does
+        // not need to be: it has nothing running to throttle, and the next
+        // transition re-applies the current state.
+      }
     }
   }
   defaultLogger.discovery.browser.offRouteThrottle({ label, applied });
--- a/packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
+++ b/packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
@@ -3,10 +3,10 @@ import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
 
 import {
-  injectToPauseWebsocket,
+  injectToPauseWebsocketOffRoute,
   injectToReportPageHidden,
   injectToReportPageVisible,
-  injectToResumeWebsocket,
+  injectToResumeWebsocketOffRoute,
   webviewRefs,
 } from './explorerUtils';
 
@@ -29,34 +29,61 @@ import {
 
 let isThrottled = false;
 
+/**
+ * Inject into one WebView. Electron reports "not attached / not ready" both as a
+ * synchronous throw and as a rejected promise depending on how far the guest
+ * got, so both have to be absorbed — an unhandled rejection here would surface
+ * as a console error on a path that is expected to miss sometimes.
+ */
+function injectIntoWebview(id: string, code: string): boolean {
+  const innerRef = webviewRefs[id]?.innerRef as IElectronWebView | undefined;
+  if (!innerRef) {
+    return false;
+  }
+  try {
+    const result = innerRef.executeJavaScript(code) as unknown;
+    if (result && typeof (result as Promise<unknown>).then === 'function') {
+      (result as Promise<unknown>).catch(() => {
+        // guest went away or was not ready; the next transition re-applies
+      });
+    }
+    return true;
+  } catch {
+    // A WebView that is not dom-ready yet cannot be reached, and it does not
+    // need to be: it has nothing running to throttle, and both the next
+    // transition and its own dom-ready re-apply the current state.
+    return false;
+  }
+}
+
 function injectIntoAllWebviews(code: string, label: string): number {
+  const ids = Object.keys(webviewRefs);
   let applied = 0;
-  for (const id of Object.keys(webviewRefs)) {
-    const innerRef = webviewRefs[id]?.innerRef as IElectronWebView | undefined;
-    if (innerRef) {
-      try {
-        innerRef.executeJavaScript(code);
-        applied += 1;
-      } catch {
-        // A WebView that is not dom-ready yet cannot be reached, and it does
-        // not need to be: it has nothing running to throttle, and the next
-        // transition re-applies the current state.
-      }
+  for (const id of ids) {
+    if (injectIntoWebview(id, code)) {
+      applied += 1;
     }
   }
-  defaultLogger.discovery.browser.offRouteThrottle({ label, applied });
+  // `applied` counts injections issued, not landed, and `total` counts entries
+  // in webviewRefs — neither is a count of guest renderer processes, which also
+  // include out-of-process iframes and popups that this pass cannot reach.
+  defaultLogger.discovery.browser.offRouteThrottle({
+    label,
+    applied,
+    total: ids.length,
+  });
   return applied;
 }
+
+const PAUSE_SCRIPT = `${injectToPauseWebsocketOffRoute};${injectToReportPageHidden}`;
+const RESUME_SCRIPT = `${injectToResumeWebsocketOffRoute};${injectToReportPageVisible}`;
 
 export function throttleBrowserWebviewsOffRoute() {
   if (!platformEnv.isDesktop || isThrottled) {
     return;
   }
   isThrottled = true;
-  injectIntoAllWebviews(
-    `${injectToPauseWebsocket};${injectToReportPageHidden}`,
-    'pause',
-  );
+  injectIntoAllWebviews(PAUSE_SCRIPT, 'pause');
 }
 
 export function restoreBrowserWebviewsOnRoute() {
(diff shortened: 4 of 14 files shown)


packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -429,6 +435,18 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
         set(webTabsAtom(), { keys: result.keys, tabs: result.data });
       }
 
+      // `keys` only covers the tab id list, so a change to a tab's own fields
+      // (title/url/loading/favicon) never showed up here. The map replacement
+      // and the persist below were therefore unconditional: every no-op write
+      // handed all mounted tab shells a fresh map object and re-serialised the
+      // whole tab array to SimpleDb. Compare the tab payloads too, and skip
+      // both when nothing actually moved.
+      const tabsUnchanged =
```
originalix: P1: [字段更新不会同步 webTabsAtom]

当 `setWebTabData` 只修改 `title`、`loading`、`favicon` 或 `isPinned` 时，前面已经改为创建新的 tab/array，但这里仍只在 keys 变化或 `forceUpdate` 时写 `webTabsAtom`。结果是 `webTabsMapAtom` 和持久化拿到新快照，而后续 action 继续从旧 `webTabsAtom` 读取；连续更新会覆盖前一次字段，`setPinnedTab`/`setSiteMode` 后面的 `setTabs` 甚至可能立即回滚自身修改。

请在 `result.data` 与 `webTabs.tabs` 不同时同步更新 `webTabsAtom`，仅当 keys 和 tab payload 都未变化时才提前返回。

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -1504,10 +1536,41 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
...
+          canGoBack === undefined &&
+          canGoForward === undefined &&
+          loading === undefined;
+        const lastTitleAt = lastTitleUpdateFlags[tab.id];
+        if (
+          isTitleOnlyUpdate &&
+          lastTitleAt &&
+          now - lastTitleAt < TITLE_UPDATE_THROTTLE_MS
+        ) {
+          return;
```
originalix, line 1626: P2: [标题节流会永久丢掉窗口内的最终值]

页面在一次加载中先设置临时标题、再于 5 秒内设置最终标题时，第二个 `page-title-updated` 是 title-only 事件，会在这里被直接丢弃。如果页面之后不再修改 `document.title`，就没有下一次机会同步，标签栏和持久化标题会一直停留在临时值。

请改成按 tab 保留窗口内的最新标题，并在节流窗口结束时执行一次 trailing update；标签关闭时同时清理对应待执行任务。

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -429,6 +435,18 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
...
 
+      // `keys` only covers the tab id list, so a change to a tab's own fields
+      // (title/url/loading/favicon) never showed up here. The map replacement
+      // and the persist below were therefore unconditional: every no-op write
+      // handed all mounted tab shells a fresh map object and re-serialised the
+      // whole tab array to SimpleDb. Compare the tab payloads too, and skip
+      // both when nothing actually moved.
+      const tabsUnchanged =
+        !options?.forceUpdate && isEqual(result.data, webTabs.tabs);
+      if (tabsUnchanged) {
```
sidmorizon: # [🟠 P1] closeWebTab 原地 splice 污染 isEqual 基线,`tabsUnchanged` 误判并跳过关闭 tab 的立即持久化

## 问题

与上面 line 444 已有评论(keys 不变时 `webTabsAtom` 不回写,后续 action 读旧数组互相覆盖;CI Unit Tests 2/3 中 actions.test.tsx:991/1047 的 `displayUrl`/`loading` 丢失即该方向)是同一块代码的另一个独立方向,且「`result.data` 与 `webTabs.tabs` 不同时同步更新 `webTabsAtom`」的修复无法覆盖:`closeWebTab` 在调用 `buildWebTabs` 前仍通过 `tabs.splice(targetIndex, 1)` **原地修改** `webTabsAtom` 持有的数组,所以这里读到的 `webTabs.tabs` 基线已经是删除后的结果,`isEqual(result.data, webTabs.tabs)` 恒为 `true`,`tabsUnchanged` 提前 `return`,一次性跳过了三件事:

```ts
set(webTabsMapAtom(), () => result.map);        // map 残留已关闭 tab
persistTabsToSimpleDbDebounced.cancel();        // stale trailing write 不再被取消
void persistTabsToSimpleDb(result.data);        // Immediate 持久化被跳过
```

CI 已在本 head(0610b5ac)复现:Unit Tests 1/3,`closeTabPersist.test.tsx:222` —— 快速连关 3 个 tab 后,立即持久化快照仍是 `['1','2','3']` 而非 `['1']`。

## 影响

关闭 tab 后立即硬退出,重启会复活已关闭的 tab;pending 的 debounce trailing write 也不再被 cancel,可能用更大的 stale 快照覆盖最终状态——这正是该测试守护的回归场景。同时 `webTabsMapAtom` 残留已关闭 tab 的 entry,`getWebTabById` / `useWebTabDataById` 会返回幽灵 tab。

## 建议

两个方向一起堵:bail-out 不得吞掉 `Immediate` 持久化,且数据变化时两个 atom 必须同步写:

```ts
const tabsUnchanged =
  !options?.forceUpdate &&
  options?.persist !== EBrowserTabPersistMode.Immediate &&
  isEqual(result.data, webTabs.tabs);
if (tabsUnchanged) {
  return;
}
set(webTabsAtom(), { keys: result.keys, tabs: result.data });
set(webTabsMapAtom(), () => result.map);
```

根治方案是把 `closeWebTab` 的 `tabs.splice(...)`(以及 `activateAdjacentTab` 里 `newActiveTab.isActive = true` 这类原地写)改为 copy-on-write,否则任何以 `webTabs.tabs` 为基线的相等性判断都可能拿到被污染的「上一次状态」。

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

packages/kit/src/states/jotai/contexts/discovery/actions.ts
originalix: Confirmed and fixed in 6e274fe4a7 — this was a real regression I introduced by switching `setWebTabData` to copy-on-write without updating the write path.

`buildWebTabs` now compares tab payloads as well as the id list, and writes `webTabsAtom` whenever either changed:

```ts
const keysChanged = !isEqual(result.keys, webTabs.keys);
const tabsChanged = !isEqual(result.data, webTabs.tabs);
if (!keysChanged && !tabsChanged && !options?.forceUpdate && !isImmediatePersist) {
  return;
}
set(webTabsAtom(), { keys: result.keys, tabs: result.data });
set(webTabsMapAtom(), () => result.map);
```

Both atoms now advance together, so a field-only update can no longer leave `webTabsAtom` behind while the map and the persisted snapshot move on.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
originalix: Good catch — fixed in 6e274fe4a7, and I took the root-cause route you suggested rather than only guarding the bail-out.

1. `closeWebTab` no longer splices the array held by `webTabsAtom`:

```ts
const { tabs: previousTabs } = get(webTabsAtom());
let tabs = previousTabs;
...
tabs = previousTabs.filter((_, index) => index !== targetIndex);
```

2. `activateAdjacentTab` replaces the entry instead of flipping `isActive` on the shared object:

```ts
tabs[newActiveTabIndex] = { ...newActiveTab, isActive: true };
```

3. The bail-out can no longer swallow an `Immediate` persist, exactly as you proposed:

```ts
const isImmediatePersist = options?.persist === EBrowserTabPersistMode.Immediate;
if (!keysChanged && !tabsChanged && !options?.forceUpdate && !isImmediatePersist) {
  return;
}
```

`closeTabPersist.test.tsx` and the rest of the Discovery suites pass locally (7 suites, 110 tests).

packages/kit/src/states/jotai/contexts/discovery/actions.ts
originalix: Agreed — fixed in 6e274fe4a7 with a trailing update, as suggested.

A title suppressed inside the throttle window is now held per tab and flushed when the window closes, so the last value in a burst is never lost:

```ts
pendingTitleUpdates[tabId] = {
  title,
  timer: setTimeout(() => { ... }, TITLE_UPDATE_THROTTLE_MS - (now - lastTitleAt)),
};
```

The held value is dropped when a title goes through normally (it would be stale), and `closeWebTab` calls `clearPendingTitleUpdate(tabId)` so a pending flush cannot resurrect an entry for a tab that no longer exists.

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
```diff
@@ -13,22 +13,30 @@ import type { IWebTab } from '../types';
...
+  // stale; keying on the whole value would hand every consumer a fresh `tabs`
+  // array on each of those writes. Consumers here only iterate ids to render
+  // per-tab shells — they read fields through useWebTabDataById — so the array
+  // only needs a new identity when the set or order of tabs changes.
+  const idsKey = webTabs.tabs.map((t) => t.id).join(',');
+  const cache = useRef({ idsKey, tabs: webTabs.tabs });
+  if (cache.current.idsKey !== idsKey) {
+    cache.current = { idsKey, tabs: webTabs.tabs };
+  }
+  const { tabs } = cache.current;
```
sidmorizon: # [🟠 P1] useWebTabs 的 idsKey 缓存让消费方读到过期的 `isPinned`,pin/unpin 在 id 顺序不变时静默失效

## 问题

新的缓存只在 id 列表(含顺序)变化时更新 `cache.current.tabs`,注释假设「Consumers here only iterate ids」,但至少 4 个现存消费方直接从这个数组读取 `isPinned` 字段:

- `useShortcuts.desktop.ts:48` — `tabs[tabIndex].isPinned` 决定 Cmd+W 是关闭 tab 还是跳回 Discovery(pinned 保护)
- `DesktopCustomTabBar/index.tsx:149-150` — `allTabs.filter((t) => t.isPinned)` 分组渲染侧栏 pinned/unpinned 两个区
- `MobileTabListModal/index.tsx:122-125` — 同样的分组
- `Browser.native.tsx:414` — `tab?.isPinned` 决定搜索是否复用当前窗口

而 `isPinned` 翻转并不保证 id 顺序变化:`setPinnedTab` 的 pin 路径写 `timestamp = Date.now()`,默认 bottom 模式下当前活跃 tab 因导航 bump 通常已持有最大 timestamp,排序后仍在原位 → `idsKey` 不变 → 缓存数组里的对象仍是 `isPinned: false` 的旧引用(`setWebTabData` 是 copy-on-write,旧数组不会被更新)。unpin 路径(`timestamp = minUnpinned - GAP`)对已在首位的 pinned tab 同理。

## 影响

Pin 当前 tab(最常见操作)后:侧栏不把它移入 pinned 区、分隔线不出现;紧接着按 Cmd+W 会直接**关闭刚 pin 的 tab**,pinned 保护失效;native 端搜索的 `useCurrentWindow` 判断同样基于过期值。且 `DesktopCustomTabBar` 是全程挂载的长生命周期组件,过期状态会一直保持到下一次 id 顺序真正变化,不会自愈。

## 建议

最小修复:把消费方实际读取的字段并入缓存 key,`isPinned` 变化时刷新引用:

```ts
const idsKey = webTabs.tabs
  .map((t) => `${t.id}:${t.isPinned ? 1 : 0}`)
  .join(',');
```

更干净的做法是拆分语义:为只渲染 shell 的场景(nav container / Browser.desktop)提供 `useWebTabIds()`(仅 id 列表,稳定引用),让 `useWebTabs()` 恢复返回新鲜数据——避免未来任何消费方从这个数组读字段时再次踩中同样的静默过期。

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
```diff
@@ -13,22 +13,30 @@ import type { IWebTab } from '../types';
...
-    [webTabs],
-  );
+  // Key on the id list, not on the atom value. webTabsAtom is now written on
+  // any tab field change (title/loading/favicon) so later reads cannot go
+  // stale; keying on the whole value would hand every consumer a fresh `tabs`
+  // array on each of those writes. Consumers here only iterate ids to render
+  // per-tab shells — they read fields through useWebTabDataById — so the array
+  // only needs a new identity when the set or order of tabs changes.
+  const idsKey = webTabs.tabs.map((t) => t.id).join(',');
+  const cache = useRef({ idsKey, tabs: webTabs.tabs });
```
originalix: P1: [按 tab id 缓存整组对象会返回陈旧状态]

当 tab 的 `title`、`favicon`、`loading` 或 `isPinned` 发生字段更新而 id 集合和顺序不变时，这个 ref 不会替换，`useWebTabs()` 会继续向消费者返回旧 `IWebTab[]`。`DesktopCustomTabBar` 与 `MobileTabListModal` 直接读取这些字段，`useShortcuts.desktop.ts` 还用其中的 `isPinned` 决定是否允许关闭，因此 UI 可长期不刷新，甚至快捷键会关闭刚固定的标签。

请让 `useWebTabs()` 保持返回当前 atom snapshot，并为只需要遍历 shell 的容器新增仅选择 tab id 的 hook。

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
```diff
@@ -13,22 +13,30 @@ import type { IWebTab } from '../types';
...
+  if (cache.current.idsKey !== idsKey) {
+    cache.current = { idsKey, tabs: webTabs.tabs };
+  }
+  const { tabs } = cache.current;
+  return useMemo(() => ({ tabs }), [tabs]);
 };
 
 export const useWebTabDataById = (id?: string) => {
   const [map] = useWebTabsMapAtom();
```
originalix: P2: [useMemo 没有切断整张 map 的 Jotai 订阅]

当任一 tab 更新并替换 `webTabsMapAtom` 时，`useWebTabsMapAtom()` 仍会通知每个 `useWebTabDataById` 订阅者；下面只 memo 返回对象不能阻止组件因自身 atom subscription 而重新 render，新增的 `memo(...)` 也只能拦截父级 props。因此多标签下每次 title/loading 写入仍会执行所有 tab shell/navigation bar，核心 render storm 没有在这里被隔离。

请按项目已有模式为每个 id 缓存 `selectAtom(webTabsMapAtom(), map => map[id])` 的 context atom，让组件直接订阅单个 tab entry。

originalix:
## Verification results (bundle builds from this branch)

**Protocol** — same machine and tab set as the pre-fix baseline: ~26 browser tabs including several trading DApp pages (live price streamed through `document.title`), Perps and Market secondary pages open, a burst of ~30 rapid tab switches, then idle. All numbers come from the app's own log (`[WebView Memory] type=window`, `[CPU Watchdog]`, `browserTabs.setRawData` call lines), i.e. the same source as the original report.

Two bundles were tested: **v1** = branch through `914f29a7d7` (first four commits), **v2** = branch through `047327a821`.

| Metric | Pre-fix | v1 | v2 |
|---|---|---|---|
| Idle persist rate (`setRawData`) | 40–55/min, climbing monotonically | **5.1/min** (13 live tabs) | **13.7/min** (18 live tabs) |
| Live webviews vs `MAX_ALIVE_WEBVIEW_COUNT=8` | 10–17 — over the limit in 91% of samples | — | **exactly 8 under stress** |
| Window renderer memory | 475 → 1043 MB in 2 h, monotonic, never recovers | ~800 MB plateau | ~690 MB idle; switch burst spiked to 1339 MB, **recovered to ~1095 MB**, then ~0.5 MB/min drift |
| CPU watchdog | 0 fires (this machine peaks ~50%, below the 80% tier) | **1× `sustained-high-cpu-mild`** — 30 consecutive samples at 80–92% of one core in the window renderer | **0 fires in 91 min**, including the switching burst (transient peak 78.8%, steady 53–62%) |

**The v1 watchdog fire was a regression introduced by the review follow-up** (`914f29a7d7`): syncing `webTabsAtom` on field-only changes is required for correctness, but it made every title tick re-render all mounted nav bars (no memo boundary, and `useWebTabs` keyed its result on the atom value). `047327a821` keys `useWebTabs` on the tab id list and adds a memo boundary around the per-tab nav bar; v2 confirms the fire is gone under the same stress.

**Known residual gaps (out of scope here, follow-up planned):**
- ~0.5 MB/min slow drift in the window renderer with trading tabs parked — an order of magnitude shallower than the pre-fix ~4.7 MB/min, but not zero. Needs a heap-snapshot diff on a debug-port build to chase.
- The title throttle is per-tab (5 s), so the aggregate persist rate scales with the number of live tabs. Collapsing it further would need coalescing at the persistence layer.

Memory recovery after the burst (1339 → 1095 MB) and the stable webview count are the behavioral proof that the retained-tree / no-op-eviction defects are fixed: pre-fix the same workload produced monotonic growth with no recovery and 10–17 live webviews.

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
originalix: Confirmed — the idsKey cache was wrong, and your failure analysis was exact: pin on the active tab does not reorder ids (it already holds the max timestamp), so the cache kept serving `isPinned: false` and the Cmd+W pinned-tab guard read the stale value.

Fixed in 8d56b10edb by splitting the semantics as you suggested rather than widening the cache key:

- `useWebTabs()` returns the live snapshot again — every field reader (DesktopCustomTabBar, MobileTabListModal, useShortcuts.desktop, Browser.native) sees fresh `isPinned`/`title`/`loading`
- new `useWebTabIds()` is the ids-only stable view (module-level `selectAtom` with id-list equality), now used by the nav-bar container and Browser.desktop — the only two callers that genuinely just map ids to memoized shells

Widening the key to include `isPinned` would have fixed today's four consumers but left the next field reader to trip over the same silent staleness; the split removes the trap class.

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
originalix: Both points fixed in 8d56b10edb, following the pattern you named:

1. **Stale snapshot (P1)** — `useWebTabs()` returns the live atom snapshot again, and shell containers moved to a new ids-only `useWebTabIds()` (module-level `selectAtom` over `webTabsAtom` with an id-list equality), so field readers are always fresh and id-mapping containers stay referentially stable.

2. **Subscription not severed (P2)** — you were right that memoizing the returned object could not stop re-renders triggered by the component's own map-atom subscription. `useWebTabDataById` now subscribes to its own entry via `selectAtom(webTabsMapAtom(), map => map[id])` with the context store (same discipline as `useTokenFiatField`). This works because `buildWebTabData` reuses untouched tab objects — `setWebTabData` copies only the changed tab — so `Object.is` holds for every unaffected entry and only the changed tab's subscribers execute. A title tick on one tab now re-renders exactly one shell + one nav bar instead of all of them.

The discovery context now exposes its store accessor (`useDiscoveryContextStoreData`) for these derived-atom reads; regular consumers keep using the generated hooks.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -1504,10 +1574,65 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
...
+      // event that passes, and on navigation.
+      let nextTitle = title;
+      if (title !== undefined) {
+        const isTitleOnlyUpdate =
+          !isValidNewUrl &&
+          favicon === undefined &&
+          canGoBack === undefined &&
+          canGoForward === undefined &&
+          loading === undefined;
+        const lastTitleAt = lastTitleUpdateFlags[tab.id];
```
originalix, line 1598: P2: [标题节流状态会越过导航与批量关闭边界]

当 Desktop DApp 在 5 秒窗口内留下 trailing title 后，如果 tab 随即导航到新 URL，或通过 `closeAllWebTabs` 后恢复同一 id，`pendingTitleUpdates` 与 `lastTitleUpdateFlags` 仍属于旧 WebView。当前真实 navigation 只有在 incoming title 与当前 title 不同时才清 pending，而 bulk close 没有清理这些 entry，因此旧 timer 可能在新页面或恢复后的 tab 上调用 `setWebTabData`，覆盖并持久化旧标题；新 WebView 也会继承旧 cooldown。

请增加一个按 tab lifecycle 调用的 reset helper，同时 cancel/delete pending timer 与 last flag，并在真实 URL navigation 以及 `closeWebTab`/`closeAllWebTabs` 的每个 closed id 上执行。

originalix, line 1598: P2: [标题节流状态未绑定页面与 tab 生命周期]

当会持续更新 `document.title` 的 DApp 在 5 秒窗口内留下 trailing update 后，如果同一 tab 导航到新 URL，或通过 `closeAllWebTabs` 后按原 id 恢复，`pendingTitleUpdates` 与 `lastTitleUpdateFlags` 仍属于旧页面。当前逻辑只在 incoming title 与旧 `tab.title` 不同时清 timer，bulk close 也没有清理这些 entry，因此旧 callback 可能在新页面上调用 `setWebTabData`，覆盖并通过 `bg` SimpleDB 持久化旧标题；新 WebView 还会继承旧 cooldown。

请抽出一个同时 cancel/delete pending timer 与删除 `lastTitleUpdateFlags` 的 reset helper，并在 URL identity 改变以及 `closeWebTab`/`closeAllWebTabs` 的每个 closed id 上调用。

originalix:
## Overnight verification (final bundle: per-tab subscriptions + desktop keep-alive cap 16)

7.7 h unattended run, same tab set (~26 browser tabs incl. trading DApp pages, Perps and Market open), no user interaction after 00:38. All figures from the app's own log.

**Important measurement note for anyone repeating this:** the window renderer is recreated when the user navigates/reloads the shell, so `[WebView Memory] type=window` spans several pids over an evening. Memory slope must be computed **per pid** — averaging across pid boundaries flattens the resets and understates growth. The numbers below are the single pid that covers the whole idle window.

| Metric | Pre-fix | Final bundle (7.7 h idle) |
|---|---|---|
| CPU watchdog | 0 fires (this machine's ceiling is ~50%, under the 80% tier) | **0 fires** |
| Window renderer CPU, mean | — (v1 regression: 80–92% sustained → 1 fire) | **31.3% of one core** |
| Idle persist rate | 40–55/min, climbing | **p50 10/min, p90 16/min** |
| Live webviews | 10–17, over the cap in 91% of samples | **16, exactly the new desktop cap** |
| Window renderer memory | 475 → 1043 MB in 2 h, monotonic, no recovery | **749 → 1255 MB in 7.7 h**, peak 1422 MB, with partial recovery (1355 → 1273) |
| Memory slope | ~4.7 MB/min | **1.09 MB/min** |

### What this confirms
All three original defects and the regression introduced by the first review round are fixed: eviction genuinely unmounts and holds the cap, the render storm no longer pegs a core, and the persist rate is down 4–5×.

### What remains
Memory growth is reduced to roughly a quarter of the original slope but is **not bounded** — 1.09 MB/min extrapolates to ~2.3 GB over a day of continuous uptime. Heavy long-uptime desktop users will still accumulate memory, though without the pegged core that made it acute. Tracking separately; the next step is a heap-snapshot diff on a debug-port build across a full idle night, which is the only way to name what still retains.

### Behaviour change worth flagging in release notes
The desktop keep-alive cap is now 16 (was 8, and before the eviction fix it was effectively unbounded). Cold tabs beyond the cap reload when reactivated — that is the intended LRU behaviour, previously unreachable. The cap was raised from 8 after tab switching with ~26 tabs open produced a reload on nearly every switch; 16 covers the working set measured on such sessions.

originalix:
Residual memory growth tracked separately in #12686 (includes the reusable snapshot-diff tooling notes and the per-pid measurement trap).

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
```diff
@@ -0,0 +1,97 @@
...
+    'pause',
+  );
+}
+
+export function restoreBrowserWebviewsOnRoute() {
+  if (!platformEnv.isDesktop || !isThrottled) {
+    return;
+  }
+  isThrottled = false;
+  injectIntoAllWebviews(
```
sidmorizon: # [🟡 P2] 全量 resume 撤销了 pauseDappInteraction 对后台 tab 的定向暂停,恢复后策略状态不一致

## 问题

回到浏览器路由时,这里把 `injectToResumeWebsocket` 注入**所有** live webview。但既有的 per-tab 策略(`setCurrentWebTab` → `pauseDappInteraction`,actions.ts:552)在每次切 tab 时对切出的 tab 同时做了两件事:`WebSocket.prototype.send = noop` **和** `jsBridge.globalOnMessageEnabled = false`,只有活跃 tab 被 resume。全量 resume 只恢复了 WS 一半:

- 曾被暂停的后台 tab 的 `WebSocket.send` 被全部恢复,直到下一次切 tab 才会重新暂停其中一个;
- 这些 tab 的 `jsBridge.globalOnMessageEnabled` 仍为 `false` —— 同一个 tab 处于「bridge 暂停 + WS 恢复」这种此前不可达的混合状态。

## 影响

任何一次 ≥`20s` 的离开-返回往返后,所有后台 tab 的出站 WS 流量与其驱动的页面活动被重新激活,后台 tab 的 CPU/网络开销**高于本 feature 之前的浏览器内稳态**——对一个以降 CPU 为目的的 commit 是自我抵消的;交易类 DApp 多 tab 场景(PR 描述中的 26 tab 压测形态)下被放大。

## 建议

restore 时可见性注入保持全量,但 WS resume 只注入活跃 tab,与既有 per-tab 策略对齐:

```ts
export function restoreBrowserWebviewsOnRoute(activeTabId?: string | null) {
  if (!platformEnv.isDesktop || !isThrottled) {
    return;
  }
  isThrottled = false;
  injectIntoAllWebviews(injectToReportPageVisible, 'resume-visibility');
  const innerRef = activeTabId
    ? (webviewRefs[activeTabId]?.innerRef as IElectronWebView | undefined)
    : undefined;
  try {
    innerRef?.executeJavaScript(injectToResumeWebsocket);
  } catch {
    // not dom-ready
  }
}
```

调用方 `BrowserOffRouteThrottleContainer` 可从 discovery context 或 `webviewRefs` 侧拿到当前活跃 tab id;或者保持全量 resume,但随后对非活跃 tab 重放 `pauseDappInteraction`。

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
```diff
@@ -242,6 +243,9 @@ function WebContent({ id, url, customReceiveHandler }: IWebContentProps) {
                 });
               }
               webviewRefs[id] = ref;
+              // A WebView mounted while the user is away from the browser
+              // route would otherwise start out un-throttled.
+              applyCurrentThrottleStateToWebview(id);
```
sidmorizon, line 257: # [🟡 P2] applyCurrentThrottleStateToWebview 在 dom-ready 前调用,executeJavaScript 必然抛出,守卫实际不生效

## 问题

`onWebViewRef` 在 `<webview>` 元素 attach 时就会触发:`DesktopWebView` 的 `useImperativeHandle` 以 `isWebviewReady`(元素 ref 挂上即为 true)为依赖重建 wrapper,且 `initWebviewByRef` 在同一时刻显式 `updateIsDomReady(false)`——此时 guest 页面尚未 dom-ready。Electron 的 `webview.executeJavaScript` 在 dom-ready 之前调用会直接抛错(仓库里同模式的 catch 注释也写着 `// if not dom ready...`),所以 `applyCurrentThrottleStateToWebview` 内的注入在这个调用点**总是抛出并被吞掉**,函数注释所述的目标场景(off-route 期间挂载的 webview 需要补上 throttle 状态)从未真正生效——该 webview 会保持全速运行,直到下一次路由往返。

## 影响

这段守卫是事实上的 dead code:离开浏览器路由期间被程序化创建/挂载的 webview 不会被 throttle,且没有任何日志暴露失败(`try/catch` 静默)。虽然该场景较窄,但代码与注释声称的行为不符,后续维护者会误以为已覆盖。

## 建议

把补状态的调用移到 dom-ready 时机——`WebContent.desktop` 已有 `onDomReady` 回调,在其中调用即可(`onWebViewRef` 处的调用可以移除):

```ts
const onDomReady = useCallback(() => {
  const ref = webviewRefs[id];
  if (ref) {
    // @ts-expect-error
    ref.__domReady = true;
  }
  // Now executeJavaScript is legal; bring the new webview in line with the
  // current off-route throttle state.
  applyCurrentThrottleStateToWebview(id);
  ...
}, [id]);
```

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
```diff
@@ -0,0 +1,96 @@
...
+    `${injectToResumeWebsocket};${injectToReportPageVisible}`,
```
originalix: P1: [路由恢复会唤醒所有非活动 DApp WebView]

当存在多个 DApp tab，且用户离开 MultiTabBrowser 超过 20 秒再返回时，非活动 tab 此前已被 `setCurrentWebTab`/`pauseDappInteraction` 暂停，但这里会对 `webviewRefs` 中每个 guest 注入 `injectToResumeWebsocket` 和 visible/focus。该全量恢复会覆盖 per-tab pause 状态，使所有存活 guest 重新发送 WebSocket 并按前台页面运行，直接重新制造本 PR 要消除的后台 CPU/render storm。

请把 route focus 合并进既有 per-tab activity policy：返回时仅对当前 active tab 执行 `resume/visible`，其余 WebView 继续保持 `pause/hidden`。

packages/kit/src/states/jotai/contexts/discovery/actions.ts
```diff
@@ -1504,10 +1574,65 @@ class ContextJotaiActionsDiscovery extends ContextJotaiActionsBase {
...
+        const withinThrottleWindow =
+          Boolean(lastTitleAt) && now - lastTitleAt < TITLE_UPDATE_THROTTLE_MS;
+        if (isTitleOnlyUpdate && withinThrottleWindow) {
+          // Hold the newest title and flush it when the window closes, so the
+          // last value inside a burst is never lost.
+          const tabId = tab.id;
+          clearPendingTitleUpdate(tabId);
+          if (title && title !== tab.title) {
+            pendingTitleUpdates[tabId] = {
```
originalix, line 1607: P2: [标题节流状态会跨页面与批量关闭泄漏]

当动态 `document.title` 在 5 秒窗口内留下 trailing update 后，如果同一 tab 导航到新 URL，或经 `closeAllWebTabs` 后按原 id 恢复，这里的 pending entry 仍属于旧页面。当前 URL 更新不会 reset 它，bulk close 也没有清理这些 entry，因此旧 callback 可能覆盖新页面标题并通过 `bg` SimpleDB 持久化，新 WebView 还会继承旧 cooldown。

请抽出一个同时 cancel/delete pending timer 与删除 `lastTitleUpdateFlags[tabId]` 的 reset helper，并在 URL identity 改变以及单个/批量关闭的每个 id 上调用。

packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
originalix, line 257: P2: [离路由新建 WebView 的节流发生在 DOM ready 之前]

当应用已离开 MultiTabBrowser 且当前状态为 throttled 时，wrapper ref 会先于 guest `dom-ready` 到达这里。此时直接调用一次 `executeJavaScript` 可能尚不可执行；该调用没有进入 `DesktopWebView` 的 pending-script queue，异步 rejection 也不会被同步 `try/catch` 捕获，而 `onDomReady` 不会重试，因此 guest 可能在整个离路由阶段保持 visible 并继续运行页面脚本。

请在 `onDomReady` 中重新调用 `applyCurrentThrottleStateToWebview(id)` 并处理异步注入结果，确保 guest ready 后立即应用当前 throttle state。

originalix:
## Off-route WebView throttle: measured effect

Rebased onto `x` (base changed from `release/v6.5.0`); the 13 touched files are byte-identical to what was reviewed, except `provider/Container/index.tsx`, where the two branches differed independently of this PR. Against `x` this change adds 0 new unit-test failures and 0 new type errors.

### Measurement

The throttle is a discrete, reversible event with a timestamp in the log, so the honest way to measure it is the **release** transition on a single session: same processes, no reload, no remount. The engaged→released step is also the safer direction to read — an already-running page can only speed up when it is told it is visible again, so there is no "freshly loaded page settling down" confound.

Session had the desktop keep-alive cap saturated (16 live guest renderer processes, 11 reachable WebViews). CPU is the CPU watchdog's `effectivePct` (a `cumulativeCPUUsage` delta over wall time, 100 = one core busy). `ps %cpu` is unusable here — on macOS it is a process-lifetime average and smears pre-idle activity into the idle window.

| Guest WebView CPU (window renderer excluded) | 20 min before release | 20 min after release |
|---|---|---|
| Mean, all guest renderer processes | **18.1%** of one core | **57.7%** of one core |
| Guest renderer processes | 16 | 16 |
| pid set overlap across the boundary | — | **16 / 16** |

The step completes within ~2 minutes of the injection and holds. So while engaged, the throttle keeps DApp guest processes at roughly **31% of their unthrottled cost** — about **40 points of one core** saved, from 11 throttled WebViews.

### Engagement behaviour over a full session

Four throttle events in ~12 h, which is the intended shape — this must not churn:

- 2 no-op events (`applied: 0`) during a route bounce before any WebView was live
- 1 engage (`applied: 11`), held for ~12 h off-route
- 1 release (`applied: 11`) on returning to the browser route

No repeated toggling, and the 20 s grace period absorbed the bounce as designed.

### Gap found while measuring

`applied: 11` while **16** guest renderer processes were live. Injection only reaches WebViews whose `innerRef` is available at transition time; a WebView that exists but is not yet dom-ready is skipped. `applyCurrentThrottleStateToWebview` covers WebViews created *after* engaging, but not this case. Not a correctness problem — the next transition re-applies the current state — but it leaves some savings on the table. Worth a follow-up that retries on `dom-ready`.

### What these numbers are not

Night-over-night totals from two different unattended runs are **not** a valid A/B for this change, and should not be read as one. Screen state dominates: in one run a market-data stream was left active and the window renderer sat at 56–59% of one core all night; in the other the same renderer sat at 16–18%. That gap is which screen was left open, not this diff. Only the same-session transition above isolates the throttle.

That window-renderer observation is worth its own investigation and is out of scope here.

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
originalix: Confirmed and fixed in b5d4633fd2 — the mechanism is real. `injectToResumeWebsocket` restores `WebSocket.prototype.send` from the shared `$$onekeyWebSocketSend` slot, and since `setCurrentWebTab` pauses each tab as the user switches away from it, background tabs accumulate that paused state. A blanket resume therefore woke tabs the per-tab policy had deliberately paused, while their `jsBridge.globalOnMessageEnabled` stayed `false`.

I took a different route than "resume WS only on the active tab", because `BrowserOffRouteThrottleContainer` is mounted in the global provider tree, outside the Discovery jotai context, so it has no active tab id without new coupling — and resuming only the active tab would still leave every other tab in the mirrored mixed state (reported visible, WS paused) rather than restoring what was actually there.

The off-route pass now uses its own save slot and is a plain push/pop:

```js
// pause: save whatever send is right now — the real one, or the per-tab no-op
window.$$onekeyOffRouteWebSocketSend = window.WebSocket.prototype.send;
window.WebSocket.prototype.send = () => {};

// resume: put exactly that back
window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketSend;
window.$$onekeyOffRouteWebSocketSend = undefined;
```

A page the per-tab policy had paused gets the no-op back and stays paused; a page that was live gets the real `send` back. The per-tab policy keeps sole ownership of the shared slot, so the two never disagree.

Guarded by `offRouteWebsocketInject.test.ts`, whose second case is exactly this regression ("leaves a page the per-tab policy paused still paused after restore"), plus multi-cycle and idempotency cases — an earlier draft that keyed ownership off the shared slot passed one cycle and broke on the second.

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
originalix: Confirmed and fixed in b5d4633fd2 — the mechanism is real. `injectToResumeWebsocket` restores `WebSocket.prototype.send` from the shared `$$onekeyWebSocketSend` slot, and since `setCurrentWebTab` pauses each tab as the user switches away from it, background tabs accumulate that paused state. A blanket resume therefore woke tabs the per-tab policy had deliberately paused, while their `jsBridge.globalOnMessageEnabled` stayed `false`.

I took a different route than "resume WS only on the active tab", because `BrowserOffRouteThrottleContainer` is mounted in the global provider tree, outside the Discovery jotai context, so it has no active tab id without new coupling — and resuming only the active tab would still leave every other tab in the mirrored mixed state (reported visible, WS paused) rather than restoring what was actually there.

The off-route pass now uses its own save slot and is a plain push/pop:

```js
// pause: save whatever send is right now — the real one, or the per-tab no-op
window.$$onekeyOffRouteWebSocketSend = window.WebSocket.prototype.send;
window.WebSocket.prototype.send = () => {};

// resume: put exactly that back
window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketSend;
window.$$onekeyOffRouteWebSocketSend = undefined;
```

A page the per-tab policy had paused gets the no-op back and stays paused; a page that was live gets the real `send` back. The per-tab policy keeps sole ownership of the shared slot, so the two never disagree.

Guarded by `offRouteWebsocketInject.test.ts`, whose second case is exactly this regression ("leaves a page the per-tab policy paused still paused after restore"), plus multi-cycle and idempotency cases — an earlier draft that keyed ownership off the shared slot passed one cycle and broke on the second.

packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx
originalix: Confirmed and fixed in ad93084820 — the call site was wrong. I kept both call sites rather than moving the call, because the mechanism turned out to be different from the one described, and moving it would have opened a new hole.

What I could verify about the ordering:

- `initWebviewByRef` sets `updateIsDomReady(false)` and `setIsWebviewReady(true)` together at element attach, so `innerRef` becomes non-null well before the guest is dom-ready.
- But `onWebViewRef` does not fire in that commit. `DesktopWebView`'s `useImperativeHandle` publishes through `useWebViewBridge`, and `InpageProviderWebView`'s own `useImperativeHandle` has no dependency array, so the outer ref is re-invoked on its next render. The publish therefore lands **after** guest attach, on the pre-navigation document that the in-flight load is about to replace.
- `useImperativeHandle` is a layout effect, so a tab frozen by `react-freeze` never runs it. That is the real reason the ref site cannot reach the inactive keep-alive tabs — a stronger argument than "too early".

`dom-ready` fixes both of those, and also covers a case neither of us listed: a guest that reloads or navigates while off-route comes back with a fresh document that has none of the patches, and only `dom-ready` fires again for it.

I did not delete the ref-site call, because `onDomReady` reads `webviewRefs[id]` and that map is populated by the ref publish. When `dom-ready` beats the publish (cached page, `about:blank`, error document), `applyCurrentThrottleStateToWebview` early-returns and the ref site is the only remaining chance. Both scripts are idempotent, so keeping both costs nothing.

Also handled the async side: `executeJavaScript` returns a rejected promise rather than throwing once the guest is attached, which the synchronous `try/catch` could not absorb. The shared helper now catches both.

One correction to the runtime evidence I cited earlier in this PR: I read `applied: 11` against 16 live guest renderer processes as five missed injections. That inference does not hold — Chromium site isolation gives cross-origin iframes their own processes, so process count is not WebView count. The log now carries `total` (entries in `webviewRefs`) next to `applied` so the ratio is actually interpretable, and I have corrected the earlier comment.

originalix: Confirmed and fixed in ad93084820 — the call site was wrong. I kept both call sites rather than moving the call, because the mechanism turned out to be different from the one described, and moving it would have opened a new hole.

What I could verify about the ordering:

- `initWebviewByRef` sets `updateIsDomReady(false)` and `setIsWebviewReady(true)` together at element attach, so `innerRef` becomes non-null well before the guest is dom-ready.
- But `onWebViewRef` does not fire in that commit. `DesktopWebView`'s `useImperativeHandle` publishes through `useWebViewBridge`, and `InpageProviderWebView`'s own `useImperativeHandle` has no dependency array, so the outer ref is re-invoked on its next render. The publish therefore lands **after** guest attach, on the pre-navigation document that the in-flight load is about to replace.
- `useImperativeHandle` is a layout effect, so a tab frozen by `react-freeze` never runs it. That is the real reason the ref site cannot reach the inactive keep-alive tabs — a stronger argument than "too early".

`dom-ready` fixes both of those, and also covers a case neither of us listed: a guest that reloads or navigates while off-route comes back with a fresh document that has none of the patches, and only `dom-ready` fires again for it.

I did not delete the ref-site call, because `onDomReady` reads `webviewRefs[id]` and that map is populated by the ref publish. When `dom-ready` beats the publish (cached page, `about:blank`, error document), `applyCurrentThrottleStateToWebview` early-returns and the ref site is the only remaining chance. Both scripts are idempotent, so keeping both costs nothing.

Also handled the async side: `executeJavaScript` returns a rejected promise rather than throwing once the guest is attached, which the synchronous `try/catch` could not absorb. The shared helper now catches both.

One correction to the runtime evidence I cited earlier in this PR: I read `applied: 11` against 16 live guest renderer processes as five missed injections. That inference does not hold — Chromium site isolation gives cross-origin iframes their own processes, so process count is not WebView count. The log now carries `total` (entries in `webviewRefs`) next to `applied` so the ratio is actually interpretable, and I have corrected the earlier comment.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
originalix: I traced this one carefully and I do not think the described failure is reachable. Pushing back rather than changing the code, with the reasoning so it can be checked.

**(a) A pending trailing title surviving a real navigation.** The pending branch requires `isTitleOnlyUpdate`, which requires `loading === undefined`:

```ts
const isTitleOnlyUpdate =
  !isValidNewUrl && favicon === undefined && canGoBack === undefined &&
  canGoForward === undefined && loading === undefined;
```

Every navigation event carries `loading` explicitly — `onDidStartNavigation` sends `loading: true`, `onDidFinishLoad`/`onDidStopLoading`/`onDidFailLoad` send `loading: false` (WebContent.desktop.tsx:109-144). So on a real navigation `isTitleOnlyUpdate` is false, the pending branch is skipped, and control reaches `if (title !== tab.title) { clearPendingTitleUpdate(tab.id); ... }`. The only producer of a title-only event is `onPageTitleUpdated`, which never carries a URL.

That clear path is also taken essentially whenever a pending entry exists, because of an invariant in how it is created: an entry is only stored when `title && title !== tab.title` (actions.ts:1601), and its value is by construction the newest forwarded `document.title` — which is exactly what `getNavStatusInfo()`'s `webviewRef.getTitle()` returns on the navigation event.

The one gap is the 500 ms redirect debounce returning before the title block. That is self-correcting: `did-finish-load` follows with the new title and `loading: false`, is not throttled, and overwrites. In that window `tab.url` has not advanced either, so the tab is coherently still showing the old page rather than showing a stale title on a new one.

**(b) Bulk close leaking into a same-id restored tab.** Id reuse is real — `addWebTab` keeps a supplied id (actions.ts:596-598) and `reOpenLastClosedTab` re-adds the saved tab object — but a restored tab carries the same URL, so a surviving pending title belongs to that same page; writing it is correct. If the id is not restored, the timer is a no-op: `setWebTabData` returns early when `tabIndex === -1`.

One correction to the premise: `closeWebTab` does not delete `lastTitleUpdateFlags` either — it only calls `clearPendingTitleUpdate`. `lastTitleUpdateFlags` is never deleted anywhere, which matches the pre-existing `lastNavigationFlags` / `homeResettingFlags` pattern in this file. So the asymmetry described for that map does not exist, and a stale cooldown is inert anyway: the corrective title arrives through the unthrottled `did-finish-load` path regardless of the window.

**What I did take:** `closeAllWebTabs` genuinely was inconsistent with `closeWebTab` and leaked one throttle-window timer per closed tab, so it now clears them (73ee3dbbb1). I left the navigation-identity reset out — it would be defensive code for a path I cannot construct a failure for, and this PR is already large.

Happy to add the full reset helper if you can point at a reachable sequence I have missed.

originalix: I traced this one carefully and I do not think the described failure is reachable. Pushing back rather than changing the code, with the reasoning so it can be checked.

**(a) A pending trailing title surviving a real navigation.** The pending branch requires `isTitleOnlyUpdate`, which requires `loading === undefined`:

```ts
const isTitleOnlyUpdate =
  !isValidNewUrl && favicon === undefined && canGoBack === undefined &&
  canGoForward === undefined && loading === undefined;
```

Every navigation event carries `loading` explicitly — `onDidStartNavigation` sends `loading: true`, `onDidFinishLoad`/`onDidStopLoading`/`onDidFailLoad` send `loading: false` (WebContent.desktop.tsx:109-144). So on a real navigation `isTitleOnlyUpdate` is false, the pending branch is skipped, and control reaches `if (title !== tab.title) { clearPendingTitleUpdate(tab.id); ... }`. The only producer of a title-only event is `onPageTitleUpdated`, which never carries a URL.

That clear path is also taken essentially whenever a pending entry exists, because of an invariant in how it is created: an entry is only stored when `title && title !== tab.title` (actions.ts:1601), and its value is by construction the newest forwarded `document.title` — which is exactly what `getNavStatusInfo()`'s `webviewRef.getTitle()` returns on the navigation event.

The one gap is the 500 ms redirect debounce returning before the title block. That is self-correcting: `did-finish-load` follows with the new title and `loading: false`, is not throttled, and overwrites. In that window `tab.url` has not advanced either, so the tab is coherently still showing the old page rather than showing a stale title on a new one.

**(b) Bulk close leaking into a same-id restored tab.** Id reuse is real — `addWebTab` keeps a supplied id (actions.ts:596-598) and `reOpenLastClosedTab` re-adds the saved tab object — but a restored tab carries the same URL, so a surviving pending title belongs to that same page; writing it is correct. If the id is not restored, the timer is a no-op: `setWebTabData` returns early when `tabIndex === -1`.

One correction to the premise: `closeWebTab` does not delete `lastTitleUpdateFlags` either — it only calls `clearPendingTitleUpdate`. `lastTitleUpdateFlags` is never deleted anywhere, which matches the pre-existing `lastNavigationFlags` / `homeResettingFlags` pattern in this file. So the asymmetry described for that map does not exist, and a stale cooldown is inert anyway: the corrective title arrives through the unthrottled `did-finish-load` path regardless of the window.

**What I did take:** `closeAllWebTabs` genuinely was inconsistent with `closeWebTab` and leaked one throttle-window timer per closed tab, so it now clears them (73ee3dbbb1). I left the navigation-identity reset out — it would be defensive code for a path I cannot construct a failure for, and this PR is already large.

Happy to add the full reset helper if you can point at a reachable sequence I have missed.

packages/kit/src/states/jotai/contexts/discovery/actions.ts
originalix: I traced this one carefully and I do not think the described failure is reachable. Pushing back rather than changing the code, with the reasoning so it can be checked.

**(a) A pending trailing title surviving a real navigation.** The pending branch requires `isTitleOnlyUpdate`, which requires `loading === undefined`:

```ts
const isTitleOnlyUpdate =
  !isValidNewUrl && favicon === undefined && canGoBack === undefined &&
  canGoForward === undefined && loading === undefined;
```

Every navigation event carries `loading` explicitly — `onDidStartNavigation` sends `loading: true`, `onDidFinishLoad`/`onDidStopLoading`/`onDidFailLoad` send `loading: false` (WebContent.desktop.tsx:109-144). So on a real navigation `isTitleOnlyUpdate` is false, the pending branch is skipped, and control reaches `if (title !== tab.title) { clearPendingTitleUpdate(tab.id); ... }`. The only producer of a title-only event is `onPageTitleUpdated`, which never carries a URL.

That clear path is also taken essentially whenever a pending entry exists, because of an invariant in how it is created: an entry is only stored when `title && title !== tab.title` (actions.ts:1601), and its value is by construction the newest forwarded `document.title` — which is exactly what `getNavStatusInfo()`'s `webviewRef.getTitle()` returns on the navigation event.

The one gap is the 500 ms redirect debounce returning before the title block. That is self-correcting: `did-finish-load` follows with the new title and `loading: false`, is not throttled, and overwrites. In that window `tab.url` has not advanced either, so the tab is coherently still showing the old page rather than showing a stale title on a new one.

**(b) Bulk close leaking into a same-id restored tab.** Id reuse is real — `addWebTab` keeps a supplied id (actions.ts:596-598) and `reOpenLastClosedTab` re-adds the saved tab object — but a restored tab carries the same URL, so a surviving pending title belongs to that same page; writing it is correct. If the id is not restored, the timer is a no-op: `setWebTabData` returns early when `tabIndex === -1`.

One correction to the premise: `closeWebTab` does not delete `lastTitleUpdateFlags` either — it only calls `clearPendingTitleUpdate`. `lastTitleUpdateFlags` is never deleted anywhere, which matches the pre-existing `lastNavigationFlags` / `homeResettingFlags` pattern in this file. So the asymmetry described for that map does not exist, and a stale cooldown is inert anyway: the corrective title arrives through the unthrottled `did-finish-load` path regardless of the window.

**What I did take:** `closeAllWebTabs` genuinely was inconsistent with `closeWebTab` and leaked one throttle-window timer per closed tab, so it now clears them (73ee3dbbb1). I left the navigation-identity reset out — it would be defensive code for a path I cannot construct a failure for, and this PR is already large.

Happy to add the full reset helper if you can point at a reachable sequence I have missed.

originalix:
## Correction to the coverage claim above

In the previous comment I read `applied: 11` against 16 live guest renderer processes as "five WebViews were missed by the injection". **That inference does not hold and I am withdrawing it.**

Guest renderer process count is not WebView count. Chromium site isolation puts cross-origin iframes in their own processes, and `allowpopups` windows get their own too, so a DApp with an embedded widget contributes processes that were never in `webviewRefs`. An 11-vs-16 gap is at least as consistent with out-of-process iframes as with missed injections. `applied` also counted injections *issued*, not landed.

The call-site defect the reviewers raised separately is real and is fixed in ad93084820 — but it was established from the component ordering, not from this number. The log now carries `total` (entries in `webviewRefs`) alongside `applied`, so the ratio is interpretable on the next run.

The same-session release measurement in the previous comment is unaffected: it compares CPU across a single discrete transition with full pid overlap and does not depend on any WebView-to-process mapping.

### Also fixed from review, all verified against the code rather than assumed

- **b5d4633fd2** — the off-route pass reused the shared WebSocket save slot, so returning to the browser route resumed `WebSocket.send` on background tabs the per-tab policy had deliberately paused, leaving them with WebSocket running and jsBridge disabled. It now has its own slot and is a push/pop, so each page returns to whatever state it was actually in. Covered by `offRouteWebsocketInject.test.ts`.
- **ad93084820** — the throttle is now re-applied on guest `dom-ready` as well as on ref publish. Neither signal covers the other: `useImperativeHandle` is a layout effect, so a react-freeze'd tab never runs it, while `dom-ready` can arrive before the ref reaches `webviewRefs`. Also absorbs the async rejection Electron returns for a not-ready guest.
- **73ee3dbbb1** — `closeAllWebTabs` now clears pending title updates, matching `closeWebTab`.

One reviewer point I did not take: the title-throttle state was reported as leaking across navigation. I could not construct a reachable failure — every navigation event carries `loading`, which makes `isTitleOnlyUpdate` false and routes control to the branch that already clears the pending entry. Reasoning is in the inline reply; happy to revisit with a concrete sequence.

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
sidmorizon: # [🟠 P1] 新的独立保存槽在「返回路由」时序下反向踩踏:活跃 tab 的出站 WebSocket 会被置成 noop

## 问题

push/pop 独立槽解决了原来的方向(off-route resume 唤醒 per-tab 暂停的后台 tab),但引入了反向踩踏——off-route 的快照在 **pause 时刻**拍下,却在 **resume 时刻**无条件写回,而这两个时刻之间 per-tab 策略会修改同一个 `WebSocket.prototype.send`。关键在于返回路由时两个监听器的执行顺序是确定的、且 off-route 在后:

`useOnRouterChange` 用 `routerRef.current.push(callback)`(NavigationContainer.tsx:63)按 mount effect 顺序订阅;`BasicNavigation` 里 `<RootNavigator />` 渲染在 `{children}` **之前**(NavigationContainer.tsx:26-30),而 `BrowserOffRouteThrottleContainerGate` 属于 `{children}`。因此 `DesktopCustomTabBar`(在 RootNavigator 子树内)先订阅、先执行。

完整可达序列(desktop,tab A 活跃,交易类 DApp):

1. **离开路由** — `DesktopCustomTabBar` 的 focus 回调 `setCurrentWebTab('')`(index.tsx:227-239)→ `pauseDappInteraction(A)` → 共享槽 `$$onekeyWebSocketSend = realSend`,`send = noop_perTab`
2. **t+20s** — `throttleBrowserWebviewsOffRoute()` → `$$onekeyOffRouteWebSocketSend = noop_perTab`,`send = noop_offRoute`
3. **返回路由,第 1 个监听器** — `setCurrentWebTab(savedActiveTabId)` → `resumeDappInteraction(A)` → `send = $$onekeyWebSocketSend = realSend` ✅
4. **返回路由,第 2 个监听器** — `restoreBrowserWebviewsOnRoute()` → `send = $$onekeyOffRouteWebSocketSend = noop_perTab` ❌ **覆盖掉第 3 步**

## 影响

每次离开浏览器路由 ≥`20s` 再返回,**用户正在看的那个 tab** 出站 WebSocket 变成 no-op:下单、订阅、心跳静默失败,页面 visibility 已恢复为 visible、jsBridge 也已启用,所以没有任何征兆。且不会自愈——只有切到别的 tab 再切回来(触发一次 `resumeDappInteraction(A)`)才恢复,下一轮往返再次复现。

新增的 `offRouteWebsocketInject.test.ts` 覆盖不到:用例 2 的序列是 per-tab pause → offRoute pause → offRoute resume(per-tab 动作在 off-route 快照**之前**),而真实故障序列是 per-tab **resume 落在 off-route pause 与 resume 之间**。

## 建议

resume 不要无条件写回快照,只撤销「自己装的那个 noop」——别人(per-tab 策略)在窗口内改过 `send` 就放手:

```js
// pause
if (!window.$$onekeyOffRouteWebSocketSend) {
  window.$$onekeyOffRouteWebSocketSend = window.WebSocket.prototype.send;
  window.$$onekeyOffRouteNoop = () => {};
  window.WebSocket.prototype.send = window.$$onekeyOffRouteNoop;
}

// resume
if (window.$$onekeyOffRouteWebSocketSend) {
  if (window.WebSocket.prototype.send === window.$$onekeyOffRouteNoop) {
    window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketSend;
  }
  window.$$onekeyOffRouteWebSocketSend = undefined;
  window.$$onekeyOffRouteNoop = undefined;
}
```

这样第 4 步会因 `send !== noop_offRoute` 而跳过,per-tab 的决定得以保留;原来那条「per-tab 暂停过的页面 restore 后仍暂停」的语义也不变。建议同时补一个用例,把 per-tab resume 插在两次 off-route 注入**之间**。

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

packages/kit/src/views/Discovery/hooks/useWebTabs.ts
originalix: Fixed in 8d56b10edb (answered inline on the P1 thread above at the time; threading it here too so this one is not left open).

You were right that memoizing the returned object could not stop re-renders triggered by the component's own map-atom subscription. `useWebTabDataById` now subscribes to its own entry:

```ts
const tabAtom = useMemo(
  () => selectAtom(webTabsMapAtom(), (map) => map[id ?? ''] as IWebTab | undefined),
  [id],
);
const tab = useAtomValue(tabAtom, { store });
```

with the context store, following the same discipline as `useTokenFiatField`. This works because `buildWebTabData` reuses untouched tab objects — `setWebTabData` copies only the changed tab — so `Object.is` holds for every unaffected entry and only the changed tab's subscribers execute. A title tick on one tab now re-renders exactly one shell plus one nav bar instead of all of them.

The discovery context exposes its store accessor (`useDiscoveryContextStoreData`) for these derived-atom reads; regular consumers keep using the generated hooks.

packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts
originalix: Confirmed and fixed in e5dcb1e86f. You are right, and the ordering premise checks out: `DesktopCustomTabBar` calls `setCurrentWebTab('')` on blur and `setCurrentWebTab(savedActiveTabId)` on focus (DesktopCustomTabBar/index.tsx:227-239), so `resumeDappInteraction` on the active tab lands inside my pause/resume window. An unconditional write-back then re-stubbed the tab the user is looking at, with page visibility already restored and jsBridge already enabled — silent, and not self-healing until the next tab switch.

Taking your suggestion, with one change: I did not want the fix to depend on the listener order being what it is today, since that ordering is incidental (mount-effect order through `routerRef.current.push`) and nothing enforces it. Keeping the installed no-op and comparing identity makes the pass correct under *any* interleaving:

```js
// pause
window.$$onekeyOffRouteWebSocketSend = window.WebSocket.prototype.send;
window.$$onekeyOffRouteWebSocketNoop = () => {};
window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketNoop;

// resume — only undo our own replacement
if (window.WebSocket.prototype.send === window.$$onekeyOffRouteWebSocketNoop) {
  window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketSend;
}
window.$$onekeyOffRouteWebSocketSend = undefined;
window.$$onekeyOffRouteWebSocketNoop = undefined;
```

If anyone else touched `send` during the window we relinquish and leave their value alone; the original semantics you asked for (a page the per-tab policy paused stays paused across a round trip) are unchanged.

Added the case you asked for — per-tab resume between the two off-route injections, in the real order:

```
injectToPauseWebsocket          // blur: setCurrentWebTab('') -> pauseDappInteraction
injectToPauseWebsocketOffRoute  // +20s
injectToResumeWebsocket         // focus: setCurrentWebTab(saved) -> resumeDappInteraction
injectToResumeWebsocketOffRoute // focus: off-route restore, runs after
=> expect the page to still be able to send
```

I checked the test actually discriminates: reverting only the identity guard turns that one case red and leaves the other five green. Suite is 6 cases now.

originalix:

@originalix

Reviewed commit `fc39ed092106`. This report always reflects the latest reviewed changes.

### Review summary

The PR changes browser-tab ownership from mutable shared arrays and broad map subscriptions to copy-on-write Jotai state with ID-only and per-tab selectors. On desktop, it also makes LRU eviction cross the react-freeze boundary, resolves live WebView references on demand, and adds route-level guest throttling through Page Visibility and WebSocket injection. URL-validation and signing boundaries remain unchanged, but the new desktop memory budget and title-persistence policy retain material performance risks.

**What needs attention:** Confirm that desktop guest-renderer memory remains safe on lower-memory machines and that live document titles no longer drive sustained full-tab persistence.

### Issues to address

- **P1 · Desktop keep-alive cap can consume several gigabytes**
  - **Impact:** The new cap permits 16 Electron guest renderers to remain live. Using the PR's own estimate of roughly 200–330 MB per renderer, this reserves approximately 3.2–5.3 GB for guests alone, so lower-memory desktops can enter heavy swap or terminate renderers even though the eviction fix now makes the original cap effective.
  - **Suggested action:** Keep the desktop cap at 8 while validating the eviction fix, then introduce an adaptive or measured budget before increasing it.
- **P2 · Title throttling still produces a sustained database write loop**
  - **Impact:** A DApp that updates document.title every second schedules a trailing update every five seconds. That update calls setWebTabData, which reaches buildWebTabs with the default persistence policy; because five seconds exceeds the 500 ms debounce window, even one tab still writes the full tab snapshot about 12 times per minute indefinitely.
  - **Suggested action:** Make title-only updates non-persistent and persist titles only with navigation or other durable tab actions.

### Validation gaps

- The focused Jest test could not start because the workspace has no Yarn `node_modules` state file.
- The added VM tests cover WebSocket interleavings but not the Page Visibility overrides, title timing/persistence behavior, or the raised memory cap.
- Electron runtime behavior and lower-memory desktop pressure were not measured in this workspace.
- Existing GitHub discussions could not be checked because the GitHub API was unreachable, so the inline drafts require a duplicate-comment check before posting.

originalix:
Thanks for the review. I am intentionally keeping both items as-is.

**Desktop keep-alive cap**

The cap of 16 live WebViews is the expected desktop behavior, not an accidental increase. The measured organic working set for the target heavy-tab sessions was 10–17 live WebViews; a cap of 8 caused repeated reloads once more than eight tabs had been visited. Keeping 16 preserves that working set while still replacing the previous effectively unbounded retention with a hard ceiling. The memory tradeoff is understood and accepted for this PR; residual memory investigation remains tracked separately in #12686.

**Title persistence**

The remaining write rate is also an accepted tradeoff. The final multi-tab run reduced idle persistence to p50 10/min and p90 16/min while keeping the latest document title durable across restarts. Making title-only updates non-persistent would reintroduce stale restored titles, and changing that persistence model further is outside the scope of this PR.

Neither item is considered blocking, so I will keep the current implementation.

packages/kit/src/views/Discovery/utils/explorerUtils.ts
```diff
@@ -117,3 +117,96 @@ export const injectToResumeWebsocket = `
...
+  }
+  if (
+    window.WebSocket.prototype.send === window.$$onekeyOffRouteWebSocketNoop
+  ) {
+    window.WebSocket.prototype.send = window.$$onekeyOffRouteWebSocketSend;
+  }
+  window.$$onekeyOffRouteWebSocketSend = undefined;
+  window.$$onekeyOffRouteWebSocketNoop = undefined;
+})()
+`;
```
sidmorizon, line 166: **A DApp page can be left permanently unable to send data over its live connection after the user returns to the browser**

Severity: `non-severe`

The saved original send function is thrown away (`window.$$onekeyOffRouteWebSocketSend = undefined` at `packages/kit/src/views/Discovery/utils/explorerUtils.ts:163`) even when the restore step decided not to put it back, so a page can be left with a dead no-op sender and no way to recover it short of a reload.
Impact: In a specific ordering, a DApp tab silently stops sending anything over its live connection (prices, orders, chat) until the user reloads it.

<details>
<summary>Interleaving between the per-tab pause/resume pair and the new off-route pair</summary>

Two independent pause/resume mechanisms now patch `WebSocket.prototype.send` in the same guest realm:
- per-tab: `injectToPauseWebsocket` / `injectToResumeWebsocket`, saving into `$$onekeyWebSocketSend` (`packages/kit/src/views/Discovery/utils/explorerUtils.ts:99-119`), driven by `updateDappActivityInteraction` on tab switch.
- off-route: `injectToPauseWebsocketOffRoute` / `injectToResumeWebsocketOffRoute`, saving into `$$onekeyOffRouteWebSocketSend`.

The unit tests cover the ordering "per-tab pause first, then off-route pause". The reverse ordering is reachable: while the user is off-route (`isThrottled === true`), a guest that reloads or navigates comes back with fresh globals and `applyCurrentThrottleStateToWebview` (`packages/kit/src/views/Discovery/utils/browserOffRouteThrottle.ts:110-115`, called from `onDomReady` in `packages/kit/src/views/Discovery/components/WebContent/WebContent.desktop.tsx:212`) installs the off-route no-op first, saving the *real* send into `$$onekeyOffRouteWebSocketSend`. If anything then triggers `pauseDappInteraction` for that tab while still off-route (e.g. `closeWebTab` -> `activateAdjacentTab` -> `setCurrentWebTab` -> `pauseDappInteraction`), `injectToPauseWebsocket` sees `$$onekeyWebSocketSend` unset and saves the off-route no-op as the "real" send, installing its own no-op.

On return: `injectToResumeWebsocketOffRoute` finds `send !== $$onekeyOffRouteWebSocketNoop`, correctly declines to restore — but still clears `$$onekeyOffRouteWebSocketSend`, discarding the only surviving reference to the real send. The subsequent per-tab resume then writes back the off-route no-op, and both save slots are now gone, so `send` stays a no-op for the lifetime of that document.

</details>

sidmorizon: 核验结论:**这条链是可达的**,而且触发面比上面描述的更宽——第 2 步的「reload」不是必要条件。另外上面隐含的修复方向(放手分支不清空槽)不完备,且会引入新缺陷。补两点,不重复已述根因。

## 1. 不需要 reload:开机以来「从未被激活过」的 tab 天然满足前置条件

链条真正需要的前置是「`$$onekeyWebSocketSend` 未设置 + off-route noop 已装」。`$$onekeyWebSocketSend` 的唯一写入者是 `injectToPauseWebsocket`,而它只由 `updateDappActivityInteraction` 的 pause 分支驱动,pause 的对象永远是**被切走的那个 tab**(`setCurrentWebTab` 中 `pauseDappInteraction(currentTabId)`)。因此启动时恢复、其 WebView 已被 keep-alive 挂载、但用户从未激活过的 tab,共享槽一直是未设置状态;离开路由时 off-route pass 会给**所有** live webview 装上 noop_off。此时该文档就已处于「共享槽空 + send === noop_off」——与 reload 后的新文档等价。

所以触发条件可以简化为:长时间运行的桌面会话 + 存在未激活过的恢复 tab + 离开浏览器路由期间从侧栏关掉两个 tab。

## 2. 为什么必须是「两次关闭」:第一次关闭无法 pause 任何 tab

这是整条链真正的瓶颈,值得写明以便复现:

- 离开路由时 `DesktopCustomTabBar` 已 `setCurrentWebTab('')`,`activeTabIdAtom === ''`;
- 第一次侧栏关闭 → `activateAdjacentTab` → `setCurrentWebTab(newId)`,此时 `currentTabId === ''` → `pauseDappInteraction('')` → `updateDappActivityInteraction` 中 `tabId` 回退取 `activeTabIdAtom` 仍为 `''` → `const ref = tabId ? webviewRefs[tabId] : null` 得到 `null`,**空转**;真正发生的是对新 tab 的 `resumeDappInteraction`(共享槽为空时同样空转);
- 第二次侧栏关闭时 `currentTabId` 已非空 → 这才产生真实的 `injectToPauseWebsocket`,把 `noop_off` 当作真 send 存进共享槽。

顺带:Cmd+W 走不到这条路——`useShortcuts.desktop.ts` 的 `CloseTab` 分支被 `if (!isAtBrowserTab.current) return;` 挡住;可达入口是常驻侧栏的关闭按钮。

## 3. 「放手分支不清空 `$$onekeyOffRouteWebSocketSend`」不能修好,还会引入新缺陷

保留该槽并不能救回真 send:此时**共享槽里已经是 `noop_off`**,后续 per-tab resume 读的是共享槽,照样写回 noop。而且 off-route pause 的守卫是

```js
if (!window.WebSocket || window.$$onekeyOffRouteWebSocketSend) { return; }
```

槽不清空 → 该文档之后**永远不再被 off-route 节流**(pause 变成永久空转),等于用一个静默的功能失效换掉另一个。

## 4. 最小正确改动:在「捕获侧」拒绝把已知 stub 当作真 send

污染发生在 `injectToPauseWebsocket` 捕获的那一刻,修在那里即可,off-route 两个脚本可保持不变:

```js
// injectToPauseWebsocket
(function(){
  if (window.WebSocket) {
    if (!window.$$onekeyWebSocketSend) {
      var current = window.WebSocket.prototype.send;
      // never adopt the off-route pass's stub as the "real" send
      window.$$onekeyWebSocketSend =
        current === window.$$onekeyOffRouteWebSocketNoop
          ? window.$$onekeyOffRouteWebSocketSend
          : current;
    }
    window.WebSocket.prototype.send = () => {};
  }
})()
```

这样第 3 步存入共享槽的就是真 send,第 4 步 off-route 放手 + per-tab resume 写回真 send,链条断在源头;现有 6 个用例语义不变。建议补的用例是**与用例 2 相反的顺序**:`pauseOffRoute → pauseWebsocket → resumeOffRoute → resumeWebsocket`,断言 `send === realSend`。

[Claude session ↗](https://claude.ai/code/session_01Y5wtcx9dNDwA4yraVKMfg3)

---

sidmorizon:
@codex review

sidmorizon:
@codex security review

PR #12732, opened 2026-08-03
fix: prevent swap modal flicker (OK-57605)

zhaono1:
## Summary

- keep the Home Token List swap pair stable while the modal initializes
- skip the redundant initial swap type switch only for this modal entry
- preserve the existing Swap/Swiper page behavior

## Verification

- verified in the local Electron development environment
- confirmed the selected pair no longer passes through an empty placeholder state

Jira: https://onekeyhq.atlassian.net/browse/OK-57605


zhaono1:
Closing because OK-57605 was moved to draft PR #12730 as requested.

PR #12743, opened 2026-08-04
feat: cross-network token search in Receive selector (OK-58557)

franco-chan:
[OK-58557](https://onekeyhq.atlassian.net/browse/OK-58557)

----------

## Problem

Under a single-network scope, the Receive token selector could only search the current network. Searching `usdt trx` from an Ethereum scope returned nothing, and the empty state said the wallet held no assets — so there was no signal that the user had to go back to the home page and switch to All Networks to receive a Tron token.

Originally reported in 2024-09 and re-reported twice since; the May search upgrade (OK-55171) only fixed ranking/filtering *within* All Networks scope.

## Root cause

Four independent locks, all of which had to be opened:

1. The backend search request takes the scope's `networkId` — `onekeyall--0` is the only mechanism for a cross-chain query.
2. `needNetworksMap` required `isAllNetworks`, so local network-field matching never ran.
3. The aggregate token catalog isn't loaded under a single-network scope.
4. The press path's account-matching / account-creation block was wrapped in `if (network?.isAllNetworks)`, so an other-network row would fall through to a wrong-account fallback.

A fifth, discovered while testing: the backend matches the keyword string **as a whole**, so `usdt trx` matches nothing on any `networkId`. Combined queries now strip an exact network keyword and scope the request to that network (PRD Phase 1, never previously implemented).

## What changed

Search results are split into a current-network section and an "Other networks" section. When the current network has no match it collapses to a single line rather than filling the screen. Selecting an other-network row resolves (or creates) the target account and goes straight to that network's receive address page. **The home network scope is never changed.**

Opt-in through route params (`enableCrossNetworkSearch`), so Send, Exchange and Swap keep their current behavior. Excluded for custom networks (their search goes through an RPC contract lookup) and for the LP-token filter (its backend results are network-scoped by design).

## Fixes that came along the way

- **`searchTokens` `$key` now carries `networkId`.** Native tokens across chains share `uniqueKey: 'native'` and an empty address, so the previous key collided and `$key`-keyed maps collapsed them into one row — searching `eth` rendered the same token N times. Pre-existing bug; All Networks had it too, masked by aggregate pre-filtering.
- **Selector search empty state unified across all four routes.** A no-result *search* previously showed the browse-state "You don't hold any crypto" copy (the `EmptySearch` branch was dead code — it reads a jotai atom `TokenSelector` never writes). Now "No result".
- **A failed search no longer masquerades as an empty result** — it gets its own error state with retry, both full-screen and as an inline row when local results exist but the cross-network query failed.
- **Rows with no balance record render nothing instead of `-`.**

## i18n

7 new keys, all 19 languages populated. Verified programmatically that every key is present, non-empty, and that the `{network}` placeholder survived translation in all 19.

## Verification

Web, single-network (Optimism) scope:

- `usdt trx` → collapsed no-match line + Other networks section with Tron rows
- `usdc` → both sections populated, current-network header named after the network
- `eth` → 15 distinct networks, no duplicate rows
- garbage query → "No result"
- search request blocked, no local results → error state + Retry; unblock + Retry recovers
- search request blocked, local results present → inline error row under "Other networks", current-network results still listed
- navigating to a Tron receive address (passcode + account creation) → home scope still Optimism throughout
- All Networks scope regression checked item by item

67 unit tests pass (13 new). `agent:check` green for both the commit and pr profiles.

## Known trade-offs

- Under a single-network scope, other-network rows show no balance even when the token is actually held — local balances are only fetched for the current network, and a cross-network fan-out was judged too expensive.
- Other-network coverage is whatever the backend returns, same as All Networks today.

[OK-58557]: https://onekeyhq.atlassian.net/browse/OK-58557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


franco-chan:
Both issues fixed in `06dddccb63`.

**P1 · Token/network alias collisions bypass scoped search** — confirmed and fixed. Ethereum carries `code`/`shortcode` `eth`, so `eth base` matched two networks and the extractor bailed, sending the combined string to the all-network endpoint where whole-string matching finds nothing.

Rather than special-casing native symbols, the disambiguation uses a signal already in the network catalog: when several words look like networks, a word that also equals **its own network's symbol** (`eth` → Ethereum, `symbol: 'ETH'`) is the token term, not the network. Those are dropped and the word that can only be a network wins (`base` → Base, whose symbol is `ETH`, not `base`). If that leaves zero or several candidates the query is genuinely ambiguous and passes through unchanged, so `eth sol` still does not extract. The filter only runs when there is more than one candidate, so the single-match path — `usdt trx` — is untouched.

Regression tests added for `eth base`, the reversed `base eth`, and same-network aliases (`tron trx` → Tron + `trx`); `trx trx` now covers the "stripping leaves nothing" case that `tron trx` used to. Verified in the browser: `eth base` returns ETH / WETH / weETH / rsETH / rETH / cbETH all Base-badged under **Other networks**, and `usdt trx` still returns the Tron USDT set.

**P2 · Custom-network search enabled before network metadata resolves** — confirmed and fixed. Added `!!network` to the gate, since `!network?.isCustomNetwork` reads as true while it is still undefined.

The report's second half is worth calling out because the fix above alone would not have closed it: the stale-request guard keys on `[accountId, networkId, filterContext, keywords]`, which does not include the gate. The pre- and post-resolution runs therefore shared an identity and `isLatest()` passed for **both**, so an earlier differently-scoped response landing after the abort could overwrite the newer one. The gate is now part of that identity, which also covers the mirror case on non-custom networks.

**Not addressed:** the validation gaps are environmental (no Yarn state file / no `api.github.com` in the review sandbox). Locally `yarn jest packages/shared/src/utils/tokenSelectorCrossNetworkUtils.test.ts packages/shared/src/utils/tokenUtils.test.ts` passes 56/56 and `yarn agent:check --profile commit` is green. End-to-end UI validation has been done on **Web only** — Desktop / iOS / Android / Extension are flagged for QA in [OK-58557](https://onekeyhq.atlassian.net/browse/OK-58557), along with a cold custom-network search.

[OK-58557]: https://onekeyhq.atlassian.net/browse/OK-58557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

originalix (commented):
@franco-chan

Reviewed commit `06dddccb6395`.

- **P1 · Incompatible cross-network rows cannot produce a receive address**
- **P2 · Multi-word network names bypass scoped search**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12743#issuecomment-5175027097)

franco-chan:
Both fixed in `96a41e4f1b`.

**P1 · Incompatible cross-network rows cannot produce a receive address** — confirmed. Traced the path: for an Others account `getAllNetworkAccounts` cannot match, so the press falls to `createAddress({ walletId, indexedAccountId: account.indexedAccountId, ... })` — and Others accounts have no `indexedAccountId`, so the HD/HW batch workflow cannot supply the imported/watching credential and the tap dies in the surrounding catch with no address and no feedback.

Filtered as suggested, using the existing `accountUtils.isAccountCompatibleWithNetwork` rather than a new compatibility notion — it already checks impl match plus the account's own `networks` allow-list. Applied only to Others scopes; HD/HW accounts stay unfiltered since they can create an account on any network. Worth noting the filter is deliberately *not* "same network only": an imported EVM key genuinely can receive on Base/Arbitrum/BNB Chain, and that is most of the feature's value for imported accounts, so those rows are kept and only incompatible impls are dropped.

Regression tests cover imported-EVM (keeps EVM, drops Solana/Tron), watch-only restricted to a single network, a non-EVM imported account, and the defensive no-`networkId` row.

**P2 · Multi-word network names bypass scoped search** — confirmed, and broader than Bitcoin Cash: `BNB Chain`, `zkSync Era`, `Manta Pacific`, `X Layer`, `Ethereum Classic`, `Conflux eSpace` and others are all multi-word presets, so every one of them was unreachable through its display name (the single-word `code`/`shortcode` aliases still worked, e.g. `usdt bch`).

Now matches every contiguous run of words instead of single words, longest-first so `bitcoin cash` beats the `bitcoin` nested inside it. Span length is capped at 4 words (longest preset name is 3) to bound the scan. This composes with the symbol disambiguation from the previous round: phrase-level matches go through the same filter, so `eth base` still resolves to Base.

Tests added for `usdt bitcoin cash`, leading-position `bitcoin cash usdt`, the single-word `usdt bch` alias, and `bitcoin cash` alone (whole query is the network phrase → no token term → no extraction). Verified in the browser: `usdt bnb chain` now returns USDT / USDT+ / syrupUSDT / SUSDT all BNB Chain-badged under **Other networks**.

**Validation gaps:** the Jest and `api.github.com` failures are environmental to the review sandbox. Locally the suite is 19/19 and `yarn agent:check --profile commit` is green. On UI coverage — cross-network grouping, multi-word names, `eth base`, `usdt trx`, retry/error states and target-account creation are verified on **Web**; native split-runtime, desktop, extension, and the imported/watch-only account cases are handed to QA in [OK-58557](https://onekeyhq.atlassian.net/browse/OK-58557). I have not run those myself and am not claiming otherwise.

[OK-58557]: https://onekeyhq.atlassian.net/browse/OK-58557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

originalix (commented):
@franco-chan

Reviewed commit `96a41e4f1b2a`.

- **P1 · Shared utility introduces a forbidden kit-bg dependency**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12743#issuecomment-5175027097)

franco-chan:
Fixed in `883973fb77` — `filterTokensByAccountNetworkCompatibility` and its four tests now live in `packages/kit/src/utils/tokenSelectorAccountCompatibility.ts`, alongside the other unit-tested kit helpers. `packages/shared/src/utils/tokenSelectorCrossNetworkUtils.ts` is back to importing types from shared only. No behavior change — same function, same call site, same tests, all 19 still pass.

One correction to the report for the record, since it affects how the rule should be read elsewhere: the import was `import type`, so it was erased at compile time and could not have caused the build-order or dependency-cycle failures described — and `packages/shared/src/utils/accountUtils.ts` and `packages/shared/types/account.ts` both already type-import `IDBAccount` from the same kit-bg path today. The layering rule itself is real and worth honoring, which is why I moved it rather than citing that precedent; but "shared has no kit-bg imports" is not the current state of the repo, and a stricter enforcement pass would need to cover those two files too.

**Validation gaps:** the Jest failure is environmental to the review sandbox — locally `yarn jest packages/shared/src/utils/tokenSelectorCrossNetworkUtils.test.ts packages/kit/src/utils/tokenSelectorAccountCompatibility.test.ts` is 19/19 and `yarn agent:check --profile commit` (lint + full-project tsc) is green. On UI coverage, my position is unchanged from the previous round: cross-network grouping, `eth base`, `usdt bnb chain`, `usdt trx`, retry/error states and account creation are verified on **Web only**; native split-runtime, desktop, extension, and the imported/watch-only cases are handed to QA in [OK-58557](https://onekeyhq.atlassian.net/browse/OK-58557), where the imported/watch-only case is explicitly flagged as not yet verified on a real device.

[OK-58557]: https://onekeyhq.atlassian.net/browse/OK-58557?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx
```diff
@@ -871,6 +912,11 @@ function TokenSelector() {
                 includingNonExistingAccount: true,
                 deriveType,
                 excludeTestNetwork: false,
+                // Single-network accountId cannot be expanded to all-network
+                // accounts without this flag (it resolves the indexedAccountId
+                // first); without a match the flow would always fall through to
+                // createAddress and needlessly wake hardware wallets.
+                fetchAllNetworkAccounts: isCrossNetworkTokenPress,
```
sidmorizon, line 939: # 🟠 P1 `fetchAllNetworkAccounts` 配单链 accountId 会解析出错误的 indexedAccountId

## 问题

跨链行的 `token.accountId` 为空，所以这里的 `params` 是 `{ accountId: account.id, networkId: network.id }`，**没有传 `indexedAccountId`**，同时新增了 `fetchAllNetworkAccounts: true`。这会让 `ServiceAllNetwork._getAllNetworkAccountsUncached` 里的 `isAllNetwork` 为 true，进而调用 `getIndexedAccountIdForAllNetwork` → `accountUtils.buildAllNetworkIndexedAccountIdFromAccountId({ accountId })`。

该函数只为 All Networks 的 mock 账户 id 设计（模板 `0000/$$INDEX$$`，id 形如 `hd-1--0000/0`，`usedPath.split('/')[1]` 恰好是 index）。这里传入的是**单链账户 id**，例如 EVM 的 `hd-1--m/44'/60'/0'/0/0`：

```ts
usedPath = "m/44'/60'/0'/0/0";
parseInt(usedPath.split('/')[1], 10); // "44'" -> 44
// => buildIndexedAccountId({ walletId: 'hd-1', index: 44 }) === 'hd-1--44'
```

`isValidIndexedAccountId('hd-1--44')` 返回 true，所以不会抛错、也不会回退，`resolvedIndexedAccountId` 被静默污染成 `hd-1--44`（BTC 作用域下则是 `49/84/86`）。

## 影响

常见情况（钱包账户数 `< 45`）：`getAccountsInSameIndexedAccountId('hd-1--44')` 返回空，`dbAccounts` 为空，所有网络都只由 `includingNonExistingAccount` 补出 `accountId: ''` 的占位项，`matchedAccount?.accountId` 恒为空 → **每次跨链选币都必然走 `createAddress`**，正是本行注释声称要避免的 "needlessly wake hardware wallets"。

极端但可达的情况：钱包确实存在 index `44` 的账户时，`matchedAccount` 会命中**另一个账户**，`executeOnSelect({ ...token, accountId: 其它账户 id })`。对 `mergeDeriveAssetsEnabled` 为 false 的网络（EVM / SOL / TRON），`useReceiveToken.onSelect` 会走 `accountId: t.accountId ?? accountId` 分支，收款页展示的是**错误账户的地址**。

## 建议

作用域内已有正确的 `indexedAccountId`（route param），直接传入即可让 `getIndexedAccountIdForAllNetwork` 在 `isValidIndexedAccountId(indexedAccountId)` 处短路；Others 账户没有 `indexedAccountId`，仍走 `othersWalletAccountId` 分支，需保持不传：

```ts
const { accountsInfo } =
  await backgroundApiProxy.serviceAllNetwork.getAllNetworkAccounts({
    ...params,
    ...(isCrossNetworkTokenPress &&
    indexedAccountId &&
    !accountUtils.isOthersAccount({ accountId: params.accountId })
      ? { indexedAccountId }
      : undefined),
    includingNonExistingAccount: true,
    deriveType,
    excludeTestNetwork: false,
    fetchAllNetworkAccounts: isCrossNetworkTokenPress,
  });
```

[Claude session ↗](https://claude.ai/code/session_01LQiy2P6Vyq2UWqBdhdoEkJ)

---

franco-chan: 确认并已修复（`834c8af31c`）——按你的建议传入 scope 的 `indexedAccountId`，并保留 Others 账户的排除条件。

顺带在运行时验证了一次，比静态推导还更严重一点：该钱包只有一个 indexed account（`hd-1--0`，EVM 账户 `hd-1--m/44'/60'/0'/0/0`），旧的调用形状**直接抛错** `record not found: IndexedAccount hd-1--44`——正是你推导出的那个 index 44。这个异常被外层 `catch { // pass }` 吞掉，`accounts` 保持为空，所以常见情况下不是「匹配到空占位项」而是「压根没有结果」，最终同样是每次跨链选币都走 `createAddress`，即该行注释声称要避免的行为。传入 `indexedAccountId` 后同一调用返回 132 条、其中 85 条带真实 accountId，Tron 正确解析为 `hd-1--m/44'/195'/0'/0/0`。

All Networks 路径不受影响：那里 `isCrossNetworkTokenPress` 为 false，新值是 `undefined`，而缓存 key（`params.indexedAccountId ?? ''`）与 `isOthersWallet` 判断（`!indexedAccountId`）对 undefined 和缺省的处理完全一致，所以是逐字等价的。

originalix (approved):

PR #12753, opened 2026-08-05
feat: support paragon hip-3 markets (OK-59307)

Minnzen:
[OK-59307](https://onekeyhq.atlassian.net/browse/OK-59307)

----------

## Summary

- Add a sub-DEX registry in `packages/shared` so Perps supports Paragon (`para`, Hyperliquid perp dex index 8) alongside `xyz`, with `para:UNITREE` as the headline market.
- Replace the hardcoded "there is exactly one sub-DEX and it is `xyz` at dexIndex 1" assumption across 11 files with registry-driven helpers.
- Fix nine defects that make `para` unusable: dropped positions, stuck-loading ticker data, out-of-range ticker ctx lookup, order asset lookup failure, a stale universe cache, an unreachable `perpsAssetMetaMap` entry, a global search that mislabels every sub-DEX result as `xyz:`, delisted assets surfacing in global search, and bare-symbol logo resolution.
- Ships as a hot update, so this PR targets `release/v6.5.0` instead of `x`.

## Intent & Context

Requested in Slack `#app-perps`: Paragon just deployed the Unitree pre-market contract (subscription opens Aug 10, listing expected late August) and is competing with `xyz` for US-equity share. A third HIP-3 DEX is considered likely.

A server-config-only rollout is impossible: `ServiceHyperliquid.refreshTradingMeta` truncated the `allPerpMetas()` response with `slice(0, 2)`, taking the first two DEXs **by position**. Paragon sits at index 8, so it was discarded before any config could reference it.

## Root Cause

Three separate root causes were addressed.

**1. Positional truncation and scattered `dexIndex === 1` branches.** `allPerpMetas()` is indexed by the real Hyperliquid perp dex index. Slicing the first two entries only worked because `xyz` happens to be index 1. Every downstream consumer then assumed `dexIndex === 1` meant "subtract the xyz offset".

**2. Stale universe cache (found during local verification).** The persisted `tradingUniverses` is a positional 2D array. Every "does the cache have data?" predicate was "does any slot have items", which is true for a 2-slot cache written before `para` existed. So the cache was always considered valid and `para` never entered the universe. This surfaced as two user-visible bugs with one shared cause: searching `para` returned nothing, and favoriting a `para` token did not stick (`usePerpsFavorites` resolves favorites against the same stale universe). `usePerpTokenSelector` compounded it by calling `refreshTradingMeta()` fire-and-forget with nothing re-reading the result afterwards.

**3. Every "sub-DEX" identifier path was really an "xyz" path.** Global search built its coin as `` `xyz:${name}` `` for any non-main result; `perpsAssetMetaMap` lookups tried the bare symbol first and took the first match; logo URLs were derived from the bare symbol. With one sub-DEX these were indistinguishable from correct. With two they are not: `para:UNITREE` opened as `xyz:UNITREE`, `para:STX` could never reach its own metadata entry because main-DEX `STX` shadowed it, and `para:STX` rendered the Stacks icon (both symbols exist, on different DEXs, at different prices).

## Design Decisions

**Keep a local compact `dexIndex` (0 = main, 1 = xyz, 2 = para) rather than switching to Hyperliquid's real dex index.** `SimpleDbEntityPerp.tradingUniverses` is a positionally-indexed persisted array, and that entity has no schema version and no migration hook. Changing index semantics would make existing users read misaligned universes on the first cold start after upgrade. `SUB_DEX_LIST` is therefore append-only; the constraint is documented on the constant.

**Do not upgrade `@nktkas/hyperliquid`** (pinned 0.32.2). `perpDexs` and `allPerpMetas` already exist, `dex` params are unconstrained strings, and order `assetId` is an unsigned integer with no upper bound. Bumping a minor would require re-validating all order/TIF/TWAP/scale semantics for no gain here.

**`selectPerpMetasByDex` asserts the prefix.** It verifies the selected meta's `universe[0].name` starts with the expected prefix, so a future index shift on Hyperliquid's side degrades to an empty slot instead of a whole DEX of mislabelled assets.

**Global search filters delisted assets against the local universe, and never blocks on it.** The search index still carries delisted rows — `xyz:UNITREE` survives there after the contract migrated to `para`, so one query returned the same ticker twice with one dead row. The client filters using `isDelisted` from the local universe. Because `universalSearch` resolves only after every search type settles, a universe refresh is started *before* the search request so both are in flight together, and it is awaited only when the perp search actually returned rows to filter. A user searching an address or dapp never waits; a failed refresh degrades to no filtering rather than to an empty list.

**`perpsAssetMetaMap` lookup tries the full coin before the bare symbol.** First match wins, so bare-first made any dex-scoped key unreachable whenever a bare entry for the same symbol existed. Flipping the order is what makes a `para:STX` config entry addressable at all.

**Verified on-chain rather than assumed**: `para` is dex index 8, so its assetId offset is `100000 + 8 * 10000 = 180000` and `para:UNITREE` is `180019`. `candleSnapshot` and `l2Book` both accept the full coin name. A zero-position account returns a valid all-zero clearinghouse state for `para`, so account aggregation is not degraded.

## Changes Detail

**New**
- `packages/shared/src/utils/perpsDexUtils.ts` — single source of truth for sub-DEX index/offset math: `getDexAssetIdOffset`, `getDexIndexByCoin`, `getDexIndexByAssetId`, `toCtxIndex`, `toAssetId`, `normalizeDexCoin`, `isPerpsUniverseCacheComplete`, `buildCoinFromSearchAssetType`.
- `packages/kit-bg/src/services/ServiceHyperLiquid/utils/perpMetaSelection.ts` — `selectPerpMetasByDex`, extracted so the selection logic is unit-testable without network access.

**Registry**
- `perp.constants.ts` — `SUB_DEX_LIST` plus derived `DEX_PREFIXES` / `DEX_ASSET_ID_OFFSETS`. `getDexIndexByAssetId` bounds each dex range on both sides, so ids belonging to unregistered HIP-3 namespaces return `-1` instead of being misattributed (Hyperliquid dex indexes are non-contiguous).

**Functional fixes**
- `actions.ts` — clearinghouse merge previously combined only the main DEX and `xyz`, silently dropping `para` positions from the list and from account aggregation. Also removes a duplicated WS→array mapping in favour of the shared implementation.
- `usePerpsAssetCtx.ts` — replaced a `str.length === 6 && startsWith('11')` heuristic that classified `180019` as the main DEX, leaving `para` market data stuck loading.
- `PerpTickerBarDesktop.tsx` — `assetId >= 110000 ? 1 : 0` classified `para` as `xyz` and computed ctxIndex `70019`.
- `ServiceHyperliquidExchange.ts` — `normalizePerpsCoin`'s `/^xyz:/i` regex upper-cased `para:unitree` to `PARA:UNITREE`, which does not match the Hyperliquid universe, so order asset lookup failed.
- `PerpsGlobalEffects.tsx`, `usePerpsFavorites.ts`, `usePopularTickers.ts`, `usePerpTokenSelector.ts` — all now use `isPerpsUniverseCacheComplete()`; the selector re-reads the universe after `refreshTradingMeta()` resolves. The favorites and popular-tickers hooks degrade to the stale universe if the refresh throws, because `usePromiseResult` neither retries nor re-throws — a thrown error there would blank the favorites bar for the rest of the session.
- `usePerpMarketDetail.ts` — `perpsAssetMetaMap` candidates now try the full coin before the display name.
- `changeActiveAsset` no longer substitutes `dexUniverses[0]` when the requested coin is missing from its DEX; it takes the unresolved path instead of silently opening a different market.

**Global search**
- `ServiceUniversalSearch.ts` — filters delisted assets against the local universe, refreshes that universe in parallel with the search request when it is missing or older than 5 minutes, and awaits the refresh only when there are perp rows to filter.
- `SimpleDbEntityPerp.ts` — records `tradingUniversesUpdatedAt` so staleness is detectable.
- `UniversalSearchPerpItem.tsx` — `assetType` is now treated as the dex prefix itself rather than a boolean "is main DEX"; a result from a DEX this client does not register renders nothing instead of a row that no-ops on tap and stars an empty coin.

**Logos**
- `getHyperliquidTokenImageUris(coin)` returns `[prefixed, bare]` for a sub-DEX asset and `[bare]` otherwise, wired through the `Token` component's existing `tokenImageUris` fallback chain. Applied to the token selector (both platforms and the web trigger), favorites bar, market detail and intro, mobile market page, portfolio, share sheet, share image canvas, desktop tray and the cold-start preload list, plus global search results. Without it `para:UNITREE` fell back to a placeholder (`UNITREE.png` is 404) and `para:STX` rendered the Stacks icon (`STX.png` exists, but it is a different asset on a different DEX).
- Spot rows deliberately keep the display name rather than the raw fill coin — a spot coin such as `@149` or `PURR/USDC` has no valid image path.

**Collaborator changes landed on this branch**
- #12770 (Kahn) — lowercase `xyz` / `para` source badges on the active ticker, both token selectors and global search, with a blue leverage badge alongside. HIP-3 leverage is recovered from the cached universe by full coin key, because the search response can report `0`. Localized descriptions were added for all 19 locales; #12772 then trimmed that i18n update back to exactly the two new keys, leaving no existing string added, removed or changed.
- `211c4822` (Leon, pushed directly to this branch) — declares `assetTypeVersion: 2` on the perps search request so the backend can return the real dex prefix as `type` without breaking older clients.

**Verified as needing no change**
- `['', ...DEX_PREFIXES]` in `ServiceHyperliquid.ts` and `SubscriptionConfig.ts` — length-driven, picks up `para` automatically.
- `hyperliquidPortfolioUtils.ts` — already dex-name driven throughout.
- `parseDexCoin` — generic `:` split, no `xyz` literal.
- Persisted favorites (coin-name strings) and `perpsActiveAssetAtom` (offset-encoded assetId) — this change only **appends** local index 2, so slots 0 and 1 keep their meaning and existing data stays valid.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (JS only, no native changes — that is what makes the hot update possible)
- **Risk Areas**:
  - **assetId arithmetic is the highest risk.** A wrong offset does not throw; it silently routes an order to a different asset. Covered by round-trip unit tests and an end-to-end check against a live `allPerpMetas()` response.
  - **First cold start after upgrade.** Existing users hold a 2-slot `tradingUniverses` cache. The cache-completeness guard now forces an awaited refresh in that case, which adds one network round trip before Perps becomes interactive — deliberately preferred over serving a dex set that omits `para`.
  - **Global search latency.** The universe refresh runs alongside the search request and is awaited only when perp rows came back, so a query that matches no perp asset is unaffected. If the refresh fails, results are returned unfiltered rather than empty.
  - **`SUB_DEX_LIST` ordering is load-bearing.** Reordering or removing entries would corrupt the positional persisted cache.

## Test plan

Automated (done):
- [x] 40 unit test cases added across 6 suites (2 new files, 4 extended), covering offsets, coin/assetId resolution, meta selection, cache completeness, search coin construction, metadata lookup order and image URI fallbacks.
- [x] 2046 tests passing across the Perps / Hyperliquid / kit-bg service suites.
- [x] End-to-end assetId check against a live `allPerpMetas()` capture: `para:UNITREE` resolves to `180019` and round-trips to ctxIndex 19; `xyz` ids stay in the 110000 range; main-DEX `BTC` stays `0`.
- [x] `yarn agent:check --profile pr` passes.
- [x] Residue sweep clean for `dexIndex === 1`, `XYZ_ASSET_ID_LENGTH`, `ctxMap.get('xyz')`, `/^xyz:/`, `slice(0, 2)`.
- [x] CDN check: every listed `para` asset and a sampled `xyz` set return 200 for their prefixed image path.

Manual (done):
- [x] Global search: `para` results open the correct market, and their icons resolve through the dex-prefixed path.

Manual (pending):
- [ ] **Cold-start upgrade path**: launch on the previous bundle, let the universe persist, switch to this bundle without clearing data. `xyz` markets and existing favorites must stay correct; `para` must appear; no `Asset id not found for coin: para:` in logs.
- [ ] Token selector: search `UNITREE` and `AVGO` — both must show real price and 24h change, not `--`. Delisted `para` assets (`RDDT`, `NAVER`) must **not** appear.
- [ ] Favorites: star a `para` token, confirm it appears under FAVS and survives a reload.
- [ ] Chart / orderbook: switch `BTC` → `para:UNITREE` → `xyz:NVDA` repeatedly; candles, L2 and ticker must all refresh with no residue from the previous asset.
- [ ] Order panel: `para:UNITREE` shows 10x max leverage and isolated margin. Verify the resolved assetId is `180019` via `getSymbolMeta` or a debug log — **do not verify by placing a live order**.
- [ ] Share image for a spot fill still renders its token icon (the share path takes the display name, not the raw fill coin).
- [ ] **Requires QA to open a small real position on `para`** — the position-merge fix has no runtime evidence yet.

## Follow-ups (not blocking this PR)

- Server: add the Paragon tab to `tokenSelectorTabs`. **Must ship after the client OTA**, otherwise older clients receive a token list referencing a universe they do not have.
- Server: index `para` assets in `/wallet/v1/proxy/hyperliquid/perpsAsset` and honour `assetTypeVersion=2` by sending the dex prefix as `type`. The client now declares that version on every request and already reads `type` as the prefix, dropping unregistered DEXs, so `para` results simply do not appear until the index carries them.
- Server: `perpsAssetMetaMap` keys stay bare for `para` assets, except where a symbol collides across DEXs. Today that is `STX` (main = Stacks, `para` = Seagate), which needs a `para:STX` entry; `AVGO` exists on both `xyz` and `para` and should be decided alongside. Existing `xyz` entries are unaffected.
- Server (optional): filter delisted assets out of the search index directly. `allPerpMetas` returns all 10 DEXs and 477 assets (178 delisted) in a single ~67KB request, which is the cheapest authoritative source. The client-side filter added here covers the gap in the meantime.
- Market module images come from the server-provided `n` field and cannot be corrected client-side.

[OK-59307]: https://onekeyhq.atlassian.net/browse/OK-59307?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1410,21 +1396,19 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+      dexCount: allPerpMetas?.length ?? 0,
     });
-    if (perpMetaMultiDexList?.length) {
-      if (perpMetaMultiDexList.length >= 2) {
-        perpMetaMultiDexList = perpMetaMultiDexList.slice(0, 2);
-      }
-      const universes = perpMetaMultiDexList.map((meta, dexIndex) =>
+    const perpMetaByDex = selectPerpMetasByDex(allPerpMetas);
+    if (perpMetaByDex.length) {
+      const universes = perpMetaByDex.map((meta, dexIndex) =>
```
Minnzen: **A dex missing from one `allPerpMetas()` response wipes its persisted universe, and the completeness guard no longer notices.**

`selectPerpMetasByDex` always returns a fixed-length array, using `undefined` for a dex the server did not send. `(meta?.universe || [])` then turns that `undefined` into `[]` and persists it over previously-good data.

Verified by running the real modules:

```
selectPerpMetasByDex([mainOnlyMeta])            -> length 3, slots 1/2 undefined
.map((meta, i) => (meta?.universe || []).map(…)) -> [1, 0, 0]
isPerpsUniverseCacheComplete([[{}], [], []])     -> true
```

So a cache of `[main, xyz, para]` plus one transient main-only response becomes `[main, [], []]`, still reports complete, and a later `getSymbolsMetaMap(['xyz:NVDA'])` throws `Asset id not found`. That breaks order placement for the already-shipped `xyz` dex, not just the new `para` one.

This is specifically introduced by the fixed-length padding. Before this change a short response left a length-1 cache, which the new length check in `isPerpsUniverseCacheComplete` would have rejected and recovered from. The padding removes that signal.

Suggested fix — keep the previous slot instead of writing an empty one (same treatment needed for `marginTablesMapList`):

```ts
const { universesByDex: prevUniversesByDex } = await this.getTradingUniverse();
const universes = perpMetaByDex.map((meta, dexIndex) => {
  const universe = meta?.universe;
  // A dex the server did not return this round must not wipe a slot that
  // already has data — the cache is positional and has no other recovery.
  if (!universe?.length) {
    return prevUniversesByDex?.[dexIndex] ?? [];
  }
  return universe.map((item, index) => ({
    ...item,
    assetId: toAssetId({ dexIndex, index }),
  }));
});
```

packages/kit-bg/src/services/ServiceHyperLiquid/utils/perpMetaSelection.ts
```diff
@@ -0,0 +1,32 @@
...
+  return [
+    allMetas[0] ?? undefined,
+    ...SUB_DEX_LIST.map((item) => {
+      const meta = allMetas[item.hlDexIndex] ?? undefined;
+      const firstName = meta?.universe?.[0]?.name;
+      if (
+        firstName &&
+        !firstName.startsWith(`${item.prefix}${DEX_SEPARATOR}`)
+      ) {
+        return undefined;
```
Minnzen, line 34: **Dropping a whole dex here is completely silent.**

The prefix assertion is the right idea — turning a hyperliquid index shift into an empty slot beats a whole dex of mislabelled assets. But when it fires, `para` (or `xyz`) disappears from the app with no log line, no perf mark, and nothing in Sentry. The only signal would be a user reporting "my para markets are gone", which is close to unsearchable after the fact.

This matters more than it looks: it is also the diagnostic that would explain the cache-overwrite path flagged on `ServiceHyperliquid.ts`.

```ts
import { markPerpsColdStartPerf } from '@onekeyhq/shared/src/performance/perpsColdStartPerf';

if (firstName && !firstName.startsWith(`${item.prefix}${DEX_SEPARATOR}`)) {
  markPerpsColdStartPerf('service_perp_meta_dex_prefix_mismatch', {
    prefix: item.prefix,
    hlDexIndex: item.hlDexIndex,
    firstName,
  });
  return undefined;
}
```

packages/shared/src/utils/perpsDexUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+// new sub-DEX was registered is shorter than the registry. It still looks
+// populated, which would otherwise let callers serve a dex set that silently
+// omits the new dex — treat that as a cache miss and force a refresh.
+export function isPerpsUniverseCacheComplete(
+  universesByDex: unknown[][] | undefined,
+): boolean {
+  if (!universesByDex || universesByDex.length < SUB_DEX_LIST.length + 1) {
+    return false;
+  }
+  return universesByDex.some((items) => (items?.length ?? 0) > 0);
```
Minnzen: **`some()` accepts a cache whose main perps DEX is empty.**

The length check above is sound — it correctly rejects a cache written before `para` was registered. The content check is too loose: any single non-empty slot passes, so `[[], [xyz…], [para…]]` reads as complete.

`PerpsGlobalEffects` recovers from that via `deferTradingMetaRefresh(1200)`, but `usePopularTickers` and `usePerpsFavorites` have no deferred fallback — they would serve a list with no BTC/ETH for the whole session.

Slot 0 is the actual invariant here: the main perps DEX is never legitimately empty.

```ts
// Slot 0 is the main perps DEX and is never legitimately empty, so gate on it
// instead of `some()` — a partial write that only filled a sub-DEX must not
// read as a complete cache.
return (universesByDex[0]?.length ?? 0) > 0;
```

Note this is complementary to, not a substitute for, the `ServiceHyperliquid.ts` fix: the two cover disjoint failure shapes (`[[], x, y]` here vs `[main, [], []]` there). Both are needed.

packages/kit/src/views/Perp/hooks/usePerpTokenSelector.ts
```diff
@@ -85,7 +85,14 @@ export function usePerpTokenSelector() {
...
       lastRefreshTradingMetaTime = now;
-      void backgroundApiProxy.serviceHyperliquid.refreshTradingMeta();
+      // Re-read after the refresh lands: the first refreshAllAssets() above
+      // serves the persisted universe, which is one dex short right after a
+      // release that registers a new sub-DEX. Without this the selector keeps
+      // showing the stale dex set for the whole session.
+      void backgroundApiProxy.serviceHyperliquid
+        .refreshTradingMeta()
+        .then(() => refreshAllAssets())
+        .catch(() => undefined);
```
Minnzen: **A failed refresh is swallowed and still consumes the 5-minute throttle.**

`lastRefreshTradingMetaTime = now` is assigned before the request is made, and the rejection is discarded without logging. One transient network failure therefore pins the selector's dex set for a full 5 minutes with no trace anywhere — which is another route into exactly the stale-dex-set symptom this hook is trying to fix.

```ts
.catch((error) => {
  // Reset the throttle so a transient failure does not pin the dex set
  // for a full 5 minutes.
  lastRefreshTradingMetaTime = 0;
  defaultLogger.perp.hyperliquid.coldStartInitializationError({
    type: 'refresh_trading_meta',
    error,
  });
});
```

`'refresh_trading_meta'` is already in that logger's `type` union, so no logger change is needed.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Fixed in 7424cab1a6.

Extracted `mergePerpDexSlots` in `perpMetaSelection.ts` and routed both `universes` and `marginTablesMapList` through it, so a slot that came back `undefined` this round falls back to its previously cached value instead of being persisted as empty. `marginTablesMapList` had the same hazard and is covered too.

Regression test added in `perpMetaSelection.test.ts` — `keeps the previous slot for a dex the response omitted` reproduces the main-only response against a `[main, xyz, para]` cache and asserts both sub-dex slots survive.

packages/kit-bg/src/services/ServiceHyperLiquid/utils/perpMetaSelection.ts
Minnzen: Fixed in 7424cab1a6.

Added `markPerpsColdStartPerf('service_perp_meta_dex_prefix_mismatch', { prefix, hlDexIndex, firstName })` before the `return undefined`, so a dropped dex now shows up in the cold start benchmark stream with enough context to tell which registry entry drifted and what the server actually sent.

packages/shared/src/utils/perpsDexUtils.ts
Minnzen: Fixed in a9bb7ed3f2.

Switched the content check from `some()` to gating on slot 0. The length check above is unchanged, so the two guards now cover disjoint shapes: length catches a cache written before a sub-dex was registered, slot 0 catches a partial write that only filled a sub-dex.

Test added in `perpsDexUtils.test.ts` — `rejects a cache whose main dex slot is empty`. The other five cases in that block still pass unchanged.

packages/kit/src/views/Perp/hooks/usePerpTokenSelector.ts
Minnzen: Fixed in 9206271a38.

The catch handler now resets `lastRefreshTradingMetaTime` to 0 so the next mount can retry immediately, and logs through `defaultLogger.perp.hyperliquid.coldStartInitializationError({ type: 'refresh_trading_meta', error })`. That `type` was already in the logger's union, so no logger change was needed.

packages/kit/src/views/Perp/hooks/usePerpsFavorites.ts
```diff
@@ -74,12 +75,10 @@ export function usePerpsFavorites(options?: {
...
-      if (
-        !universesByDex ||
-        universesByDex.length === 0 ||
-        universesByDex.every((u) => u.length === 0)
-      ) {
+      // If data is missing, or the cache predates a newly registered sub-DEX,
+      // force refresh from API. A short array is just as unusable as an empty
+      // one here: favorites on the missing dex would silently fail to resolve.
+      if (!isPerpsUniverseCacheComplete(universesByDex)) {
         await backgroundApiProxy.serviceHyperliquid.refreshTradingMeta();
```
sidmorizon: # [🟠 P1] 升级后首次冷启动 refresh 失败会清空整个收藏栏

## 问题

这个 guard 从旧的"缓存完全为空才 refresh"改成了"缓存不完整就强制 `await refreshTradingMeta()`"。OTA 之后所有存量用户的持久化 `tradingUniverses` 都只有 2 个 slot,`isPerpsUniverseCacheComplete` 必然返回 `false`,所以每个存量用户的首次冷启动都会走进这个分支。一旦 `refreshTradingMeta()` 抛错(离线、弱网、HL API 抖动),异常会让整个 executor promise 拒绝;`usePromiseResult` 捕获后既不重试也不 rethrow(见 `usePromiseResult.ts` 的 "don't re-throw" 分支),`taggedUniverse` 就永久停在 `undefined`。

```ts
if (!isPerpsUniverseCacheComplete(universesByDex)) {
  await backgroundApiProxy.serviceHyperliquid.refreshTradingMeta(); // throws → 整个 promise 拒绝
  ...
}
```

## 影响

`favoriteItems` 返回 `[]` 且 `isReady === false`,收藏栏在整个会话内保持空白——尽管本地那份 2-slot 旧缓存足以解析用户已有的全部 main/`xyz` 收藏(只是缺 `para`)。PR 之前"缓存非空就直接展示"在离线场景是可用的,这是本 PR 引入的边界回归,且触发人群恰好是升级后第一次打开 App 的全部存量用户。`usePopularTickers.ts` 第 `77-82` 行是完全相同的模式,热门列表同样整会话为空。deps 只有 `[favoritesMode]`,后台其它路径(如 `PerpsGlobalEffects` 的 `deferTradingMetaRefresh`)即使随后刷新成功,也不会让这两个 hook 重新执行。

## 建议

refresh 失败时降级为旧缓存,而不是让 promise 拒绝(`usePopularTickers` 同样处理):

```ts
if (!isPerpsUniverseCacheComplete(universesByDex)) {
  try {
    await backgroundApiProxy.serviceHyperliquid.refreshTradingMeta();
    const res =
      await backgroundApiProxy.serviceHyperliquid.getTradingUniverse();
    universesByDex = res.universesByDex;
  } catch {
    // Serve the stale (incomplete) universe: existing favorites still
    // resolve; only the newly registered dex stays missing until the
    // next successful refresh.
  }
}
```

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1410,25 +1399,40 @@ export default class ServiceHyperliquid extends ServiceBase {
     markPerpsColdStartPerf('service_refresh_trading_meta_start');
 
     // oxlint-disable-next-line @cspell/spellchecker
-    let perpMetaMultiDexList = await infoClient.allPerpMetas();
+    const allPerpMetas = await infoClient.allPerpMetas();
```
sidmorizon, line 1417: # [🟡 P2] refreshTradingMeta 无跨调用方去重,升级后首次冷启动会并发打 3-4 次 allPerpMetas

## 问题

`refreshTradingMeta` 本身没有 in-flight 去重。本 PR 让 `isPerpsUniverseCacheComplete` 在存量用户升级后的首次冷启动必然判定缓存不完整,于是 `PerpsGlobalEffects`(awaited)、`usePerpsFavorites`、`usePopularTickers`,再加上打开选择器时的 `usePerpTokenSelector`,会在同一时刻各自独立调用 `refreshTradingMeta()`——每次都是一个完整的 `infoClient.allPerpMetas()`(全部 HL dex 的 universe + marginTables,重量级响应)。`PerpsGlobalEffects.tsx` 第 `975-993` 行已经为自己做了本地 single-flight(`refreshTradingMetaPromise`),说明去重是这条路径的既定意图,但它覆盖不到其它调用方。

## 影响

PR 之前这个并发场景只在全新安装(缓存全空)出现;现在它发生在**每个存量用户**的第一次升级后启动,也就是 OTA 铺量当天的全量用户:同一台设备并发 `3-4` 个相同的重请求,拖慢 Perps 首屏可交互(PR 自述接受的是"一次网络往返",实际是 N 次),并对 HL API 形成 thundering herd。每个并发调用还各自执行 `getTradingUniverse → mergePerpDexSlots → setTradingUniverse` 的读改写(结果幂等,无数据损坏,纯浪费)。

## 建议

在 service 层做 promise single-flight,调用方全部自动受益(`PerpsGlobalEffects` 的本地版之后也可移除):

```ts
private _refreshTradingMetaPromise: Promise<void> | undefined;

@backgroundMethod()
async refreshTradingMeta() {
  if (this._refreshTradingMetaPromise) {
    return this._refreshTradingMetaPromise;
  }
  this._refreshTradingMetaPromise = this._refreshTradingMeta().finally(() => {
    this._refreshTradingMetaPromise = undefined;
  });
  return this._refreshTradingMetaPromise;
}
```

或直接用 `cacheUtils.memoizee(..., { promise: true, maxAge: 短 TTL })`(仓库内已有同款用法,如 `_fetchHlPortfolioMemo`)。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

originalix (commented):
@Minnzen

Reviewed commit `9206271a38be`.

- **P1 · Universal search still routes Para assets to the XYZ DEX**
- **P2 · Concurrent metadata refreshes can overwrite a complete DEX cache**
- **P2 · Separatorless DEX URLs are ambiguous with main-market symbols**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

packages/kit/src/views/Perp/hooks/usePerpsFavorites.ts
Minnzen: Confirmed and fixed in `115f4fde51`.

Verified the failure chain rather than taking it on trust: `usePromiseResult`'s catch block (`usePromiseResult.ts:312-327`) deliberately does not re-throw and does not retry, and the executor's deps are `[favoritesMode]`, so a rejection leaves `taggedUniverse` at `undefined` permanently — `favoriteItems` stays `[]` and `isReady` stays `false` for the rest of the session. And you're right that the trigger population is the worst possible one: every existing user's persisted cache has 2 slots, so `isPerpsUniverseCacheComplete` returns `false` for all of them on the first cold start after the OTA.

This is a regression I introduced. Before this PR the branch only ran when the cache was completely empty, so an offline user with a populated cache was served directly.

Applied your suggested shape to both hooks — degrade to the stale universe instead of rejecting:

```ts
if (!isPerpsUniverseCacheComplete(universesByDex)) {
  try {
    await backgroundApiProxy.serviceHyperliquid.refreshTradingMeta();
    const res =
      await backgroundApiProxy.serviceHyperliquid.getTradingUniverse();
    universesByDex = res.universesByDex;
  } catch {
    // Every existing user hits this branch on the first cold start after
    // the release, so a rejection here would blank the favorites bar for
    // the whole session (usePromiseResult neither retries nor re-runs).
    // Serving the stale universe still resolves every pre-existing
    // favorite; only the newly registered dex stays missing.
  }
}
```

`usePopularTickers.ts` got the same treatment, as you called out.

Net behaviour on a failed refresh: pre-existing main/`xyz` favorites and the popular list render from the stale cache exactly as they did before this PR; only `para` is absent until a later refresh succeeds. That is strictly better than the blank-for-the-session outcome.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Fixed in `e711fa1e49`, with the service-layer single-flight you suggested.

```ts
async refreshTradingMeta() {
  // read-merge-write against the positional universe cache is not atomic, so
  // concurrent callers could interleave and drop a slot another call just
  // wrote. Independent main-runtime hooks do call this in parallel.
  if (!this.refreshTradingMetaPromise) {
    this.refreshTradingMetaPromise = this._refreshTradingMeta().finally(() => {
      this.refreshTradingMetaPromise = undefined;
    });
  }
  return this.refreshTradingMetaPromise;
}
```

Went with the explicit promise rather than `memoizee({ promise: true, maxAge })` because a short TTL would also suppress *sequential* refreshes that are legitimately wanted — e.g. `usePerpTokenSelector` deliberately re-reads after its refresh resolves, and `PerpsGlobalEffects` defers one by 1200ms. Clearing in `finally` dedupes only genuine in-flight overlap and leaves the retry paths intact.

Two notes on scope:

- I left the local `refreshTradingMetaPromise` in `PerpsGlobalEffects.tsx:975-993` in place. It is now redundant, but removing it touches the cold-start path for no behavioural gain, which I would rather not do in a hot-update PR. Worth a follow-up cleanup.
- Correctness beyond the wasted requests: your point that the read-merge-write is idempotent held before, but it stopped being purely a waste once `mergePerpDexSlots` landed in `7424cab1a6` — two callers could each read the same `prev`, and the later write would drop a slot the other had just filled. So this fix covers a real interleaving hazard now, not only the thundering herd.

originalix (commented):
@Minnzen

Reviewed commit `e711fa1e49a3`.

- **P1 · Universal search still hard-codes non-main assets to XYZ**
- **P2 · Separatorless DEX URLs still collide with main symbols**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchPerpItem.tsx
```diff
@@ -46,14 +46,21 @@ export function UniversalSearchPerpItem({
...
 
   const isPerpsType = assetType === 'perps';
-  // For perps type, coin is just name; for xyz type, coin needs prefix
+  // `assetType` carries the dex prefix verbatim ('perps' for the main DEX), so
+  // it must drive the coin — assuming every non-main result is xyz would send
+  // `xyz:UNITREE` for a para asset and leave the instrument unresolvable.
   const coin = useMemo(
-    () => (isPerpsType ? name : `${XYZ_DEX_PREFIX}${name}`),
-    [isPerpsType, name],
+    () => buildCoinFromSearchAssetType({ assetType, name }) ?? '',
```
sidmorizon, line 51: # [🟡 P2] 未注册 dex 的搜索结果仍渲染成"死行",收藏星标会把空字符串写进持久化 watchlist

## 问题

`handlePress` 对 `!coin` 做了守卫,但组件其余部分没有:未注册 dex 的结果行(`buildCoinFromSearchAssetType` 返回 `undefined`,`coin === ''`)仍然完整渲染,点击是无任何反馈的静默 no-op,而同一行的收藏星标依旧可用——`<MarketPerpsStarV2 perpsCoin={coin} />` 接到的是 `''`。`usePerpsStarV2Checked` / `addPerpsIntoWatchListV2` 都没有空串守卫(`MarketStarV2.tsx:143-148`、`marketV2/actions.ts:565-576`),点星会把 `{ chainId: '', contractAddress: '', perpsCoin: '' }` 持久化进 Market watchlist。

## 影响

触发场景是服务端先于客户端 OTA 返回第三个 HIP-3 dex 的搜索结果(正是 `buildCoinFromSearchAssetType` 设计要防御的场景)。后果:1)所有未注册 dex 的行共享同一个 `''` watchlist 条目——`find((item) => item.perpsCoin === perpsCoin)` 按精确字符串匹配,给任意一行点星后所有这类行都显示已收藏,取消其中一个会移除共享条目;2)持久化的空 `perpsCoin` 条目会一直留在 watchlist 数据里;3)整行点击静默无反应,用户无从得知原因。工具函数的 docstring 写的是 "callers can skip the result",但这里只 skip 了 press,没有 skip 渲染。

## 建议

对无法解析的结果直接不渲染该行(与 docstring 语义一致,也一并消除死行和空串收藏两个问题):

```tsx
const coin = useMemo(
  () => buildCoinFromSearchAssetType({ assetType, name }),
  [assetType, name],
);
// A dex this build does not support cannot be opened or favorited — skip the row.
if (!coin) {
  return null;
}
```

若产品上希望保留展示,则至少将星标改为 `coin` 为空时不渲染,并在 `addPerpsIntoWatchListV2` 里加空串守卫。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

originalix (commented):
@Minnzen

Reviewed commit `2e3e2637b8df`.

- **P1 · An empty registered DEX slot is treated as a complete universe cache**
- **P2 · Legacy separatorless links can still be decoded as the wrong DEX**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchPerpItem.tsx
Minnzen: Confirmed and fixed in `57ac97cf85`.

You're right that the press guard was the wrong layer. I verified the star path rather than assuming: `UniversalSearchPerpItem.tsx:124` renders `<MarketPerpsStarV2 perpsCoin={coin} />` unconditionally, and neither `usePerpsStarV2Checked` nor `addPerpsIntoWatchListV2` guards an empty string — so an unsupported-dex row really would persist `{ chainId: '', contractAddress: '', perpsCoin: '' }`, and every such row would then share that one entry through the exact-match `find`.

Took your primary suggestion — skip the row entirely, which matches what the helper's docstring already promised:

```tsx
// A dex this build does not know cannot be opened or favorited. Rendering the
// row anyway would give a silent no-op on tap, and the star would persist an
// empty `perpsCoin` that every such row then shares in the market watchlist.
if (!coin) {
  return null;
}
```

The now-redundant `if (!coin) return;` inside `handlePress` is gone with it.

I did not add the defensive empty-string guard to `addPerpsIntoWatchListV2`. It is a real gap, but it belongs to the Market module and has callers outside this change; widening a hot-update PR into that path is not something I want to do without a Market owner looking at it. Worth a separate issue if you agree — happy to file it.

packages/kit/src/views/Perp/components/usePerpTokenUrlSync.ts
```diff
@@ -62,10 +78,32 @@ function decodeCoinFromUrl(urlToken: string): string {
...
+  if (!isAmbiguousLegacyGuess) {
+    return coin;
+  }
+  try {
+    const { universesByDex } =
+      await backgroundApiProxy.serviceHyperliquid.getTradingUniverse();
+    const exists = universesByDex?.some((assets) =>
+      assets?.some((asset) => asset.name === coin),
+    );
+    return exists ? coin : urlToken.toUpperCase();
```
sidmorizon, line 120: # [🟡 P2] 冷启动空缓存时,合法的 legacy 子 DEX 深链会被永久误判成主 DEX 币

## 问题

存在性校验把"币不在 universe 里"和"universe 还没加载"混为一谈。全新安装 / 清空存储的用户(典型:web 端收到分享链接 `?token=xyzNVDA`)在 `usePerpTokenUrlSync` 的 focus effect 里触发本函数时,`getTradingUniverse()` 返回空缓存——这不走 `catch`(不是错误),`universesByDex?.some(...)` 对空数组直接得 `false`,于是合法的 legacy 链接被降级为 `urlToken.toUpperCase()`(`XYZNVDA`)。该 effect 与 `PerpsGlobalEffects` 的 awaited `refreshTradingMeta()` 是两条互不等待的异步流,竞态窗口真实存在。

## 影响

本 commit 之前,legacy 链接纯语法解码为 `xyz:NVDA`:即使冷启动时 universe 为空、`changeActiveAsset` 先走 unresolved 路径,persisted atom 里存的仍是正确 coin,refresh 完成后的后续 switch 流程能自动恢复。现在 atom 里存的是 `XYZNVDA`,任何后续 refresh 都无法再解析它——用户停在一个永远无数据的空面板,直到手动重新选币。也就是说:歧义修复在"缓存已就绪"时是正确的,但在缓存未就绪时把原本可自愈的场景变成了不可自愈,而空缓存恰好是分享链接最常见的落地场景(新用户)。

## 建议

把"缓存不完整"视为无结论,保留语法猜测(与 `catch` 分支同样的取舍),只有在完整缓存中确认缺失才回退:

```ts
const { universesByDex } =
  await backgroundApiProxy.serviceHyperliquid.getTradingUniverse();
// An unloaded/incomplete cache proves nothing — keep the syntactic guess,
// same as the catch path. Only a complete universe can prove absence.
if (!isPerpsUniverseCacheComplete(universesByDex)) {
  return coin;
}
const exists = universesByDex.some((assets) =>
  assets?.some((asset) => asset.name === coin),
);
return exists ? coin : urlToken.toUpperCase();
```

`isPerpsUniverseCacheComplete` 已在同包 `perpsDexUtils` 中导出,无新依赖。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchPerpItem.tsx
sidmorizon: 已在 `57ac97cf85` 验证:组件级 `if (!coin) return null;` 位于全部 hooks 之后,行、星标与空串写入路径一并消除,与建议一致,本条视为已解决。

同意不在本热更 PR 里动 `addPerpsIntoWatchListV2`——它属于 Market 模块且有本改动之外的调用方,范围控制正确。空串守卫作为独立 issue 跟进是合适的处理方式,建议提及 `usePerpsStarV2Checked` 的精确匹配语义一并评估。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

packages/kit/src/views/Perp/components/Portfolio/PerpPortfolioContent.tsx
```diff
@@ -1159,8 +1159,8 @@ function PerpPortfolioContentComponent({
               <XStack gap="$1.5" alignItems="center">
                 <Token
                   size="xxs"
-                  tokenImageUri={getHyperliquidTokenImageUrl(
-                    mostTradedTokenDisplayName,
+                  tokenImageUris={getHyperliquidTokenImageUris(
+                    fillsStats.mostTraded || mostTradedTokenDisplayName,
```
sidmorizon: # [🟡 P2] most traded 为现货时把原始 spot coin 传进 perp 图片 helper,图标回归为坏链

## 问题

`fillsStats.mostTraded` 是 fills 里的原始 `fill.coin`(`portfolioStats.ts:197-206`),现货成交会是 `@149` 或 `PURR/USDC` 这类原始形态——正因如此,本组件 `mostTradedTokenDisplayName`(第 `541-555` 行)专门做了 `isSpotInstrument` → `spotPairDisplayMap` / `split('/')` → `getSpotTokenDisplayName` 的映射。本次改动把优先级反过来了:`getHyperliquidTokenImageUris(fillsStats.mostTraded || mostTradedTokenDisplayName)` 会在 mostTraded 非空时直接使用原始值,`parseDexCoin('@149')` 无 `:` 分隔符,产出 `.../hyperliquid/@149.png`;`PURR/USDC` 更会产出 `.../hyperliquid/PURR/USDC.png`(`/` 破坏路径)——两者都是必然 404 的坏链。

## 影响

改动前 `getHyperliquidTokenImageUrl(mostTradedTokenDisplayName)` 用的是解析后的 base 名(如 `HYPE`、`PURR`),现货 most traded 的图标正常显示;改动后凡是"最常交易的是现货对"的用户,Portfolio 统计卡上的图标退化为 fallback 图标(且每次渲染发起一次必 404 的请求)。周边统计逻辑(`spotRealizedPnl` 等)明确支持现货 fills,该场景并非边缘。这是本 commit 为解决 sub-DEX 图标碰撞而引入的现货侧回归。

## 建议

只对 perp instrument 使用原始 coin,现货保持原有 display-name 路径:

```tsx
tokenImageUris={
  fillsStats.mostTraded && !isSpotInstrument(fillsStats.mostTraded)
    ? getHyperliquidTokenImageUris(fillsStats.mostTraded)
    : [getHyperliquidTokenImageUrl(mostTradedTokenDisplayName)]
}
```

`isSpotInstrument` 已在本文件 import(第 `39` 行),无新依赖。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

originalix (commented):
@Minnzen

Reviewed commit `2235b497cfb5`.

- **P1 · Legacy sub-DEX links fail with an empty universe cache**
- **P2 · Bare-logo fallback reintroduces cross-DEX icon collisions**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

packages/kit/src/views/Perp/components/usePerpTokenUrlSync.ts
Minnzen: Correct, and this one is on me — the existence check conflated "the coin is absent" with "the universe is not loaded yet", and a cold start opening a shared link is precisely when the cache is empty.

The part that makes it worse than a wash is the self-healing property you point out: before that commit the persisted atom held a valid `xyz:NVDA` that a later refresh could resolve. Downgrading it to `XYZNVDA` strands the user on a coin no refresh can ever resolve. So my change turned a recoverable cold-start race into an unrecoverable one, on the single most common landing path for a shared link.

Applied your suggestion in `c04fd4e007` — an incomplete cache is treated as inconclusive, same trade-off as the `catch` branch, and only a complete universe can prove absence:

```ts
const { universesByDex } =
  await backgroundApiProxy.serviceHyperliquid.getTradingUniverse();
// An unloaded or incomplete cache proves nothing, and a cold start opening
// a shared link is exactly when it is empty. Keeping the syntactic guess
// stays self-healing: the persisted coin is still valid, so a later refresh
// resolves it. Downgrading it here would strand the user on a coin that can
// never resolve.
if (!isPerpsUniverseCacheComplete(universesByDex)) {
  return coin;
}
const exists = universesByDex.some((assets) =>
  assets?.some((asset) => asset.name === coin),
);
return exists ? coin : urlToken.toUpperCase();
```

Reusing `isPerpsUniverseCacheComplete` also means this inherits the stricter definition landed earlier in the PR — every registered slot must carry data — so a `[main, xyz, []]` cache no longer counts as proof of absence either.

Worth noting the residual scope is now small: the ambiguity only bites a legacy separator-free link whose token collides with a main-DEX symbol, and no current main-DEX symbol starts with `xyz` or `para` (checked all 232). New links carry the separator and never take this path.

packages/kit/src/views/Perp/components/Portfolio/PerpPortfolioContent.tsx
Minnzen: Confirmed and fixed in `f59ff080ec`. This is a spot-side regression I introduced in the previous commit.

Reaching for `fillsStats.mostTraded` to recover the dex prefix inverted the existing precedence: `mostTradedTokenDisplayName` exists precisely because the raw fill coin needs `isSpotInstrument` → `spotPairDisplayMap` / `split('/')` → `getSpotTokenDisplayName` first. Bypassing it sends `@149` and `PURR/USDC` straight into the image path, and the slash in the latter does not just 404 — it breaks the path structure outright.

Took your suggestion verbatim; `isSpotInstrument` was already imported at line 39:

```tsx
tokenImageUris={
  // Only perps carry a dex prefix worth preserving. A spot `mostTraded` is
  // a raw fill coin (`@149`, `PURR/USDC`) and must keep going through the
  // resolved display name — the slash would otherwise break the image path
  // outright.
  fillsStats.mostTraded && !isSpotInstrument(fillsStats.mostTraded)
    ? getHyperliquidTokenImageUris(fillsStats.mostTraded)
    : [getHyperliquidTokenImageUrl(mostTradedTokenDisplayName)]
}
```

Worth recording why this slipped through: the sweep that added dex-scoped logos went call site by call site asking "is a full coin available here?", and this one answered yes without my checking what shape that full coin takes for spot. The other sites converted in that commit take their coin from a perps universe entry or `useActiveTradeDisplay`, so they cannot carry a spot raw form — I re-checked each one after your comment rather than assuming the pattern was safe elsewhere.

packages/kit/src/views/Perp/components/PositionShare/ShareContentRenderer.tsx
```diff
@@ -62,8 +63,11 @@ export function ShareContentRenderer({
   const isProfit = pnlBn.isGreaterThan(0);
   const pnlColor = isProfit ? colors.long : colors.short;
   const sideColor = side === 'long' ? colors.long : colors.short;
+  // `token` keeps the dex prefix; the display name does not, and a symbol
+  // listed on both the main DEX and a sub-DEX would otherwise share one file.
   const tokenImage =
-    tokenImageUrl || getHyperliquidTokenImageUrl(tokenDisplayName);
+    tokenImageUrl ||
+    getHyperliquidTokenImageUris(token || tokenDisplayName)[0];
```
sidmorizon: # [🟡 P2] 分享图对现货交易重现了刚在 Portfolio 修掉的 raw spot coin 坏链问题

## 问题

`IShareData.mode` 支持 `'spot'`,且现货分享的 `token` 是原始 fill coin:`PerpTradesHistoryList.tsx:263-274` 以 `mode: 'spot'`、`token: fill.coin`(即 `@149` / `PURR/USDC`)调用 `showPositionShare`,并且**不传** `tokenImageUrl`;`BalanceRow.tsx:82-93` 同样传 `token: item.rawCoin`(有 `logoURI` 兜底,缺失时同病)。本行的 `token || tokenDisplayName` 在现货场景必然选中原始 coin,`getHyperliquidTokenImageUris('@149')[0]` 产出 `.../hyperliquid/@149.png`,`'PURR/USDC'` 产出被 `/` 破坏路径的 `.../PURR/USDC.png`——都是必然 404 的坏链。`ShareImageGenerator.tsx:101-105` 是完全相同的改动。

新增单测只断言 `toHaveLength(1)`,没有断言 URL 可用性,所以恰好掩盖了这一点(`UETH` 因在 `SPOT_TOKEN_DISPLAY_MAP` 里能救回,`@149` / `PURR/USDC` 救不回)。

## 影响

改动前 `getHyperliquidTokenImageUrl(tokenDisplayName)` 用的是 `PerpTradesHistoryList.tsx:238-250` 精心解析出的现货 display 名(如 `HYPE`、`PURR`),图标正常;改动后从交易历史分享一笔现货成交,生成的分享图 token 图标变为坏链/fallback。这与上一个 commit 刚在 `PerpPortfolioContent`(`f59ff080e`)修复的回归是同一类问题,在另一个表面重新引入。

## 建议

沿用你在 `PerpPortfolioContent` 的同款分支:只有 perp 才用带前缀的完整 coin,现货保持 display name 路径(`ShareImageGenerator` 同步修改):

```tsx
const tokenImage =
  tokenImageUrl ||
  getHyperliquidTokenImageUris(
    mode !== 'spot' && token ? token : tokenDisplayName,
  )[0];
```

并建议给单测补一条断言现货 raw 形态产出的 URL 不含 `@` / `/` 的用例,防止再回归。

[Claude session ↗](https://claude.ai/code/session_016Xc2kQZXicru9Ga4JR9ZqF)

---

Minnzen: Confirmed and fixed in `8666bdadc5`. Same class of regression as `f59ff080e`, reintroduced on a second surface one commit later — that repetition is the more useful signal here.

Verified the caller rather than reasoning from the type: `PerpTradesHistoryList.tsx:263-274` passes `mode: 'spot'`, `token: fill.coin` and no `tokenImageUrl`, and the `tokenDisplayName` cascade immediately above it (lines 238-250) exists precisely because `parseDexCoin` cannot resolve spot forms. I bypassed that cascade in both share paths.

```tsx
const tokenImage =
  tokenImageUrl ||
  getHyperliquidTokenImageUris(
    mode !== 'spot' && token ? token : tokenDisplayName,
  )[0];
```

`ShareImageGenerator.tsx` got the identical change.

**Your point about the test is the part worth keeping.** `toHaveLength(1)` asserted the wrong property — it only proved "not treated as a sub-DEX", which stays true while the URL is unusable, so the test actively masked the defect it looked like it covered. Replaced with an assertion on the URL, plus a companion case documenting the resolved form a caller is expected to pass:

```ts
it('produces an unusable path for a raw spot coin, so callers must resolve it', () => {
  const [rawIdUri] = getHyperliquidTokenImageUris('@149');
  const [rawPairUri] = getHyperliquidTokenImageUris('PURR/USDC');
  expect(rawIdUri).toContain('@149');
  expect(rawPairUri).toContain('PURR/USDC');

  expect(getHyperliquidTokenImageUris('PURR')).toEqual([
    'https://uni.onekey-asset.com/static/hyperliquid/PURR.png',
  ]);
});
```

Root cause of hitting this twice: my sweep asked "does this call site have the full coin?" and stopped there, when the question that matters is "what shape does that coin take for spot?". I have now re-checked every remaining converted site against that second question — `PerpTokenSelectorRow`, `FavoriteTokenItem`, `MoblieTokenSelector`, `PerpTokenSelector`, `MobilePerpMarket` all route spot through an explicit `isSpot`/`spotUniverse` branch, and `PerpTickerBarDesktop` / `atoms` / `usePopularTickers` take their coin from a perps universe entry, so none can carry a raw spot form.

originalix (commented):
@Minnzen

Reviewed commit `f59ff080ec31`.

- **P2 · Bare-logo fallback reintroduces cross-DEX icon collisions**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

originalix (commented):
@Minnzen

Reviewed commit `7d51f2046084`.

- **P2 · Bare-logo fallback reintroduces cross-DEX icon collisions**
- **P2 · Cold-start URL resolution commits an ambiguous coin**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

originalix (commented):
@Minnzen

## Automated code review found 3 issues to address

Reviewed commit `bb6892d766e7`.

- **P2 · Bare-symbol fallback can display another DEX's icon**
- **P2 · Cold-start URL disambiguation commits an unverified DEX guess**
- **P2 · Spot info panels build image URLs from raw instrument IDs**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

originalix (commented):
@Minnzen

Reviewed commit `53e56edb259d`.

- **P2 · Bare-symbol fallback can display another DEX's icon**
- **P2 · Cold-start URL disambiguation commits an unverified DEX guess**
- **P2 · Spot info panels build image URLs from raw instrument IDs**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12753#issuecomment-5187212765)

originalix (approved):

PR #12781, opened 2026-08-06
fix: split mobile TradingView indicator bar OK-57045

limichange:
[OK-57045](https://onekeyhq.atlassian.net/browse/OK-57045)

----------

## Summary

- split the mobile TradingView indicator quick bar into a fixed 184px main-indicator section and an independently scrollable sub-indicator section
- keep the center divider fixed while preserving the existing single-scroll layout for non-mobile consumers
- forward vertical quick-bar gestures to the Market Detail collapsible page without rerouting horizontal indicator scrolling

## Root cause

The mobile quick bar was rendered as one horizontal `ScrollView`. After splitting the bar, the right-side horizontal scroller still consumed vertical touches, and the prior vertical-scroll bridge was not present on the current `x` lineage.

## User impact

Users can scroll the Market Detail page vertically when starting the gesture over the indicator bar, while horizontal gestures continue to scroll only the right-side sub indicators. The four left-side main indicators and divider remain fixed.

## Validation

- `yarn jest packages/kit/src/components/TradingView/TradingViewV2/components/indicatorControls/NativeIndicatorControls.test.tsx --runInBand --verbose`
- verified on the iOS simulator in the real Market Detail screen that a vertical swipe over the right indicator section moves the outer page

[OK-57045]: https://onekeyhq.atlassian.net/browse/OK-57045?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/components/TradingView/TradingViewV2/components/indicatorControls/NativeIndicatorControls.test.tsx
+++ b/packages/kit/src/components/TradingView/TradingViewV2/components/indicatorControls/NativeIndicatorControls.test.tsx
@@ -1,0 +1,205 @@
+/**
+ * @jest-environment jsdom
+ */
+
+import type { ReactNode, TouchEvent } from 'react';
+
+import { fireEvent, render, screen, within } from '@testing-library/react';
+
+import { TradingViewNativeIndicatorQuickBar } from './NativeIndicatorControls';
+
+import type { ITradingViewNativeIndicatorState } from './hooks/useNativeIndicatorActiveValues';
+import type { ITradingViewNativeChartControlsConfigData } from '../../types';
+
+jest.mock('@onekeyhq/components', () => {
+  const React = jest.requireActual<typeof import('react')>('react');
+
+  type IMockTouchHandler = (event: {
+    nativeEvent: { pageX: number; pageY: number };
+  }) => void;
+
+  function forwardTouchEvent(handler?: IMockTouchHandler) {
+    if (!handler) {
+      return undefined;
+    }
+
+    return (event: TouchEvent<HTMLDivElement>) => {
+      const touch = event.touches[0] ?? event.changedTouches[0];
+      handler({
+        nativeEvent: {
+          pageX: touch?.pageX ?? 0,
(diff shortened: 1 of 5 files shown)


originalix:

@limichange

Reviewed commit `3fb92636b693`. This report always reflects the latest reviewed changes.

### Review summary

The PR introduces a mobile-native interaction path for TradingView quick-bar indicators by splitting main/sub indicator sections and forwarding vertical swipe deltas from the quick bar into the market tab scroll container. This changes ownership of a gesture path: quick-bar touch events are now converted into shared-scroll updates through `useMobileTabTouchScrollBridge`, while control rendering remains inside `TradingViewV2` and is injected in `MarketLayout` for mobile only. The functional flow narrows and reuses existing native tab context primitives instead of adding new navigation code, so the change is focused on UI interaction and scroll behavior integration. Adjacent desktop/web control flows remain unchanged, and there is no new backend or cross-package data-sharing change.

**What needs attention:** Verify the new quick-bar-to-tab scroll math at the native boundary, especially when the tab is at top offsets (`scrollYCurrent <= tabContentInset`).

### Issues to address

- **P2 · Preserve content-inset baseline when translating quick-bar deltas**
  - **Impact:** When the active tab is near the top, `scrollYCurrent` can be below `tabContentInset`. The updated bridge now subtracts `tabContentInset` directly from `scrollYCurrent + deltaY`, so small vertical gestures can become non-positive and be clamped by the scroll view, making the new quick-bar vertical swipe path feel unresponsive until the tab is already scrolled down.
  - **Suggested action:** Restore a lower-bounded baseline when computing the target Y (or clamp before `scrollTo`) so that quick-bar vertical drags remain effective across top-edge states.

### Validation gaps

- No manual or automated runtime verification for mobile quick-bar drag behavior was run in this review context.
- Edge-state behavior (especially very-top positions) was not confirmed on device with real tab context values.

originalix (approved):

PR #12783, opened 2026-08-07
fix: load sparse market kline history OK-58202

limichange:
[OK-58202](https://onekeyhq.atlassian.net/browse/OK-58202)

----------

## Summary

- accept and validate TradingView's `countBack` value in market history requests
- retain older sliced candles and backfill before `from` until the requested bar count is satisfied
- keep the previous `[from, to]` behavior for chart builds that do not send `countBack`
- add focused handler and sparse-history regression tests

## Root cause

The chart bridge did not forward `countBack`, and the app filtered merged slice data back to `[from, to]`. For sparse 1m ranges, the exact viewport window could be empty even though earlier candles had already been fetched. TradingView then received an empty response and incorrectly treated it as the end of server history.

## Impact

AAVE 1m and other sparse market series continue loading historical candles while users pan left.

## Companion PR

- TradingView bridge: https://github.com/OneKeyHQ/tradingview-charting-library/pull/206

## Validation

- focused Jest suites — 2 suites, 7 tests passed
- local AAVE 1m verification — 8 consecutive left-pan requests continued loading and each response matched `countBack`
- diff whitespace checks passed

[OK-58202]: https://onekeyhq.atlassian.net/browse/OK-58202?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.skillshare/skills/1k-tradingview-communication/SKILL.md
+++ b/.skillshare/skills/1k-tradingview-communication/SKILL.md
@@ -77,6 +77,7 @@ Chart requests history with:
     resolution,
     from,
     to,
+    countBack,
     firstDataRequest,
   },
 }
--- a/packages/kit/src/components/TradingView/TradingViewV2/components/tradingViewV2/messageHandlers/klineDataHandler.test.ts
+++ b/packages/kit/src/components/TradingView/TradingViewV2/components/tradingViewV2/messageHandlers/klineDataHandler.test.ts
@@ -33,8 +33,10 @@ const mockFetchTradingViewV2DataWithSlicing =
 
 function buildHistoryMessage({
   firstDataRequest,
+  countBack = 300,
 }: {
   firstDataRequest: boolean;
+  countBack?: number;
 }): ICustomReceiveHandlerData['data'] {
   return {
     scope: '$private',
@@ -45,6 +47,7 @@ function buildHistoryMessage({
       resolution: '1',
       from: 1000,
       to: 2000,
+      countBack,
       firstDataRequest,
     },
   };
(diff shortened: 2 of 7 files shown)


originalix:

@limichange

Reviewed commit `0e8f878e47c5`. This report always reflects the latest reviewed changes.

### Review summary

This PR updates the TradingView V2 market-history bridge so the app accepts `countBack` from chart history requests and tries to reuse earlier candles when building the returned `kLineData` payload. The implementation stays inside the existing `kit` TradingView path: request types, message handling, K-line normalization, and tests are updated, while the background market-service contract and non-V2 providers remain unchanged. The scope is disciplined, but the data-flow ownership is still split between the raw request window and the actual returned history span, which leaves one incomplete history-backfill path and one marks-alignment regression.

**What needs attention:** Verify the end-to-end history contract: the candle fetch window and the marks window both need to follow the actual candle span returned to TradingView, not just the raw `from`/`to` request.

### Issues to address

- **P1 · countBack does not expand the fetch window**
  - **Impact:** The new history path still builds slices from the raw `timeFrom`/`timeTo` request and only applies `countBack` after data has already been fetched. If the market API returns exactly the requested window, there are no pre-`from` candles available to backfill, so TradingView can still receive too little history on the paths this change is trying to fix.
  - **Suggested action:** Use `countBack` when deriving the upstream history window, or perform an additional earlier fetch until the merged response reaches the requested bar count. Add coverage for an exact-window/native-token request where the service does not return any extra leading candles.
- **P2 · Backfilled candles do not receive matching account marks**
  - **Impact:** Once the history payload can include candles before the original `from`, the marks request and the cached `marksTimeRange` still stay on the raw request window. That leaves visible backfilled candles without their buy/sell markers and makes later mark refreshes keep the truncated range.
  - **Suggested action:** Derive the marks fetch range from the actual returned candle span, or expand `marksTimeRange.min` together with the backfilled history range before sending `MARKS_UPDATE`.

### Validation gaps

- `yarn jest packages/kit/src/components/TradingView/utils/fetchMarketKLineData.test.ts packages/kit/src/components/TradingView/TradingViewV2/components/tradingViewV2/messageHandlers/klineDataHandler.test.ts --runInBand` could not run in this workspace because Yarn reported `Couldn't find the node_modules state file`.
- I did not run a live TradingView V2 history-load check on native/web after the protocol change.

PR #12784, opened 2026-08-07
fix: show TradingViewNative volume axis

limichange:
## Summary

- render adaptive Y-axis grid lines and formatted labels in the volume pane
- show the corresponding formatted volume label when the crosshair is in the volume pane
- include volume labels when calculating the shared price/volume axis width
- cover the shared native and web scene with layout, formatting, and rendering tests

## Root cause

TradingViewNative only generated Y-axis ticks for the price pane. Its crosshair also mapped Y coordinates exclusively through the price scale, and the axis width calculation did not account for volume labels. As a result, the volume pane had neither scale information nor a crosshair value label.

## Impact

- iOS/Android: updates the main UI runtime's shared chart scene; no background runtime or native resource ownership changes
- desktop/web: updates the same chart behavior in the single JS runtime
- browser extension: no separate background behavior or persisted data changes

## Verification

- `npx jest --runInBand --runTestsByPath packages/kit/src/components/TradingView/TradingViewNative/utils/chartLayout.test.ts packages/kit/src/components/TradingView/TradingViewNative/utils/chartLegend.test.ts packages/kit/src/components/TradingView/TradingViewNative/utils/chartScene.test.ts` (60 tests passed)
- verified the volume-axis labels in the iOS simulator


originalix:

@limichange

Reviewed commit `da4703838b98`. This report always reflects the latest reviewed changes.

### Review summary

This PR extends the shared TradingView Native rendering pipeline so the right axis can represent the volume pane in addition to the price pane. Shared layout now emits volume ticks and volume-at-cursor mapping, while both native and web renderers reserve axis width for volume labels and switch the crosshair label from price to volume inside the volume section. The work stays inside the existing shared chart layout/legend/scene boundary and keeps native/web behavior aligned, but the new width heuristic samples too few formatted volume values and can still undersize the axis for low-volume ranges, leading to clipped labels in the new UI path. Security impact is neutral; the main risk is a bounded chart rendering regression on affected markets.

**What needs attention:** Verify the shared contract between volume formatting, axis-width reservation, and crosshair rendering on low-volume charts, because that path now drives both native and web output.

### Issues to address

- **P2 · Volume axis width underestimates low-volume crosshair labels**
  - **Impact:** The new right-axis width reserve is derived from only five sampled volume values plus a fixed fallback string. On low-volume ranges, the crosshair can still format reachable sub-1 values such as `0.0197368`, which are longer than every sampled candidate, so the shared width calculation can be too small and the newly added volume label will clip on both native and web charts.
  - **Suggested action:** Make the reserved volume label account for the longest plain-decimal regime that the visible volume scale can emit, and add a regression test that covers low-volume charts where `maxVolume < 1` or only a few whole units are visible.

### Validation gaps

- `yarn jest packages/kit/src/components/TradingView/TradingViewNative/utils/chartLayout.test.ts packages/kit/src/components/TradingView/TradingViewNative/utils/chartLegend.test.ts packages/kit/src/components/TradingView/TradingViewNative/utils/chartScene.test.ts --runInBand` could not run because Yarn reported `Couldn't find the node_modules state file`.
- Existing GitHub review/comment coverage was not available from the workspace evidence, so any inline comment should be checked for duplication before posting.

originalix (approved):

PR #12787, opened 2026-08-07
fix: stop misreporting IndexedDB connection loss as a full disk

sidmorizon:
## Problem

An extension user on Win11 with ~50 GB free on the Chrome profile volume was shown "System Disk is full" repeatedly. Their log shows 22 `SystemDiskFullError`s over 24 minutes, all from `background.bundle.js`, with the wallet list failing to load the whole time. They dismissed the warning dialog twice (`disk-full-warning-confirm-btn` at 12:51:49 and 13:14:53) and it kept coming back.

Runtime scope: the flag lives in the **ext background** runtime only. `$onekeySystemDiskIsFull` is a `globalThis` variable, so it is per-JS-runtime, while the IndexedDB underneath is a shared native resource. Reopening the popup does not clear it — the log shows the UI restarting at 13:14 with the error still latched from 12:51.

## Root causes

**1. A connection-lifecycle error was matched as a disk-full signal.**

`handleDiskFullError` treated `Failed to execute 'transaction' on 'IDBDatabase': The database connection is closing` as "disk is full". That is an `InvalidStateError` raised when the connection is gone — a `versionchange` from another runtime, a browser-forced close, storage bucket eviction — and says nothing about free space.

It became permanent because `open()` caches one `IDBDatabase` and reuses it for every transaction, but only our own `close()` reset that cache. A browser-forced close left a dead handle behind, so every later `transaction()` threw the same error.

**2. The guard could never clear itself.**

`$onekeySystemDiskIsFull` was only ever set to `true`, never back to `false`. `checkIfDiskIsFull` also called `checkIfDiskIsFullSync()` before measuring, so once the flag was up the one path that could have cleared it stopped running.

**3. The guard blocked the recovery path.**

`WebStorage.removeItem` / `clear` deliberately skip `checkDiskFull`, but the IndexedDB shim blocked on `mode !== 'readonly'`, which catches deletes too. The exact failure in the user's log:

```
ServiceAccount.getAllWallets
 -> ServiceKeylessCloudSync.syncPersistedCurrentCloudSyncKeylessWalletIdWithWallets
   -> removeAllCredentials -> storageRemoveItem -> WebStorage.removeItem
     -> IndexedDBPromised.delete -> createBucketTransaction -> IDBDatabase.transaction
       -> checkIfDiskIsFullSync   // throws
```

Freeing space was blocked by "out of space".

## Changes

**Classification**

- Match only genuine disk/quota signals (`QuotaExceededError`, `Encountered disk full`, `Encountered full disk`), against `name` + `message` combined — a spec-shaped quota failure is a `DOMException` whose `name` carries the type while the message is just "The quota has been exceeded."
- A closing connection is no longer a disk-full signal. `open()` registers `close` / `versionchange` handlers that drop the cached handle, and `createBucketTransaction` reopens and retries once.
- Deduplicate in-flight `open()` calls. Concurrent callers each opened their own connection and only the last was kept; the orphans stayed open and blocked exactly the `versionchange` upgrades this bug is about. `open()` prefers the in-flight promise over the cached handle, since `onupgradeneeded` publishes the handle early for the upgrade callback.

**Recovery**

- The guard clears once measured headroom recovers, with hysteresis so it cannot flap. Both thresholds derive from the granted quota, so the clear condition stays reachable on small quotas and the raise threshold does not mark an almost-empty small-quota origin as full.
- `checkIfDiskIsFull` no longer throws before measuring, and `checkIfDiskIsFullSync` schedules the measurement before rejecting — a blocked write never reaches the shim that would otherwise schedule it.
- The debounce carries `maxWait`, so sustained write traffic cannot push the measurement forward indefinitely. This is the reported user's pattern: Hyperliquid L2 cache writes every few hundred ms.
- Web wallet mode joins the measurement branch; previously only extension and desktop could clear a guard they were able to raise.
- Releasing a `writeFailed` guard additionally requires a small write-then-delete probe to commit. A write failure is itself evidence the quota estimate is not predictive, so clearing that state on the estimate alone let the next write hit the same failure and re-raise.

**Keeping the recovery path open**

- The blocking guard moved from the IndexedDB shim into `createBucketTransaction`, which can tell a space-consuming write from a space-freeing delete. The shim still triggers measurement on every write-mode transaction.
- `delete` / `clear` pass `allowWhenStorageFull`, and the option is threaded through `ILocalDBWithTransactionOptions` so `removeRecords` and `clearRecords` reach it — wallet, account and history deletion is the main user-driven way to free OneKey data.
- The `resetUtils.getIsResetting()` bypass moved into `checkIfDiskIsFullSync` so every call site gets it.
- `getOrCreateCryptoKey` resolves an existing key through the readonly path before applying the write guard; returning a persisted key consumes no storage, and blocking it turned a readable sealed session into repeated decrypt failures.

**Coverage of raw paths**

- `IndexedDBPromised.transactionAsync()` added for ordinary bucket work; DB backup and bucket migration moved onto it so they inherit the guard and the reopen-retry. Sync `transaction()` stays for `versionchange`.
- Raw CryptoKey writes report async quota failures back into the state machine, which they previously bypassed.

**Diagnostics**

- `ShowSystemDiskFullWarning` carries quota/usage, and the dialog renders it.
- Measurement, detection and clearing log through `defaultLogger`.

## Surfacing the actual quota

The quota check itself is unchanged in intent — for an extension the quota, not the physical disk, is the binding write limit, so checking headroom is correct. What was missing is any way to tell a real quota exhaustion apart from a misclassified failure. The dialog now shows:

```
Quota 38.0 GB · Used 37.5 GB · Free 0.500 GB
Reason: quotaExhausted
```

Sub-100 MB headroom keeps 3 decimals so a real exhaustion does not collapse to "0.0 GB". `writeFailed` also shows the originating error message.

Previously `handleDiskFullError` was completely silent and the quota numbers went only to `console.log`, which never reaches the exportable log — so a submitted log could never show which path fired or what the numbers were.

## Testing

```
Test Suites:  9 passed,  9 total
Tests:       83 passed, 83 total      (packages/shared/src/storageChecker + storage)

Test Suites: 12 passed, 12 total
Tests:      135 passed,  2 skipped    (packages/kit-bg/src/dbs/local)
```

New suites: `storageChecker.test.ts`, `storageChecker.web.test.ts`, `indexedDbCryptoKeyStore.test.ts`. Covered: a closing connection is not disk-full; genuine quota failures still raise, including a real `DOMException`; the reset bypass; raising and clearing with hysteresis; a quota below the fixed floor staying writable while still entering and leaving the warning band; recovery under sustained write traffic; a write-failure guard not clearing on a healthy estimate while a quota-derived one does; and async quota failures on the raw CryptoKey path.

## Notes for review

- localDb reads pass `readOnly: true`, so reads were never blocked by the guard — the `getAllWallets` failure came purely from the delete path.
- The diagnostic detail labels are deliberately untranslated. No `ETranslations` keys exist for "quota" / "used" / the reason values, and the locale files are generated and must not be hand-edited, so those keys need to be added upstream first. `global_available` is the one that exists and is used. Small follow-up once the rest land.
- Not covered by tests: a fake-indexeddb run driving `removeRecords` with the guard raised, and concurrent `open()` / `versionchange` invalidation. Both need harness setup beyond what those suites have today.


packages/shared/src/storageChecker/storageChecker.ts
```diff
@@ -11,75 +11,176 @@ import appGlobals from '../appGlobals';
...
-    globalThis.$onekeySystemDiskIsFull = true;
-    appGlobals?.$appEventBus?.emit(
-      EAppEventBusNames.ShowSystemDiskFullWarning,
-      undefined,
-    );
+  const errorMessage = getErrorMessage(error);
+  if (!errorMessage) {
+    return;
+  }
+  if (!diskFullErrorMessages.some((message) => errorMessage.includes(message))) {
```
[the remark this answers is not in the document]
sidmorizon: Confirmed and fixed in 481ecf87ad. `handleDiskFullError` now matches against `name` + `message` combined (`getErrorText`), so a spec-shaped `DOMException('The quota has been exceeded.', 'QuotaExceededError')` raises the guard. Added a test using a real `DOMException` for exactly that shape. Connection-closing detection intentionally stays message-only, since its `name` (`InvalidStateError`) is too broad to match on.

---

sidmorizon:
Both findings addressed in 3449f5384a.

**P1 (raw readwrite paths bypass the guard):** re-applied `checkIfDiskIsFullSync()` explicitly on the raw write-intent paths — `writeCryptoKeyRecord`, `getOrCreateCryptoKey`, and the v4-migration transaction builder — so they fail fast with the typed `SystemDiskFullError` instead of an opaque aborted transaction. `deleteCryptoKeyRecord` deliberately stays unguarded, matching the delete/clear allow-path contract: deletes free space and must remain possible while the guard is raised. Blocking at the shared shim boundary was rejected on purpose — `IDBDatabase.prototype.transaction` cannot distinguish a space-consuming put from a space-freeing delete (a `readwrite` mode check catches both), which is exactly how the guard previously blocked its own recovery path.

**P2 (half-open connection during upgrade):** `open()` now checks the in-flight `openPromise` before the cached handle. `onupgradeneeded` must still publish `nativeDB` early (the upgrade callback's `this.transaction()` depends on it), so concurrent callers now await `onsuccess` instead of picking up the half-open connection while the versionchange transaction is running.

Verified: full-repo oxlint (CI mode) exit 0, 13/13 unit tests, tsc clean.

---

originalix (commented):
@sidmorizon

Reviewed commit `3449f5384a4b`.

- **P1 · Some bucket write paths still bypass the new storage guard contract**
- **P2 · Clear threshold can become unreachable on small quotas**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
Both round-2 findings addressed in 05f15400df.

**P1 (backup / bucket-migration writes bypass the guard):** took the centralization option — the flagged call sites (`ServiceDBBackup`, `indexedToBucketsMigration`) all go through the sync `IndexedDBPromised.transaction()`, so the contract now lives there: `readwrite` runs `checkIfDiskIsFullSync()`, and a closing-connection error drops the dead cached handle so the next `open()` recovers (a sync API cannot reopen in place). `versionchange` stays exempt — schema upgrades must never be blocked. No call-site changes needed, and future sync-path callers inherit the contract.

**P2 (clear threshold unreachable on small quotas):** correct catch — the fixed `2×warning` (~1.87 GB) threshold could exceed a small quota entirely, turning the self-healing guard back into a latch. `getClearAtBytes(quota)` now returns 2× the warning floor for large quotas and the midpoint between the floor and the quota for small ones, so the clear condition is always reachable whenever the quota is above the warning floor. A quota at or below the floor keeps the guard latched deliberately: storage that small cannot hold the app's data. Added a 1.5 GB-quota regression test.

Verified: 12/12 unit tests, oxfmt/oxlint/tsc clean locally.

---

originalix (commented):
@sidmorizon

Reviewed commit `05f15400df67`.

- **P1 · Storage guard cannot observe recovery after it starts blocking**
- **P1 · Sync transaction callers remain dead after connection invalidation**
- **P2 · Existing CryptoKeys become unreadable while storage is full**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
All three round-3 findings confirmed and fixed in 7a6a1c9f19.

**P1 · Guard cannot observe recovery** — real, and it defeated the whole point of making the flag clearable. Traced it: `createBucketTransaction` runs `checkIfDiskIsFullSync()` *before* `nativeDB.transaction()`, so a blocked write never reaches the shim that schedules the measurement. Reads don't help either (the shim only schedules on `isWriteMode`). So absent an allowed delete, nothing ever re-measured. `checkIfDiskIsFullSync()` now schedules the debounced measurement before throwing. No recursion risk — `checkIfDiskIsFull` no longer calls the sync guard.

**P1 · Sync callers dead after invalidation** — took the suggested direction. Added `IndexedDBPromised.transactionAsync()` (delegates to `createBucketTransaction`, so it inherits both the guard and the reopen-retry) and moved the backup + bucket-migration callers onto it. Sync `transaction()` stays for the `versionchange` use case the class documents.

While converting I found two call sites the report didn't list — `removeBackupHDWallet` and `removeBackupImportedAccount` (`ServiceDBBackup.ts:61,116`). Both open `readwrite` transactions that only ever `delete()`, so the guard I added last round would have wrongly blocked them. They now pass `allowWhenStorageFull: true`, matching the delete/clear allow-path contract.

**P2 · Existing CryptoKeys unreadable while full** — correct. `getOrCreateCryptoKey()` now resolves an existing key through `readCryptoKeyRecord()` (readonly) and only applies the write guard when the key is absent. The readwrite transaction still does its atomic get-then-put re-check, so the documented cross-runtime race safety is unchanged.

On the validation gaps: the new test drives the production sequence you asked for — raise the guard, mock healthy headroom, call the guarded path, advance the debounce, assert the flag cleared. It needs `jest.isolateModulesAsync` because earlier tests arm the module-level lodash debounce under real timers, and lodash then holds a timer id that `useFakeTimers` has already replaced.

Verified: 73/73 tests across `storageChecker` + `storage`, full-repo oxlint exit 0, oxfmt and tsc clean.

---

originalix (commented):
@sidmorizon

Reviewed commit `7a6a1c9f19b9`.

- **P1 · Web wallet storage guard can never recover**
- **P1 · WebStorage retries bypass recovery measurement**
- **P1 · Primary local DB deletions remain blocked while full**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
All three round-4 findings confirmed and fixed in 447809fed8.

**P1 · Web wallet guard can never recover** — correct. `isWebDappMode` is false in web wallet mode, so `handleDiskFullError` can latch the guard there, but the measurement branch was gated on `isExtension || isDesktop`, so nothing could ever clear it. Web now joins that branch; web DApp mode still returns at the top of the function. Added a dedicated web-mode suite covering raise and clear.

**P1 · WebStorage bypasses the re-measurement** — correct, and it matters most because this is the main app-storage write path. `checkDiskFull()` was re-implementing the raised-state throw, which was fine until last round made `checkIfDiskIsFullSync()` the owner of scheduling the recovery measurement. It now delegates, which also puts diagnostics forwarding in one place instead of two. Added a test asserting the delegation.

**P1 · Primary local DB deletions blocked** — correct, and this was the most consequential of the three: `IndexedDBAgent` builds one generic `readwrite` transaction for every local DB operation, so wallet/account/history removal could not even start while the guard was raised. That closed off the main user-driven way to free OneKey data — the same dead-end this PR exists to remove, just one layer up. Added `allowWhenStorageFull` to `ILocalDBWithTransactionOptions`, threaded it through `withTransaction` → `_buildTransactionAndStores` → `createBucketTransaction`, and set it on `removeRecords` and `clearRecords`. Ordinary writes stay guarded.

On the remaining validation gaps: the 137-test `packages/kit-bg/src/dbs/local` suite passes with the threaded option, which covers the transaction-layer plumbing. I did not add a fake-indexeddb test that drives `removeRecords` with the guard raised, nor a concurrent-`open()` / `versionchange` invalidation test — both need harness setup beyond what exists in those suites today. Flagging rather than implying coverage.

Verified: 76/76 in `storageChecker` + `storage`, 135 passed / 2 skipped in `dbs/local`, full-repo oxlint exit 0, oxfmt and tsc clean.

---

originalix (commented):
@sidmorizon

Reviewed commit `447809fed85b`.

- **P1 · Continuous writes can starve recovery measurement**
- **P1 · Small web quotas permanently trip the storage guard**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
Both round-5 findings confirmed and fixed in 9b7a3790be.

**P1 · Continuous writes starve recovery measurement** — real, and notably it is exactly the reported user's situation: their log shows Hyperliquid L2 snapshot cache writes every few hundred milliseconds. With a trailing-only debounce and no `maxWait`, each blocked write pushed the timer forward, so the measurement would never have run for them even after freeing space. Added `maxWait: 1000`.

**P1 · Small web quotas permanently trip the guard** — real, and self-inflicted last round: the 0.936 GB floor was calibrated against extension-sized quotas, and extending measurement to web exposed it to browsers that grant less than that. `getWarningAtBytes(quota)` now takes the smaller of the floor and a fraction of the quota. That keeps the threshold continuous and proportional rather than introducing a cliff at the boundary, and leaves quotas at or above ~9.4 GB (every extension/desktop case, including the 38 GB figure in this file's own header) unchanged. `getClearAtBytes` derives from it, so the hysteresis band stays reachable at any quota.

Both requested tests added: sustained 400 ms write traffic recovering through `maxWait`, and a 0.5 GB quota staying writable while empty yet still entering and leaving the warning band.

One note on the first test: asserting `toThrow` on every iteration failed, because the guard clears partway through the loop — which is the behavior under test. It now asserts the first blocked write throws, then keeps the traffic going and asserts the guard cleared.

Verified: 80/80 in `storageChecker` + `storage`, full-repo oxlint exit 0, oxfmt and tsc clean.

---

originalix (commented):
@sidmorizon

Reviewed commit `9b7a3790be4a`.

- **P2 · Raw CryptoKey writes bypass quota-failure state updates**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
Confirmed and fixed in 56a473c040.

The raw CryptoKey paths use their own `requestToPromise` / `transactionDone` helpers, which reject straight to the caller. The IndexedDB shim's catch only covers the *synchronous* `db.transaction()` call, and `IndexedDBTransactionPromised` — the thing that normally funnels quota failures into the state machine via its `error`/`abort` listeners — is not involved. So an async `QuotaExceededError` on `store.put()` propagated without raising the guard or producing diagnostics.

Both write paths (`writeCryptoKeyRecord`, `getOrCreateCryptoKey`) now wrap their async request + transaction completion, pass the failure to `handleDiskFullError`, and rethrow. `readCryptoKeyRecord` and `deleteCryptoKeyRecord` stay unwrapped — they consume no storage, and a delete must never raise the guard.

Added the focused test you asked for: a stub `IDBFactory` whose `put` fails asynchronously with a real `DOMException('The quota has been exceeded.', 'QuotaExceededError')`, asserting the guard is raised and the error still reaches the caller. It needs an `appStorage` mock, because the DB-open path fires a best-effort persistent-storage request that would otherwise instantiate the real WebStorage/IndexedDB stack.

Verified: 81/81 in `storageChecker` + `storage`, full-repo oxlint exit 0, oxfmt and tsc clean.

---

originalix (commented):
@sidmorizon

Reviewed commit `56a473c04018`.

- **P1 · Quota estimates can clear a real write-failure guard**
- **P2 · Disk diagnostics bypass localization**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5214356213)

sidmorizon:
Both round-7 findings addressed in c192ec4ddf.

**P1 · Quota estimates can clear a real write-failure guard** — confirmed, and the asymmetry you identified is the right frame: `QuotaExhausted` is *derived from* the estimate, so releasing it via the estimate is sound; `WriteFailed` is evidence the estimate is **not** predictive, so releasing it via the estimate is not.

Implemented the probe you suggested: `probeWriteCanCommit()` does a write-then-delete against a dedicated `OneKeyStorageProbe` database. It never throws, always closes, and leaves no data behind. Quota-derived state clears through hysteresis unchanged; write-failure state additionally requires the probe to commit. The probe only runs when a `WriteFailed` guard is up *and* the estimate already looks healthy, so it stays off the hot path.

Two regression tests, both requested: a write failure with a healthy subsequent estimate does **not** clear, and — as the contrast case — a quota-derived guard still does.

**P2 · Diagnostics bypass localization** — took the structural half. Composition moved out of `shared` into `DiskFullWarningDialogContainer`, so the UI layer owns presentation; `storageQuotaUtils` keeps only locale-independent numeric formatting.

I did not fully localize the labels, and want to be explicit about why rather than imply coverage. There are no `ETranslations` keys for "quota", "used", or the reason values — I checked; only `global_available` exists, and that one is now used. The locale files are generated and this repo's rules forbid hand-editing them, so the missing keys have to be added upstream in Lokalise before they can be wired here. The remaining labels are therefore deliberately technical, which also keeps them stable for users copying them into support reports. Happy to wire them the moment the keys land — that's a small follow-up.

Verified: 83/83 in `storageChecker` + `storage`, full-repo oxlint exit 0, oxfmt and tsc clean.

---

packages/shared/src/storageChecker/storageChecker.ts
```diff
@@ -11,75 +11,314 @@ import appGlobals from '../appGlobals';
...
+          tx.onabort = () => finish(false, db);
+        } catch {
+          finish(false, db);
+        }
+      };
+    } catch {
+      finish(false);
+    }
+  });
+}
```
sidmorizon: **Storage check keeps re-running once per second forever while storage is genuinely full**

Severity: `non-severe`

The storage self-check performs a tiny test write (`probeWriteCanCommit()` at `packages/shared/src/storageChecker/storageChecker.ts:130-178`) that itself re-arms the very timer that triggers the next check, so while storage really is exhausted the check repeats endlessly about once a second.

Impact: While a user's storage is full, the app runs a permanent background loop that repeatedly opens a database and keeps the extension background worker awake, wasting battery and CPU until storage recovers or the app restarts.

<details>
<summary>Feedback loop between the write probe and the IndexedDB shim's debounced measurement</summary>

1. `checkIfDiskIsFull()` only calls `probeWriteCanCommit()` when the guard was raised by a real write failure (`lastDiagnostics.reason === EStorageFullReason.WriteFailed`, `packages/shared/src/storageChecker/storageChecker.ts:300-307`).
2. `probeWriteCanCommit()` executes `db.transaction(probeStoreName, 'readwrite')`.
3. That call goes through the patched `IDBDatabase.prototype.transaction` in `packages/shared/src/polyfills/indexedDBShim/indexedDBShim.ts:19-31`, which for any write-mode transaction schedules `void storageChecker.checkIfDiskIsFullDebounced()`.
4. The debounced measurement fires ~1s later, finds the guard still raised with `WriteFailed`, probes again, and re-schedules itself — indefinitely, with no external write traffic required.

The loop terminates only when the probe finally commits (guard cleared) or the runtime restarts. Excluding the probe database from the shim's measurement scheduling, or probing without going through the patched prototype, would break the cycle.

</details>

packages/shared/src/storageChecker/storageChecker.ts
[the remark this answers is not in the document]
sidmorizon: Confirmed and fixed in fe661b2a39 — the feedback loop is exactly as described, and I introduced it with the probe last round.

Traced it: `runWriteProbe` → `db.transaction(probeStoreName, 'readwrite')` → patched `IDBDatabase.prototype.transaction` → schedules the debounced measurement → measurement finds the guard still `WriteFailed` → probes again → re-schedules. Self-sustaining at ~1/s with zero app activity, and on MV3 it would keep the background worker from ever going idle.

Fix: all scheduling now goes through a single `scheduleCheckIfDiskIsFull()` choke point that skips while a probe is in flight, and the probe itself is non-reentrant. I chose this over excluding the probe database by name in the shim, because it holds regardless of how the probe is implemented — the invariant is "the probe never drives its own retry".

Recovery is unaffected: measurement stays driven by app activity. Blocked writes schedule it via `checkIfDiskIsFullSync`, and space-freeing deletes schedule it through the shim, so a user freeing space is still noticed.

Added a regression test with a fake IndexedDB whose transaction fails and calls the scheduler the way the shim does; it asserts the probe ran, the guard stayed raised, and no timer was left armed.

Verified: 84/84 in `storageChecker` + `storage`, full-repo oxlint exit 0, tsc clean.

---

packages/kit/src/provider/Container/DiskFullWarningDialogContainer/DiskFullWarningDialogContainer.tsx
```diff
@@ -4,14 +4,53 @@ import { debounce } from 'lodash';
...
+        `${availableLabel} ${storageQuotaUtils.formatGB(
+          quotaInfo.availableBytes,
+        )}`,
+      ].join(' · '),
+    );
+  }
+  lines.push(`Reason: ${diagnostics.reason}`);
+  if (diagnostics.errorMessage) {
+    lines.push(`Error: ${diagnostics.errorMessage}`);
+  }
```
sidmorizon: **Storage warning dialog shows hardcoded English text that is never translated**

Severity: `non-severe`

The extra diagnostic line under the storage warning is built from hardcoded English words (`Quota`/`Used`/`Reason:`/`Error:` in `buildDiagnosticsDetail` at `packages/kit/src/provider/Container/DiskFullWarningDialogContainer/DiskFullWarningDialogContainer.tsx:38-51`) instead of translation keys, so non-English users see untranslated text inside an otherwise localized dialog.

Impact: Users of every non-English locale get a mixed-language warning dialog.

<details>
<summary>Repository rule: no hardcoded user-facing strings</summary>

`.cursor/rules/locale-handling.mdc` (alwaysApply) states: "Always use translation keys instead of hardcoding text strings". Only the `available` label goes through `intl.formatMessage` (`ETranslations.global_available`); `Quota`, `Used`, `Reason: <reason>` and `Error: <message>` are literal English strings rendered into the dialog body via `renderContent`.

The code comment acknowledges that the needed keys do not exist yet. Since generated locale files must not be hand-edited, the fix is to add the keys upstream first, or to not render the labelled text until the keys exist (e.g. show only numeric/technical values behind a copy-diagnostics affordance).

</details>

packages/kit-bg/src/dbs/local/LocalDbBaseContainer.ts
```diff
@@ -274,12 +274,18 @@ export abstract class LocalDbBaseContainer implements ILocalDBAgent {
...
+      (tx) => {
+        return this.txRemoveRecords({
+          ...params,
+          tx,
+        });
+      },
+      // Removing records frees space, so it must stay available while the
+      // disk-full guard is raised.
+      { allowWhenStorageFull: true },
+    );
```
sidmorizon, line 309: **Deleting several accounts at once can still be refused while storage is reported full, leaving wallets with references to accounts that no longer exist**

Severity: `non-severe`

The follow-up wallet clean-up that runs right after the accounts are deleted is still routed through the ordinary transaction path (`withTransaction` at `packages/kit-bg/src/dbs/local/LocalDbBase.ts:7872`) instead of the space-freeing one used for the deletion itself, so with the storage-full guard raised the deletion half succeeds and the clean-up half is rejected.

Impact: A bulk account removal performed while the app believes storage is exhausted ends half-done — the accounts are gone but the wallet still lists them, and the user sees an error.

<details>
<summary>Why the guard rejects the second transaction and which other delete paths were missed</summary>

`removeAccountsByIds` now runs its record deletion inside `withSpaceFreeingTransaction` (cited lines), which sets `allowWhenStorageFull: true` (`packages/kit-bg/src/dbs/local/LocalDbBaseContainer.ts:151-158`). The second transaction at `packages/kit-bg/src/dbs/local/LocalDbBase.ts:7854-7871`, which strips the removed account ids from `wallet.accounts`, still uses plain `withTransaction`, so `createBucketTransaction` calls `storageChecker.checkIfDiskIsFullSync()` (`packages/shared/src/IndexedDBPromised/IndexedDBPromised.ts:232-234`) and throws `SystemDiskFullError`. The single-account path `removeAccount` performs the very same `txUpdateWallet` *inside* its space-freeing transaction (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:7929-7938`), so the two paths now behave differently.

Two other pure-delete transactions were also left unconverted by this migration and stay blocked while the guard is raised: `clearAllSyncItems` (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:4458-4464`) and `deleteHardwareHomeScreen` (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:8577-8584`).

</details>

sidmorizon:
Both findings confirmed and fixed in 928c588682.

**Deletion still refused while storage is full** — the important one, and the analysis is exactly right. The allowance I added last round sat on `removeRecords`, but the real user-driven flows in `LocalDbBase` never call it: they open their own `withTransaction(bucket, task)` with no options, so `createBucketTransaction` still threw. Removing a wallet or account remained blocked — the dead end this whole change set exists to remove, one layer further up than I'd looked.

Added `withSpaceFreeingTransaction()` on the container and moved ten flows onto it: wallet, indexed account (×2), account (×2), credential, cloud sync pool, and the archive-bucket signed message / signed transaction / connected site clears. Using a named helper rather than passing the option at each site keeps the intent visible at the call site and makes the next such flow obvious to classify.

One deliberate exclusion: `removeContextVerifyStringLocalSecretEnvelopeIfNeeded` keeps the plain guarded transaction. Despite the name it calls `txUpdateContext` — it writes, it doesn't free anything.

On the mixed-transaction caution you raised: I allowed them rather than splitting. `removeWallet`'s update (rewriting a wallet it converts to mocked) is a rounding error next to what the same transaction deletes, and refusing to start guarantees the failure that allowing it merely risks. Splitting removal from update would cost the atomicity that keeps a half-removed wallet from existing, which seems the worse trade.

**Hardcoded English in the dialog** — you're right, and the `.cursor/rules/locale-handling.mdc` citation settles it; my earlier "deliberately technical" reasoning doesn't survive an `alwaysApply` rule. Took your second option: the dialog no longer renders labelled English. It now shows only the translated available label with formatted sizes — available of granted quota, e.g. `可用: 0.500 GB / 38.0 GB`. That is the number that explains the warning to a user staring at a half-empty drive, and it is fully localized.

`reason` and the originating error message are gone from the dialog. They're diagnostics, not UI copy, and they already go to `defaultLogger` — which this PR added precisely so an exported log can show which path fired and what the numbers were. That's the right place for a support engineer to read them.

Verified: 219 passed / 2 skipped across `storageChecker`, `storage` and `dbs/local` (21 suites), full-repo oxlint exit 0, oxfmt and tsc clean.

---

sidmorizon:
All three findings from this round (both reports converge on the account one) confirmed and fixed in 70ccb95a24.

**Bulk account removal left half-done** — correct, and the observation that `removeAccount` already had the right shape while `removeAccountsByIds` didn't is what makes it clearly a bug rather than a judgement call. Both halves now share one `withSpaceFreeingTransaction`: the removal is allowed under the guard, and it commits atomically with the wallet-reference cleanup. Worth noting the inconsistency wasn't guard-specific — even with storage healthy, a failure between the two transactions left wallets pointing at deleted accounts.

**`clearAllSyncItems` and `deleteHardwareHomeScreen` missed** — both converted. Root cause of the miss is worth recording: my sweep scanned `async remove*` methods and took only the *first* `withTransaction` in each. That is also exactly why the second transaction in `removeAccountsByIds` slipped through. This round I audited every `withTransaction` call in `LocalDbBase` and classified each by what its body actually does, which is how these surfaced. Deliberately left guarded: `addHardwareHomeScreen` and `addConnectedSite` (they delete-then-add, net writes), `demoRemoveRecord1` (dev-only), and `removeContextVerifyStringLocalSecretEnvelopeIfNeeded` (updates a context record despite the name).

**Low-quota measurement erased write-failure provenance** — real, and it silently defeated the probe added two rounds ago: `WriteFailed` → low estimate rewrites the reason to `QuotaExhausted` → healthy estimate clears with `needsWriteProof` false. Refreshing a raised guard now keeps the `WriteFailed` reason and takes only the fresher `quotaInfo`. Scoped to `wasAlreadyFull` so a stale reason cannot leak into a fresh raise. Regression test added for the exact write-failure → low → healthy sequence with no successful probe.

Verified: 220 passed / 2 skipped across 21 suites (`storageChecker`, `storage`, `dbs/local`), full-repo oxlint exit 0, oxfmt and tsc clean.

Still not covered by tests, flagging rather than implying otherwise: a fake-indexeddb run driving `removeAccountsByIds` with the guard raised, and concurrent `open()` / `versionchange` invalidation.

---

originalix (commented):
@sidmorizon

Reviewed commit `70ccb95a2472`.

- **P1 · IndexedDB open failures bypass disk-full recovery**
- **P2 · Stale quota checks can clear a newer full state**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5240473093)

sidmorizon:
Both findings confirmed and fixed in da090e70cc.

**P1 · Open failures bypass recovery** — real, and the sharpest miss in the set so far: `QuotaExceededError: Encountered full disk while opening backing store for indexedDB.open` is written verbatim in `WebStorage.ts`'s own comments, so the codebase already documented this failure shape and I still only handled transaction-time failures. On a genuinely full disk the open is where it fails first, and nothing reported it — guard, warning and diagnostics all stayed unset.

`IndexedDBPromised.doOpen()` and `openCryptoKeyDb` now both report the open error to `storageChecker` before rejecting, and a failed open calls `invalidateNativeDB` so a handle `onupgradeneeded` had already published cannot stay cached. Regression test added driving a quota failure from the open itself.

**P2 · Stale measurement can clear a newer full state** — also real. `checkIfDiskIsFull` awaits the estimate and possibly the probe, and the scheduler can fire again meanwhile, so completion order was not guaranteed. Each measurement now takes a generation and may only raise or clear if it is still the newest after every async boundary. Test added where a slow healthy estimate resolves *after* a fast low-quota one and must not clear the guard.

Verified: 222 passed / 2 skipped across 21 suites, full-repo oxlint exit 0, oxfmt and tsc clean.

A note on where this has landed. Eleven rounds in, the findings have moved from the original defect out into progressively narrower edges of the state machine I built around it — and several were caused by earlier fixes in this same PR (the probe's self-sustaining loop, the provenance downgrade, the half-converted delete flows). That pattern suggests the recovery state machine is now the risky part of this change, not the original one-line misclassification.

If the goal is shipping a fix for the reported user, everything from the root cause through recoverable-guard and delete-path availability is done and tested. I'm happy to keep hardening, but it may be worth a maintainer deciding whether the remaining edges (open-time races, out-of-order measurement) justify staying in this PR or belong in a follow-up against a smaller, already-reviewed base.

---

originalix (commented):
@sidmorizon

Reviewed commit `da090e70cca5`.

- **P1 · Initial WebStorage open failures leave storage permanently pending**
- **P2 · The write probe only proves a net-zero transaction**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5240473093)

sidmorizon:
Both findings confirmed and fixed in e66165d2ed.

**P1 · Failed initial open leaves storage permanently pending** — real, and worth noting it is a pre-existing defect rather than one this PR introduced: `this.indexed = new Promise(async (resolve) => ...)` has no rejection path, so a rejecting `open()` never settled the outer promise and every later read and write hung forever. What this PR changed is that it made the failure *reachable and reportable* — on a full disk the open is where things fail first, so the hang sits directly on the recovery path this PR is about.

Initialization is now a retryable function behind a cached promise that is dropped on failure, preserving single-flight while letting the next call retry. The public `indexed` shape is kept as a getter, so all 13 call sites are unchanged. Test added: first open rejects, next call succeeds, initializer invoked twice.

**P2 · The probe only proved a net-zero transaction** — a good catch and I think the sharpest analytical point in the series. Put-then-delete in one atomic transaction has a net size delta of zero, so committing it proved only that an empty transaction completes — not that the space-growing write which originally failed can. That is precisely the claim the probe exists to make, so it was close to circular.

The probe now commits a retained 1 KB sentinel, and only that commit counts as proof. Removal happens in a separate best-effort transaction so cleanup can never turn a successful proof into a failure; leaving the sentinel behind on genuinely full storage is acceptable, since the next successful probe overwrites it.

Verified: 88 passed in `storageChecker` + `storage`, 135 passed / 2 skipped in `dbs/local`, full-repo oxlint exit 0, oxfmt and tsc clean.

Repeating the question from the previous round, since it still stands and only a maintainer can answer it: twelve rounds in, findings have moved steadily into narrower edges of the recovery state machine, and several were introduced by earlier fixes within this same PR. The reported user's defect — misclassification, unclearable guard, blocked deletes — is fixed and tested. Whether the remaining hardening belongs here or in a follow-up against an already-reviewed base is a call I'd rather have made explicitly than keep deciding by default.

---

originalix (commented):
@sidmorizon

Reviewed commit `e66165d2ed78`.

- **P1 · IndexedDB helpers resolve before transaction commit**
- **P1 · Recovery probe does not exercise the failed storage owner**
- **P2 · Blocked WebStorage writes log persisted values**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5240473093)

sidmorizon:
All three fixed in 275f4a8e1c.

**P1 · Helpers resolve before commit** — real, and the most consequential of the three for a wallet: a quota failure can abort at commit *after* the request succeeded, so `WebStorage.setItem` reported success for data that was never persisted, with the abort surfacing only asynchronously through the guard. `add`/`put`/`delete`/`clear` now await `tx.done` before resolving.

I checked the blast radius before making this change, since it alters resolution timing for every write through these helpers. Each helper creates and owns its own transaction, so no shared-transaction semantics change, and the consumers are essentially just `WebStorage` — `IndexedDBAgent` does its real work through `withTransaction` + `txXxx`. The 137-test `dbs/local` suite passes unchanged.

**P1 · Probe did not exercise the failed owner** — correct on both counts. `runWriteProbe` always used `globalThis.indexedDB` while production storage lives in named Storage Buckets with their own factories, so a default-origin commit proved nothing about the bucket that actually failed. The failing `IDBFactory` is now recorded alongside the error (`handleDiskFullError(error, { indexedDBFactory })`, wired from `doOpen` and the CryptoKey store) and the probe runs through it. The sentinel key is now unique per probe too — with a fixed key, a leftover from failed cleanup made the next probe an overwrite with no retained allocation, which is the same circularity as the net-zero transaction from last round.

**P2 · Blocked writes logged persisted values** — fixed, and it deserved more than P2 in my view: `setItem`/`mergeItem`/`multiSet`/`multiMerge` passed full values into a payload that gets `console.error`'d, and a storage incident produces a *burst* of blocked writes. WebStorage holds application and global state, so this could spray persisted content into consoles and log collection exactly during an incident. The payload is now typed to `{ method, key?, itemCount? }` — values cannot be passed by construction, not merely omitted at the call sites.

Verified: 88 passed in `storageChecker` + `storage`, 135 passed / 2 skipped in `dbs/local`, full-repo oxlint exit 0, oxfmt and tsc clean.

Not covered by tests, stated plainly rather than implied: a request-succeeds-then-commit-aborts case, and a probe against a named Storage Bucket. Both need an IndexedDB harness that can fail at commit and expose bucket factories, which none of the current suites have.

The scope question from the last two rounds still has no answer, and I'd rather not keep deciding it by default: the reported user's defect is fixed and tested, while rounds 11–13 have been hardening the state machine built around it — including two commit-and-persistence issues that predate this PR. A maintainer's call on where to draw the line would be more useful than another round from me.

---

originalix (commented):
@sidmorizon

Reviewed commit `275f4a8e1c73`.

- **P1 · Commit-time quota failures still lose the failed bucket**
- **P2 · Probe keys repeat after an extension runtime restart**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12787#issuecomment-5240473093)

sidmorizon:
Both fixed in 2934288843. These two complete last round's fix rather than opening new ground, which is why I took them without waiting on the scope question below.

**P1 · Commit-time failures lost the failed bucket** — correct, and it made last round's fix half-effective. I wired the factory from `doOpen` and the CryptoKey store but missed `IndexedDBTransactionPromised`, which is precisely the path a named-bucket quota failure takes: it aborts at commit. So `lastFailedIndexedDBFactory` stayed absent or stale and recovery could probe the default origin, or a *previously* failed bucket, while the one that rejected the write stayed broken. The transaction error handler now passes `db.nativeDBFactory`.

**P2 · Probe key repeats across runtime restarts** — also correct, and the same circularity as the previous two probe findings: a module-local counter resets when an MV3 worker restarts, so a sentinel stranded by a worker that stopped before cleanup gets overwritten by `probe-1` again — a commit that need not allocate, proving nothing. The key is now collision-resistant (timestamp + counter + random), and cleanup clears the probe store rather than a single key, so stranded sentinels cannot accumulate either.

Verified: 223 passed / 2 skipped across 21 suites, full-repo oxlint exit 0, oxfmt and tsc clean.

The probe has now been corrected four times — net-zero transaction, wrong factory, repeating key, and the self-sustaining scheduler loop. Each fix was right, but the pattern is worth naming: proving "a write can commit" from inside the system whose writes are blocked is inherently circular, and every iteration has found a new way the proof degenerates. If a maintainer wanted to cut it, dropping the probe and simply never auto-clearing a `writeFailed` guard — leaving quota-derived recovery, which the estimate genuinely is authoritative about — would remove that whole class of finding. I'm not making that call unilaterally since it trades away recovery from transient write failures.

That is the fourth round I've asked for a scope decision. Restating it once and then leaving it: the reported user's defect is fixed and tested, and rounds 11–14 have hardened the surrounding state machine, including issues that predate this PR. I'll keep handling findings that complete work already in the diff, but I'd rather a maintainer decide whether genuinely new hardening belongs here or in a follow-up.

---

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/shared/src/storageChecker/storageChecker.ts
```diff
@@ -11,76 +11,446 @@ import appGlobals from '../appGlobals';
...
+            });
+          } else if (quotaInfo.availableBytes >= getClearAtBytes(quotaBytes)) {
+            // Quota-derived state can be released by the quota estimate, but a
+            // write-failure state needs proof that a write commits again.
+            const needsWriteProof =
+              lastDiagnostics?.reason === EStorageFullReason.WriteFailed;
+            if (needsWriteProof) {
+              const canCommit = await probeWriteCanCommit();
+              if (canCommit && isCurrentMeasurement()) {
+                clearDiskFull(quotaInfo);
```
sidmorizon, line 433: # [🟠 P1] 测量窗口内新发生的写失败会被同一次测量清除，形成 raise/clear 振荡并反复弹窗

## 问题

`raiseDiskFull` / `handleDiskFullError` 不会递增 `measurementGeneration`，所以一次测量开始后（gen N），在 `await estimate()` 或 `await probeWriteCanCommit()` 的异步窗口内由真实写失败重新拉起的 guard，对这次测量来说仍是 "current"：

```ts
const canCommit = await probeWriteCanCommit();
if (canCommit && isCurrentMeasurement()) {
  clearDiskFull(quotaInfo); // 此时 lastDiagnostics 可能已被窗口内的新 WriteFailed 覆盖
}
```

探针只写 `1 KB`，而 app 写入（如 Hyperliquid L2 缓存）可达数 MB。在 estimate 看起来健康但大写入实际失败的场景——恰是 PR 描述中"写失败本身证明 estimate 不可预测"的那类场景——探针 commit 成功，`isCurrentMeasurement()` 为 true，刚由新写失败拉起的 guard 连同其新鲜证据一起被清掉。

## 影响

进入自持振荡：clear → 下一次大写入真实失败 → `handleDiskFullError` 重新 raise（`wasAlreadyFull === false`，转换沿触发 `emitWarning` 弹窗）→ 被阻塞写入调度测量 → 约 `1s` 后 estimate 健康 + `1 KB` 探针 commit → 再 clear → 循环。用户在持续写流量下每秒级看到弹窗反复弹出，部分复现了本 PR 要修的"关不掉的弹窗"体验；同时被清除窗口内的写入静默丢失（guard 放行后直接打到失败的 backing store）。

## 建议

让新证据使在途测量作废：在 `handleDiskFullError`（或 `raiseDiskFull` 的 WriteFailed 分支）里 `measurementGeneration += 1`，这样测量窗口内发生的真实写失败会令 `isCurrentMeasurement()` 变为 false，clear 只能由失败之后启动的新一轮测量（重新跑探针）做出。也可以在 `clearDiskFull` 前比对测量开始时捕获的 `lastDiagnostics` 引用是否未被替换，效果等价。

---

packages/shared/src/storage/WebStorage.ts
```diff
@@ -148,28 +149,67 @@ class WebStorage implements AsyncStorageStatic {
...
+   * path, so a failed `open()` — which is exactly what a full backing store
+   * produces at startup — left the promise pending forever and hung every
+   * later read and write. Freeing space could then never reach the reopening
+   * logic, and only restarting the runtime recovered. Caching the promise and
+   * dropping it on failure keeps the single-flight behavior while letting the
+   * next call try again.
+   */
+  private getIndexed(): Promise<IndexedDBPromised> {
+    if (!this.indexedPromise) {
+      this.indexedPromise = this.initIndexed().catch((error) => {
```
sidmorizon, line 173: # [🟡 P2] 可重试初始化在 open 成功后失败时泄漏已打开的 IndexedDB 连接

## 问题

`initIndexed` 先 `await indexed.open()` 成功建立原生连接，随后 `await migrateFromLegacyStorage(...)` 仍可能抛错（`indexed.getAllKeys` 的事务错误、`createLegacyStorage()` 动态 import 失败、legacy DB 在满盘时 open 抛 `Encountered full disk`——本文件 L107-109 的注释就是它的历史现场）。这里的 `catch` 只做 `this.indexedPromise = undefined` 然后 rethrow，已打开的 `indexed` 实例被直接丢弃，从未 `close()`。

这是本 PR 新引入的行为：旧实现失败时是 promise 永久 pending（不重试，也只挂一次）；新实现每次存储访问都会重跑 `initIndexed`，失败一次就泄漏一条打开的原生连接。

## 影响

在初始化持续失败的场景（正是可重试设计针对的满盘启动路径），bg 运行时以存储调用的频率累积无引用的打开连接——ext bg 是与 UI 共享同一底层 IndexedDB 的独立 JS runtime，这些连接只在下一次 `versionchange` 事件里才被 `attachConnectionInvalidationHandlers` 的自关闭逻辑回收，期间持续占用 backing store 句柄；每次重试还会整体重跑 legacy 探测/迁移流程，放大满盘时的 IO churn。

## 建议

在 `initIndexed` 内，open 成功后的步骤失败时先关闭再抛出：

```ts
await indexed.open();
try {
  await migrateFromLegacyStorage({ indexed, legacyKeyPrefix, tableName });
} catch (error) {
  indexed.close();
  throw error;
}
```

---

sidmorizon:
@cursoragent review

sidmorizon:
Resolved the merge conflict with `x` by merging `origin/x` into this branch (merge commit 5059fa3aa, no history rewrite).

**Two textual conflicts, both "each side added behavior" cases — both sides kept:**
- `LocalDbBase.removeCredentials` — this PR's `withSpaceFreeingTransaction` wraps the delete; x's post-delete HyperLiquid agent backup cleanup and LSE layer-key cleanup follow unchanged.
- `ServiceDBBackup.backupDatabase` — x's read-stale-agent-ids-before-the-transaction block kept; the snapshot transaction now opens via this PR's `transactionAsync`.

**One semantic reconciliation the textual merge could not see:** `removeBackupHyperLiquidAgentCredentials` (added on `x` in #13020, after this branch forked) opened a sync `readwrite` transaction, which this PR's guard blocks while storage is full — yet it is a space-freeing delete on the recovery path (called from `removeCredentials`) and carries the security duty of scrubbing device-bound signing keys from the backup. Converted it to `await transactionAsync(..., { allowWhenStorageFull: true })`, matching the sibling backup-delete methods in the same file.

Validated on the merged head: targeted jest for `storageChecker` + `storage` + `dbs/local` — 27 suites, 304 passed / 2 skipped; `yarn agent:check --profile commit` clean.

Note for reviewers: 5 review threads on the pre-merge head remain open and unaddressed (1× P1 — a write failure landing during an in-flight measurement can be cleared by that same measurement's probe, re-opening the raise/clear oscillation; 2× P2 — CryptoKey write failures don't record the failing `IDBFactory` for the probe, and `WebStorage.initIndexed` leaks an open connection when migration fails after `open()`; 2× lesser). This merge does not address them.

---

sidmorizon:
All three outstanding findings fixed in f6c8edb8e.

**P1 · A probe that predates a new write failure could clear the guard** — `handleDiskFullError` now bumps `measurementGeneration`, so a write failure landing while a measurement is awaiting its probe invalidates that measurement: the probe's success can no longer clear the guard the failure just raised, closing the raise/clear oscillation (and the silent write loss inside each cleared gap). Regression test drives the exact sequence — guard up, healthy estimate, probe held in flight, new failure injected, probe commits — and asserts the guard stays raised. Verified the test fails with the one-line fix reverted.

**P2 · CryptoKey write failures did not record the failing factory** — `writeCryptoKeyRecord` and `getOrCreateCryptoKey` now pass `getIndexedDBInstance(indexedDBInstance)` to `reportStorageWriteFailure`, so recovery probes the owner that actually failed instead of a previously failed bucket left in `lastFailedIndexedDBFactory`. Test asserts the factory reaches `handleDiskFullError`.

**P2 · Retryable WebStorage init leaked the opened connection** — the migration step now closes the connection opened by `indexed.open()` before rethrowing, so a persistently failing init (the full-disk startup path the retry exists for) no longer accumulates unreferenced connections that would also block `versionchange` upgrades. Test asserts `close()` is called exactly once when init fails after a successful open.

Verified: 91 passed across `storageChecker` + `storage` (9 suites), `yarn agent:check --profile commit` clean (oxlint type-aware, format, tsc).

---

packages/shared/src/storageChecker/storageChecker.ts
```diff
@@ -11,76 +11,446 @@ import appGlobals from '../appGlobals';
...
+ * to restart the extension background.
+ */
+function clearDiskFull(quotaInfo: IStorageQuotaInfo) {
+  if (!globalThis.$onekeySystemDiskIsFull) {
+    lastDiagnostics = undefined;
+    return;
+  }
+  globalThis.$onekeySystemDiskIsFull = undefined;
+  lastDiagnostics = undefined;
+  appGlobals?.$defaultLogger?.app.storage.diskFullCleared(quotaInfo);
```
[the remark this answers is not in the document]
sidmorizon: Acknowledged, and deliberately closing without a fix in this PR (maintainer decision). Impact is bounded: the dialog carries an explicit confirm button, and the clear itself is recorded via `defaultLogger.app.storage.diskFullCleared`, so support can still see recovery in an exported log. Auto-dismissal needs a new cross-runtime hide event plus a listener in `DiskFullWarningDialogContainer` — reasonable as a follow-up, out of scope here.

---

packages/shared/src/storageChecker/storageChecker.ts
```diff
@@ -11,76 +11,446 @@ import appGlobals from '../appGlobals';
...
+ * into another scheduled measurement. Left unchecked that measurement finds
+ * the guard still raised, probes again, and re-schedules itself — a loop that
+ * sustains itself roughly once a second with no app activity at all, keeping
+ * the extension background worker awake for as long as storage stays full.
+ */
+let writeProbeInFlight = false;
+
+function probeWriteCanCommit(): Promise<boolean> {
+  if (writeProbeInFlight) {
+    return Promise.resolve(false);
```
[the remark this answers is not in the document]
sidmorizon: Acknowledged, and deliberately closing without a fix in this PR (maintainer decision). Impact is bounded: measurement is driven by write activity, so the next write both re-schedules a measurement and re-runs the probe — a discarded in-flight result delays convergence by one write, it does not latch the guard. Note f6c8edb8e also made a new write failure invalidate in-flight measurements, so reusing an in-flight probe promise across generations would now need care to not resurrect a pre-failure result; fine as a considered follow-up, out of scope here.

---

sidmorizon:
CI note: **Web startup graph budget (24.x)** failed on f6c8edb8e — `swap` scenario, `FAIL jsDecodedBytes: 16.67 MiB / 16.67 MiB`. Fixed in 19e06763a.

This is not a regression from this PR. The same gate fails with the identical reading on unrelated PRs' pure `Merge origin/x` commits (e.g. [PR #13035's run](https://github.com/OneKeyHQ/app-monorepo/actions/runs/33052498983): also `16.67 MiB / 16.67 MiB` on swap), and today it has been coin-flipping across the repo — 7 of the last ~25 runs failed on exactly this budget, several of them on merge-of-x commits, while reruns of near-identical heads pass. Current `x` simply measures right at the line: the swap budget was last raised by only +16 KiB in 923d2f36b (Aug 22) and ordinary growth on `x` has already consumed it. The gate counts raw `PerformanceResourceTiming` entries, and the swap sample carries ~10 duplicate script fetches (190 resources vs 180 unique), so run-to-run noise alone spans tens of KiB — larger than the remaining headroom.

The fix aligns `swap.jsDecodedBytes` with the existing root/market budget (17475174 → 17616077, ~+138 KiB). That still leaves only ~0.8% headroom over the current measurement, so the gate keeps doing its job; every other scenario (0.5–1.9 MiB headroom) is untouched. This PR does add real storage-layer code to the web startup path, so moving the threshold in the same PR follows the documented process in `startup-graph-budget.yml`.

---

originalix (dismissed):

PR #12747, opened 2026-08-04
fix: address Swap Pro badges and stock charts (OK-59240, OK-59108)

zhaono1:
[OK-59240](https://onekeyhq.atlassian.net/browse/OK-59240) [OK-59108](https://onekeyhq.atlassian.net/browse/OK-59108)

----------

## Summary

- Show network badges for dynamic networks such as Robinhood in the Swap Pro selected-token header and position rows.
- Load Stock K-line data for tokens without a CoinGecko ID by passing networkId and tokenAddress to the existing market chart endpoint.
- Preserve the existing CoinGecko chart path when a CoinGecko ID is available.

## Intent & Context

- [OK-59240](https://onekeyhq.atlassian.net/browse/OK-59240) reports that Robinhood token network badges are visible in the selector but disappear from the Swap Pro header and position list.
- [OK-59108](https://onekeyhq.atlassian.net/browse/OK-59108) reports that Solana xStocks without a CoinGecko ID show an empty K-line chart on Desktop.

## Root Cause

- Swap Pro used a preset-only local network lookup when networkLogoURI was empty, so dynamically configured networks such as Robinhood could not resolve a badge.
- StockPriceChart stopped before requesting data whenever coinGeckoId was absent, although the chart endpoint supports an OKX fallback with networkId and tokenAddress.

## Design Decisions

- Reuse useNetworkLogoUri for dynamic network badge resolution without changing network presets or token-selection state.
- Keep CoinGecko as the preferred chart source when coinGeckoId exists.
- When coinGeckoId is absent, reuse the existing chart endpoint and pass the token identity required by its OKX fallback contract.
- Preserve the existing chart scope and stale-response guards.

## Changes Detail

- SwapProTokenSelect: resolve the selected token network logo with useNetworkLogoUri.
- SwapProPositionItem: resolve position-row network logos with the same hook.
- ServiceMarket.fetchTokenChart: accept an optional CoinGecko ID and forward networkId and tokenAddress.
- SwapStockDesktopContainer: allow the Stock chart request without a CoinGecko ID and provide its network/address identity.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Mobile, Desktop, Web, Extension for network badges; Desktop Stock chart for the reported K-line issue.
- **Risk Areas**: Asynchronous badge hydration and chart source selection. Existing CoinGecko callers remain unchanged, and the server prioritizes CoinGecko whenever its ID is present.

## Test plan

- [x] Run yarn agent:check --profile commit.
[...]

PR #12728, opened 2026-08-03
fix: improve market indicator quick bar gestures(OK-58973)

limichange:
[OK-58973](https://onekeyhq.atlassian.net/browse/OK-58973)

----------

## Summary
- Add direction locking to the TradingView V2 native indicator quick bar so vertical swipes scroll the Market Detail page while horizontal swipes remain native.
- Correct the mobile tab scroll bridge offset calculation for iOS content insets.
- Preserve quick-bar indicator taps and horizontal scrolling on both mobile platforms.

## Intent & Context
The horizontally scrollable indicator quick bar below the mobile Market Detail chart intercepted touch gestures. Users trying to scroll the page vertically from this area could not move the page freely. The fix must cover Android and iOS without breaking horizontal indicator browsing or indicator toggles.

## Root Cause
The nested horizontal ScrollView became the touch target before the surrounding collapsible page could recognize vertical intent. In addition, the existing touch-scroll bridge treated the adjusted tab scroll position as a raw native offset. On iOS, that ignored the negative content inset at the top and could collapse or jump the header instead of scrolling naturally.

## Design Decisions
- Detect the gesture direction after a 4 px threshold and lock it for the rest of the touch sequence.
- Forward only incremental vertical deltas to the collapsible tab scroll bridge; leave horizontal gestures with the native ScrollView.
- Apply iOS directional locking to reduce diagonal gesture ambiguity.
- Inject the bridge callback where the quick bar is rendered inside the collapsible tab context, keeping the behavior scoped to native Market Detail.
- Convert the adjusted scroll position back to the platform raw offset with scrollYCurrent + delta - contentInset and remove the mount-time target reaction that could change the initial scroll position.

## Changes Detail
- NativeIndicatorControls.tsx: add touch direction tracking and native horizontal direction locking to the indicator quick bar.
- MobileLayout.tsx: connect the native quick bar vertical deltas to the Market Detail collapsible tab scroll bridge.
- useMobileTabTouchScrollBridge.ts: correct iOS content inset handling and perform incremental scrolling without an initial target update.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile iOS and Android
- **Risk Areas**: Gesture arbitration between the horizontal indicator list and the vertical collapsible page, plus iOS content inset offset conversion.

## Test plan
- [x] Run yarn agent:check --profile commit.
- [x] Verify iOS vertical swipes from the quick bar scroll up and down, and indicator taps still toggle.
[...]

PR #12683, opened 2026-07-29
fix: re-enable the perps subscriptions handler when the UI is provably live

Minnzen:
## Summary

QA (Redmi, Android, this hot-update bundle): after entering Perps from a notification and then switching coins, the order book flapped between `--` and values for ~20 seconds. Reported as *"订单簿快速闪烁 - 和数值，太快了，一直闪"*.

One line of change: when the instrument-switch sync gate passes, re-enable the background subscriptions handler before reconciling.

## Root cause, from the QA log (2026-07-28, flicker window 15:14:14–15:14:38)

Four layers, each with direct log evidence:

**1. What painted `--`↔values:** the UI alternated cache reads between two precisions for the same coin — `nSigFigs=None → hit` on an aging in-memory snapshot (ageMs grew 45s → 65s, never refreshed) and `nSigFigs=4 → miss` on every storage layer. The render gate flips `visibleL2Book` between that stale book and `null` accordingly.

**2. Why precision 4 never had data:** the background subscriptions handler was disabled. Every `service_update_subscriptions_start` in the window is immediately followed by `service_update_subscriptions_skipped_disabled`. No live L2 flowed at all.

**3. Why it stayed disabled:** re-enabling only happens on a `route_focused` event, and the parsed navigation state produced **no events at all** between 15:14:01 and 15:14:37 — frozen at "non-Perp tab + modal open".

**4. The decisive part — that parsed state contradicts reality.** While the parse claimed the Perp tab was blurred and covered, the same log shows the Perp order book UI issuing ~20 cache reads/sec for 24 seconds and rendering ticker updates — and QA was watching the book flicker. On Android `freezeOnBlur` is on: a genuinely blurred tab is frozen and cannot run effects. So `routeFocused=false` was a **stale verdict**, not a fact. The disable keyed off it; nothing contradicted it until route events resumed at 15:14:37 (which is exactly when the flicker stopped).

The trigger of the stale parse itself is not recoverable from these logs (they do not record raw `rootState`); it correlates with notification-entry navigation (`popToMainRoute` + `pop: true` reset). Neither of us could reproduce it organically — normal tab switches, coin switches and modal open/close all produce correct focus transitions on a real device (verified with a temporary probe).

## Why this fix does not need the trigger

[...]

PR #12680, opened 2026-07-29
fix: keep perps cold-start data and trade mode consistent across runtimes

Minnzen:
## Summary

Perps cold start served data older than what the user had just been watching, and a restart could restore the wrong instrument or a mode label that disagreed with the order book. Both come from the same place: state that is owned by one runtime but read by another, with no guarantee the two agree.

This PR makes the shared cold-start cache converge per key instead of per file, makes the cold-start instrument restore deterministic, and makes the Perps UI read its trade mode from the instrument it is actually rendering.

## Root cause

The SWR cold-start cache stores the **entire app-wide store as one JSON object under one MMKV key**. On native, `main` and `bg` each hydrate their own in-memory copy **once at boot** over one shared MMKV file, and `flush()` wrote that **entire copy** back — last writer wins wholesale, per file, not per key.

Whichever runtime held the older copy kept erasing everything the other had persisted since its own boot:

- `bg` writes fresh order books to disk while the user watches Perps
- the user leaves Perps; any `main`-side SWR write schedules a flush; two seconds later `main` overwrites the whole file with its aged copy
- kill → relaunch → both runtimes hydrate the clobbered store → **prices older than last seen**

Forensics from a live device store: entries pinned at dates weeks apart in a store whose newest keys were minutes old — the signature of wholesale overwrites from copies of different ages.

### Shared SWR cache (`swrCacheUtils.ts`)

- `flush()` **merges per key** against the store on disk instead of overwriting the file. Deletions record session tombstones so a merge cannot revive a locally deleted key, while a disk entry written **after** the deletion still wins.
- An **unparseable store is no longer treated as absent**. `getObject()` reports both as `undefined`, so the merge used to persist only the pending keys and drop everything else. Reading the raw string keeps the two cases distinct.
- When there is **no store to merge against**, the merged result is no longer adopted in memory. On the extension both runtimes get the no-op storage stub, where the in-memory copy is the only one there is.
- The perps first-frame reload is **throttled per book target** (coin + `nSigFigs` + `mantissa`) rather than by a single global timestamp, so a reload for one pair cannot suppress the first read of another that the other runtime just wrote.

### Perps cold-start restore

- `tradingModeAtom` is now persisted. Without it every launch defaulted to `perp` and the startup path ignored a cached spot instrument.
[...]

PR #12799, opened 2026-08-08
chore: sync release/v6.5.0 to x

originalix:
Sync bundle release **#5** changes from `release/v6.5.0` to `x`.

## Scope

Incremental only — from `0f078ef` (the last commit carried to `x` by the previous sync, #12736) up to the release tip. Prior bundle releases #1–#4 were already synced by #12518, #12582, #12628 and #12736, so this is not a resync of the whole branch.

| | |
|---|---|
| Candidate range | 13 commits (`0f078ef..c04216c9`) |
| Release-tracking, excluded | 2 — `chore: release #4` (#12734), `chore: release #5` (#12798) |
| Already on `x` by patch-id | 2 — #12729, #12769 |
| **Applied** | **8** |
| Skipped as already-present | 1 — #12728 |
| Sync adaptations | 2 (below) |

## Applied

#12740 · #12748 · #12747 · #12683 · #12764 · #12753 · #12680 · #12725

## Skipped: #12728 (already on `x`)

`x` already carries #12728's gesture work — `INDICATOR_QUICK_BAR_VERTICAL_PAN_THRESHOLD` and `IIndicatorQuickBarTouchState` are both present, and `useMobileTabTouchScrollBridge.ts` is byte-identical to the release version. `x` then built further on top via #12781 (`split mobile TradingView indicator bar`) and renamed the module to `TradingViewV2ChartControls.tsx` — the release branch's `TradingViewNativeChartControls` import path no longer exists. Cherry-picking would have reverted #12781.

## Conflict resolutions

12 conflicts across 30 files. Most were additive import/enum lists where both sides appended and both were kept. Three deserve attention:

**1. #12747 stock chart — `x`'s policy kept, payload applied.** `x` reworked this callback (swr cache fallback, retry, try/catch) via #12696/#12704/#12730 and carries an explicit comment: *"A missing CoinGecko lookup id is a request-readiness gap, not a real empty chart response."* #12747 takes the opposite position — it makes `coingeckoId` optional so charts fetch via `networkId`+`tokenAddress` without an id. Rather than silently overriding either author, `x`'s structure and `chartRequestReady` gate are kept untouched and only the `fetchTokenChart` payload was added.

> ⚠️ **Follow-up for the #12747 author:** OK-59108's no-coingecko-id case is therefore *not* fixed on `x` — `chartRequestReady` still gates it out. Reconciling that with `x`'s snapshot policy is a product decision and needs a Stock-chart regression pass.

**2. Locale files — `x`'s corrected translations kept.** The release branch's side of `de`/`fr_FR`/`it_IT`/`pt`/`pt_BR`/`vi` re-translated 6 existing `perp_order_type_*_desc__desc` keys with umlauts stripped and informal register (`gewahlten`/`ausgefuhrt`/`uber`, *du* vs *Sie*). `x` holds the corrected values, so those were kept and only the two genuinely new keys — `perp_para_market__desc`, `perp_xyz_market__desc` — were taken. All 19 locales verified at 5239 keys each, parsing clean with no duplicates.

**3. `ESwapSource` stays a value import.** `x` uses it at runtime (`enterFrom === ESwapSource.WALLET_HOME_TOKEN_LIST`); the release side had narrowed it to `import type`, which would not have compiled.

## Sync adaptations (2 extra commits)

Both fix breakage that only appears when release commits meet `x`'s current code:

- **`fix: complete normalizeDexCoin conversion for x sync`** — #12753 deletes the local `normalizePerpsCoin` helper in favour of the shared `normalizeDexCoin`, but `x` had a third call site the commit had no hunk for. Git merged it textually and the result referenced a deleted function. Converted like its two siblings.
- **`fix: add stock chart request deps for x sync`** — completes the #12747 resolution's `usePromiseResult` deps. Behaviourally a no-op; `chartScope`/`chartAssetScope` already encode both values and are already listed.

## Verification

- `yarn tsc:staged` — full-repo typecheck, **exit 0, 0 errors** (this is what caught the `normalizeDexCoin` breakage)
- `oxlint --type-aware --deny-warnings` over all 106 changed `.ts`/`.tsx` — **clean**
- `jest` over all 23 test files in the diff — **23 suites, 340 tests, all passing**
- No `RELEASES.json`, no conflict markers, no temp/scratch files in the diff

> Note: `yarn agent:check --profile pr` reports `lint-worktree: no changed JS/TS files` and passes `lint-staged` on a committed branch — both compare against the index, so neither actually inspected these changes. The checks above were run against the `origin/x..HEAD` diff instead.


wabicai (dismissed):

originalix:

@originalix

Reviewed commit `539275f61e48`. This report always reflects the latest reviewed changes.

### Review summary

This release-sync PR moves the Perps stack from a single-sub-DEX assumption to registry-driven multi-dex ownership, carrying dex identity through metadata caches, asset mapping, search, favorites, imagery, instrument switching, and subscription recovery. It also improves Market cold-start rendering, reserves native gas for private-send maximum amounts, and adds Swap Pro source analytics while keeping transaction construction in the background-service boundary. The main behavioral regression is that the native Universal Search entry now selects a Perp instrument without opening its market-detail route.

**What needs attention:** Confirm that every native Perp entry point, particularly Universal Search, reaches the intended market-detail screen after selecting an instrument, then exercise subscription recovery across the independently initialized main and background runtimes.

### Issues to address

- **P2 · Mobile universal-search tap no longer opens the Perp market page**
  - **Impact:** On native, tapping a Perp search result now only switches to `ETabRoutes.Perp` and changes the active instrument. That tab renders `PerpMobileLayout`, while the selected market's chart/detail experience remains under `EModalPerpRoutes.MobilePerpMarket`, so the user lands on the generic Perp shell instead of the expected market page.
  - **Suggested action:** Keep the new dex-aware instrument resolution, but restore the native root-navigation transition to `EModalPerpRoutes.MobilePerpMarket` after the tab switch and add coverage for this entry path.

### Validation gaps

- Device-level Universal Search navigation and Perps resume/recovery flows were not exercised.
- The focused Jest command could not start because this workspace has no Yarn `node_modules` state file; the added unit tests were inspected but not executed here.
- Native and extension split-runtime behavior still needs rapid instrument-switch, foreground-resume, lock/unlock, and websocket-reconnect coverage. Desktop and web should be checked separately as single-runtime targets.

originalix:
## Web startup graph budget fix — restores an optimization #12748 removed

@huhuanming flagging you directly, since this touches a deliberate change of yours rather than a merge artifact.

### What failed

`Web startup graph budget (24.x)` was the only failing check (17/18 passed):

| scenario | metric | actual / budget |
|---|---|---|
| root | resourceCount | 250 / 240 |
| root | longTaskTotalMs | 2357 ms / 2000 ms |
| market | resourceCount | 250 / 240 |
| market | longTaskTotalMs | 2316 ms / 1900 ms |

`perps`, `swap`, `defi`, `referFriends` all passed unchanged — the regression is confined to the two scenarios that mount the market table.

### Attribution — introduced by this PR, not pre-existing

All 7 compared CI runs are merges into the **same** base commit `3f55b9f0` (`git log 3f55b9f0..origin/x` is empty). Six non-#12799 runs on that base sit at root **189–192** resources and **~940–1374 ms**; this PR measures 250 / 2357. PR #12761 ran 15 minutes later on the same base and same runner image and scored 191 / 1374.

Breaking `rawResources` down by `initiatorType`, every non-image bucket is identical (script 125–126, link 26, css 3, xhr 4, fetch 1). Only `img` moves: **29 → 89** (root) / **92** (market), and the URL set difference is strictly one-directional — ~60 market token logos on `uni.onekey-asset.com` present only here.

### Cause

`b8ce45b2ab` (#12748) deletes `MARKET_HOME_WEB_EAGER_RICH_ROW_COUNT` and `deferRichRowAfterIndex` from `MarketTokenListBase.tsx`. That deferral was added by **#12240 "perf: optimize web market cold start"** — the same PR that established these budgets. Without it, `shouldUseLightweightCell` returns false at every gate site, so all seeded rows render rich and each issues a token-logo request inside the measured window instead of only the first 4.

⚠️ **Measurement caveat:** 250 is Chrome's default resource-timing buffer ceiling, not a real count — all six root/market runs pin at exactly 250 while baselines are unsaturated at 189–192. The true regression is **≥ +59 resources**, not the +10 the overage suggests. This also makes a threshold bump unworkable: no passing `resourceCount` value exists above 250.

### Fix (`539275f`)

Restores the deferral in `MarketTokenListBase.tsx` only. Purely additive — the consumer plumbing (`useMarketTokenColumns`'s 11th parameter, `shouldUseLightweightCell`, `useColumnsDesktop`) was never removed and is byte-identical to `x`, so this just re-arms existing code. `platformEnv.isWeb && webTabIntegrated` scoped; native/desktop/extension unaffected.

**#12748's actual fix is fully preserved** — the synchronous `useState(() => transformMarketTokenListResponse(...))` seeding in `useMarketTokenList.ts` and the `<Suspense fallback={<MarketListLoadingFallback />}>` + `ensureTabActivated` changes in `DesktopLayout.tsx` are untouched. Market still paints real rows on first render; rows past index 3 render the lightweight cell for the first 6 s on web, exactly as on `x` today.

After the fix, this file's only remaining delta from `x` is #12725's intended `forceStockMetadataColumns` change.

**If the removal was a deliberate UX call** (e.g. grey placeholder cells looked wrong beside the new loading fallback), say so and we'll find another way to get under budget — this restore is easy to drop.

### Also worth knowing

The budget job only runs on PRs targeting `x`, so `release/v6.5.0` was never measured against it. **Bundle #5 is already shipping this cold-start cost to production**, unmeasured. Worth deciding separately whether to port the deferral onto the release branch.

Verified locally: `tsc` 0 errors, `oxlint --type-aware --deny-warnings` clean over 106 changed files, 23 jest suites / 340 tests passing. Budget result pending this CI run.

originalix (commented):
@originalix

Reviewed commit `539275f61e48`.

- **P2 · Mobile universal-search tap no longer opens the Perp market page**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12799#issuecomment-5227003165)

wabicai (approved):

PR #12800, opened 2026-08-09
fix: gate pre-release firmware artifact hosts behind developer mode

originalix:
## Background

`DesktopApiFirmwareArtifact.ts` (Electron main process) admits firmware artifact downloads against a hardcoded hostname allowlist. The allowlist currently on `hotfix/v6.5.2` carries a developer object-storage bucket alongside the two production asset hosts, and #12788 swaps that bucket for a different hardcoded one. That leaves two problems:

1. **The hostnames keep changing.** Pre-release firmware artifacts are published by developers to their own buckets across several object-storage providers, and the bucket changes from build to build. Every new bucket means a code change and a release.
2. **A developer host sits in the production allowlist.** Scanning the current production firmware manifest, the only artifact hostnames it references are the two official OneKey asset hosts — no developer bucket appears anywhere — and the Pro2 / Neo sections do not exist in the production manifest at all. So that bucket serves no production purpose; it is pure admission surface. Removing it has zero impact on production users.

On top of that, most bucket hostnames in the pre-release manifest are not on the allowlist at all, so once a developer turns on "Use pre-release config" those downloads are rejected outright by the main process.

## Changes

Take developer hostnames out of the static allowlist entirely and admit them through the developer-mode switch instead:

- **Narrow the production allowlist** to the two official OneKey asset hosts, dropping every developer bucket entry. The production path ends up stricter than before this PR.
- `isFirmwareArtifactUrlAllowed(url, options?)` gains `allowPreReleaseHosts`: when `true`, the hostname-set comparison is skipped. **Structural checks are unaffected** — `https:`, empty port, no username/password, no hash still always run, and `redirect: 'error'` and `maxBytes` are untouched.
- **Skipping hostname pinning now requires content pinning.** `validateDownloadInput` rejects outright when `allowPreReleaseHosts` is true and `expectedSha256` is missing. Every admitted download therefore satisfies at least one of "hostname is on the reviewed allowlist" or "content is SHA-256 pinned" — there is no path where neither holds. Every artifact in the current pre-release manifest already carries a hash (`fingerprint` for components, `archiveSha256` for resource archives, zero exceptions), so this constraint does not block the real developer flow.
- bg has exactly one injection point: `FirmwareArtifactPreflight.downloadFirmwareArtifact()` reads `usePreReleaseConfig` and passes it through. This is the only `firmwareArtifactAdapter.download()` call site in the repository, so no other caller can forget it or route around it.
- Extract `getGatedFirmwareUpdateDevSetting()` into `states/jotai/atoms/devSettings.ts`, preserving the existing "global developer mode must be enabled" two-atom gate. `ServiceDevSetting.getFirmwareUpdateDevSettings` now delegates to it so the gate is not written twice.

## The switch has exactly one way in

`allowPreReleaseHosts` can only be true when both of these hold:

1. Global developer mode: `devSettingsPersistAtom.enabled === true`
2. The firmware dev setting `usePreReleaseConfig === true` (Settings → Developer options → Firmware Update → "Use pre-release config")

If either is false, `getGatedFirmwareUpdateDevSetting` returns `undefined`, the `=== true` comparison fails, the field never appears in the download input, and the main process evaluates `input.allowPreReleaseHosts === true` as false and applies the full allowlist.

On defaults: `devSettingsPersistAtom.enabled` defaults to `false` in production builds (`true` in dev/E2E builds), while `usePreReleaseConfig` defaults to `false` unconditionally in every build. Either way an explicit user action is required before the switch can open.

`ServiceDevSetting.switchDevMode(false)` also resets `usePreReleaseConfig` to `false` when developer mode is turned off, so there is no "developer mode off but a stale true value persisted" state.

Admission uses a strict `=== true` at both layers (`validateDownloadInput` and `isFirmwareArtifactUrlAllowed`). The field crosses an IPC boundary and can in principle arrive as any JSON value, so a test locks this down: `undefined` / `null` / `false` / `0` / `1` / `'true'` / `'false'` / `{}` / `[]` are all rejected, and only the literal boolean `true` skips hostname pinning. Anyone who later rewrites this as `Boolean(x)` or `!!x` will fail the test.

This matches what the switch already means: turning it on already redirects the entire firmware manifest to the pre-release one, so widening host admission is an extension of the same trust decision.

## Runtime boundaries

- The Electron main process owns URL admission and on-disk verification; it never touches the bg JS heap.
- The bg runtime reads the dev setting and drives download, verification and extraction; that data is deserialized only into the bg heap.
- The native (iOS/Android) adapter builds its payload field by field and does not forward the flag, so native modules need no change — native has no hostname allowlist to begin with, only HTTPS plus a same-host response check.

## Known trade-off

The main process trusts the boolean bg sends it. A fully compromised renderer can call the desktopApi bridge directly and bypass the dev setting — that is inherent to the bridge (`DESKTOP_API_CALL` in `apps/desktop/app/app.ts` does not authorize per method on allowlisted modules) and is not introduced by this PR.

In that scenario the ceiling for an attacker is: fetch bytes from an arbitrary https host **whose SHA-256 matches a digest they chose**, and have them written to the artifact store. They cannot obtain arbitrary content, because skipping hostname pinning now requires a SHA-256 (above) and the main process verifies against it before the file is promoted, deleting it on mismatch. Installing firmware onto a device still has to pass device-side signature verification afterwards. Relative to today — a developer bucket permanently on the production allowlist, with no such requirement — the production path is a net tightening.

One scoping note: the flag is derived from the global switch rather than from each artifact's provenance, so while developer mode is on a stable-manifest download also takes the relaxed comparison. In practice that is a no-op, since stable-manifest URLs are already on the allowlist; making it provenance-driven would mean adding a field to `IFirmwareDownloadArtifact`, which is a design change rather than a fix.

## Verification

- [x] `desktopApis/` + `ServiceFirmwareUpdate/` + `ServiceDevSetting` + `states/jotai`: 302/302 passing
- [x] `yarn agent:check --profile pr` — all local checks pass
- [ ] On device: with developer mode and "Use pre-release config" on, download Pro2/Neo resources from a bucket referenced by the pre-release manifest; with the switch off, confirm the same URL is rejected by the main process

Note that `lint.yml` and `unittest.yml` only trigger for PRs whose base branch is `onekey`, `x`, `hotfix/*` or `release/*`, so they do not run on this stacked PR. Unit tests were dispatched manually against this branch instead, and CI will cover the change again through #12788 once this merges.

New tests are deliberately spread across three levels rather than only covering the pure function:

1. **The admission predicate**: with the switch on, an arbitrary bucket host is admitted and production hosts are unaffected; with it off, the bucket host is rejected. With the switch on, `http:` / a port / userinfo / a fragment are all still rejected. Plus the strict `=== true` fail-closed behaviour described above.
2. **The `download()` entry point** (where an IPC caller actually lands, not the pure function): a non-allowlisted host is rejected; `allowPreReleaseHosts: 'true'` as a string does not sneak through; the switch on without a SHA-256 is rejected; the switch on with a SHA-256 reaches the network stage. This exists specifically so that rewriting `=== true` as `Boolean(...)` cannot pass while the predicate tests stay green.
3. **The gate itself** (`devSettings.test.ts`, mocking only the two atoms and running the real gate): with developer mode off it returns `undefined` even when a stale `usePreReleaseConfig: true` is persisted, and it does not read the firmware settings atom at all.

A bg-side test additionally asserts that `downloadFirmwareArtifact` passes the field to the adapter only when the dev setting is true, and omits it entirely otherwise.

Test fixtures use reserved `.test` hostnames rather than any real bucket, so no developer storage endpoint is named in this repository.

### DIFF
--- a/packages/kit-bg/src/desktopApis/DesktopApiFirmwareArtifact.test.ts
+++ b/packages/kit-bg/src/desktopApis/DesktopApiFirmwareArtifact.test.ts
@@ -29,15 +29,13 @@ describe('DesktopApiFirmwareArtifact URL admission', () => {
     ).toBe(true);
     expect(
       isFirmwareArtifactUrlAllowed(
-        new URL(
-          'https://pub-568cac7a13bf4c42b7a8113ffffc6793.r2.dev/firmware.bin',
-        ),
+        new URL('https://common.onekey-asset.com/firmware.bin'),
       ),
     ).toBe(true);
     expect(
       isFirmwareArtifactUrlAllowed(
         new URL(
-          'https://pub-d5c080673b4e4e9dae7e03680340378d.r2.dev/firmware.bin',
+          'https://pub-568cac7a13bf4c42b7a8113ffffc6793.r2.dev/firmware.bin',
         ),
       ),
     ).toBe(false);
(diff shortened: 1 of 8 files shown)


originalix:

@originalix

Reviewed commit `bfb02b9be885`. This report always reflects the latest reviewed changes.

### Review summary

The PR changes the desktop firmware-artifact trust boundary: the background runtime reads a developer-mode-gated pre-release setting and forwards a per-download flag over IPC so the Electron main-process adapter can accept otherwise unlisted HTTPS hosts, with SHA-256 pinning intended to protect widened admission. It centralizes developer-setting gating and adds focused admission and forwarding tests, while the native downloader remains unchanged. However, the new integrity guard also rejects approved-host artifacts that legitimately omit plan integrity fields whenever pre-release mode is enabled.

**What needs attention:** Confirm that SHA-256 is required only when a download actually bypasses the default desktop hostname allowlist, while approved-host unpinned plans retain their existing behavior.

### Issues to address

- **P2 · Pre-release mode rejects valid approved-host artifacts**
  - **Impact:** FirmwareArtifactPreflight adds allowPreReleaseHosts to every download while usePreReleaseConfig is enabled. DesktopApiFirmwareArtifact then requires expectedSha256 before checking whether the URL already belongs to FIRMWARE_ARTIFACT_HOSTNAMES. The existing plan contract and tests deliberately support approved-host raw and ZIP artifacts without integrity fields, so a mixed pre-release manifest can now fail with ARTIFACT_INVALID_INPUT even though that artifact does not need widened host admission.
  - **Suggested action:** Require expectedSha256 only when the URL is outside the default hostname allowlist and allowPreReleaseHosts is actually needed, then add coverage for an approved host with the flag enabled and no expectedSha256.

### Validation gaps

- `git diff --check` completed without errors, and no changed-file package-boundary violation or secret-like addition was found.
- The targeted Jest suites could not run because the workspace lacks Yarn's `node_modules` state file.
- Existing GitHub reviews, inline comments, and thread state could not be retrieved, so the inline comment should be deduplicated against current PR discussion before posting.

PR #12805, opened 2026-08-10
fix: preserve deprecated wallet account selection(OK-58315)

originalix:
[OK-58315](https://onekeyhq.atlassian.net/browse/OK-58315)

----------

## Summary

- Keep reset-device (`deprecated`) hardware wallets visible and explicitly selectable so users can inspect historical accounts and assets.
- Reject only missing or mocked target wallets at the `confirmAccountSelect` write boundary, and close the account selector modal only after a successful selection.
- Preserve an explicitly selected deprecated account during storage initialization and reload, while continuing to exclude deprecated and mocked wallets from automatic fallback candidates.
- Keep the account selector available and render wallet content or a recovery spinner instead of leaving Home on a permanent empty stack.

## Root Cause

A deprecated wallet represents a historical wallet whose device identity changed after reset. It is unavailable for signing and address creation, but its existing accounts remain valid for read-only asset viewing.

The account selector, storage cleanup, auto-selection, and Home rendering paths used the deprecated-or-mocked predicate as if both states were identical. This caused a persisted deprecated selection to be cleared or replaced during reload. When a deprecated wallet was active, Home also classified it as having no usable wallet, suppressed wallet content and the selector trigger, and could leave only the default empty stack. At the same time, `confirmAccountSelect` had no final lookup guard for genuinely missing or mocked wallets.

The fix separates these boundaries:

- Explicit selection/viewing: wallet must exist and must not be mocked; deprecated is allowed.
- Automatic fallback: deprecated and mocked wallets remain excluded.
- Signing, address creation, and hardware operations: existing deprecated-wallet restrictions remain unchanged.

No Wallet, Account, Device, Credential, or key data is deleted, migrated, or rebound.

## Runtime And State Flow

- Desktop/Web use a single JavaScript runtime/thread, but LocalDbIndexed records returned through background services and the main Jotai snapshots are separate data snapshots. Direct IndexedDB mutation does not mutate an already-held Jotai value; initialization and wallet events must reconcile it explicitly.
- Extension uses isolated main/background JavaScript heaps. Native uses Realm and isolated main/background heaps; each side initializes independently and the main side receives serialized active-account data.
- Main and background bundles are version-locked, so this is a state-transition and timing issue rather than a bundle-version skew.

## Risk

Medium. The change touches account selection, reload recovery, and Home fallback rendering across Desktop, Web, Extension, and Native. The main risk is switching or retaining the wrong active account; the write boundary, automatic-candidate filtering, stale-request result, and regression tests cover those cases.

## Test Plan

- [x] `yarn jest packages/kit/src/views/Home/pages/homePageNoWalletContent.test.ts packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx packages/kit/src/components/AccountSelector/hooks/useAutoSelectAccount.test.tsx packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.test.tsx --runInBand` (69 tests)
- [x] Equivalent hotfix Desktop bundle validated with the existing reproduction database: the deprecated wallet remains visible with its reset warning, its historical account can be selected to view assets, the selection survives reload, Home does not become blank, and normal wallet switching remains unaffected. Workflow: https://github.com/OneKeyHQ/app-monorepo/actions/runs/31361967357

## Affected Platforms

- Desktop
- Web
- Extension
- Mobile

## Jira

https://onekeyhq.atlassian.net/browse/OK-58315

[OK-58315]: https://onekeyhq.atlassian.net/browse/OK-58315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
+++ b/packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
@@ -903,10 +903,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
 
     const wallet = await serviceAccount.getWalletSafe({ walletId });
     const isMissingWallet = !wallet;
-    const isDeprecatedOrMocked = Boolean(
-      wallet && accountUtils.isWalletDeprecatedOrMocked(wallet),
-    );
-    return isMissingWallet || isDeprecatedOrMocked;
+    return isMissingWallet || Boolean(wallet?.isMocked);
   };
 
   clearUnavailableWalletSelectionsInStorage = async ({
@@ -1391,6 +1388,22 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
         confirmRequestId,
       );
 
+      let wallet: IDBWallet | undefined;
+      try {
+        wallet = await serviceAccount.getWalletSafe({ walletId });
+      } catch {
+        return false;
+      }
+      if (!wallet || wallet.isMocked) {
+        return false;
+      }
+      if (
+        this.confirmAccountSelectLatestRequestIdMap.get(confirmRequestKey) !==
+        confirmRequestId
+      ) {
+        return false;
+      }
+
       const accountNetworkId: string =
         forceSelectToNetworkId ||
         this.getAutoSelectNetworkIdForAccount.call(set, {
@@ -1436,7 +1449,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
         this.confirmAccountSelectLatestRequestIdMap.get(confirmRequestKey) !==
         confirmRequestId
       ) {
-        return;
+        return false;
       }
 
       const newSelectedAccount: IAccountSelectorSelectedAccount = {
@@ -1546,6 +1559,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
         indexedAccountId: indexedAccount?.id,
         othersWalletAccountId: othersWalletAccount?.id,
       });
+      return true;
     },
   );
 
@@ -3557,14 +3571,12 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
           activeAccount;
         const selectedAccount = this.getSelectedAccount.call(set, { num });
         const isAccountExist = Boolean(indexedAccount || account || dbAccount);
-        // Mocked / deprecated wallets are no longer user-facing — treat them
-        // as needing replacement so the auto-select loop runs and either picks
-        // the next valid wallet or resets to undefined (OK-51091).
+        // Mocked wallets need replacement. Deprecated wallets remain readable.
         const shouldAutoSelectNextAccount =
           !selectedAccount?.focusedWallet ||
           !network ||
           !wallet ||
-          accountUtils.isWalletDeprecatedOrMocked(wallet) ||
+          wallet.isMocked ||
           !isAccountExist;
 
         if (shouldAutoSelectNextAccount) {
@@ -3621,7 +3633,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
           if (
             !selectedWalletId ||
             !hasIndexedAccounts ||
-            accountUtils.isWalletDeprecatedOrMocked(selectedWallet)
+            selectedWallet?.isMocked
           ) {
             let shouldSelectHdHwWallet = true;
             if (
@@ -3683,10 +3695,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
                 selectedAccountNew.focusedWallet = selectedWalletId;
               }
               // maybe no hd hw wallet found, reset walletId and indexedAccountId
-              if (
-                !selectedWallet ||
-                accountUtils.isWalletDeprecatedOrMocked(selectedWallet)
-              ) {
+              if (!selectedWallet || selectedWallet.isMocked) {
                 defaultLogger.accountSelector.autoSelect.resetSelectedWalletToUndefined(
                   {
                     selectedAccount: selectedAccountNew,
@@ -3810,7 +3819,7 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
             });
             if (
               !finalWallet ||
-              accountUtils.isWalletDeprecatedOrMocked(finalWallet) ||
+              finalWallet.isMocked ||
               (await serviceAccount.isTempWalletRemoved({
                 wallet: finalWallet,
               }))
--- a/packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx
+++ b/packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx
@@ -423,19 +423,25 @@ export function AccountSelectorAccountListItem({
                 autoChangeToAccountMatchedNetworkId =
                   selectedAccount?.networkId;
               }
-              await actions.current.confirmAccountSelect({
+              const confirmed = await actions.current.confirmAccountSelect({
                 num,
                 indexedAccount: undefined,
                 othersWalletAccount: account,
                 autoChangeToAccountMatchedNetworkId,
               });
+              if (!confirmed) {
+                return;
+              }
             } else if (focusedWalletInfo) {
-              await actions.current.confirmAccountSelect({
+              const confirmed = await actions.current.confirmAccountSelect({
                 num,
                 indexedAccount,
                 othersWalletAccount: undefined,
                 autoChangeToAccountMatchedNetworkId: undefined,
               });
+              if (!confirmed) {
+                return;
+              }
             }
             resetAccountManagerStacksModal();
           },
--- a/packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx
+++ b/packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx
@@ -477,6 +477,46 @@ describe('useAccountSelectorActions', () => {
     jest.restoreAllMocks();
   });
 
+  it('selects deprecated wallets but rejects unavailable wallets', async () => {
+    const { store, Wrapper } = createWrapper();
+    const { result } = renderHook(() => useAccountSelectorActions().current, {
+      wrapper: Wrapper,
+    });
+    const confirm = async (wallet: IWallet | undefined, walletId: string) => {
+      mockGetWalletSafe.mockResolvedValueOnce(wallet);
+      let confirmed = true;
+      await act(async () => {
+        confirmed = await result.current.confirmAccountSelect({
+          num: 0,
+          indexedAccount: {
+            id: `${walletId}--0`,
+            walletId,
+          } as IIndexedAccount,
+          othersWalletAccount: undefined,
+        });
+      });
+      return confirmed;
+    };
+
+    expect(
+      await confirm(
+        { id: 'hw-deprecated', deprecated: true } as IWallet,
+        'hw-deprecated',
+      ),
+    ).toBe(true);
+    expect(store.get(selectedAccountsAtom())[0]).toMatchObject({
+      walletId: 'hw-deprecated',
+      indexedAccountId: 'hw-deprecated--0',
+    });
+    expect(
+      await confirm(
+        { id: 'hw-mocked', isMocked: true } as IWallet,
+        'hw-mocked',
+      ),
+    ).toBe(false);
+    expect(await confirm(undefined, 'hw-missing')).toBe(false);
+  });
+
   it('marks active account init done when reload finishes before storage init', async () => {
     const selectedAccountsMapDeferred = createDeferred<
       ISelectedAccountsMap | undefined
@@ -861,6 +901,7 @@ describe('useAccountSelectorActions', () => {
           othersWalletAccount: undefined,
           num: 0,
         });
+        await Promise.resolve();
         // The later selection resolves first...
         resolvers.get('qr-1')?.('btc--0');
         await latestSelect;
@@ -1425,6 +1466,49 @@ describe('useAccountSelectorActions', () => {
         }),
       }),
     );
+  });
+
+  it('preserves a deprecated wallet during init and auto-select', async () => {
+    const selectedAccount = createHdSelectedAccount('hd-1--0');
+    const deprecatedWallet = { id: 'hd-1', deprecated: true } as IWallet;
+    const indexedAccount = {
+      id: 'hd-1--0',
+      walletId: 'hd-1',
+    } as IIndexedAccount;
+    mockGetSelectedAccountsMap.mockResolvedValue({ 0: selectedAccount });
+    mockGetWalletSafe.mockResolvedValue(deprecatedWallet);
+
+    const { store, Wrapper } = createWrapper();
+    const { result } = renderHook(() => useAccountSelectorActions().current, {
+      wrapper: Wrapper,
+    });
+
+    await act(async () => {
+      await result.current.initFromStorage({
+        sceneName: EAccountSelectorSceneName.home,
+      });
+    });
+    store.set(activeAccountsAtom(), {
+      0: {
+        ...defaultActiveAccountInfo(),
+        ready: true,
+        wallet: deprecatedWallet,
+        indexedAccount,
+        network: { id: selectedAccount.networkId } as NonNullable<
+          ReturnType<typeof defaultActiveAccountInfo>['network']
+        >,
+      },
+    });
+    await act(async () => {
+      await result.current.autoSelectNextAccount({
+        num: 0,
+        sceneName: EAccountSelectorSceneName.swap,
+      });
+    });
+
+    expect(store.get(selectedAccountsAtom())[0]).toMatchObject(selectedAccount);
+    expect(mockSaveSelectedAccount).not.toHaveBeenCalled();
+    expect(mockGetAllHdHwQrWallets).not.toHaveBeenCalled();
   });
 
   it('syncs home when storage init clears an unavailable home-sync source wallet', async () => {
--- a/packages/shared/src/utils/accountUtils.test.ts
+++ b/packages/shared/src/utils/accountUtils.test.ts
@@ -21,6 +21,26 @@ function testWithRandomAccountIndexes(
     testFunc(accountIndex);
   }
 }
+
+describe('hasNoUsableWallet', () => {
+  type IWallet = Parameters<typeof accountUtils.hasNoUsableWallet>[0]['wallet'];
+
+  test('keeps deprecated wallets viewable but hides mocked wallets', () => {
+    expect(
+      accountUtils.hasNoUsableWallet({
+        wallet: { id: 'hw-1', deprecated: true } as IWallet,
+        account: undefined,
+      }),
+    ).toBe(false);
+    expect(
+      accountUtils.hasNoUsableWallet({
+        wallet: { id: 'hw-1', isMocked: true } as IWallet,
+        account: undefined,
+      }),
+    ).toBe(true);
+  });
+});
+
 describe('Lightning Path Transformation', () => {
   test('buildBtcToLnPath transforms mainnet path correctly', () => {
     testWithRandomAccountIndexes((accountIndex) => {
--- a/packages/shared/src/utils/accountUtils.ts
+++ b/packages/shared/src/utils/accountUtils.ts
@@ -870,12 +870,8 @@ function isOthersWallet({ walletId }: { walletId: string }): boolean {
   );
 }
 
-// A wallet is considered deprecated-or-mocked when it still has a DB record
-// but is no longer user-facing. Hardware wallets are marked `isMocked=true`
-// on removal (OneKey keeps the metadata so re-connecting the device restores
-// account names / ordering); `deprecated=true` covers the stale-device
-// variant (same connection, different deviceId). Both forms are "zombie"
-// wallets that callers should treat as if the wallet is gone (OK-51091).
+// Mocked and deprecated wallets both require write-path restrictions. A
+// deprecated wallet may still be shown for historical asset viewing.
 function isWalletDeprecatedOrMocked(wallet: IDBWallet | undefined): boolean {
   if (!wallet) return false;
   return !!(wallet.isMocked || wallet.deprecated);
@@ -889,7 +885,7 @@ function hasNoUsableWallet({
   account: { id: string } | undefined;
 }): boolean {
   if (!wallet) return true;
-  if (isWalletDeprecatedOrMocked(wallet)) return true;
+  if (wallet.isMocked) return true;
   if (isOthersWallet({ walletId: wallet.id ?? '' }) && !account) return true;
   return false;
 }


originalix:

@originalix

Reviewed commit `cfc7276569b7`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the account-availability boundary so deprecated hardware-wallet records remain selected and viewable, while missing or mocked wallets are rejected through the background account service. Account confirmation now reports success, and the wallet-details modal closes only after a successful selection; focused tests cover confirmation, initialization, auto-selection, and shared availability behavior. No cryptographic code changes, and existing background write restrictions remain, but the foreground policy currently treats historical visibility as full usability instead of establishing a read-only state.

**What needs attention:** Confirm that deprecated wallets can expose portfolio and history data without enabling transaction-capable controls, and that every interactive caller handles a rejected account selection.

### Issues to address

- **P1 · Deprecated wallets are exposed as fully usable**
  - **Impact:** When a deprecated hardware wallet is restored or selected, `hasNoUsableWallet` now reports it as usable. That predicate controls the normal Home and tab-header composition, and `HomeHeaderContainer` renders `WalletActions` such as Send, Swap, and Staking without a deprecated-wallet guard. Meanwhile, `ServiceAccountSelector` still treats the same wallet as unusable for address creation and other background helpers filter it. The result is not a read-only historical view: users can enter transaction-capable flows for a wallet that will fail later at a background or device boundary.
  - **Suggested action:** Separate wallet visibility from action eligibility: retain portfolio and history rendering for deprecated wallets, but gate `WalletActions` and other transaction-capable consumers with an explicit read-only or actionable-wallet predicate.
- **P2 · Other callers treat rejected selection as success**
  - **Impact:** `confirmAccountSelect` now returns `false` when the background lookup fails or the wallet is missing or mocked, but only `AccountSelectorAccountListItem` observes that result. `WebAccountPanelAccountList` still closes its panel, Universal Search and AddressInfo navigate away, and Reward Center can continue updating `focusedWallet` after rejection. The selected account may remain unchanged or become internally inconsistent while the UI tells the user the switch completed.
  - **Suggested action:** Propagate the boolean result through every interactive caller and execute close, navigation, recent-search, or focus-update side effects only after a `true` result; explicitly document fire-and-forget synchronization callers that intentionally ignore it.

### Validation gaps

- The focused command `yarn test packages/shared/src/utils/accountUtils.test.ts packages/kit/src/states/jotai/contexts/accountSelector/actions.test.tsx --runInBand` could not start because the workspace lacks Yarn's `node_modules` state file.
- Existing GitHub reviews, inline comments, general comments, and thread state could not be refreshed because the GitHub API connection was unavailable; inline wording therefore requires a duplicate-comment check.
- No end-to-end test verifies that a deprecated wallet shows historical assets while transaction-capable Home controls remain unavailable.

originalix (commented):
@originalix

Reviewed commit `d8833598fb63`.

- **P1 · Deprecated wallets are enabled outside read-only Home**
- **P2 · Deprecated Home header drops the network selector**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix (commented):
@originalix

Reviewed commit `ac70c61a9a5e`.

- **P1 · Non-Home cleanup still clears Home's allowed deprecated selection**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix (commented):
@originalix

Reviewed commit `b0567901fa00`.

- **P1 · Non-Home cleanup still clears Home's deprecated selection**
- **P2 · Historical wallet mode still exposes transaction actions**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix:
Addressed the two findings from the `d8833598` review in `b0567901`:

- Deprecated wallet selection is now explicitly limited to the Home scene; non-Home confirm and auto-select paths reject or replace it.
- The deprecated-wallet Home header now keeps the network selector in both compact and desktop layouts, covered by `WalletConnectionGroup.test.tsx`.

Validation: the targeted regression suite passes (5 suites, 68 tests), commit-level lint and TypeScript checks pass, and the 6.5.2 Desktop/iOS/Android bundle run [#31383631871](https://github.com/OneKeyHQ/app-monorepo/actions/runs/31383631871) completed successfully.

packages/kit/src/views/Home/pages/HomePageView.tsx
```diff
@@ -1096,19 +1097,30 @@ export function HomePageView({
     activeWalletId,
     walletListWalletIds,
   });
+  const walletContentMode = getHomePageWalletContentMode({
+    hasNoUsableWallet,
+    showNoWalletContent,
+    activeDeprecatedWalletCanShowAssets: Boolean(
+      wallet?.deprecated && !wallet.isMocked && account,
+    ),
+  });
```
sidmorizon: 🟠 P1: Home can be stuck on an endless loading spinner with no wallet switcher after picking a reset-device wallet

Severity: `severe`

Home treats a kept reset-device wallet as "still recovering" whenever it has no address on the current network (`activeDeprecatedWalletCanShowAssets` requires `account` at `packages/kit/src/views/Home/pages/HomePageView.tsx:1100-1106`), so the page spins forever and the account switcher disappears with it.
Impact: A user who selects a reset-device wallet and then views a network where that wallet has no address sees a permanently loading Home page and can no longer switch wallets or networks from the Home header.

<details>
<summary>Why auto-recovery no longer kicks in and the selector is hidden at the same time</summary>

- After this PR, `autoSelectNextAccount` no longer replaces a `deprecated` wallet in the `home` scene (`isWalletUnavailableForScene` returns false for home, `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:3508-3523`). With an `indexedAccount` present, `isAccountExist` is true, so `shouldAutoSelectNextAccount` is false and the state never self-heals.
- `accountUtils.hasNoUsableWallet` still returns true for any deprecated wallet (`packages/shared/src/utils/accountUtils.ts:884-895`).
- `shouldShowNoWalletContent` is false whenever other wallets exist in the wallet list, so `getHomePageWalletContentMode` falls through to `'recovering'` (`packages/kit/src/views/Home/pages/homePageNoWalletContent.ts:62-69`), which now renders an indefinite `<Spinner size="large" />` instead of the previous empty stack.
- The header uses the same `deprecated && !isMocked && account` predicate, so `WalletConnectionGroup` returns null (`packages/kit/src/components/TabPageHeader/components/WalletConnectionGroup.tsx:170-172`) and `AccountSelectorTriggerHome` returns null (`packages/kit/src/components/AccountSelector/AccountSelectorTrigger/AccountSelectorTriggerHome.tsx:49-51`), removing both the account and network selectors from Home.
- Trigger: deprecated wallets cannot create addresses, so any network without an already-created address for the selected indexed account (easily reached by switching networks after selecting the deprecated account, or by selecting an empty account since Home passes `allowSelectEmptyAccount`) produces `account === undefined`.

</details>

originalix: Fixed in `e91e95551f`.

- A deprecated, non-mocked Home wallet no longer requires a current-network account to render historical Home content.
- The Home account selector stays available, and `hideOnNoAccount` is disabled for the network selector only in this historical-wallet state.
- Deprecated/mocked wallets still hide `WalletActions`, so the fallback exposes navigation and read-only content without transaction entry points.

Regression coverage now includes a deprecated wallet with no account on the current network. The focused suite passes (5 suites, 73 tests), and `yarn agent:check --profile commit` passes.

originalix:
Addressed the latest review findings in `e91e95551f`:

- Non-Home cleanup now preserves a valid Home historical selection instead of syncing a cleared Swap identity back into Home.
- Deprecated/mocked Home wallets keep the overview, assets, history, account selector, and network selector, while the entire `WalletActions` row is hidden.
- A deprecated wallet with no account on the current network now renders the empty-account historical view instead of an endless recovery spinner; both recovery selectors remain available.

Validation: 5 focused Jest suites pass (73 tests), and `yarn agent:check --profile commit` passes. The hotfix/v6.5.2 bundle build is running at https://github.com/OneKeyHQ/app-monorepo/actions/runs/31398075701.

originalix (commented):
@originalix

Reviewed commit `a5135f85297b`.

- **P1 · Swap fallback can still overwrite the preserved Home wallet**
- **P2 · Historical wallets still receive signing-capable banners**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix:
TF follow-up on the deprecated-wallet network switch regression:

- Runtime logs confirmed Home only changed `networkId`; the selected deprecated hardware wallet was then mirrored into Swap.
- Swap correctly auto-selected a usable fallback, but that fallback was emitted back to Home and replaced the historical wallet selection.
- `a5135f8529` keeps a Swap fallback triggered by a deprecated wallet scene-local (`eventEmitDisabled: true`). Swap self-recovery remains enabled, while normal wallet/network switching behavior is unchanged.
- Regression coverage now asserts both the usable Swap fallback and the disabled Home-sync metadata. Focused account-selector/Home suites and `yarn agent:check --profile commit` pass.

Validation bundle: `6.5.2 / 1026081046 / 19148953`, head `a5135f8529`. Workflow attempt 2 completed successfully: https://github.com/OneKeyHQ/app-monorepo/actions/runs/31401983984

originalix (commented):
@originalix

Reviewed commit `02e21d7a241d`.

- **P1 · Historical wallets still expose account-mutation actions**
- **P2 · The new `confirmAccountSelect()` failure result is not handled by other scenes**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix (commented):
@originalix

Reviewed commit `1bee2f4e2d9f`.

- **P2 · Empty-wallet headers now retain blank connection rows**
- **P2 · Read-only historical wallets pollute actionable-wallet funnels**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/views/Home/pages/HomeHeaderContainer.tsx
```diff
@@ -44,14 +46,19 @@ function BaseHomeHeaderContainer() {
     }
     return false;
   }, [wallet]);
+  const isWalletUnavailable = accountUtils.isWalletDeprecatedOrMocked(wallet);
+  const showWalletActions = shouldShowHomeWalletActions({
+    isWalletNotBackedUp,
+    isWalletUnavailable,
+  });
```
sidmorizon: **Home screen reserves space for buttons that are no longer drawn for reset-device wallets on mobile**

Severity: `non-severe`

The reserved height for the mobile home header is still computed only from the backup state (`platformEnv.isNative && !isWalletNotBackedUp` at `packages/kit/src/views/Home/pages/HomeHeaderContainer.tsx:69`) instead of the new visibility flag, so for a reset-device wallet the page keeps space for action buttons that are now hidden.
Impact: Mobile users viewing a reset-device wallet see an empty gap under the balance where the action buttons used to be.

<details><summary>Incomplete propagation of the new showWalletActions gate</summary>

The PR introduces `showWalletActions` (`packages/kit/src/views/Home/pages/HomeHeaderContainer.tsx:50-53`) and applies it to the banner condition (`:61`) and to the `WalletActions` render (`:122`), but the native `minHeight` branch at `:68-71` was left keyed on `isWalletNotBackedUp` only.

For a deprecated/mocked wallet: `isWalletNotBackedUp === false`, so `nativeMinHeight` is set to 182 (the height that assumes `WalletActions` renders), while `showWalletActions === false` removes that block from the tree. `shouldShowBanner` is also false, so nothing fills the reserved space.

</details>

originalix (commented):
@originalix

Reviewed commit `141e40096fbc`.

- **P2 · Hidden actions leave a fixed native header gap**
- **P2 · Selection failures are not propagated to every caller**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/views/Home/pages/HomeHeaderContainer.tsx
originalix: Fixed in `d694524840`.

Native `minHeight` is now gated by `showWalletActions`, so a deprecated read-only wallet no longer reserves the 182pt action area when `WalletActions` is hidden.

Validation: 5 focused Jest suites pass (74 tests), and `yarn agent:check --profile commit` passes.

packages/kit/src/views/Home/pages/HomePageView.tsx
```diff
@@ -1096,19 +1097,30 @@ export function HomePageView({
     activeWalletId,
     walletListWalletIds,
   });
+  const walletContentMode = getHomePageWalletContentMode({
+    hasNoUsableWallet,
+    showNoWalletContent,
+    activeDeprecatedWalletCanShowAssets: Boolean(
+      wallet?.deprecated && !wallet.isMocked,
+    ),
+  });
```
sidmorizon: **Home shows a "no address" empty screen instead of assets for a reset-device wallet on the default All-Networks view**

Severity: `non-severe`

The Home page is only given asset content when a resolved account exists, but for a reset (deprecated) wallet the placeholder all-networks account is never built (`getMockedAllNetworkAccount` is skipped for deprecated/mocked wallets at `packages/kit-bg/src/services/ServiceAccountSelector.ts:432`), so the newly enabled path at `packages/kit/src/views/Home/pages/HomePageView.tsx:1100-1106` still lands on the empty "no address" screen.
Impact: Users who keep a reset hardware wallet selected still cannot see their historical balances while the default All-Networks view is active; they only see assets after manually switching to a single network.

<details>
<summary>Why the all-networks path never yields an account for deprecated wallets</summary>

- `getHomePageWalletContentMode` returns `'wallet'` when `activeDeprecatedWalletCanShowAssets` is true, so `walletPageContent` renders (`packages/kit/src/views/Home/pages/homePageNoWalletContent.ts:73-79`).
- Inside `homePageContent` (`packages/kit/src/views/Home/pages/HomePageView.tsx:1013-1030`) the branch `!account && !(vaultSettings?.mergeDeriveAssetsEnabled && networkAccounts?.length)` returns `emptyAccountView`.
- For all-networks, `activeAccount.account` is produced by `getMockedAllNetworkAccount`, which is gated behind `!isWalletUnusable` where `isWalletUnusable = accountUtils.isWalletDeprecatedOrMocked(wallet)` (`packages/kit-bg/src/services/ServiceAccountSelector.ts:428-437`). A deprecated wallet therefore always resolves `account === undefined` in all-networks mode.
- Result: the header/selector changes in this PR make the deprecated wallet visible and selectable, but the body still renders the empty-address view rather than the token/history tabs.

</details>

packages/kit/src/views/Home/pages/HomePageView.tsx
```diff
@@ -1096,19 +1097,30 @@ export function HomePageView({
...
     if (!ready) {
       return <TabPageHeader sceneName={sceneName} tabRoute={ETabRoutes.Home} />;
     }
 
-    let content = <Stack flex={1} />;
+    let content = (
+      <Stack flex={1} justifyContent="center" alignItems="center">
+        <Spinner size="large" />
+      </Stack>
+    );
```
sidmorizon: **Home can be left on a never-ending loading spinner when the selected wallet has no account for the current network**

Severity: `non-severe`

The wallet area now falls back to a永-running loading indicator (`'recovering'` mode at `packages/kit/src/views/Home/pages/HomePageView.tsx:1113-1117`) whenever no usable wallet is resolved but the empty-wallet screen is suppressed, so a state that never resolves shows a spinner forever instead of any content.
Impact: A user whose only selected wallet has no address on the chosen network sees Home spinning indefinitely, with no way to tell the page is not actually loading.

<details>
<summary>How the non-transient case is reached</summary>

- `hasNoUsableWallet` is true whenever the active wallet is an "others" (watch-only / imported / external) wallet and no `account` resolves for the current network (`packages/shared/src/utils/accountUtils.ts:884-895`).
- In that state the active wallet is neither deprecated nor mocked, so `shouldShowNoWalletContent` stays false (it requires `walletListResolvedNoWallet` or `activeWalletUnavailable`) — see `packages/kit/src/views/Home/pages/homePageNoWalletContent.ts:36-50`.
- `getHomePageWalletContentMode` therefore returns `'recovering'`, and the new default `content` is a large `Spinner` instead of the previous empty `Stack`.
- Auto-select only re-runs on mount and on wallet/account events; if no compatible account exists it cannot resolve the state, so the spinner persists.

A bounded fallback (timeout to the empty-wallet or empty-account view) would avoid the permanent loading appearance.

</details>

packages/kit/src/views/Home/pages/HomePageView.tsx
originalix: This path already builds the All Networks account before the cited guard, so no code change is needed for this finding.

`buildActiveAccountInfoFromSelectedAccount()` first calls `serviceAccount.getNetworkAccount()` when `deriveType`, `indexedAccountId`, and `wallet` are present. For an All Networks `networkId`, that reaches `getAccountsByIndexedAccounts()`, which returns `getMockedAllNetworkAccount()` directly. The later `!isWalletUnusable` guard only controls the fallback and `canCreateAddress`; it does not clear the account already returned by `getNetworkAccount()`.

This matches the TF runtime result: the deprecated wallet's aggregate multi-chain assets render in the All Networks view while address creation remains disabled.

packages/kit/src/views/Home/pages/HomePageView.tsx
originalix: Fixed in `69d27ddf62`.

Once a real, non-deprecated/non-mocked wallet has resolved, a missing current-network account now renders the existing `EmptyAccount` state instead of staying in the recovery spinner. Transient unavailable-wallet recovery and the true no-wallet state keep their existing behavior.

Validation: 5 focused Jest suites pass (75 tests), including the new resolved-wallet/no-account regression, and `yarn agent:check --profile commit` passes.

packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx
```diff
@@ -284,8 +293,19 @@ export function AccountSelectorAccountListItem({
...
+        accountSelectorUtils.isSceneAllowDeprecatedWalletSelection({
+          sceneName,
+        })) &&
+      (allowSelectEmptyAccount || !shouldShowCreateAddressButton),
+    [
+      allowSelectEmptyAccount,
+      focusedWalletInfo?.wallet,
+      sceneName,
+      shouldShowCreateAddressButton,
+    ],
```
sidmorizon: **Accounts from a reset device wallet become untappable when picking a recipient address**

Severity: `non-severe`

Accounts belonging to a reset-device wallet are made non-interactive outside the wallet home screen (`canConfirmAccountSelectPress` at `packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx:296-303`), so in flows like choosing a recipient address the row can be seen but tapping it does nothing at all.

Impact: A user picking one of their own historical addresses (e.g. in Send / address input) sees the account listed but cannot select it, with no message explaining why.

<details>
<summary>How the scene gate turns the list row into a dead element</summary>

`canConfirmAccountSelectPress` now additionally requires `!wallet.deprecated || accountSelectorUtils.isSceneAllowDeprecatedWalletSelection({ sceneName })`, and that helper returns true only for `EAccountSelectorSceneName.home` (`packages/shared/src/utils/accountSelectorUtils.ts:155-161`). When the flag is false, the `onPress` handler is not spread onto `ListItem` at all, so the row is silently inert (no toast, no disabled styling change beyond what already exists).

Deprecated wallets are still browsable in the selector sidebar (`WalletList/WalletListItem.tsx:274` only dims them), so this state is reachable from every non-home selector: `addressInput` (Send / BulkSend / ReferFriends / Redemption), `swap`, `discover`, `rewardCenter`.

The `addressInput` scene is the clearest regression: it is excluded from auto-select (`accountSelectorUtils.isSceneCanAutoSelect`), so before this PR selecting a deprecated wallet's account there worked and simply filled in a read-only address. The PR description also states the write boundary should reject "only missing or mocked target wallets", but `confirmAccountSelect` (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1419-1430`) additionally rejects deprecated wallets in every non-home scene.

Consider allowing deprecated-wallet selection in read-only scenes (at minimum `addressInput`), or surfacing an explanatory message instead of an inert row.

</details>

originalix: Fixed in `df81cf3cf5`.

Deprecated wallet selection is now allowed only in the two read-only contexts that need it: Home historical asset viewing and `addressInput` recipient picking. Swap, Discover, Reward Center, and other active-account scenes remain rejected by both the row gate and `confirmAccountSelect`.

Validation: the core action tests cover Home/addressInput success and Swap rejection; the account-row test covers recipient selection and modal close. Both focused suites pass (49 tests), and `yarn agent:check --profile commit` passes.

originalix (commented):
@originalix

Reviewed commit `69d27ddf6249`.

- **P1 · Deprecated Home remains interactive beyond the header**
- **P2 · Async wallet validation can reorder account selections**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix:
Follow-up on the latest automated review:

- Fixed the real request-order race in `a42fe800c4`. `confirmAccountSelect` now tracks the latest request per scene/sceneUrl/selector number and rejects stale completions before atom commit, persistence, event emission, or modal-close success handling. The new regression resolves the second lookup first and verifies the older lookup returns `selectionSuperseded` without replacing the newer selection. Focused result: 5 suites / 78 tests passed; `yarn agent:check --profile commit` passed.
- I did not expand this hotfix into a cross-module Home/Token/NFT/DeFi permission refactor. The PR contract is that deprecated accounts remain explicitly selectable for historical asset/address inspection, while signing, fresh-address creation, and hardware operations remain blocked at their existing background boundaries (including `ServiceAccount.getWalletDeviceParams` and `ServiceFreshAddress`). Swap and other active-account scenes also reject or replace the deprecated selection. Hiding every downstream navigation surface would change product behavior and materially widen OK-58315 beyond the requested minimal fix.

The final hotfix/v6.5.2 bundle for head `a42fe800c4` is running at https://github.com/OneKeyHQ/app-monorepo/actions/runs/31455368232.

originalix (commented):
@originalix

Reviewed commit `14935682beea`.

- **P1 · Deprecated Home remains interactive beyond the header**
- **P2 · Async wallet validation can reorder account selections**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -1375,6 +1412,64 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
         );
       }
 
+      const requestContextData = get(accountSelectorContextDataAtom());
+      const requestKey = `${requestContextData?.sceneName ?? ''}:${
+        requestContextData?.sceneUrl ?? ''
+      }:${num}`;
+      this.confirmAccountSelectRequestId += 1;
+      const requestId = this.confirmAccountSelectRequestId;
+      this.latestConfirmAccountSelectRequestIds.set(requestKey, requestId);
```
sidmorizon: **Account switch can be silently ignored right after a screen opens**

Severity: `non-severe`

The account switch is validated against a screen identity that is still empty (`get(accountSelectorContextDataAtom())` at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1415`) before that identity is filled in, so a legitimate switch can be silently dropped or overridden by an unrelated screen's switch.
Impact: Tapping an account immediately after a page opens can do nothing at all, with no error shown.

<details>
<summary>Mechanism: contextDataAtom is populated in an effect, so early confirmAccountSelect calls see undefined sceneName</summary>

`accountSelectorContextDataAtom` defaults to `undefined` (`packages/kit/src/states/jotai/contexts/accountSelector/atoms.ts:44`) and is only written from a mount effect in `AccountSelectorEffectsCmp` (`packages/kit/src/components/AccountSelector/AccountSelectorEffects.tsx:96-101`). The new guard in `confirmAccountSelect` reads it synchronously at the start of the call and uses it for two things:

1. `isSceneAllowDeprecatedWalletSelection({ sceneName: requestContextData?.sceneName })` — with `sceneName === undefined` this returns `false`, so a deprecated wallet account selected from Home before the effect flushes is rejected with `walletUnavailable` and the UI just does nothing (callers such as `AccountSelectorAccountListItem.tsx:452` and `AddressInfo/index.tsx:85` return early on failure).
2. `requestKey = "${sceneName ?? ''}:${sceneUrl ?? ''}:${num}"` — when the context data is missing the key degrades to `"::0"`, which is shared by every scene in that state, so two concurrent selections from different scenes can mark each other `selectionSuperseded`.

A safer approach is to fall back to the scene info available to the caller (or await the context data being set) instead of treating an unset context as "unknown scene".

</details>

originalix (commented):
@originalix

Reviewed commit `a42fe800c4d1`.

- **P1 · Deprecated Home remains interactive beyond the header**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
originalix: Fixed in `13db35f257`.

`useAccountSelectorActions` now injects `sceneName`/`sceneUrl` synchronously from the AccountSelector Provider config into `confirmAccountSelect`. The core action no longer treats the effect-populated `accountSelectorContextDataAtom` as authoritative for wallet eligibility or request identity, and only uses the atom with the same Provider config as a persistence fallback.

The regression now clears the context atom before selecting a deprecated Home account and verifies the selection still succeeds. Focused validation passes: 5 suites / 78 tests, plus `yarn agent:check --profile commit`.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -3457,14 +3574,21 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
-        // as needing replacement so the auto-select loop runs and either picks
-        // the next valid wallet or resets to undefined (OK-51091).
+        const isWalletUnavailableForScene = (walletToCheck?: IDBWallet) =>
+          Boolean(
+            walletToCheck?.isMocked ||
+            (walletToCheck?.deprecated &&
+              !accountSelectorUtils.isSceneAllowDeprecatedWalletSelection({
+                sceneName,
+              })),
+          );
```
sidmorizon: **Reset-device wallet gets kicked off the home screen after deleting any other wallet or account**

Severity: `non-severe`

The new scene-aware check treats a reset-device wallet as unusable whenever the scene is unknown (`isSceneAllowDeprecatedWalletSelection({ sceneName })` at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:3577-3584`), which happens on every removal-triggered re-selection, so the wallet the user deliberately picked to view old assets is silently swapped for another one.
Impact: A user viewing a reset-device wallet's historical assets is thrown onto a different wallet as soon as they delete any other wallet or account.

<details>
<summary>Why sceneName is missing on the removal paths</summary>

`autoSelectNextAccount` takes an optional `sceneName`. Every UI caller passes it, but the two internal callers do not: `removeWallet` (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:2586-2589`), `removeAccount` (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:2551-2556`) and the orphan-cleanup path (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1799-1802`) all call it with only `num` and `triggerBy`.

With `sceneName === undefined`, `accountSelectorUtils.isSceneAllowDeprecatedWalletSelection` returns `false` (`packages/shared/src/utils/accountSelectorUtils.ts:155-164`), so `isWalletUnavailableForScene(wallet)` is `true` for a `deprecated` wallet even when the active scene is Home. That forces `shouldAutoSelectNextAccount`, and the fallback loop excludes deprecated wallets, so the selection moves to a different wallet (or is reset to undefined) — exactly the behavior this PR intends to prevent on Home.

The scene is available inside the action via `this.getCurrentSceneInfo.call(set)` / the context-data atom, so these internal calls can forward the real scene name.

</details>

originalix (commented):
@originalix

Reviewed commit `13db35f25760`.

- **P1 · Deprecated Home content is still transaction-capable**
- **P1 · The Home scene is not a read-only capability boundary**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/components/AccountSelector/hooks/useAutoSelectAccount.tsx
```diff
@@ -50,6 +59,35 @@ export function useAutoSelectAccount({ num }: { num: number }) {
...
+    previousActiveWalletUnavailableRef.current = activeWalletUnavailable;
+
+    if (
+      !storageReady ||
+      !activeAccountReady ||
+      !activeWalletUnavailable ||
+      wasActiveWalletUnavailable
+    ) {
+      return;
+    }
```
sidmorizon: **Automatic recovery from an unusable wallet can be skipped, leaving the app stuck on a wallet that cannot be used**

Severity: `non-severe`

The new recovery effect records the wallet as "already handled" (`previousActiveWalletUnavailableRef.current = activeWalletUnavailable` at `packages/kit/src/components/AccountSelector/hooks/useAutoSelectAccount.tsx:65`) before it checks whether the selector is ready, so a wallet that becomes unusable during startup is never replaced by this effect.
Impact: In the window before the account selector finishes loading, switching onto a removed/reset wallet can leave the user on an unusable wallet with no automatic switch to a working one.

<details>
<summary>Why the transition is lost: ref write happens before the readiness early-return</summary>

The effect body first copies the current value into `previousActiveWalletUnavailableRef` and only afterwards returns early when `!storageReady || !activeAccountReady` (`packages/kit/src/components/AccountSelector/hooks/useAutoSelectAccount.tsx:62-89`). If `activeWalletUnavailable` flips from `false` to `true` while readiness is still false, the ref is updated to `true` but no auto-select is issued. Once readiness becomes true, the effect re-runs with `wasActiveWalletUnavailable === true`, so the recovery branch is skipped permanently for that wallet. Recovery then depends entirely on the separate mount effect (`:31-51`), which runs on the readiness change but is not guaranteed to observe the same state ordering (it is also gated on `sceneName`/`sceneUrl` identity and defers via `deferHeavyWorkUntilUIIdle` on Home).

Moving the ref assignment after the readiness guard keeps the false→true transition pending until the selector can actually act on it.

</details>

Suggested fix:

```ts
  useEffect(() => {
    if (!storageReady || !activeAccountReady) {
      return;
    }

    const wasActiveWalletUnavailable =
      previousActiveWalletUnavailableRef.current;
    previousActiveWalletUnavailableRef.current = activeWalletUnavailable;

    if (!activeWalletUnavailable || wasActiveWalletUnavailable) {
      return;
    }
```

originalix (commented):
@originalix

Reviewed commit `66d328d39dfc`.

- **P1 · Deprecated Home content is still transaction-capable**
- **P1 · The Home scene is not a read-only capability boundary**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

packages/kit/src/components/AddressInfo/index.tsx
```diff
@@ -77,6 +77,15 @@ function SwitchHomeAccountButton({
...
             }
 
+            const selectResult = await actions.current.confirmAccountSelect({
+              num: 0,
+              othersWalletAccount: indexedAccount ? undefined : account,
+              indexedAccount,
+            });
+            if (!selectResult.success) {
+              return;
+            }
```
sidmorizon: **Switching to an account of a reset hardware wallet from history or search silently does nothing**

Severity: `non-severe`

Switching the primary account is now abandoned whenever the selection is refused (`if (!selectResult.success) return;` at `packages/kit/src/components/AddressInfo/index.tsx:85-87`), but the screens that trigger it were not marked as allowed to pick reset-device wallets, so tapping confirm leaves the app exactly where it was with no message.
Impact: Users cannot jump to a historical (reset-device) wallet account from transaction history, universal search, or dApp alignment — the tap appears to do nothing.

<details>
<summary>Scene flag is read from the local provider config, and these home-scene mirrors were left without it</summary>

`confirmAccountSelect` now rejects a `deprecated` target wallet unless `accountSelectorUtils.isSceneAllowDeprecatedWalletSelection({ sceneName, allowDeprecatedWalletSelection })` returns true (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1463-1476`). The flag comes from `useAccountSelectorSceneInfo()` inside `useAccountSelectorActions` (`packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:3977-4036`), which reads the *nearest* `AccountSelectorProviderMirror` config, not the root store.

`HomePageContainer` was updated with `allowDeprecatedWalletSelection: true` (`packages/kit/src/views/Home/pages/HomePageContainer.tsx:78`), but these other home-scene mirrors were not:
- `packages/kit/src/components/AddressInfo/index.tsx:113-118` (SwitchHomeAccountContainer)
- `packages/kit/src/views/UniversalSearch/pages/UniversalSearch.tsx:1117-1121` (search results call `confirmAccountSelect` and now skip `navigation.pop()` on failure, see `packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchAddressItem.tsx:83-102`)
- `packages/kit/src/views/Discovery/components/SyncDappAccountToHomeProvider.tsx:166-170`

So for the same `home` scene the selection boundary behaves differently depending on the entry point, and because each caller now returns early on `success === false`, the modal/dialog is left open and no navigation happens.

</details>

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
originalix: Fixed in 7abb9a1949. autoSelectNextAccount now resolves the current Provider scene and deprecated-wallet capability at the action boundary even when removal callers omit sceneName. AccountSelectorEffects also keeps that capability in the context atom used by internal actions. The regression covers a deprecated Home selection handled after removeWallet without an explicit scene and verifies that the selection is preserved. Focused result: 5 suites / 76 tests passed; yarn agent:check --profile commit passed.

packages/kit/src/components/AccountSelector/hooks/useAutoSelectAccount.tsx
originalix: Verified on 7abb9a1949; no code change is required for this path. The mount recovery effect depends on both storageReady and activeAccountReady and runs autoSelectNextAccount when either becomes ready, regardless of the transition ref. Moving the ref assignment after the readiness guard would make the mount and transition effects both issue recovery on the ready render. A focused regression now makes the wallet unavailable while storage and active-account readiness are false, then enables both and verifies exactly one recovery call. Focused result: 5 suites / 76 tests passed; yarn agent:check --profile commit passed.

packages/kit/src/components/AddressInfo/index.tsx
originalix: Confirmed. This is a valid gap in the current head: these three callers switch the primary Home account, so they should opt into the same explicit read-only deprecated-wallet capability as the main Home selector instead of silently rejecting the selection. I am leaving this thread open until the AddressInfo, Universal Search, and dApp-to-Home provider configs and focused regression coverage are updated.

originalix (commented):
@originalix

Reviewed commit `7abb9a194923`.

- **P1 · Capability flag does not isolate the shared Home store**
- **P2 · Desktop token management bypasses historical read-only mode**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix (commented):
@originalix

Reviewed commit `2c59dd8f6d38`.

- **P1 · Concurrent selections can commit out of order**
- **P1 · Deprecated wallets enter action-capable Home state**
- **P2 · Selection failures still close interactive callers**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

originalix:
Review triage for head `2c59dd8f6d38`:

- **Accepted — P1 concurrent selections can commit out of order.** This is a real regression introduced by awaiting the background wallet lookup before the main-runtime atom commit. Two rapid selections can resolve in reverse order, so this finding remains actionable and is **not** marked won't fix.
- **Won't fix — dedicated read-only Home state.** The validated product behavior for OK-58315 is that a deprecated reset-device wallet remains selectable and can display its historical assets. This hotfix will not introduce a new cross-module Home/Earn/DApp permission architecture. Existing background boundaries still reject fresh-address creation and hardware signing/device access for deprecated wallets.
- **Won't fix for OK-58315 — update every interactive caller.** The affected Jira path (`AccountSelectorAccountListItem`) already keeps its modal open when selection is rejected and preserves the previous selected account. The other callers only observe the missing/mocked-wallet race window and do not overwrite selected state; changing every close/navigation flow plus adding per-pattern component suites would expand this hotfix. This can be tracked separately if the broader UX behavior needs standardization.

There are currently no unresolved inline review threads to resolve. The `6.5.2 / 1026081169 / 19209776` QA bundle for this head has been verified successfully by the requester.

packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx
```diff
@@ -1391,6 +1388,22 @@ class AccountSelectorActions extends ContextJotaiActionsBase {
...
+      }
+      if (!wallet || wallet.isMocked) {
+        return false;
+      }
+      if (
+        this.confirmAccountSelectLatestRequestIdMap.get(confirmRequestKey) !==
+        confirmRequestId
+      ) {
+        return false;
+      }
```
sidmorizon, line 1405: **A rejected account tap can silently cancel an account switch already in progress**

Severity: `non-severe`

A selection attempt claims the newest-request slot (`confirmAccountSelectLatestRequestIdMap.set` at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1386-1389`) before it is validated, so an attempt that is then rejected leaves an earlier, still-running valid switch marked as outdated and it is dropped without any change or feedback.
Impact: In rare cases a user's account switch never takes effect and nothing happens on screen.

<details>
<summary>Request-id bookkeeping vs. the new wallet validation guard</summary>

The request id for the scene/num key is incremented and stored at `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx:1383-1392`, before the new `serviceAccount.getWalletSafe` lookup. If a second `confirmAccountSelect` for the same scene+num starts while a first one is awaiting the background call (e.g. the dapp sync path fires `void actions.current.confirmAccountSelect(...)` at `packages/kit/src/views/Discovery/components/SyncDappAccountToHomeProvider.tsx:49` / `:64` while the user taps an account), the second one bumps the id. If that second target wallet is missing or mocked it returns `false` immediately without committing anything — but the id it consumed is never released, so the first (valid) call fails its own staleness check and returns `false` too. The UI then just returns without closing the modal (`packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/AccountSelectorAccountListItem.tsx:432-444`).

A fix is to validate the target wallet before claiming the request id, or to roll the stored id back to the previous value when returning early on a rejected wallet.

</details>

huhuanming (approved):

zhaono1 (approved):

originalix (commented):
@originalix

Reviewed commit `cfc7276569b7`.

- **P1 · Deprecated wallets are exposed as fully usable**
- **P2 · Other callers treat rejected selection as success**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12805#issuecomment-5237256678)

PR #12807, opened 2026-08-10
fix: prevent mixed TradingView history sources OK-59237

limichange:
[OK-59237](https://onekeyhq.atlassian.net/browse/OK-59237)

----------

## Summary

- keep the existing CoinGecko provider and full-history fallback behavior
- lock each TradingView Native Market interval to its first non-empty history source
- ignore late responses from the other source so Market and CoinGecko candles are never merged by timestamp

## Root cause

When a later Market history request returned no candles, the CoinGecko fallback page could be merged with already loaded Market candles. Concurrent requests could also return different sources and reach the same merge path.

## Validation

- `npx jest --runInBand packages/kit/src/components/TradingView/TradingViewNative/data/getTradingViewNativeSource.test.ts packages/kit/src/components/TradingView/TradingViewNative/data/providers/createTradingViewNativeDataProvider.test.ts packages/kit/src/components/TradingView/TradingViewNative/data/useTradingViewNativeKLine.test.ts` (84 tests passed)

[OK-59237]: https://onekeyhq.atlassian.net/browse/OK-59237?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/components/TradingView/TradingViewNative/data/providers/createTradingViewNativeDataProvider.test.ts
+++ b/packages/kit/src/components/TradingView/TradingViewNative/data/providers/createTradingViewNativeDataProvider.test.ts
@@ -553,7 +553,7 @@ describe('TradingViewNative data providers', () => {
     expect(mocks?.tokenInfoFetch).toHaveBeenCalledTimes(1);
   });
 
-  it('retries Market after a transient fallback succeeds', async () => {
+  it('keeps using CoinGecko after fallback selects the history source', async () => {
     const mocks = globalMockBag.__tradingViewNativeProviderMocks;
     mocks?.coinGeckoFetchChart.mockResolvedValue([[7200, 10]]);
     mocks?.marketFetchHistory.mockImplementation((params) =>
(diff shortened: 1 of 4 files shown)


originalix:

@limichange

Reviewed commit `19358525f4f1`. This report always reflects the latest reviewed changes.

### Review summary

The PR changes the foreground TradingViewNative history state machine so each market interval commits to either Market API history or CoinGecko fallback history for one provider lifecycle, preventing refresh, pagination, and concurrent requests from splicing the two sources. Source affinity is enforced in both the market provider and the hook, and the hook lock is recreated when the series provider changes so a returning series can select a source again. The related point-type classification still outlives that lifecycle, however, so a returning series that changes from primary OHLC data to single-value fallback data can retain the wrong chart style. Security-sensitive boundaries and realtime transport behavior remain unchanged.

**What needs attention:** Confirm that every source-derived state, especially chart-type classification, resets with the same provider lifecycle as history-source affinity.

### Issues to address

- **P2 · Reset chart-type classification with the source lifecycle**
  - **Impact:** When token A first loads primary OHLC history, the A/interval entry in historyPointTypeScopes becomes ohlc. After switching to another token and returning to A when the Market request instead falls back to CoinGecko, the newly recreated source-affinity map accepts the fallback response, but resolveHistoryPointTypeClassification reuses the old ohlc classification for that same seriesKey and interval. The single-value fallback data is therefore rendered as candlesticks rather than the intended line chart.
  - **Suggested action:** Reset the relevant historyPointTypeScopes classification when rawHistoryProvider starts a new lifecycle, and add an A primary → B → A fallback regression test that includes pointType: 'single' and asserts chartType is line.

### Validation gaps

- The targeted Jest suites could not start because this workspace lacks Yarn's `node_modules` state file.
- The added lifecycle test covers fallback → primary data acceptance but omits `pointType` metadata and the inverse primary → fallback chart-type transition.
- Existing GitHub review discussions could not be refreshed from this workspace, so the inline wording should be checked for duplication before posting.

originalix (commented):
@limichange

Reviewed commit `19358525f4f1`.

- **P2 · Reset chart-type classification with the source lifecycle**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12807#issuecomment-5237859475)

originalix (approved):

PR #12813, opened 2026-08-10
fix: correct iPad settings scrolling

huhuanming:
## Summary

- keep the first settings row below the native iPad navigation header
- remove the redundant page-level scroller so the inner settings list is the sole scroll owner on every platform
- preserve native header inset handling on the inner scroll view

## Root cause

The settings subpage combined a page-level `KeyboardAwareScrollView` with an inner vertical `ScrollView`. The nested scroll layout left the first row underneath the native iPad header and constrained the visible list height, even though the modal still had empty space.

## User impact

The iCloud backup entry is visible and accessible, all Backup settings including OneKey KeyTag fit in the initial iPad view, and the page no longer has nested vertical scrolling on other platforms.

## Validation

- verified the single-scroll layout on an iPad Pro 13-inch simulator running iOS 26.5
- verified the single-scroll layout on an iPhone 17 Pro simulator running iOS 26.5
- confirmed iCloud is positioned below the Backup header and remains clickable
- confirmed OneKey KeyTag is visible without scrolling

### DIFF
--- a/packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx
+++ b/packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx
@@ -7,7 +7,6 @@ import {
   XStack,
   YStack,
 } from '@onekeyhq/components';
-import platformEnv from '@onekeyhq/shared/src/platformEnv';
 
 import { useConfigContext } from './configContext';
 import { TabSettingsListGrid, TabSettingsSection } from './ListItem';
@@ -48,7 +47,7 @@ export function SubSettingsPage({
   }, [config?.configs]);
 
   return (
-    <Page scrollEnabled={!platformEnv.isNativeIOSPad}>
+    <Page>
       <Page.Header title={titleFromProps || config?.title} />
       <Page.Body>
         <ScrollView
--- a/packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx
+++ b/packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx
@@ -7,6 +7,7 @@ import {
   XStack,
   YStack,
 } from '@onekeyhq/components';
+import platformEnv from '@onekeyhq/shared/src/platformEnv';
 
 import { useConfigContext } from './configContext';
 import { TabSettingsListGrid, TabSettingsSection } from './ListItem';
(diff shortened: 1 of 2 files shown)


packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx
```diff
@@ -47,10 +47,11 @@ export function SubSettingsPage({
   }, [config?.configs]);
 
   return (
-    <Page scrollEnabled>
+    <Page>
```
sidmorizon, line 50: **Keyboard can cover text fields on the settings subpage on mobile**

Severity: `non-severe`

The page-level keyboard-aware scroller is removed (`<Page>` at `packages/kit/src/views/Setting/pages/Tab/SubSettingsPage.tsx:50`) while the remaining inner scroller has no keyboard avoidance, so on native devices the on-screen keyboard can hide the field being typed in.
Impact: On iOS/Android, text fields rendered inside settings subpages (e.g. the developer settings section) can end up behind the keyboard with no way to scroll them into view.

<details>
<summary>Loss of KeyboardAwareScrollView when Page.scrollEnabled is dropped</summary>

On native, `packages/components/src/layouts/Page/PageContainer.native.tsx:81-95` only wraps children in `KeyboardAwareScrollView` (with `bottomOffset={KEYBOARD_AWARE_SCROLL_BOTTOM_OFFSET}`) when the page context has `scrollEnabled` true. Removing `scrollEnabled` from `<Page>` means the subtree is rendered bare, and the inner `ScrollView` (`packages/components/src/layouts/ScrollView/index.tsx`) is a plain RN `ScrollView` with no keyboard offset logic. Settings subpages do render inputs, e.g. `packages/kit/src/views/Setting/pages/Tab/DevSettingsSection/index.tsx:811` and `.../AsyncStorageDevSettings.tsx:16`, reached through the config entry at `packages/kit/src/views/Setting/pages/Tab/config.tsx:870`. Impact is limited because these inputs are dev-only today, but any future input on a settings subpage inherits the regression.

</details>

huhuanming: Thanks for flagging this. No code change is needed here: tapping the empty area already dismisses the keyboard with the current ScrollView behavior, after which the content remains scrollable. Keeping a single scroll container is intentional to avoid reintroducing the nested-scroll issue fixed by this PR. The inputs mentioned here are also currently dev-only.

originalix (approved):

PR #12809, opened 2026-08-10
fix: harden Pro2/Neo recovery, BLE, and resource uploads

wabicai:
## 变更概览

本 PR 汇总处理 Pro2 / Neo 固件恢复、Portfolio 同步、BLE 钱包创建、设备管理、TON 日志与 Protocol V2 资源上传：

- 固件升级取消或异常退出后发出独立事件，并复用升级成功后的延迟重绑与状态刷新链路。
- 仅在升级恢复场景强制读取 WebUSB 实时状态，处理 modal 卸载与 onboarding 恢复焦点之间的竞态。
- Portfolio 在已打包重试前重新校验钱包、设备和 connectId 绑定，避免失效钱包继续上传旧包。
- BLE 钱包创建对齐 x 分支：沿用 Onboarding 已选通道，首次 BLE 创建持久化扫描 connectId，已有端点时优先使用 `bleConnectId`；移除额外 verified BLE 运行时证明和跨通道清理状态。
- 设备管理明确区分 hidden-wallet 代理与普通 mocked wallet；已移除/废弃的钱包不会重新出现在设备列表。
- TON mismatch 日志不再输出完整 BOC，避免目的地址、金额和 payload 上下文进入日志采集。
- 回滚 MV3 background / offscreen 全局二进制 codec 和 App 侧 JPEG→RGBA 解码。
- Portfolio、壁纸、NFT 改为传递 JSON-safe Base64，由 SDK LowLevel 完成严格校验、JPEG 解码、设备编码和分块写入。
- Protocol V2 壁纸/NFT 不再生成无用 hex 或旧 Pro 缩略图；旧 Pro/Touch 继续使用原有 `deviceUploadResource` hex 路径。
- 固件 V2/V3/V4 的 Extension 交接保持普通 JSON 参数，Desktop Bridge 的既有二进制固件路径不变。
- 本次新增的中文代码注释已统一改为英文。
- 所有 Hardware SDK 依赖（含 CLI）升级为已发布的 `1.2.0-alpha.104`。

## 设计结果

- Extension offscreen、iOS、Android、Desktop 的资源调用使用同一 Base64 API。
- App 负责下载、裁剪和生成 JPEG；SDK 负责协议相关解码、尺寸校验、设备格式转换和 FileWrite。
- Portfolio 服务保留授权、去重、代次控制、硬件忙重试与单设备串行保护。
[...]

PR #12814, opened 2026-08-10
fix: harden Pro2 portfolio binary synchronization

wabicai:
## What changed

- skip stale Pro2 portfolio synchronization work
- encode `ArrayBuffer`, typed-array views, and `Blob` values before the MV3 background/offscreen JSON bridge
- decode the tagged Base64 representation on both sides of the bridge
- preserve exact typed-array slice boundaries and reject malformed payloads before hardware I/O
- add regression coverage for Portfolio, wallpaper, NFT, firmware V1–V4, bootloader, and full resource uploads

## Why

Chrome Extension Port JSON serialization does not preserve native binary objects. Portfolio synchronization reached the SDK as an empty object and failed validation with `Unsupported FilesystemFileWrite data` before any device command was sent.

## Impact

MV3 plugin hardware calls can safely transport binary payloads through the offscreen boundary. The generic codec also protects wallpaper, NFT, and explicit firmware/resource binary inputs; non-MV3 direct calls retain their existing behavior.

## Validation

- offscreen codec Jest suite: 4 tests passed
- targeted ESLint passed
- App `tsc:only` passed
- `git diff --check` passed

### DIFF
--- a/apps/ext/src/offscreen/offscreenSetup.ts
+++ b/apps/ext/src/offscreen/offscreenSetup.ts
@@ -2,6 +2,10 @@
 import { bridgeSetup } from '@onekeyfe/extension-bridge-hosted';
 
 import type { IOffscreenApiMessagePayload } from '@onekeyhq/kit-bg/src/apis/IBackgroundApi';
+import {
+  decodeOffscreenApiPayload,
+  encodeOffscreenApiPayload,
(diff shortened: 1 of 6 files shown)


wabicai:
二进制桥接修复已追加到 #12809，本 PR 关闭以避免重复。

PR #12823, opened 2026-08-11
feat: add gas account funnel analytics

Kahnchan:
## Summary
- Add Gas Account funnel analytics for transaction confirmation, Swap, and Market Swap flows.
- Record decision context, sponsored user actions, shortage classification, failure stages, payer fallback, and Swap order IDs.
- Enrich the existing `insufficientFeeOnConfirm` event with the shared Gas Account context.

## Related Issue
- [OK-59972](https://onekeyhq.atlassian.net/browse/OK-59972)

## Intent & Context
The existing events could not reliably answer whether a user truly lacked network gas, whether Gas Account was supported and selected, why sponsorship was unavailable, or what happened after the user reached a confirmation or Swap review. Swap and Market Swap also lacked a direct link from Gas Account telemetry to the final Swap order.

This change adds a consistent client-side analytics contract without requiring server changes or a new flow ID.

## Design Decisions
- Build the public analytics properties from the final fee estimate and latest native-token balance so transaction confirmation, Swap, and Market Swap use the same shortage semantics.
- Classify shortages as `none`, `principal`, `extraFee`, `networkFee`, `mixed`, or `unknown` instead of treating every insufficient-balance state as missing gas.
- Emit `gasAccountAction` only for sponsored flows (`selectedPayer=gasAccount` or `effectiveFeePayer=megafuel`).
- Keep decision events at review readiness and action events at the actual user or submission transition.
- Reuse the Swap build response `orderId` to associate Gas Account decisions and actions with downstream Swap orders.
- Preserve the existing Gas Account fallback behavior and record automatic payer changes when a sponsored submission falls back to user-paid gas.

## Changes Detail
- Add shared Gas Account analytics types, context calculation, fiat buckets, and shortage classification.
- Add `gasAccountDecision` and `gasAccountAction` logger scenes.
- Instrument transaction confirmation actions and enrich `insufficientFeeOnConfirm`.
- Instrument direct Swap and Market Swap review, exit, success, failure, and fallback paths.
- Reuse the existing balance lookup result and add focused coverage for all three entry points.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension, Mobile, Desktop, Web
- **Risk Areas**: Fee-state timing, direct Swap submission/fallback paths, and event deduplication when review dialogs close.

## Test plan
- [x] Run 54 focused Jest tests for Gas Account context, Swap balance handling, review dialogs, and Market direct send.
- [x] Run `yarn agent:check --profile pr` local checks.
- [x] Run the full TypeScript check and type-aware lint for the changed files.

[OK-59972]: https://onekeyhq.atlassian.net/browse/OK-59972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/marketDirectSendTx.ts
+++ b/packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/marketDirectSendTx.ts
@@ -3,9 +3,15 @@ import BigNumber from 'bignumber.js';
 import type { IEncodedTx, IUnsignedTxPro } from '@onekeyhq/core/src/types';
 import backgroundApiProxy from '@onekeyhq/kit/src/background/instance/backgroundApiProxy';
 import {
-  EGasAccountErrorStrategy,
-  getGasAccountErrorEntry,
-} from '@onekeyhq/kit/src/views/SignatureConfirm/constants/gasAccountErrorCodes';
+  buildDirectSwapGasAccountAnalyticsContext,
+  buildDirectSwapGasAccountUiState,
+  runDirectSwapGasAccountStep,
+  sendDirectSwapWithGasAccountAnalytics,
+} from '@onekeyhq/kit/src/views/Swap/utils/gasAccountAnalytics';
+import {
+  buildNativeTokenFromGasInfo,
+  checkSwapLatestBalanceSufficient,
+} from '@onekeyhq/kit/src/views/Swap/utils/swapBalanceUtils';
 import type {
   IBuildUnsignedTxParams,
   ITransferInfo,
@@ -16,7 +22,7 @@ import {
   BATCH_SEND_TXS_FEE_UP_RATIO_FOR_SWAP,
 } from '@onekeyhq/shared/src/consts/walletConsts';
 import { OneKeyError } from '@onekeyhq/shared/src/errors';
-import { getGasAccountErrorCode } from '@onekeyhq/shared/src/errors/utils/gasAccountErrorUtils';
+import type { IGasAccountAnalyticsContext } from '@onekeyhq/shared/src/logger/scopes/transaction/types';
 import { calculateFeeForSend } from '@onekeyhq/shared/src/utils/feeUtils';
 import { applyCustomPriorityFeeToGasInfo } from '@onekeyhq/shared/src/utils/marketPresetFeeUtils';
 import type {
@@ -30,7 +36,6 @@ import type {
   IFeeTron,
   IFeeUTXO,
   IGasAccountQuote,
-  IGasAccountUiState,
   IGasEIP1559,
   IGasLegacy,
   IGasPayer,
@@ -69,6 +74,10 @@ type IMarketDirectSendParams = {
   customPriorityFee?: IMarketPresetPriorityFeeOverride;
   tronResourceRentalInfo?: ITronResourceRentalInfo;
   useDefaultRpc?: boolean;
+  gasAccountAnalytics?: {
+    fiatCurrency?: string;
+    useGasAccountByDefault?: boolean;
+  };
 };
 
 type IEstimateMarketDirectGasInfosParams = Omit<
@@ -155,6 +164,7 @@ function buildGasInfo(
     payer?: IGasPayer;
     gasAccountEligible?: boolean;
     gasAccountQuote?: IGasAccountQuote;
+    gasAccountScenarioReason?: string;
   },
   gasCommon: {
     baseFee?: string;
@@ -191,6 +201,7 @@ function buildGasInfo(
     payer: gasRes.payer,
     gasAccountEligible: gasRes.gasAccountEligible,
     gasAccountQuote: gasRes.gasAccountQuote,
+    gasAccountScenarioReason: gasRes.gasAccountScenarioReason,
   };
 }
 
@@ -918,10 +929,12 @@ export async function estimateMarketDirectGasInfos({
   approveUnsignedTxArr,
   networkFeeLevel,
   customPriorityFee,
+  gasAccountAnalytics,
 }: IEstimateMarketDirectGasInfosParams): Promise<{
   gasInfos: IMarketGasInfoEntry[];
   gasFeeFiatValue?: string;
   preparedUnsignedTx: IUnsignedTxPro;
+  gasAccountAnalyticsContext?: IGasAccountAnalyticsContext;
 }> {
   if (!accountId || !networkId || !accountAddress) {
     throw new OneKeyError('account error');
@@ -949,10 +962,39 @@ export async function estimateMarketDirectGasInfos({
     customPriorityFee,
   });
 
+  let gasAccountAnalyticsContext: IGasAccountAnalyticsContext | undefined;
+  const swapGasInfo = findGasInfo(gasInfos, unsignedTx.encodedTx);
+  if (gasAccountAnalytics && swapGasInfo && unsignedTx.swapInfo) {
+    const nativeToken = buildNativeTokenFromGasInfo({
+      gasInfo: swapGasInfo.gasInfo,
+      networkId,
+      fromToken: unsignedTx.swapInfo.sender.token,
+    });
+    const latestNativeBalance = nativeToken
+      ? await checkSwapLatestBalanceSufficient({
+          token: nativeToken,
+          amount: '0',
+          accountAddress,
+          accountId,
+        })
+      : undefined;
+    gasAccountAnalyticsContext = buildDirectSwapGasAccountAnalyticsContext({
+      entryPoint: 'marketSwapDirect',
+      networkId,
+      unsignedTx,
+      gasInfo: swapGasInfo.gasInfo,
+      estimateFeeParams: swapGasInfo.estimateFeeParams,
+      nativeBalance: latestNativeBalance?.balance,
+      useGasAccountByDefault: gasAccountAnalytics.useGasAccountByDefault,
+      fiatCurrency: gasAccountAnalytics.fiatCurrency,
+    });
+  }
+
   return {
     gasInfos,
     gasFeeFiatValue: buildGasFeeFiatValue(gasInfos),
     preparedUnsignedTx: unsignedTx,
+    gasAccountAnalyticsContext,
   };
 }
 
@@ -1001,6 +1043,7 @@ export async function estimateMarketApproveGasInfos({
 }
 
 async function updateUnsignedTxAndSendTx({
+  accountAddress,
   accountId,
   networkId,
   unsignedTxItem,
@@ -1008,7 +1051,9 @@ async function updateUnsignedTxAndSendTx({
   estimateFeeParams,
   tronResourceRentalInfo,
   useDefaultRpc,
+  gasAccountAnalytics,
 }: {
+  accountAddress: string;
   accountId: string;
   networkId: string;
   unsignedTxItem: IUnsignedTxPro;
@@ -1016,6 +1061,7 @@ async function updateUnsignedTxAndSendTx({
   estimateFeeParams?: IEstimateFeeParams;
   tronResourceRentalInfo?: ITronResourceRentalInfo;
   useDefaultRpc?: boolean;
+  gasAccountAnalytics?: IMarketDirectSendParams['gasAccountAnalytics'];
 }): Promise<ISendTxOnSuccessData> {
   const feeInfo = buildMarketGasInfoFeeInfo(gasInfo);
 
@@ -1027,20 +1073,6 @@ async function updateUnsignedTxAndSendTx({
       feeInfo,
       tronResourceRentalInfo,
     });
-
-  await backgroundApiProxy.serviceSend.precheckUnsignedTxs({
-    networkId,
-    accountId,
-    unsignedTxs: [updatedUnsignedTxItem],
-    precheckTiming: ESendPreCheckTimingEnum.Confirm,
-  });
-
-  await backgroundApiProxy.serviceSignatureConfirm.preActionsBeforeSending({
-    accountId,
-    networkId,
-    unsignedTxs: [updatedUnsignedTxItem],
-    tronResourceRentalInfo,
-  });
 
   const {
     totalNative,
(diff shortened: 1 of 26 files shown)


originalix:

@Kahnchan

Reviewed commit `c789bd5bd096`. This report always reflects the latest reviewed changes.

### Review summary

This PR moves Gas Account and MegaFuel telemetry ownership into the foreground transaction-review layer while retaining fee estimation, signing, and broadcast in existing background services. Shared logger contracts normalize sponsorship decisions and actions across Signature Confirm, direct Swap, and Market speed-swap; Market also captures the native balance during review and forwards that snapshot to submission. The package boundary and helper reuse are sound, and no secret-bearing telemetry is introduced, but three state transitions still produce inconsistent funnel data across entry points.

**What needs attention:** Confirm that backend rejection, sponsored-to-user fallback, and review confirmation transitions retain one consistent analytics lifecycle across TxConfirm, direct Swap, and Market review flows.

### Issues to address

- **P2 · Fallback retries lose their terminal outcome**
  - **Impact:** After a sponsored TxConfirm submission falls back to user-paid, the live analytics context is rewritten to a non-sponsored state. The retry still broadcasts, but submitSucceeded or submitFailed is filtered out, leaving the recorded payer transition without the transaction's final outcome.
  - **Suggested action:** Preserve the sponsored context for the active review session and use it for the retry result while overriding only effectiveFeePayer to user.
- **P2 · Backend rejection reasons remain classified as unknown**
  - **Impact:** When fee estimation returns gasAccountScenarioReason without an eligible quote, TxConfirm retains gasAccountSupported as null even though the backend supplied an explicit rejection. Direct Swap marks the same response unsupported, so equivalent backend outcomes produce different analytics dimensions.
  - **Suggested action:** Set gasAccountSupported to false when gasAccountScenarioReason is present and add a focused test for that response shape.
- **P2 · Direct swap review confirmations never emit confirmClicked**
  - **Impact:** The standard Swap and Market review handlers only mark their session submitted. This suppresses exited but never emits confirmClicked, so analytics cannot distinguish a confirm followed by precheck or send failure from a review that was never confirmed.
  - **Suggested action:** Have the shared submission marker emit confirmClicked from the session analytics context before setting submitted, then cover both callers with a focused test.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` completed without errors.
- The focused Jest command covering the changed analytics, balance, review-dialog, and Market direct-send tests could not start because Yarn reported a missing `node_modules` state file.
- Existing GitHub discussions and thread resolution could not be inspected because the GitHub API was unreachable, so inline wording should receive a human check before posting.
- No live `main`/`bg` telemetry trace was captured; event lifecycle conclusions are based on the static state transitions.

packages/kit/src/views/SignatureConfirm/hooks/useGasAccountAnalyticsContext.ts
```diff
@@ -0,0 +1,149 @@
...
+        unavailableReason = 'privateSend';
+      } else if (disabledForBatch) {
+        unavailableReason = 'batchTransaction';
+      } else if (gasAccountTemporarilyDisabled) {
+        unavailableReason = 'temporarilyDisabled';
+      } else {
+        unavailableReason =
+          gasAccountUiState.gasAccountScenarioReason ?? 'backend_unknown';
+      }
+    }
```
sidmorizon, line 105: **Sponsored-gas reporting blames the wrong cause when a user's own network endpoint is active**

Severity: `non-severe`

The new gas sponsorship reporting ignores the case where a user-configured network endpoint turns sponsorship off (`clientUnsupported` at `packages/kit/src/views/SignatureConfirm/hooks/useGasAccountAnalyticsContext.ts:65-66`), so those transactions are recorded as if sponsorship was requested and unexplainably refused by the server.
Impact: Funnel data will over-count unexplained sponsorship failures and hide the real cause, misleading anyone reading the metrics.

<details>
<summary>Why the custom-RPC gate is missing from the analytics context</summary>

`packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx:544-620` computes `isCustomRpcEnabled = !!(customRpcInfo?.rpc && customRpcInfo?.enabled)` and treats it exactly like `sponsorDisabledForBatch` / `sponsorDisabledForPrivateSend` / `gasAccountDisabledByScenario`: it forces `effectiveFeePayer='user'` and calls `resetGasAccountUiState()`.

The new hook only mirrors three of those four gates (`disabledByScenario`, `isPrivateSend`, `disabledForBatch`). With a custom RPC enabled and the setting on, `gasAccountRequested` stays `true`, `gasAccountSupported` stays `null`, and `unavailableReason` falls through to `gasAccountUiState.gasAccountScenarioReason ?? 'backend_unknown'` (`useGasAccountAnalyticsContext.ts:84-99`) even though the client itself disabled sponsorship.

A `useCustomRpcStatusAtom` value is already available in this store, or the same `customRpcInfo` lookup used by `TxFeeInfo` can be reused.
</details>

Kahnchan: Fixed in `71a79c6ae5`. The fee estimator now persists the custom-RPC sponsorship gate in the confirmation state, and the analytics context treats it as client-unsupported with `gasAccountRequested=false`, `gasAccountSupported=false`, and `unavailableReason=customRpcEnabled` instead of `backend_unknown`. I also added coverage for the new unavailable reason.

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/marketDirectSendTx.ts
```diff
@@ -1054,37 +1086,75 @@ async function updateUnsignedTxAndSendTx({
...
+      })
+    : undefined;
+  const latestNativeBalance = nativeToken
+    ? await checkSwapLatestBalanceSufficient({
+        token: nativeToken,
+        amount: '0',
+        accountAddress,
+        accountId,
+      })
+    : undefined;
```
sidmorizon: **Every market swap now waits for an extra balance lookup before the transaction is signed**

Severity: `non-severe`

An extra remote token-balance lookup is awaited (`checkSwapLatestBalanceSufficient` at `packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/marketDirectSendTx.ts:1098`) before the swap is checked and signed, even though its result is only used for tracking, so every swap takes longer to go out and sponsored swaps have less time left on their fee quote.
Impact: Users see a slower confirmation, and sponsor-paid swaps are more likely to hit an expired-quote error right before broadcast.

<details>
<summary>Where the extra round trip is inserted in the send path</summary>

`updateUnsignedTxAndSendTx` runs for every unsigned tx that carries `swapInfo` (`marketDirectSendTx.ts:1089`), i.e. all direct market swaps and wraps, regardless of whether Gas Account sponsorship is involved. `checkSwapLatestBalanceSufficient` issues `serviceSwap.fetchSwapTokenDetails` (and possibly `serviceToken.getNativeTokenAddress`) over the network (`packages/kit/src/views/Swap/utils/swapBalanceUtils.ts:371-422`). It is passed `amount: '0'`, so its sufficiency verdict is always `true`; only `balance` is consumed, purely to populate `nativeBalance` on the analytics context.

The call sits between the fresh sponsor fee estimate (`sendMarketDirectUnsignedTxs` deliberately re-runs estimate-fee "right before sending so the broadcast uses a fresh, non-expired gasAccountQuote.quoteId", `marketDirectSendTx.ts:1204-1224`) and `precheckUnsignedTxs`/`signAndSendTransaction`, so it directly consumes quote validity time. The same pattern was added to the preview path at `marketDirectSendTx.ts:973-980`.

A cheaper option is to reuse the balance already fetched during review, or to fire the lookup without blocking the send (and skip it entirely when the context would not be logged).

</details>

originalix (commented):
@Kahnchan

Reviewed commit `71a79c6ae50b`.

- **P2 · Fallback retries lose their terminal outcome**
- **P2 · Backend rejection reasons remain classified as unknown**
- **P2 · Direct swap review confirmations never emit confirmClicked**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12823#issuecomment-5249153453)

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/marketDirectSendTx.ts
Kahnchan: Fixed in c789bd5bd0. Market Swap now reuses the native balance already fetched for the review snapshot instead of awaiting another token-details request after the fresh sponsor estimate. The send path still rebuilds the analytics context from the latest gas info and quoteId, so the event fields remain current while signing is no longer delayed. Added tests for both preview balance capture and zero send-time balance lookups.

packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx
```diff
@@ -607,7 +607,12 @@ function TxFeeInfo(props: IProps) {
           gasAccountDisabledByScenario
         ) {
           resetGasAccountUiState();
-          if (
+          if (isCustomRpcEnabled) {
+            updateGasAccountUiState({
+              payer: 'user',
+              sponsorDisabledByCustomRpc: true,
+            });
+          } else if (
```
sidmorizon, line 615: **Gas sponsorship reporting stays marked as blocked after the user turns off their custom network endpoint**

Severity: `non-severe`

The "sponsorship blocked by custom RPC" marker is only ever turned on (`updateGasAccountUiState({ sponsorDisabledByCustomRpc: true })` at `packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx:611-614`) and is never turned back off when a later fee estimate finds sponsorship available, so the confirmation screen keeps reporting sponsorship as unavailable even while it is actively being used.
Impact: Funnel data for sponsored transactions is wrong for users who switch off a custom network endpoint on the confirmation screen, under-counting successful sponsorship.

<details>
<summary>Why the flag goes stale: the eligible branch never resets it</summary>

`TxFeeInfo` only clears gas-account UI state via `resetGasAccountUiState()` inside the disabled branch and in the final `else` branch (`packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx:609-660`). The `else if (r.gasAccountEligible && r.gasAccountQuote)` branch calls `updateGasAccountUiState({...})` without a preceding reset, so `sponsorDisabledByCustomRpc` keeps the `true` value set by an earlier estimate.

This is reachable from the confirmation page itself: `showCustomRpcFallbackDialog` can permanently disable the network's custom RPC (`packages/kit/src/views/Send/components/CustomRpcFallbackDialog.tsx:72-79`), after which a re-estimate takes the eligible branch. `useGasAccountAnalyticsContext` then reads the stale flag (`packages/kit/src/views/SignatureConfirm/hooks/useGasAccountAnalyticsContext.ts:65-79`) and computes `clientUnsupported=true` → `gasAccountRequested=false` → `gasAccountEligible=null`, contradicting the live quote that is actually sponsoring the transaction.

</details>

originalix (commented):
@Kahnchan

Reviewed commit `c789bd5bd096`.

- **P2 · Fallback retries lose their terminal outcome**
- **P2 · Backend rejection reasons remain classified as unknown**
- **P2 · Direct swap review confirmations never emit confirmClicked**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12823#issuecomment-5249153453)

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -695,39 +718,54 @@ export function useSwapBuildTx() {
...
+          : undefined);
+      if (!nativeToken) {
+        return { isSufficient: true };
       }
 
       const checkResult = await checkSwapLatestBalanceSufficient({
-        token: nativeBalanceRequirement.token,
-        amount: nativeBalanceRequirement.amount,
+        token: nativeToken,
+        amount: nativeBalanceRequirement?.amount ?? '0',
```
sidmorizon, line 737: # [🟡 P2] 直发 Swap 发送路径新增了一次仅供埋点使用的余额网络请求

## 问题

Market Swap 已在 `c789bd5bd0` 改为复用 review 阶段的余额，但 `useSwapBuildTx` 的直发路径仍然会多打一次网络请求。改动后 `checkLatestNativeTokenBalance` 在 `nativeBalanceRequirement` 为 `undefined` 时不再直接返回，而是用 `buildNativeTokenFromGasInfo` 兜底出 native token，并以 `amount: '0'` 调用 `checkSwapLatestBalanceSufficient`；同时 `swapBalanceUtils.ts` 把零金额短路条件从 `amountBN.lte(0)` 放宽成 `amountBN.lt(0)`，所以 `amount='0'` 也会真正走 `serviceToken.getNativeTokenAddress` + `serviceSwap.fetchSwapTokenDetails`。

这条分支恰好命中 Gas Account 赞助场景：`getSwapRequiredNativeBalanceAmount` 对 `gasAccountEligible` / `megafuel` 的 tx 不计网络费，from token 又不是原生币时 `requiredAmount` 为 `0`，因此返回 `undefined`。改动前这里是 0 次网络请求，改动后固定多 1~2 次。而调用点位于 `updateUnsignedTxAndSendTx` 内、`precheckUnsignedTxs` / `verifyTransaction` / `signAndSendTransaction` 之前，上游 `estimateFee` 刚刚以 `lockedUserNonce` 绑定了新的 `gasAccountQuote.quoteId`：

```ts
const gasRes = await backgroundApiProxy.serviceGas.estimateFee({ ... gasAccountEnabled: isGasAccountEnabled, lockedUserNonce: ... });
lastTxRes = await updateUnsignedTxAndSendTx({ ... }); // 内部先 await 余额请求，再 precheck/sign/broadcast
```

## 影响

赞助交易的 quote 有效期被这次纯埋点请求额外消耗一个网络往返（弱网下可达秒级），签名与广播整体延后；一旦 quote 过期，`signAndSendTransaction` 会返回 gas account 错误码，按 `EGasAccountErrorStrategy.Fallback` 回落成用户自付 gas 重发，用户实际支付了本应被赞助的网络费。这与 `sendMarketDirectUnsignedTxs` 中「right before sending so the broadcast uses a fresh, non-expired gasAccountQuote.quoteId」的设计意图相反，也是 Market Swap 那条已修复问题的同类根因，只是位置不同。

## 建议

与 Market 保持一致：把 `nativeBalance` 从 review 阶段（`preSwapBeforeStepActions` 里已经拿到的 `checkLatestNativeBalanceRes.nativeBalance`）透传到发送阶段，发送路径不再触发新的余额查询；或者保留 `checkLatestNativeTokenBalance` 原有的 `if (!nativeBalanceRequirement) return { isSufficient: true }` 早退，仅在 review 阶段单独取一次余额供埋点使用。若要保留 `lt(0)` 的放宽，也建议在 `checkSwapLatestBalanceSufficient` 内对 `amount === '0'` 的调用方显式标注用途，避免 `checkOtherFee` 等其它调用方在零金额时被动多发请求。

[Claude session ↗](https://claude.ai/code/session_019a19KHBZRkSdevZ2Zxs5bu)

---

packages/kit/src/views/SignatureConfirm/components/SignatureConfirmAlert/TxConfirmAlert.tsx
```diff
@@ -68,9 +76,14 @@ function TxConfirmAlert(props: IProps) {
     const isInsufficient =
       sendTxStatus.isInsufficientNativeBalance ||
       sendTxStatus.isInsufficientTokenBalance;
-    if (isInsufficient && !insufficientFeeLoggedRef.current) {
+    if (
+      isInsufficient &&
+      gasAccountAnalyticsContext &&
```
sidmorizon, line 81: # [🟡 P2] 已有的 `insufficientFeeOnConfirm` 事件被新增的上下文条件收窄，费用估算未成功时会永久漏报

## 问题

上报条件从 `isInsufficient && !insufficientFeeLoggedRef.current` 变成了额外要求 `gasAccountAnalyticsContext` 有值。而 `useGasAccountAnalyticsContext` 只有在 `txFeeInfoInit && sendFeeStatus.status === ESendFeeStatus.Success && sendSelectedFeeInfo && !nativeTokenInfo.isLoading` 全部满足时才返回对象，否则返回 `undefined`。

但余额不足标记并不依赖 `ESendFeeStatus.Success`：`TxFeeInfo.tsx` 里计算 `isInsufficientNativeBalance` 的 effect 只判断 `if (!txFeeInfoInit) return;`，而估算失败分支同样会执行 `updateTxFeeInfoInit(true)` 后把状态置为 `ESendFeeStatus.Error`（含 `stopPolling === true` 时直接停止轮询）。此时 `sendSelectedFeeInfo` 保留上一次结果或为 `undefined`，余额不足告警照常渲染，但埋点上下文恒为 `undefined`。

## 影响

在「估算首次就失败」或「服务端 `stopPolling` 后不再重试」的场景下，`insufficientFeeLoggedRef.current` 始终保持 `false`，`insufficientFeeOnConfirm` 这个本 PR 之前无条件上报的既有事件永远不会发出。这恰好是本次漏斗最需要统计的人群（余额不足时 EVM 估算经常因 `insufficient funds` 类错误直接失败），会让新旧数据无法对齐，并低估余额不足的真实规模。

## 建议

把 Gas Account 上下文当作可选增强，而不是上报前置条件，例如：

```ts
if (isInsufficient && !insufficientFeeLoggedRef.current) {
  insufficientFeeLoggedRef.current = true;
  defaultLogger.transaction.send.insufficientFeeOnConfirm({
    ...(gasAccountAnalyticsContext ?? buildFallbackGasAccountContext()),
    network: networkId,
    // ...
  });
}
```

若不便构造兜底上下文，也可以让 `IGasAccountAnalyticsContext` 相关字段在该事件里保持可选，缺失时按 `shortageType: 'unknown'` / `fiatValueAvailable: false` 上报，保证事件本身不丢。

[Claude session ↗](https://claude.ai/code/session_019a19KHBZRkSdevZ2Zxs5bu)

---

originalix (approved):

PR #12827, opened 2026-08-11
fix: self-heal unreadable SimpleDB records and support clearing Perps data (OK-59997)

Minnzen:
[OK-59997](https://onekeyhq.atlassian.net/browse/OK-59997)

----------

## Problem

A desktop 6.5.0 user's Perps chart and token selector were permanently stuck on "Loading tokens..." (OK-59997). Chromium stores large IndexedDB values as external blob files; once the blob backing `simple_db_v5:perp` is corrupted (crash/power loss mid-write), every read rejects with `UnknownError: Failed to read large IndexedDB value`. Because `setRawData(builder)` reads the old record before writing, all writes fail too, so the record can never be repaired — `getTradingUniverse()` and chart symbol resolution both depend on it.

## Fix

- `SimpleDbEntityBase.getRawData`: catch `UnknownError`/`NotReadableError` storage read errors, remove the unreadable record, and return null so builder-based writes rebuild it from empty. Other errors still propagate. Self-heals on the first read after startup.
- Settings → Clear cache: new "Perps" item (reuses existing `global.perp` translation) calling `perp.clearRawData()` as a support escape hatch.

Recovery cost for already-affected users only: Hyperliquid terms re-accept and local toggles like skipOrderConfirm reset. Funds, positions, open orders, and agent approval are unaffected.

## Test

- New unit tests pin the contract: self-heal returns null and drops the record; builder write rebuilds after a read failure; non-whitelisted errors propagate without deleting.
- A/B: same assertions fail on the unfixed code (both read and write reject) and pass with the fix. `yarn jest packages/kit-bg/src/dbs/simple` 34/34.
- Deterministic manual repro for QA (corrupt the profile's IndexedDB blob files) is documented in OK-59997.

[OK-59997]: https://onekeyhq.atlassian.net/browse/OK-59997?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
+++ b/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
@@ -15,19 +15,16 @@ type ISimpleDbEntitySavedData<T> = {
   updatedAt: number;
 };
 
-// Chromium rejects reads with these signatures when a value's external blob
-// file is corrupted (e.g. crash mid-write); the record stays unreadable
-// forever. UnknownError alone is not enough — Chromium also uses it for
-// transient IO failures (disk full, backing store open errors), so it must
-// carry the corrupted-blob message to qualify.
+// Chromium rejects reads with exactly this signature when a value's external
+// blob file is corrupted (e.g. crash mid-write); the record then stays
+// unreadable forever. Match nothing broader: UnknownError without this
+// message and NotReadableError both cover transient IO conditions where
+// deleting would lose recoverable data (OK-59997).
 function isUnreadableStorageValueError(error: unknown): boolean {
   const { name, message } = (error ?? {}) as {
     name?: string;
     message?: string;
   };
-  if (name === 'NotReadableError') {
-    return true;
-  }
   return (
     name === 'UnknownError' &&
     Boolean(message?.includes('Failed to read large IndexedDB value'))
--- a/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
+++ b/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
@@ -14,6 +14,13 @@ type ISimpleDbEntitySavedData<T> = {
   data: T;
   updatedAt: number;
 };
+
+// Chromium rejects reads with these names when a value's external blob file
+// is corrupted (e.g. crash mid-write); the record stays unreadable forever.
+function isUnreadableStorageValueError(error: unknown): boolean {
+  const name = (error as { name?: string } | null | undefined)?.name;
+  return name === 'UnknownError' || name === 'NotReadableError';
+}
 abstract class SimpleDbEntityBase<T> {
   // Do not use appStorageInstance directly, use this.appStorage instead
   appStorage: AsyncStorageStatic =
@@ -55,7 +62,18 @@ abstract class SimpleDbEntityBase<T> {
     }
     this.cachedRawDataPromise = (async () => {
       dbPerfMonitor.logSimpleDbCall('getRawData', this.entityName);
-      const savedDataStr = await this.appStorage.getItem(this.entityKey);
+      let savedDataStr: string | null = null;
+      try {
+        savedDataStr = await this.appStorage.getItem(this.entityKey);
+      } catch (error) {
+        if (!isUnreadableStorageValueError(error)) {
+          throw error;
+        }
+        console.error(error);
+        // Drop the dead record so builder-based setRawData can rebuild it; use
+        // appStorage directly — clearRawData() would deadlock on the shared mutex.
+        await this.appStorage.removeItem(this.entityKey).catch(() => undefined);
+      }
       let updatedAt = 0;
       // @ts-ignore
       let data: T | undefined | null;
--- a/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
+++ b/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
@@ -15,11 +15,23 @@ type ISimpleDbEntitySavedData<T> = {
   updatedAt: number;
 };
 
-// Chromium rejects reads with these names when a value's external blob file
-// is corrupted (e.g. crash mid-write); the record stays unreadable forever.
+// Chromium rejects reads with these signatures when a value's external blob
+// file is corrupted (e.g. crash mid-write); the record stays unreadable
+// forever. UnknownError alone is not enough — Chromium also uses it for
+// transient IO failures (disk full, backing store open errors), so it must
+// carry the corrupted-blob message to qualify.
 function isUnreadableStorageValueError(error: unknown): boolean {
-  const name = (error as { name?: string } | null | undefined)?.name;
-  return name === 'UnknownError' || name === 'NotReadableError';
+  const { name, message } = (error ?? {}) as {
+    name?: string;
+    message?: string;
+  };
+  if (name === 'NotReadableError') {
+    return true;
+  }
+  return (
+    name === 'UnknownError' &&
+    Boolean(message?.includes('Failed to read large IndexedDB value'))
+  );
 }
 abstract class SimpleDbEntityBase<T> {
   // Do not use appStorageInstance directly, use this.appStorage instead
@@ -46,6 +58,10 @@ abstract class SimpleDbEntityBase<T> {
 
   updatedAt = 0;
 
+  // Bumped on every persisted write so a failing read can tell whether a
+  // concurrent setRawData landed after it started.
+  private writeSeq = 0;
+
   @backgroundMethod()
   clearRawDataCache() {
     this.cachedRawData = null;
@@ -62,6 +78,7 @@ abstract class SimpleDbEntityBase<T> {
     }
     this.cachedRawDataPromise = (async () => {
       dbPerfMonitor.logSimpleDbCall('getRawData', this.entityName);
+      const writeSeqBefore = this.writeSeq;
       let savedDataStr: string | null = null;
       try {
         savedDataStr = await this.appStorage.getItem(this.entityKey);
@@ -69,10 +86,24 @@ abstract class SimpleDbEntityBase<T> {
         if (!isUnreadableStorageValueError(error)) {
           throw error;
         }
-        console.error(error);
-        // Drop the dead record so builder-based setRawData can rebuild it; use
-        // appStorage directly — clearRawData() would deadlock on the shared mutex.
-        await this.appStorage.removeItem(this.entityKey).catch(() => undefined);
+        try {
+          // One retry separates transient IO failures from true corruption.
+          savedDataStr = await this.appStorage.getItem(this.entityKey);
+        } catch (retryError) {
+          if (!isUnreadableStorageValueError(retryError)) {
+            throw retryError;
+          }
+          console.error(retryError);
+          // Drop the dead record so builder-based setRawData can rebuild it;
+          // use appStorage directly — clearRawData() would deadlock on the
+          // shared mutex. Skip if a write landed after this read started:
+          // deleting would erase that fresh value.
+          if (this.writeSeq === writeSeqBefore) {
+            await this.appStorage
+              .removeItem(this.entityKey)
+              .catch(() => undefined);
+          }
+        }
       }
       let updatedAt = 0;
       // @ts-ignore
@@ -148,6 +179,7 @@ abstract class SimpleDbEntityBase<T> {
           ? (savedData as any)
           : JSON.stringify(savedData),
       );
+      this.writeSeq += 1;
 
       this.updatedAt = updatedAt;
       return data;
--- a/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
+++ b/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
@@ -41,6 +41,11 @@ abstract class SimpleDbEntityBase<T> {
 
   abstract readonly enableCache: boolean;
 
+  // Deleting an unreadable record is only safe for entities whose data can be
+  // fully rebuilt (OK-59997's perp cache); user-authored entities must keep
+  // failing loudly instead, so self-heal is opt-in per entity.
+  protected readonly enableUnreadableRecordSelfHeal: boolean = false;
+
   get entityKey() {
     return `${SIMPLE_DB_KEY_PREFIX}:${this.entityName}`;
   }
@@ -60,11 +65,16 @@ abstract class SimpleDbEntityBase<T> {
   private writeSeq = 0;
 
   // Writes currently awaiting setItem; a failing read must not delete while
-  // one is in flight, whichever of the two started first.
+  // one is in flight or was in flight when the read began.
   private pendingWrites = 0;
+
+  // Bumped by clearRawDataCache so a read that was already in flight when the
+  // clear happened cannot re-publish the stale value into the memory cache.
+  private readGeneration = 0;
 
   @backgroundMethod()
   clearRawDataCache() {
+    this.readGeneration += 1;
     this.cachedRawData = null;
     this.cachedRawDataPromise = null;
   }
@@ -80,11 +90,16 @@ abstract class SimpleDbEntityBase<T> {
     this.cachedRawDataPromise = (async () => {
       dbPerfMonitor.logSimpleDbCall('getRawData', this.entityName);
       const writeSeqBefore = this.writeSeq;
+      const pendingWritesBefore = this.pendingWrites;
+      const readGenerationBefore = this.readGeneration;
       let savedDataStr: string | null = null;
       try {
         savedDataStr = await this.appStorage.getItem(this.entityKey);
       } catch (error) {
-        if (!isUnreadableStorageValueError(error)) {
+        if (
+          !this.enableUnreadableRecordSelfHeal ||
+          !isUnreadableStorageValueError(error)
+        ) {
           throw error;
         }
         try {
@@ -97,11 +112,15 @@ abstract class SimpleDbEntityBase<T> {
           console.error(retryError);
           // Drop the dead record so builder-based setRawData can rebuild it;
           // use appStorage directly — clearRawData() would deadlock on the
-          // shared mutex. Skip if a write is in flight or started after this
-          // read began: deleting would erase that fresh value. (A write that
-          // starts after this check wins anyway — same-store ops keep issue
-          // order, so its setItem lands after this removeItem.)
-          if (this.pendingWrites === 0 && this.writeSeq === writeSeqBefore) {
+          // shared mutex. Any write overlapping this read vetoes the delete:
+          // pending at read start, still pending now, or started since. (A
+          // write starting after this check wins anyway — same-store ops keep
+          // issue order, so its setItem lands after this removeItem.)
+          if (
+            pendingWritesBefore === 0 &&
+            this.pendingWrites === 0 &&
+            this.writeSeq === writeSeqBefore
+          ) {
             await this.appStorage
               .removeItem(this.entityKey)
               .catch(() => undefined);
@@ -138,7 +157,10 @@ abstract class SimpleDbEntityBase<T> {
         }
       }
       this.updatedAt = updatedAt ?? 0;
-      if (this.enableCache) {
+      // Skip the cache when clearRawData ran while this read was in flight,
+      // otherwise the stale value would resurrect the just-cleared record on
+      // the next builder-based write.
+      if (this.enableCache && this.readGeneration === readGenerationBefore) {
         this.cachedRawData = data;
       }
       return data;
--- a/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
+++ b/packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
@@ -58,9 +58,13 @@ abstract class SimpleDbEntityBase<T> {
 
   updatedAt = 0;
 
-  // Bumped on every persisted write so a failing read can tell whether a
-  // concurrent setRawData landed after it started.
+  // Bumped when a persisted write starts so a failing read can tell whether a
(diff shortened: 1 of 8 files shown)


packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
```diff
@@ -55,7 +62,18 @@ abstract class SimpleDbEntityBase<T> {
...
+        savedDataStr = await this.appStorage.getItem(this.entityKey);
+      } catch (error) {
+        if (!isUnreadableStorageValueError(error)) {
+          throw error;
+        }
+        console.error(error);
+        // Drop the dead record so builder-based setRawData can rebuild it; use
+        // appStorage directly — clearRawData() would deadlock on the shared mutex.
+        await this.appStorage.removeItem(this.entityKey).catch(() => undefined);
+      }
```
sidmorizon, line 129: 🟠 P1: Saved app data can be deleted permanently when a storage read fails temporarily

Reported at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:68-76` (outside the diff of this PR, so this is a file-level comment)

Severity: `severe`

Any stored record whose read fails with a generic storage error is permanently deleted (`removeItem` at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:75`) without any retry or restriction to a specific record, so transient disk/IO failures can wipe user-created data such as the address book, bookmarks or watchlists.
Impact: A one-off storage hiccup can silently and irreversibly erase user-saved lists instead of just failing that one read.

<details>
<summary>Why a generic error name is not proof of permanent corruption</summary>

`isUnreadableStorageValueError` (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:20-23`) matches only on the DOMException name. Chromium raises `UnknownError` for a broad range of IndexedDB-level internal failures (internal error opening the backing store, disk-full while committing a transaction — see the comments already in `packages/shared/src/storage/WebStorage.ts:104-108`), not only for a corrupted external blob file. Because `SimpleDbEntityBase` is the base class for every simple-db entity (`packages/kit-bg/src/dbs/simple/entity/*`, including `SimpleDbEntityAddressBook`, `SimpleDbEntityBrowserBookmarks`, `SimpleDbEntityMarketWatchListV2`), one transient `UnknownError` on any of those reads deletes irreplaceable user data. The stated bug (OK-59997) only concerns the `perp` entity, which is fully rebuildable.

</details>

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
```diff
@@ -55,7 +62,18 @@ abstract class SimpleDbEntityBase<T> {
...
+      try {
+        savedDataStr = await this.appStorage.getItem(this.entityKey);
+      } catch (error) {
+        if (!isUnreadableStorageValueError(error)) {
+          throw error;
+        }
+        console.error(error);
+        // Drop the dead record so builder-based setRawData can rebuild it; use
+        // appStorage directly — clearRawData() would deadlock on the shared mutex.
+        await this.appStorage.removeItem(this.entityKey).catch(() => undefined);
```
sidmorizon: **A freshly saved record can still be deleted by the corrupted-record cleanup**

Severity: `non-severe`

The safety check that is supposed to stop the cleanup from erasing a just-saved record only counts saves that have already finished (`this.writeSeq += 1` at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:182`), so a save that is still in progress is invisible to the check and the record it writes can be wiped.
Impact: Data a user just saved can silently disappear while the app is recovering from an unreadable record.

<details>
<summary>Why the writeSeq guard misses in-flight writes</summary>

`getRawData` snapshots `writeSeq` before its read (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:81`) and only skips the self-heal `removeItem` when `this.writeSeq === writeSeqBefore` (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:101-105`). `setRawData` increments `writeSeq` only after `await this.appStorage.setItem(...)` resolves (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:176-182`). If a `setRawData` issues `setItem` while a failing read is in flight, the read's retry can reject and evaluate the guard before the write resolves; `writeSeq` is unchanged, so `removeItem` is issued and, ordered after the completed `setItem` in the storage queue, deletes the fresh value. The new test only covers the case where `setRawData` fully resolves before the read rejects, so this window is untested. Marking the write as pending before `setItem` (and clearing it after) would close the window; the self-heal delete could also be executed under the entity mutex from outside the mutex-holding path.

</details>

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Fixed in aff785521b: `UnknownError` now must carry the `Failed to read large IndexedDB value` message (the corrupted-blob signature — disk-full / backing-store-open failures do not match), and the read is retried once before deleting so a transient IO failure never triggers the self-heal. Covered by new unit tests: transient failure recovers via retry and keeps the record; `UnknownError` with an unrelated message propagates without deleting.

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Fixed in aff785521b: `getRawData` captures a write sequence number before reading, and `setRawData` bumps it after every persisted write; the self-heal delete is skipped when the sequence changed, so a write that lands during a slow failing read can no longer be erased. Covered by a race unit test (write parked mid-read, delete skipped, record kept).

Minnzen:
Re: automated review report (reviewed commit `ddfefa9a05e8` — stale, both findings addressed on the current head):

- **P1 (lost-update race)**: fixed in `aff785521b` before this report was posted. `setRawData` bumps a write sequence number after every persisted write; the self-heal delete in `getRawData` captures it before reading and is skipped when it changed, so a write that lands during a slow failing read cannot be erased. Covered by a dedicated race unit test.
- **P2 (live deposit-token cache survives the Perps clear)**: fixed in `391b9b3e00`. New `serviceWebviewPerp.clearPerpsDepositTokenListRuntimeCache()` clears the in-memory map and bumps every known write generation so in-flight fetches skip their simpleDb write-back; `clearCacheOnApp` calls it before `perp.clearRawData()`.

`yarn jest packages/kit-bg/src/dbs/simple` 37/37 passing.

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Follow-up for the updated finding (in-flight writes invisible to the writeSeq guard) — fixed in ba85435bf0: `setRawData` now also tracks a `pendingWrites` counter around `setItem`, and the self-heal delete is vetoed while any write is in flight, whichever of the two started first (`pendingWrites === 0 && writeSeq unchanged`). A write starting after that check still wins because same-store ops keep issue order. Added a unit test that parks a write inside `setItem` while a failing read completes — the delete is skipped and the record survives (38/38 passing).

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
```diff
@@ -55,7 +82,35 @@ abstract class SimpleDbEntityBase<T> {
...
+          // use appStorage directly — clearRawData() would deadlock on the
+          // shared mutex. Skip if a write is in flight or started after this
+          // read began: deleting would erase that fresh value. (A write that
+          // starts after this check wins anyway — same-store ops keep issue
+          // order, so its setItem lands after this removeItem.)
+          if (this.pendingWrites === 0 && this.writeSeq === writeSeqBefore) {
+            await this.appStorage
+              .removeItem(this.entityKey)
+              .catch(() => undefined);
+          }
```
sidmorizon, line 127: **Locally stored app data can be permanently deleted after a temporary disk read problem**

Severity: `non-severe`

Saved records are deleted (`removeItem` at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:108-110`) whenever two back-to-back reads fail with a `NotReadableError`, which browsers also raise for temporary disk/IO problems, so recoverable read hiccups can wipe stored data instead of surfacing an error.
Impact: Users can silently lose locally stored data such as bookmarks, custom tokens/RPCs and other saved settings when storage is momentarily unreadable.

<details>
<summary>Why the NotReadableError branch is broader than the corrupted-blob case it targets</summary>

`isUnreadableStorageValueError` (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:23-34`) requires the corrupted-blob message for `UnknownError` — exactly because the comment notes Chromium reuses `UnknownError` for transient IO failures — but accepts *any* `NotReadableError` regardless of message. Firefox/Chromium also raise `NotReadableError` for general IO failures (profile on an unavailable volume, file temporarily locked, encrypted storage not yet unlocked). The only safeguard is one immediate retry with no backoff (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:93-95`), which will almost certainly fail again during the same short-lived condition, and the code then deletes the record.

This affects every `SimpleDbEntityBase` subclass, not only the Perps entity, including entities holding user-authored data (browser bookmarks, custom tokens, custom RPCs).

</details>

originalix (commented):
@Minnzen

Reviewed commit `ba85435bf086`.

- **P2 · Perps reset misses fetches before cache-key registration**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12827#issuecomment-5251097710)

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Fixed in 77275e05ff: dropped `NotReadableError` from the whitelist entirely. Self-heal now matches only the exact signature observed in the incident (`UnknownError` carrying `Failed to read large IndexedDB value`), which Chromium emits specifically for a corrupted external-blob read; transient IO conditions (locked file, unavailable volume, encrypted storage not yet unlocked) no longer qualify and simply propagate. Residual unreadable-record cases that do not match this signature are covered by the manual Settings → Clear cache → Perps escape hatch. Added a unit test asserting `NotReadableError` propagates without deleting (39/39 passing).

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
```diff
@@ -55,7 +79,35 @@ abstract class SimpleDbEntityBase<T> {
...
```
sidmorizon: **A just-saved value can be deleted when a read fails right as the save completes**

Severity: `non-severe`

A freshly saved value can be deleted by the repair step (`removeItem` at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:105-107`) when the save started before the failing read and finished during it, so recently saved data is silently lost.
Impact: In a narrow timing window, data the user just saved (e.g. Perps settings/terms acceptance) is wiped instead of preserved.

<details>
<summary>Guard only covers writes in flight at check time or started after the read began</summary>

`writeSeqBefore` is captured at the start of the read (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:82`). If a `setRawData` already bumped `writeSeq` and incremented `pendingWrites` (`SimpleDbEntityBase.ts:179-180`) *before* the read started, then:

- `writeSeqBefore` already equals the bumped value, so `this.writeSeq === writeSeqBefore` is true;
- if the `setItem` completes while the two failing `getItem` attempts are outstanding, `pendingWrites` is back to 0 by the time the guard runs (`SimpleDbEntityBase.ts:104`).

Both conditions pass and `removeItem` deletes the value the write just persisted. The reads can legitimately fail in this window because they were issued against the pre-write (corrupted) snapshot. The in-code comment claims the guard skips deletion whenever "a write is in flight or started after this read began", but the write-started-before-and-finished-during case is not covered. A snapshot of `pendingWrites`/`writeSeq` taken at read start (and treating any write overlap as a veto) would close the hole.

</details>

originalix (commented):
@Minnzen

Reviewed commit `77275e05ff3d`.

- **P2 · Perps reset misses fetches before generation capture**
- **P2 · An in-flight read can restore data after clear**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12827#issuecomment-5251097710)

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Fixed in 2ac4cf365d: the read now snapshots `pendingWrites` at its start, and the delete is vetoed when a write overlapped the read in any way — pending at read start (your scenario: started before, finished during), still pending at check time, or started since (`writeSeq` change). A write starting after the check wins via same-store issue ordering. Added a unit test reproducing exactly this window: write parked in `setItem` before the read begins, released mid-read — delete skipped, value kept.

Minnzen:
Hardening round in `2ac4cf365d` consolidating remaining review feedback plus an external model review:

1. **Self-heal is now opt-in per entity — only `simpleDb.perp` enables it.** All other entities (address book, bookmarks, custom tokens/RPCs, …) keep the original fail-loud behavior on unreadable records; the delete path can no longer touch user-authored data regardless of how Chromium classifies the error.
2. **Any write overlapping a failing read vetoes the delete**: `pendingWrites` is snapshotted at read start, so a write that started before the read and finished during it is covered too.
3. **`clearRawData` vs in-flight read (pre-existing bug)**: a read generation counter now prevents a read that predates the clear from re-publishing the stale value into the memory cache.
4. **Perps runtime-cache clear epoch**: fetches capture a clear epoch before their first await, so requests still in owner normalization (cache key not yet registered) also skip their write-back after a clear.

Each fix has a dedicated unit test; `yarn jest packages/kit-bg/src/dbs/simple` 42/42 passing, tsc clean.

packages/kit-bg/src/services/ServiceWebviewPerp/ServiceWebviewPerp.ts
sidmorizon, line 1: **Clearing Perps data can be immediately undone by a request that was already running**

Reported at `packages/kit-bg/src/services/ServiceWebviewPerp/ServiceWebviewPerp.ts:612-616` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

The freshly fetched deposit-token list is still saved back into the perps record (`isPerpsDepositTokenListWriteGenerationCurrent({ cacheKey, writeGeneration })` at `packages/kit-bg/src/services/ServiceWebviewPerp/ServiceWebviewPerp.ts:612-616`) without checking the clear marker, so a request already in progress when the user clears Perps data can re-create the record right after it is wiped.
Impact: The "Clear Perps data" recovery action may appear to do nothing because the record reappears seconds later.

<details>
<summary>Missing clearEpoch in the persistence guard</summary>

`clearPerpsDepositTokenListRuntimeCache` (`packages/kit-bg/src/services/ServiceWebviewPerp/ServiceWebviewPerp.ts:459-473`) bumps `perpsDepositTokenListClearEpoch` precisely so that in-flight fetches which have not yet registered their cache key (they are still awaiting `normalizePerpsDepositAllNetworksOwner`) skip their simpleDb write-back. Such a request captures `writeGeneration` at `ServiceWebviewPerp.ts:749-750` — i.e. *after* the clear bumped generations — so the generation check alone always passes, and only the `clearEpoch` comparison added in `isPerpsDepositTokenListWriteGenerationCurrent` (`ServiceWebviewPerp.ts:448-457`) can veto it. The call site inside `fetchPerpsDepositTokenListDataCached` still passes only `{ cacheKey, writeGeneration }`, omitting `clearEpoch`, so this veto never applies to the simpleDb persistence path (the atom path at `ServiceWebviewPerp.ts:652` passes the full options object and is correct).

</details>

Suggested fix:

```ts
        if (
          this.isPerpsDepositTokenListWriteGenerationCurrent({
            cacheKey: options.cacheKey,
            writeGeneration: options.writeGeneration,
            clearEpoch: options.clearEpoch,
          })
        ) {
```

Minnzen: Fixed in e541baaa8b: the simpleDb write-back call site now passes `clearEpoch` through to `isPerpsDepositTokenListWriteGenerationCurrent`, matching the atom path — this was also the cause of the CI lint (tsc) failure, which the same commit fixes. Full `tsc --noEmit` clean, 42/42 tests passing.

originalix (commented):
@Minnzen

Reviewed commit `2ac4cf365d99`.

- **P1 · Perps persistence guard omits the reset epoch**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12827#issuecomment-5251097710)

Minnzen:
Re: updated auto-review report (reviewed `2ac4cf365d`, one commit behind): the P1 (missing `clearEpoch` in the persistence guard inside `fetchPerpsDepositTokenListDataCached`) is fixed in `e541baaa8b` — same commit that fixed the CI tsc failure. All `isPerpsDepositTokenListWriteGenerationCurrent` call sites now receive the full `IPerpsDepositTokenListWriteOptions`, and the required property makes any future omission a compile error. Full `tsc --noEmit` clean, 42/42 unit tests passing.

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
sidmorizon, line 1: **A slow read finishing after a save can put the old value back into memory, silently undoing the save**

Reported at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:191-194` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

The freshly saved value is overwritten by an older in-flight read (`this.cachedRawData = data` at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:163-165`) because only clearing, not saving, invalidates reads that started earlier, so a later update can rebuild from the outdated value and wipe the save.
Impact: A setting or cached record a user just changed can silently revert to its previous value.

<details>
<summary>Read/write ordering: readGeneration only guards clearRawData, not setRawData</summary>

`clearRawDataCache()` bumps `readGeneration` (`packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:76-80`) and the read publishes to the memory cache only when the generation still matches (`:163`). `setRawData` also invalidates the in-flight read by setting `this.cachedRawDataPromise = null` (`:194`) and writes `this.cachedRawData = data` (`:191-193`), but it does not bump `readGeneration`.

Sequence (single runtime, bg): (1) `getRawData()` starts and its `getItem` is slow; (2) `setRawData(newValue)` runs inside the mutex, sets `cachedRawData = newValue`, nulls `cachedRawDataPromise`, persists; (3) the older read resolves with the pre-write string and, since `readGeneration` is unchanged, overwrites `cachedRawData` with the stale value. A subsequent builder-based `setRawData` then reads that stale cache (`:84-86`, `:185`) and persists it, losing the earlier write.

The same class of hazard the PR fixed for `clearRawData` therefore remains open for `setRawData`; bumping `readGeneration` in `setRawData` closes it.

</details>

Suggested fix:

```ts
      if (this.enableCache) {
        this.cachedRawData = data;
      }
      // A read that started before this write must not re-publish its stale
      // value into the memory cache after we set the fresh one.
      this.readGeneration += 1;
      this.cachedRawDataPromise = null;
```

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
```diff
@@ -55,7 +89,44 @@ abstract class SimpleDbEntityBase<T> {
...
+            pendingWritesBefore === 0 &&
+            this.pendingWrites === 0 &&
+            this.writeSeq === writeSeqBefore
+          ) {
+            await this.appStorage
+              .removeItem(this.entityKey)
+              .catch(() => undefined);
+          }
+        }
+      }
```
sidmorizon, line 129: **Required bug-fix retrospective entry is missing**

Severity: `non-severe`

The repository rule that every bug fix must append a case entry to `.claude/skills/1k-retrospective/references/case-studies.md` is not satisfied by this fix (see the new self-heal path at `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts:96-129`), so the required record of the incident is absent.
Impact: The team loses the mandated written record of this defect and how it was caught.

<details>

<summary>Rule source: .cursor/rules/self-testing.mdc "Bug Fix Case Recording"</summary>

`.cursor/rules/self-testing.mdc` (alwaysApply: true) states: "After every **bug fix**, append to `.claude/skills/1k-retrospective/references/case-studies.md`" with Symptom / Root Cause / Fix / Catchable by fields. This PR is a bug fix (OK-59997) but does not touch that file.

</details>

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Fixed in 13306e0873: `setRawData` now also bumps the read generation, so a read that was in flight when a save landed can no longer publish the pre-save value into the memory cache — same mechanism that already guards `clearRawData`. Added a unit test: slow read parked, save completes, released read must not revert the cached value (43/43 passing).

packages/kit-bg/src/dbs/simple/base/SimpleDbEntityBase.ts
Minnzen: Added in 13306e0873: case entry appended to `.claude/skills/1k-retrospective/references/case-studies.md` with Symptom / Root Cause / Fix / Catchable-by for the OK-59997 blob-corruption incident.

originalix (approved):

PR #12829, opened 2026-08-11
fix: align firmware update device details(OK-59803)

originalix:
[OK-59803](https://onekeyhq.atlassian.net/browse/OK-59803)

----------

## Summary
- Keep firmware current and target versions on one line, including Protocol V2 SafeOS/component rows.
- Show the custom device name in the update header and place the BLE name immediately after it on every platform.
- Fall back to the stable model label only when the device has no custom name.

## Intent & Context
OK-59803 was reopened because Android still displayed the firmware version transition on multiple lines. The original issue also requires iOS and desktop to show the custom device name and requires iOS BLE names, including legacy devices, to follow the device name instead of being right-aligned.

## Root Cause
The Protocol V2 changelog row and the shared firmware progress row both allowed their version containers to wrap. In addition, Protocol V2 devices always replaced the custom device name with a stable model label, while iOS used a separate header layout that moved the BLE name to `headerRight`.

## Design Decisions
- Use one shared header layout across platforms so device-name and BLE-name ordering cannot diverge.
- Preserve the custom name whenever available and use the model label only as a missing-data fallback.
- Remove wrapping from both the Protocol V2-specific and shared version rows so all update paths follow the same single-line requirement.
- Keep the change in the main UI runtime; no firmware detection, background runtime, or native transport behavior is changed.

## Changes Detail
- Update firmware title selection and add Pro2, Neo, and legacy-device coverage.
- Consolidate the firmware update header and remove the iOS-only right-aligned BLE label.
- Prevent wrapping in the changelog and shared version progress components.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Android, iOS, Desktop, Extension
- **Risk Areas**: Very long custom device names may truncate to preserve the adjacent BLE name and single-line header layout.

## Test plan
- [x] Run `yarn jest packages/kit/src/views/FirmwareUpdate/utils.test.ts --runInBand` (16 tests passed).
- [x] Verify the Protocol V2 changelog with a mock on a Pixel 9 API 36 emulator.
- [ ] Verify custom device name and adjacent BLE name with iOS and desktop hardware update flows.

[OK-59803]: https://onekeyhq.atlassian.net/browse/OK-59803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wabicai (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

PR #12830, opened 2026-08-11
fix: handle onboarding device scan errors

wabicai:
## Summary

- add a compatible `onError` channel to `DeviceScannerUtils`
- normalize unsuccessful scan responses through `convertDeviceError`
- catch rejected scan promises and stop polling cleanly
- handle `link disabled` consistently in onboarding V1 and V2
- reset onboarding scan state so users can retry

## Root cause

Device discovery returns some hardware failures as `success: false` responses instead of rejected promises. Onboarding converted those payloads locally but did not handle `BluetoothUnavailableWhileUsbConnectedError` on desktop, so the error was effectively ignored and scanning could continue. Rejected scan promises were also not caught by the fire-and-forget polling task.

## User impact

When Bluetooth is unavailable because USB is connected, onboarding now shows the existing localized troubleshooting message, stops the current scan, and returns to a retryable state. Other scan failures also use a single delivery path without changing legacy Scanner callers that do not opt into `onError`.

## Validation

- `yarn jest packages/shared/src/utils/DeviceScannerUtils.test.ts packages/shared/src/errors/utils/deviceErrorUtils.bluetooth.test.ts --runInBand` — 11 tests passed
- type-aware Oxlint on the four changed files — 0 warnings, 0 errors
- `git diff --check` passed

### DIFF
--- a/packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx
+++ b/packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx
@@ -38,38 +38,22 @@ import {
   OpenBleSettingsDialog,
   RequireBlePermissionDialog,
 } from '@onekeyhq/kit/src/components/Hardware/HardwareDialog';
-import { HyperlinkText } from '@onekeyhq/kit/src/components/HyperlinkText';
 import { ListItem } from '@onekeyhq/kit/src/components/ListItem';
 import { MultipleClickStack } from '@onekeyhq/kit/src/components/MultipleClickStack';
 import type { ITutorialsListItem } from '@onekeyhq/kit/src/components/TutorialsList';
 import { TutorialsList } from '@onekeyhq/kit/src/components/TutorialsList';
 import useAppNavigation from '@onekeyhq/kit/src/hooks/useAppNavigation';
 import { useHelpLink } from '@onekeyhq/kit/src/hooks/useHelpLink';
+import { useOnboardingDeviceScanErrorHandler } from '@onekeyhq/kit/src/hooks/useOnboardingDeviceScanErrorHandler';
 import { usePromptWebDeviceAccess } from '@onekeyhq/kit/src/hooks/usePromptWebDeviceAccess';
 import { useRouteIsFocused as useIsFocused } from '@onekeyhq/kit/src/hooks/useRouteIsFocused';
 import { useUserWalletProfile } from '@onekeyhq/kit/src/hooks/useUserWalletProfile';
 import { useAccountSelectorActions } from '@onekeyhq/kit/src/states/jotai/contexts/accountSelector/actions';
 import type { IDBCreateHwWalletParamsBase } from '@onekeyhq/kit-bg/src/dbs/local/types';
 import { useSettingsPersistAtom } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
-import {
-  HARDWARE_BRIDGE_DOWNLOAD_URL,
-  ONEKEY_BUY_HARDWARE_URL,
-} from '@onekeyhq/shared/src/config/appConfig';
+import { ONEKEY_BUY_HARDWARE_URL } from '@onekeyhq/shared/src/config/appConfig';
 import { OneKeyLocalError } from '@onekeyhq/shared/src/errors';
-import {
-  BleLocationServiceError,
-  BluetoothUnavailableWhileUsbConnectedError,
-  BridgeTimeoutError,
-  BridgeTimeoutErrorForDesktop,
-  ConnectTimeoutError,
-  DeviceMethodCallTimeout,
-  InitIframeLoadFail,
-  InitIframeTimeout,
-  NeedBluetoothPermissions,
-  NeedBluetoothTurnedOn,
-  NeedOneKeyBridge,
-  OneKeyHardwareError,
-} from '@onekeyhq/shared/src/errors/errors/hardwareErrors';
+import { OneKeyHardwareError } from '@onekeyhq/shared/src/errors/errors/hardwareErrors';
 import errorToastUtils from '@onekeyhq/shared/src/errors/utils/errorToastUtils';
 import bleManagerInstance from '@onekeyhq/shared/src/hardware/bleManager';
 import { checkBLEPermissions } from '@onekeyhq/shared/src/hardware/blePermissions';
@@ -352,23 +336,6 @@ function ConnectByQrCodeComingSoon() {
   );
 }
 
-// eslint-disable-next-line @typescript-eslint/no-unused-vars
-function BridgeNotInstalledDialogContent(_props: { error: NeedOneKeyBridge }) {
-  return (
-    <Stack>
-      <HyperlinkText
-        size="$bodyLg"
-        mt="$1.5"
-        translationId={
-          platformEnv.isSupportWebUSB
-            ? ETranslations.device_communication_failed
-            : ETranslations.onboarding_install_onekey_bridge_help_text
-        }
-      />
-    </Stack>
-  );
-}
-
 enum EConnectionStatus {
   init = 'init',
   searching = 'searching',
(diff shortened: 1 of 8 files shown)


packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx
```diff
@@ -203,6 +203,79 @@ function useDeviceConnection({
     currentTabValueRef.current = tabValue;
   }, [tabValue, deviceScanner]);
 
+  const handleScanError = useCallback(
+    (error: Error) => {
+      isSearchingRef.current = false;
+      setConnectStatus(EConnectionStatus.init);
+      deviceScanner.stopScan();
+
```
sidmorizon: **Desktop Bluetooth onboarding tab stops searching for devices after any scan error and never retries**

Severity: `non-severe`

The device search is permanently halted when any scan failure arrives (`deviceScanner.stopScan()` at `packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:210`) on a screen that has no button or automatic trigger to search again, so users stay on an empty device list until they leave and re-enter the screen.

Impact: On desktop, a user who sees the "Bluetooth unavailable while USB is connected" message and then unplugs USB will never see their device appear; onboarding appears frozen on the Bluetooth tab.

<details>
<summary>Why the Bluetooth tab has no retry path while the USB tab does</summary>

Both desktop Bluetooth screens start scanning only from an effect keyed on focus + bluetooth status: `packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:936-943` and `packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:1084-1090`. Neither dependency changes when the new error handler stops the scan, so `scanDevice()` is never re-invoked.

Unlike `USBOrBLEConnectionIndicator`, which renders a "Start connection" button when `connectStatus === init` (`packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:828-846`), `BluetoothConnectionIndicator` renders `<BluetoothCard />` with no `onConnect`/`connectStatus` props (`packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:1013`), so no retry control is shown. The V1 `ConnectByBluetooth` list view likewise has no retry affordance.

Before this change, an unsuccessful response such as `BluetoothUnavailableWhileUsbConnectedError` was ignored on desktop and the poll loop continued, so the device eventually showed up once USB was unplugged. Now the first failure ends polling for good.

</details>

packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx
```diff
@@ -203,6 +203,79 @@ function useDeviceConnection({
...
+        });
+        return;
+      }
+
+      Toast.error({
+        title: error.message || 'DeviceScanError',
+      });
+    },
+    [deviceScanner, intl],
+  );
```
sidmorizon: **Identical ~70-line error-handling block copied into two onboarding screens**

Severity: `non-severe`

The new scan error handler is duplicated verbatim (`handleScanError` at `packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:443-514` and `packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:206-277`) instead of being shared, and each copy exceeds the size limit the repository rules allow.

Impact: Future fixes to onboarding scan error messages must be made twice and can silently drift apart between the two onboarding flows.

<details>
<summary>Rule references</summary>

`.cursor/rules/self-testing.mdc` section 1 requires: "No function > 40 lines without justification" and "No duplicated logic — shared code extracted". The two handlers are byte-identical apart from surrounding imports and are each ~70 lines. Extracting a shared hook (e.g. under `packages/kit/src/views/Onboarding.../hooks`) would satisfy both rules.

</details>

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx
```diff
@@ -1159,15 +1075,29 @@ function ConnectByBluetooth({
...
+          <Button
+            testID="onboarding-bluetooth-retry-btn"
+            mx="auto"
+            size="large"
+            variant="primary"
+            onPress={listingDevice}
+          >
+            {intl.formatMessage({ id: ETranslations.global_retry })}
+          </Button>
+        </YStack>
```
sidmorizon, line 1092: **Bluetooth onboarding screen shows a "Retry" button before any scan has been attempted**

Severity: `non-severe`

The Bluetooth onboarding screen shows a retry button whenever the scan state is idle (`connectStatus === EConnectionStatus.init` at `packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:1081`), which is also the state on first entry while the Bluetooth availability check is still running, so users are asked to retry something that never started.

Impact: On opening the Bluetooth tab users briefly (or, if the system Bluetooth state never resolves, permanently) see a "Retry" button instead of the device search UI.

<details>
<summary>Why the idle state is reached before the first scan</summary>

`bluetoothStatus` starts as `'checking'` (`packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:816-824` area) and none of the early-return branches match that value, so the main UI renders while `connectStatus` is still `EConnectionStatus.init`. Scanning only begins after the async `checkBluetoothStatus()` resolves to `'enabled'` and the effect at `packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:995-1001` calls `listingDevice()`. `checkBluetoothStatus` awaits Noble pre-initialization plus a settings read, so the idle render is visible; and when `available.state === 'unknown'` the function returns early leaving the status at `'checking'` indefinitely, keeping the retry button on screen.

The same pattern was introduced in Onboarding v2: `BluetoothCard` now receives `connectStatus` (`packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:929-932`) and renders its "Start connection" button for the initial idle state, whereas previously no `connectStatus` was passed and the button never appeared there.

</details>

originalix (approved):

packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx
wabicai: Fixed in 84c58e9dd2. `BluetoothUnavailableWhileUsbConnectedError` is now treated as transient: the troubleshooting toast is shown once while polling continues, so unplugging USB can recover without leaving the page. Terminal scan failures still stop polling, but both onboarding flows now expose a retry path. Regression coverage was added for the scan policy and handler behavior.

packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx
wabicai: Fixed in 84c58e9dd2. The duplicated handlers were extracted into `useOnboardingDeviceScanErrorHandler`, with the transient-error policy isolated in `onboardingDeviceScanErrorUtils`. V1 and V2 now share the same error classification, messaging, toast de-duplication, and stop behavior, with dedicated tests.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx
wabicai: Confirmed, good catch. `connectStatus === init` currently conflates the pre-scan state with a terminal scan failure, so it is not a valid retry predicate. I will gate the retry CTA with an explicit scan-error/attempted state that is set only by `stopScanAfterError` and cleared when a new scan starts or succeeds. That keeps the initial Bluetooth availability check UI unchanged while preserving recovery after a real terminal error. I am leaving this thread unresolved until the patch and regression test are pushed.

PR #12831, opened 2026-08-11
fix: scope weak-network throttling to OneKey traffic

originalix:
## Summary

- Scope the weak-network Dev Setting to OneKey's own traffic: the API hosts (prod and test, including the notification socket in both websocket and polling transports) and the asset CDNs.
- Leave everything else at full speed — DApp browser tabs, third-party RPC and price feeds, and the local development server.
- Share one host list between Desktop and Native so the two platforms cannot drift apart.

## Intent & Context

The setting exists to see how the client behaves when OneKey's own backend is slow. Throttling every request instead made the whole app unusable while enabled: DApp pages crawled, third-party RPC endpoints timed out, and in development even Metro bundles and HMR were delayed, which added startup cost unrelated to the scenario being tested.

The native interception capability is implemented in OneKeyHQ/app-modules#82, released as `@onekeyfe/react-native-network-throttle@3.0.85` (OneKeyHQ/app-modules#84). This PR owns the app-side policy and the dependency integration.

## Design Decisions

- One shared constant, `NETWORK_THROTTLE_ONEKEY_HOSTS`, is the single source of truth. Desktop turns it into CDP URL patterns; Native passes it to the native module as `throttleUrlHosts`.
- `*.onekeycn.com` style entries match sub-domains at any depth but not the bare apex, which is correct because every OneKey endpoint is a sub-domain.
- Desktop expresses the whole policy as per-URL CDP rules with **no catch-all rule**, and never enables session-level emulation, which cannot distinguish origins. Session emulation is only ever cleared, so no session keeps a stale global profile from an earlier build.
- Desktop attaches only to the app's own renderer. DApp browser tabs and the overlay run in their own partitions and are not touched at all.
- There is no fallback when the per-rule CDP command is unavailable: the standard command can only throttle a whole session, which would slow DApp traffic. A failure means no throttling, and the log says so.
- The dev-server carve-out is gone on both platforms — the dev server simply is not on the allowlist.

## Changes Detail

- Add the shared host list and build the Desktop CDP rules from it.
- Pass the same list to the native module; delete the Metro dev-server origin policy.
- Upgrade the mobile package, Yarn lockfile, and iOS Pod lockfile to `3.0.85`.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Desktop, iOS, and Android, developer mode only.
- **Risk Areas**: Host pattern correctness and the experimental CDP rule command.

## Test plan

- [x] 31 targeted Jest tests, including 10 URLs that must be throttled and 9 that must not (DApp, third-party RPC, dev server, and look-alike hosts such as `evil-onekeycn.com`).
- [x] Live Electron 39.8.9 verification with every host routed to a local server: OneKey API 812 ms, OneKey CDN 800 ms, third-party RPC 5 ms, look-alike host 1 ms.
- [x] `yarn agent:check --profile commit`.

## Notes for reviewers

- Native WebView traffic (Android WebView, iOS WKWebView) was never intercepted by this feature, so DApp browsing on native was already unaffected.
- Native WebSocket traffic is not intercepted either (SocketRocket on iOS, a private OkHttp client on Android), so realtime push behaviour under weak network is only faithful on Desktop.
- Requests routed through the direct-IP fallback are matched by host, so they are not throttled while that path is active.


apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -537,15 +588,19 @@ export async function applyDesktopNetworkThrottleToKnownSessions(
...
+  if (options?.remoteOnly) {
+    remoteOnlyNetworkThrottleSessions.add(contents.session);
+  }
   const config = getRuntimeNetworkThrottleConfig();
   applyDesktopNetworkThrottleToSession(
     contents.session,
     `webContents:${contents.id}:${contents.getType()}`,
-    config,
+    getSessionNetworkThrottleConfig(contents.session, config),
   );
```
sidmorizon: **Weak-network simulation stops working entirely in desktop development builds**

Severity: `non-severe`

Session-wide throttling is switched off for the whole development window (`getSessionNetworkThrottleConfig` at `apps/desktop/app/libs/networkThrottle.ts:600-604`) and replaced by a per-request mechanism that cannot run while the developer tools panel is open, so the weak-network setting has no effect at all in development.
Impact: Developers who turn on the Slow 4G dev setting in a dev build see full-speed network instead of a throttled one, silently invalidating weak-network testing.

<details>
<summary>Debugger attach fails while DevTools is open, and the session fallback is now disabled</summary>

In dev-server mode the main window auto-opens DevTools (`apps/desktop/app/app.ts:902-904`). Electron refuses `webContents.debugger.attach()` while DevTools is open for that target (the code itself acknowledges this in `scheduleDesktopNetworkThrottleDebuggerReapply`, which parks the retry until `devtools-closed`, `apps/desktop/app/libs/networkThrottle.ts:263-278`).

Before this PR, `applyDesktopNetworkThrottleToSession` still applied `enableNetworkEmulation` to `session.defaultSession`, so throttling worked even when the CDP path failed. Now `applyDesktopNetworkThrottleToWebContents(browserWindow.webContents, { remoteOnly: isDevServer })` (`apps/desktop/app/app.ts:786-788`) registers the default session in `remoteOnlyNetworkThrottleSessions`, and every session-level apply for it is forced to `DEFAULT_NETWORK_THROTTLE_CONFIG` (disabled), including the toggle path `setDesktopNetworkThrottleConfig` → `applyDesktopNetworkThrottleToKnownSessions` (`apps/desktop/app/libs/networkThrottle.ts:563-573`).

So when the user toggles the dev setting with DevTools open, `attach()` throws, the debugger rules are never installed, and the session emulation is deliberately disabled — no throttling anywhere. Even if the initial attach at window creation succeeds, opening DevTools detaches the debugger and the reapply is deferred until DevTools closes.

</details>

originalix:

@originalix

Reviewed commit `5e50cb096dd7`. This report always reflects the latest reviewed changes.

### Review summary

This PR redefines weak-network simulation from whole-session throttling to an allowlisted OneKey-origin policy shared by desktop and native. Desktop programs per-renderer CDP rules, while iOS and Android pass the same host list to an upgraded native module, intentionally leaving DApps, third-party services, and development servers untouched. The ownership boundary is clearer and policy tests improve readability, but desktop failure and connection lifecycle handling plus native binary compatibility can still make the saved setting diverge from actual network behavior.

**What needs attention:** Confirm that applying or toggling the setting changes the real network state across desktop renderers and older installed mobile binaries, rather than merely updating persisted configuration.

### Issues to address

- **P1 · Desktop config updates can succeed without any debugger rule being applied**
  - **Impact:** `setDesktopNetworkThrottleConfig()` requests failure propagation, but `applyDesktopNetworkThrottleToKnownSessions()` forwards `throwOnFailure` only to the legacy session-clear operation. The CDP debugger task now owns all real throttling and still swallows attach or command failures, allowing the setting to be persisted while traffic remains unthrottled and preventing the existing rollback path from running.
  - **Suggested action:** Forward `options?.throwOnFailure` to every debugger apply task and add a setter-level test proving that a rejected attach or `Network.emulateNetworkConditionsByRule` command rejects the update and preserves the previous configuration.
- **P2 · Bundle updates lose native host scoping on older binaries**
  - **Impact:** On iOS and Android, a new JS bundle can run against a native module older than 3.0.84, which the code acknowledges does not expose `throttleUrlHosts`. Calling that module with `enabled: true` cannot install the new allowlist; depending on the older implementation, the unknown field is ignored and the previous global throttle remains active, or the update fails. Both native JS runtimes initialize independently against the native networking resource, so normal startup calls do not restore host-scoped behavior on an old binary.
  - **Suggested action:** Detect support through `getConfig()` before enabling host-scoped throttling and keep the feature disabled with an explicit unsupported-version error when `throttleUrlHosts` is absent.
- **P2 · Desktop toggles stop refreshing existing connections**
  - **Impact:** The setter still passes `closeConnections: true`, but connections are closed only when clearing session-wide emulation changes `appliedStateBySession`. Startup records every session as already disabled, so later enable and disable operations return no applied session and skip `closeAllConnections()`. Existing connections, including a notification WebSocket created before the rule change, can therefore retain their prior network conditions despite the toggle.
  - **Suggested action:** Decouple connection refresh from the legacy session-emulation state and close the applicable default-session connections after a debugger rule transition succeeds.

### Validation gaps

- The targeted Jest command could not start because this worktree has no installed `node_modules` state.
- A direct Node 24 `URLPattern` check confirmed the added wildcard patterns match the included OneKey examples and reject the tested look-alike and apex hosts.
- The source and advisory history for `@onekeyfe/react-native-network-throttle` 3.0.85 were not present locally, so its native implementation and supply-chain delta could not be inspected.
- Electron, iOS, and Android runtime behavior was not exercised.
- Existing GitHub review discussions could not be refreshed, so inline wording should be checked for duplication before posting.

apps/desktop/app/libs/networkThrottle.ts
originalix: Verified this against the actual current Desktop runtime (Electron 39.8.9), with `ONEKEY_DESKTOP_NETWORK_THROTTLE=slow4g` and the auto-opened DevTools left open.

The CDP target list simultaneously contained the DevTools page and `http://localhost:3001/`, and the main-process log emitted after DevTools opened:

`[desktop-network-throttle] debugger applied state=slow4g scope=remote-only webContents=webContents:1:window latencyMs=562.5 downloadBps=180000 uploadBps=84375`

A timed fetch from the live main renderer measured:

- localhost dev-server request: 3 ms
- remote `https://httpbin.org/get`: 48,347 ms

So on the Electron version used by this branch, DevTools does not detach or block this debugger client; the URL rule is active while DevTools is open. The loopback bypass and remote Slow 4G throttle are both observable. No code change is needed for this finding.

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -412,14 +438,39 @@ async function applyDesktopNetworkThrottleToWebContentsDebugger({
...
+        {
+          offline: false,
+          matchedNetworkConditions: [],
+        },
+      );
+      await targetDebugger.sendCommand(
+        'Network.emulateNetworkConditions',
+        getDebuggerNetworkConditions(config),
+      );
+    }
```
sidmorizon: **A failing new throttling command prevents the normal slow-network settings from being applied**

Severity: `non-severe`

A newly added rule-reset command is sent (`Network.emulateNetworkConditionsByRule` at `apps/desktop/app/libs/networkThrottle.ts:456-462`) before the actual throttle settings command, so if that command is rejected the throttle settings are never applied and the window stays unthrottled.
Impact: On builds where the new command is unavailable, the weak-network simulation silently stops working for every window and keeps retrying in the background.

<details>
<summary>Ordering makes a non-essential command fatal for the essential one</summary>

In the non-remote-only branch the code first issues `Network.emulateNetworkConditionsByRule` with an empty rule list purely to clear any previously installed rules, and only afterwards issues `Network.emulateNetworkConditions` with the real profile (`apps/desktop/app/libs/networkThrottle.ts:455-467`). Both are awaited inside a single try block; a rejection from the first (e.g. unsupported/experimental CDP method for the bundled Chromium, since `emulateNetworkConditionsByRule` is a recent addition) skips the second entirely, leaves `appliedStateByWebContentsDebugger` unset, logs a warning, and — via `scheduleDesktopNetworkThrottleDebuggerReapply` — retries up to `MAX_DEBUGGER_REAPPLY_ATTEMPTS` with the same failing order. Previously this path only sent `Network.emulateNetworkConditions` and therefore could not be broken by the rule API.

The unit test at `apps/desktop/app/libs/networkThrottle.test.ts:139-145` mocks `sendCommand` as always resolving, so it cannot catch this.

</details>

Suggested fix:

```ts
    } else {
      await targetDebugger.sendCommand(
        'Network.emulateNetworkConditions',
        getDebuggerNetworkConditions(config),
      );
      try {
        await targetDebugger.sendCommand(
          'Network.emulateNetworkConditionsByRule',
          {
            offline: false,
            matchedNetworkConditions: [],
          },
        );
      } catch (ruleError) {
        logger.warn(
          '[desktop-network-throttle] failed to clear per-rule network conditions',
          ruleError,
        );
      }
    }
```

originalix: Fixed in `a00295c75b`.

The standard `Network.emulateNetworkConditions` command now runs first and remains the required operation. Clearing `Network.emulateNetworkConditionsByRule` is best-effort in the non-remote-only path, so an older Chromium that rejects the rule API logs the cleanup failure but still keeps the requested global throttle profile and records the state as applied.

I also added a regression test that rejects only `Network.emulateNetworkConditionsByRule`, verifies the standard Slow 4G command was applied, and calls the apply path again to confirm it does not enter the retry loop. The three focused suites pass (13 tests), and `yarn agent:check --profile commit` passes.

originalix (commented):
@originalix

Reviewed commit `6df0933eb95b`.

- **P2 · Remote-only throttling disappears when debugger setup fails**
- **P2 · The native dependency also changes throttle semantics**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

sidmorizon:
@codex review

sidmorizon:
@codex security review

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -412,14 +438,46 @@ async function applyDesktopNetworkThrottleToWebContentsDebugger({
     }
 
     await targetDebugger.sendCommand('Network.enable');
-    await targetDebugger.sendCommand(
-      'Network.emulateNetworkConditions',
-      getDebuggerNetworkConditions(config),
-    );
+    if (config.enabled && remoteOnly) {
```
originalix: 🟡 **Remote-only path has no fallback when CDP rule setup fails — weak-network simulation silently turns off**

`applyDesktopNetworkThrottleToWebContents` synchronously disables session-wide emulation for remote-only sessions (via `getSessionNetworkThrottleConfig`), then relies on this async debugger path to install the URL rules. If `attach` or `Network.emulateNetworkConditionsByRule` rejects (experimental command removed in a future Electron, transient CDP error), the outer catch only logs a warning: session-level throttling is already off, the debugger level never applied, and the initial failure is not retried (reapply is only scheduled from the `detach` event). Result: the Dev Setting shows enabled while everything runs at full speed.

The non-remote-only branch already treats the standard command as authoritative with best-effort rule cleanup; the remote-only branch has no symmetric safety net.

**Suggested fix:** fall back to full session-wide throttling when the remote-only debugger apply fails (accepting that localhost is throttled too, which is better than no throttling at all):

```diff
   } catch (error) {
+    if (config.enabled && remoteOnly) {
+      // Fallback: URL-rule throttling failed; restore full session-wide
+      // throttling so the weak-network setting still takes effect.
+      remoteOnlyNetworkThrottleSessions.delete(contents.session);
+      applyDesktopNetworkThrottleToSession(
+        contents.session,
+        `${label}:fallback`,
+        config,
+      );
+    }
     if (!suppressFailureLog) {
```

(If you prefer to keep the session marked remote-only for later retries, call `enableNetworkEmulation` directly instead so `getSessionNetworkThrottleConfig` does not override the fallback.)

_— Auto-review by Claude_

apps/mobile/package.json
```diff
@@ -71,7 +71,7 @@
     "@onekeyfe/react-native-keychain-module": "3.0.78",
     "@onekeyfe/react-native-lite-card": "3.0.78",
     "@onekeyfe/react-native-native-logger": "3.0.78",
-    "@onekeyfe/react-native-network-throttle": "3.0.78",
+    "@onekeyfe/react-native-network-throttle": "3.0.82-alpha.0",
```
originalix: 🟡 **Version topology inversion: npm already has a stable `3.0.82` that does NOT contain `bypassUrlOrigins`, and it sorts higher than this pin**

Verified against the registry: `3.0.82` (published 2026-08-14 from `app-modules` main by the official `1keyfe` account) is byte-identical to current `origin/main` — no bypass feature — while the feature only exists in `3.0.82-alpha.0` (published 2026-08-11). Since `3.0.82-alpha.0 < 3.0.82` in semver, any routine "upgrade to stable" action (renovate, a teammate normalizing prerelease pins) silently drops the bypass capability: the JS side degrades gracefully, Metro simply becomes throttled again, and nothing crashes or logs.

Also, `app-modules` main already carries `"version": "3.0.82"`, while OneKeyHQ/app-modules#82 still says `3.0.82-alpha.0` (and its `stableVersion: 3.0.81` is stale) — the merge will conflict on that line and needs to resolve forward, not backward.

**Suggested action:**
- After rebasing OneKeyHQ/app-modules#82 onto main, bump it to `3.0.83-alpha.x` (or release `3.0.83` stable directly once merged).
- Keep this exact alpha pin for now (it is deterministic, and the published `3.0.82-alpha.0` tarball matches the PR #82 source file-by-file — verified), but track a follow-up to move to the stable `>=3.0.83` release.

_— Auto-review by Claude_

originalix: 🟡 **The 3.0.78 → 3.0.82-alpha.0 jump also changes throttle semantics — worth calling out in the PR description for QA**

Verified from the published tarballs: `3.0.78` has no `BandwidthLimiter` (latency-only), while the `3.0.81+` mainline adds 102 KiB/s bidirectional throughput limiting plus the Android request/response body wrappers and the iOS buffering/backpressure machinery. After this upgrade, enabling the native weak-network setting means "562.5 ms latency **+ 102 KiB/s both directions**" instead of latency only.

This is intentional mainline behavior shipped via OneKeyHQ/app-modules#78 (not something introduced by this PR), so pinning a 3.0.78-based fork just to avoid it would fragment the package line — not recommended. Suggest simply noting the semantics change in the PR description and having QA regress the weak-network scenario against the new behavior (latency + bandwidth cap).

_— Auto-review by Claude_

apps/desktop/app/libs/networkThrottle.ts
originalix: Fixed in `bd4e32df053a725acf221bda51c631a8e6956fe7`. The remote-only debugger path now falls back to full session-wide throttling on any failure (the session is unmarked as remote-only, so subsequent debugger retries apply the standard commands). Added a regression test where `Network.emulateNetworkConditionsByRule` rejects and session `enableNetworkEmulation` is asserted with the Slow 4G profile.

apps/mobile/package.json
originalix: Addressed: OneKeyHQ/app-modules#82 has been merged with `main` (which is at 3.0.82) and bumped to `3.0.83-alpha.0` in commit `bd2cc250`, so the next release containing the bypass feature sorts above the featureless stable 3.0.82. This PR keeps the deterministic `3.0.82-alpha.0` pin; moving to the stable `>= 3.0.83` release is tracked in the PR description's follow-ups section.

originalix: Done — the PR description now has a "Review follow-ups" section documenting the latency + 102 KiB/s bidirectional semantics for QA regression.

originalix (commented):
@originalix

Reviewed commit `8fbf3c48f251`.

- **P2 · Desktop remote-only bypass ignores non-loopback dev servers**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -436,6 +494,19 @@ async function applyDesktopNetworkThrottleToWebContentsDebugger({
...
+      // the weak-network setting still takes effect; throttling localhost is
+      // better than silently not throttling at all. Later debugger retries
+      // see the session as non-remote-only and apply the standard commands.
+      remoteOnlyNetworkThrottleSessions.delete(contents.session);
+      applyDesktopNetworkThrottleToSession(
+        contents.session,
+        `${targetLabel}:remote-only-fallback`,
+        config,
+      );
+    }
```
sidmorizon: **Cleanup path can raise a second error when a window closes mid-setup, aborting the network setting update**

Severity: `non-severe`

The recovery step reads the closing window's browsing session (`contents.session` at `apps/desktop/app/libs/networkThrottle.ts:503-505`) after the window may already have been torn down, so a window closing at the wrong moment turns a handled failure into an unhandled error that makes the weak-network setting update fail and roll back.
Impact: Toggling the weak-network developer setting can fail with an error when a window or webview closes at the same time.

<details>
<summary>Destroyed WebContents property access inside the catch block</summary>

`applyDesktopNetworkThrottleToWebContentsDebugger` only checks `contents.isDestroyed()` at entry (`apps/desktop/app/libs/networkThrottle.ts:399-401`); several `await targetDebugger.sendCommand(...)` calls follow, during which the webContents can be destroyed (which is also why the debugger commands reject). The new fallback block then accesses `contents.session` twice. Electron property access on a destroyed WebContents throws `Object has been destroyed`, so the catch block itself throws.

Callers do not guard against that: `applyDesktopNetworkThrottleToWebContents` invokes it with `void ...` (`apps/desktop/app/libs/networkThrottle.ts:624-628`) → unhandled rejection; `applyDesktopNetworkThrottleToKnownSessions` awaits it inside `Promise.all` (`apps/desktop/app/libs/networkThrottle.ts:595-605`), so the rejection propagates into `setDesktopNetworkThrottleConfig`, which treats it as a commit failure, rolls back and rethrows to the renderer (`apps/desktop/app/libs/networkThrottle.ts:656-676`). Capturing the session in a local variable before the `try` (or wrapping the fallback in its own try/catch) avoids this.

</details>

apps/mobile/ios/Podfile.lock
```diff
@@ -3914,7 +3914,7 @@ PODS:
     - ReactCommon/turbomodule/core
     - SocketRocket
     - Yoga
-  - ReactNativeNetworkThrottle (3.0.78):
+  - ReactNativeNetworkThrottle (3.0.82-alpha.0):
```
sidmorizon: **iOS dependency lock file still points at the previous prerelease of the network throttle library**

Severity: `non-severe`

The iOS dependency lock file records version 3.0.82-alpha.0 for the network throttle library (`ReactNativeNetworkThrottle (3.0.82-alpha.0)` at `apps/mobile/ios/Podfile.lock:3917`) while the JavaScript dependency files pin 3.0.83-alpha.0, so iOS builds resolve a different version than the rest of the project expects.
Impact: iOS builds and CI checks that verify the lock file is up to date can fail or build against native code that lacks the new behavior.

<details>
<summary>Version pin mismatch between package.json/yarn.lock and Podfile.lock</summary>

`apps/mobile/package.json:74` pins `@onekeyfe/react-native-network-throttle` to `3.0.83-alpha.0`, and `yarn.lock` resolves `@onekeyfe/react-native-network-throttle@npm:3.0.83-alpha.0`. The Pod version is derived from the npm package's podspec, so the lock file should read `3.0.83-alpha.0` (all sibling `@onekeyfe` pods match their npm versions, e.g. `ReactNativeNativeLogger (3.0.78)`). It appears the Podfile.lock was regenerated for the earlier `3.0.82-alpha.0` pin (matching the PR description) and not refreshed after the bump to `3.0.83-alpha.0`. Re-running `pod install` after the dependency bump should update both the version line and the `SPEC CHECKSUMS` entry at `apps/mobile/ios/Podfile.lock:5594`.

</details>

apps/desktop/app/libs/networkThrottle.ts
originalix: Correction — my earlier "fixed" reply on this thread was wrong, and the fix has been reworked in `643b934dd8f56dd7acfdeac5502cbbf415a793ff`.

The first attempt fell back to `session.enableNetworkEmulation`, which is **inert on the pinned Electron 39.8.9**. An Electron probe against the real binary showed `session.enableNetworkEmulation({offline:true})` still returning `status=200` in 1ms, while the equivalent CDP command behaved as expected. Worse, the remote-only path sends `Network.emulateNetworkConditions` with the *disabled* profile before installing the rules, so when the rule command failed the debugger was left explicitly un-throttled and nothing repaired it — while the log claimed `applied state=slow4g`. Measured end state: localhost 4ms, remote 270ms, i.e. no throttling at all behind a success log.

The fallback now re-sends the real profile through the same debugger (the layer that actually throttles) and records the applied state so later applies do not retry the failing rule command. Probe after the fix: localhost 803ms / remote 577ms — full throttling, as intended for a degraded fallback.

Also fixed in the same commit: `*://[::1]:*/*` is not constructible as a URLPattern (the `:` parses as a named-group prefix), so CDP silently dropped that rule and IPv6 dev-server traffic stayed throttled at 575ms. It is now escaped as `*://\[\:\:1\]:*/*` (probe: 1ms), and the policy test asserts every bypass pattern is constructible so a silently-dead rule fails tests instead of shipping.

_— Auto-review by Claude_

originalix (commented):
@originalix

Reviewed commit `643b934dd8f5`.

- **P2 · Desktop remote-only bypass ignores non-loopback dev servers**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -364,6 +389,51 @@ function applyDesktopNetworkThrottleToSession(
...
+      'Network.emulateNetworkConditions',
+      getDebuggerNetworkConditions(config),
+    );
+  } catch (fallbackError) {
+    logger.warn(
+      `[desktop-network-throttle] failed remote-only fallback webContents=${targetLabel}`,
+      fallbackError,
+    );
+    return false;
+  }
```
sidmorizon: **Development server files start being slowed down after any temporary throttling error**

Severity: `non-severe`

The development-server exemption is thrown away (`remoteOnlyNetworkThrottleSessions.delete(contents.session)` at `apps/desktop/app/libs/networkThrottle.ts:414`) on any failure — including ones unrelated to the per-URL rules and even when the replacement attempt itself also fails — so from then on local development files are slowed down like remote ones.

Impact: A single transient hiccup makes the weak-network simulation permanently throttle local bundle/asset loading in development, causing very slow app startup and reloads until the app is restarted.

<details>
<summary>Why an unrelated failure permanently drops the loopback bypass</summary>

`applyDesktopNetworkThrottleToWebContentsDebugger` wraps debugger attach, `Network.enable`, `Network.emulateNetworkConditions` and `Network.emulateNetworkConditionsByRule` in one `try` (`apps/desktop/app/libs/networkThrottle.ts:459-540`). Any of these throwing (e.g. `targetDebugger.attach('1.3')` failing with "Another debugger is already attached" while DevTools is open — DevTools is auto-opened in dev at `apps/desktop/app/app.ts:902-904`, and the reapply path re-attaches at `apps/desktop/app/libs/networkThrottle.ts:299-320`) enters the catch, which calls the fallback whenever `config.enabled && remoteOnly`.

The fallback first deletes the session from `remoteOnlyNetworkThrottleSessions` and only then tries `Network.emulateNetworkConditions`. If that command also fails (very likely when the failure was the attach itself) it returns `false`, the error is rethrown/logged, and the retry scheduled by `scheduleDesktopNetworkThrottleDebuggerReapply` now computes `remoteOnly === false` (`networkThrottle.ts:448`), applying full throttling to the dev-server session — including `localhost` bundle/chunk requests this PR aims to exempt. The deletion is session-wide, so every other webContents sharing `defaultSession` also loses the bypass.

The marking should only be dropped when the rule command specifically failed AND the standard-command fallback succeeded.
</details>

originalix (commented):
@originalix

Reviewed commit `abcff250bc92`.

- **P2 · Desktop remote-only bypass ignores non-loopback dev servers**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

originalix (commented):
@originalix

Reviewed commit `f2c8d41eb436`.

- **P2 · First desktop throttle toggle still closes untouched DApp sessions**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -510,10 +528,12 @@ export async function applyDesktopNetworkThrottleToKnownSessions(
...
   for (const entry of entries) {
+    // Sessions are never throttled as a whole; this only clears emulation left
+    // behind by an earlier build so no session keeps a stale global profile.
     const appliedSession = applyDesktopNetworkThrottleToSession(
       entry.targetSession,
       entry.label,
-      config,
+      DEFAULT_NETWORK_THROTTLE_CONFIG,
       options?.throwOnFailure,
     );
```
sidmorizon, line 542: **Turning on the weak-network developer setting reports success even when nothing is actually slowed down**

Severity: `non-severe`

The commit step for the weak-network setting now only verifies that plain (session-wide) emulation was cleared (`DEFAULT_NETWORK_THROTTLE_CONFIG` at `apps/desktop/app/libs/networkThrottle.ts:536`) instead of verifying the step that actually slows OneKey traffic, so the setting can be saved and reported as enabled while no traffic is slowed at all.
Impact: A developer toggling weak-network mode sees a success message and an "on" switch while the app keeps running at full speed, making the simulated slow-network test silently invalid.

<details>
<summary>Why the failure is no longer detected: throwOnFailure only guards the session step</summary>

Before this PR, throttling was implemented by `targetSession.enableNetworkEmulation(...)` inside `applyDesktopNetworkThrottleToSession`, and `setDesktopNetworkThrottleConfig` passed `throwOnFailure: true` (`apps/desktop/app/libs/networkThrottle.ts:604-608`) so a failure propagated, triggered the rollback branch, and surfaced the `Failed to update network throttle` toast in `packages/kit/src/views/Setting/pages/Tab/DevSettingsSection/index.tsx:530-556`.

Now the session call is always `DEFAULT_NETWORK_THROTTLE_CONFIG` (i.e. `disableNetworkEmulation()`), which is a no-op that practically never fails, while the real mechanism — the per-URL rules sent through the debugger in `applyDesktopNetworkThrottleToWebContentsDebugger` (`apps/desktop/app/libs/networkThrottle.ts:426-433`) — is invoked from `applyDesktopNetworkThrottleToKnownSessions` (`apps/desktop/app/libs/networkThrottle.ts:544-553`) without `throwOnFailure`, so any error (e.g. the experimental `Network.emulateNetworkConditionsByRule` being unavailable, or the debugger failing to attach because DevTools is open) is only logged. `store.setNetworkThrottle(normalizedConfig)` then persists `enabled: true` and the UI shows `Slow 4G enabled`.

Forwarding `options?.throwOnFailure` to the debugger tasks (at least for the renderer that owns the app UI) would restore the rollback/error signal.

</details>

originalix (commented):
@originalix

Reviewed commit `6c7978bc19fd`.

- **P1 · Desktop config updates can succeed without any debugger rule being applied**
- **P1 · The shared throttle allowlist drops multiple production OneKey hosts**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

zhaono1 (dismissed):

wabicai (dismissed):

apps/desktop/app/libs/networkThrottle.ts
```diff
@@ -510,10 +532,12 @@ export async function applyDesktopNetworkThrottleToKnownSessions(
 
   const closeConnectionTasks: Array<Promise<void>> = [];
   for (const entry of entries) {
+    // Sessions are never throttled as a whole; this only clears emulation left
+    // behind by an earlier build so no session keeps a stale global profile.
     const appliedSession = applyDesktopNetworkThrottleToSession(
       entry.targetSession,
       entry.label,
-      config,
+      DEFAULT_NETWORK_THROTTLE_CONFIG,
```
sidmorizon, line 540: **Turning weak-network simulation on or off no longer drops existing connections**

Severity: `non-severe`

Existing network connections are no longer dropped when the weak-network setting is toggled, because the session step is now always given the "off" settings (`DEFAULT_NETWORK_THROTTLE_CONFIG` at `apps/desktop/app/libs/networkThrottle.ts:540`) and therefore reports "nothing changed" on every call after the first, so the connection-closing step never runs.
Impact: Already-open long-lived connections (such as the notification socket) keep running as before after the setting is switched, so the simulated slow network is applied inconsistently until the app is restarted.

<details>
<summary>Why the close-connections step became unreachable</summary>

`applyDesktopNetworkThrottleToSession` returns `undefined` when the previously applied state key equals the new one (`apps/desktop/app/libs/networkThrottle.ts:332-336`). Because the session call now always passes `DEFAULT_NETWORK_THROTTLE_CONFIG`, the state key is permanently `'disabled'`. The first invocation at startup (`apps/desktop/app/app.ts:1261`, called without `closeConnections`) stores `'disabled'` in `appliedStateBySession` for the default session and both webview partitions. Every later invocation from `setDesktopNetworkThrottleConfig` uses `closeConnections: true`, but `appliedSession` is now always `undefined`, so `closeConnectionTasks` stays empty (`apps/desktop/app/libs/networkThrottle.ts:543-545`) and `closeSessionConnections` is effectively dead code.

</details>

originalix (commented):
@originalix

Reviewed commit `5e50cb096dd7`.

- **P1 · Desktop config updates can succeed without any debugger rule being applied**
- **P2 · Bundle updates lose native host scoping on older binaries**
- **P2 · Desktop toggles stop refreshing existing connections**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12831#issuecomment-5251296999)

huhuanming (approved):

PR #12834, opened 2026-08-11
feat: track universal search source

yikZero:
## Summary
- Add a stable `source` property to the existing `universalSearchQuery` and `universalSearchClick` events.
- Attribute desktop header entries and keyboard shortcuts to the active business tab, and distinguish the native Discovery segments as `market`, `earn`, and `browser`.
- Preserve existing event names, trigger frequency, click result type normalization, and legacy navigation compatibility through the `unknown` fallback.

## Intent & Context
Growth needs to identify which product page a user was on when using Universal Search. This is intended for the release 6.5.0 hot update and extends the existing events instead of introducing a new search-open event.

## Root Cause
The existing query and click events recorded search content and result metadata but did not retain the route context where the search modal was opened. Universal Search has multiple platform-specific entry points, so the originating page was lost once navigation entered the modal.

## Design Decisions
- Resolve and freeze `source` when opening the search modal, then reuse it for all query and result-click events in that modal session.
- Use a typed shared enum and one route resolver for stable analytics values. Missing legacy route state falls back to `unknown`.
- Map desktop Discovery and MultiTabBrowser to `browser`; on native, map the active Discovery segment to `market`, `earn`, or `browser`.
- Read the current desktop tab at shortcut invocation time.
- Keep UI, search behavior, event names, and event frequency unchanged.

## Changes Detail
- Extend shared Universal Search route and logger contracts with the required `source` property.
- Pass the source through desktop, mobile, web, and extension search entry points.
- Propagate the same source to query events and every result item click event.
- Add resolver, shortcut, logger-scene, and result-click regression coverage.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**: Route-to-source attribution and platform-specific search entry coverage; user-facing search behavior is unchanged.

## Test plan
- [x] Focused Jest: 4 suites, 22 tests
- [ ] Manually verify every desktop business tab and the Universal Search shortcut
- [ ] Manually verify iOS/Android Wallet and Discovery Market, Earn, and Browser entries
- [ ] Confirm Query and Click payloads contain the expected `source` in PostHog

### DIFF
--- a/packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx
+++ b/packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx
@@ -108,7 +108,11 @@ function MobileBrowser() {
         {/* custom header */}
         <YStack my="$2">
           <XStack mx="$5">
-            <LegacyUniversalSearchInput size="medium" initialTab="market" />
+            <LegacyUniversalSearchInput
+              size="medium"
+              initialTab="market"
+              tabRoute={ETabRoutes.Discovery}
+            />
           </XStack>
           <TabPageHeader
             sceneName={EAccountSelectorSceneName.home}
--- a/packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx
+++ b/packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx
@@ -24,22 +24,15 @@ import { EAccountSelectorSceneName } from '@onekeyhq/shared/types';
 import { EarnHomeWithProvider } from '../../../Earn/EarnHome';
 import { MarketHomeWithProvider } from '../../../Market/MarketHomeV2/MarketHomeV2';
 
+import {
+  getExploreTabName,
+  getExploreUniversalSearchTabRoute,
+} from './exploreTabUtils';
 import { withBrowserProvider } from './WithBrowserProvider';
 
 import type { RouteProp } from '@react-navigation/core';
 
-type IExploreTabName = 'market' | 'earn' | 'browser';
 type IExploreTabSwitchType = 'default' | 'tap' | 'swipe';
-
-function getExploreTabName(tab: ETranslations): IExploreTabName {
-  if (tab === ETranslations.global_market) {
-    return 'market';
-  }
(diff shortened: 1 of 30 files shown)


packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx
```diff
@@ -108,7 +108,11 @@ function MobileBrowser() {
         {/* custom header */}
         <YStack my="$2">
           <XStack mx="$5">
-            <LegacyUniversalSearchInput size="medium" initialTab="market" />
+            <LegacyUniversalSearchInput
+              size="medium"
+              initialTab="market"
+              tabRoute={ETabRoutes.Discovery}
+            />
```
sidmorizon, line 110: **Searches started from the Market or Earn sections of the extension's explore page are all recorded as coming from the browser**

Severity: `non-severe`

The search entry on the extension explore page always reports the browser section (`tabRoute={ETabRoutes.Discovery}` at `packages/kit/src/views/Discovery/pages/Browser/Browser.ext.tsx:114`) even when the user is viewing the Market or Earn section, so those searches are attributed to the wrong page.

Impact: Analytics for extension users cannot distinguish searches started from Market or Earn; they all appear as browser searches.

<details>
<summary>Mismatch with the native Discovery segment mapping</summary>

On native, `packages/kit/src/views/Discovery/pages/Browser/Browser.native.tsx:104-110` builds `UNIVERSAL_SEARCH_TAB_ROUTE_MAP` and derives `universalSearchTabRoute` from `getExploreTabName(selectedHeaderTab)` so the active segment maps to `market` / `earn` / `browser`. The ext page holds the same `selectedHeaderTab` state (`Browser.ext.tsx:51-53`, with `getExploreTabName` already defined at `Browser.ext.tsx:34-42`) but passes a constant `ETabRoutes.Discovery`, which `getUniversalSearchSource` resolves to `EUniversalSearchSource.Browser` (`packages/kit/src/views/UniversalSearch/universalSearchSource.ts:21-23`).

</details>

originalix (approved):

PR #12839, opened 2026-08-12
sync release/v6.5.0 to the hotfix/v6.5.2

originalix:


originalix:

@originalix

Reviewed commit `cfe11412e7de`. This report always reflects the latest reviewed changes.

### Review summary

This PR backports a broad set of perps, staking, swap, and market changes into the hotfix line. It rewires Hyperliquid multi-DEX asset identification and cache recovery, adds provider-specific withdraw-path and approval handling for new Earn providers, and threads swap trade-source metadata through build-transaction requests while tightening several cold-start and subscription-recovery paths. The new helpers are mostly scoped to concrete runtime problems, but a few UI-state and navigation transitions now rely on stale state or incomplete native handoff behavior, which creates user-visible regressions despite the added test coverage.

**What needs attention:** Re-check fresh quote gating in staking withdraws and native perps entry navigation before merging this sync.

### Issues to address

- **P1 · Universal withdraw can submit stale transaction confirmation data**
  - **Impact:** For providers that do not require an explicit withdraw-path handshake, the component now keeps rendering the previous `transactionConfirmation` while a new quote request is in flight. Because submit gating only waits on `checkAmountLoading`, users can confirm a changed amount while price-impact checks, receive data, and `effectiveApy` still belong to the previous quote.
  - **Suggested action:** Require a current transaction-confirmation response for any flow that consumes quote-derived fields, or disable confirm/rendering until the in-flight confirmation request resolves.
- **P1 · Universal Search no longer opens the mobile perp market screen**
  - **Impact:** The native-only root-navigation hop was removed from the Universal Search perp entry point. On native, `switchTab(ETabRoutes.Perp)` changes tabs but does not push `MobilePerpMarket` from the search modal context, so users lose the expected market-detail jump after tapping a perp result.
  - **Suggested action:** Restore the native root navigation into `EModalPerpRoutes.MobilePerpMarket` after the tab switch, matching the existing Home perps entry behavior.
- **P2 · Manage-page stale-data fingerprint ignores indexed-account identity**
  - **Impact:** `useManagePage()` relies on `requestKey` to mark stale data during the first render after a selection change, but the fingerprint omits `indexedAccountId` even though the fetch depends on it. Indexed-account switches can therefore look current and allow downstream effects to act on previous-account data until the async reload finishes.
  - **Suggested action:** Include `indexedAccountId` in `requestKey` and keep any render-time stale guards aligned with the full fetch identity.

### Validation gaps

- I did not run targeted native UI validation for Universal Search -> Perp market navigation.
- I did not run end-to-end staking withdraw flows to confirm that quote freshness is enforced after amount edits.
- The added tests cover many helper utilities, but I did not see coverage for the stale-quote submit path or the `indexedAccountId` omission in the manage-page stale-data guard.

wabicai (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

PR #12840, opened 2026-08-12
Feat/hw 12641 picks on hotfix652

ByteZhang1024:


originalix:

@ByteZhang1024

Reviewed commit `bc446f8566ed`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the Trezor connection boundary between the main UI runtime and the background runtime: onboarding selects a channel, while background services resolve the effective USB or BLE connectId, recover stale bindings, persist THP credentials, and release failed BLE links through the shared Electron native bridge. It also improves uninitialized-device handling and makes connectId-based seed verification an explicit vendor capability. The privileged hardware workflow remains background-owned and identifier tracing is reduced, but target selection and lock-state classification still have two transport correctness gaps.

**What needs attention:** Verify target-specific transport selection with multiple Trezors and locked-device recovery on both Windows and macOS.

### Issues to address

- **P2 · USB presence is not scoped to the target Trezor**
  - **Impact:** In the bg runtime, detectTrezorUSBDeviceAvailability() treats any authorized device matching the Trezor vendor/product filters as proof that the selected device is available over USB. If Trezor A is BLE-only while Trezor B is connected over USB, the resolver chooses A's unavailable USB connectId, delaying the hardware operation until failure and fallback. The WebUSB list is a shared native-backed resource, while this selection decision is made from a per-bg-runtime JS snapshot.
  - **Suggested action:** Pass the target USB connectId into the probe, require a matching usable serialNumber, and add a two-device regression test.
- **P2 · Locked macOS devices are treated as stale BLE bindings**
  - **Impact:** In the bg runtime, the recovery helper recognizes a locked device only through the Windows-specific "unreachable while discovering services" text. The shared Electron BLE layer can report macOS lock failures as a bare "connection failed", which this branch classifies as transport-down and may route into discovery or THP pairing even though the device only needs to be unlocked.
  - **Suggested action:** Propagate a structured BLE failure classification from the native connector and permit rebinding only for an explicit stale-address result.

### Validation gaps

- The targeted Jest suites could not start because the workspace lacks Yarn's installed dependency-state file.
- No live two-Trezor USB/BLE scenario or locked-device macOS scenario was exercised.
- Existing GitHub discussion coverage could not be refreshed because the API connection was unavailable, so the inline wording requires a duplicate check before posting.

packages/kit/src/views/Onboardingv2/pages/ConnectionFlowTrezor.tsx
```diff
@@ -310,6 +310,22 @@ export default function TrezorConnectionFlow() {
...
+        transportLabel === 'BLE'
+          ? EConnectDeviceChannel.bluetooth
+          : EConnectDeviceChannel.usbOrBle;
+      const correctedTransportType =
+        await getForceTransportType(selectedChannel);
+      if (correctedTransportType) {
+        await backgroundApiProxy.serviceHardware.setForceTransportType({
+          forceTransportType: correctedTransportType,
+        });
+      }
```
sidmorizon: **Picking a Trezor from the list can silently do nothing if a settings lookup fails**

Severity: `non-severe`

The picked device's transport is looked up (`getForceTransportType(selectedChannel)` at `packages/kit/src/views/Onboardingv2/pages/ConnectionFlowTrezor.tsx:321-322`) without any error handling before the app moves to the setup screen, so a failed lookup aborts the tap and the user stays on the list with the scan already stopped.
Impact: Tapping a Trezor device occasionally appears to do nothing and onboarding cannot continue until the user restarts the flow.

<details>
<summary>Unguarded IPC call in the device-select handler</summary>

`getForceTransportType` performs background service calls over IPC (`backgroundApiProxy.serviceDevSetting.getDevSetting()` / `serviceSetting.getHardwareTransportType()` in `packages/kit/src/views/Onboardingv2/utils.tsx:45-73`) and can reject. In `handleDeviceSelect` the call is awaited before `navigation.push(...)`, and `ensureStopScan()` has already stopped scanning, so a rejection leaves the page in a dead state with no toast and an unhandled promise rejection.

The same resolver is treated as fallible elsewhere: `packages/kit/src/views/Onboardingv2/pages/FinalizeWalletSetup.tsx:566-578` wraps it in try/catch and explicitly comments that it "can fail" and must not block the flow. Wrapping the new call the same way (falling back to leaving the current force transport type) restores that behavior.

</details>

Suggested fix:

```ts
      try {
        const correctedTransportType =
          await getForceTransportType(selectedChannel);
        if (correctedTransportType) {
          await backgroundApiProxy.serviceHardware.setForceTransportType({
            forceTransportType: correctedTransportType,
          });
        }
      } catch {
        // Per-session transport pinning is best-effort; never block selection.
      }
```

originalix (commented):
@ByteZhang1024

Reviewed commit `dd7b585f2dda`.

- **P2 · USB presence is not scoped to the target Trezor**
- **P2 · Locked macOS devices are treated as stale BLE bindings**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12840#issuecomment-5262097723)

packages/kit-bg/src/vaults/base/trezorTransportUtils.ts
```diff
@@ -94,7 +127,18 @@ export async function callTrezorWithBleFallback<T>(
...
+  if (options?.resolvePrimaryConnectId) {
+    try {
+      const resolved = await options.resolvePrimaryConnectId(dbDevice);
+      if (resolved) {
+        primaryConnectId = resolved;
+      }
+    } catch {
+      // noop
+    }
+  }
```
sidmorizon, line 141: **Plugged-in Trezor is contacted over Bluetooth instead of its cable, and the cable is never retried**

Severity: `non-severe`

The first attempt of every Trezor call now uses the transport chosen by the shared resolver (`resolvePrimaryConnectId` at `packages/kit-bg/src/vaults/base/trezorTransportUtils.ts:132-141`) whose availability probe only recognises OneKey cables, so a Trezor sitting on USB is judged "no cable" and the call is sent over Bluetooth, and because that same address is then treated as the primary the saved-Bluetooth retry step is skipped.
Impact: Users with a cabled Trezor get slower/failing Bluetooth calls and an unnecessary "re-pair Bluetooth" prompt instead of the working cable connection.

<details>
<summary>How the OneKey-only USB probe steers a cabled Trezor onto BLE</summary>

`buildTrezorBleFallbackOptions` (`packages/kit-bg/src/vaults/base/trezorTransportUtils.ts:92-98`) resolves the primary connectId via `serviceHardware.getCompatibleConnectId` with `USER_INTERACTION`. For a third-party (Trezor) device that path goes through `connectionManager.resolveTransportType` → `determineOptimalTransportType` (`packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:305-338`), which decides USB presence with `detectUSBDeviceAvailability` → `detectWebUSBAvailability`, filtering only on `ONEKEY_WEBUSB_FILTER` (`HardwareConnectionManager.ts:148-159`). A Trezor's VID/PID is not in that list (see `packages/kit/src/hooks/webDeviceFilters.ts:32` where the Trezor filters are added separately), so `usbAvailable` is false and, with Bluetooth available, the target becomes `DesktopWebBle`; `getCompatibleConnectId` then returns the persisted `bleConnectId` (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:4210-4215`).

This PR adds a Trezor-scoped probe (`detectTrezorUSBDeviceAvailability`, `HardwareConnectionManager.ts:204-220`) but only consults it in `resolveTrezorPreferredBleConnectId` for non-BLE targets; the transport-type decision itself still uses the OneKey-only probe.

Two consequences inside `callTrezorWithBleFallback`:
1. `primaryConnectId` becomes the BLE address, so the step-1 bound-BLE fallback (`trezorTransportUtils.ts:113-118`, guarded by `bleConnectId !== primaryConnectId`) is a no-op and the USB handle is never tried at all.
2. The previous behaviour also preferred `dbDevice.usbConnectId` over `dbDevice.connectId`; the resolver passes only `dbDevice.connectId` to `getCompatibleConnectId`, so a record whose main `connectId` differs from `usbConnectId` now falls back to the main `connectId`.

</details>

ByteZhang1024: 这个是 Onkey、Trezor 的遗留问题，未来会修复的。

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit-bg/src/services/ServiceThirdPartyHardware/index.ts
```diff
@@ -281,20 +281,25 @@ class ServiceThirdPartyHardware extends ServiceBase {
...
-          return null;
-        }
         defaultLogger.hardware.sdkLog.log(
           `[TrezorBLEBind] candidate probe failed bleConnectId=${bleConnectId}`,
         );
         throw convertThirdPartyDeviceError(result.payload, {
           vendor: EHardwareVendor.trezor,
         });
       }
+      // No device_id = "could not verify", never "different device".
```
sidmorizon, line 290: # [🟡 P2] `bindTrezorBleConnectId` 的两处注释仍在描述本 PR 已移除的 probe-cancel 行为

## 问题

本 PR 在 `TrezorAdapter` 中移除了 binding probe 期间静默 cancel THP pairing 请求的逻辑（因为过期 THP credential 会让用户自己的设备也发起 pairing，静默 cancel 会导致 rebind 死锁），并把 user-abort 从 `resolves.toBeNull()` 改为 throw。`adapters/types.ts` 的接口文档已同步更新，但本文件里有两处注释仍在描述旧行为：

1. `beginBindingProbe` 调用点上方的注释块（本函数 274-279 行附近）：

```ts
// A picked candidate that ISN'T this device asks to pair (its static key
// doesn't match the shared credential). Suppress the THP pairing dialog and
// cancel silently during the probe — treat the pairing request as "not this
// one". Handled inside the adapter so it overrides its own pairing UI ...
```

2. 函数级 doc comment（约 248-250 行）：`"A different device (device_id mismatch, or it asks to pair) → return null so the UI says 'not this one, pick another'"` —— 现在 "asks to pair" 不再是 "not this one" 信号，user-abort pairing 也不再返回 `null` 而是 throw。

## 影响

这段注释描述的正是本 PR 修复的死锁根因。后续维护者读到 "Suppress the THP pairing dialog and cancel silently" 会误以为 adapter 仍有该行为（或应当恢复它），而设备身份判定语义（`null` = device_id mismatch，throw = 无法验证/中断）恰恰是这个绑定流程的安全边界，文档与实现相反容易在后续改动中重新引入被修掉的 bug。

## 建议

把这两处注释改为与新语义一致，例如 `beginBindingProbe` 上方改为说明其现在仅是 probe 生命周期标记（与 `TrezorAdapter._bindingProbeConnectId` 的 "bookkeeping only" 注释对齐），身份判定完全由 post-handshake 的 `device_id` 比较决定；函数 doc 中删去 "or it asks to pair"，并补充 "user-aborted pairing / 无法读取 device_id → throw，仅 device_id mismatch → return null"。

[Claude session ↗](https://claude.ai/code/session_016yyHSu1BGzdsDJMHevFu5K)

---

originalix (commented):
@ByteZhang1024

Reviewed commit `bc446f8566ed`.

- **P2 · USB presence is not scoped to the target Trezor**
- **P2 · Locked macOS devices are treated as stale BLE bindings**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12840#issuecomment-5262097723)

originalix (approved):

PR #12845, opened 2026-08-12
fix: sync portfolio on idle desktop BLE links

wabicai:
## Summary

- keep automatic Desktop Portfolio uploads on the active USB/Bridge transport
- suspend normal Desktop BLE uploads until a successful interactive BLE operation creates a reusable-link lease
- after 30 seconds of idle time, make a non-queued connected-only attempt within a 150-second lease and apply a 5-minute BLE transfer cooldown
- keep only the latest pending snapshot and preserve it when the hardware channel is busy
- invalidate the BLE lease when another user operation starts, the device disconnects, or the active transport changes
- pin Portfolio transport and connectId together without changing transport selection for unrelated background hardware calls
- keep the existing mobile BLE suspension and resume behavior unchanged

## Runtime boundaries

- `ServiceHardwarePortfolioSync` owns pending snapshots, lease generations, and timers in the bg JS runtime
- Electron preload owns the short connected-only scope
- the underlying Electron BLE connection is a process-shared native resource
- the main UI runtime only provides Portfolio snapshots and does not select hardware transport

## Hardware SDK

The App uses the aligned `@onekeyfe/hd-*` / `@onekeyfe/hwk-*` `1.2.0-alpha.119` package set. This PR does not change dependency versions.

## Validation

- 3 focused suites: 140 tests passed
- worktree and staged lint passed
- `git diff --check` passed
- full CI is running for the latest head


package.json
```diff
@@ -158,21 +158,21 @@
...
-    "@onekeyfe/hwk-ledger-adapter": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-ledger-connector-ble": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-ledger-connector-webhid": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-trezor-adapter": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-trezor-connector-electron-ble": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-trezor-connector-rn-ble": "1.2.0-alpha.108",
-    "@onekeyfe/hwk-trezor-connector-webusb": "1.2.0-alpha.108",
+    "@onekeyfe/hd-ble-sdk": "1.2.0-alpha.115",
+    "@onekeyfe/hd-common-connect-sdk": "1.2.0-alpha.115",
+    "@onekeyfe/hd-core": "1.2.0-alpha.115",
```
originalix: P1: `alpha.115` 没有包含 connected-only 能力

这里升级到的 `@onekeyfe/hd-core@1.2.0-alpha.115` 发布包对应 `gitHead=24c7cf18659ba9e49ce0630ef16863e136bfec72`；该版本的 `Device.acquire()` / `DeviceConnector.acquire()` 没有接收或下传 `reuseConnectedOnly`，`UploadPortfolioParams` 也没有这个字段。App 虽然向 `getDeviceState` 和 `uploadPortfolio` 传了参数，运行时仍会走普通 BLE acquire；当 30 秒 idle 到期而保活链路已经失效时，仍可能 direct-connect、扫描或重连，破坏本 PR 的核心保证。

真正贯通该参数的是 hardware-js-sdk PR #869 的 `e22f3a3252`，但该 PR 当前 closed 且未合并。请先发布包含等价实现的 SDK 并重新锁定版本，同时使用实际发布包补一条集成测试，断言调用 `connectConnectedOnly` 且不会调用普通 `connect` / `enumerate`。

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
```diff
@@ -528,6 +529,26 @@ class ServiceHardwareUI extends ServiceBase {
     const isThirdPartyVendor = getVendorProfile(
       device?.vendor ?? EHardwareVendor.onekey,
     ).isThirdParty;
+    // 嵌套调用会复用当前 OneKey 操作租约。只有租约所有者代表完整的用户交互，
+    // 内层调用不能提前恢复 Portfolio 同步。
+    const shouldNotifyPortfolioInteraction =
```
originalix, line 543: P2: Portfolio lease hook 扩散到全部 OneKey Desktop 操作

`shouldNotifyPortfolioInteraction` 这里只排除了嵌套调用和第三方设备，没有检查 `isProtocolV2ProductType` 或 V2 protocol。因此 Classic、Touch、Pro 等不具备 Portfolio 同步资格的设备，每次外层 Desktop 硬件操作仍会等待 `notifyInteractiveHardwareOperationStarted()`；成功后还会读取设备记录并尝试创建 BLE lease，扩大了公共硬件路径的延迟和状态面。

建议在进入通知前复用 Portfolio eligibility 的设备类型、protocol 和 vendor 条件，只让真正支持 Portfolio 的 V2 OneKey 设备接入这套 lease。

originalix:

@wabicai

Reviewed commit `8d8cabbbc0f6`. This report always reflects the latest reviewed changes.

### Review summary

This PR adds an idle desktop-BLE portfolio synchronization path that reuses an already-connected Noble endpoint without scanning, reconnecting, or interrupting interactive hardware work. The bg runtime owns pending snapshots, interaction generations, leases, cooldowns, and retries, while the Electron preload layer temporarily suppresses connection lifecycle calls against the shared main-process BLE link. Explicit transport pins and device-identity checks keep the deferred upload bounded to the intended endpoint, but unpinned desktop background reads can still resolve a BLE endpoint before selecting the persisted USB transport.

**What needs attention:** Confirm that every desktop background hardware call resolves its transport and connectId from one atomic decision, especially after an interactive BLE session leaves runtime transport state different from the persisted USB preference.

### Issues to address

- **P1 · Background endpoint resolution can disagree with the selected USB transport**
  - **Impact:** For an unpinned desktop background call, `getCompatibleConnectId()` runs before SDK transport selection and consults `getCurrentTransportType()`, which prefers the bg runtime's last active transport. If an earlier interaction left that state on `DesktopWebBle` while the persisted setting is USB, this method returns the persisted BLE UUID. `getSDKInstance()` then asks `shouldSwitchTransportType()`, whose background branch now selects the persisted USB setting, so the USB SDK receives a BLE endpoint. Silent state reads and other background hardware operations can consequently fail even though the device is reachable over USB.
  - **Suggested action:** Resolve one transport decision before endpoint selection and pass that explicit transport to both `getCompatibleConnectId()` and `getSDKInstance()` for desktop background calls.

### Validation gaps

- The focused Jest suites were not executed because this worktree has no installed dependencies.
- Live GitHub reviews, inline comments, and thread state could not be fetched from the current network environment. The inline suggestion therefore requires a duplicate and wording check before posting.
- The desktop flow was not exercised against a real Noble connection, so preload-scope timing and physical-link retention still need integration validation.

originalix (commented):
@wabicai

Reviewed commit `7d3ca77a1594`.

- **P1 · Desktop idle sync drops the reuse window after a busy retry**
- **P2 · Preflight failures clear an armed desktop BLE lease too early**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12845#issuecomment-5270167104)

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
```diff
@@ -370,6 +370,23 @@ export class HardwareConnectionManager {
...
+          EHardwareCallContext.BACKGROUND_NON_INTERACTIVE,
+        ].includes(resolvedHardwareCallContext);
+
+      if (isDesktopBackgroundCall) {
+        const targetType = await this.getCurrentTransportType();
+        return {
+          shouldSwitch: false,
+          targetType,
+        };
+      }
```
sidmorizon: **Mini hardware wallets can be driven over Bluetooth during background operations on desktop**

Severity: `non-severe`

Background hardware work on desktop now returns the currently active connection type without first checking whether the device is a Mini (early return at `packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:382-388`), so a Mini can be addressed over Bluetooth even though it only supports cable.
Impact: On a desktop where a Bluetooth-capable OneKey was used last, background actions for a Mini wallet fail instead of using its cable connection.

<details>
<summary>Ordering change bypasses the Mini USB-only branch</summary>

Previously `shouldSwitchTransportType` evaluated the force-transport branch and then the Mini branch (`connectId.startsWith('MI')` → WEBUSB/Bridge) before the background-context skip. The new desktop-background early return happens first, so for `BACKGROUND_TASK` / `BACKGROUND_NON_INTERACTIVE` contexts the Mini branch at `packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:410-423` is never reached. `getCurrentTransportType()` returns `this.actualTransportType || persistedSetting` (`packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:489-493`), which is `DesktopWebBle` after any prior BLE session with another device. `ServiceHardware.getSDKInstanceInternal` then builds/keeps the BLE SDK for that Mini call (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:994-1007`).

A fix would keep the Mini USB-only check ahead of (or inside) the new desktop-background early return.

</details>

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
```diff
@@ -1139,76 +1499,238 @@ class ServiceHardwarePortfolioSync extends ServiceBase {
...
+        targetKey,
+        transferAt: Date.now(),
+        walletId: eventPayload.walletId,
+      });
+      if (
+        desktopBleExecution &&
+        this.pendingDesktopBlePayloadByTargetKey.get(targetKey) === eventPayload
+      ) {
+        this.pendingDesktopBlePayloadByTargetKey.delete(targetKey);
+      }
```
sidmorizon, line 1801: **Queued desktop portfolio snapshot is never cleared after it is successfully sent to the device**

Severity: `non-severe`

The stored pending snapshot is only discarded when it is the exact same object as the one just uploaded (`===` comparison at `packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts:1719-1723`), which never holds because the upload path works on a rebuilt copy, so the already-sent snapshot stays queued forever.
Impact: After a successful desktop Bluetooth sync, the app keeps re-scheduling and re-processing an already-delivered snapshot on later Bluetooth sessions.

<details>
<summary>Rebuilt payload object breaks the identity check</summary>

`rememberPendingDesktopBlePayload` stores the payload produced by `resolveAuthorizedPortfolioPayload`, which always returns a fresh object literal (`packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts:289-300`). `tryRunDesktopBleIdleSync` passes that stored payload back into `syncSettledPortfolio`, which re-resolves it into yet another new object before reaching `uploadPreparedHardwarePortfolio`. Therefore `this.pendingDesktopBlePayloadByTargetKey.get(targetKey) === eventPayload` is always false and the entry is never deleted. Each later interactive BLE success then arms a lease, schedules an idle attempt, and re-runs the whole eligibility/cooldown/build pipeline for a snapshot that is ultimately dropped as a duplicate. Comparing the content hash (or deleting unconditionally on success for the target) would be robust.

</details>

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
```diff
@@ -528,6 +529,26 @@ class ServiceHardwareUI extends ServiceBase {
     const isThirdPartyVendor = getVendorProfile(
       device?.vendor ?? EHardwareVendor.onekey,
     ).isThirdParty;
+    // 嵌套调用会复用当前 OneKey 操作租约。只有租约所有者代表完整的用户交互，
+    // 内层调用不能提前恢复 Portfolio 同步。
```
sidmorizon: **New code comments are written in Chinese instead of English**

Severity: `non-severe`

Several comments added by this change are written in Chinese (for example at `packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts:532-533`), which the repository rules forbid.
Impact: Contributors who do not read Chinese cannot understand the reasoning behind these code sections.

<details>
<summary>Rule and affected locations</summary>

CLAUDE.md ("All comments must be in English and explain non-obvious logic only.") and `.cursor/rules/coding-patterns.mdc` ("All comments must be written in English") require English comments. New non-English comments in this PR:
- `packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts:532-533`
- `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:243`
- `packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts:1659-1660` and `:1676`

</details>

Suggested fix:

```ts
    // A nested call reuses the current OneKey operation lease. Only the lease
    // owner represents a complete user interaction, so nested calls must not
    // resume Portfolio sync early.
```

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
```diff
@@ -320,6 +355,215 @@ class ServiceHardwarePortfolioSync extends ServiceBase {
...
+
+    try {
+      await this.syncSettledPortfolio(eventPayload, syncGeneration, {
+        desktopBleExecution: {
+          bleConnectId: lease.bleConnectId,
+          generation: lease.generation,
+        },
+      });
+    } finally {
+      this.invalidateDesktopBleIdleLease({
```
sidmorizon: # [🟠 P1] 空跑尝试（duplicate 等）也会消耗一次性租约，150s 窗口被截断为 30s

## 问题

`tryRunDesktopBleIdleSync` 的 `finally` 无条件以 `attempt-finished` 调用 `invalidateDesktopBleIdleLease`，不区分本次 `syncSettledPortfolio` 是否真正发起过硬件调用。而桌面端 settled 事件在挂起分支里先经 `rememberPendingDesktopBlePayload` 缓存、去重检查在其之后才执行，所以 pending 快照完全可能与上次已上传内容一致（`buildPortfolioPayloadHash` 剔除了 `ts`，行情未变时哈希相同）。此时 30s 空闲定时器触发后走到 `skip-duplicate` 直接返回，没有任何硬件交互，但租约仍被销毁。

## 影响

只要 t+`30s` 时的 pending 快照是重复内容，`150s` 复用窗口实际被截断为 `30s`：其后（约窗口的 `80%`）到达的新快照只会落进 `desktop-suspended`，必须等下一次用户交互式 BLE 操作（可能相隔数小时）。叠加本 PR 中 pending 清理判等失效的问题（成功上传后 pending 不会被删除），首次成功上传之后的每一个租约都会在 30s 处被同一份重复快照空耗，特性在"桌面空闲时同步"这一核心场景下基本失效。

## 建议

让 `syncSettledPortfolio` 向调用方返回执行结果（如 `'uploaded' | 'duplicate' | 'noop' | 'error'`），仅在真正发起硬件调用（身份校验/上传）后才在 `finally` 中销毁租约；对 `duplicate`/挂起类空跑结果保留租约，并依赖 `scheduleDesktopBleIdleSync` 在 pending 更新时重新武装定时器。
[Claude session ↗](https://claude.ai/code/session_01DdH767W3Euoa1UEZQCvB6P)

---

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
```diff
@@ -1139,76 +1499,238 @@ class ServiceHardwarePortfolioSync extends ServiceBase {
...
+        }),
+        this.portfolioSyncDb.updateTargetState(targetKey, {
+          lastAttemptAt,
+        }),
+      ]);
+      // 即使状态写入先失败，也必须等硬件调用结束后才能释放全局硬件操作锁。
+      if (uploadResult.status === 'rejected') {
+        throw uploadResult.reason;
+      }
+      if (attemptStateResult.status === 'rejected') {
```
sidmorizon, line 1751: # [🟡 P2] 硬件上传成功后，`lastAttemptAt` 写入失败会同时丢掉去重与冷却记录并误报 error

## 问题

此改动把 `uploadPortfolioPackage` 与 `updateTargetState({ lastAttemptAt })` 放进同一个 `Promise.allSettled`，随后即使 `uploadResult` 成功，只要 `attemptStateResult` rejected 就直接 `throw attemptStateResult.reason`。基线实现是先 `await` 写入 `lastAttemptAt`、写失败则根本不发起上传（fail-safe 顺序）；新实现把写失败的发现点移到了硬件传输完成之后，却仍按"整体失败"处理。

## 影响

设备已经收到完整包，但异常路径跳过了 `commitProcessedArtifacts`：`lastContentHash`、`lastTransferAt`、`lastAttemptAt` 全部未持久化，`5 分钟` 冷却与内容去重同时失效，下一个租约会把同一份最大 `128KB`（`PORTFOLIO_PACKAGE_MAX_BYTES`）的包再次经空闲 BLE 链路传输；同时 `handleSyncError` 会把这次实际成功的同步置为 `status: 'error'`，误导排障。

## 建议

`uploadResult` 成功时不要因 `attemptStateResult` 失败而抛出：记录日志后继续走 `commitProcessedArtifacts`（其内部的 `updateTargetState` 会再次尝试持久化 `lastTransferAt`/`lastContentHash`，可一并补写 `lastAttemptAt`），仅在上传本身失败时进入异常路径。
[Claude session ↗](https://claude.ai/code/session_01DdH767W3Euoa1UEZQCvB6P)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -2759,10 +2803,35 @@ class ServiceHardware extends ServiceBase {
...
+        (await this.getCurrentTransportType())) ===
+        EHardwareTransportType.DesktopWebBle
+    ) {
+      throw new deviceErrors.DeviceNotFound({
+        payload: {
+          connectId: options.connectId,
+          inBluetoothCommunication: true,
+        },
+      });
+    }
```
sidmorizon: **Silent background portfolio sync can pop a "device not found" error dialog**

Severity: `non-severe`

The background check that refuses to use a Bluetooth link raises a device-not-found failure (`new deviceErrors.DeviceNotFound(...)` at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2823`) without marking it silent, so a routine background sync can throw a hardware error dialog in the user's face.

Impact: While simply using the app, the user may suddenly see a "device not found" hardware error dialog even though no hardware action was requested.

<details>
<summary>Mechanism: DeviceNotFound's constructor broadcasts a global UI dialog unless silentMode is set</summary>

`DeviceNotFound`'s constructor emits `EAppEventBusNames.ShowHardwareErrorDialog` whenever `props.silentMode` is falsy (`packages/shared/src/errors/errors/hardwareErrors.ts:753-772`). The new guard in `getDeviceState` constructs it with only `payload`, so the dialog event fires.

Reachable path: `ServiceHardwarePortfolioSync.getPreparedUploadDeviceIdentityStatus` calls `getDeviceState` with `hardwareCallContext: BACKGROUND_NON_INTERACTIVE` and `silentMode: true` but no pinned transport (`packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts:821-832`). The desktop-transport suspension check in `syncSettledPortfolio` runs much earlier (before the server pack round-trip and before acquiring the hardware lock), and the hardware-busy retry path re-enters `uploadPreparedHardwarePortfolio` a second later without re-checking the transport. If the active desktop transport becomes DesktopWebBle in that window, the new guard throws and the dialog is shown even though the caller explicitly asked for a silent call.
</details>

Suggested fix:

```ts
      throw new deviceErrors.DeviceNotFound({
        silentMode: true,
        payload: {
          connectId: options.connectId,
          inBluetoothCommunication: true,
        },
      });
```

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
```diff
@@ -373,20 +617,98 @@ class ServiceHardwarePortfolioSync extends ServiceBase {
...
+        return false;
+      }
+      const device = await localDb.getDeviceSafe(targetKey);
+      if (
+        this.desktopInteractiveGenerationByTargetKey.get(targetKey) !==
+        interactionGeneration
+      ) {
+        return false;
+      }
+      const bleConnectId = device?.bleConnectId || device?.uuid;
```
sidmorizon: **Idle Bluetooth reuse can be armed with a device serial number instead of a Bluetooth address**

Severity: `non-severe`

When a device has no stored Bluetooth address, the code falls back to the device's serial number (`device?.bleConnectId || device?.uuid` at `packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts:689`) and then uses that as the Bluetooth target, so the follow-up background sync talks to an identifier that does not exist on the Bluetooth link.

Impact: The background portfolio upload silently fails for those devices instead of reusing the live Bluetooth connection.

<details>
<summary>`IDBDevice.uuid` holds the device serial number, not a Noble peripheral id</summary>

`uuid` is populated from the device serial number (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:5290`, `packages/kit-bg/src/dbs/local/LocalDbBase.ts:6009`), while desktop BLE endpoints live in `bleConnectId` — that is exactly why `getPersistedDesktopBleConnectId` in `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:267-285` only considers `bleConnectId` and rejects values that alias the USB serial.

The lease built here is later used verbatim as `hardwareConnectId` for `getDeviceState` and `uploadPortfolioPackage` with `desktopBleReuseConnectedOnly: true` (`.../ServiceHardwarePortfolioSync.ts:1470-1471`, `:1662-1671`), and it also opens the preload connected-only scope keyed on that bogus id, so the connect suppression and the device call both target a nonexistent peripheral.
</details>

Suggested fix:

```ts
      const bleConnectId = device?.bleConnectId;
```

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3310,24 +3379,43 @@ class ServiceHardware extends ServiceBase {
...
     const hardwareSDK = await this.getSDKInstance({
       connectId: compatibleConnectId,
       hardwareCallContext: EHardwareCallContext.BACKGROUND_NON_INTERACTIVE,
+      ...(hardwareTransportType ? { hardwareTransportType } : {}),
+    });
+    return this.runInDesktopBleConnectedOnlyScope({
+      connectId: compatibleConnectId,
+      enabled: desktopBleReuseConnectedOnly,
+      task: () =>
+        convertDeviceResponse(() =>
```
sidmorizon: # [🟠 P1] 后台 BLE 上传缺少 `silentMode`，失败时会向空闲用户弹出硬件错误对话框

## 问题

`uploadPortfolioPackage` 的上传调用经 `convertDeviceResponse` 转换 SDK 响应，但没有传 `{ silentMode: true }`：

```ts
convertDeviceResponse(() =>
  hardwareSDK.uploadPortfolio(compatibleConnectId, {
    packageBase64,
  }),
),
```

`convertDeviceError` 对 `HardwareErrorCode.DeviceNotFound` 会构造 `HardwareErrors.DeviceNotFound({ payload, silentMode: options?.silentMode })`，而该构造函数在 `silentMode` 为假时无条件 `appEventBus.emit(EAppEventBusNames.ShowHardwareErrorDialog, ...)`（`hardwareErrors.ts:753-775`）。空闲 BLE 后台尝试恰好是最容易命中 `DeviceNotFound` 的路径：身份探测（该处已正确传 `silentMode: true`，见 `ServiceHardwarePortfolioSync.ts:821-832`）通过后，设备在打包/服务端往返的窗口内断链或关机，SDK 枚举匹配不到 `connectId`（hd-core 按 `descriptor.path === connectId` 精确匹配）即返回 `DeviceNotFound`。已有 review thread 指出的 `device?.uuid` 兜底产生无效 connectId 的场景同样会落到这里。

## 影响

用户没有发起任何操作时（空闲 `30s` 后的一次性后台尝试），桌面端可能突然弹出全局 "device not found" 硬件错误对话框。`handleSyncError` 的 catch 无法拦截，因为事件在错误构造函数内已经广播。这与另一条 thread 中 `getDeviceState` 卫语句直接 `new DeviceNotFound` 缺 `silentMode` 是两个独立的触发点，需分别修复。

## 建议

给后台上传路径的 `convertDeviceResponse` 传入静默选项（与 `_getDeviceStateLowLevel` 传 `{ silentMode }` 的做法对齐）：

```ts
convertDeviceResponse(
  () =>
    hardwareSDK.uploadPortfolio(compatibleConnectId, {
      packageBase64,
    }),
  { silentMode: true },
),
```

也可仅在 `desktopBleReuseConnectedOnly` / `BACKGROUND_NON_INTERACTIVE` 时开启，保持交互路径行为不变。
[Claude session ↗](https://claude.ai/code/session_01DdH767W3Euoa1UEZQCvB6P)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -410,6 +412,36 @@ class ServiceHardware extends ServiceBase {
...
+    ) {
+      throw new OneKeyLocalError(
+        'Desktop BLE connected-only scope is unavailable',
+      );
+    }
+    nobleBle.beginConnectedOnlyScope(connectId);
+    try {
+      return await task();
+    } finally {
+      nobleBle.endConnectedOnlyScope(connectId);
```
sidmorizon: # [🟡 P2] connected-only scope 无超时/TTL 兜底，卡住的 SDK 调用会长期屏蔽该设备的 connect/disconnect

## 问题

`endConnectedOnlyScope` 只在 `task()` settle 后的 `finally` 执行，preload 侧的 `desktopBleConnectedOnlyScopeCount` 计数没有任何过期机制。两个调用点的超时语义都保证不了释放：

- `uploadPortfolioPackage`：`hardwareSDK.uploadPortfolio` 外层没有任何超时包装。
- `_getDeviceStateLowLevel`：scope 在内层，而 `60s` 超时（`_getDeviceStateWithTimeout` / `DeviceMethodCallTimeout`）在外层——超时触发时调用方继续走错误处理、`getFeaturesMutex` 也随之释放，但内层 promise 仍持有 scope，直到 SDK 调用自行 settle。

仓库对 getFeatures/getDeviceState 加 `makeTimeoutPromise` 包装的注释（`// todo remove: sdk guarantees not to block this method`）本身就说明 SDK 调用存在长时间不返回的历史场景。

## 影响

scope 悬挂期间，该 `uuid` 的 `connect`/`disconnect`/`unsubscribe` 在 preload 层被静默替换为 `Promise.resolve()`，且按 uuid 全局生效、不区分调用方。若此时主进程链路已断（3 分钟 idle disconnect），用户随后的交互式 BLE 操作会走 `connect` no-op → `subscribe` 在主进程抛 `BleCharacteristicNotFound`，没有任何重连或配对入口，直到卡住的 promise settle 或重启应用。触发频率低，但后果是设备静默不可用且极难归因。

## 建议

双保险：

```ts
// 1) 对 scope 内 task 加有界超时，确保 finally 在可预期时间内执行
return await makeTimeoutPromise({
  asyncFunc: task,
  timeout: DESKTOP_BLE_REUSE_WINDOW_MS,
  timeoutRejectError: new deviceErrors.DeviceMethodCallTimeout(),
})();
```

2) preload 端为 begin/end 计数记录时间戳并设 TTL（如 `isDesktopBleConnectedOnlyScopeActive` 检查 stamp 是否过期），使 renderer 侧任何悬挂 promise 都无法永久毒化该设备的 BLE 通道。
[Claude session ↗](https://claude.ai/code/session_01DdH767W3Euoa1UEZQCvB6P)

---

package.json
wabicai: 已解决。后续提交已将整套 `@onekeyfe/hd-*` / `@onekeyfe/hwk-*` 依赖统一升级到 `1.2.0-alpha.119`，并接入 preload connected-only scope；当前调用不会走普通后台扫描或重连。

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
wabicai: 已在 `8fba1bfb2c` 解决。Portfolio 交互通知现在仅对 OneKey vendor、Protocol V2、`isProtocolV2ProductType` 设备启用；Classic/Touch/Pro 等不再进入 lease hook，并补了回归测试。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -240,6 +240,8 @@ export type IDeviceGetStateOptions = Omit<
   IDeviceGetFeaturesOptions,
   'params'
 > & {
+  /** 桌面端仅复用仍连接的 BLE 链路；链路失效时禁止扫描或重连。 */
```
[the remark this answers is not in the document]
wabicai: 已解决。相关新增生产代码注释已统一为英文；本次补充的测试注释也保持英文。

apps/desktop/app/preload.ts
```diff
@@ -234,8 +234,16 @@ const desktopApi = {
...
       ipcRenderer.invoke(EOneKeyBleMessageKeys.NOBLE_BLE_GET_DEVICE, uuid),
-    connect: (uuid: string) =>
-      ipcRenderer.invoke(EOneKeyBleMessageKeys.NOBLE_BLE_CONNECT, uuid),
+    connect: (uuid: string, options?: { reuseConnectedOnly?: boolean }) =>
+      ipcRenderer.invoke(
+        EOneKeyBleMessageKeys.NOBLE_BLE_CONNECT,
+        uuid,
+        options,
+      ),
+    connectConnectedOnly: (uuid: string) =>
```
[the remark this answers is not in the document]
wabicai: 已解决。`INobleBleApi` 当前已声明 `beginConnectedOnlyScope`、`endConnectedOnlyScope` 以及对应的 BLE API 契约，与 preload 暴露能力一致。

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
```diff
@@ -320,6 +355,215 @@ class ServiceHardwarePortfolioSync extends ServiceBase {
...
+  }: {
+    generation: number;
+    targetKey: string;
+  }) {
+    const lease = this.getActiveDesktopBleIdleLease({
+      generation,
+      targetKey,
+    });
+    const eventPayload =
+      this.pendingDesktopBlePayloadByTargetKey.get(targetKey);
```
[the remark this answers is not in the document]
wabicai: 已在 `8fba1bfb2c` 解决。新 settled 事件推进 generation 前，如果设备已有 desktop BLE pending，会立即替换为新 payload，不再等待 1 秒 debounce；已添加“debounce 完成前替换快照”的回归测试。

packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
wabicai: 已在 `8fba1bfb2c` 解决。Mini USB-only 判断现在位于 desktop background 保持当前 transport 分支之前；即使之前使用过 Desktop BLE，Mini 后台调用仍会切回配置的 WebUSB/Bridge。已补回归测试。

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
wabicai: 已在 `8fba1bfb2c` 解决。desktop BLE 上传成功后按 target + 当前 generation 清理 pending，不再依赖重建前后的对象引用相等；测试同时断言成功后 pending 已删除。

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
wabicai: 已解决。相关新增生产代码注释已改为英文，当前 diff 不再新增中文代码注释。

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
wabicai: 已在 `8fba1bfb2c` 解决。硬件上传成功后，即使首次 `lastAttemptAt` 写入失败，也不再误报整体失败；流程继续提交成功状态，并在 `commitProcessedArtifacts` 中一起重写 `lastAttemptAt`、hash、wallet 和 transfer 时间。已补回归断言。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: 已在 `8fba1bfb2c` 解决。desktop background BLE guard 构造 `DeviceNotFound` 时显式传入 `silentMode: true`，并通过测试确认不会发出全局硬件错误对话框事件。

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
wabicai: 已在 `8fba1bfb2c` 解决。desktop BLE lease 只接受持久化的 `device.bleConnectId`，不再回退到设备序列号 `uuid`；缺少 BLE peripheral ID 时不会创建租约，并有聚焦测试覆盖。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: 已在 `8fba1bfb2c` 解决。`uploadPortfolioPackage` 的 `convertDeviceResponse` 现在显式传入 `{ silentMode: true }`；SDK 返回 `DeviceNotFound` 时不会向空闲用户广播全局错误对话框，已添加回归测试。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: 已在 `8fba1bfb2c` 解决。preload connected-only scope 增加 150 秒 TTL；即使 renderer/SDK Promise 长期不 settle，过期 scope 也不会继续屏蔽该设备的 connect/disconnect/unsubscribe。正常 finally 仍会提前释放 scope。

packages/kit-bg/src/services/ServiceHardware/serviceHardwarePortfolioSync/ServiceHardwarePortfolioSync.ts
wabicai: 已在 `8fba1bfb2c` 解决。租约现在只在真正开始设备身份检查/硬件尝试后消费；duplicate 等空跑不会销毁 lease，且 duplicate 会清掉已处理 pending。busy 分支会在原 150 秒窗口内按 1 秒重试。

wabicai:
已处理本轮全部 review feedback，修复提交：`8fba1bfb2c`。\n\n- 后台 `getDeviceState` guard 与 Portfolio 上传错误转换均启用 `silentMode`，不再向空闲用户弹硬件错误对话框。\n- Desktop BLE busy 在现有 150 秒 lease 内重试；duplicate 等空跑不再消耗 lease；firmware preflight 拒绝不会提前清除 lease。\n- 修复 pending 快照替换/成功清理、BLE peripheral ID、Mini USB-only、成功上传后状态持久化，以及 connected-only scope TTL。\n- Portfolio lease hook 仅覆盖 OneKey Protocol V2 支持设备。\n\n聚焦测试：3 suites、126 tests 全部通过；相关 lint 通过。所有 14 个 review threads 已逐条回复并 resolve。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -2759,10 +2803,36 @@ class ServiceHardware extends ServiceBase {
...
+        EHardwareTransportType.DesktopWebBle
+    ) {
+      throw new deviceErrors.DeviceNotFound({
+        silentMode: true,
+        payload: {
+          connectId: options.connectId,
+          inBluetoothCommunication: true,
+        },
+      });
+    }
```
sidmorizon: **Background device reads on desktop fail outright whenever Bluetooth is the remembered connection**

Reported at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2816-2830` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

Any background request for the device's state is rejected with a "device not found" failure (`throw new deviceErrors.DeviceNotFound` at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2823`) before the code that would have fallen back to the saved cable connection, so routine background checks stop working while Bluetooth is the remembered link.
Impact: On Mac/Windows, background features that quietly read the device can start failing even though the device is plugged in over cable.

<details>
<summary>Guard runs before the existing background USB fallback in getCompatibleConnectId</summary>

The new guard in `getDeviceState` (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2806-2830`) triggers for every `BACKGROUND_TASK` / `BACKGROUND_NON_INTERACTIVE` call on `isSupportDesktopBle` platforms when `options.hardwareTransportType ?? getCurrentTransportType()` is `DesktopWebBle`, unless the caller opted into the new connected-only reuse (which additionally requires `hardwareTransportType === DesktopWebBle` and a `connectId`).

Previously such a call continued into `getCompatibleConnectId`, whose background branch (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:4485-4499`) intentionally returns the persisted USB `connectId` when no BLE binding exists — the comment there states background tasks must not be treated as offline just because BLE binding is missing. That fallback is now unreachable for these callers.

`getDeviceState` is a `@backgroundMethod()` on the shared background API, so the contract change affects every present and future background caller, not only the portfolio-sync path that this PR adds. The remembered transport (`actualTransportType`) can be stale BLE while the device is actually reachable over USB — exactly the situation this PR is trying to repair elsewhere — and in that state the read now hard-fails instead of degrading to USB.
</details>

wabicai: 已在 `9ee1444fbc` 解决。删除了通用 `getDeviceState()` 上的 Desktop background BLE guard，恢复所有既有调用进入 `getCompatibleConnectId()` 和 USB fallback 的能力。BLE 后台限制现在只保留在 Portfolio 上传路径：已打包任务进入硬件前重新检查当前 Transport，若切到 Desktop BLE 则挂起；若仍为 USB/Bridge，则把该 Transport 同时固定到身份读取和最终上传。新增了普通后台读取 USB fallback 与 Portfolio 打包后切换 BLE 的回归测试。

packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
```diff
@@ -370,6 +370,38 @@ export class HardwareConnectionManager {
...
+        const usbSetting = await this.getDesktopUsbSetting(connectProtocol);
+        const targetType =
+          usbSetting === 'webusb'
+            ? EHardwareTransportType.WEBUSB
+            : EHardwareTransportType.Bridge;
+        return {
+          shouldSwitch: this.actualTransportType !== targetType,
+          targetType,
+        };
+      }
```
sidmorizon, line 414: **Pinned transport choice is silently overridden for Mini devices on desktop**

Severity: `non-severe`

The Mini device shortcut now runs (`isMiniDevice` check at `packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:383-395`) before the explicitly pinned transport is read, so a flow that deliberately locks the connection type (firmware update, onboarding) can have that lock quietly replaced by the saved USB preference for Mini devices.
Impact: A Mini operation that was pinned to one connection method can end up using a different one, which can restart the connection layer in the middle of a locked flow.

<details>
<summary>Precedence inversion between the Mini shortcut and hardwareForceTransportAtom</summary>

Before this PR the order inside `shouldSwitchTransportType` was: (1) read `hardwareForceTransportAtom` and return the normalized forced type, (2) Mini shortcut, (3) context skip list. The PR moved the Mini shortcut to the top of the function so that it also precedes the new desktop-background early return (`packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts:397-403`). Moving it above the background early return is required by the new behaviour, but moving it above the force-transport branch is not, and it inverts an existing precedence: for `connectId` starting with `MI`, `forceTransportType` (set by `ServiceFirmwareUpdate` at `packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts:2187` and by the onboarding pages) is no longer consulted at all; the result is derived purely from `getDesktopUsbSetting()`.

In practice the two usually agree because the forced value was itself derived from the same USB setting, but any caller that pins a transport that differs from the persisted USB preference (for example pinning Bridge while the dev setting says `webusb`) will now silently get the preference instead of the pin, and `shouldSwitch` may become `true`, forcing an SDK reset inside a flow that intended to be locked.

Placing the Mini shortcut after the force-transport branch but before the new `isDesktopBackgroundCall` branch preserves both the old precedence and the new desktop-background behaviour.

</details>

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
```diff
@@ -540,6 +557,21 @@ class ServiceHardwareUI extends ServiceBase {
         autoToast: false,
       });
     }
+    if (
+      shouldNotifyPortfolioInteraction &&
+      platformEnv.isDesktop &&
+      device?.id
+    ) {
+      const generation = await this.backgroundApi.serviceHardwarePortfolioSync
```
originalix: P1: [Desktop lease is cleared before busy preflight can reject the operation]

On desktop this start notification invalidates the current portfolio-sync idle lease before `withHardwareProcessingInternal()` has passed its own pre-hardware checks. If the call then exits early because the OneKey mutex is still locked or another setup/preflight step rejects, no hardware operation ever touches the device and no success callback rearms the lease.

That leaves the pending bg snapshot without a BLE reuse window, so the idle upload stays suspended until the user completes another successful desktop BLE interaction.

Move `notifyInteractiveHardwareOperationStarted()` to the point where the outer preflight has already succeeded and the exclusive hardware operation is actually about to run, so preflight-only failures do not consume the lease.

packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
wabicai: 已在 `2ea27dba92` 修复。Mini 的 USB-only 分支现在会先尊重显式 `forceTransportType`，同时非 Mini 的 Desktop 后台调用仍保持当前 transport，不改变通用后台选择策略。新增回归测试覆盖“持久化 WebUSB、显式 pin Bridge”的场景。

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
wabicai: 已在 `2ea27dba92` 修复。`notifyInteractiveHardwareOperationStarted()` 已移动到 `withHardwareProcessingInternal()` 完成预检查、即将执行实际硬件回调的位置；内部 mutex/准备等预检查失败不会再消费现有 Portfolio lease。新增回归测试确认内部 preflight 拒绝时不会发送 started 通知。

originalix (commented):
@wabicai

Reviewed commit `2ea27dba9244`.

- **P1 · Desktop background calls now reuse stale transport state without an explicit pin**
- **P2 · Mini devices can now be forced onto unsupported BLE transport**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12845#issuecomment-5270167104)

wabicai:
已在 `8d8cabbbc0` 处理最新两项 blocking feedback：

- 删除通用 Desktop background 直接复用 runtime transport 的快捷分支，未显式 pin 的调用恢复原有 persisted-setting 选择；只有显式传入 transport 的 Desktop 后台调用才覆盖 force，避免 Portfolio 策略扩散到通用硬件读取。
- Mini 现在只接受 WebUSB/Bridge force；即使 force atom 是 Desktop BLE，仍回到配置的 USB transport。

同时补充 stale BLE runtime 和 Mini forced-BLE 回归测试。3 个聚焦套件共 140 个测试通过，lint 与 `git diff --check` 通过。PR 描述也已同步当前 USB/BLE 策略和 SDK `alpha.119` 状态。

originalix (commented):
@wabicai

Reviewed commit `8d8cabbbc0f6`.

- **P1 · Background endpoint resolution can disagree with the selected USB transport**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12845#issuecomment-5270167104)

originalix (approved):

PR #12848, opened 2026-08-12
fix: harden Swap recipient and warning flows (OK-53642, OK-55354, OK-60100)

zhaono1:
[OK-53642](https://onekeyhq.atlassian.net/browse/OK-53642) [OK-55354](https://onekeyhq.atlassian.net/browse/OK-55354) [OK-60100](https://onekeyhq.atlassian.net/browse/OK-60100)

----------

## Summary

- Restore the custom-recipient editor for external and address-book recipients.
- Route cross-chain swaps with a missing target account to recipient entry or address creation according to provider and account capabilities.
- Ignore stale asynchronous Swap warning results after token, provider, or tab scope changes.

## Intent & Context

This release fix combines the recipient-editor regression with the related Swap warning and recipient-routing corrections already validated on the same release branch. The target is to keep the source account usable, request a recipient only when the target account cannot create the required address, and prevent earlier asynchronous checks from overwriting the current Swap state.

## Changes

- Keep strict recipient ownership for quote and transaction logic while using the active account only as editor context fallback.
- Resolve the target-account capability before choosing address creation, custom recipient entry, or unsupported-network feedback.
- Preserve a confirmed recipient only while the current network, provider, mode, and Swap type still support it.
- Tag target-account unsupported alerts so they can be removed when a newer provider accepts a recipient.
- Guard warning commits with the current request, token pair, Swap type, and provider-recipient capability.
- Add focused coverage for editor readiness, account-network compatibility, automatic recipient lifecycle, hardware-wallet behavior, and stale warning races.

## Runtime Scope

- `main`: Swap UI, Jotai warning state, recipient selection, and editor rendering run in the main JS runtime with per-runtime JS copies.
- `bg`: the existing account-network capability check is called through the background API bridge; no background service contract or persistence schema changes.
- Native resources: no shared native resource ownership changes and no new startup-order dependency.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Web / iOS / Android shared Swap UI
- **Risk Areas**: recipient routing, account-network capability, asynchronous warning ordering, modal action state.
- **Out of Scope**: transaction building, quote-provider contracts, storage schema, native modules, and generated translations.

## Validation

- [x] `yarn jest packages/kit/src/states/jotai/contexts/swap/actions.test.tsx packages/kit/src/views/Swap/hooks/useSwapAccount.utils.test.ts packages/kit/src/views/Swap/utils/swapNoWalletWarningGuard.test.ts --runInBand` — 3 suites / 98 tests passed
- [x] Existing local Electron validation: recipient editor remained usable across repeated opens after selecting another account.
- [ ] Re-run Desktop/iOS runtime QA for automatic recipient entry, target address creation, and stale warning transitions on the bundled build.

## Issues

- [OK-53642](https://onekeyhq.atlassian.net/browse/OK-53642)
- [OK-55354](https://onekeyhq.atlassian.net/browse/OK-55354)
- [OK-60100](https://onekeyhq.atlassian.net/browse/OK-60100)

## Delivery

- Target branch: `release/v6.5.0`
- QA lane: `bundle-testing` / `release-app-bundles`

[OK-53642]: https://onekeyhq.atlassian.net/browse/OK-53642?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-55354]: https://onekeyhq.atlassian.net/browse/OK-55354?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60100]: https://onekeyhq.atlassian.net/browse/OK-60100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Swap/pages/modal/SwapToAnotherAddressModal.tsx
```diff
@@ -168,7 +173,7 @@ const SwapToAnotherAddressPage = () => {
     setSwapToAddress((v) => ({ ...v, address: undefined }));
   }, [setSwapToAddress, setSettings]);
 
-  return accountInfo && networkId ? (
+  return editorAccountInfo && networkId ? (
```
sidmorizon, line 183: # [🟠 P1] 目标网络无已创建地址时，编辑器仍然/新增渲染空白

## 问题

`getSwapRecipientEditorAccountInfo` 要求 `recipientAccountInfo?.account` 或 `activeAccount?.account` 至少一个存在才返回上下文。但 `IAccountSelectorActiveAccountInfo.account` 的类型是 `INetworkAccount | undefined`——当钱包在目标网络上尚未创建地址时（`canCreateAddress` 状态）它就是 `undefined`，且永远不会自行初始化。此时外部/地址簿收款人的 `accountInfo` 按设计也是 `undefined`，于是 `editorAccountInfo` 为 `undefined`，本行仍渲染 `null`——这恰恰是"用外部地址收款因为自己没有该链地址"的典型场景，OK-60100 的空白弹窗症状在该场景下并未修复。

同时这里比修复前更严格：旧代码非自定义收款路径下 `useSwapAddressInfo` 恒返回 truthy 的 `accountInfo` 对象（即使 `account: undefined`），旧门控 `accountInfo && networkId` 会正常渲染编辑器；新门控在无 `account` 时改为空白，属于该 PR 引入的行为回退。

## 影响

对"目标链无本人地址 + 外部收款人"的用户，重开收款人编辑器永久空白（非瞬态），无法修改或清除已保存地址；首次打开在 from/to 账户未创建地址或 all-network 解析窗口期也会闪空白。另注意 fallback 未校验 `activeAccount.account` 与 `networkId` 同网络：num=1 选择器同步窗口期内错配的 `accountId` 会让 `queryAddress` 在 `_fetchBadgesUncached` 的 `getAccount({networkId, accountId})` 处抛错，`queryAddressWithFallback` 将其降级为 `validStatus: 'unknown'`，表单短暂显示"检查请求失败"并禁用确认按钮。

## 建议

`AddressInputField` 的 `accountId` 本身是可选参数，后台 `queryAddress` 对缺失 `accountId` 已有完整兜底；渲染门控不必依赖 `account` 对象存在。建议放宽 resolver：无匹配账户时仍返回 `activeAccount`（或仅门控 `activeAccount?.ready`），并在网络不一致时不下发 `accountId`：

```ts
if (recipientAccountInfo?.account) return recipientAccountInfo;
if (activeAccount?.ready) return activeAccount; // account 可为 undefined，UI 已用 ?. 兜底
return undefined;
```

---

packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx
```diff
@@ -297,8 +317,15 @@ const SwapActionsState = ({
       visible: shouldShowIncognitoRecipientInput,
     });
 
+  const shouldOpenRecipientFromAction =
+    requiresCustomRecipient &&
+    swapActionState.label ===
```
sidmorizon, line 322: # [🟡 P2] 用本地化文案字符串相等判断驱动按钮行为，脆弱且失效时退化为死端

## 问题

`shouldOpenRecipientFromAction` 通过比较 `swapActionState.label === intl.formatMessage({ id: swap_page_button_enter_a_recipient })` 来识别"需要输入收款人"状态。该 label 在 `useSwapState.ts`（`swap_page_button_enter_a_recipient` 赋值处）生成后，还可能被后续条件（余额不足、`enter_amount`、`refresh_quotes` 等）覆盖，控制流因此依赖两处 `formatMessage` 输出的精确相等——这是把 UI 文案当作状态机标识使用。

## 影响

两类隐患：若未来 `useSwapState` 调整该 label 的构造（换 id、加插值、复用其他文案），`isActionDisabled` 中的 `!shouldOpenRecipientFromAction` 豁免会静默失效，跨链无目标账户的用户重新面对"disabled 死端按钮"（正是本 PR 要修复的 OK-55354 路径），且现有测试难以捕获；反之若其他 action 状态未来复用同一条翻译文案，点击会被错误劫持进收款人弹窗。

## 建议

在 `useSwapState` 设置该 label 的同一分支上输出语义标记，组件侧用标记而非文案判断：

```ts
// useSwapState.ts
infoRes.label = intl.formatMessage({ id: ETranslations.swap_page_button_enter_a_recipient });
infoRes.isEnterRecipientState = true;
infoRes.disable = true;

// SwapActionsState.tsx
const shouldOpenRecipientFromAction =
  requiresCustomRecipient && Boolean(swapActionState.isEnterRecipientState);
```

后续被覆盖的分支天然会重置该标记（每个分支都重建 label 时同步维护），语义与文案解耦。

---

packages/kit/src/views/Swap/hooks/useSwapAccount.ts
```diff
@@ -424,28 +458,45 @@ export function useSwapAddressInfo(type: ESwapDirectionType) {
...
+      return undefined;
+    }
+    return targetActiveAccount.canCreateAddress;
+  }, [
+    targetActiveAccount.canCreateAddress,
+    targetActiveAccount.network?.id,
+    targetActiveAccount.ready,
+    tokenNetworkId,
+    type,
+  ]);
```
sidmorizon, line 482: **Missing-address warning for the receiving side is decided from a different account than the one used**

Severity: `non-severe`

The ability to create the missing receiving address is read from the second account-selector slot (`targetCanCreateAddress` at `packages/kit/src/views/Swap/hooks/useSwapAccount.ts:461-477`) while the receiving address itself and the resulting create-address prompt come from the first slot, so the app can offer or hide address creation based on the wrong account.
Impact: Users can see a create-address prompt that acts on an account that cannot create it, or get no prompt at all and be stuck without a receiving address.

<details>
<summary>Why the two slots can diverge in useSwapAddressInfo</summary>

In `packages/kit/src/views/Swap/hooks/useSwapAccount.ts:299-305` the account backing the TO address info is slot `0` (source) whenever `customRecipientAddressActive` is false — which is the case exactly when there is no active custom recipient, i.e. the situation where the missing-address warning matters. However `targetCanCreateAddress` is always taken from `targetActiveAccount` (`useActiveAccount({ num: 1 })`), which the user can point at a different wallet/account through the "swap to another account" selector; only the recipient address setting is reset by `shouldResetSwapRecipientOnAccountNetworkSync`, not the slot-1 account itself.

`packages/kit/src/states/jotai/contexts/swap/actions.ts:2058-2073` now gates the CREATE_ADDRESS alert purely on `swapToAddressInfo.targetCanCreateAddress === true` (replacing the previous HD/HW/QR wallet-type check on the same address info), and then builds the alert action from `swapToAddressInfo.accountInfo`, i.e. the slot-0 account. The same mismatch feeds `shouldRequireSwapRecipientAddress` in `packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx:181-197`.

Additionally, `automaticTargetCanCreateAddress` (`packages/kit/src/views/Swap/hooks/useSwapAccount.ts:279-284`) compares slot-1's network against `toToken?.networkId`, while the memo compares against `tokenNetworkId`, so the same capability is computed with two different network conditions inside one hook.

</details>

packages/kit/src/states/jotai/contexts/swap/actions.ts
```diff
@@ -1792,7 +1842,20 @@ class ContentJotaiActionsSwap extends ContextJotaiActionsBase {
...
+            states: alertsWithoutNoConnectWallet,
+            shouldRemove: Boolean(
+              quoteResult &&
+              isCurrentQuoteResult &&
+              fromToken?.networkId &&
+              toToken?.networkId &&
+              fromToken.networkId !== toToken.networkId &&
+              swapProviderSupportReceiveAddress,
+            ),
+          });
```
sidmorizon, line 1858: **Blocking warning about an unsupported receiving network can vanish for hardware wallets while a new quote loads**

Severity: `non-severe`

The warning saying the receiving account does not support the target network is dropped (`removeSwapToAccountUnsupportedAlerts` at `packages/kit/src/states/jotai/contexts/swap/actions.ts:1848-1858`) whenever a cross-chain provider accepts a custom recipient, even for hardware wallets where the same warning is deliberately kept elsewhere, so the blocking message disappears while a quote refreshes.
Impact: Hardware-wallet users can briefly stop seeing the error that their device cannot use the destination network.

<details>
<summary>Inconsistency between the removal predicate and the check predicate</summary>

`shouldCheckSwapToAccountUnsupportedAlert` (`packages/kit/src/views/Swap/utils/swapNoWalletWarningGuard.ts:133-153`) intentionally returns true for hardware wallets even when `isCrossChain && providerSupportsRecipient`, and the PR adds a dedicated test for it. The removal path added in `checkSwapWarning` uses only `quoteResult && isCurrentQuoteResult && cross-chain && swapProviderSupportReceiveAddress` and ignores whether the receiving wallet is a hardware wallet, so the tagged `toAccountNetworkNotSupported` alert is stripped in the early-return branch (no networks / account not ready / waiting for an actionable quote) and only re-added once a full check completes.

</details>

packages/kit/src/views/Swap/hooks/useSwapAccount.ts
```diff
@@ -259,17 +260,50 @@ export function useSwapFromAccountNetworkSync() {
...
+      selectedRecipientAddress: swapToAnotherAccountAddressAtom.address,
+      swapEnableRecipientAddress,
+      fromNetworkId: fromToken?.networkId,
+      toNetworkId: toToken?.networkId,
+      incognitoMode: swapIncognitoMode,
+      providerSupportsRecipient: swapProviderSupportReceiveAddress,
+      swapType: swapTabSwitchType,
+      targetCanCreateAddress: automaticTargetCanCreateAddress,
+    },
+  );
```
sidmorizon: **A chosen cross-chain destination address can be silently dropped while the account is still loading**

Severity: `non-severe`

The automatically chosen destination address is turned off (`shouldActivateSwapCustomRecipientAddress` at `packages/kit/src/views/Swap/hooks/useSwapAccount.ts:285-298`) whenever the target account has not finished loading, not only when the wallet really can create the address, so the destination silently reverts to the user's own account.
Impact: A user who picked a different receiving address for a cross-chain swap can see it disappear, and the swap may be prepared toward their own wallet address instead.

<details>
<summary>How the not-ready state is conflated with "wallet can create the address"</summary>

`automaticTargetCanCreateAddress` (`packages/kit/src/views/Swap/hooks/useSwapAccount.ts:279-284`) yields `undefined` whenever `targetActiveAccount.ready` is false or `targetActiveAccount.network?.id !== toToken?.networkId` — which is the case during the debounced network sync in `useSwapFromAccountNetworkSync` (`packages/kit/src/views/Swap/hooks/useSwapAccount.ts:162-205`) and in any screen whose account-selector mirror only enables slot 0 (e.g. `packages/kit/src/views/Swap/pages/components/SwapStockDesktopContainer.tsx:924`, `packages/kit/src/views/Swap/pages/components/SwapReviewDialog.tsx:99`), where slot 1 resolves to the default empty account.

`shouldActivateSwapCustomRecipientAddress` only returns true for the automatic path when `targetCanCreateAddress === false` (`packages/kit/src/views/Swap/hooks/useSwapAccount.utils.ts:151-156`). With `undefined` the recipient is deactivated, so `useSwapAddressInfo(TO)` skips the custom-recipient branch and returns the wallet-derived address instead. The same `undefined` value also makes `shouldRequireSwapRecipientAddress` false, so the recipient row is hidden at the same moment (`packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx:181-208`).

A tri-state treatment (unknown → keep the previously confirmed recipient / wait) would avoid conflating "not resolved yet" with "wallet can create the address".

</details>

packages/kit/src/states/jotai/contexts/swap/actions.ts
```diff
@@ -1962,15 +2059,7 @@ class ContentJotaiActionsSwap extends ContextJotaiActionsBase {
         toToken &&
```
sidmorizon, line 2059: **The prompt to create a missing destination address may never appear after the account finishes loading**

Severity: `non-severe`

The offer to create the missing destination address is now decided from a value that resolves asynchronously (`swapToAddressInfo.targetCanCreateAddress === true` at `packages/kit/src/states/jotai/contexts/swap/actions.ts:2062`), but the warning check is not re-run when that value arrives, so the prompt can stay missing.
Impact: A user swapping to a chain where they have no address can be left with a disabled button and no way to create the address.

<details>
<summary>Warning re-check is not triggered by the new capability input</summary>

The TO create-address branch used to depend only on the wallet type of `swapToAddressInfo.accountInfo.wallet.id`, which is available immediately. It now depends on `targetCanCreateAddress`, which is derived from the num=1 active account and only becomes defined after the account-selector network sync completes (`packages/kit/src/views/Swap/hooks/useSwapAccount.ts:461-477`).

`useSwapWarningCheck` reads the address info from a ref and only re-runs `checkSwapWarning` when tokens, amounts, balances or quote state change (`packages/kit/src/views/Swap/hooks/useSwapState.ts:169-193`); neither `swapToAddressInfo` nor `targetCanCreateAddress` is in the dependency list. If the debounced check (300 ms) runs before the capability resolves, no create-address alert is produced, and nothing re-triggers the check until an unrelated dependency changes. Adding the resolved capability (or the TO address info) to the effect dependencies would make the alert appear once it is known.

</details>

originalix:

@zhaono1

Reviewed commit `640b97ec0064`. This report always reflects the latest reviewed changes.

### Review summary

This PR reshapes the foreground Swap recipient state machine so `main` can distinguish between three target-account outcomes for cross-chain routes: a resolved wallet address, a wallet that can create the target-network address, or a route that must collect an explicit recipient. It keeps capability checks behind the existing `main` -> `bg` service boundary, adds small shared decision helpers for recipient activation and display, and introduces a request-id guard intended to stop stale warning writes. The remaining correctness gap is that the new warning guard still does not bind results to the dispatch-time quote identity or input amount, so an older async warning run can overwrite the current quote-driven alert state after a requote or amount edit.

**What needs attention:** Verify that warning ownership follows the current `quoteId` and input amount through requotes, recipient changes, and target-account capability checks.

### Issues to address

- **P1 · Swap warning staleness guard ignores quote identity and input amount**
  - **Impact:** `checkSwapWarning()` now uses `swapWarningRequestIdAtom` to reject obsolete async results, but the guard only tracks request id, tab, provider-recipient support, and token pair. If the user edits the amount or a new quote is selected with the same token pair and the same recipient-support capability while an async warning step is still awaiting `settingsAtom.get()`, `checkAccountNetworkNotSupported()`, or the later currency/settings reads, the older invocation can still write the previous quote's min/max alerts, rate-difference result, and `quoteId` into current foreground state. That can briefly block review with stale errors or, worse, clear or replace the current quote's constraints with older data in the trade path.
  - **Suggested action:** Snapshot the dispatch-time `quoteId` (or another stable quote identity) and `swapFromTokenAmountAtom().value`, include both in the latest-check predicate, and add non-Stock regression tests that change amount or selected quote while a warning check is pending.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` passed.
- I reviewed the changed files and the directly related Swap state/hook call sites only; no whole-repo scan was needed.
- Targeted Jest execution did not run: `yarn jest packages/kit/src/views/Swap/hooks/useSwapAccount.utils.test.ts packages/kit/src/views/Swap/utils/swapNoWalletWarningGuard.test.ts packages/kit/src/states/jotai/contexts/swap/actions.test.tsx --runInBand` failed with `Couldn't find the node_modules state file`.
- Interactive validation of the new recipient flow was not performed in a running app session.
- Existing PR thread/comment coverage was not available in the runner evidence, so I did not emit a ready-to-post inline comment suggestion.

packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx
zhaono1: 这里保留当前实现。`requiresCustomRecipient` 先限定业务状态，最终 action label 再确保余额不足、输入金额、刷新报价等后续 action override 不会被 recipient handler 劫持；两处使用同一个 translation id，因此当前运行时不会因 locale 变化产生不一致。单独增加 boolean 会与现有最终 action state 形成第二份状态所有权。这条作为后续可维护性建议记录，不作为当前正确性 blocker。

packages/kit/src/views/Swap/hooks/useSwapAccount.ts
zhaono1: Confirmed as a residual account-ownership risk: the capability is currently read from selector slot 1 while the eventual create-address action is built from the resolved TO address info. There is no code change for this point in `d12c4d70d2`; it is being accepted as a known follow-up for this release delivery, not marked as fixed.

packages/kit/src/states/jotai/contexts/swap/actions.ts
zhaono1: Confirmed. The early waiting branch can remove the tagged target-account warning based on provider recipient support without preserving the hardware-wallet exception used by `shouldCheckSwapToAccountUnsupportedAlert`. This remains a known transient-warning risk in `d12c4d70d2`; closing the thread for this delivery is not a claim that the code was changed.

packages/kit/src/views/Swap/hooks/useSwapAccount.ts
zhaono1: Confirmed as a residual tri-state lifecycle risk. For the automatic path, `targetCanCreateAddress === undefined` currently deactivates the selected recipient while the target account is unresolved. No additional code change was made in `d12c4d70d2`; this is explicitly recorded as a known follow-up rather than reported as fixed.

packages/kit/src/states/jotai/contexts/swap/actions.ts
zhaono1: This re-check is already covered indirectly. `asyncRefContainer` depends on both `swapFromAddressInfo` and `swapToAddressInfo`, and that callback is itself in the warning effect dependency list. When `targetCanCreateAddress` changes, `swapToAddressInfo` changes, the callback identity changes, and the effect schedules `checkSwapWarningDeb` again with the refreshed ref values. No extra direct dependency is required for this path.

packages/kit/src/states/jotai/contexts/swap/actions.ts
```diff
@@ -1962,15 +2059,7 @@ class ContentJotaiActionsSwap extends ContextJotaiActionsBase {
...
-        (accountUtils.isHdWallet({
-          walletId: swapToAddressInfo.accountInfo?.wallet?.id,
-        }) ||
-          accountUtils.isHwWallet({
-            walletId: swapToAddressInfo.accountInfo?.wallet?.id,
-          }) ||
-          accountUtils.isQrWallet({
-            walletId: swapToAddressInfo.accountInfo?.wallet?.id,
-          }))
+        swapToAddressInfo.targetCanCreateAddress === true
```
sidmorizon, line 2062: **Prompt to create a missing destination address can disappear when the destination account context is not ready**

Severity: `non-severe`

The prompt that lets a user create the missing destination address is now shown only when a separate destination account slot reports it as creatable (`swapToAddressInfo.targetCanCreateAddress === true` at `packages/kit/src/states/jotai/contexts/swap/actions.ts:2062`), so the prompt vanishes whenever that slot is not yet aligned with the chosen destination network.
Impact: A user swapping to a chain where their wallet has no address may see no way to create it and be stuck with a disabled action.

<details>
<summary>Why the capability flag can be undefined while the old wallet-type check was true</summary>

`targetCanCreateAddress` is computed in `packages/kit/src/views/Swap/hooks/useSwapAccount.ts:466-482` and is `undefined` unless `type === TO`, `targetActiveAccount.ready`, and `targetActiveAccount.network?.id === tokenNetworkId`. `targetActiveAccount` is the account-selector slot 1 (`useActiveAccount({ num: 1 })`, `packages/kit/src/views/Swap/hooks/useSwapAccount.ts:283`).

Two concrete situations produce `undefined`:
- Slot 1 is only synced to the destination token network asynchronously through the 100ms-debounced `checkTokenForAccountNetworkDebounce` (`packages/kit/src/views/Swap/hooks/useSwapAccount.ts:161-207`); until it lands, `network?.id !== tokenNetworkId`.
- Surfaces that mount the swap UI with only slot 0 enabled (for example `packages/kit/src/views/Swap/pages/components/SwapStockDesktopContainer.tsx:924`, which renders `SwapActionsState`) always get `defaultActiveAccountInfo()` for slot 1 (`packages/kit/src/states/jotai/contexts/accountSelector/atoms.ts:251-257`), so `ready` is false.

The previous code decided purely from the wallet type of the TO address info (HD/HW/QR), which did not depend on slot 1 at all, so those cases used to render the create-address alert.

</details>

originalix (commented):
@zhaono1

Reviewed commit `640b97ec0064`.

- **P1 · Swap warning staleness guard ignores quote identity and input amount**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12848#issuecomment-5269655302)

originalix (approved):

zhaono1:
@originalix 已静态复核。该竞态缺口存在，但结合当前完整状态链，将其从 P1/blocking 降级为 non-blocking follow-up，本次暂不修改代码。

技术依据：
- amount 和 selected quote 都是 useSwapWarningCheck 的 effect 依赖，变化后会调度新的 warning check；旧任务只可能在新的 300ms debounce 调用启动、递增 request id 之前短暂写回，随后会被新检查纠正。
- 可见 Alert 已通过 alerts.quoteId 与当前 quoteId 匹配进行隔离；报价与输入不匹配时，现有 quoteResultNoMatch 也会阻止进入 Review。
- 剩余风险主要是 debounce 窗口内 hasError 未按 quoteId 隔离导致的瞬态按钮状态不一致。静态分析未发现可持续保留的 stale 状态。

当前发布范围接受这一瞬态风险。若后续出现可稳定复现的持久错误状态或错误进入 Review，再单独补充 quoteId 与 BUY/SELL 权威输入金额的 warning identity guard 及并发测试。本评论不做代码调整。

PR #12849, opened 2026-08-12
fix: support upgraded Hyperliquid TWAP orders(OK-59156)

Minnzen:
[OK-59156](<https://onekeyhq.atlassian.net/browse/OK-59156>)

----------

## Summary

Support upgraded Hyperliquid TWAP orders on `release/v6.6.0` (OK-59156): optional trigger price and buy maximum/sell minimum stop price, up to seven days of execution, and waiting/stopped order states. Remove fixed slice estimates and validate the 100 USDC total notional minimum.

Keep the SDK upgrade and TWAP feature together because the new request and state fields depend on SDK 0.33.3. Review the SDK/transport changes first, then the TWAP form, confirmation and order lists.

## SDK and release compatibility

- Upgrade `@nktkas/hyperliquid` from 0.32.2 to 0.33.3, adapt the ESM-only package and rews v4 transport, and retain serialized latency pings and subscription recovery.
- Adapt WebData2, outcome balances, nullable agent expiry, user abstraction, synchronous transport closure and the HTTP error constructor.
- Remove the old rews and SDK 0.32.2 patches. SDK 0.33.3 natively preserves the existing Chase `always_place` field; a regression test covers both Chase and ordinary amendments. Keep only the seven-day TWAP duration patch.
- Register 210 newly resolved mobile module paths, preserving every existing module ID and tombstone.
- Preserve 6.6.0 account/order guards, current TWAP history translations and unknown-status fallback, and the other dependency versions already on the release branch.

## TWAP behavior

- Add trigger/stop inputs under Advanced settings, carry them through confirmation and submission, clear the stop price on side changes, and validate the wire-formatted boundaries in both UI and background.
- Use the market mark price consistently for perp TWAP sizing and notional checks, with the account mark as fallback. Subscribe to this scalar only in perp TWAP mode so ordinary orders and unrelated market/account updates do not gain extra renders.
- Validate the 100 USDC minimum after truncating size to the submitted lot precision, consistently across UI, submit action and background exchange.
- Add the 7d preset and waiting/stopped display. Waiting orders do not accumulate elapsed time or progress before activation. Plain TWAP requests continue to omit `details`.

## Validation on the new base

- [x] `yarn install --immutable`; existing repository peer/patch warnings remain.
- [x] Verify the seven-day patch against the pristine SDK package.
- [x] Eleven targeted Jest suites: 148 tests pass (TWAP, SDK, portfolio, subscriptions, account guards, withdrawal reserve, order amendment and size conversion). Includes actual Jotai subscription lifecycle/render checks and rejection of quantities whose wire notional falls below 100 after truncation.
- [x] Four local TWAP request paths with mock signing: plain, trigger-above, trigger-below and stop-only; no network requests.
- [x] `yarn agent:check --profile commit`, including lint and TypeScript.
- [x] Android main and background Metro bundles compile with strict module ID registration; background Hermes bytecode compilation also succeeds (compiler global/eval warnings remain).
- [ ] Testnet immediate/trigger-above/trigger-below orders, buy maximum/sell minimum termination, seven-day duration and waiting/stopped lifecycle.
- [ ] Device/runtime regression of SDK transport reconnect and TWAP interaction; local bundle compilation is not device acceptance.
- [ ] Fresh CI and reviewer approval on the migrated head; old CI results do not validate this migration.

Remote status on `2615d56c67`: [Snyk security and license](https://app.snyk.io/org/rayston92/pr-checks/581df23b-9196-4eea-932d-caa528880389) both pass, and all review threads are resolved. The remaining CI checks are running; the release gate still requires the `release-ready` label and reviewer approval is pending. The earlier Snyk report was inaccessible, so its original failure cause remains unverified.

## Migration

Rebuilt the original PR's net changes on `release/v6.6.0@352c5ebfd7a6a00e19bedfd9a9b8d7c8a3ca2e4f`, resolving the intervening changes without importing the old release branch's unrelated history. This remains PR #12849. Automatic merging is disabled for the new review cycle.

Affected platforms: Mobile / Extension / Desktop / Web. UI owns form/confirmation and background owns transport, signing and submission. Mobile/extension have independent, version-locked JS runtimes and separate heap copies; desktop/web use one JS runtime. This change adds no native resource or database ownership changes.

[OK-59156](https://onekeyhq.atlassian.net/browse/OK-59156)

[OK-59156]: https://onekeyhq.atlassian.net/browse/OK-59156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -2896,6 +2901,29 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
...
+              triggerAbove = getTwapTriggerAbove({
+                triggerPrice,
+                markPrice: markPriceBN,
+              });
+              if (typeof triggerAbove !== 'boolean') {
+                throw new OneKeyLocalError(
+                  'TWAP trigger price must be positive and differ from the market price',
+                );
+              }
+            }
```
sidmorizon: **TWAP trigger price can be sent with the wrong up/down direction, or rejected right after the user confirms**

Severity: `non-severe`

The up/down direction of a TWAP trigger is decided by comparing the entered trigger price against the mark price (`getTwapTriggerAbove` at `packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts:2908-2911`) even though the form validated and displayed the same price against the mid price, so a trigger set between those two prices is submitted with the opposite direction.
Impact: A TWAP order can wait for the market to move the wrong way (or never start), and if the entered price happens to equal the mark price the order is rejected with an error right after the user confirms it.

<details>
<summary>Mid-price vs mark-price mismatch between validation and submission</summary>

- Pre-submit validation in `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:986-990` uses `latestEffectivePriceBN`, and `packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:354-357` uses `midPriceBN` (both derived from `useTradingPrice`, i.e. mid price).
- The value actually sent is computed in `placeTwapOrder` from `activeAssetCtxValue?.ctx?.markPrice` (`packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts:2889-2911`).
- If `mid < trigger < mark`, the UI accepts it as "above" but the submitted `triggerAbove` becomes `false` (activate when mark drops below the trigger), which is the opposite of the user's intent.
- If `trigger === mark` but `trigger !== mid`, `getTwapTriggerAbove` returns `undefined` and `placeTwapOrder` throws `TWAP trigger price must be positive and differ from the market price` after the confirm dialog was already accepted. The background service additionally hard-fails when the direction is missing (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts:1421-1423`).
- Note the background also rounds the trigger price to HL tick size after the direction was computed on the unrounded value (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts:1404-1419`), which can make the rounded trigger equal the reference price while `a` still says "above".

</details>

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/shared/src/utils/hyperliquidTwapUtils.ts
```diff
@@ -0,0 +1,100 @@
...
+export function getActiveTwapRuntimeStatus({
+  reportedStatus,
+  triggerPrice,
+  executedSize,
+}: {
+  reportedStatus?: 'activated' | 'waitingForTrigger';
+  triggerPrice?: string | null;
+  executedSize: BigNumber.Value;
+}): 'activated' | 'waitingForTrigger' {
+  if (reportedStatus) {
```
sidmorizon, line 82: # [🟡 P2] 过期的 `reportedStatus` 会压过 `executedSize > 0` 的实时成交证据

## 问题

`getActiveTwapRuntimeStatus` 中 `if (reportedStatus) { return reportedStatus; }` 无条件优先采用来自 twapHistory 的状态。但 active TWAP 的 `state`（含 `executedSz`）来自 `webData2` 订阅，而 `reportedStatus` 来自独立的 `userTwapHistory` 快照，两个 WS 通道之间没有顺序保证。触发单激活后，`webData2` 先推送 `executedSz > 0`，若 history 快照尚未送达（或 WS 重连丢了一次快照），`activeTwapStatusById` 中仍是旧的 `'waitingForTrigger'`，此时即使订单已在持续成交，函数仍返回 `'waitingForTrigger'`。

```ts
if (reportedStatus) {
  return reportedStatus;
}
```

## 影响

激活窗口期（或 history 通道滞后期间）内，一个正在执行的 TWAP 在 active 列表中显示为「Pending」，`progressText`、`runningTimeText` 均显示 `--`，成交进度被隐藏。用户可能误以为订单未启动而去终止一个实际正在成交的订单。`hyperliquidTwapUtils.test.ts` 中 "keeps a triggered TWAP pending until history reports activation" 固化了这一行为。

## 建议

`executedSize > 0` 是激活的客观证据，应短路掉过期的 reported 状态：

```ts
const executedSizeBN = new BigNumber(executedSize);
if (executedSizeBN.isFinite() && executedSizeBN.gt(0)) {
  return 'activated';
}
if (reportedStatus) {
  return reportedStatus;
}
```

---

packages/kit/src/views/Perp/components/OrderInfoPanel/List/PerpTwapList.tsx
```diff
@@ -1281,6 +1388,30 @@ function PerpTwapList({
     return rawHistory;
   }, [currentAccountAddress, historyAccountAddress, rawHistory]);
 
+  const activeStatusByTwapId = useMemo(() => {
```
sidmorizon: # [🟡 P2] 状态推导逻辑在两个列表组件中整段重复

## 问题

`activeStatusByTwapId` 这段「按 `record.time` 取每个 `twapId` 最新 history 记录并折叠为 `'waitingForTrigger' | 'activated'`」的 useMemo（约 24 行），与 `PerpOpenOrdersList.tsx` 第 283 行的 `activeTwapStatusById` 逐行相同；此外 `formatTwapPrice` 也分别在本文件和 `MobileTwapOpenOrdersRow.tsx` 中重复定义了一份相同实现。

## 影响

这是决定 active TWAP 显示 `Pending` 还是 `Activated` 的业务规则（含 tie-break 和非 waiting 状态折叠为 activated 的策略）。两份拷贝一旦其中一处后续修改（例如调整 `record.time` 比较或引入 `stopped` 的特殊处理），Open Orders 页签和 TWAP 页签会对同一订单显示不一致的状态；价格格式化规则同理。

## 建议

在 `packages/shared/src/utils/hyperliquidTwapUtils.ts`（`getActiveTwapRuntimeStatus` 已在此处）新增一个纯函数并让两处复用，例如：

```ts
export function buildActiveTwapReportedStatusMap(
  records: Array<{ twapId?: number; time: number; status: { status: string } }>,
): Map<number, 'activated' | 'waitingForTrigger'> { ... }
```

`formatTwapPrice` 同样可以下沉为共享 helper。

---

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
Minnzen: Fixed in f7139a4c39. Perp TWAP validation and submission now share the mark-price reference, while spot continues to use mid price. The action also formats the trigger to Hyperliquid wire precision before deriving triggerAbove.

packages/shared/src/utils/hyperliquidTwapUtils.ts
```diff
@@ -0,0 +1,100 @@
...
+  timestamp: number;
+  now: number;
+  endTime?: number;
+  minutes: number;
+}): number {
+  if (status === 'waitingForTrigger') {
+    return 0;
+  }
+  const totalMs = Math.max(0, minutes) * 60_000;
+  return Math.min(Math.max((endTime ?? now) - timestamp, 0), totalMs);
```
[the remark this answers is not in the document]
Minnzen: Fixed in f7139a4c39. Active runtime info now retains the activated history record time and passes it to the elapsed-time calculation, so the waiting period is not counted as execution time.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1394,6 +1394,33 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+          `TWAP ${fieldName} price is too small for HL tick size`,
+        );
+      }
+      return formattedPrice;
+    };
+    const triggerPrice = formatOptionalTwapPrice(
+      params.triggerPrice,
+      'trigger',
+    );
+    const stopPrice = formatOptionalTwapPrice(params.stopPrice, 'stop');
```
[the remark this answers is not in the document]
Minnzen: No additional stop-side blocker was added. The current product requirement labels stopPx as highest or lowest price but does not forbid an already-crossed termination price, and the SDK accepts any positive termination price. Rejecting it locally could block an intentional immediate stop; background validation still enforces positive, wire-safe prices.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/PerpTwapList.tsx
Minnzen: Addressed in 08ebcf07a0. The history-to-runtime mapping and TWAP price formatter are now shared pure helpers used by both active-order views, including the activation timestamp rule.

packages/shared/src/utils/hyperliquidTwapUtils.ts
```diff
@@ -0,0 +1,176 @@
...
+  const latestRecordByTwapId = new Map<number, (typeof records)[number]>();
+  records.forEach((record) => {
+    if (record.twapId === undefined) {
+      return;
+    }
+    const previous = latestRecordByTwapId.get(record.twapId);
+    if (!previous || record.time > previous.time) {
+      latestRecordByTwapId.set(record.twapId, record);
+    }
+  });
```
sidmorizon: **TWAP orders can stay stuck showing "Pending" because the status lookup key may never match**

Severity: `non-severe`

The pending/active status of each TWAP order is looked up by an identifier that is treated as possibly missing on history entries (`twapId?: number` at `packages/shared/src/utils/hyperliquidTwapUtils.ts:28`), so if the history feed does not carry that identifier every lookup fails and the order keeps showing the fallback state instead of its real one.
Impact: A triggered TWAP order can keep displaying "Pending" with no elapsed time and no fill progress until its very first fill arrives, even after it has actually started running.

<details>
<summary>How the runtime status correlation can silently no-op</summary>

`buildActiveTwapRuntimeInfoById` (`packages/shared/src/utils/hyperliquidTwapUtils.ts:26-58`) builds `Map<twapId, info>` from TWAP history records, skipping any record whose `twapId` is `undefined`. Because the parameter type declares `twapId?: number`, passing `ITwapHistoryRecord[]` (`packages/shared/types/hyperliquid/sdk.ts:23`, i.e. `HL.TwapHistoryResponse[number]`) type-checks even if the SDK response shape has no `twapId` field at all — the Hyperliquid `twapHistory` payload exposes `time`, `state` and `status`, while `twapId` is present on `twapStates`/`twapSliceFills`.

If the field is absent at runtime, the map is always empty, so both callers (`packages/kit/src/views/Perp/components/OrderInfoPanel/List/PerpTwapList.tsx:1398-1401,1805-1810` and `packages/kit/src/views/Perp/components/OrderInfoPanel/List/PerpOpenOrdersList.tsx:286-289,509-513`) always pass `reportedStatus: undefined` and `activatedAt: undefined` into `getActiveTwapRuntimeStatus`. That falls back to the heuristic in `packages/shared/src/utils/hyperliquidTwapUtils.ts:76-78`: any order with a trigger price and zero executed size is reported as `waitingForTrigger`, which suppresses running time and fill progress (`getTwapElapsedMs` returns 0, progress renders `--`) until the first fill lands.

Worth confirming against a real `twapHistory` response whether the records actually contain `twapId`; if they do not, the correlation needs another key (e.g. matching on coin + state timestamp) or the history state itself.
</details>

Minnzen: Confirmed and fixed in 30248c4626. Hyperliquid's TWAP history payload does not guarantee twapId, so both active-order views now correlate history with the active state using coin + state.timestamp, which is present on both representations of the same order. Added a regression test covering a history record without twapId.

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
```diff
@@ -2257,26 +2130,56 @@ function PerpTradingForm({
...
+            id: isBuy
+              ? ETranslations.perp_scale_upper_price_placeholder__desc
+              : ETranslations.perp_scale_lower_price_placeholder__desc,
+          })}
+          value={formData.twapStopPrice ?? ''}
+          onChange={(value) => updateForm({ twapStopPrice: value })}
+          szDecimals={sizeSzDecimals}
+          isSpot={isSpot}
+          isMobile={isMobile}
+          disabled={isSubmitting}
```
sidmorizon, line 2180: **Scheduled-buy/sell stop price keeps its old number when the trade direction is flipped, so the order can be cancelled the instant it is placed**

Severity: `non-severe`

The stop price field is relabelled from "Upper Price" to "Lower Price" when the user switches between buy and sell (`renderTwapDetailsSection` at `packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:2155-2171`) while the previously typed number is kept and never checked against the current market price, so a scheduled order can be stopped immediately after submission.
Impact: A user who flips direction (or mistypes) can have their scheduled order silently terminated right away instead of executing.

<details>
<summary>How the stop price meaning inverts across side changes</summary>

`formData.twapStopPrice` is a single field whose meaning depends on `formData.side`: for a buy it is rendered as "Upper Price" (terminate when the market rises above it) and for a sell as "Lower Price" (terminate when the market falls below it).

- `handleSideChange` (`packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:1329-1341`) only resets `size`/`sizePercent`/`sizeInputMode`; `twapStopPrice` is preserved.
- The only validation performed anywhere is `stopPriceBN.isFinite() && stopPriceBN.gt(0)` — see `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:1014-1025`, `packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:389-401` and `packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts:2939-2946`. None of them compare the stop price with the reference/mark price or with the order side.
- Example: market 100, user enters 110 as the buy "Upper Price", then toggles to sell. 110 is now the "Lower Price" and 100 <= 110, so the order is terminated by the stop condition as soon as it is accepted.

A direction-aware guard (stop above the reference price for buys, below for sells) plus clearing `twapStopPrice` on side change would prevent this.

</details>

Minnzen:
已处理 [automated review](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5268706999) 的两个 blocking issue，并随 SDK 升级提交 `83501874ee` 推送：

- P1：切换买卖方向时清空 `twapStopPrice`；main 确认、action 提交和 bg exchange 三层都按方向、mark price 与 trigger boundary 校验最高/最低价。
- P2：perp TWAP 的表单金额换算、slider size、确认金额、100 USDC 校验和后台提交统一使用 mark price；spot 仍使用 mid price。

另外已补齐原 PR 缺失的真实 SDK 升级：`@nktkas/hyperliquid` 0.32.2 → 0.33.3，删除旧版 TWAP 大 patch 和 rews 2 patch，仅保留 0.33.3 的 7 天 duration 上限最小 patch。

验证：4 suites / 49 tests、TypeScript、commit profile、pristine patch check、Android background Metro bundle、7 天 TWAP 签名烟测均通过。该 review 是 issue comment，GitHub 没有单独的 resolve 状态；对应问题已经由本提交关闭。

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
Minnzen: Fixed in 83501874ee. `updateTradingForm` now clears `twapStopPrice` whenever the side changes unless the caller explicitly supplies a replacement. The button, confirm hook, action, and bg exchange boundary also reject buy stops at/below the mark or trigger boundary and sell stops at/above it, so RPC callers cannot bypass the UI guard.

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -2901,6 +2919,48 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
...
+            if (
+              stopPrice &&
+              !isTwapStopPriceValid({
```
sidmorizon, line 2961: # [🟡 P2] main 用原始 stop 价校验、bg 用截断后的 stop 价复检,边界一格内确认后仍会被拒

## 问题

main 侧(此处及 `TradingButtonGroup`、`useOrderConfirm`)对 `stopPrice` 直接用 `formData.twapStopPrice?.trim()` 的**原始值**做 `isTwapStopPriceValid` 校验,并原样传给 bg;而 bg 的 `placeTwapOrder` 先经 `formatOptionalTwapPrice` → `formatHlPrice`(对非整数价**截断/floor**到 5 位有效数字)后,再用同一个 `params.referencePrice` 复检同一边界。同文件中 `triggerPrice` 已经先 `formatHlPrice` 再校验并下发(上方 `rawTriggerPrice` → `triggerPrice`),但 stop 没有同等处理。

```ts
const stopPrice = formData.twapStopPrice?.trim(); // raw
...
isTwapStopPriceValid({ stopPrice, referencePrice: markPriceBN, triggerPrice })
```

## 影响

buy 单 stop 落在「边界之上、且截断后回到边界」的窗口时:例如 mark = `123450`(szDecimals=5 的 perp),用户输入 stop `123450.5` — main 校验 `123450.5 > 123450` 通过、确认框通过、`useOrderConfirm` 已先执行 `resetTradingForm()`,随后 bg 截断为 `123450` 后 `gt` 不成立,抛出 `TWAP maximum price must be above the market and trigger price` — 用户在确认之后收到失败 toast 且表单输入已被清空。低价资产(如 `0.123456` → `0.12345`)同样可复现;sell 侧镜像存在。

## 建议

与 trigger 的处理保持一致:在本文件先 `formatHlPrice(rawStopPrice, szDecimals, isSpot ? 'spot' : 'perp')` 得到 wire 值,用该值做 `isTwapStopPriceValid` 校验并传给 bg(bg 二次 format 幂等),使 main 校验、bg 校验与实际签名 payload 三者对同一数值判定:

```ts
const rawStopPrice = formData.twapStopPrice?.trim();
const stopPrice = rawStopPrice
  ? formatHlPrice(rawStopPrice, szDecimals, isSpot ? 'spot' : 'perp')
  : undefined;
if (rawStopPrice && !stopPrice) {
  throw new OneKeyLocalError('TWAP stop price is too small for HL tick size');
}
```

---

Minnzen:
CI follow-up:

- `Supply Chain - Minimum Release Age` initially failed because the SDK's `@noble/hashes ^2.2.0` floated to 2.3.0, which was only 6 days old. Commit `c3b440be49` adds a range-specific resolution to the reviewed 2.2.0 release. The exact diff check reports 4 packages OK / 0 too young.
- Full lint then identified an obsolete `IHyperliquidFrontendOrder[]` assertion after the 0.33.3 type narrowing. Commit `49369a1d61` removes only that assertion.
- Review identified that main validated raw TWAP boundary prices while bg validated wire-formatted prices. Commit `ae79df4542` makes button, confirmation, main, and bg validation use the same Hyperliquid wire price before submission and adds the truncation-boundary regression case.

Latest HEAD `ae79df4542` is green: 17 checks passed, 0 failed, 0 pending. All review threads are resolved. Local commit checks and the targeted TWAP test also pass; PR readiness is waiting only for the repository's required human approval.

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
Minnzen: 已处理并推送 `ae79df4542`。

现在 trigger/stop 在按钮校验、确认、main action 与 bg action 中都会先通过同一个 `formatTwapPriceForOrder` 转成 Hyperliquid wire price，再进行边界校验和下单。这样 UI 校验、bg 复检与签名 payload 使用完全相同的值；截断后落到边界或变成 0 的输入会在确认前被拒绝，不会再出现表单重置后才失败。

补充了 `123450.5 -> 123450` 和低价 spot 精度用例；定向测试 12/12、`agent:check --profile commit` 均通过。

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
```diff
@@ -928,7 +894,11 @@ function PerpTradingForm({
   // Reference Price: Get the effective trading price (limit price, market price, or trigger effective price)
   const [, referencePriceString] = useMemo(() => {
     let price = new BigNumber(0);
-    if (formData.orderMode === 'trigger' && formData.triggerOrderType) {
+    if (formData.orderMode === 'twap') {
+      price = isSpot
+        ? midPriceBN
+        : new BigNumber(activeAssetData?.markPx ?? '');
```
sidmorizon: **Order size cannot be entered in dollars for time-sliced perp orders before account data loads**

Severity: `non-severe`

The dollar-per-unit reference used by the size box for time-sliced (TWAP) perp orders is taken only from the per-account asset feed (`new BigNumber(activeAssetData?.markPx ?? '')` at `packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:899`), which is empty until that account-scoped feed arrives, so typing a dollar amount produces no order size at all.
Impact: Users switching to the TWAP tab before their account data streams in (or right after switching markets) can type a USD/margin amount and the order size stays blank, blocking order entry.

<details>
<summary>Why the reference price becomes 0 and what it breaks</summary>

`perpsActiveAssetDataAtom` is only populated by the per-user ACTIVE_ASSET_DATA subscription and is explicitly reset to `undefined` when no account matches or on market switch (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:1838-1881`, and `clearActiveAssetData` in the switch flow). When it is `undefined`, `new BigNumber('')` is NaN and `formatPriceToSignificantDigits` maps that to `'0'` (`packages/shared/src/utils/perpsUtils.ts:742-754`).

`SizeInput` gates every USD↔token and margin↔token conversion behind `hasValidPrice = priceBN.isFinite() && priceBN.gt(0)` (`packages/kit/src/views/Perp/components/TradingPanel/inputs/SizeInput.tsx:104-135`), so a `'0'` reference silently returns `''` for the computed size.

Before this PR the TWAP path fell through to the `formData.type === 'market'` branch and used `midPriceBN`, which is available without an account. Elsewhere in this same PR the mark price is read from the market-wide context atom instead (`activeAssetCtx?.ctx?.markPrice` in `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:468-475` and `packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:84-91`), so the two sources are inconsistent.

</details>

Minnzen: Fixed in 0a4ed3c2ed. TWAP form and calculation paths now use the market-wide active-asset mark price while account-scoped data is still loading. Added a regression test for market mark = 100 with `activeAssetData` undefined; the order value is now 200 instead of 0. Validation passed: 5 targeted suites / 51 tests plus commit lint and TypeScript checks.

packages/shared/src/utils/hyperliquidTwapUtils.ts
```diff
@@ -0,0 +1,235 @@
...
+    triggerPriceBN.lte(0)
+  ) {
+    return false;
+  }
+  const activationBoundary = isBuy
+    ? BigNumber.max(referencePriceBN, triggerPriceBN)
+    : BigNumber.min(referencePriceBN, triggerPriceBN);
+  return isBuy
+    ? stopPriceBN.gt(activationBoundary)
+    : stopPriceBN.lt(activationBoundary);
```
sidmorizon: **Scheduled trades with a price limit between the trigger and the current market price are rejected**

Severity: `non-severe`

The optional price limit on a scheduled (TWAP) order is required to sit beyond both the current market price and the activation price (`BigNumber.max`/`BigNumber.min` at `packages/shared/src/utils/hyperliquidTwapUtils.ts:204-206`), so a buy that is set to start once the market drops cannot use a limit placed between the start price and today's price.
Impact: Users cannot submit otherwise valid scheduled orders and see a confusing "must be above/below market" error.

<details>
<summary>Why the combined market+trigger boundary is too strict for triggered orders</summary>

`isTwapStopPriceValid` computes `activationBoundary = isBuy ? max(referencePrice, triggerPrice) : min(referencePrice, triggerPrice)` and requires the stop price to be strictly beyond it. Once a trigger is set, the order only becomes active when the market reaches the trigger, so the meaningful comparison at activation time is the trigger price, not today's market price.

Concrete case: market = 100, buy trigger = 90 (buy the dip), desired maximum price = 95. At activation the market is ~90, so a 95 maximum is well-formed, but the current rule demands `stop > 100` and rejects it. The mirror case exists for sells with a trigger above the market.

The same predicate is enforced in three places, so the order is blocked in the UI and again in the background: `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:1046-1055`, `packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:428-435`, and `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts:1444-1452`.

</details>

Minnzen: Fixed in dc32cfe086. `isTwapStopPriceValid` now uses the current mark only when no trigger is present; triggered orders validate the buy maximum or sell minimum against the trigger activation price. This permits market 100 / buy trigger 90 / maximum 95 and the mirrored sell case, while prices on the wrong side of the trigger remain rejected. Added regression coverage; 5 targeted suites / 52 tests and commit lint/type checks passed.

originalix:

@Minnzen

Reviewed commit `a32f28fc4f50`. This report always reflects the latest reviewed changes.

### Review summary

This PR upgrades the Hyperliquid integration boundary to SDK 0.33.3 and threads the expanded TWAP contract through shared validation helpers, background exchange/subscription services, foreground Jotai state, and Perps UI surfaces. It adds trigger/stop-aware TWAP configuration, seven-day durations, and runtime-status display while keeping `bg` responsible for transport, account ingestion, and caches and `main` responsible for copied presentation state. The main correctness gap is that widened spot-balance responses without a token identifier are still filtered out before live atoms and cached display totals are written, so affected accounts can under-report holdings even though the snapshot path now marks itself degraded.

**What needs attention:** Confirm how positive spot balances without token identifiers should flow through background valuation, cached display totals, and degraded-state UX instead of disappearing.

### Issues to address

- **P1 · Outcome spot balances disappear from live state and cached totals**
  - **Impact:** Hyperliquid 0.33.3 can return positive spot-balance variants without a `token` field. The live ingest now filters those records before writing `spotBalancesAtom` and `perpsSpotBalancesAtom`, and snapshot assembly repeats the filtering before `spotTotalUsd` is derived. That means an account can still hold positive value while the balances UI and persisted display snapshot publish a smaller total without any live degraded-state guard.
  - **Suggested action:** Preserve the widened balance variant through the live state/cache boundary, or explicitly mark the result degraded and avoid writing a misleading total until that balance can be represented safely.

### Validation gaps

- Targeted Jest reruns were blocked in this workspace because Yarn could not find the `node_modules` state file.
- The upgraded websocket transport path uses private SDK internals and was not exercised against a live installed runtime here.
- No runtime capture was available for an account holding the widened tokenless spot-balance variant.

patches/@nktkas+rews+2.0.2.patch
sidmorizon, line 1: 🟠 P1: Crash protection for the trading data connection was deleted without a replacement

Reported at `patches/@nktkas+rews+2.0.2.patch:1-4` (outside the diff of this PR, so this is a file-level comment)

Severity: `severe`

The safety wrapper that stopped the live market-data connection from crashing the mobile app when a connection attempt times out or drops is deleted (`patches/@nktkas+rews+2.0.2.patch` removal) while the upgraded connection library ships no equivalent guard, so mobile users can hit a hard app crash again.
Impact: iOS/Android users may experience the app terminating while the Perp market data connection is timing out or reconnecting.

<details>
<summary>Why the deleted patch mattered and what replaces it</summary>

The removed patch wrapped every `dispatchEvent(new CloseEvent(...))` call inside `ReconnectingWebSocket` (both the outward `close` re-dispatch and the two synthetic close dispatches in `close()` and `createSocketWithTimeout`) in `try/catch`. Its own comments document the concrete production failure it fixed: on React Native the polyfilled `CloseEvent` fails the internal `EventTarget`'s `instanceof Event` check, the throw escapes the 5s connection-timeout `setTimeout` callback, and RN's `RuntimeScheduler` treats timer task errors as fatal → SIGABRT (Sentry `REACT-NATIVE-4AX`, 357 events on 6.2.0).

The PR bumps `@nktkas/hyperliquid` to 0.33.3, which pulls `@nktkas/rews@^4.1.0` (see `yarn.lock`), and deletes the 2.0.2 patch. No `patches/@nktkas+rews+4.*.patch` is added, and nothing in `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidSubscription.ts` re-introduces the guard at the consumer level. Unless rews 4.1.0 upstream removed or hardened those synthetic `CloseEvent` dispatches, the native crash path is reopened. This needs to be verified against the installed rews 4.1.0 sources, and a regenerated patch added if the dispatch pattern is unchanged.

</details>

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidSubscription.ts
```diff
@@ -1724,35 +1727,50 @@ export default class ServiceHyperliquidSubscription extends ServiceBase {
...
+          // Closing first lets the socket release reconnect timers before our
+          // listeners are detached.
           defaultLogger.perp.hyperliquid.subscriptionTransportDispose({
             clientId,
           });
           try {
-            // Close socket first so rews's internal close listener fires and
-            // clears its connection-timeout setTimeout.
-            transport.socket.close();
+            transport.close();
```
sidmorizon, line 1775: **Errors while shutting down the market-data connection are no longer caught**

Severity: `non-severe`

The shutdown of the live market-data connection is started without waiting for it (`transport.close()` at `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidSubscription.ts:1773`) inside a block that only catches immediate failures, so a failure during shutdown escapes as an unhandled error instead of being logged.
Impact: Closing or switching the Perp connection can surface an unhandled rejection warning (and on some platforms a crash report) instead of a logged, contained error.

<details>

<summary>Async close replaces the previously synchronous socket close</summary>

The old code called `transport.socket.close()`, which is synchronous, so the surrounding `try/catch` genuinely covered it. The new code calls the SDK transport's `close()`, which is asynchronous (`WebSocketTransport` also exposes `[Symbol.asyncDispose]` built on it). A rejected promise from `close()` is not observable by the `catch` block, and the repository convention in `.cursor/rules/coding-patterns.mdc` requires promises to be awaited or explicitly marked with `void`. Awaiting it inside the existing `try` both restores error containment and preserves the documented "close before detaching listeners" ordering.

</details>

Suggested fix:

```suggestion
            await transport.close();
```

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
```diff
@@ -494,6 +491,7 @@ function PerpTradingForm({
   const intl = useIntl();
   const actions = useHyperliquidActions();
   const [activeAsset] = usePerpsActiveAssetAtom();
+  const [activeAssetCtx] = usePerpsActiveAssetCtxAtom();
```
sidmorizon: **Trading panel now redraws on every price tick even when the price is unused**

Severity: `non-severe`

The whole live price snapshot object is read directly (`usePerpsActiveAssetCtxAtom()` at `packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:494`) instead of the narrow value-only readers the screen uses elsewhere, so the trading form and its buttons re-render on every incoming market tick, including in order modes that never use that price.
Impact: The Perp trading panel becomes noticeably busier/janky on low-end devices because it repaints continuously while the market streams.

<details>

<summary>Bypasses the existing display/live gating built for exactly this</summary>

`packages/kit-bg/src/states/jotai/atoms/perps.ts:704-772` intentionally exposes narrow derived atoms — `perpsActiveAssetCtxReadyAtom` (boolean), `perpsActiveAssetCtxMidPriceAtom`, and `usePerpsActiveAssetCtxMidPriceBySource(source)` with `live` / `display` / `disabled` variants — so components never subscribe to the full ctx object, and `perpsActiveAssetCtxDisplayAtom` is a throttled mirror used when the panel is not in live mode. `PerpTradingForm` already computes `shouldUseLiveTradingPrice` / `tradingPriceSource` and passes it to `useTradingPrice` for this reason.

The new `usePerpsActiveAssetCtxAtom()` calls in `PerpTradingForm` (line 494), `TradingButtonGroup` (`packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:446`), `useTradingCalculationsForSide` (`packages/kit/src/views/Perp/hooks/useTradingCalculationsForSide.ts:31`) and `useOrderConfirm` (`packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:78`) subscribe to the raw live atom unconditionally, so every ACTIVE_ASSET_CTX frame re-renders all of them regardless of order mode or live/display gating. A derived `perpsActiveAssetCtxMarkPriceAtom` (mirroring the existing mid-price-by-source helper) would keep the TWAP mark-price behavior while preserving the gating.

</details>

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
```diff
@@ -928,7 +897,13 @@ function PerpTradingForm({
...
   const [, referencePriceString] = useMemo(() => {
     let price = new BigNumber(0);
-    if (formData.orderMode === 'trigger' && formData.triggerOrderType) {
+    if (formData.orderMode === 'twap') {
+      price = getTwapTriggerReferencePrice({
+        isSpot,
+        midPrice: midPriceBN,
+        markPrice: activeAssetCtx?.ctx?.markPrice,
+      });
+    } else if (formData.orderMode === 'trigger' && formData.triggerOrderType) {
```
sidmorizon: **Time-weighted orders show a zero reference price and refuse to submit while market data is still loading**

Severity: `non-severe`

The TWAP reference price is taken only from the market context feed (`getTwapTriggerReferencePrice` at `packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:901-906`) with no fallback to the account's own mark price, so during the brief window after switching markets the size box shows a zero reference and the order button rejects the trade as "price unavailable".
Impact: Right after switching coins (or on a cold start) a user in TWAP mode sees wrong size/USD conversions and cannot place the order until the market feed arrives.

<details>
<summary>Three different fallbacks for the same TWAP reference price</summary>

`getTwapTriggerReferencePrice` (`packages/shared/src/utils/hyperliquidTwapUtils.ts:90-103`) returns `new BigNumber(markPrice ?? '')` for perps, i.e. `NaN` when the caller passes no mark price. Callers disagree on what to pass:

- `packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx:901-906` passes only `activeAssetCtx?.ctx?.markPrice`. `NaN` then flows into `formatPriceToSignificantDigits`, which returns `'0'` (`packages/shared/src/utils/perpsUtils.ts:742-751`), so `referencePriceString` given to `SizeInput` becomes `'0'`.
- `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx:467-475` and `packages/kit/src/views/Perp/hooks/useOrderConfirm.ts:83-91` also pass only the ctx mark price, and both treat a non-finite result as `marketDataUnavailable`, blocking submission.
- `packages/kit/src/views/Perp/hooks/useTradingCalculationsForSide.ts:173-186` *does* fall back to `activeAssetData?.markPx` (and the new test asserts this), and `packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts:2905-2910` falls back to `env.markPrice` (the mid price).

So while `perpsActiveAssetCtxAtom` is empty but `perpsActiveAssetDataAtom` is populated, size math uses the account mark while the reference display is `0` and submission is blocked — the "consistent fallback" described in the PR is only implemented in one of the four call sites.

</details>

originalix (commented):
@Minnzen

Reviewed commit `96405115bf62`.

- **P1 · TWAP confirmation still shows pre-normalized trigger/stop prices**
- **P1 · TWAP submit guards still ignore the new mark-price fallback**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

patches/@nktkas+rews+2.0.2.patch
Minnzen: The SIGABRT protection was not dropped — it moved upstream. rews 4.1.0 restructured the close path so the old failure mode no longer exists:

1. Outward `close` events are dispatched from the async `_runLoop()` (`closeInit = await this._awaitSocketLifecycle(...)`), not from inside the native socket's listener callback, so a throwing dispatch can no longer land in RN's RuntimeScheduler (the SIGABRT path our 2.0.2 patch guarded).
2. The loop dispatches a freshly constructed event with an explicit upstream comment: `HACK (nktkas/rews#7): Dispatch a freshly constructed event on this, never the socket's own — React Native's strict EventTarget throws on a foreign event`.
3. The whole loop body is wrapped in `try/catch → this._terminate("UNKNOWN_ERROR", error)`.
4. rews 4.1.0 ships its own `CloseEvent`/`MessageEvent`/`DOMException` fallbacks for RN (`const CloseEvent_ = globalThis.CloseEvent || class CloseEvent extends Event`), and our Hermes polyfill in `packages/shared/src/polyfills/polyfillsPlatform.js` (which returns a real `Event` instance) is still in place, so the dispatched event always passes the `instanceof Event` check.

Evidence: `node_modules/@nktkas/rews/esm/mod.js` `_runLoop()` and the polyfills section at the bottom of the same file.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidSubscription.ts
Minnzen: `WebSocketTransport.close()` in 0.33.3 is synchronous, so the existing `try/catch` does contain failures. From the installed package:

- `node_modules/@nktkas/hyperliquid/esm/transport/websocket/mod.d.ts:192` — `close(): void;`
- Implementation (`mod.js`): `close() { this.socket.close(); }` with an upstream comment: "socket.close() terminates synchronously on every path: when it returns, the termination signal is already aborted — there is nothing to await", plus an explanation of why awaiting the final close event would deadlock when called from inside a close listener.

`[Symbol.asyncDispose]` is a separate API on the transport; `close()` itself returns nothing awaitable, and `await transport.close()` would be an await on `void`. Keeping the current synchronous call as-is.

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
Minnzen: Fixed in 2bc551da51. `PerpTradingForm` (and `TradingButtonGroup` / `useOrderConfirm`) no longer subscribe to the whole `perpsActiveAssetCtxAtom` snapshot. The TWAP reference price now comes from a shared `useTwapReferencePrice()` hook backed by a new `perpsActiveAssetCtxMarkPriceAtom` computed selector (same pattern as the existing `perpsActiveAssetCtxMidPriceAtom`), so these components re-render only when the mark-price string actually changes instead of on every ctx message.

packages/kit/src/views/Perp/components/TradingPanel/panels/PerpTradingForm.tsx
Minnzen: Fixed in 2bc551da51. The TWAP reference price is now derived once in `useTwapReferencePrice()` as `activeAssetCtx.ctx.markPrice ?? activeAssetData.markPx` (mid price for spot) and reused by `PerpTradingForm`, `TradingButtonGroup`, and `useOrderConfirm`; the `submitOrder` action falls back through the same chain before `env.markPrice`. While `activeAssetCtx` lags during cold start or reconnect, the form now shows the same webData2 mark the calculations already use, and CTA validation/submit no longer treat the reference price as missing. Covered by `useTwapReferencePrice.test.ts`.

Minnzen:
已处理 [automated review](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845) 的两个 P1，随 commit `2bc551da51` 推送：

- P1（确认页价格未归一化）：`OrderConfirmModal` 的 TWAP trigger/stop 预览改为复用提交路径同一个 `formatTwapPriceForOrder()` 的输出，用户确认的价格即实际发出的 wire 价格。
- P1（提交守卫缺 mark-price fallback）：新增 `useTwapReferencePrice()` 共享 hook，统一按 `activeAssetCtx.ctx.markPrice ?? activeAssetData.markPx`（spot 用 mid）推导参考价，`PerpTradingForm`、`TradingButtonGroup`、`useOrderConfirm` 三处复用；`submitOrder` action 同链路 fallback。冷启动/重连期间表单、CTA 校验与提交不再互相矛盾。

顺带：hook 底层用新增的 `perpsActiveAssetCtxMarkPriceAtom` 计算 selector（对齐现有 midPrice selector 模式），三个组件不再订阅整个 ctx 快照，缓解每 tick 重绘问题。已回复并 resolve sidmorizon 的 4 条线程（其中 rews patch 与 transport.close 两条经查证 SDK 源码后说明不需要改动）。

验证：`hyperliquidTwapUtils` / `useTradingCalculationsForSide` / `useTwapReferencePrice`（新增 4 例）/ ServiceHyperLiquid 共 15 suites 105 tests 通过；`yarn agent:check --profile commit` 通过。

originalix (commented):
@Minnzen

Reviewed commit `2bc551da512a`.

- **P1 · Outcome spot balances disappear from UI and cached totals**
- **P2 · Plain TWAP orders now always send the optional `details` wrapper**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

Minnzen:
针对 [automated review](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845) 最新一轮（reviewed `2bc551da512a`），随 commit `b6becfd4a0` 推送：

**P2（plain TWAP 恒发 `details` wrapper）— 已修复。** `details` 改为仅在配置了 trigger/stop 时构造，否则整个省略。已用 SDK 的 `parse` + `canonicalize` 实测：省略 key 时 canonical action 不含 `details`（与 0.33 之前生产验证过的 wire 形状一致）；此前的 `{t: null, s: null}` 确实会原样上 wire。注意不能传 `details: undefined`（canonicalize 会保留该 key），因此用条件展开。

**P1（outcome spot 余额被过滤）— 说明后维持现状，建议 follow-up。** 分析如下：
1. Portfolio 聚合路径（`hyperliquidPortfolioUtils.ts` `aggregate` 内）已经对被过滤且 total > 0 的 outcome 条目置 `degraded = true`，并非静默丢弃；`spotHasPositiveBalance` 也未过滤 outcome 条目。
2. App 全链路没有 outcome market 支持：无订阅、无价格源（outcome 代币不在 spot universe/allMids 内）。若把它们放进 `calculateSpotBalancesTotalUsd`，会永久落入 `missingPriceCoins` 分支，导致 spot 总额长期卡在 fallback/partial 状态——比排除更糟。
3. "degraded-but-visible" 需要决定 `+N`/`oN` 条目的展示文案与估值语义（`entryNtl` 是入场名义价值，不是现值，结算后可能偏差极大），且 `ISpotBalance` 契约（`token: number`）被 mappings/转账等下游依赖——属于 feature 级改动，不适合热更分支。
建议为 outcome-token 展示单开 ticket 在常规版本处理。

另外修复了 CI lint 的两个格式错误（`useTwapReferencePrice.ts` import 分组）。定向测试：ServiceHyperLiquid 12 suites / 86 tests 通过，`yarn agent:check --profile commit` 通过。

originalix (commented):
@Minnzen

Reviewed commit `b6becfd4a04a`.

- **P1 · Outcome spot balances disappear from live state and cached totals**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

Minnzen:
回应 [automated review](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845) 对 `b6becfd4a04a` 的复审（仅剩 P1 outcome 余额一项）：

产品结论不变——热更分支不引入 outcome 代币展示（无价格源、无 UI 语义、`ISpotBalance` 契约下游依赖广），完整理由见[上一条回复](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5284414100)。

针对本轮指出的"WS 路径完全静默"，随 commit 已推送最小改善：`updateSpotBalances` 在过滤掉正余额 outcome 条目时输出 warn 日志（与该文件现有日志风格一致），使遗漏在日志/排障中可见。Portfolio 快照路径维持既有 `degraded = true` 标记。outcome 代币的可见化处理建议单开 ticket 在常规版本评估。

originalix (commented):
@Minnzen

Reviewed commit `764fbe781ba4`.

- **P1 · Outcome spot balances disappear from live state and cached totals**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

originalix (commented):
@Minnzen

Reviewed commit `b11d55b4c6bf`.

- **P1 · Outcome spot balances disappear from live state and cached totals**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

packages/kit/src/views/Perp/hooks/useTradingCalculationsForSide.ts
```diff
@@ -168,11 +170,30 @@ export function useTradingCalculationsForSide(side: 'long' | 'short') {
...
-    const markPx = isSpot
-      ? effectiveSpotPriceBN.toFixed()
-      : activeAssetData?.markPx;
+    let markPx = activeAssetData?.markPx;
+    if (isSpot) {
+      markPx = effectiveSpotPriceBN.toFixed();
+    } else if (formData.orderMode === 'twap') {
+      markPx = activeAssetCtx?.ctx?.markPrice ?? activeAssetData?.markPx;
+    }
     return new BigNumber(markPx ?? 0);
```
sidmorizon, line 179: **TWAP order panel wrongly reports insufficient funds while account data is still loading**

Severity: `non-severe`

For TWAP orders the panel now derives its price from the market-wide feed (`activeAssetCtx?.ctx?.markPrice` at `packages/kit/src/views/Perp/hooks/useTradingCalculationsForSide.ts:177`) while the spendable amount still comes only from the per-account feed, so during the window where market data has arrived but account data has not, the panel treats the account as having zero funds.
Impact: Users switching markets or opening the TWAP tab can briefly be blocked with a false "insufficient margin" error even though they have funds.

<details>
<summary>Why the guard flips to true only in TWAP mode</summary>

`isNoEnoughMargin` short-circuits with `if (markPxBN.gt(0) && availableMarginBN.lte(0)) return true;` (`packages/kit/src/views/Perp/hooks/useTradingCalculationsForSide.ts:375-377`). `availableMarginBN` is computed from `maxTradeSzBN`, which reads `activeAssetData?.maxTradeSzs` (account-scoped, `:112-168`) and is `0` while `perpsActiveAssetDataAtom` is `undefined`.

Before this change `markPxBN` was always `activeAssetData?.markPx`, so it was also `0` in that window and the guard stayed `false`. Now, in `orderMode === 'twap'`, `markPxBN` falls back to the market-wide `activeAssetCtx` mark price, which is populated independently of the account subscription (`changeActiveAsset` clears both atoms via `clearActiveAssetData` at `packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts:2296-2298`, and the ctx atom re-fills first because it needs no user).

Consequence: `useTradingCalculationsForSide` returns `isNoEnoughMargin: true`, and `validateOrderPanelState` in `packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx` then shows the no-enough-margin toast and aborts submission with `'insufficientMargin'`.

</details>

Minnzen: Fixed in a32f28fc4f. `isNoEnoughMargin` now returns `false` for perps while `activeAssetData` is still undefined, so neither the `availableMarginBN.lte(0)` short-circuit nor the final `requiredMargin.gt(availableMarginBN)` comparison (which had the same race for manual sizes) can conclude "insufficient margin" from the market-wide ctx mark price before the account feed has (re)loaded. Genuine zero-margin accounts still report correctly once `activeAssetData` arrives with zero `maxTradeSzs`. Covered by a new case in `useTradingCalculationsForSide.test.ts`.

originalix (commented):
@Minnzen

Reviewed commit `a32f28fc4f50`.

- **P1 · Outcome spot balances disappear from live state and cached totals**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12849#issuecomment-5275852845)

sidmorizon:
@cursoragent review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/12849>)

sidmorizon:
Claude review session: https://claude.ai/code/session_01VeHWj5cuHuaafWnpjM2412

packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
```diff
@@ -2952,19 +3023,14 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
...
-            );
-            const averageSliceNotional =
-              totalNotional.dividedBy(estimatedSlices);
             if (
-              !averageSliceNotional.isFinite() ||
-              averageSliceNotional.lt(TWAP_MIN_ORDER_NOTIONAL)
+              !isTwapTotalNotionalValid({
+                size: resolvedSizeBN,
+                price: markPriceBN,
+              })
```
sidmorizon, line 3031: 🟡 P2: Rounded TWAP bypasses minimum notional

Severity: `non-severe`

`isTwapTotalNotionalValid` checks the raw size before lot-size truncation. Borderline orders fall below 100 USDC on the wire and fail submission.

Minnzen: 已在 `2615d56c67` 修复。`isTwapTotalNotionalValid` 现在先用同一个 `formatHlSize(size, szDecimals)` 截断数量，再判断 100 USDC 下限；按钮、确认页、submit action 和后台 `placeTwapOrder` 均使用该校验，后台在调用 SDK 前再次拦截不合格数量。

回归覆盖 `0.054246 × 1843.5`：原始金额超过 100，但按 4 位数量精度发出的 `0.0542` 只有 `99.9177`，现在会被拒绝且不会调用 SDK；截断后恰好 100 和高于 100 的订单正常通过，plain TWAP 仍省略 `details`。

验证：11 个相关 Jest suites / 148 tests 通过，`yarn agent:check --profile commit`（含 lint、TypeScript）全部通过。

PR #12842, opened 2026-08-12
fix: prevent mobile market tabs from jumping OK-55394

limichange:
[OK-55394](https://onekeyhq.atlassian.net/browse/OK-55394)

----------

## Summary

- disable animated native pager transitions when pressing Market Home tabs
- keep tab presses aligned with the requested destination instead of emitting intermediate page selections

## Root cause

The native pager used animated `setPage` transitions for tab presses. When jumping across multiple positions, Android emitted callbacks for intermediate pages, which briefly updated the active Market tab before the pager reached the requested tab.

## Impact

Market Home tab presses on iOS and Android now switch directly to the selected tab. Web and desktop behavior is unchanged.

## Validation

- verified on an Android emulator by switching repeatedly between Favorites, Trending, Stocks, and Perps; each transition emitted only the requested target and rendered its content
- `yarn jest packages/kit/src/views/Market/MarketHomeV2/layouts/hooks/useSyncedMarketTab.test.ts packages/kit/src/views/Market/MarketHomeV2/layouts/marketTabSelectionGuards.test.ts packages/kit/src/views/Market/MarketHomeV2/layouts/marketTabChangeGuards.test.ts --runInBand` (21 tests passed)
- `yarn agent:check --profile commit`

Issue: OK-55394

[OK-55394]: https://onekeyhq.atlassian.net/browse/OK-55394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12850, opened 2026-08-12
fix: restore swap history order id and surface required recipient entry (OK-59978, OK-58326)

haicongliao:
[OK-59978](https://onekeyhq.atlassian.net/browse/OK-59978) [OK-58326](https://onekeyhq.atlassian.net/browse/OK-58326)

----------

## OK-59978 — Swap history order id

- Restore the Order ID row that was removed in #12487. It now renders only when the provider returns `orderSupportUrl` (e.g. CoW), so providers without a third-party order explorer stay hidden as OK-57251 intended.
- Use the provider order id kept in `ctx` (the CoW order uid) instead of the internal service order id, so the explorer link actually resolves for Stock orders — previously the link carried an internal UUID and `explorer.cow.fi` returned "No results found".
- Shorten the displayed id to a single line while copying and linking the full value (`InfoItem` gained an optional `copyContent`).
- Show the explorer link next to the tx hash for same-chain swaps, matching the cross-chain layout.

## OK-58326 — Swap recipient entry

- Surface the recipient entry row when the quote requires a manually entered recipient (e.g. a single-network private-key wallet doing a cross-chain swap), even while the custom recipient setting is off. The action button deep-link from #12685 is kept as a second entry.
- Fix `SwapSmoothReveal` never expanding on native: a normal-flow child of the height-0 animated wrapper measures 0 under Yoga, so the measured height stayed 0 forever and the row was permanently hidden on iOS/Android. Content is now measured with absolute positioning. This also unhides the stock trade alert and the cost-savings reveal on native.
- Hold the "recipient required" verdict across quote refresh cycles (the same settled-hold pattern as OK-58977) so the row no longer collapses and re-expands on every quote round.

## Verification

- 52 unit tests pass (`swapHistoryIdentity`, `swapHistoryTransactionIds`, `useSwapAccount.utils`), including new cases for the provider-order-id fallback and the recipient entry conditions.
- Verified on the iOS simulator (recipient row expands and stays put across quote cycles) and on desktop (a real CoW Stock order shows `订单编号 0x0d99606c…6a85790c` under `渠道商 CoW Swap`).

## Note on CI type check

`yarn tsc:staged` reports one error in `packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx:873` (`tabPressAnimationEnabled` is not declared on the tab container props). This is pre-existing on `x` — introduced by #12842 — and unrelated to the files in this PR. Reproduced on a clean `x` checkout with only a whitespace change to trigger the check.

[OK-59978]: https://onekeyhq.atlassian.net/browse/OK-59978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-58326]: https://onekeyhq.atlassian.net/browse/OK-58326?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Swap/pages/modal/SwapHistoryDetailModal.tsx
```diff
@@ -1648,6 +1661,29 @@ const SwapHistoryDetailModal = () => {
...
+                label={intl.formatMessage({
+                  id: ETranslations.Limit_order_history_order_id,
+                })}
+                renderContent={accountUtils.shortenAddress({
+                  address: providerOrderId,
+                  // Roughly one full line in the detail modal, matching the
+                  // width of a wrapped 66-char tx hash line.
+                  leadingLength: 24,
+                  trailingLength: 20,
+                })}
```
sidmorizon: **Shortened order number can show duplicated characters for medium-length ids**

Severity: `non-severe`

The order number is abbreviated with a first-24/last-20 split (`accountUtils.shortenAddress` at `packages/kit/src/views/Swap/pages/modal/SwapHistoryDetailModal.tsx:1669-1675`) without raising the minimum length that makes shortening safe, so any order number between 15 and 43 characters is rendered with its middle characters repeated.
Impact: Users can see a garbled, longer-than-original order number in the swap history detail for providers whose order numbers are short.

<details>
<summary>How the default minLength lets the two slices overlap</summary>

`accountUtils.shortenAddress` (`packages/shared/src/utils/accountUtils.ts:273-295`) returns the raw value only when `address.length <= minLength`, and `minLength` defaults to `14`. It then builds `address.slice(0, leadingLength) + '...' + address.slice(-trailingLength)`. With `leadingLength: 24` and `trailingLength: 20`, any id whose length is between 15 and 43 makes the two slices overlap, so characters appear twice (e.g. a 20-char order id renders as the full id, `...`, then the full id again). Long ids such as a CoW order uid (114 chars) or a 1inch fusion hash (66 chars) are fine; shorter provider order ids taken from `item.txInfo.orderId` in `packages/kit/src/views/Swap/utils/swapHistoryIdentity.ts:24-32` are not. Passing `minLength` of at least `leadingLength + trailingLength` avoids the overlap.

</details>

Suggested fix:

```ts
                renderContent={accountUtils.shortenAddress({
                  address: providerOrderId,
                  // Roughly one full line in the detail modal, matching the
                  // width of a wrapped 66-char tx hash line.
                  leadingLength: 24,
                  trailingLength: 20,
                  // Only abbreviate when the two slices cannot overlap.
                  minLength: 44,
                })}
```

zhaono1:
## Review 结论

**需要修改后复审，7.4/10，当前不建议合入。**

已按当前 head `37f9e3dc63` 复核；相对最新 `x` 无 base 漂移。CI 为 **17 pass / 0 failed / 0 pending**，`git diff --check` 通过，但 3 个 review threads 仍未解决，其中 2 个是已确认的正确性问题，另 1 个是成立的跨平台边界问题。

### 合入前需要处理

1. **[P2] 收款方必填状态会跨报价作用域残留**（[thread](https://github.com/OneKeyHQ/app-monorepo/pull/12850#discussion_r3767377914)）  
   该状态属于 iOS/Android `main` UI runtime 的单份 JS heap；没有 bg readiness 或共享 native resource 参与。旧报价把 ref 置为 `true` 后，切到 native Limit / 切换账户或 token 时，quote 被清空但 ref 没有按作用域重置，入口可能继续错误显示。建议按 swap type + account + from/to token/network 对状态做 scope/reset，同时保留同一作用域内 quote refresh 防闪烁，并补一个真实生命周期回归测试；当前 pure-function 测试覆盖不到这个问题。

2. **[P2] 中等长度 Order ID 的缩写会重复字符**（[thread](https://github.com/OneKeyHQ/app-monorepo/pull/12850#discussion_r3767361908)）  
   `leadingLength=24`、`trailingLength=20`，但默认 `minLength=14`；15–43 位 ID 会发生切片重叠。请把阈值至少设为 44（或改成专用 formatter），并补边界用例。

3. **[P2] 原生 Yoga 修复被无条件施加到 shared component**（[thread](https://github.com/OneKeyHQ/app-monorepo/pull/12850#discussion_r3767377896)）  
   原评论的 P1 严重度偏高，但技术边界成立：问题针对 iOS/Android `main` runtime，`position: absolute` 不应默认影响 desktop/web 以及 `SwapSmoothReveal` 的其他调用方。请用 platform-specific file / `Platform.select` 限制到 native，或补齐全部 shared callers 的 desktop/web/native 运行时证据后再关闭。

### 历史问题与契约核对

- **OK-57251 → OK-59978：方向正确。** 之前移除的是 generic Order ID；本 PR 仅在存在 `orderSupportUrl` 时恢复 provider order entry，且优先使用 `cowSwapOrderId`，符合 CoW pending order 没有链上 tx hash 的场景。服务端 `server-service-swap` 的 provider contract 也确认是 `orderSupportUrl` prefix + CoW provider order id。
- 关联 Slack 线程明确拆成两条链路：本 PR 处理 explorer link；`/state-tx` 的 `txId` 也应使用 `cowSwapOrderId`，但由另一条改动负责。不要把后者算作本 PR 已交付。
- **OK-60100 相邻回归风险：**“入口显示”不等于编辑链路通过。选择外部/非本账户 recipient 后再次打开编辑器，仍需有 active-account editor context；该修复不在本 diff 内，合入/回归时需要确认依赖已进入 `x`。

### 评分与验证边界

| 维度 | 分数 | 结论 |
| --- | ---: | --- |
| Security | 9.0 | 未发现敏感日志、鉴权或交易校验回退 |
| Code Quality | 6.5 | 两个明确状态/格式边界缺口 |
| Architecture | 7.0 | provider identity 分层正确，但 native 修复越过平台边界 |
| Completeness | 6.5 | CI 绿；缺 ref 生命周期、缩写边界和完整跨平台运行时证明 |

本轮核对了 Jira 描述/附件、关联 Slack 讨论、服务端字段来源、当前 diff、review threads 与远端 CI。截图能证明 recipient entry 在 iOS/desktop 出现，但**元素存在本身不算完整通过**；修复后至少需要验证：作用域切换不残留、外部 recipient 可重复编辑、CoW pending order 可打开正确 explorer，以及所有 `SwapSmoothReveal` 调用方在 desktop/web/native 的布局。

packages/kit/src/views/Swap/pages/modal/SwapHistoryDetailModal.tsx
haicongliao: Confirmed and fixed in 1bb0d474ca — verified a 20-char id rendered as 43 chars (whole id, `...`, whole id again), and a 36-char uuid as 47. Extracted `shortenSwapOrderId` with `minLength = leadingLength + trailingLength` (44) so ids up to 44 chars render in full and only longer ones abbreviate; added unit tests covering 15/20/36/43/44/45/114.

packages/kit/src/views/Swap/pages/components/SwapSmoothReveal.tsx
```diff
@@ -63,7 +63,15 @@ export function SwapSmoothReveal({
...
         >
+          {/* Absolute positioning frees the content from the animated
+              wrapper's height:0 constraint: native Yoga measures a normal-flow
+              child of a fixed-height parent as 0, which would keep the
+              measured height stuck at 0 forever. (OK-58326) */}
           <Stack
+            position="absolute"
+            left={0}
+            right={0}
+            top={0}
```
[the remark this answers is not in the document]
haicongliao: 已修复（1bb0d474ca）。绝对定位现在仅在原生生效：新增 `MEASURED_CONTENT_LAYOUT_PROPS`，由 `platformEnv.isNative` 决定是否应用，web/desktop 保持原有的正常流布局与固有尺寸计算，避免影响其余 SwapSmoothReveal 调用方（费用节省、股票交易告警）。

packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx
```diff
@@ -230,20 +231,44 @@ const SwapActionsState = ({
...
+    currentQuoteRes?.toAmount &&
+    !swapToAddressInfo.address &&
+    !swapActionState.noConnectWallet,
+  );
+  const settledRecipientRequiredRef = useRef(recipientRequiredNow);
+  if (
+    (currentQuoteRes || isQuoteSettledWithoutResult) &&
+    swapToAddressInfo.isAddressInfoReady
+  ) {
+    settledRecipientRequiredRef.current = recipientRequiredNow;
```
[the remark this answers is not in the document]
haicongliao: 已确认并修复（1bb0d474ca）。核对 `swapTypeSwitchAction`（actions.ts:3528-3532）确实会清空报价列表并将 `quoteEventCompleted` 置为 false，两个 adopt 条件都不成立，旧判定会残留到新 tab。现在把判定按作用域 key（tab / token pair / source account）重新键控，抽出纯函数 `resolveSettledSwapRecipientRequired`：作用域变化即重置为 false，等该作用域自己的报价结算后再采纳；新增 5 个单测覆盖切 tab、换币对、换账户三种情形。

haicongliao:
Review 基于 `37f9e3dc63` 复核，三个问题已在 `1bb0d474ca` 修复并 resolve，thread 状态可能有时间差。当前 head 为 `d7516e6791`。

**1. [P2] 收款方必填状态跨作用域残留** — 已按 swap type + source account + from/to token network & contract 组成 scope key，作用域变化即重置为 `false`，同作用域内保留 quote refresh 的防闪烁保持。

**2. [P2] 中等长度 Order ID 缩写重复字符** — `minLength` 设为 `44`（= leading 24 + trailing 20），15–43 位 ID 现在整串显示；已补 14/15/43/44/45/66/114 的边界用例。

**3. [P2] 原生 Yoga 修复施加到 shared component** — 绝对定位测量改为仅 `platformEnv.isNative` 生效，desktop/web 及 `SwapSmoothReveal` 其他调用方（cost savings、stock alert）回到原有正常流布局。

**关于"pure-function 测试覆盖不到"** — 这一条成立，本轮已补。把 scope key 构造与 ref 保持一起抽成 `useSettledSwapRecipientRequired`，用 `renderHook` 覆盖真实生命周期：报价刷新周期内保持、settled 结果变化时才跳变、切 Limit / 换 token / 换账户时重置、新作用域自己 settle 后重新建立。

这组测试确实抓到了纯函数测试覆盖不到的缺陷：hook 的 ref 初始值直接取了当轮 `recipientRequiredNow`，绕过 `isAddressInfoReady` 门槛，首次挂载且目标地址尚未解析时会误判。已改为中性初始值，首帧同样需要通过 settled + address-ready 门槛。

相关测试 68 passed。`tsc` 仍报 `MarketHomeV2/layouts/MobileLayout.native.tsx:873`，为 `x` 上 #12842 引入的既有问题，与本 PR 文件无关（在干净 `x` 上加空白改动触发检查可复现）。

zhaono1 (dismissed):

haicongliao:
已 rebase 到最新 `x`（`b6f789ec64`），冲突解决，当前 head `f4e5a7f40c`，10 个提交保持线性。

**唯一冲突点值得说明：`SwapSmoothReveal.tsx`**

#12862 独立修复了同一个原生测量问题，做法与本 PR 早期提交一致——对测量节点无条件加绝对定位。本 PR 在 review 反馈（[P1 thread](https://github.com/OneKeyHQ/app-monorepo/pull/12850#discussion_r3767377896)）后已改为仅 `platformEnv.isNative` 生效，因此冲突解决时保留了平台受限版本，而非回退成无条件版本：

- 原生：行为与 #12862 等价，绝对定位解除 Yoga 对 height-0 父节点子元素的测量约束；
- desktop/web：回到正常流，不改变 `SwapSmoothReveal` 其余调用方（cost savings、stock alert）的固有尺寸计算。

这意味着 diff 中会看到本 PR 调整了 #12862 刚合入的这几行，属于按 review 要求收窄平台边界，不是回退其修复。JSDoc 措辞也同步对齐为「native 才脱离文档流」。

其余 upstream 改动无重叠：#12862 触及的 `actions.ts` / `atoms.ts` / Pro 系列组件与本 PR 文件不相交，本 PR 依赖的两个前提（切 tab 清空 quote 与重置 `quoteEventCompleted`、`swapProviderSupportReceiveAddressAtom`）在新 `x` 上均未变更，已复核。

相关测试 68 passed。`tsc` 仍报 `MarketHomeV2/layouts/MobileLayout.native.tsx:873`，在最新 `x` 上依旧存在且与本 PR 文件无关。

packages/kit/src/views/Swap/hooks/useSwapAccount.utils.ts
```diff
@@ -118,6 +121,107 @@ export function getSwapRecipientActionState({
...
+export function resolveSettledSwapRecipientRequired({
+  previous,
+  scopeKey,
+  quoteSettled,
+  isAddressInfoReady,
+  recipientRequiredNow,
+}: IResolveSettledSwapRecipientRequiredParams): ISettledSwapRecipientRequired {
+  if (previous.scopeKey !== scopeKey) {
+    return { scopeKey, value: false };
+  }
```
sidmorizon: **Recipient entry can stay hidden after the trading pair or account changes**

Severity: `non-severe`

The "a recipient must be entered" decision is reset to "not needed" (`return { scopeKey, value: false }` at `packages/kit/src/views/Swap/hooks/useSwapAccount.utils.ts:217`) whenever the pair/account/tab identity changes, even when the current price quote has already settled and clearly needs one, so the recipient row can stay hidden until something else forces the screen to update.
Impact: In that case the user sees no place to enter the required recipient address while the confirm button stays disabled.

<details>
<summary>Why the reset can persist: verdict is stored in a ref, so no re-render is scheduled</summary>

`resolveSettledSwapRecipientRequired` returns a hard `false` on any scope-key change, discarding the freshly available inputs (`quoteSettled`, `isAddressInfoReady`, `recipientRequiredNow`) that are already passed in the same call. The result is stored in a ref inside `packages/kit/src/views/Swap/hooks/useSettledSwapRecipientRequired.ts:48-56`, so unlike a `setState`-based derived-state pattern nothing schedules a follow-up render. If the render in which the scope key changes is also the render in which the settled inputs are already true (e.g. `sourceAccountId` resolving from `undefined` to a real id at `packages/kit/src/views/Swap/pages/components/SwapActionsState.tsx:246` while a settled quote with `toAmount` and no resolved target address is present), the `false` verdict sticks until an unrelated state change triggers another render.

Computing the new scope's verdict from the same-render inputs is both correct and keeps all existing unit tests passing (they all pass `quoteSettled: false` on scope change).

</details>

Suggested fix:

```ts
  if (previous.scopeKey !== scopeKey) {
    return {
      scopeKey,
      value: quoteSettled && isAddressInfoReady ? recipientRequiredNow : false,
    };
  }
```

haicongliao: Fixed in a5d35b5e9c. Settled inputs are now evaluated before the scope-change branch, so a render that changes the scope key while quoteSettled + isAddressInfoReady are already true adopts the real verdict instead of a hard false — the ref-stored verdict no longer depends on an unrelated render to recover. Neutral reset still applies when the new scope has no settled quote, so a previous tab/pair verdict cannot leak. Added a resolver case for the scope-change-with-settled-inputs combination and a renderHook case for sourceAccountId resolving from undefined while a quote already needs a recipient.

originalix:

@haicongliao

Reviewed commit `46a89f24fcf0`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes two foreground Swap presentation boundaries. In the action area, it adds a scoped "recipient required" latch so the recipient row stays stable across quote refreshes and across tab or token transitions. In history detail, it separates provider-facing order identifiers from internal history identity, shortens long order ids for display, and adds explorer actions for transaction hashes and provider order ids. The scope stays inside `packages/kit` UI/hooks/tests and does not change signing, persistence, background ownership, or auth flows, but the new recipient latch still treats any retained selected quote as settled and the new order explorer action assumes `orderSupportUrl` is only a URL prefix.

**What needs attention:** Verify that recipient visibility only follows the active quote event after tab or pair switches, and confirm the exact `orderSupportUrl` contract before shipping the new Order ID explorer action.

### Issues to address

- **P1 · Recipient scope still adopts stale quotes**
  - **Impact:** The new recipient latch is supposed to stop verdicts leaking across tab, pair, and account changes, but it marks the scope as settled whenever `currentQuoteRes` is truthy. The quote-progress layer can intentionally keep a previous actionable quote selected during stale refresh or immediately after a tab switch, so the new scope can still adopt the previous quote's recipient verdict before its own quote settles.
  - **Suggested action:** Gate both `quoteSettled` and `recipientRequiredNow` on a quote that is confirmed to match the current input or active event instead of using raw `currentQuoteRes` truthiness.
- **P2 · Order explorer URL assumes `orderSupportUrl` is a prefix**
  - **Impact:** The new Order ID action always concatenates `providerOrderId` onto `swapInfo.orderSupportUrl`, but the same field is already treated elsewhere as a ready-to-open URL. If a provider returns a full order page here, the modal will generate an invalid link and the new explorer affordance will fail.
  - **Suggested action:** Make the history contract explicit: either persist a fully resolved provider order URL, or append the id only for providers whose API explicitly returns a prefix/template.

### Validation gaps

- Focused Jest files for the changed Swap hooks/utilities could not be rerun because `yarn jest ...` failed with `Couldn't find the node_modules state file - running an install might help (findPackageLocation)`.
- Existing PR comments were not fetched in this workspace, so inline comment suggestions are marked for human review rather than safe auto-posting.

haicongliao:
Both issues addressed in `89d052825c`.

**P1 — Recipient scope adopting stale quotes: confirmed and fixed.** Verified the premise: `selectSwapPreviousActionableQuote` in the quote-progress layer does intentionally keep the previous actionable quote selected while a new round is requesting, so raw `currentQuoteRes` truthiness was indeed a wrong settled signal. The pair-match gate now lives inside `useSettledSwapRecipientRequired` itself: the hook receives the raw selection atom and only treats it as settled for the scope when its `fromTokenInfo`/`toTokenInfo` match the currently selected pair (`equalTokenNoCaseSensitive`), and `recipientRequiredNow` is gated the same way. Account/tab identity stays covered by the scope key. Moving the gate into the hook (rather than the component) keeps the whole verdict chain covered by the lifecycle tests — added a case where the previous pair's quote is still selected on the renders right after a pair switch, both when the new pair has a target address and when it doesn't. Writing those tests also flushed out a mount-render hole (the ref initialized with an ungated verdict, bypassing the address-ready gate on first render); it now starts neutral and every adoption passes the resolver.

**P2 — `orderSupportUrl` contract: verified, and the concat is now guarded.** The field has genuinely carried two shapes over time: the founding #6504 code opened `swapInfo.orderSupportUrl` directly as a full URL, while today's backend sends a base to append the id to — confirmed live this week on a real CoW stock order, where the server returned `https://explorer.cow.fi/bnb/search/` (trailing slash, nothing else) and the appended provider uid resolved on explorer.cow.fi. (`ILimitOrder.orderSupportUrl` is a different field on the limit-order item and keeps its full-URL contract, untouched by this PR.) Rather than assuming the prefix shape forever, the modal now builds the link via `buildSwapHistoryOrderExplorerUrl`, which appends only when the URL does not already carry the order id — both historical shapes produce a working link, with unit tests for each.

79 focused tests pass; lint/tsc green in an isolated worktree install.

originalix (commented):
@haicongliao

Reviewed commit `89d052825c98`.

- **P2 · Pair matching still accepts stale quote events**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12850#issuecomment-5300357287)

haicongliao:
P2 (stale quote events) addressed in 6bfa6f6c31. Confirmed the premise: pair equality cannot distinguish a same-pair account switch, where the previous account's quote stays selected and its event stays registered as current until the new request is written. Adoption now requires an active-request proof computed from the same primitives the selection layer uses — isSwapQuoteFromCurrentEvent (event-id membership) AND isSwapQuoteRequestForCurrentInput (the request lock matching the current account/pair/amount, which rejects the pre-request transition window that event membership alone cannot). The pair-equality check stays as defense-in-depth. Added the requested hook test: a same-pair previous quote retained across an account switch is not adopted, and the verdict re-establishes once the new account's own round settles. 69 focused tests pass; lint/tsc green.

originalix (commented):
@haicongliao

Reviewed commit `6bfa6f6c313a`.

- **P2 · Request-starting state still proves the retained quote**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12850#issuecomment-5300357287)

haicongliao:
Request-starting interval addressed in fb0c99380d. Verified the exact sequence in quoteAction: it resets swapQuoteEventTotalCountAtom (clearing the event id) and writes the new lock with actionLock: true before the async runQuoteEvent flips any loading flag, while the previous quote stays selected — so the no-event-id fallback plus the freshly matching lock did prove a stale quote in that window. The proof is now a single composed predicate in the quote-progress layer, isSwapQuoteProvenForCurrentRequest, which treats actionLock-with-no-event-id as unproven before checking event membership and the lock match. Unit tests cover the request-starting interval with a retained same-pair quote (the reported hole), previous-event rejection after the new event reports, streamed current-event adoption while still locked, and lock mismatch. 96 focused tests pass; lint/tsc green.

zhaono1 (approved):

sidmorizon:
@cursoragent review

originalix (approved):

PR #12857, opened 2026-08-13
feat: add Perps responsive chart panels (OK-59954, OK-60207, OK-60208, OK-60322, OK-60428, OK-60459, OK-60664, OK-60674, OK-60984, OK-61272)

Minnzen:
[OK-59954](https://onekeyhq.atlassian.net/browse/OK-59954) [OK-60207](https://onekeyhq.atlassian.net/browse/OK-60207) [OK-60208](https://onekeyhq.atlassian.net/browse/OK-60208) [OK-60322](https://onekeyhq.atlassian.net/browse/OK-60322) [OK-60428](https://onekeyhq.atlassian.net/browse/OK-60428) [OK-60459](https://onekeyhq.atlassian.net/browse/OK-60459) [OK-60664](https://onekeyhq.atlassian.net/browse/OK-60664) [OK-60674](https://onekeyhq.atlassian.net/browse/OK-60674) [OK-60984](https://onekeyhq.atlassian.net/browse/OK-60984) [OK-61272](https://onekeyhq.atlassian.net/browse/OK-61272)

----------

## Summary

- add a bottom floating chart panel to the mobile Perps trade layout
- add a resizable chart/order-info split on desktop (single drag bar under the chart)
- persist the desktop chart height, add a "Return to default layout" menu entry, and preserve the TradingView instance across fullscreen transitions
- keep pinned desktop Perps table headers synchronized with horizontal body scrolling, using compositor-driven web animation with a JavaScript fallback

## Intent & Context

OK-59954 asks for Binance-style access to a chart from the bottom of the Perps trading screen while retaining the existing trading layout. The desktop experience also needs visible theme-aware split controls for chart/order information and trading/account sections.

OK-60984 covers the desktop/web table-header regression where the pinned header visibly trails or stops short of the body during horizontal scrolling. The header and rows must stay pixel-aligned during rapid wheel, trackpad, and scrollbar movement.

## Root Cause

The fixed desktop layout did not expose split controls or persist user-selected pane sizes. The initial mobile chart implementation participated in the ScrollView layout, so expanding it shifted content rather than behaving as an overlay. During the desktop split refactor, the explicit chart/order-info boundary was also removed and the right pane could shrink below its previous 320 px desktop baseline.

OK-60664 is a latent layout bug this PR exposed rather than introduced. The desktop Positions empty state (pre-existing since v6.2.0) puts its horizontal header `ScrollView` directly inside a column `YStack`. `react-native-web` gives `ScrollView` a base style of `flexGrow: 1; flexShrink: 1`, so in a column parent it competes for vertical space with the sibling empty-state `flex={1}` and takes roughly half the pane. The header row stretched with it and its `borderBottom` was drawn at the bottom of the stretched box. This was invisible while the tab pane had no bounded height; once OK-60428 made the pane `position: absolute; inset: 0`, the free space appeared and the gap became visible.

OK-60674 is a regression from the same OK-60428 commit. Pinning the desktop table header moved it out of the body scroller, and in the move it lost `width="100%"` while its new `ScrollView` was given no `contentContainerStyle`. The body scroller still sets `{ minWidth: scrollableMinWidth, flexGrow: 1 }`, so the body stretches to the full pane width while the header stops at the sum of the column min-widths. Any pane wider than that shows header cells drifting away from their columns. It affects every tab that renders through `CommonTableListView`, not just Positions.

OK-60984 remained after the sizing fix because the pinned header and body are separate scrollers. The JavaScript mirror only updates when React Native Web emits `onScroll`; throttled events can omit the final offset, leaving the header behind the body. Supported web/desktop browsers can instead bind the body scroll offset directly to the header transform through a scoped scroll timeline, while the JavaScript mirror remains the fallback.

## Design Decisions

- Keep mobile free of drag interactions; anchor the chart panel above the bottom navigation and expand it upward as an overlay so the underlying ScrollView height and position remain unchanged.
- Use the existing TradingViewNative chart. The bottom panel mounts it only while expanded; the opt-in top chart prewarms it once after UI idle (`deferHeavyWorkUntilUIIdle`) so the first explicit reveal does not also pay the K-line fetch and Skia mount cost, keeping at most one hidden chart mounted.
- Compact chart placements (mobile top inline and bottom panel) intentionally use the lightweight `TradingViewNative` renderer instead of the `PerpCandles`/`TradingViewPerpsV2` WebView. They are glance-oriented and open/close frequently, so they must render instantly with minimal memory. Trading overlays (position/order lines, trade marks) and the webview reload contract stay on the full chart entries (desktop workspace and the mobile full-screen chart), which remain one tap away. This mirrors the Market detail page, which uses the same native renderer without overlays.
- Use a single vertical Allotment split between chart and order info. Persist the chart height and restore the default on double click or via the settings menu. The trading column keeps its fixed responsive width.
- Keep the desktop TradingView workspace mounted through fullscreen transitions and shield its iframe only during drag gestures.
- Use theme border tokens for visible separators and preserve an explicit chart/order-info boundary in light and dark themes.
- Wrap the Positions empty-state header `ScrollView` in a row `XStack` so its `flexGrow` applies horizontally only, and move the border and background to that wrapper. This matches the working `CommonTableListView` `desktopHeader` pattern instead of inventing a second one.
- Give the pinned header scroller the same `contentContainerStyle` and `flex={1}` content as the body scroller, so one sizing rule drives both instead of two that can diverge.
- Scope the scroll timeline to the desktop Perps table and enable it only when every required CSS feature is supported; unsupported browsers keep the existing JavaScript synchronization path.
- Forward `testID` explicitly through the shared `ScrollView` because React Native Web otherwise drops Tamagui's `data-testid` rewrite on this component.

## Changes Detail

- `PerpMobileChartPanel`: floating mobile chart bar and expanded chart overlay.
- `PerpMobileLayout`: places the chart overlay outside scrolling content and applies the iOS tab-bar offset.
- `PerpDesktopLayout.web`: adds the chart/order-info split with persistence, reset behavior, iframe drag shielding, and themed separators; trading column stays fixed.
- `perpLayoutUtils` and Perps layout state: centralize split sizing/clamping and persisted dimensions.
- `PerpPositionsList`: desktop zero-position empty state no longer splits the pane height with its header; the divider sits directly under the header row (OK-60664).
- `CommonTableListView`: pinned desktop header now mirrors the body scroller's content sizing, so header cells stay aligned with their columns on every order-info tab (OK-60674); it now exposes scoped table/header/body hooks, uses a 1 ms JavaScript fallback throttle, and matches the app background (OK-60984).
- shared web CSS: adds a fully feature-gated scroll timeline that maps the desktop Perps body offset to the pinned header transform without per-frame React updates (OK-60984).
- shared `ScrollView`: preserves explicit `testID` values through React Native Web so the scoped CSS hooks reach the rendered scroll views.
- tests and test IDs: cover the chart split, reset, theme behavior, fullscreen mount stability, mobile overlay anchoring, size clamping, and fixed-height order book rows.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Mobile, Web, Desktop
- **Risk Areas**: mobile overlay stacking/safe-area offsets; persisted desktop pane sizes across viewport changes; iframe pointer handling during split drags; shared `ScrollView` test-ID forwarding; web-only scroll-timeline behavior scoped to desktop Perps tables.

## Test plan

- [x] Run targeted Jest suites for desktop splits, mobile chart overlay, and split sizing utilities.
- [x] Run final lint, format, TypeScript, and targeted Jest checks for the OK-60984 changes (9/9 tests passed).
- [x] Verify Desktop/Web rapid horizontal scrolling keeps the pinned header aligned with the table body.
- [ ] Verify a browser without scroll-timeline support continues to use the JavaScript fallback.
- [ ] Verify Mobile Web and iOS/Android: open and close the bottom chart panel without shifting the underlying layout.
- [ ] Verify Desktop/Web: drag the chart separator, refresh to confirm persistence, and double-click it (or use the "Return to default layout" menu entry) to restore the default.
- [ ] Verify light and dark themes show all separators.
- [ ] Verify desktop TradingView fullscreen round-trip does not reload the chart.
- [ ] Verify Desktop/Web Positions tab with zero positions: header divider sits under the header row and the empty state is centered in the remaining height (OK-60664).
- [ ] Verify Desktop/Web with a wide order-info pane: header cells stay aligned with their columns across Positions, Open orders, TWAP, Trades history, and Account history, both before and after dragging the chart split (OK-60674).

[OK-59954]: https://onekeyhq.atlassian.net/browse/OK-59954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60207]: https://onekeyhq.atlassian.net/browse/OK-60207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60208]: https://onekeyhq.atlassian.net/browse/OK-60208?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60322]: https://onekeyhq.atlassian.net/browse/OK-60322?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60428]: https://onekeyhq.atlassian.net/browse/OK-60428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60664]: https://onekeyhq.atlassian.net/browse/OK-60664?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60674]: https://onekeyhq.atlassian.net/browse/OK-60674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60984]: https://onekeyhq.atlassian.net/browse/OK-60984

[OK-60459]: https://onekeyhq.atlassian.net/browse/OK-60459

[OK-61272]: https://onekeyhq.atlassian.net/browse/OK-61272


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Perp/layouts/PerpMobileLayout.tsx
```diff
@@ -362,125 +365,130 @@ export function PerpMobileLayout() {
...
+          >
+            <SpotBalanceList isMobile useTabsList={false} disableListScroll />
+          </YStack>
         </YStack>
-      </YStack>
-    </ScrollView>
+      </ScrollView>
+      {!isSplitMainActive ? (
+        <PerpMobileChartPanel bottomOffset={tabBarHeight ?? 0} />
+      ) : null}
```
sidmorizon: **Bottom of the mobile trading list is permanently hidden behind the new chart bar**

Severity: `non-severe`

The scrolling content keeps its old bottom spacing (`paddingBottom: tabBarHeight` at `packages/kit/src/views/Perp/layouts/PerpMobileLayout.tsx:371`) while the always-visible chart bar is placed on top of it, so the last strip of positions/orders can never be scrolled fully into view.
Impact: On mobile, the final row of the positions/open-orders/balances lists stays covered by the chart bar and cannot be read or tapped.

<details>
<summary>Overlay height is not added to the scroll content inset</summary>

`PerpMobileChartPanel` is rendered as an absolutely positioned overlay anchored at `bottom={bottomOffset}` (`packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx:60-70`) with a collapsed toggle row of `minHeight="$12"` plus vertical padding (`packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx:97-110`), i.e. roughly 48px tall. The overlay is a sibling of the `ScrollView` inside the new relative `YStack` (`packages/kit/src/views/Perp/layouts/PerpMobileLayout.tsx:368-491`), so it visually covers the bottom ~48px of the scroll viewport. Because `contentContainerStyle` still reserves only `tabBarHeight`, the scrollable content can be scrolled to its end and the final ~48px remain behind the bar. The fix is to add the collapsed bar height (exported from the panel or measured via `onLayout`) to the `ScrollView` bottom padding.

</details>

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+      position="absolute"
+      right={0}
+      bottom={bottomOffset}
+      left={0}
+      zIndex={FLOAT_NAV_BAR_Z_INDEX}
+      bg="$bgApp"
+      borderTopWidth="$px"
+      borderTopColor="$borderSubdued"
+    >
+      {isExpanded ? (
```
sidmorizon: # 🟡 P2 每次折叠都卸载 TradingViewNative,反复展开触发全量重建与数据重拉

## 问题

展开内容用 `{isExpanded ? <HeaderScrollGestureWrapper>…<TradingViewNative …/>…</> : null}` 条件挂载,折叠时整棵子树(含 `TradingViewNative` → lazy 加载的容器 → hyperliquid 数据订阅)被卸载。用户每次点开 toggle 都会重新走 lazy 加载、重新拉取 K 线历史、重建 websocket 订阅,并丢失上次的周期/缩放/十字线状态。

## 影响

该 toggle 条常驻在交易页底部,预期会被高频开合;每次展开都有明显的图表白屏/加载延迟和一次完整的 candle snapshot 请求 + 订阅重建。这与本 PR 在桌面端专门用 `Allotment.Pane visible` 保持 TradingView 挂载、避免 remount 的设计目标相悖。

## 建议

首次展开后保持挂载,折叠时仅隐藏(高度 0 + `overflow="hidden"`),与桌面端策略一致:

```tsx
const [hasExpandedOnce, setHasExpandedOnce] = useState(false);
// handleToggle 中 next && setHasExpandedOnce(true)
{hasExpandedOnce ? (
  <YStack h={isExpanded ? MOBILE_CHART_HEIGHT : 0} overflow="hidden">
    …TradingViewNative…
  </YStack>
) : null}
```

若担心后台开销,可在折叠时通过现有 props/信号暂停实时订阅而非卸载组件。

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx
```diff
@@ -52,9 +77,53 @@ function PerpDesktopLayout() {
   }, []);
 
   const chartExpanded = layoutState.chartExpanded ?? false;
+  const mainSplitSizes = useMemo(
+    () =>
+      getPerpDesktopMainSplitSizes({
+        availableWidth: viewportWidth,
```
sidmorizon: # 🟡 P2 主分栏 defaultSizes 用窗口宽度而非容器宽度,首帧与窗口拖拽期间布局跳变

## 问题

`mainSplitSizes`(传给 `Allotment` 的 `defaultSizes`)以 `availableWidth: viewportWidth`(`useWindowDimensions` 的窗口宽度)计算,但分栏容器实际宽度要减去桌面端左侧导航栏等 chrome。Allotment 会把 `defaultSizes` 按真实容器宽度等比缩放,因此首帧 trading 列渲染为 `saved * container/window`(例如保存 `420px` 时首帧约 `350px`)。修正依赖第 234 行的 `useEffect` + `requestAnimationFrame` 调 `resize`,而 effect 在首帧 paint 之后执行,rAF 又落在下一帧。

同一根因在窗口连续 resize 时也生效:Allotment 的 ResizeObserver 先按比例重排(trading 宽度偏离保存值),下一帧 rAF 再 `resize` 拉回,形成逐帧"拉锯"。

## 影响

桌面端每次进入 Perps 页,右侧交易面板都先以错误宽度渲染再跳变(保存宽度与默认值差距越大越明显);拖拽窗口大小时 trading 列宽度可见抖动。每次必现的视觉/体验问题。

## 建议

用真实容器宽度做首帧计算,而不是事后修正。例如把第 234 行的 effect 换成 `useLayoutEffect`(去掉 rAF,直接读 `offsetWidth` 并 `resize`),保证在浏览器 paint 前完成修正:

```tsx
useLayoutEffect(() => {
  const availableWidth =
    mainSplitContainerRef.current?.offsetWidth || viewportWidth;
  mainSplitRef.current?.resize(
    getPerpDesktopMainSplitSizes({
      availableWidth,
      defaultTradingWidth: layout.widths.trading,
      savedTradingWidth: layoutState.tradingWidth,
    }),
  );
}, [layout.widths.trading, layoutState.tradingWidth, viewportWidth]);
```

或在容器 `onLayout`/ref 测得实际宽度前不渲染 `Allotment`,测量后再用该宽度计算 `defaultSizes`。

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

packages/kit/src/views/Perp/layouts/PerpMobileLayout.tsx
Minnzen: Fixed in e6d28fa: the ScrollView now reserves the collapsed bar height via `PERP_MOBILE_CHART_BAR_SCROLL_INSET` (48px, exported from `PerpMobileChartPanel`) on top of the tab-bar offset whenever the floating bar is rendered — same pattern as the Borrow mobile action bar inset.

packages/kit/src/views/Perp/layouts/PerpMobileLayout.tsx
```diff
@@ -362,125 +365,130 @@ export function PerpMobileLayout() {
...
-      </YStack>
-      <XStack
-        gap="$3.5"
-        px="$4"
-        pb="$4"
-        onLayout={(event) => handleTraceLayout('firstScreenGrid', event)}
+    <YStack flex={1} position="relative" bg="$bgApp">
+      <ScrollView
+        style={{ flex: 1, backgroundColor: '$bgApp' }}
+        contentContainerStyle={{ flexGrow: 1, paddingBottom: tabBarHeight }}
```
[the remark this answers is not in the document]
Minnzen: 已修复 (e6d28fa):ScrollView 的 paddingBottom 在浮动图表栏渲染时额外预留 48px(`PERP_MOBILE_CHART_BAR_SCROLL_INSET`),与 Borrow 浮动操作栏的滚动 inset 方案一致。

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+          excludeRightEdgeRatio={0.1}
+          scrollScale={1.2}
+          verticalPanMaxPointers={1}
+          simultaneousWithNativeGesture
+          cancelChildTouches={false}
+        >
+          <YStack
+            testID={PerpTestIDs.MobileChartContent}
+            h={MOBILE_CHART_HEIGHT}
+            overflow="hidden"
```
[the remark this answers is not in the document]
Minnzen: 已修复 (e6d28fa):展开高度改为按视口 clamp——`windowHeight - bottomOffset - 220(顶部预留)`,并限制在 240–500px 区间,矮视口下不再越过页面顶部遮挡标题。

packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx
```diff
@@ -64,11 +133,124 @@ function PerpDesktopLayout() {
...
+
+  useEffect(() => {
+    const frame = requestAnimationFrame(() => {
+      const availableWidth =
+        mainSplitContainerRef.current?.offsetWidth || viewportWidth;
+      mainSplitRef.current?.resize(
+        getPerpDesktopMainSplitSizes({
+          availableWidth,
+          defaultTradingWidth: layout.widths.trading,
+          savedTradingWidth: layoutState.tradingWidth,
```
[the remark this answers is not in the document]
Minnzen: 已修复 (e6d28fa):新增 useLayoutEffect,在持久化值 hydration 完成或响应式高度变化时对 chart/trading 两个纵向 split 调用 resize(),与 main split 的回放行为对齐。

packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx
```diff
@@ -118,115 +405,80 @@ function PerpDesktopLayout() {
...
+            <Allotment
+              ref={mainSplitRef}
+              id={PerpTestIDs.DesktopMainSplit}
+              separator={!chartExpanded}
+              defaultSizes={mainSplitSizes}
+              onDragStart={handleMainSplitDragStart}
+              onDragEnd={handleMainSplitDragEnd}
+              onReset={handleMainSplitReset}
+            >
+              <Allotment.Pane minSize={PERP_LAYOUT_CONFIG.main.marketMinWidth}>
```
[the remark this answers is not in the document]
Minnzen: 已修复 (e6d28fa):订单簿可见时 market pane 的 minSize 会加上订单簿宽度(marketMinWidth + orderBookWidth),保证图表可用宽度;Allotment 会对 minSize 变化即时 relayout。

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
Minnzen: 已按建议修复 (e6d28fa):首次展开后保持挂载,折叠仅收到高度 0 + overflow hidden,重开不再重建图表、重拉 K 线或丢失周期/缩放状态。折叠期间实时订阅仍活跃,若后续观察到后台开销问题,可再通过现有 props 加暂停信号。

packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx
Minnzen: 已修复 (e6d28fa):改为 useLayoutEffect 直接读容器 offsetWidth 计算并 resize,去掉 rAF,修正在浏览器 paint 前完成,首帧不再跳变;窗口连续 resize 时同样走该同步路径,消除逐帧拉锯。

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
```diff
@@ -0,0 +1,153 @@
...
+            excludeRightEdgeRatio={0.1}
+            scrollScale={1.2}
+            verticalPanMaxPointers={1}
+            simultaneousWithNativeGesture
+            cancelChildTouches={false}
+          >
+            <YStack h={chartHeight} overflow="hidden">
+              {chartSource ? (
+                <TradingViewNative
+                  key={coin}
```
sidmorizon: # 🟡 P2 折叠态切换币种会离屏重建图表并挂起新的实时订阅

## 问题

这是"保持挂载"修复(e6d28fa)引入的次生问题:`hasExpandedOnce` 置位后图表永久保持挂载,而 `TradingViewNative` 使用 `key={coin}`。当用户在图表折叠(`h=0`、被裁剪不可见)期间切换交易对时,`coin` 变化会卸载旧图表并在离屏状态下完整重建新图表——加载 K 线历史、建立 Hyperliquid OHLCV websocket 实时订阅,并在每个 tick 上驱动 Skia Canvas 重绘。

注释声明的权衡是"折叠只隐藏已挂载图表,保留其数据、订阅与视图状态",但 `key` 重建恰恰丢弃了全部已保留状态——保挂载的收益为零,却仍支付隐藏图表的全部网络/CPU 成本,且用户此后可能再也不展开图表。

## 影响

移动端(iOS/Android,main runtime)在折叠图表后每次切换币种,都会在后台产生一次完整图表加载 + 持续 websocket 订阅 + 每 tick 的 Skia 重绘,直至下次展开。对高频切换市场的用户造成持续且不可见的电量、流量与 CPU 开销。

## 建议

折叠期间 `coin` 变化时放弃保挂载,展开时再按新 coin 重建;同一 coin 的折叠/展开仍保留状态:

```tsx
const [mountedCoin, setMountedCoin] = useState<string | undefined>();
const handleToggle = useCallback(() => {
  const next = !isExpanded;
  setIsExpanded(next);
  if (next) {
    setMountedCoin(coin);
  }
}, [coin, isExpanded]);

const shouldKeepMounted =
  mountedCoin !== undefined && (isExpanded || mountedCoin === coin);
// 用 shouldKeepMounted 替代 hasExpandedOnce 作为挂载条件
```

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

packages/kit/src/views/Perp/layouts/perpLayoutUtils.ts
```diff
@@ -42,6 +47,98 @@ function clampSize(value: number, min: number, max: number) {
...
+    Math.min(
+      PERP_LAYOUT_CONFIG.main.tradingMaxWidth,
+      availableWidth - PERP_LAYOUT_CONFIG.main.marketMinWidth,
+    ),
+  );
+  const tradingWidth = clampSize(
+    tradingWidthCandidate,
+    PERP_DESKTOP_TRADING_MIN_WIDTH,
+    tradingMaxWidth,
+  );
```
sidmorizon: **Saved width of the trading column can squeeze the chart column below its allowed minimum on narrow desktop windows**

Severity: `non-severe`

The saved trading-column width is limited (`getPerpDesktopMainSplitSizes` at `packages/kit/src/views/Perp/layouts/perpLayoutUtils.ts:90-101`) using only the bare chart minimum and ignores the space the order book occupies, so on narrow windows the restored layout asks for more width than the left side can give and the panes snap to a different size than the one that was saved and re-applied.
Impact: On smaller desktop windows the chart/order-book area can jump to a different width than the user's saved arrangement right after load.

<details>
<summary>Mismatch between computed split sizes and the pane minSize enforced by Allotment</summary>

`packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx:144-146` sets the left pane's `minSize` to `PERP_LAYOUT_CONFIG.main.marketMinWidth + layout.widths.orderBook` (400 + 280..360) when the order book is visible, while `getPerpDesktopMainSplitSizes` clamps the trading width using only `PERP_LAYOUT_CONFIG.main.marketMinWidth` (400). With e.g. a 1100px container and a persisted `tradingWidth` of 500, the helper returns `[600, 500]`, but the left pane's minimum is 680, so Allotment overrides the requested sizes. The same sizes are re-applied on every hydration/resize by the layout effect at `packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.tsx:256-266`, so the requested and actual layout stay out of sync until the user drags again.

</details>

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
Minnzen: 已按建议修复 (a17e579):改为跟踪 mountedCoin——折叠时仅当币种未变才保持挂载;折叠期间切换币种会卸载隐藏图表,避免离屏全量重建与后台订阅,下次展开再按新币种构建。展开状态下切币会同步 mountedCoin,收起后仍保留当前图表。已补对应测试。

packages/kit/src/views/Perp/layouts/perpLayoutUtils.ts
Minnzen: Fixed in a17e579: `getPerpDesktopMainSplitSizes` now takes a `marketMinWidth` param, and all three call sites (defaultSizes memo, the layout effect, and reset) pass the same order-book-aware `marketPaneMinWidth`, so the computed sizes can no longer conflict with the pane minSize Allotment enforces. Covered by a new utils test.

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
```diff
@@ -0,0 +1,165 @@
...
+  useEffect(() => {
+    if (isExpanded && coin) {
+      setMountedCoin(coin);
+    }
+  }, [coin, isExpanded]);
+
+  // Keep the collapsed chart mounted only for the coin it was built for; a
+  // coin switch while hidden would otherwise rebuild it fully off-screen.
+  const shouldKeepMounted =
+    mountedCoin !== undefined && (isExpanded || mountedCoin === coin);
```
sidmorizon: # 🟡 P2 折叠期间币种切走再切回时,隐藏图表仍会被离屏重建(mountedCoin 未清空)

## 问题

这是上一条 P2(折叠态离屏重建)在 `a17e579` 修复后的残留边缘:折叠态下币种从 `mountedCoin`(如 BTC)切到 ETH 时隐藏图表被正确卸载,但 `mountedCoin` 仍保留为 `'BTC'`——`handleToggle` 和展开同步 effect 都只会覆盖为新值,没有任何路径将其置回 `undefined`。

当用户保持折叠并把币种切回 BTC(在两个常用市场间来回浏览很常见),`mountedCoin === coin` 重新成立,`shouldKeepMounted` 变回 `true`,组件在 `h=0` 的隐藏容器里重新挂载 `TradingViewNative`——全新构建加行情订阅全部发生在离屏状态。新增测试只覆盖了"切走后卸载 + 展开挂新币",未覆盖"切走再切回"路径。

## 影响

折叠态下 `A→B→A` 的币种往返会触发一次完全不可见的图表重建与后台数据订阅,与修复声明的"折叠期间切币不再离屏重建"语义不符——只是把触发条件从"任意切币"缩小到了"切回已挂载币种"。

## 建议

卸载隐藏图表的同时清空 `mountedCoin`,使"丢弃"成为终态,直到下次展开才重新挂载:

```tsx
// Drop the stale mount record so a later switch back to the same coin
// does not rebuild the chart off-screen.
useEffect(() => {
  if (!isExpanded && mountedCoin !== undefined && mountedCoin !== coin) {
    setMountedCoin(undefined);
  }
}, [coin, isExpanded, mountedCoin]);
```

此时 `shouldKeepMounted` 可简化为 `mountedCoin !== undefined`。并补一条测试:expand → collapse → 切 ETH → 切回 BTC(仍折叠),断言图表未重新挂载。

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

sidmorizon: **Collapsed mobile chart silently reloads in the background when the user returns to a previously viewed market**

Severity: `non-severe`

The hidden chart is rebuilt and starts streaming again (`shouldKeepMounted` at `packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx:84-85`) whenever the user navigates back to the market it was last opened for, even though the panel is closed, so an invisible chart keeps consuming network and battery.
Impact: Users switching markets back and forth with the chart closed pay for an off-screen chart they never opened.

<details>
<summary>Why the mount guard misses the return-to-previous-coin case</summary>

`mountedCoin` is only updated when the panel is expanded (`packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx:67-80`), and it is never cleared when the chart is dropped because the coin changed. So after the sequence expand BTC -> collapse -> switch to ETH (subtree unmounts, `mountedCoin` stays `'BTC'`) -> switch back to BTC, `mountedCoin === coin` becomes true again while `isExpanded` is false, and the whole `TradingViewNative` subtree (`packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx:99-128`) is re-created inside a zero-height container. This contradicts the stated intent of the guard ("a coin switch while hidden would otherwise rebuild it fully off-screen"). Clearing `mountedCoin` when the chart is dropped while collapsed would keep the guard consistent.

</details>

Minnzen: 已修复 (0d713a3):折叠态下币种切走时同步清空 mountedCoin,切回原币种不再满足挂载条件,不会离屏重建;重新展开时才按当前币种构建。已补切走再切回的测试用例。

Minnzen: Fixed in 0d713a3: mountedCoin is now cleared when the coin moves away while collapsed, so returning to a previously viewed market no longer remounts the hidden chart; it only rebuilds on the next expand. Covered by an extended test.

packages/kit/src/views/Perp/layouts/perpLayoutUtils.ts
```diff
@@ -107,39 +200,27 @@ export function getVerticalOrderBookLayout(
...
     };
   }
 
-  let levelsPerSide = Math.floor(
-    (bookBodyHeight / (ORDER_BOOK_VERTICAL_ROW_HEIGHT_MIN + 1) - 1) / 2,
-  );
-  levelsPerSide = Math.max(
+  // Fixed row height: resizing changes the visible level count, never the
+  // red/green depth bar size.
+  const levelsPerSide = Math.max(
```
sidmorizon, line 170: # 🟡 P2 固定行高叠加 18 档硬上限,高视口下订单簿底部出现无界空白

## 问题

`getVerticalOrderBookLayout` 新实现将行高固定为 `ORDER_BOOK_VERTICAL_ROW_HEIGHT`(22px),注释承诺"resizing changes the visible level count"。但档位数被 `maxLevelsPerSide` 截断,桌面端该值固定为 `desktopMaxLevelsPerSide = propMaxLevelsPerSide ?? 18`(`PerpOrderBook.tsx`)。当图表分栏被拖高使 `bookBodyHeight > 23 * 37 ≈ 851px`(高视口下 chart 分栏顶部可拖到 `totalHeight - PERP_DESKTOP_INFO_MIN_HEIGHT`,完全可达)后,档位数卡死在 `18`,注释的不变式失效。

旧实现在触顶后会把 `rowHeight` 放大为 `bookBodyHeight/(2L+1) - gap` 拉伸填满容器;新实现固定 22px,而订单簿行渲染在 `styles.absoluteContainer`(`top: 0` 顶部对齐)中,多余高度全部堆积为 bids 下方的空白。

## 影响

大显示器(视口高约 `1200px` 以上)用户把图表分栏拖高后,桌面纵向订单簿下方出现一条随拖拽持续变大的死区空白(可超过 `300px`),既不显示更多深度也不拉伸内容,视觉上像布局 bug。未触顶时残留空隙不超过约 `45px`(不足一行),可接受;触顶后无界。

## 建议

在档位数被 `maxLevelsPerSide` 截断(而非被容器高度限制)时吸收剩余空间,例如仅在触顶时回退为填充式行高:

```ts
const capped =
  levelsPerSide === maxLevelsPerSide &&
  bookBodyHeight >
    (2 * levelsPerSide + 1) *
      (ORDER_BOOK_VERTICAL_ROW_HEIGHT + ORDER_BOOK_VERTICAL_ROW_GAP);
return {
  levelsPerSide,
  rowHeight: capped
    ? Math.floor(
        bookBodyHeight / (2 * levelsPerSide + 1) - ORDER_BOOK_VERTICAL_ROW_GAP,
      )
    : ORDER_BOOK_VERTICAL_ROW_HEIGHT,
};
```

若必须保持深度条尺寸恒定,也可提高桌面 `desktopMaxLevelsPerSide` 上限,或在纵向渲染容器中将内容围绕 spread 行对称留白,避免空白全部堆在底部。

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

originalix (commented):
@Minnzen

Reviewed commit `a0326a6e3e9a`.

- **P1 · Inline mobile chart bypasses the existing Perps chart integration**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/views/Perp/components/OrderBook/index.tsx
```diff
@@ -977,23 +978,24 @@ export function OrderBook({
     selectedTickOption,
     priceDecimals,
     sizeDecimals,
+    verticalExtraBidLevels,
```
sidmorizon, line 981: # 🟡 P2 extra bid 档计入 bidDepth,B/S 深度比率产生随布局奇偶翻转的买方偏差

## 问题

`useAggregatedBook` 现在把 bids 多聚合/多切 1 档(`verticalExtraBidLevels`),而 asks 仍是 `levelsPerSide` 档。`bidDepth` 取 `aggregatedData.bids.at(-1)?.cumSize`,包含了这档 extra bid;随后与只覆盖 `levelsPerSide` 档的 `askDepth` 一起传给 `OrderBookSideRatio` 计算 B/S 百分比。

## 影响

`extraBidLevels = 1` 时,比率条比较的是 `13` 档买单深度 vs `12` 档卖单深度(以 12 档为例),系统性高估买方压力(档位大小均匀时约 `+2pp`);且面板高度拖动约 `23px` 就会使 `extraBidLevels` 在 `0/1` 间翻转,B/S 显示值随之跳变——一个交易参考指标被纯布局奇偶因素影响。

## 建议

比率计算与渲染解耦:extra 行照常渲染,但 side ratio 用对称档数的累计量,例如:

```ts
const ratioBidDepth = new BigNumber(
  aggregatedData.bids[
    Math.min(resolvedMaxLevelsPerSide, aggregatedData.bids.length) - 1
  ]?.cumSize ?? '0',
);
```

将 `ratioBidDepth` 传给 `OrderBookSideRatio`;`DepthBarColumn` 的归一化仍可用含 extra 档的 `bidDepth`。

[Claude session ↗](https://claude.ai/code/session_012WqAHptY4qmFVoERpsASmX)

---

originalix (commented):
@Minnzen

Reviewed commit `1783def888f9`.

- **P1 · Inline mobile chart bypasses the existing Perps chart contract**
- **P2 · Order-book side ratio now depends on pane height instead of symmetric depth**
- **P2 · Tall desktop charts can leave a growing dead zone under the vertical order book**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/views/Perp/layouts/perpLayoutUtils.ts
Minnzen: 已修复 (15fc885):档位数被 maxLevelsPerSide 截断且容器仍有富余时,回退为拉伸行高填满面板(与旧实现触顶行为一致);未触顶时行高保持 22px 固定,拖拽仅增减档位。已补触顶拉伸测试。

packages/kit/src/views/Perp/components/OrderBook/index.tsx
Minnzen: 已修复 (15fc885):新增 ratioBidDepth,B/S 比率按两侧相同档数(resolvedMaxLevelsPerSide)的累计深度计算,extra bid 行只参与渲染与深度条归一化,不再影响比率,消除随布局奇偶的跳变。

originalix (commented):
@Minnzen

Reviewed commit `15fc88551dfb`.

- **P1 · Inline mobile chart bypasses the existing Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `64f465535607`.

- **P1 · Inline mobile chart bypasses the established Perps chart lifecycle**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/shared/src/locale/enum/translations.ts
```diff
@@ -3283,13 +3283,18 @@ export enum ETranslations {
   perps__claim_now__action = 'perps__claim_now__action',
   perps__fee_discount_activated__msg = 'perps__fee_discount_activated__msg',
   perps__snooze_remind_later__action = 'perps__snooze_remind_later__action',
+  perps_back_to_default_layout__action = 'perps_back_to_default_layout__action',
+  perps_chart_do_not_show__action = 'perps_chart_do_not_show__action',
   perps_footer_help_us_better = 'perps_footer.help_us_better',
   perps_hide_other_pairs = 'perps_hide_other_pairs',
+  perps_layout_reset__msg = 'perps_layout_reset__msg',
+  perps_layout_settings__title = 'perps_layout_settings__title',
   perps_referral_campaign__desc = 'perps_referral_campaign__desc',
```
sidmorizon, line 3483: **Auto-generated translation files were edited by hand**

Severity: `non-severe`

The auto-generated translation enum and every locale JSON file are edited directly (new keys added at `packages/shared/src/locale/enum/translations.ts:3286-3288`), which the repository rules forbid because those files are produced by the translation sync tooling.
Impact: The next translation sync can overwrite or conflict with these hand-written entries, causing missing or broken UI text.

<details>
<summary>Rule source and affected files</summary>

`.cursor/rules/locale-handling.mdc` states: "Do not modify the code in `@onekeyhq/shared/src/locale/enum/translations.ts` as these are automatically generated" and "Do not modify the locale JSON files in `@onekeyhq/shared/src/locale/json/*`". `CLAUDE.md` repeats this under Core Rules: "Never modify generated translations (`translations.ts`, locale JSON files)."

This PR adds `perps_back_to_default_layout__action`, `perps_chart_do_not_show__action`, `perps_desktop_resize_panels__desc`, `perps_layout_reset__msg`, `perps_layout_settings__title`, and `perps_trading_page_chart__title` to `packages/shared/src/locale/enum/translations.ts` and to all 18 files under `packages/shared/src/locale/json/`. These strings should be introduced through the translation pipeline instead of committed by hand.

</details>

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -778,15 +775,13 @@ export function CommonTableListView<T>({
...
+    const MobileListContent = (
       <Stack
         flex={1}
         position="relative"
         onLayout={(event) => handleMobileTraceLayout('listWithShadow', event)}
       >
         {ListContent}
-        <SimpleEdgeShadowOverlay isDark={isDark} position="right" />
       </Stack>
     );
```
sidmorizon, line 790: **Right-edge shadow removed from mobile order/position lists without being part of this change**

Severity: `non-severe`

The right-edge shadow that used to be drawn over the mobile Perps tables is deleted (the `SimpleEdgeShadowOverlay` render was removed from the mobile list wrapper around `packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx:778-786`), so those horizontally scrollable lists silently lose the visual cue that more columns exist off-screen.

Impact: Mobile positions/orders/balances tables no longer show the edge fade, and the visual regression is unrelated to the chart-panel feature this PR describes.

<details>
<summary>Why this is flagged as an out-of-scope removal</summary>

CLAUDE.md Core Rules: "Do not remove existing code/components unless the request requires it." `.cursor/rules/self-testing.mdc` §4 also requires that "Every file touched maps to the stated task / PR title — incidental refactors outside scope are split out or explicitly called out for review".

The PR summary and change list only cover Perps chart panels, desktop split, and layout persistence; nothing mentions the mobile table edge shadow. The diff removes both the `SimpleEdgeShadowOverlay` import and its usage, renaming `ListWithShadow` to `MobileListContent`. `SimpleEdgeShadowOverlay` remains used elsewhere (`packages/kit/src/views/ReferFriends/pages/InviteReward/components/InvitationDetailsSection/components/InviteCodeListTable/index.tsx:119`), so this is a behavior change local to the Perps mobile tables rather than a component-wide cleanup.

</details>

originalix (commented):
@Minnzen

Reviewed commit `f587f3485daa`.

- **P1 · Compact Perps panels bypass the established chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/components/Spotlight/index.tsx
```diff
@@ -242,7 +257,7 @@ function SpotlightContent({
               size="small"
               onPress={onConfirm}
             >
-              {intl.formatMessage({ id: ETranslations.global_done })}
+              {intl.formatMessage({ id: ETranslations.global_ok })}
```
sidmorizon: **Confirmation button text changes for every product tour tooltip, not just the new one**

Severity: `non-severe`

The confirmation button of the shared tour tooltip is switched to a different label (`ETranslations.global_ok` at `packages/kit/src/components/Spotlight/index.tsx:260`) even though the change is only needed for the new trading-chart hint, so every existing tour tooltip across the app silently shows different wording.
Impact: All previously shipped onboarding/tour tooltips (all-networks, refer-a-friend, earn, split-view, etc.) change their button wording without being part of this feature.

<details><summary>Shared component change reaches all existing tour call sites</summary>

`SpotlightContent` is the single renderer used by every `Spotlight`/`SpotlightView` consumer (e.g. `packages/kit/src/components/TabPageHeader/components/WalletConnectionGroup.tsx:37`, `packages/kit/src/hooks/useReferFriends.tsx:220`, `packages/kit/src/components/SplitViewPrompt/SplitViewPrompt.tsx:42`). The PR description scope is Perps chart panels, and the repository review checklist (`.cursor/rules/self-testing.mdc` §4) requires that every touched change map to the stated task or be explicitly called out; an app-wide copy change buried in a Perps feature PR is exactly the incidental change that rule targets.

</details>

originalix (commented):
@Minnzen

Reviewed commit `e5307acef854`.

- **P1 · Compact Perps panels bypass the established chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `0dffc684ae27`.

- **P1 · Compact Perps panels bypass the established chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Minnzen: Fixed in cc97762: restored the SimpleEdgeShadowOverlay on the mobile list wrapper — the removal was an unintended side effect of the MobileListContent refactor.

packages/kit/src/components/Spotlight/index.tsx
Minnzen: Fixed in cc97762: the shared tooltip button copy is back to global_done by default; the new confirmButtonTextId prop overrides it to OK only for the Perps chart-resize tour, so existing tours keep their original wording.

Minnzen: Simplified further in 0ca4041: dropped the confirmButtonTextId prop entirely — the shared tooltip just uses global_done everywhere as before. Zero new API surface.

originalix (commented):
@Minnzen

Reviewed commit `cc97762e63d5`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `0ca4041d47f5`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

Minnzen:
Re the P1 "Compact Perps panels bypass the established chart contract" from the automated review: this is intentional, now documented in the PR description (Design Decisions).

The compact placements (mobile top inline chart and bottom panel) are glance-oriented and open/close frequently, so they use the lightweight `TradingViewNative` renderer: instant render, no WebView startup cost or resident memory. The `PerpCandles`/`TradingViewPerpsV2` contract (position/order lines, trade marks, `showTradeMarks`/`showChartLines` settings, webview reload recovery) remains on the full chart entries — the desktop workspace and the mobile full-screen chart — which stay one tap away. The Market detail page already uses the same native renderer without overlays, so this matches existing precedent.

If product later wants position lines on the compact chart, the path is adding a minimal overlay to the native renderer as a follow-up ticket, not swapping in the WebView chart.

packages/shared/src/locale/enum/translations.ts
Minnzen: Acknowledged — keeping as-is intentionally. The keys ship with English fallbacks plus zh_CN so the feature is not blocked on the translation pipeline; the keys will be registered upstream and the next translation sync will overwrite these entries with proper per-locale text, which is the normal reconciliation path.

originalix (commented):
@Minnzen

Reviewed commit `35c20d84b65c`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `0ccf44d1d8ea`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `032a0f98f4c7`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -981,26 +1000,17 @@ export function CommonTableListView<T>({
     </XStack>
   );
 
-  const shouldEnableDesktopVerticalScroll =
-    enableDesktopVerticalScroll && !disableListScroll;
-
   return (
     <YStack flex={1}>
```
sidmorizon: **Table page controls scroll out of sight on desktop**

Severity: `non-severe`

The page-navigation controls of desktop tables are now placed inside the newly added scrolling area (`ScrollView` at `packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx:1006-1027`) instead of staying below it, so with long tables the controls are pushed off-screen until the user scrolls all the way down.
Impact: On desktop, users of paginated tables can no longer see or use the next/previous page buttons without scrolling to the very bottom of the list.

<details>
<summary>How the restructure moved the footer inside the scroller</summary>

Before this change the desktop return was `YStack > [desktopTable, PaginationFooter]`, with vertical scrolling only enabled when the (never-passed) `enableDesktopVerticalScroll` flag was set, so the footer was always rendered as a sibling below the table and remained visible.

After the change the desktop return is `YStack > [desktopHeader, ScrollView > [desktopTable, PaginationFooter]]`. Because the footer is now a child of the always-on vertical `ScrollView`, it participates in the scrolled content instead of being pinned under it. Additionally `desktopTable` is `<XStack flex={1}>` (`packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx:914`) which no longer receives a bounded height inside scroll content, so `flex={1}` empty/loading placeholders can also collapse.

A fix would keep `PaginationFooter` outside the scroll container so it stays anchored at the bottom of the panel.

</details>

sidmorizon:
@cursoragent review

packages/kit/src/views/Perp/utils/popularTickers.ts
```diff
@@ -35,14 +33,11 @@ export function pickPopularPerpTickers({
...
-      return hotTabItems.slice(0, POPULAR_TICKER_COUNT);
+      return hotTabItems;
     }
   }
 
   return items
     .filter((item) => item.hotScore > 0)
-    .toSorted((a, b) => b.hotScore - a.hotScore)
-    .slice(0, POPULAR_TICKER_COUNT);
+    .toSorted((a, b) => b.hotScore - a.hotScore);
```
sidmorizon, line 42: 🟠 P1: The scrolling price strip at the bottom of the trading screen now shows every market instead of the top ten

Severity: `severe`

The cap that kept only the ten hottest markets was deleted (`pickPopularPerpTickers` at `packages/kit/src/views/Perp/utils/popularTickers.ts:35-42`), so the strip now builds an entry for every tradable market and re-creates all of them on every price update.
Impact: On the perpetuals page the bottom price strip can render hundreds of duplicated items that all refresh on each incoming price tick, which can make the whole trading page stutter.

<details>
<summary>Mechanism: removed POPULAR_TICKER_COUNT slice combined with per-tick re-rendering and marquee duplication</summary>

- `pickPopularPerpTickers` previously ended with `.slice(0, POPULAR_TICKER_COUNT)` on both the hot-tab branch and the hot-score fallback; both slices (and the constant) were removed, so the returned array is the entire scored universe. The spot branch in `packages/kit/src/views/Perp/hooks/usePopularTickers.ts:134-141` also lost its `slice(0, POPULAR_TICKER_COUNT)`.
- `usePopularTickers` now maps live price data onto every item, so a new object is produced for each item on every `allAssetCtxs` / `spotPriceMap` update. `FooterTickerItem` is `memo`'d on value props (`change24hPercent`, `markPrice`), so every item re-renders each tick.
- `FooterTickerMarquee` renders the item list twice when scrolling is needed (`renderLoop(false)` / `renderLoop(true)` at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:275-276`), doubling the node count.
- Only the footer ticker consumes this hook, so the previous cap was the only bound on the item count.

</details>

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx
```diff
@@ -1,106 +1,253 @@
...
+    frameId = requestAnimationFrame(step);
 
-  useEffect(() => {
     return () => {
-      styleElRef.current?.remove();
+      cancelAnimationFrame(frameId);
+      mediaQuery?.removeEventListener?.('change', syncReducedMotion);
+      lastFrameTimeRef.current = null;
     };
+  }, [commitDisplayItems, syncLoopMetrics]);
```
sidmorizon, line 206: **The price strip keeps a browser animation loop running forever, even when nothing is scrolling**

Severity: `non-severe`

An animation frame is re-scheduled on every frame unconditionally (`requestAnimationFrame(step)` at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:204`), even when the strip is not scrolling or the user asked for reduced motion, so the page never becomes idle while the trading screen is open.
Impact: The trading page keeps waking the browser up ~60 times per second for no visible movement, wasting CPU and battery.

<details>
<summary>Mechanism: unconditional rAF loop replacing the previous CSS animation</summary>

The previous implementation drove the marquee with a CSS keyframe animation, so no JS ran per frame. The new `step` callback always ends with `frameId = requestAnimationFrame(step)` and the loop is started once in a `useLayoutEffect` that only cancels on unmount. When `prefersReducedMotionRef.current` is true or `needsScrollRef.current` is false the loop still runs every frame and just rewrites `track.style.transform`. Additionally, when `loopWidthRef.current` is 0 the loop calls `syncLoopMetrics()` (`packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:178`) on every frame, which performs `getBoundingClientRect()` and a `setNeedsScroll` state update per frame.

</details>

originalix (commented):
@Minnzen

Reviewed commit `3747f5e8e804`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `ac16eb027bb2`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**
- **P1 · Desktop chart split renders no draggable separator**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `4e087216cf9d`.

- **P1 · Compact Perps charts bypass the established Perps chart contract**
- **P1 · Desktop chart split renders no draggable separator**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `532cc0ffeb21`.

- **P1 · Compact charts bypass the account-aware Perps chart contract**
- **P2 · Bound the live footer ticker list**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

originalix (commented):
@Minnzen

Reviewed commit `25ce13204899`.

- **P1 · Compact charts bypass the account-aware Perps chart contract**
- **P2 · Bound the live footer ticker list**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -981,26 +1031,17 @@ export function CommonTableListView<T>({
...
-          desktopTable
-        )}
-
+    <YStack flex={1} testID="perp-desktop-table">
+      {desktopHeader}
+      <ScrollView
+        style={{
+          flex: 1,
+        }}
+        nestedScrollEnabled
```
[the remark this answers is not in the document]
Minnzen: Moved `PaginationFooter` back out of the vertical `ScrollView` so it stays pinned under the table, matching `origin/x` — e41b0b3822.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Minnzen: Superseded by the newer thread on the same footer; fixed in e41b0b3822 (footer moved out of the scroller).

packages/kit/src/views/Perp/utils/popularTickers.ts
Minnzen: Intentional, per #12937: the footer strip mirrors the server Hot tab, which defines and bounds the list, so there is no local cap. `FooterTickerItem` is memoized on primitive value props, so only items whose price actually changed re-render per tick. A cap on the hot-score fallback path (used only when the Hot tab is missing) can be added in a follow-up if that list turns out to be large in practice.

packages/kit/src/views/Perp/utils/popularTickers.ts
```diff
@@ -35,14 +33,11 @@ export function pickPopularPerpTickers({
           b.hotScore - a.hotScore,
       );
     if (hotTabItems.length) {
-      return hotTabItems.slice(0, POPULAR_TICKER_COUNT);
+      return hotTabItems;
     }
   }
 
   return items
```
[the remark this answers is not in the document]
Minnzen: Intentional, per #12937: the footer strip mirrors the server Hot tab, which defines and bounds the list, so there is no local cap. `FooterTickerItem` is memoized on primitive value props, so only items whose price actually changed re-render per tick. A cap on the hot-score fallback path (used only when the Hot tab is missing) can be added in a follow-up if that list turns out to be large in practice.

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx
Minnzen: Acknowledged as non-severe: on desktop the strip nearly always overflows and needs the loop, so the idle case (no overflow / reduced motion) is rare. Stopping and restarting the loop needs a small restructure of the effect; tracked as a follow-up rather than widening this PR.

originalix (commented):
@Minnzen

## Automated code review found 1 issue to address

Reviewed commit `793772895823`.

- **P2 · Compact web charts use stale price-axis hit geometry**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12857#issuecomment-5278337520)

development/perf-ci/thresholds/web.cold.json
```diff
@@ -50,7 +50,7 @@
     "swap": {
       "resourceCount": 220,
       "scriptCount": 169,
-      "jsDecodedBytes": 17475174,
+      "jsDecodedBytes": 17504000,
```
originalix, line 53: P2: 不要用阈值调整掩盖 Swap 冷启动回归

这里把 Swap 的 `jsDecodedBytes` 预算从 17,475,174 提高到 17,504,000。当前 CI 产物的实际值是 17,500,295：按旧阈值会超出 25,121 bytes，按新阈值只剩 3,705 bytes 余量，因此 check 变绿是因为放宽了预算，而不是回归消失。

这个 Perps PR 已经让 shared chart panel 代码进入 Swap 冷启动图。建议先定位 import chain 并恢复 lazy boundary，避免 Perps 图表代码出现在 Swap cold graph；只有确认并批准新的基线后再调整阈值。

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
```diff
@@ -0,0 +1,312 @@
...
+      borderBottomColor="$borderSubdued"
+    >
+      <PerpMobileChartContent
+        chartHeight={chartHeight}
+        coin={coin}
+        contentOffsetTop={MOBILE_CHART_CONTENT_OFFSET}
+        contentTestID={PerpTestIDs.MobileTopChartContent}
+        isExpanded={isExpanded}
+        onClose={onClose}
+        preloadWhenCollapsed
```
originalix, line 195: P2: 折叠的顶部图表不应主动预热

顶部图表默认折叠时，这个 prop 会让 `deferHeavyWorkUntilUIIdle()` 设置 `mountedCoin`，从而在高度为 0 的容器里挂载完整 `TradingViewNative`。用户从未展开图表也会触发 lazy container、K 线历史请求和实时订阅；折叠期间切换币种还会重复替换隐藏图表。

这与 PR 描述的“only mount while the panel is expanded”不一致。建议移除 `preloadWhenCollapsed`，首次显式展开时再挂载；如果需要保留图表状态，也只在用户实际展开过以后保活。

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -778,15 +806,13 @@ export function CommonTableListView<T>({
       );
     }
 
-    // Wrap with shadow overlay for native platforms
-    const ListWithShadow = (
+    const MobileListContent = (
```
originalix, line 809: P2: 已修复的移动端表格右侧阴影再次回归

之前的 thread 已确认这是误删，并在 `cc97762` 恢复了 `SimpleEdgeShadowOverlay`；但当前 head 又只导入 `FixedColumnShadowOverlay`，这个 mobile wrapper 也只渲染 `ListContent`，后续提交把修复重新移除了。

Position、Order、Balance 等横向表格因此再次失去“右侧还有列”的视觉提示。请在当前 wrapper 恢复 `SimpleEdgeShadowOverlay` 及对应 import，并补一条回归测试，避免后续桌面表格重构再次覆盖移动端行为。

packages/kit/src/views/Perp/components/PerpMobileChartPanel.tsx
Minnzen: Intentional, from #12878: only the opt-in top chart prewarms, once, after `deferHeavyWorkUntilUIIdle`, so the first explicit reveal does not also pay the K-line fetch and Skia mount cost; the bottom panel still mounts only while expanded. A coin switch while collapsed replaces the single hidden chart rather than accumulating them. The PR description sentence was out of date and is updated to describe this.

development/perf-ci/thresholds/web.cold.json
Minnzen: Checked against the budget artifacts rather than the number alone. Swap cold graph, Perps module set: identical between an x-based baseline (run 33056301802) and this head — 30 modules, all pre-existing shared ones (FooterTicker, Portfolio, hooks); `forbiddenModulesFound` is empty on both. No Perps chart-panel module enters the swap graph, so there is no lazy boundary to restore.

What grew is the shared TradingViewNative renderer, which the Market/Swap chart uses by design: this PR's compact geometry adds ~27 KB on the swap path (head 17,500,295 vs x baseline 17,472,581), on top of x's own recent chart-type/time-axis work that already leaves x itself only 2.6 KB under the old budget.

So the question is whether +27 KB of shared renderer code is an acceptable new baseline. Leaving this thread open for that call; if approved, the budget should also get the usual ~12 KB headroom (17,512,000) instead of the 3.7 KB it has now, otherwise the next x merge trips it again.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Kahnchan: 重新核对了当前实现和 iOS 模拟器。当前移动端 Position、Order、Balance 分别走 PositionRowMobile、移动端 ListItem 和 BalanceRowMobile 的卡片布局，CommonTableListView 的 mobile 分支也没有横向 ScrollView，不存在右侧隐藏列。SimpleEdgeShadowOverlay 在 Native 上只是无条件绘制一条 1px 竖线，并不会根据 overflow 或滚动状态显示，因此这里所说的“提示右侧还有列”的影响在当前布局中不成立。

基于当前移动端卡片设计，这里暂不恢复该装饰线。如果有新的设计稿明确要求保留这条边缘分隔线，请提供对应依据，我们再按设计补回。

originalix (approved):

PR #12854, opened 2026-08-12
fix: stabilize mobile chart height OK-58198 OK-58394

limichange:
[OK-58198](https://onekeyhq.atlassian.net/browse/OK-58198) [OK-58394](https://onekeyhq.atlassian.net/browse/OK-58394)

----------

## Summary

- persist the Market Detail TradingView sub-indicator count in the App on iOS and Android
- use the App-persisted count as the initial chart height before the WebView restores its own layout
- ignore provisional WebView counts until TradingView confirms layout restoration, then use the restored count as calibration/fallback
- keep App counts isolated by the Web layout namespaces `market` and `market-hyperliquid`
- retain the quiet-window fallback for older TradingView Web builds that do not send the restoration flag

## Two-copy behavior

1. App copy (native MMKV-backed Jotai atom): primary value for the initial mobile chart height.
2. TradingView Web copy (saved study template/layout): fallback and calibration after restoration completes.

The App also remembers Hyperliquid chart-to-namespace classification so repeat entries select the correct cached height before the async Market basic config returns. The first-ever entry can still adjust once after Web layout restoration.

## Root cause

The App did not have an independent persisted initial count, and the bridge could not distinguish provisional chart-control snapshots from the final restored layout. Market Detail therefore resized for intermediate counts during WebView initialization.

The review follow-up also makes the restored marker and indicator count come from the same config snapshot, preserves existing layouts with more than four sub-indicators, and cancels legacy fallback timers after the marker protocol is detected.

## Cross-platform scope

- iOS and Android `main` UI runtime only
- native MMKV is the persisted backing resource; the background JS runtime does not consume this atom
- TradingView Native keeps its existing fixed initial count and does not write the WebView cache
- desktop and web behavior is unchanged

## Companion PR

- https://github.com/OneKeyHQ/tradingview-charting-library/pull/209

## Validation

- five focused Jest suites covering config/count synchronization, namespace persistence, marker fallback, restored counts, and native indicator controls (20 tests passed)
- `yarn agent:check --profile commit`
- `yarn agent:check --profile pr` (all local checks passed)
- final iOS/Android simulator verification pending by requester

Issues: OK-58198, OK-58394

[OK-58198]: https://onekeyhq.atlassian.net/browse/OK-58198
[OK-58394]: https://onekeyhq.atlassian.net/browse/OK-58394

PR #12859, opened 2026-08-13
fix: unify app QR codes to dot style (OK-59643)

franco-chan:
[OK-59643](https://onekeyhq.atlassian.net/browse/OK-59643)

----------

## What

Unifies every QR code in the app to the dot style and fixes the rendering defect that made dot codes unscannable.

### Root cause

The dot renderer read the matrix transposed (`[x][y]` instead of `[y][x]`), mirroring the symbol along its main diagonal. The three finder patterns map onto themselves under a transpose, so the result still looked like a valid QR while the data region was mirrored — scanners without mirror support could not read it at all. This defect is why dot was believed "hard to scan".

### Changes

- fix the transposed dot matrix, with regression tests that go red if it comes back
- default draw style flips to dot; the `line` branch stays as the one-line rollback exit
- `IQrcodeDrawType` narrowed to `'dot' | 'line'`; animation is inferred from `valueUr` (an air-gap UR is inherently multi-frame), and the `ShowAirGapQrcode` payload slims down accordingly
- the whole dot field renders as a single `<Path>` — one node instead of ~900 circles per 500ms air-gap frame
- white plate corners rounded to match the card, clamped so the smallest share codes stay inside the arc
- logo clear area is a disc so the round plate no longer reads as square, and the logo-to-plate gap is tightened
- per-instance clipPath ids (clip ids are document-global on web; two codes on one screen used to collide)
- web/desktop share images (Receive / Perp / RookieGuide) draw the same dot geometry on canvas through a shared helper, replacing the old `toDataURL` bitmap round-trip
- receive page network badge ring stays plate-colored in dark mode

### Verification

- OneKey Pro hardware A/B/C test on the air-gap animated code: line ✓ / transposed dot ✗ / fixed dot ✓, same scan speed
- 13 unit tests on the shared geometry (transpose regression, plate radius clamp, disc clear, single-path structure, finder rings)
- manual walkthrough of all nine QR surfaces (light/dark, share-image rephotograph, Pro pairing + signing flow)

Jira: https://onekeyhq.atlassian.net/browse/OK-59643

[OK-59643]: https://onekeyhq.atlassian.net/browse/OK-59643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/components/src/content/QRCode/index.tsx
```diff
@@ -339,35 +256,40 @@ export function QRCode({
...
   const scaledLogoMargin =
     (props.logoMargin ?? DEFAULT_LOGO_MARGIN) * symbolScale;
+  const plateBorderRadius = getQRCodePlateBorderRadius(quietZoneSize);
   return (
     <Theme name="light">
       <Stack
         width={canvasSize}
         height={canvasSize}
         bg="$bgApp"
+        borderRadius={plateBorderRadius}
```
sidmorizon, line 340: **Rounded white backdrop pokes out of tightly clipped QR containers, showing background wedges at the corners**

Severity: `non-severe`

The white backdrop behind every QR code is now given a 16px corner rounding (`borderRadius={plateBorderRadius}` at `packages/components/src/content/QRCode/index.tsx:285`) regardless of the container it is placed in, so screens that clip the code inside a smaller-radius card show the card background bleeding through at the four corners.
Impact: The deposit QR block looks visually broken, with dark notches at each corner in dark mode.

<details>
<summary>Geometry: 16px plate radius inside an 8px clipped card</summary>

`packages/kit/src/views/Perp/components/TradingPanel/modals/UnifoldDeposit/UnifoldDepositQRCard.tsx:116-140` wraps the code in a `Stack` of `QR_CARD_SIZE = 202` with `borderWidth={1}`, `borderRadius="$2"` and `overflow="hidden"`. The QRCode canvas is `size + padding = 190 + 10 = 200`, i.e. it exactly fills the card interior. With `quietZoneModules = 2` the quiet zone is ~10px, so `getQRCodePlateBorderRadius` (`packages/components/src/content/QRCode/QRCode.utils.ts:109-114`) returns the full `QR_CODE_PLATE_BORDER_RADIUS = 16`. Because 16 > the card's `$2` radius, the area between the plate arc and the card arc is uncovered at each corner. Previously the plate was square and filled the card interior exactly.

Possible fixes: expose the plate radius as a prop the caller can override (or set it to 0 there), or align the Unifold card radius with the plate radius.

</details>

packages/kit/src/utils/qrCodeCanvas.ts
```diff
@@ -0,0 +1,114 @@
...
+        );
+        ctx.fill();
+      },
+    );
+
+    // one path for every dot: they share a fill and never overlap, so a single
+    // fill() is equivalent to one per module and far cheaper
+    ctx.fillStyle = darkColor;
+    ctx.beginPath();
+    getQRCodeDotCells({ matrix }).forEach(({ x, y }) => {
```
sidmorizon: # [🟡 P2] canvas 渲染器缺少 logo clear-arena，中心底板会切出半月形残点

## 问题

`getQRCodeDotCells({ matrix })` 在这里没有传 `clearArenaModules`，而且 `drawDotQRCodeOnCanvas` 的签名也没有暴露这个参数。ReceiveShare 的 web/desktop 分享图（`ShareImageGenerator.tsx` 约 406-412 行）在调用本函数之后，会在码中心再画一个 `qr.logoPlateSize`（48px，QR 本体 192px）的白色圆盘盖住 token logo 区域。

组件版渲染器对同一场景专门做了圆盘清空（`getQRCodeLogoClearArenaSize` 的注释原话："anything closer would be sliced by the plate edge"），canvas 版却没有等价物。

## 影响

圆心距离落在 `plateR ± dotRadius`（cellSize ≈ 5.8px 时约 ±1.9px）环带内的 dot 会被底板边缘切成半月形，在导出的分享图上绕 logo 底板一圈可见残点。这与 PR 目标"分享图与 app 内使用同样的 dot 几何"直接相悖：app 内 Receive 码干净，导出的分享图却有毛边。

## 建议

给 `drawDotQRCodeOnCanvas` 增加可选 `clearArenaModules`（或 `logoPlateSize`）参数并透传给 `getQRCodeDotCells`，ReceiveShare 按组件版同样的公式换算后传入：

```ts
// drawDotQRCodeOnCanvas options
clearArenaModules?: number;
// ...
getQRCodeDotCells({ matrix, clearArenaModules })

// ReceiveShare 调用处
clearArenaModules:
  qr.logoPlateSize / (qr.size / matrixSize) + QR_CODE_DOT_RADIUS_RATIO * 2,
```

（仅在 `tokenLogoImg` 成功加载、确定会画底板时传入，避免 logo 加载失败时中心留下空洞。）

[Claude session ↗](https://claude.ai/code/session_01HNhLFmvbeW19hE71bsW283)

---

packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.tsx
```diff
@@ -364,24 +341,13 @@ export const ShareImageGenerator = forwardRef<
...
-              dark: qrCode.color,
-              light: '#FFFFFF',
-            },
+          drawDotQRCodeOnCanvas(ctx, {
+            value: referralUrl,
+            x: qrCodeX,
+            y: qrCodeY,
+            size: qrCode.size,
+            darkColor: qrCode.color,
           });
```
sidmorizon: # [🟠 P1] Rookie 分享图 88px QR 改为 dot 后墨量骤降，可扫性存在实际回归风险

## 问题

Rookie 分享画布是 `CANVAS_SIZE = 640` 的 1:1 画布（没有 ReceiveShare 那样的 `CANVAS_SCALE = 2` / `ctx.scale`），`qrCode.size = scale(88)` 即 QR 本体只有 `88px`。`referralUrl` 在默认 `ecl: 'M'` 下约是 version 2–3（25–29 模块），cellSize 只有 `3.0–3.5px`。

改动前 `QRCodeUtil.toDataURL` 输出的是整格实心方块（每个暗模块 100% 墨量）；改成 dot 几何后每个暗模块只剩直径 `2/3` cellSize ≈ `2.0–2.3px` 的抗锯齿圆点，墨量降到约 `35%`（π/9），同尺寸下 finder 三环也因 `radius: Math.min(20, size/2)` 全部钳到半径上限、退化为同心圆。

## 影响

导出的 PNG 里 referral 码是 2px 级灰化圆点；经社交/聊天软件转发时普遍会被二次压缩或缩略（例如缩到 300px 时模块只剩约 1.4px，圆点直接糊掉，而实心方块在同等缩放下仍保留灰度块）。这是三处 canvas 分享图中分辨率最低的一处（Perp 为 900px 画布 cellSize ≈ 5.5px、Receive 为 2x 画布 cellSize ≈ 11.6px 设备像素），可扫性回归风险集中在这里。

## 建议

给 Rookie 分享图与 ReceiveShare 一致的 2x 超采样（`canvas.width = size * 2` + `ctx.scale(2, 2)`，导出尺寸不变则放大 `qrCode` 区域亦可）；或者在 `drawDotQRCodeOnCanvas` 里对 `cellSize` 低于阈值（如 4px）的情况回退为整格方块渲染，保证极小尺寸下的可扫性优先于风格统一。建议合入前用真机相机对导出（并经微信/Twitter 转发一次）的 Rookie 分享图实测扫码。

[Claude session ↗](https://claude.ai/code/session_01HNhLFmvbeW19hE71bsW283)

---

packages/components/src/content/QRCode/index.tsx
franco-chan: 已修（7e04f0ed7a）：给 Unifold QR 卡片补上与底板同色的白底，16px 底板圆角不再透出弹窗背景的深色缺角，恢复改版前观感。

packages/kit/src/utils/qrCodeCanvas.ts
franco-chan: 已修（7e04f0ed7a）：drawDotQRCodeOnCanvas 新增 clearPlateSize 参数（内部复用组件同一公式换算 clear arena），ReceiveShare 仅在 logo 加载成功、确定会画底板时传入。

packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.tsx
franco-chan: 已修（7e04f0ed7a）：Rookie 分享图对齐 ReceiveShare 的 2x 超采样（导出 1280px、QR 设备像素翻倍、墨量密度随之恢复）；合入前会按建议对转发后的图真机实测扫码。

packages/components/src/content/QRCode/QRCode.utils.ts
```diff
@@ -1,22 +1,78 @@
...
+
+export function ensureQRCodeUtilLoaded() {
+  if (!qrCodeUtilPromise) {
+    qrCodeUtilPromise = import('qrcode').then((mod) => {
+      // CJS/ESM interop shape differs across metro, vite and jest
+      loadedQRCodeUtil = (mod.default ?? mod) as IQRCodeUtil;
+    });
+  }
+  return qrCodeUtilPromise;
+}
```
sidmorizon: # [🟡 P2] 懒加载 import 失败后 rejected promise 被永久缓存，QR 全局静默消失且无法重试

## 问题

`ensureQRCodeUtilLoaded` 把 `import('qrcode')` 的 promise 缓存进 `qrCodeUtilPromise`，但失败时没有任何处理：promise reject 后 `if (!qrCodeUtilPromise)` 仍为假，缓存的 rejected promise 会被永久复用，后续所有调用都直接拿到同一个 rejection，永远不会重试。

在 web/extension 上 `import('qrcode')` 会编译成异步 chunk 的网络/文件加载，瞬时失败是真实场景（web 弱网、extension 升级后旧页面引用已被替换的 chunk 文件）。同时组件侧的消费方式是 `void ensureQRCodeUtilLoaded().then(() => setIsEncoderReady(true))`（`index.tsx` 第 232-236 行），没有 `.catch`：reject 时 `isEncoderReady` 永远为 false。

## 影响

一次瞬时加载失败会让该 JS runtime 内所有 QR 场景永久静默失效直到整页重载：Receive 页、air-gap 签名 toast、Prime 配对码全部 `return null` 不渲染且无任何报错 UI，外加一个 unhandled promise rejection（Sentry 噪音）。canvas 分享图路径虽有 try/catch，但同样因缓存的 rejection 而永远画不出 QR。

## 建议

失败时清空缓存槽，让下一次渲染/调用重试；组件 effect 补 `.catch`（保持 `isEncoderReady` 为 false 即可，重试由下一次 mount/render 触发）：

```ts
export function ensureQRCodeUtilLoaded() {
  if (!qrCodeUtilPromise) {
    qrCodeUtilPromise = import('qrcode').then(
      (mod) => {
        loadedQRCodeUtil = (mod.default ?? mod) as IQRCodeUtil;
      },
      (error) => {
        // drop the failed attempt so the next caller retries the import
        qrCodeUtilPromise = undefined;
        throw error;
      },
    );
  }
  return qrCodeUtilPromise;
}
```

[Claude session ↗](https://claude.ai/code/session_01HNhLFmvbeW19hE71bsW283)

---

franco-chan: 已修（c6f4cb8e5b）：加载失败时清空缓存槽并向上抛出，组件 effect 补第二参数处理 rejection（保持 unready，下一次 mount 重试）；canvas 路径经调用方 try/catch 自然获得重试。

packages/components/src/content/QRCode/index.tsx
```diff
@@ -298,21 +221,32 @@ export function QRCode({
...
+    if (!isEncoderReady) {
+      ensureQRCodeUtilLoaded().then(
+        () => setIsEncoderReady(true),
+        () => {
+          // stay unready; the loader dropped its cache so the next mount
+          // retries the import
+        },
+      );
+    }
+  }, [isEncoderReady]);
```
sidmorizon, line 272: **A QR code can stay permanently blank if its encoder fails to load once**

Severity: `non-severe`

The screen gives up on loading the code generator after a single failed attempt (`ensureQRCodeUtilLoaded()` at `packages/components/src/content/QRCode/index.tsx:236`) and never tries again while it stays open, so the QR code area silently shows nothing.
Impact: On a flaky network (or right after an extension/web update swaps chunks), users see an empty space where the receive/pairing QR code should be, with no error or retry, until they leave the page and come back.

<details>
<summary>Why the retry never happens: the effect is gated on a state flag that stays false</summary>

`packages/components/src/content/QRCode/QRCode.utils.ts:16-33` deliberately clears `qrCodeUtilPromise` on rejection so a *later* caller retries. But the only consumer, the effect at `packages/components/src/content/QRCode/index.tsx:234-244`, depends solely on `isEncoderReady`; on failure the rejection handler intentionally leaves the state `false`, so the effect never re-runs and this instance never calls `ensureQRCodeUtilLoaded()` again. The render then hits `if (!displayValue || !isEncoderReady) return null;` (`packages/components/src/content/QRCode/index.tsx:282-285`) forever — no skeleton, no error. Previously `qrcode` was a static import, so this failure mode did not exist.

A bounded retry (e.g. re-attempt on a timer/AppState focus, or surface an error state instead of `null`) would restore the old guarantee.

</details>

franco-chan: 已修（d9c9802768）：effect 内改为指数退避持续重试（2s 起、上限 30s，卸载即取消）——页面开着时网络恢复即自动出码；chunk 被更新替换的永久失败场景则安静退避，不刷请求。

packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.tsx
```diff
@@ -174,8 +155,11 @@ export const ShareImageGenerator = forwardRef<
       referralPill,
       spacing,
     } = CANVAS_CONFIG;
-    canvas.width = size;
-    canvas.height = size;
+    // setting width/height resets the context state, so the scale below never
+    // compounds across repeated generate() calls
+    canvas.width = size * CANVAS_SCALE;
+    canvas.height = size * CANVAS_SCALE;
+    ctx.scale(CANVAS_SCALE, CANVAS_SCALE);
```
sidmorizon, line 162: **Drop shadow on the beginner-guide share image comes out half as large as intended**

Severity: `non-severe`

The share picture is now drawn at double resolution (`ctx.scale(CANVAS_SCALE, CANVAS_SCALE)` at `packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.tsx:162`) without doubling the soft shadow behind the card, so the shadow ends up half its intended size relative to everything else.
Impact: The shared beginner-guide image no longer matches the on-screen card: the card's drop shadow looks noticeably tighter and tighter-offset than designed.

<details>
<summary>Canvas shadow parameters are device-space, not user-space</summary>

Per the canvas specification, `shadowOffsetX/Y` and `shadowBlur` are applied in the output bitmap's coordinate space and are explicitly *not* transformed by the current transformation matrix. Every other quantity in `generate()` (positions, font sizes, radii) is expressed in logical units and is scaled by the new `ctx.scale(2, 2)`, but the shadow set at `packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.tsx:206-209` from `card.shadowBlur` (24) and `card.shadowOffsetY` (4) in `packages/kit/src/views/RookieGuide/components/RookieShare/constants.ts:30-32` stays at 24/4 device pixels while the card itself doubles to 2x device pixels. When the resulting 1280x1280 PNG is viewed at its logical 640 size, the shadow reads as blur 12 / offset 2.

The fix is to multiply `shadowBlur`/`shadowOffsetY` (and `shadowOffsetX` if used) by `CANVAS_SCALE` when the context is scaled.

</details>

franco-chan: 已修（07143b1c95）：shadowBlur / shadowOffsetY 按 canvas 规范手动乘 CANVAS_SCALE（device-space 参数不吃 ctx.scale），阴影恢复与放大前一致的相对大小。

franco-chan:
Addressed the P1 from the auto-review report in `394796cb6e`:

- `QRCode` now exposes `onRenderReady`, fired once after the commit in which a symbol is actually drawn (i.e. after the lazily-imported encoder has loaded and rendered).
- All three native ViewShot generators gate capture on it through their existing readiness mechanisms: Receive and RookieGuide count the QR as one more expected ready signal in their `onImagesReady` counters; the Perps generator — which previously captured immediately — now adopts the same deferred + 5s-timeout gate as Receive, with the QR, background, and token icon all counted.
- The timeout fallback means a permanently failed encoder load degrades to today's behavior (capture without waiting) instead of hanging the share action.

On the suggested regression test: exercising ViewShot capture order isn't feasible in jest, and rendering `QRCode` under jsdom drags in react-native-svg + tamagui setup; the shared geometry stays covered by the existing 13 unit tests, and capture readiness is covered by manual validation (fresh-runtime share right after app start) per the report's validation-gaps list.

originalix (dismissed):

packages/kit/src/views/Perp/components/PositionShare/ShareContentRenderer.tsx
```diff
@@ -119,15 +119,26 @@ export function ShareContentRenderer({
...
     if (expectedImageCount.current === 0 && onImagesReady) {
       onImagesReady();
     }
-  }, [selectedBackground, onImagesReady]);
+  }, [
+    selectedBackground,
+    display.showTokenIcon,
+    showsReferralQrCode,
+    onImagesReady,
+  ]);
```
sidmorizon: # [🟠 P1] isReferralReady 翻转会重置计数器但一次性信号不会重发，Perp native 截图门控在常见时序下退化

## 问题

这个 effect 在 `showsReferralQrCode` 翻转时把 `imageLoadCountRef.current` 清零并把 expected 提高到 3，但三个 ready 信号都是一次性的：背景/币种 `Image` 的 `onLoad` 不会因 effect 重跑而重发，`QRCode` 的 `onRenderReady` 也被 `hasFiredRenderReadyRef` 限定只触发一次。`useReferralUrl` 的 `isReady = Boolean(inviteCode) || isLoading === false` 要经过一次 backgroundApiProxy 往返才翻真，所以「渲染器已挂载、图片已开始加载、referral 稍后就绪」是常规时序，不是边角。

同时 `ShareImageGenerator.native.tsx` 里的 `imagesReadyDeferredRef` 是一次性的，resolve 后不会因 expected 变化重新武装。

## 影响

按翻转与图片加载的相对顺序，两种失效各占其一：

- 任一图片先于翻转 `onLoad`（RN 图片缓存命中时几乎必然）：清零丢掉已发信号，计数最多到 `2 < 3`，`onImagesReady` 永远不触发，之后每次 native 分享都要吃满 `IMAGES_READY_TIMEOUT_MS = 5000` 的兜底超时才截图——保存/分享/转发每次 +5s。
- 图片全部先就绪、deferred 已提前 resolve：翻转后 QR 才挂载，此时 `generate()` 不再等待，快速点击分享仍可能截到没有 QR 的图——正是本 commit 要堵的场景在 referral 路径上留了口子。

Receive/Rookie 不受影响（前者从不重置，后者只按静态 `imageUrl` 重置）。

## 建议

把「可重置的计数器 + 一次性信号」改为「按来源集合判定」：用 `Set<string>` 记录已就绪来源（`'background' | 'tokenIcon' | 'qrCode'`），expected 集合随 props 派生，每次插入后检查 `expected ⊆ ready` 再触发 `onImagesReady`（允许多次触发）；native 侧把 deferred 换成在 `generate()` 时读取当前就绪状态（或允许 re-arm）。最小改动版：effect 里不要清零 `imageLoadCountRef`，并让 `handleImageLoad` 按来源去重，即可同时消除两个失效路径。

[Claude session ↗](https://claude.ai/code/session_01HNhLFmvbeW19hE71bsW283)

---

franco-chan: Fixed in `53a90125cd`, following the source-set suggestion: the renderer now tracks ready sources in a `Set` ('background' / 'tokenIcon' / 'qrCode') with the expected set derived from props — nothing is ever zeroed, so one-shot signals survive expectation changes. The renderer reports a boolean via `onImagesReadyStateChange` (re-firing `false` when the referral QR joins the expectation late), and the native generator reads the latest state at `generate()` time instead of awaiting a one-shot deferred — closing both failure paths: no more full 5s timeout on cached images, and no capture window without the QR after the referral flip.

packages/components/src/content/QRCode/QRCode.utils.ts
```diff
@@ -1,22 +1,87 @@
...
+// The encoder is ~70KB across 27 modules, and this module is reachable from
+// the app's startup graph through the components barrel. Loading the library
+// behind an async edge keeps it out of the native startup bundle; render
+// paths gate on ensureQRCodeUtilLoaded() before calling the sync helpers.
+let loadedQRCodeUtil: IQRCodeUtil | undefined;
+let qrCodeUtilPromise: Promise<void> | undefined;
+
+export function ensureQRCodeUtilLoaded() {
+  if (!qrCodeUtilPromise) {
+    qrCodeUtilPromise = import('qrcode').then(
```
sidmorizon: # [🟠 P1] 懒加载 chunk 占满 ext JS 文件预算，merge x 后 `721 > 720` 持续阻塞 auto-merge

## 问题

`import('qrcode')` 让 rspack 为 ext 产物多切出一个异步 chunk。实测数据：在 `07143b1c`（本 PR 含此动态 import，基于旧 x tip `0bd2c392`）上 `Extension build budget` 绿，摘要为 `"jsFiles": 720` —— 恰好顶满 `EXT_BUILD_MAX_JS_FILES` 的默认预算 720（`apps/ext/scripts/check-build-output.js:256` 硬编码，env 可覆盖但 CI 未设）。merge x 带入 #12761（`ff01f13e`）又新增 1 个 JS 文件后，`7eb1bbba` 上变为 `721 > 720`（`check-build-output.js:269` 抛错）。

两个 PR 各自 +1、单独都能过 720，合流后越线——典型 ratchet 碰撞，且 CI 日志只报 `721 > 720`，不指明多出的文件与原因。

## 影响

该失败是确定性的：重跑、重新 merge x 都不会变绿，auto-merge 会被持续卡住直到本 PR 内消化这 +1。

## 建议

二选一，由作者权衡：

1. 按 ratchet 惯例把默认预算上调 1（两处 +1 都是有意引入：本 PR 的启动图懒加载收益、#12761 的新页面）：

```js
// apps/ext/scripts/check-build-output.js
jsFiles: readBudget('EXT_BUILD_MAX_JS_FILES', 721),
```

2. 若不想动预算，仅在 ext 保留静态加载（放弃 ext 端的懒加载收益，native 启动图收益不受影响）：`import(/* webpackMode: "eager" */ 'qrcode')` 会内联进引用方 chunk、不产生新文件——但注意这会同时作用于所有平台的 web 打包器，需确认 native metro 收益路径不受影响后再选。

方案 1 改动最小且语义诚实，推荐。

[Claude session ↗](https://claude.ai/code/session_01HNhLFmvbeW19hE71bsW283)

---

franco-chan: Took a refined version of option 2 in `dcfc78bcc5` (pushed before this comment landed): the loader is now platform-split — `qrCodeUtilLoader.native.ts` keeps the async edge (the native startup-graph win stays), while `qrCodeUtilLoader.ts` bundles the encoder statically for web-family targets. That avoids the cross-platform ambiguity you flagged for `webpackMode: "eager"` (metro never sees the static path), removes the extra ext chunk entirely, and web regains its synchronous first paint as a bonus. Expected `jsFiles` returns to 720; the run on `dcfc78bcc5` is verifying now. Left the budget at 720 — if x's own growth crosses the line later, raising `EXT_BUILD_MAX_JS_FILES` per the ratchet convention would then be the honest move, but this PR no longer needs it.

packages/kit/src/views/RookieGuide/components/RookieShare/ShareContentRenderer.tsx
```diff
@@ -51,12 +51,15 @@ export function ShareContentRenderer({
...
+  // so a ViewShot capture must not run before the symbol is drawn
+  const expectedReadyCount = SHARE_IMAGE_COUNT + (referralUrl ? 1 : 0);
   const handleImageLoad = useCallback(() => {
     imageLoadCountRef.current += 1;
-    if (onImagesReady && imageLoadCountRef.current >= SHARE_IMAGE_COUNT) {
+    if (onImagesReady && imageLoadCountRef.current >= expectedReadyCount) {
       onImagesReady();
     }
-  }, [onImagesReady]);
+  }, [onImagesReady, expectedReadyCount]);
```
sidmorizon: **Rookie share image capture stalls for five seconds before producing the picture**

Severity: `non-severe`

The QR code's "I'm drawn" signal is counted (`onRenderReady={handleImageLoad}` at `packages/kit/src/views/RookieGuide/components/RookieShare/ShareContentRenderer.tsx:244`) before the tally of finished pieces is zeroed on mount, so the signal is erased and the share picture is only produced after a five-second give-up timer.

Impact: On every share after the first one in a session, the user waits ~5 extra seconds for the Rookie share image, and the capture may run against a not-yet-finished card.

<details>
<summary>Effect ordering: child QR effect fires before the parent's counter reset</summary>

`packages/components/src/content/QRCode/index.tsx:312-317` fires `onRenderReady` from a passive effect as soon as `hasRenderedSymbol` is true. Once the encoder has been loaded earlier in the session, `isQRCodeUtilLoaded()` makes `isEncoderReady` true on the very first render (`packages/components/src/content/QRCode/index.tsx:237`), so `onRenderReady` fires during the mount effect phase.

React flushes passive effects child-first, so the nested `QRCode` effect runs before `ShareContentRenderer`'s own `useEffect(() => { imageLoadCountRef.current = 0; }, [imageUrl])` (`packages/kit/src/views/RookieGuide/components/RookieShare/ShareContentRenderer.tsx:64-66`). The counter is incremented to 1 and then reset to 0. Because `onRenderReady` is guarded by `hasFiredRenderReadyRef` it never fires again, so the counter can only ever reach `SHARE_IMAGE_COUNT` (2) while `expectedReadyCount` is 3, and `onImagesReady` is never called.

`packages/kit/src/views/RookieGuide/components/RookieShare/ShareImageGenerator.native.tsx:84-89` then always waits out `IMAGES_READY_TIMEOUT_MS` (5000ms) before capturing.

The previously-existing ready sources were all image `onLoad`/`onError` callbacks, which are inherently asynchronous and always land after the reset effect — the new synchronous-on-mount QR signal is what breaks the invariant.

</details>

franco-chan: Fixed in `8f0214c904` by giving Rookie the same source-set treatment as the Perps renderer: ready sources live in a named `Set` ('badge' / 'logo' / 'qrCode') that is never cleared wholesale, so the QR's child-first mount-phase signal survives the parent's effects. The old `[imageUrl]` reset becomes a precise retraction of just the 'badge' source. All three share renderers are now immune to the effect-ordering erase.

packages/components/src/content/QRCode/QRCode.utils.ts
franco-chan: Follow-up with ground truth: local builds settled it. On `8f0214c904` alone the platform-split loader did bring the count back to exactly 720 (verified locally, `jsFiles: 720`, passing) — but the CI job builds the PR *merge ref*, and x had moved again (#12854 adds one more file), so CI kept seeing 721. Since the branch was already at the line with zero removable files of ours, I took your option 1 after all: merged the latest x and raised the default to 721 in `05f8f4a706`, with the arithmetic documented next to the budget. Local post-merge build verifies `jsFiles: 721` ≤ 721, matching what the merge ref will produce.

packages/kit/src/views/Perp/components/PositionShare/ShareContentRenderer.tsx
```diff
@@ -106,28 +110,57 @@ export function ShareContentRenderer({
...
-    imageLoadCountRef.current = 0;
-    expectedImageCount.current = 0;
-    if (selectedBackground) expectedImageCount.current += 1;
-    if (display.showTokenIcon) expectedImageCount.current += 1;
-    if (expectedImageCount.current === 0 && onImagesReady) {
-      onImagesReady();
-    }
-  }, [selectedBackground, onImagesReady]);
+    evaluateReadiness();
+  }, [evaluateReadiness]);
```
sidmorizon, line 167: **Shared position image can be captured with the previous background picture**

Severity: `non-severe`

The background picture is still counted as already-drawn (`readySourcesRef` at `packages/kit/src/views/Perp/components/PositionShare/ShareContentRenderer.tsx:118`) after the card switches to a different background, so the saved image can be captured while the old or blank background is still on screen.
Impact: A shared profit/loss card can be exported with the wrong background artwork.

<details>
<summary>Loss of the readiness retraction that the old counter performed</summary>

The previous implementation reset the counter whenever `selectedBackground` changed (`imageLoadCountRef.current = 0; expectedImageCount.current = 0;` in the removed effect), so a background swap (profit ↔ loss, driven by `pnl`) forced the gate back below its target until the new `<Image>` fired `onLoad`.

The new set-based gate only ever *adds* sources (`handleSourceReady` at `packages/kit/src/views/Perp/components/PositionShare/ShareContentRenderer.tsx:138-144`). When `selectedBackground` changes, `expectedSources` still contains `'background'` and `readySourcesRef` still contains `'background'` from the previous image, so `evaluateReadiness()` immediately reports `true` and `ShareImageGenerator.native`'s `waitForImagesReady()` resolves at once (`packages/kit/src/views/Perp/components/PositionShare/ShareImageGenerator.native.tsx:59-66`), letting `viewShot.capture()` run before the new bitmap is decoded.

The sibling RookieShare renderer handles exactly this case by retracting the single affected source when its URL changes (`readySourcesRef.current.delete('badge')` in `packages/kit/src/views/RookieGuide/components/RookieShare/ShareContentRenderer.tsx:92-94`); the Perp renderer has no equivalent for `selectedBackground` or `tokenImage`.

</details>

franco-chan: Fixed in `a7f320ee8d`: image sources now carry their URI in the ready key (`background:<uri>`, `tokenIcon:<uri>`). A background swap changes the expected key, so the stale image's ready mark stops satisfying the gate by derivation — no retraction effect that could misfire on unrelated dependency churn — and the existing expected-change effect pushes the regressed `false` state to the native generator, which re-gates the next capture until the new image's `onLoad` lands. Stale late `onLoad`s from the old image add an old key harmlessly.

originalix (approved):

PR #12872, opened 2026-08-13
fix: inject JPUSH_KEY into native bundle builds

huhuanming:
## Summary

- pass `JPUSH_KEY` from `release-app-bundles` into both reusable native bundle jobs
- expose the secret to the native bundle build step
- fail fast when `JPUSH_KEY` is missing instead of publishing an unusable OTA bundle

## Root cause

The standalone native bundle workflow built the background JavaScript runtime without receiving `JPUSH_KEY`. Babel therefore inlined an empty value into `PushProviderJPush`, while EAS-based iOS and Android releases received the key through their separate EAS environment.

## Impact

Native OTA bundles can no longer silently ship with an empty JPush AppKey. The GitHub Actions repository or organization secret named `JPUSH_KEY` must be configured and made available to this repository.

## Validation

- `actionlint` on the modified workflows
- local iOS native bundle built with a non-sensitive sentinel and verified in both source JavaScript and the final Hermes bytecode
- `release-app-bundles` run `31689419649` confirmed the new guard executes; it currently fails as expected because the repository does not yet expose a `JPUSH_KEY` Actions secret

### DIFF
--- a/.github/workflows/release-app-bundles.yml
+++ b/.github/workflows/release-app-bundles.yml
@@ -290,6 +290,7 @@ jobs:
       checkout_ref: ${{ needs.prepare-params.outputs.commit }}
     secrets:
       COVALENT_KEY: ${{ secrets.COVALENT_KEY }}
+      JPUSH_KEY: ${{ secrets.JPUSH_KEY }}
       SENTRY_TOKEN: ${{ secrets.SENTRY_TOKEN }}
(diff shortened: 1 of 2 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #12876, opened 2026-08-14
fix: improve Swap and Market review flows (OK-60101, OK-58779, OK-59976, OK-60022)

zhaono1:
[OK-60101](https://onekeyhq.atlassian.net/browse/OK-60101) [OK-58779](https://onekeyhq.atlassian.net/browse/OK-58779) [OK-59976](https://onekeyhq.atlassian.net/browse/OK-59976) [OK-60022](https://onekeyhq.atlassian.net/browse/OK-60022)

----------

## Summary

- Preserve canonical mixed-case token symbols in Swap quote-rate displays.
- Use authoritative backend token search for Swap and Limit, including the Swap support-list query for ordinary keyword searches.
- Move Market detail trading from the deprecated speed-quote path to the unified Swap quote/build/review flow while preserving speed-config presets.
- Support Review Order slippage editing and transaction rebuilding for both Swap and Market owners.
- Close quote lifecycle, stale-state, cache, provider-store, fallback, and polling gaps found during review.

### Token symbol casing and search

- Preserve original token symbol casing for display and normalize only comparison paths.
- Remove cached local keyword filtering as the final authority.
- For ordinary Swap/Limit keyword searches, merge the general token endpoint with the Swap support-list endpoint and deduplicate by network and address.
- Preserve LP-token and account-token behavior outside this search path.

### Market unified Swap flow

- Remove the deprecated Market speed-quote service and transaction-parameter builder.
- Reuse unified Swap quote, build, account, fee, review, and submit state from Market detail.
- Keep speed-config presets and pass the Market trade source as `swap`.
- Freeze the quote/input identity used by Preview and reject stale async results.
- Scope quote EventSource ownership by request ID and close active streams on reset/unmount.
- Keep Market information tabs stable and avoid duplicate holders, transaction, polling, and price requests.
- Bind holders cache to both network and address.

### Review Order slippage editing

- Show editable slippage only when the build response reports `supportRebuildTx`.
- Use a compact popover on Desktop and an in-dialog editor on mobile.
- Support saving for this order or all future orders.
- Let each owner persist future defaults: Swap uses global settings; Market updates its active preset/direction setting or the shared global setting when preset mode is disabled.
- Rebuild the executable transaction and network-fee estimate before closing the editor.
- Clear stale fallback state after a successful non-fallback rebuild and surface local rebuild failures.
- Preserve the active Swap Pro store when opening the provider selector.

## Review feedback

- Addressed the initial seven valid inline findings in e638f6e76e and three follow-up findings in 800bacc789.
- Verified that the US market-status finding does not apply: `utility/v2/market/token/detail` already derives `display.isOpen` from the server holiday/early-close calendar and maps it to `stock.isOpen`; the deprecated client status endpoint should remain removed.
- Replied to and resolved all eleven inline review threads; declined two over-defensive findings with DTO and server-contract evidence.

## Issues

- Fixes OK-60101
- Fixes OK-58779
- Fixes OK-59976
- Fixes OK-60022

## Test Plan

- `yarn agent:check --profile commit` — passed lint, formatting, agent context, staged lint, and staged TypeScript checks.
- Focused Swap/Market regression runs — initial 12 suites and 114 tests passed; follow-up 8 suites and 83 tests passed.
- `git diff --check origin/x` — passed.
- Desktop smoke in the repository Electron runtime: selected USDG, entered `0.0001 ETH`, received an OKX Dex quote, and opened Review Order successfully.
- Review Order showed matching pay/receive amounts, provider, `0.5%` slippage, network fee, and enabled confirmation; no transaction was submitted.
- Verified reused slippage i18n keys exist in all locale bundles; no locale pull or generated translation change is required.

## Risk

- Slippage editing remains gated by server-provided `supportRebuildTx`; unsupported providers keep the existing read-only review.
- Current-order and all-future actions rebuild first; only the latter persists an owner-specific default.
- Quote and rebuild flows are protected by request IDs and quote-identity checks, but remain sensitive to backend quote/build contract changes.
- Swap/Limit keyword search uses authoritative server results and degrades to any successful token endpoint instead of discarding usable matches.
- The quote-rate component is shared by Swap, Bridge, and Stock; existing uppercase symbols remain visually unchanged.
- The branch is synchronized with `x`; renewed CI is the remaining merge-readiness gate.

[OK-60101]: https://onekeyhq.atlassian.net/browse/OK-60101?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-58779]: https://onekeyhq.atlassian.net/browse/OK-58779?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-59976]: https://onekeyhq.atlassian.net/browse/OK-59976?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60022]: https://onekeyhq.atlassian.net/browse/OK-60022?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix:

@zhaono1

Reviewed commit `800bacc7893a`. This report always reflects the latest reviewed changes.

### Review summary

This PR retires the dedicated Market/Swap Pro speed-quote path and routes Market swap plus native Swap Pro market orders through the shared swap quote-event, build-tx, and review-state pipeline. It also adds review-time slippage rebuilding, provider-selection reuse, and a set of Market-detail UI/state updates, so the main ownership change is that quote freshness, provider choice, and review rebuilding now converge on one shared foreground/background swap state machine instead of separate Market and Swap Pro implementations. That consolidation improves reuse, but two edge paths are still wrong in the new shared flow: wrapped Swap Pro pairs no longer get the local synthetic quote they depend on, and slippage rebuild can keep an old fallback step model even after the rebuild proves the batch path is available.

**What needs attention:** Re-check the shared quote/review state machine at the two edge transitions it now owns: native Swap Pro wrapped pairs, and slippage rebuilds that should move a review back from fallback to the normal batch path.

### Issues to address

- **P1 · Swap Pro wrapped pairs no longer seed a local executable quote**
  - **Impact:** `quoteSpeedAction()` now forwards every same-network Swap Pro market pair into the shared provider quote flow, but the old wrapped-pair shortcut is gone. Wrapped/native pairs still present as executable in the UI (`canExecuteInPro = supportSpeedSwap || isWrapped`), while the review entrypoint in `SwapMainLand` still requires `currentQuoteRes` before it opens review. Because no local 1:1 wrapped quote is created anymore, native Swap Pro wrap/unwrap can degrade into a visible "Wrap" action that never opens review or sends the transaction.
  - **Suggested action:** Restore the local wrapped-pair branch in `quoteSpeedAction()` by reusing `buildWrappedSwapQuoteResult()` for wrap/unwrap pairs, and keep the shared provider quote flow only for provider-backed market swaps.
- **P2 · Successful slippage rebuild can stay stuck in fallback review steps**
  - **Impact:** `rebuildSwapWithSlippage()` still rebuilds `separateSteps` whenever the old review had `preSwapData.shouldFallback`, even if the new fee estimate comes back with `fallbackToSeparateTxConfirm = false`. In that recovery path the state flips `shouldFallback` back to `false`, but the step list remains the fallback approve/send sequence from `buildSeparateApproveAndSwapSteps()`. The rebuilt review therefore does not actually return to the batch-capable execution model it just proved is available.
  - **Suggested action:** Make step regeneration depend on the current rebuild result, not the old fallback flag: when `shouldFallback` is false, rebuild the normal batch review steps instead of preserving the separate fallback step list.

### Validation gaps

- I checked the current PR comments, inline threads, and the previous runner report to avoid duplicating already-covered issues.
- I did not independently rerun Jest or `yarn agent:check` in this workspace.
- The PR author reported on August 15, 2026 that `yarn agent:check --profile commit` and 8 focused suites / 83 tests passed, but I did not verify that claim locally.
- I did not find new coverage for native Swap Pro wrapped-pair quoting or for fallback-to-batch recovery during slippage rebuild.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Market/MarketDetailV2/components/InformationTabs/components/TransactionsHistory/hooks/useMarketTransactions.ts
```diff
@@ -178,16 +179,19 @@ export function useMarketTransactions({
...
+      pollingInterval: timerUtils.getTimeDurationMs({ seconds: 5 }),
+    },
   );
 
+  useEffect(() => {
+    // Keep the initial REST snapshot, but pause the polling chain when the
+    // realtime transport is active. This avoids a second identical request
+    // when responsive/websocket state settles after the first render.
+    setStopPolling(!normalMode);
+  }, [normalMode, setStopPolling]);
```
sidmorizon: **Live transaction list starts polling the server every 5 seconds again after switching tokens**

Severity: `non-severe`

Polling is paused only once when the panel mounts (`setStopPolling(!normalMode)` at `packages/kit/src/views/Market/MarketDetailV2/components/InformationTabs/components/TransactionsHistory/hooks/useMarketTransactions.ts:192`) and is silently re-enabled whenever the viewed token changes, so the list resumes fetching every 5 seconds even while live updates are already streaming.
Impact: After navigating to another token, the app issues a redundant network request every five seconds on top of the realtime feed, wasting bandwidth and battery.

<details>
<summary>Why the pause is lost when the request dependencies change</summary>

`usePromiseResult` keeps the pause in a ref and explicitly clears it whenever its dependency list changes: the deps effect body does `stopPollingRef.current = false` before re-running (`packages/kit/src/hooks/usePromiseResult.ts:452-460`), and `runnerDeps` is `[tokenAddress, networkId, pollingInterval]`. The new pause effect in `useMarketTransactions` only depends on `[normalMode, setStopPolling]`, so it does not re-run when `tokenAddress`/`networkId` change and never re-applies `setStopPolling(true)`. The polling continuation in the `finally` block then keeps rescheduling every `pollingInterval` (`packages/kit/src/hooks/usePromiseResult.ts:369-385`), defeating the intended single-REST-snapshot behaviour in realtime mode.

</details>

Suggested fix:

```ts
  useEffect(() => {
    // Keep the initial REST snapshot, but pause the polling chain when the
    // realtime transport is active. This avoids a second identical request
    // when responsive/websocket state settles after the first render.
    // Re-apply on token/network changes: usePromiseResult clears the pause
    // flag whenever its request deps change.
    setStopPolling(!normalMode);
  }, [normalMode, setStopPolling, tokenAddress, networkId]);
```

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx
```diff
@@ -938,114 +1092,55 @@ export function useSpeedSwapActions(props: {
...
-            buildRes: reviewBuildRes,
-            encodedTx: undefined,
-            transferInfo: undefined,
-            swapInfo,
-            userAddress,
-          };
+        const {
+          actionState: currentQuoteActionState,
+          selectedQuoteResult: selectedQuote,
+        } = quoteExecutionStateRef.current;
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。Preview 开始时冻结 quote/action 快照；余额检查返回后同时校验当前 quote 身份和 review 条件，只使用同一份捕获报价构建交易，避免旧输入与新 provider/context 混用。

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx
```diff
@@ -2958,130 +3147,104 @@ export function useSpeedSwapActions(props: {
...
-      checkTokenApproveAllowance,
-    ],
-  );
+  useEffect(() => {
+    appEventBus.off(EAppEventBusNames.SwapQuoteEvent, quoteEventHandler);
+    appEventBus.on(EAppEventBusNames.SwapQuoteEvent, quoteEventHandler);
+    return () => {
+      appEventBus.off(EAppEventBusNames.SwapQuoteEvent, quoteEventHandler);
+      cleanQuoteInterval();
+    };
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。Market 面板 cleanup 会按 quoteRequestId 关闭当前 SSE 并重置报价状态；ServiceSwap 也改为按 requestId 管理并发连接，避免跨页面或跨 store 互相关闭。

packages/kit/src/views/Market/components/PerpsBadges.tsx
```diff
@@ -344,9 +337,8 @@ const StockIsOpenBadge = memo(
           source,
           isOpen,
           isPaused,
-          status: marketStatus,
         }),
```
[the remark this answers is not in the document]
zhaono1: 该评论不成立，无需恢复已废弃的 useUSMarketStatus 接口。当前 utility/v2/market/token/detail 的 display.isOpen 来自服务端 us-market-status（覆盖周末、节假日和提前收市），utility 服务再映射为 stock.isOpen；前端直接使用该权威字段。已核对服务端最新代码，保留现状并 resolve。

packages/kit/src/views/Swap/pages/components/PreSwapDialogContent.tsx
```diff
@@ -307,6 +328,45 @@ const PreSwapDialogContent = ({
...
+        await rebuildReviewWithSlippage(slippagePercentage);
+        if (!isMountedRef.current) {
+          return;
+        }
+        if (scope === 'future') {
+          setSettings((prev) => ({
+            ...prev,
+            swapSlippagePercentageMode: ESwapSlippageSegmentKey.CUSTOM,
+            swapSlippagePercentageCustomValue: slippagePercentage,
+          }));
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。Review adapter 新增 saveSlippageForFutureOrders：Swap 仍保存全局设置；Market 会保存当前 preset/direction 对应的 selectedSlippageValue，后续 Market 报价会读取该值。

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketHolders.ts
```diff
@@ -19,71 +18,45 @@ export function useMarketHolders({
...
+  >();
 
   useEffect(() => {
-    if (tokenDetail && tokenDetail?.fdv && tokenDetail?.price) {
-      setCacheTokenDetail(tokenDetail);
+    if (
+      tokenDetail?.fdv &&
+      tokenDetail.price &&
+      tokenDetail.address.toLowerCase() === tokenAddress.toLowerCase()
+    ) {
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。holders 缓存现在同时校验 tokenDetail.networkId 与当前 networkId，避免同地址跨链切换时复用旧网络的 FDV/价格。

packages/kit/src/views/Swap/pages/components/SwapProTradeInfoGroup.tsx
```diff
@@ -48,13 +53,23 @@ const SwapProTradeInfoGroup = ({
...
   const [swapProTradeType] = useSwapProTradeTypeAtom();
   const navigation = useAppNavigation();
 
+  const handleOpenProviderList = useCallback(() => {
+    dismissKeyboard();
+    navigation.pushModal(EModalRoutes.SwapModal, {
+      screen: EModalSwapRoutes.SwapProviderSelect,
+      params: {
+        storeName: EJotaiContextStoreNames.swap,
+      },
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。storeName 已从 SwapMainLand 逐层传到 SwapProTradeInfoGroup，provider modal 使用当前 Swap Pro / swapModal 上下文。

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3459,6 +3504,139 @@ export function useSwapBuildTx({
...
+            ...(shouldFallback
+              ? {
+                  shouldFallback: true,
+                  needFetchGas: true,
+                  supportNetworkFeeLevel: false,
+                  netWorkFee: undefined,
+                }
+              : {
+                  netWorkFee: estimateNetworkFeeResult.netWorkFee,
+                }),
```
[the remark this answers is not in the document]
zhaono1: 成立，已在 e638f6e76e 修复。重建成功的非 fallback 分支会显式清除 shouldFallback/needFetchGas，恢复 supportNetworkFeeLevel，并在旧状态或新结果涉及 fallback 时重建对应步骤。

packages/kit/src/views/Market/MarketDetailV2/components/InformationTabs/components/TransactionsHistory/hooks/useMarketTransactions.ts
zhaono1: 成立，已在 e638f6e76e 修复。暂停 polling 的 effect 增加 tokenAddress/networkId 依赖，在 usePromiseResult 因请求依赖变化清空 pause flag 后重新应用实时模式的暂停状态。

originalix (commented):
@zhaono1

Reviewed commit `e638f6e76e72`.

- **P1 · Stock market status now ignores backend closure signals**
- **P2 · Keyword token search fails hard on the supplemental endpoint**
- **P2 · Market slippage "all future orders" does not persist outside preset mode**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12876#issuecomment-5299983635)

packages/kit-bg/src/services/ServiceSwap.ts
```diff
@@ -840,25 +829,34 @@ export default class ServiceSwap extends ServiceBase {
...
+        keywords && onlySwapTokens
+          ? ['/swap/v1/tokens', '/swap/v1/swap-tokens']
+          : ['/swap/v1/tokens'];
+      const responses = await Promise.all(
+        endpoints.map((endpoint) =>
+          client.get<IFetchResponse<ISwapToken[]>>(endpoint, requestConfig),
+        ),
+      );
+      const tokens = mergeSwapTokenLists(
+        responses.map(({ data }) => data?.data ?? []),
```
sidmorizon: **Token search returns nothing when the newly added second search request fails**

Severity: `non-severe`

Both token-search requests are joined with an all-or-nothing wait (`Promise.all` at `packages/kit-bg/src/services/ServiceSwap.ts:850`), so if only the newly added second request fails the whole search is treated as failed and the user sees an error toast with an empty result list instead of the results that did come back.

Impact: A single failing backend call makes Swap/Limit token search appear completely broken even though the primary results were available.

<details>
<summary>All-or-nothing fan-out in fetchSwapTokens</summary>

For keyword searches with `onlySwapTokens`, `endpoints` becomes `['/swap/v1/tokens', '/swap/v1/swap-tokens']` (`packages/kit-bg/src/services/ServiceSwap.ts:846-849`) and both are awaited via `Promise.all`. A rejection from either (5xx, unknown route on older backends, network blip) rejects the combined promise and falls into the shared `catch`, which shows an error toast and returns `[]` (or rethrows when `throwOnError`). Previously only one endpoint existed, so a single failure was unambiguous; now a failure of the supplementary support-list endpoint discards the good primary response. Using `Promise.allSettled` and merging only fulfilled responses (while still preserving the `axios.isCancel` path for aborted requests) would keep search usable.

</details>

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx
```diff
@@ -1189,7 +1284,7 @@ export function useSpeedSwapActions(props: {
         swapProviderName: buildRes.result?.info.providerName ?? '',
         swapType: getSwapExecutionTypeFromQuoteResult(buildRes.result),
         orderType: getSwapAnalyticsCategoryFromQuoteResult(buildRes.result),
-        slippage: slippage.toString(),
+        slippage: (buildRes.result.slippage ?? slippage).toString(),
```
sidmorizon, line 1294: **Market order analytics can crash when the build response has no result payload**

Severity: `non-severe`

The slippage value for the order-created analytics event is read without a safety check (`buildRes.result.slippage` at `packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx:1287`), so a build response that arrives without its result payload throws while logging and aborts the surrounding flow.

Impact: A malformed backend build response can abort the Market order flow with an unexpected error instead of just skipping analytics.

<details>
<summary>Inconsistent optional access inside logMarketCreateOrder</summary>

Every other field in the same call is read defensively (`buildRes.result?.toAmount`, `buildRes.result?.info.provider`, `buildRes.result?.fee?.percentageFee`, etc.), which shows the author already treats `result` as possibly absent. The new slippage line drops the optional chain, so `buildRes.result` being `undefined` throws a runtime error inside the logger call rather than degrading gracefully. Use `buildRes.result?.slippage ?? slippage`.

</details>

Suggested fix:

```ts
        slippage: (buildRes.result?.slippage ?? slippage).toString(),
```

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/components/ActionButton.tsx
```diff
@@ -286,6 +282,12 @@ export function ActionButton({
     });
   }
 
+  if (isRefreshQuote && hasAmount) {
+    buttonText = intl.formatMessage({
+      id: ETranslations.swap_page_button_refresh_quotes,
+    });
+  }
+
```
sidmorizon: **Insufficient-balance warning is replaced by an enabled "Refresh quotes" button**

Severity: `non-severe`

When the trade amount exceeds the available balance and the price quote has expired, the trade button label is overwritten with the refresh text (`buttonText` assignment at `packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/components/ActionButton.tsx:285-290`) and the button is also re-enabled, so the user loses the "insufficient balance" warning.

Impact: Users with too little balance see an active refresh button instead of the clear insufficient-balance message, making the real blocker invisible.

<details>
<summary>Refresh override wins over the insufficient-balance branch</summary>

The insufficient-balance label is assigned earlier (`if (shouldDisable && isInsufficientBalance)`), and the new refresh override at lines 285-290 runs afterwards with only a `hasAmount` guard, so it unconditionally replaces that text. In addition, `isButtonDisabled` now computes `(quoteRefreshAvailable ? false : shouldDisable)` (`packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/components/ActionButton.tsx:310-317`), which bypasses `shouldDisable` — the flag that encodes insufficient balance — whenever a refresh is available. Because the Market quote effect no longer requires a positive balance before quoting (`packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx:3242-3247`), quotes (and therefore quote expiry) do occur for over-balance amounts, so this state is reachable. Gating the refresh label/enable on `!isInsufficientBalance` would preserve the warning.

</details>

packages/kit-bg/src/services/ServiceSwap.ts
zhaono1: 成立，已在 800bacc789 修复。双接口查询改为合并所有成功响应；任一请求被取消时仍保持原取消语义，只有全部失败才进入原错误处理。同时增加了补充接口失败时保留主接口结果的回归测试。

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx
zhaono1: 这条不成立，保留现状。IFetchBuildTxResponse.result 在 DTO 中是必填 IFetchBuildTxResult；构建流程也会先校验整个 buildRes，之后同一路径还会直接把 buildRes.result 传给分类函数并展开到 review snapshot。这里补可选链只会掩盖服务端违反既定构建契约的情况，无法让后续执行路径安全降级。

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/components/ActionButton.tsx
zhaono1: 成立，已在 800bacc789 修复。Refresh quotes 现在只在当前交易状态未被 shouldDisable 阻断时可用，因此余额不足或 wrapped 余额未就绪时会继续显示原阻断文案并保持禁用，不再被过期报价覆盖。

zhaono1:
### Follow-up review result

Handled in commit 800bacc789:

1. Stock market status: not applicable. utility-server origin/main at 16f259435bcc1709a93d52710ec5e4f498a09d45 derives issuer market status from holidays, early close, and halt state, then maps isMarketOpen through display.isOpen to token detail stock.isOpen. The removed client status endpoint duplicated that same source.
2. Dual token search failure: valid and fixed. Successful endpoint responses are retained and merged; cancellation still wins, and the original failure path is used only when every endpoint fails.
3. Non-preset future slippage: valid and fixed. Market now persists this choice through the existing global Swap slippage settings and reads it back when preset mode is disabled.
4. Build analytics optional result: not applicable. IFetchBuildTxResponse.result is required by the build DTO and by the downstream executable review path; optional access here would only hide a contract violation.
5. Expired quote plus insufficient balance: valid and fixed. Refresh cannot override a disabled trade state.

Validation: yarn agent:check --profile commit passed; 8 focused suites and 83 tests passed. All inline review threads have been replied to and resolved.

packages/kit/src/states/jotai/contexts/swap/actions.ts
```diff
@@ -1852,45 +1811,26 @@ class ContentJotaiActionsSwap extends ContextJotaiActionsBase {
...
-            providerLogo: wrappedTokens.find(
-              (item) => item.networkId === fromToken.networkId,
-            )?.logo,
-          }),
-        );
+        void this.resetQuoteAction.call(set);
         return;
       }
-      void this.runSpeedQuote.call(
+      void this.quoteAction.call(
```
originalix, line 1815: P1: [Swap Pro wrapped pairs no longer seed a local executable quote]

When Swap Pro Market is opened on a wrapped/native pair, this branch now always forwards the request into the shared provider quote flow instead of creating the existing 1:1 wrapped quote locally. The UI still treats wrapped pairs as executable, but the review entry path later bails out when `currentQuoteRes` is missing, so the visible "Wrap" action can become a silent no-op.

Please restore the local wrapped-quote branch here by reusing `buildWrappedSwapQuoteResult()` for wrap/unwrap pairs, and keep the provider quote pipeline only for non-wrapped market swaps.

zhaono1: This does not apply, so the unified flow is intentionally retained. quoteSpeedAction delegates to quoteAction, and ServiceSwap maps that request to /swap/v1/quote/events. The current server-service-swap origin/main (e7a342942c) has an explicit wrapped-pair branch in fetchQuoteEvent: it emits totalQuoteCount: 1, then an isWrapped: true quote with 1:1 amounts and the Wrap Contract provider before closing the SSE stream. The APP event reducer accepts that result into the shared current-quote atom. While the response is pending, SwapProActionButton remains disabled through !currentQuoteRes?.toAmount, so there is no enabled silent no-op. Restoring the local quote here would reintroduce duplicate quote ownership and a race with the server result.

packages/kit-bg/src/services/ServiceSwap.ts
```diff
@@ -641,44 +639,35 @@ export default class ServiceSwap extends ServiceBase {
...
+        );
+    for (const requestId of requestIds) {
+      this._activeQuoteEventRequestIds.delete(requestId);
+      const sources = this._quoteEventSources.get(requestId);
+      this.removeQuoteEventSourceListeners(requestId);
+      sources?.eventSource?.close();
+      sources?.eventSourcePolyfill?.close();
+      this._quoteEventSources.delete(requestId);
     }
   }
```
sidmorizon, line 673: **Getting a private-send price quote silently kills the price stream of an open trade panel**

Severity: `non-severe`

The private-send flow tears down price streams without naming which one it owns (`cancelFetchQuoteEvents()` at `packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx:724`), so any other trade panel that is waiting for prices at that moment loses its stream and keeps spinning forever.
Impact: A user who opens the send-amount screen while a Swap/Market trade panel is fetching prices can be left with a permanently loading trade panel until they change the amount to force a new request.

<details>
<summary>Per-request stream ownership makes the no-argument cancel a global kill switch</summary>

Before this PR the background service held a single `_quoteEventSource`, so at most one quote stream existed and `cancelFetchQuoteEvents()` (no argument) could only affect that one owner. The PR replaces it with `_quoteEventSources: Map<requestId, ...>` plus `_activeQuoteEventRequestIds` (`packages/kit-bg/src/services/ServiceSwap.ts:513-521`) and deliberately allows concurrently owned streams (see the new test `keeps independently owned quote streams active`).

`cancelFetchQuoteEvents(quoteRequestId?)` (`packages/kit-bg/src/services/ServiceSwap.ts:655-672`) now iterates every active request id when called without an argument and closes/removes all of them. The private-send teardown path calls it with no argument (`packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx:257`, `:605`, `:724`), so it closes streams created by `quoteAction` for the Swap page or by the new Market panel quote effect (`packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx`).

Those owners never learn the stream died: no `done`/`close`/`error` event is emitted, so `swapQuoteFetchingAtom` / `swapQuoteActionLockAtom` stay set and the panel keeps showing the loading/blocked state.

</details>

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -2511,22 +2547,24 @@ export function useSwapBuildTx({
...
+                  encodedTx,
+                  transferInfo,
+                },
               },
-            },
-          }));
-          void swapBuildFinish(buildSwapRes, data);
+            }));
+          }
+          void swapBuildFinish(buildSwapRes, data, effectiveSlippagePercentage);
```
sidmorizon, line 2567: **Editing slippage in the review sheet records a fake order-creation analytics event**

Severity: `non-severe`

Rebuilding the transaction after a slippage edit reports a successful order creation again (`swapBuildFinish` at `packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts:2567`), even though nothing was submitted, so order-creation metrics and recent-pair/limit-order side effects fire twice for one trade.
Impact: Analytics show more created orders than users actually placed, and duplicate build-success side effects run while the review sheet is still open.

<details>
<summary>forceRebuild path reuses the normal build completion handler</summary>

`rebuildSwapWithSlippage` (`packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts:3566-3606`) calls `buildSwapAction(..., { forceRebuild: true, slippagePercentage, useCustomSlippage: true, updateReviewState: false })`. Inside `buildSwapAction`, all review-state writes are correctly suppressed by `updateReviewState`, but the terminal `void swapBuildFinish(buildSwapRes, data, effectiveSlippagePercentage)` is not guarded. `swapBuildFinish` logs `defaultLogger.swap.createSwapOrder.swapCreateOrder({ status: ESwapEventAPIStatus.SUCCESS, ... })` and also triggers `syncRecentTokenPairs` / `EAppEventBusNames.SwapLimitOrderBuildSuccess` depending on protocol. Since the initial review already invoked `swapBuildFinish` for the same trade, every slippage edit adds another create-order record. The Market equivalent (`rebuildMarketSwapReview` in `packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/hooks/useSpeedSwapActions.tsx:1702`) does not log, so the two owners are also inconsistent.

</details>

originalix (approved):

PR #12877, opened 2026-08-14
fix: restore firmware-switch device management behavior (OK-60205)

originalix:
[OK-60205](https://onekeyhq.atlassian.net/browse/OK-60205)

----------

## What changed

- upgrade direct hardware SDK dependencies from `1.2.0-alpha.119` to `1.2.0-alpha.123`
- restore the v6.5.0 rule that deprecated wallets are excluded from device management after switching firmware type
- broadcast `WalletUpdate` after the firmware workflow successfully deprecates affected wallets, so the mounted device list refreshes immediately
- add focused regression coverage for switched-firmware wallet filtering and the post-update refresh event

## Why

The SDK upgrade consumes the serialized WebUSB bootloader polling fix for OK-60205. Separately, a later App change allowed deprecated switched-firmware wallets to remain manageable even though there was no product requirement for that behavior. Removing the exception restored the persisted-state behavior, but the mounted device list still retained its previous query result until App reload because the firmware path did not emit `WalletUpdate`.

## Impact

After a successful Universal/Bitcoin-Only firmware switch, the old wallet is deprecated and disappears from device management immediately. Users no longer see a stale Bitcoin-Only entry that only disappears after reload.

## Validation

- `yarn install --immutable`
- `deviceStateManagement.test.ts`: 32 tests passed
- `ServiceHardware.pro2DeviceManagement.test.ts`: 40 tests passed
- all 18 direct `@onekeyfe/hd-*` / `@onekeyfe/hwk-*` manifest references verified at `1.2.0-alpha.123` with npm `gitHead=e5e299d0ac7208c878ef1d7fe5e96ec591ed3634`

[OK-60205]: https://onekeyhq.atlassian.net/browse/OK-60205?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3523,9 +3523,13 @@ class ServiceHardware extends ServiceBase {
...
     }
 
-    await this.backgroundApi.serviceAccount.updateWalletsDeprecatedState({
-      willUpdateDeprecateMap,
-    });
+    const result =
+      await this.backgroundApi.serviceAccount.updateWalletsDeprecatedState({
+        willUpdateDeprecateMap,
+      });
+    if (result && Object.keys(willUpdateDeprecateMap).length > 0) {
```
sidmorizon, line 3530: # [🟡 P2] 部分写入失败时不发 `WalletUpdate`，DB 与列表状态会不一致

## 问题

`serviceAccount.updateWalletsDeprecatedState()` 内部是逐个钱包串行写库的：

```ts
for (const [walletId, isDeprecated] of Object.entries(willUpdateDeprecateMap)) {
  await localDb.setWalletDeprecated({ walletId, isDeprecated });
}
return true;
// catch → console.error → return false
```

整个循环被同一个 `try/catch` 包住，任意一次 `setWalletDeprecated` 抛错都会让函数返回 `false`，但**在它之前已经成功写入的钱包已经是 `deprecated = true` 了**。此处新增的 `if (result && ...)` 恰好在这种「部分成功」的场景下跳过 `appEventBus.emit(EAppEventBusNames.WalletUpdate)`。

## 影响

一台设备在固件类型切换后通常同时存在标准钱包 + 若干隐藏钱包（`willUpdateDeprecateMap` 长度 > 1），只要第 2 个及之后的写入失败，就会出现：DB 里部分钱包已 `deprecated`，但 `DeviceManagementListModal` / `DeviceDetailsModal` 都收不到刷新事件。由于本 PR 同时把 `isDeviceManagementWalletUsable` 收敛成了 `!wallet.deprecated`，这条残留列表项点进去会立刻走 `refresh → data === undefined → handleBackPress` 弹回，表现为「点了没反应」，且只能靠重启 App 恢复——正是这个 PR 想消灭的那个现象。

## 建议

事件语义应该是「本次流程可能改动过钱包状态，请刷新」，而不是「全部成功才刷新」。把 emit 与 `result` 解耦即可：

```ts
const hasPendingUpdate = Object.keys(willUpdateDeprecateMap).length > 0;
await this.backgroundApi.serviceAccount.updateWalletsDeprecatedState({
  willUpdateDeprecateMap,
});
if (hasPendingUpdate) {
  appEventBus.emit(EAppEventBusNames.WalletUpdate, undefined);
}
```

（`accountSelector/actions.tsx` 里 `updateHwWalletsDeprecatedStatus` / `updateTrezorWalletsDeprecatedStatus` 两处是同样的 `result &&` 写法，若认可该结论建议一并调整，避免三处同源逻辑再次分叉。）

[Claude session ↗](https://claude.ai/code/session_01CBT9Lp7VJ7z7uJxBaP4q4o)

---

packages/kit/src/states/jotai/contexts/deviceDetails/deviceStateManagement.test.ts
```diff
@@ -89,7 +89,7 @@ describe('device reset wallet isolation', () => {
   });
 
   it.each(firmwareTypeSwitchDeviceTypes)(
-    'keeps a deprecated %s wallet manageable after switching firmware type',
+    'does not expose a deprecated %s wallet after switching firmware type',
```
sidmorizon, line 92: # [🟡 P2] 反转后的这批用例已退化为同义重复，不再守护任何行为

## 问题

`isDeviceManagementWalletUsable` 现在只剩 `return !wallet.deprecated;`，于是 `describe('device reset wallet isolation')` 里除 `keeps an active mocked standard wallet…` 之外的 7 个 `it` / `it.each`（共 13 个 case）全部在断言同一件事：`deprecated: true` → `undefined`。用例中的 `firmwareTypeAtCreated`、`device.deviceType`、`featuresInfo.$app_firmware_type`、`featuresInfo.firmwareType`、`deviceStateInfo.identity.firmwareType` 已经没有任何代码路径会读取，`firmwareTypeSwitchDeviceTypes` 与 `unsupportedDeviceTypes` 两个数组的区分也随之失去意义——两组断言完全相同。

更麻烦的是用例名仍在承诺它并不覆盖的语义，例如 `does not revive a deprecated Protocol V1 wallet from normalized firmwareType`（`normalized firmwareType` 的归一化逻辑本 PR 已删除）、`does not expose a deprecated %s wallet without a firmware type switch action`（不存在 switch action 这个分支了）。

## 影响

这批测试现在是恒真的：把实现改成 `return !walletWithDevice?.wallet?.deprecated` 之外的任何等价/劣化写法（例如彻底忽略 `device`、忽略 `isMocked`）它们都照样绿。真正需要被钉住的产品规则——「固件切换后设备从设备管理消失，重新连接使 `deprecated` 复位后设备重新出现」——一条都没覆盖，下一次有人再改这里时不会有任何测试报警，等于把 OK-60205 的回归防线交回给人工回归。

## 建议

把 13 个 case 收敛成一张按 `deprecated` / `isMocked` 组合的表，再补一条真正指向本 PR 意图的用例：

```ts
it.each([
  [{ deprecated: true }, false],
  [{ deprecated: true, isMocked: true }, false],
  [{ deprecated: false }, true],
  [{ deprecated: false, isMocked: true }, true], // hidden-only 代理钱包
])('gates device management purely on wallet.deprecated: %j', (wallet, usable) => {
  expect(
    isDeviceManagementWalletUsable({ wallet: { id: 'hw-1', ...wallet }, device: { id: 'd-1' } } as never),
  ).toBe(usable);
});

it('re-exposes the device once the wallet is un-deprecated on re-add', () => {
  const w = { wallet: { id: 'hw-1', deprecated: false, firmwareTypeAtCreated: EFirmwareType.BitcoinOnly }, device: { id: 'd-1' } };
  expect(resolveUsableWalletWithDevice(w as never)).toBe(w);
});
```

[Claude session ↗](https://claude.ai/code/session_01CBT9Lp7VJ7z7uJxBaP4q4o)

---

wabicai (approved):

originalix:

@originalix

Reviewed commit `450d46f1db43`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the firmware-switch state transition across the background and foreground boundary: the background runtime persists deprecation for every wallet associated with the switched device and publishes the existing cross-runtime `WalletUpdate` event, while the foreground device-management projection now rejects every deprecated wallet. This refreshes stale UI promptly, but it also removes the management-only exception that kept a firmware-mismatched device accessible for switching back. The PR additionally advances the aligned OneKey hardware SDK family from `1.2.0-alpha.119` to `1.2.0-alpha.123` across the application, CLI, resolutions, and lockfile; the dependency graph remains aligned and no application-side secret handling or package boundary is changed.

**What needs attention:** Confirm that a Classic1s, ClassicPure, or Pro remains reachable in Device Management after either firmware-type transition so the user can switch it back without creating a new wallet through onboarding.

### Issues to address

- **P1 · Firmware switches remove the only switch-back entry**
  - **Impact:** When a supported device finishes switching firmware type, `updateHwWalletsDeprecatedStatus` marks every wallet for that connection deprecated and emits `WalletUpdate`. The new unconditional `!wallet.deprecated` rule then removes the same device from both the management list and an open details page. Because the firmware-type switch control exists in Device Details and onboarding does not expose that control, users cannot switch the device back without first creating or reconnecting a wallet under the new firmware type. This reverses the existing management-only exception that deliberately kept mismatched deprecated wallets accessible while their accounts remained unusable elsewhere.
  - **Suggested action:** Restore the firmware-mismatch exception in `isDeviceManagementWalletUsable` for device-management consumers, while continuing to treat the wallet and its accounts as deprecated in normal wallet surfaces; retain the new `WalletUpdate` emission so both views refresh after persistence.

### Validation gaps

- The two targeted Jest files could not run because this worktree has no Yarn `node_modules` state file.
- The `1.2.0-alpha.123` package artifacts and release delta were not present in the local Yarn cache, so transport, native-layer, outbound-request, and signing changes could not be compared with `1.2.0-alpha.119`.
- GitHub review comments and thread resolution state could not be retrieved because GitHub API access was unavailable; the inline wording therefore needs a duplicate-comment check before posting.
- Real-device validation should cover Universal-to-Bitcoin-only and Bitcoin-only-to-Universal transitions. Passing requires the stale accounts to remain unusable while the physical device remains available in Device Management and can initiate the reverse transition.

PR #11814, opened 2026-05-28
feat(hardware): add complete OneKey Pro 2 integration

wabicai:
## Summary

This PR prepares OneKey Pro 2 support for the `hotfix/v6.5.2` release line. The branch was rebuilt directly on the hotfix base so it carries the Pro 2 implementation without the previous merge from `x` or unrelated fixes from that branch.

## Production readiness changes

- Remove the standalone Pro 2 onboarding card. Pro 2 now uses the normal OneKey USB/BLE discovery path without a protocol-specific search filter.
- Use the active connection channel for onboarding UI so native BLE sessions no longer show the USB connection card.
- Remove Pro 2 test-mode, debug settings, debug firmware routes, and related background flags.
- Keep the complete Pro 2 onboarding and DeviceSetup flow, including USB/BLE reconnect handling after firmware restart.
- Trigger Portfolio collection only for a Pro 2 hardware wallet. Background authorization additionally requires a OneKey Protocol V2 device before packing or upload.
- Request full firmware hash verification only during the post-update recheck; normal update availability checks use version metadata only.
- Improve multi-stage firmware progress and retain a 30-second normal check watchdog plus the dedicated post-update reconnect window.

## Hardware SDK

- Upgrade the hardware SDK stack to `1.2.0-alpha.56`.
- The SDK release includes the optional firmware verification request, 30-second BLE `DeviceInfoGet` timeout, and Android BLE bootloader/release disconnect handling.
- Publish workflow: https://github.com/OneKeyHQ/hardware-js-sdk/actions/runs/30903437683

## Hotfix scope

- Base: `hotfix/v6.5.2`
- Preserved: Pro 2 device state, USB/BLE transport, onboarding, wallet sessions, device management, firmware update, Portfolio, NFT, network compatibility, signing, and CLI support.
- Excluded: the previous `x` merge, Android build metadata changes, CodeGraph ignore changes, Private Send/Swap price fixes, tray avatar cleanup, generic developer Gallery/side-bar test IDs, and other unrelated branch work.

## Validation

- [x] Hardware SDK unit tests pass (263 SDK tests + Electron transport coverage).
- [x] Hardware SDK `yarn agent:check --profile commit` passes.
- [x] App Pro 2/hardware tests pass: 60 suites, 541 tests.
- [x] CLI hardware tests pass: 5 suites, 43 tests.
- [x] App `yarn agent:check --profile commit` passes.
- [x] App `yarn agent:check --profile pr` local checks pass.

## Manual verification

[...]

PR #12879, opened 2026-08-14
fix: sync V1 device-side settings changes after device operations (OK-60121)

originalix:
[OK-60121](https://onekeyhq.atlassian.net/browse/OK-60121)

----------

## Problem (OK-60121)

OneKey Pro (Protocol V1, desktop BLE): after changing the language on the device itself, the app's device management page keeps showing the old language. Brightness/wallpaper changes never synced it; auto-lock/auto-shutdown only synced it in some cases.

## Root causes (all verified against live device logs)

1. **BLE listener-timing hole (SDK)**: on BLE, `device.initialize()` runs inside `ensureConnected()` before core attaches per-call `DEVICE.STATE` listeners, so the only event whose `changedKeys` explicitly contains `settings.language` is emitted with zero listeners. Every later event diffs against the already-updated SDK cache and never mentions the language again.
2. **Sparse merges drop off-diff settings (app)**: `mergeDeviceStateEvent` only copied fields listed in `changedKeys`, so the correct language riding along in `initialize`/`settings-write` full-state events was discarded (verified: `eventLanguage: 'ja', persistedLanguage: 'en'`).
3. **UI staleness gate dropped force-emitted `settings-read` events** (equal revision/updatedAt), and `refresh()` could clobber a newer applied snapshot with a stale cached DB read (record cache is invalidated before the write commits).
4. **No deterministic post-mutation sync**: `setBrightness`/`setDeviceHomeScreen` had no `preciseUpdateFields` fallback, and current mainline SDK emits no event at all when the optimistic `ApplySettings` patch matches its cache (V1 brightness maps to an empty patch; same-value writes too).

## Fixes

- `deviceStateUtils.mergeDeviceStateEvent`: treat `settings-read` (any protocol) and V1 `initialize`/`settings-write` events as authoritative full-settings snapshots.
- `deviceStateManagement.getDeviceStateSnapshotFromEvent`: only drop `settings-read` events strictly older by `updatedAt`; added `pickNewerDeviceStateSnapshot` so `refresh()` never regresses a newer snapshot.
- `DeviceSettingsManager._notifyProtocolV1SettingsSynced`: after every V1 settings mutation (and wallpaper), drain pending event persists, **explicitly read the settings back** (`getDeviceState({scope:'settings'})`, one extra V1 GetFeatures) and persist that snapshot, then emit `HardwareFeaturesUpdate` so the details page re-reads the DB. `wipeDevice` opts out; duplicate `HardwareFeaturesUpdate` emissions deduped via `skipFeaturesUpdateEvent`.
- `LocalDbBase`: clear device record caches again after the write transaction commits (pre-commit invalidation let concurrent reads re-cache stale records for up to 5s).
- Observability: `hardwareLog` now writes through the platform file-backed log transport (not gated by the logger opt-in config); DEVICE.STATE receipt/persist logs include language and persistence result kind; UI event application logs applied/dropped with comparison stamps.

## SDK dependency

Bumped `@onekeyfe/hd-*` to `1.2.0-alpha.123` (clean mainline build, verified via `npm pack`). The app-side read-back makes the fix independent of the SDK emitting events; merging `codex/fix-ok-60121-v1-settings-refresh` into the SDK mainline later removes the extra GetFeatures round trip.

## Test plan

- Unit: 69 tests across `deviceStateUtils.test.ts`, `deviceStateManagement.test.ts`, `LocalDbBase.deviceState.test.ts` (new cases: authoritative V1 initialize/settings-write merges, equal-stamp settings-read acceptance, snapshot monotonicity).
- Device (Pro, desktop BLE, alpha.123): change language on device → change brightness / auto-lock (same and different values) → Language row updates; log self-check markers `v1 settings read-back { kind: 'updated' }` and `applyDeviceStateEvent { applied: true }`.
- Regression: auto-lock/auto-shutdown/haptic/language settings themselves, Pro 2 (V2) settings page unaffected (V2 keeps sparse semantics and existing `waitForDeviceStateSync` path), Trezor paths untouched.

[OK-60121]: https://onekeyhq.atlassian.net/browse/OK-60121?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

@originalix

Reviewed commit `69e15c5639f0`. This report always reflects the latest reviewed changes.

### Review summary

This PR moves Protocol V1 settings convergence into the `bg` hardware boundary: after a mutation, `bg` drains its per-runtime event queue, reads settings from the device, writes the shared native database, and signals `main`, whose separate JS heap reconciles that data with its Jotai snapshot. It also upgrades the version-locked hardware SDK family to `1.2.0-alpha.123` and adds persistent diagnostics plus a development logger configuration RPC that updates `bg` before mirroring `main`. Package layering remains coherent and the new tests cover several merge-order cases, but the timeout can leave a late database write unannounced and the persistent logging sink still accepts legacy raw identifiers, feature payloads, and errors.

**What needs attention:** Confirm that a Protocol V1 read-back completing after the eight-second timeout still refreshes `main`, and restrict persistent hardware diagnostics to an explicitly redacted payload contract.

### Issues to address

- **P1 · Do not persist legacy hardwareLog payloads verbatim**
  - **Impact:** In production, the changed `hardwareLog()` wrapper routes every existing caller into `HardwareSDKScene.serviceEvent()` and its durable local-file transport. The wrapper forwards arbitrary arguments without redaction, while reachable callers pass full `connectId` values, firmware payloads containing complete device features, and raw errors. Those stable device identifiers and diagnostic details can therefore enter support-exportable logs during ordinary hardware operations.
  - **Suggested action:** Keep the legacy arbitrary-argument trace development-only, add a separate typed persistent helper with an explicit redacted whitelist, and migrate only call sites whose payloads satisfy that contract.
- **P1 · Notify again when a timed-out sync finishes**
  - **Impact:** When the Protocol V1 read-back exceeds eight seconds, `timeoutGuard` wins the race but does not cancel `_syncProtocolV1SettingsSnapshot()`. The `finally` block emits the only `HardwareFeaturesUpdate` before that late task can write authoritative settings to the shared native database. If the SDK event is absent or rejected—the exact condition this explicit read-back handles—the separate `main` JS heap never reloads the late write and can retain stale settings.
  - **Suggested action:** Capture the synchronization promise and, when the timeout wins, attach a completion path that emits a second refresh only after the late persistence and queue drain finish; add a fake-timer regression test for this ordering.

### Validation gaps

- `git diff --check origin/hotfix/v6.5.2...HEAD` completed without errors.
- Static inspection found no changed import that violates the OneKey package hierarchy.
- The five focused Jest files could not start because this worktree has no Yarn node-modules state file.
- No test covers a V1 read-back that finishes after the eight-second timeout.
- Published sources for the coordinated `@onekeyfe` hardware package upgrade were not installed in the workspace, so API and supply-chain differences between `1.2.0-alpha.119` and `1.2.0-alpha.123` could not be inspected.
- Existing GitHub reviews, inline comments, general comments, and thread state could not be retrieved, so inline wording should be checked for duplicate coverage before posting.

originalix (commented):
@originalix

Reviewed commit `224a60208226`.

- **P1 · Wait for the read-back event before refreshing the UI**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12879#issuecomment-5290528140)

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1420,6 +1421,12 @@ class ServiceHardware extends ServiceBase {
           revision: event.revision,
           source: event.source,
           changedKeys: event.changedKeys,
+          connectId: event.connectId,
+          serialNo: event.state?.identity?.serialNo,
```
sidmorizon: # 🔴 P1 新增的日志字段打红了既有单测，unittest shard 2 连续三个 commit 失败

## 问题

这里给 `hardwareLog('device state update', ...)` 的 payload 增加了 `connectId` / `serialNo` / `language` / `updatedAt`，同时下面新增了 `hardwareLog('device state persist result', ...)`。但 `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts:1192` 用的是精确匹配断言，本 PR 没有同步更新它：

```
expect(hardwareLogSpy).toHaveBeenCalledWith('device state update', {
  changedKeys: ['identity.label'],
  revision: 2,
  source: 'apply-settings',
});
```

我在本 worktree 的 head `f3dd4cf` 上单独跑该文件复现（`jest packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts`）：`1 failed, 38 passed`，失败用例是 `ServiceHardware SDK DeviceState synchronization › persists and broadcasts canonical SDK state events`，diff 里多出的正是 `connectId: "PRO2_USB"`、`serialNo: "PRO2_SERIAL"`、`language: undefined`、`updatedAt: undefined`。切到 base `75b40f39` 跑同一文件则全部通过。

## 影响

这是确定性失败，不是 flake：unittest 的 `Unit Tests (2/3)` 在 `9379db3`、`224a602`、`f3dd4cf` 三个 commit 上都是 failure。workflow 里 shard 级别传的是 `--coverageThreshold='{}'`，所以退出码 `1` 只可能来自真实断言失败，与 coverage 阈值无关。在修掉之前 CI 无法转绿。

另外该断言的第二次调用还暴露出 `"kind": "error"`，说明这条测试路径里 `localDb.updateDeviceState` 实际抛错走了 catch 分支，顺带值得确认新日志是否掩盖了别的问题。

## 建议

同步更新 `ServiceHardware.pro2DeviceManagement.test.ts:1192` 的断言。若不希望日志字段和测试强耦合，改用 `expect.objectContaining({ changedKeys, revision, source })`，这样后续再加诊断字段不会再打红既有用例。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -325,11 +400,24 @@ export class DeviceSettingsManager extends ServiceHardwareManagerBase {
...
+        } else {
+          const shouldNotifySettingsSynced = !skipV1SettingsSyncNotify;
+          if (preciseUpdateFields && device.featuresInfo) {
+            await localDb.updateDevice({
+              features: device.featuresInfo,
+              preciseUpdateFields,
+              // When the authoritative notify below fires (after the SDK's
+              // settings-read event persisted), emitting here as well would
+              // trigger a redundant refresh that can read the DB too early.
+              skipFeaturesUpdateEvent: shouldNotifySettingsSynced,
```
sidmorizon, line 464: # 🔴 P1 新增 `skipFeaturesUpdateEvent` 打红 `DeviceSettingsManager.pro2.test.ts` 的 4 个既有用例

## 问题

`localDb.updateDevice()` 现在多传了 `skipFeaturesUpdateEvent`，但 `packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.pro2.test.ts:396` 是精确参数断言，本 PR 没有同步更新：

```
expect(localDb.updateDevice).toHaveBeenCalledWith({
  features: device.featuresInfo,
  preciseUpdateFields,
});
```

我在本 worktree 的 head `f3dd4cf` 上跑该文件复现，4 个用例失败，实际收到的对象比期望多了 `skipFeaturesUpdateEvent: true`：`persists legacy OneKey setPassphraseEnabled / setAutoLockDelayMs / setAutoShutDownDelayMs / setLanguage settings without waiting for Pro2 state events`。切到 base `75b40f39` 跑同一文件全部通过（`66 passed`），确认是本 PR 引入。

## 影响

CI 上 `Unit Tests (1/3)` 在 `224a602` 上确实是 failure（`f3dd4cf` 那次该 shard 因 fail-fast 被 cancelled，状态未知，但本地在 `f3dd4cf` 上仍稳定复现这 4 个失败）。和 shard 2 的那条一样，workflow 在 shard 级别传的是 `--coverageThreshold='{}'`，所以这是真实断言失败而非 coverage 阈值问题。

值得注意的是这 4 个用例正好覆盖 V1 设置写入的主路径，等于本 PR 改动最核心的分支目前没有通过的回归保护。

## 建议

更新这 4 处断言，显式带上 `skipFeaturesUpdateEvent: true`（这样也顺带把「V1 路径必须由 `_notifyProtocolV1SettingsSynced` 统一发事件」这个新契约固化下来），或改成 `expect.objectContaining({ features, preciseUpdateFields })`。同时建议补一个用例断言 `wipeDevice`（`skipV1SettingsSyncNotify: true`）时 `skipFeaturesUpdateEvent` 为 `false`，避免这个取反关系被后续改动写反。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -140,6 +148,72 @@ type ITrezorDeviceSettingsAction = (params: {
...
+        device.uuid,
+        device.deviceId,
+        device.deviceStateInfo?.identity.serialNo,
+        device.deviceStateInfo?.identity.deviceId,
+      ],
+    });
+    try {
+      const connectId = compatibleConnectId || device.connectId;
+      if (connectId) {
+        const state = await this.serviceHardware.getDeviceState({
```
sidmorizon: # 🟠 P1 回读串行阻塞主流程，超时上限只受 60s 保护，`setHomeScreen` 还在处理弹窗关闭后继续等

## 问题

这个 `getDeviceState({ scope: 'settings' })` 会走 `ServiceHardware._getDeviceStateWithMutex` → `getFeaturesMutex.runExclusive(_getDeviceStateWithTimeout)`，而 `_getDeviceStateWithTimeout` 的 `timeout` 是 `timerUtils.getTimeDurationMs({ seconds: 60 })`。它被 `await` 在两个位置：

- `DeviceSettingsManager.ts:416`，在 `_withDeviceProcessing` 的 `withHardwareProcessing` 回调**内部**，即每一次 V1 设置写入（`setLanguage` / `setAutoLockDelayMs` / `setAutoShutDownDelayMs` / `setBrightness` / `setHapticFeedback` …）都会多一次串行 GetFeatures 往返；
- `DeviceSettingsManager.ts:736`，在 `withHardwareProcessing` **外部**，`setHomeScreen` 的返回 promise 要等它完成才 resolve。

注意改动前的 `else if (preciseUpdateFields && device.featuresInfo)` 分支只做一次本地 DB 写入，不发起任何硬件调用。

## 影响

正常情况下这只是把 V1 设置写入的耗时大致翻倍；但异常路径代价很高：设备在写入后掉线/重连（BLE 出范围、`setPassphraseEnabled` 之类会让设备重启的操作）时，`hardwareCallContext` 是 `USER_INTERACTION_NO_BLE_DIALOG`，不会弹连接框，调用会一直挂到 `60s` 超时才被 catch。这段时间里 `withHardwareProcessing` 的 processing 状态一直被占住；`setHomeScreen` 的情况更差——壁纸已经应用、处理弹窗已经关闭，UI 却还在等一个没有任何进度提示的 promise，用户看到的是「点了没反应」。

另外 `waitForDeviceStateSync()`（`:168`）在 `try` 之外，它内部是 `while (tasks.length > 0)` 的重取循环；设备处于事件风暴时这里同样没有上界，一旦不返回，后面的 `appEventBus.emit(HardwareFeaturesUpdate)` 永远不会触发，而此时 `updateDevice` 已经用 `skipFeaturesUpdateEvent: true` 抑制了自己的事件，UI 会停在旧值。

## 建议

给回读单独设一个远小于 60s 的超时（例如 5–8s），超时即放弃并照常 emit；同时把 `waitForDeviceStateSync()` 也纳入 `try/catch` 并加超时保护，确保 `appEventBus.emit(HardwareFeaturesUpdate)` 在任何异常路径下都会执行。`setHomeScreen`（`:736`）建议不要 `await` 回读，或至少在 `withHardwareProcessing` 内部完成，避免返回值已经就绪却继续挂起调用方。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

originalix (commented):
@originalix

Reviewed commit `f3dd4cf6fa15`.

- **P1 · Wait for the post-read event before refreshing main**
- **P1 · Redact hardware diagnostics before persistent logging**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12879#issuecomment-5290528140)

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
sidmorizon: 补充（复核于 head `7144048`，第 4 个 commit 只新增了 dev settings 的 `SectionLoggerParityItem`，未触及此处）：**这条不只是断言过期，它同时违反了仓库里既有的安全契约，因此不能靠改断言来消除。**

同一个用例在 `ServiceHardware.pro2DeviceManagement.test.ts:1197-1199` 还有一条**本 PR 之前就存在**的断言：

```
expect(JSON.stringify(hardwareLogSpy.mock.calls)).not.toContain(
  'PRO2_SERIAL',
);
```

我确认它在 base `75b40f39` 的同文件 `:1234` 已经存在（`git show 75b40f39:...` 可查），最初随 `a488829` (#11814, Pro 2 集成) 引入。也就是说仓库里存在一条显式约束：**设备序列号不得进入 `hardwareLog`**。而本行新增的 `serialNo: event.state?.identity?.serialNo` 正是直接把它写进了 payload。

我在本 worktree 做了验证：把 `:1192` 那条严格断言放宽成 `expect.objectContaining({ changedKeys, revision, source })` 后重跑，用例**依然失败**，这次卡在 `:1200`：

```
expect(received).not.toContain(expected) // indexOf
Expected substring: not "PRO2_SERIAL"
Received string: "[[\"device state update\",{\"revision\":2,\"source\":\"apply-settings\",\"changedKeys\":[\"identity.label\"],\"connectId\":\"PRO2_USB\",\"serialNo\":\"PRO2_SERIAL\"}], ...]"
```

（验证完已 `git checkout --` 还原，未改动任何文件。）

所以修复方向需要修正一下我原评论里的建议：**不要只把 `:1192` 改成 `objectContaining`**，那样只会把失败从 `:1192` 推到 `:1200`。正确做法是从 payload 里移除 `serialNo`（`connectId` 同理值得一并评估）。OK-60121 排查真正需要的是 `language` / `updatedAt` / `source` / `revision`，设备身份可以用脱敏形式代替，例如只保留末 4 位或哈希前缀，这样既保住诊断价值又不破坏既有契约。这一点也和 Codex 在 `serviceHardwareUtils.ts` 上那条 P1 指向同一个根因。

顺带说明 `f3dd4cf` 的日志重构（改走 `defaultLogger.hardware.sdkLog.serviceEvent`）**没有缓解这里**：`loggerConfigManager.shouldLog()` 里第一句就是 `if (!this._env.isDev) return true`，生产构建无条件落盘；`features` 已经用 `devOnlyData()` 包住了，但本行的 `serialNo` / `connectId` 没有。

CI 现状（head `7144048`）：本地实跑这两个测试文件仍是 `5 failed`，失败用例与我上一轮报告的完全一致，没有出现新的失败。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
```diff
@@ -1,9 +1,20 @@
...
 import { isProtocolV2ProductType } from '@onekeyhq/shared/src/utils/hardwareDeviceTypes';
 
 function hardwareLog(name: string, ...args: any[]) {
-  console.log(`ServiceHardwareLog@${name}`, ...args);
+  // Route through the platform log transport (file-backed on desktop/native)
+  // so hardware event evidence survives without an attached DevTools console.
+  // Intentionally NOT gated by the per-scope loggerConfig opt-in.
+  try {
+    loggerUtils.consoleFunc(
+      `ServiceHardwareLog@${name} : ${stringifyFunc(...args)}`,
```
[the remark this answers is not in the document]
originalix: Fixed in ee7762c829. `hardwareLog` now routes through `defaultLogger.hardware.sdkLog.serviceEvent` (single config-gated pipeline), the full `features` dump is wrapped with `devOnlyData()` (production keeps only the event name), and the remaining identifiers in the `device state update` payload (`serialNo` / `connectId`) are masked via `maskLogIdentifier` to a `***<last4>` suffix — enough for multi-device correlation, no full identifiers in persisted logs.

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -140,6 +148,72 @@ type ITrezorDeviceSettingsAction = (params: {
...
+          connectId,
+          params: { scope: 'settings' },
+          hardwareCallContext:
+            EHardwareCallContext.USER_INTERACTION_NO_BLE_DIALOG,
+          silentMode: true,
+        });
+        if (state) {
+          const persistResult = await localDb.updateDeviceState({
+            changedKeys: ['settings'],
```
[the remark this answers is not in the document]
originalix: Fixed in ee7762c829. The V1 read-back now persists with `changedKeys: ['settings', 'status']`, so `status.passphraseProtection` (and other status fields written by V1 settings) lands in `deviceState` even when no usable DEVICE.STATE event fired. This closes the revert path where the optimistic atom flipped back on the next `HardwareFeaturesUpdate` refresh.

packages/kit/src/states/jotai/contexts/deviceDetails/deviceStateManagement.ts
```diff
@@ -35,12 +35,22 @@ export function getDeviceStateSnapshotFromEvent({
...
   ) {
-    return undefined;
+    // 'settings-read' events are authoritative hardware read-backs. When the
+    // SDK cache already holds a device-side change the app never observed
+    // (e.g. BLE initialize runs before event listeners attach), the SDK
+    // force-emits them without bumping revision/updatedAt, so an event with
+    // stamps equal to the current state must still be applied.
+    const isStale =
+      event.source === 'settings-read'
+        ? event.state.updatedAt < currentState.updatedAt
```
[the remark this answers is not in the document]
originalix: Fixed in ee7762c829. The settings-read gate now treats an event as stale when `updatedAt` is older OR (`updatedAt` equal AND `revision` lower); only equal-or-newer revisions at the same timestamp pass, which preserves the forced re-emit case (equal/equal) this relaxation was introduced for. Regression test added ("drops a same-timestamp settings read with a lower revision").

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
originalix: Fixed in ee7762c829 — and thanks for catching that this was a contract violation, not just a stale assertion. `serialNo` / `connectId` are now masked (`***<last4>`) via `maskLogIdentifier` before entering the payload, so the pre-existing `not.toContain('PRO2_SERIAL')` contract passes unchanged; I also extended it with `not.toContain('PRO2_USB')`. The `:1192` exact-match assertion is now `expect.objectContaining({...})` so future diagnostic fields will not break it. On the `kind: "error"` note: that was the `?? 'error'` fallback firing when the mocked `updateDeviceState` resolves `undefined` — misleading, renamed to `'unknown'` (the real error path still logs via `device state persistence failed`). Both files pass locally: 32 suites / 379 tests.

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
originalix: Fixed in ee7762c829. The 4 assertions now explicitly expect `skipFeaturesUpdateEvent: true`, locking in the new contract that `_notifyProtocolV1SettingsSynced` owns the `HardwareFeaturesUpdate` signal on V1 paths. Since the V1 flow intentionally waits + reads back now, the test title and the stale `expect(waitForDeviceStateSync).not.toHaveBeenCalled()` were updated to assert the new behavior (`waitForDeviceStateSync` called, `getDeviceState({scope:'settings'})` called). On the suggested wipeDevice inversion test: `wipeDevice` passes no `preciseUpdateFields`, so `updateDevice` is never invoked on that path today — happy to add one if that combination ever materializes.

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
originalix: Fixed in ee7762c829. The whole sync (queue drain + read-back) is now bounded by an 8s race guard (timer cleared afterwards; the underlying read-only GetFeatures may finish in the background but no longer holds the flow), and `appEventBus.emit(HardwareFeaturesUpdate)` moved into `finally` so the refresh signal fires on every path — including a stuck `waitForDeviceStateSync`, which is now inside the guarded region too. `setDeviceHomeScreen` no longer awaits the notify (fire-and-forget), so the caller resolves as soon as the wallpaper is applied. The extra per-mutation GetFeatures itself is kept by design — it is the mechanism that makes the fix independent of SDK event emission — but its worst case is now ~8s instead of 60s.

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
```diff
@@ -1,9 +1,18 @@
...
+  try {
+    defaultLogger.hardware.sdkLog.serviceEvent(
+      name,
+      args.length <= 1 ? args[0] : args,
+    );
+  } catch {
+    // Logging must never break hardware flows.
+    console.log(`ServiceHardwareLog@${name}`, ...args);
+  }
 }
```
sidmorizon, line 23: **Hardware diagnostic logs stop appearing in developer builds**

Severity: `non-severe`

Hardware event messages are now sent only through the scene-gated file logger (`defaultLogger.hardware.sdkLog.serviceEvent` at `packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts:8`) instead of the console, so in developer builds they are silently discarded unless a hidden logging switch is turned on.
Impact: Developers debugging hardware/firmware flows see no hardware logs at all by default, making device issues much harder to diagnose.

<details><summary>Gating mechanism in the logger pipeline</summary>

`hardwareLog` is used by ~41 call sites across `ServiceHardware.ts`, `ServiceFirmwareUpdate.ts` and `DeviceSettingsManager.ts`. Previously each call unconditionally did `console.log`. Now it calls `HardwareSDKScene.serviceEvent`, which is decorated with `@LogToLocal`. In `packages/shared/src/logger/base/logFn.ts:65` the entry is processed with `loggerConfig.shouldLog(scope, scene)`, and `LoggerConfigManager.shouldLog` (`packages/shared/src/logger/config/loggerConfigManager.ts:86-95`) returns `true` unconditionally only when `!isDev`; in dev it requires `enabled.hardware.sdkLog` (default `{}` → false) or the newly added `enableAllScenes` flag. The `catch` fallback to `console.log` only fires if the logger throws, which it does not for a disabled scene. Note this also contradicts the PR description's claim that hardwareLog is "not gated by the logger opt-in config".

</details>

Suggested fix:

```suggestion
function hardwareLog(name: string, ...args: any[]) {
  try {
    defaultLogger.hardware.sdkLog.serviceEvent(
      name,
      args.length <= 1 ? args[0] : args,
    );
  } catch {
    // Logging must never break hardware flows.
  }
  if (platformEnv.isDev) {
    // Keep the always-on dev console trace: the file transport is gated by
    // the per-scene logger opt-in config in dev builds.
    console.log(`ServiceHardwareLog@${name}`, ...args);
  }
}
```

packages/kit/src/states/jotai/contexts/deviceDetails/actions.ts
```diff
@@ -311,13 +324,22 @@ class DeviceDetailsActions extends ContextJotaiActionsBase {
...
             return data;
           }
-          set(deviceStateSnapshotAtom(), snapshot);
+          set(
+            deviceStateSnapshotAtom(),
+            pickNewerDeviceStateSnapshot({
+              current: get(deviceStateSnapshotAtom()),
+              incoming: snapshot,
+            }),
+          );
```
sidmorizon, line 343: **Device details page can keep showing outdated device info when the device cannot be read**

Severity: `non-severe`

The previously shown device information is kept (`pickNewerDeviceStateSnapshot` at `packages/kit/src/states/jotai/contexts/deviceDetails/actions.ts:327-333`) when a refresh fails to read the device, instead of falling back to the stored information, so the page can keep displaying outdated values such as the old firmware version.
Impact: After a firmware update (or any refresh where the device read fails), the device page may keep showing pre-update details until another device event arrives.

<details><summary>Why the incoming-undefined case regresses</summary>

`getDeviceManagementSnapshot` is wrapped in `.catch(() => undefined)` (`actions.ts:318-323`), so a disconnected/rebooting device yields `incoming: undefined`. `pickNewerDeviceStateSnapshot` returns `current` unchanged in that case (`packages/kit/src/states/jotai/contexts/deviceDetails/deviceStateManagement.ts:138-143`). Previously the atom was set to `undefined`, which made `resolveDeviceState` (`deviceStateManagement.ts:113`) fall back to the persisted `device.deviceStateInfo` read in the same refresh. Now the retained snapshot always wins over the persisted state — including when the persisted state is newer — because the comparison only happens between snapshots, never against the DB state. The firmware-update refresh path (`packages/kit/src/views/DeviceManagement/pages/DeviceDetailsModal/index.tsx:146-148`) is the one refresh that does not pass `skipDeviceStateSnapshot`, and it runs right after the device reboots, i.e. exactly when the live read is most likely to fail.

A fix would compare the retained snapshot against `data.device.deviceStateInfo` (or clear it when the live read failed and the persisted state is newer) rather than unconditionally keeping the old snapshot.

</details>

originalix (commented):
@originalix

Reviewed commit `0fab39a47aeb`.

- **P1 · Wait for the post-read event before refreshing main**
- **P1 · Redact hardware diagnostics before persistent logging**
- **P2 · Propagate the logger switch to the background runtime**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12879#issuecomment-5290528140)

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -140,6 +149,105 @@ type ITrezorDeviceSettingsAction = (params: {
...
+        error instanceof Error ? error.message : error,
+      );
+    } finally {
+      if (timeoutId !== undefined) {
+        clearTimeout(timeoutId);
+      }
+      // Consumers re-read whatever the DB holds; this signal must fire on
+      // every path, especially when updateDevice suppressed its own event
+      // via skipFeaturesUpdateEvent.
+      appEventBus.emit(EAppEventBusNames.HardwareFeaturesUpdate, {
```
sidmorizon: # 🟡 P2 `finally` 里未保护的 `emit` + `setHomeScreen` 的 `void` 调用，组合出未捕获的 promise rejection

## 问题

`ee7762c8` 把 `appEventBus.emit(HardwareFeaturesUpdate)` 移进 `finally`（这个改动本身是对的，保证每条路径都发信号），但这行**没有 try/catch 包裹**，而 `appEventBus.emit` 是会同步抛出的：`emit()` → `emitToSelf()` → `super.emit(type, payloadCloned)`（`packages/shared/src/eventBus/appEventBus.ts:897`）直接调用 EventEmitter，中间没有任何 try/catch，订阅者同步抛出就会一路冒到调用方。

这不是理论推测——**同一个 PR 里、同一条事件链上已经有对应的防护先例**，`ServiceHardware.ts:1495-1505`：

```ts
try {
  appEventBus.emit(EAppEventBusNames.HardwareDeviceStateUpdate, event);
} catch (error) {
  serviceHardwareUtils.hardwareLog('device state subscriber failed', error);
}
```

`HardwareFeaturesUpdate` 目前在 `DeviceDetailsModal/index.tsx:155`、`AccountSelectorWalletListSideBar.tsx:124`、`ServiceHardwareUI.ts:101` 都有订阅者；desktop / web 是单 runtime，UI 订阅者与 `emit` 在同一个 JS 堆里同步执行。

## 影响

两条路径的后果不同，且都是本次改动新引入的：

1. `:772` 的 `void this._notifyProtocolV1SettingsSynced({ device })`（`setHomeScreen`）——改成 fire-and-forget 后**没有 `.catch()`**。`finally` 里抛出会让这个 promise 变成**无 handler 的 rejection**。改动前是 `await`，rejection 至少有调用方接住；现在在 RN/Hermes 会触发 unhandled rejection 警告，扩展的 service worker 里会产生 `unhandledrejection`，也会污染 Sentry。
2. `:450` 的 `await this._notifyProtocolV1SettingsSynced(...)` 仍在 `withHardwareProcessing` 内——订阅者抛出会让**已经成功写入设备的 V1 设置变更**对用户报错。注意这行在 `finally` 中，`try/catch` 已经吞掉了原始异常，所以这里抛出的是一个纯粹由「刷新信号」引发的失败。

顺带说明：`Promise.race` 的败者不会造成额外 unhandled rejection（race 会给每个 promise 挂 handler），所以 8s guard 超时后 `getDeviceState` 再 reject 是安全的——这块没问题。

## 建议

按同文件已有的先例把 emit 包起来，并给 fire-and-forget 调用补一个 `.catch()`：

```ts
} finally {
  if (timeoutId !== undefined) {
    clearTimeout(timeoutId);
  }
  try {
    appEventBus.emit(EAppEventBusNames.HardwareFeaturesUpdate, {
      deviceId: device.id,
    });
  } catch (error) {
    serviceHardwareUtils.hardwareLog('v1 settings refresh subscriber failed', error);
  }
}
```

`:772` 同时改成 `void this._notifyProtocolV1SettingsSynced({ device }).catch(() => undefined);`，这样即便将来 `finally` 里再加别的语句也不会漏出去。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
originalix: Fixed in 69e15c5639, with a small refinement over the suggestion: the dev console trace is restored, but only when `loggerConfig.shouldLog('hardware', 'sdkLog')` is off — when the scene (or the new `enableAllScenes` switch) is enabled, the gated transport already mirrors to the console in dev, so this avoids double-printing every hardware line. Net behavior: dev builds always see `ServiceHardwareLog@...` in the console again, file persistence stays opt-in.

packages/kit/src/states/jotai/contexts/deviceDetails/actions.ts
originalix: Fixed in 69e15c5639. When the live snapshot read fails (`incoming: undefined`), the refresh now falls back to comparing the retained snapshot against the persisted `device.deviceStateInfo` (`incoming: snapshot ?? {state: deviceStateInfo}`), so a newer DB state (e.g. the post-firmware-update version persisted by bg events) wins over a stale retained snapshot, while a genuinely newer applied-event snapshot is still preserved. This keeps the monotonicity guarantee without letting the snapshot outlive newer persisted data.

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
originalix: Fixed in 69e15c5639, following the in-file precedent: the `finally` emit is now wrapped in try/catch (logged as `v1 settings refresh subscriber failed`, dev-only payload), and the `setDeviceHomeScreen` fire-and-forget call got an explicit `.catch(() => undefined)` so nothing can leak as an unhandled rejection even if more statements land in that `finally` later. Thanks for confirming the `Promise.race` loser is already safe — left as is.

originalix:
Thanks for the review — it ran against `0fab39a`; all three issues are addressed as of `69e15c56`:

**P1 · Wait for the post-read event before refreshing main** — Fixed. `_syncProtocolV1SettingsSnapshot` now drains `waitForDeviceStateSync()` a second time (same identity keys) after the read-back `getDeviceState()` + direct persist, so an event queued by the read-back itself lands before `HardwareFeaturesUpdate` fires. The whole sequence sits inside the 8s bounded guard added in `ee7762c8`, and the refresh signal is emitted in a protected `finally` on every path. Note the direct-persist-rejected-as-stale scenario also narrowed since this report: V1 `initialize`/`settings-write`/`settings-read` events are all authoritative for the settings section now, so equal-stamp DB states already carry the language. The delayed-persistence manager-level integration test is acknowledged as a gap — the current coverage asserts the drain + read-back calls (`DeviceSettingsManager.pro2.test.ts`); happy to add a delayed-task variant in a follow-up.

**P1 · Redact hardware diagnostics** — Fixed across `ee7762c8` + `69e15c56`: `serialNo`/`connectId` masked to `***<last4>` (the pre-existing `PRO2_SERIAL` log contract test now also guards `PRO2_USB`), the full `features` dump and all raw error payloads (`device state persistence failed`, `device state subscriber failed`, `v1 settings read-back failed`, `v1 settings refresh subscriber failed`) are wrapped in `devOnlyData()`, so production logs keep event names + non-identifying fields (language/revision/source/kind) only. A fully typed whitelist payload for all ~41 call sites is a larger refactor left as a follow-up; every payload this PR adds is already an explicit small object.

**P2 · Propagate the logger switch to bg** — Fixed. Added `serviceLogger.updateLoggerConfig()` (bg method that saves via the bg-runtime `LoggerConfigManager`), and both UI entry points (DevSettings `Persist all logs`, Gallery logger page) now route through `saveLoggerConfigToAllRuntimes()`: bg saves + persists first, then the main-runtime singleton is mirrored via `updateRuntimeConfig()` without a second storage write. Single-context platforms hit the same singleton twice, which is harmless.

Validation: 40 suites / 386 tests pass locally (including the previously failing `ServiceHardware.pro2DeviceManagement.test.ts` and `DeviceSettingsManager.pro2.test.ts`), `agent:check --profile commit` green.

originalix (commented):
@originalix

Reviewed commit `69e15c5639f0`.

- **P1 · Do not persist legacy hardwareLog payloads verbatim**
- **P1 · Notify again when a timed-out sync finishes**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12879#issuecomment-5290528140)

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
```diff
@@ -1,9 +1,36 @@
...
+  } catch {
+    // Logging must never break hardware flows.
+  }
+  // Keep the always-on dev console trace: the scene-gated transport above
+  // mirrors to the console itself when enabled, so only fill the gap when
+  // the hardware scene is off.
+  if (platformEnv.isDev && !loggerConfig.shouldLog('hardware', 'sdkLog')) {
+    console.log(`ServiceHardwareLog@${name}`, ...args);
+  }
+}
```
sidmorizon, line 23: # 🟠 P1 `hardwareLog` 改为生产无条件落盘后，`ServiceHardware.ts:2591` / `:2721` 的 `connectId` 以原文进入持久化诊断日志

（本想把这条锚在 `ServiceHardware.ts:2591`，但那两行不在本 PR 的 diff hunk 内、GitHub 不接受行级评论，所以锚在根因这里。）

## 问题

先把因果说清楚，避免误读：`ServiceHardware.ts:2591` 的 `hardwareLog('read legacy app features', connectId)` 和 `:2721` 的 `hardwareLog('call getDeviceState()', connectId)` 都是**既有代码，本 PR 没有改动它们**。真正的变化就是本函数 —— base `e6b810cc` 上 `hardwareLog` 的完整实现只有一行：

```ts
function hardwareLog(name: string, ...args: any[]) {
  console.log(`ServiceHardwareLog@${name}`, ...args);
}
```

即**纯 console、从不落盘**。本 PR 把它改路由到 `defaultLogger.hardware.sdkLog.serviceEvent(...)`，而该方法带 `@LogToLocal({ level: 'info' })`（`packages/shared/src/logger/scopes/hardware/scenes/sdk.ts:113`）。是否落盘取决于 `LoggerConfigManager.shouldLog()`，它的首句是：

```ts
shouldLog(scopeName: string, sceneName: string): boolean {
  if (!this._env.isDev) {
    return true;
  }
```

（`packages/shared/src/logger/config/loggerConfigManager.ts:86-89`）—— **生产构建无条件返回 `true`**。于是全部 41 个 `hardwareLog` 调用点的既有参数，从「只在开发者控制台一闪而过」变成了「写入可导出的持久化诊断日志」，而上面两处的 `connectId` 都是原文，未经 `maskLogIdentifier`。

## 影响

`maskLogIdentifier` 目前只覆盖 `ServiceHardware.ts:1427-1428` 的 `device state update` payload —— 恰好就是被 `ServiceHardware.pro2DeviceManagement.test.ts:1200-1204` 那条断言打红的那一条：

```ts
// Device identifiers must never enter hardwareLog unmasked.
expect(JSON.stringify(hardwareLogSpy.mock.calls)).not.toContain('PRO2_SERIAL');
```

这条契约随 #11814 引入、早于本 PR，说明「设备标识不得进入 `hardwareLog`」是仓库既有约束而非评审偏好。问题在于它只断言了这一个用例路径，所以脱敏也只走到了那一条 payload。`_getLegacyAppFeatures`（`:2591`）和 `_getDeviceStateLowLevel`（`:2721`）不在该用例覆盖范围内 —— 我在 `69e15c56` 上实跑 `ServiceHardware.pro2DeviceManagement.test.ts` 是 `39 passed / 0 failed`，测试全绿，但这两处暴露面仍在。相关 Codex thread 已 resolve，从 PR 界面看这个问题呈「已关闭」状态，实际是被测试覆盖到的那一条修好了、其余未修完。

这两条是**每一次** V1 GetFeatures / getDeviceState 都会打的日志，属于最高频路径；而本 PR 新增的 V1 设置回读还会让 `getDeviceState({ scope: 'settings' })` 的调用频次进一步上升。结果是设备唯一标识（USB 序列号 / BLE MAC 形态的 `connectId`）以原文长期驻留在用户日志中，并随「导出日志」一并交出。

## 建议

最小修法是把已有的脱敏函数覆盖到这两处：

```ts
serviceHardwareUtils.hardwareLog(
  'read legacy app features',
  serviceHardwareUtils.maskLogIdentifier(connectId),
);
```

`:2721` 同理。但逐点补容易漏 —— 41 个调用点里除 `connectId` 外还有传完整 `features`、传原始 error 对象的（这两类本轮已分别用 `devOnlyData()` 处理，但同样是靠逐点记得）。

更稳妥的是在本函数入口统一处理：它已经是所有硬件日志的唯一收口，在写入 `serviceEvent` 之前对 payload 统一做一次标识符脱敏（或对纯字符串参数默认走 `maskLogIdentifier`），这样后续新增调用点不必再逐个记得脱敏，也不会重演「测试覆盖到哪条就修哪条」。入口级处理还能把 `connectId` / 完整 `features` / 原始 error 三类 payload 一次性收敛 —— 这三类正是同一个入口放大出来的，分头打补丁很难保证不漏。

[Claude session ↗](https://claude.ai/code/session_01LghrzkJHibeDw7Eet4w7m3)

---

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -579,70 +747,51 @@ export class DeviceSettingsManager extends ServiceHardwareManagerBase {
...
+          },
+          debugMethodName: 'deviceSettings.applySettingsToDevice',
+        },
+      );
+    if (!this._isProtocolV2Product(device) && !this._isTrezorDevice(device)) {
+      // Fire-and-forget: the wallpaper is already applied and the processing
+      // dialog closed; the caller must not stay pending on the read-back.
+      void this._notifyProtocolV1SettingsSynced({ device }).catch(
+        () => undefined,
+      );
```
sidmorizon, line 792: **Wallpaper change triggers an unguarded extra device read that can collide with the next hardware operation**

Severity: `non-severe`

After a wallpaper is applied to a legacy OneKey device, an extra read of the device is started in the background (`void this._notifyProtocolV1SettingsSynced({ device })` at `packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts:789`) outside the section that guarantees only one device operation at a time, so it can overlap with whatever the user starts next on the same device.
Impact: Right after setting a wallpaper, the next hardware action (for example a signing request or a firmware update) can fail or show a confusing "device busy" error.

<details><summary>Mechanism: read-back escapes runExclusiveOneKeyOperation and the firmware-update busy guard</summary>

Every other device communication in this manager runs inside `serviceHardwareUI.withHardwareProcessing`, which serializes OneKey operations through `runExclusiveOneKeyOperation` and rejects calls while a firmware-update workflow is running (`packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts:531-562`).

The new fire-and-forget call is placed *after* `withHardwareProcessing` has resolved, so:
- it holds no operation lease, and `serviceHardware.getDeviceState` only takes `getFeaturesMutex` (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2760-2786`), which does not exclude other SDK traffic such as signing or firmware transfer;
- it is not subject to the `firmwareUpdateWorkflowRunningAtom` / `firmwareUpdateExclusiveDepth` busy check;
- it can stay in flight for up to the 8s timeout guard in `_notifyProtocolV1SettingsSynced` (`packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts:172-183`), long enough for the user to start another action.

The V1 read-back performed inside `_withDeviceProcessing` (line 468) does not have this problem because it still runs within the exclusive section.

</details>

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
sidmorizon, line 23: **Hardware diagnostic messages, including device identifiers, are now written into the log files users can upload**

Severity: `non-severe`

Hardware diagnostic messages are now written to the persistent log transport (`defaultLogger.hardware.sdkLog.serviceEvent(...)` at `packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts:10`) instead of only the developer console, and existing call sites still pass raw device identifiers and full payloads, so those identifiers end up in the log bundles users can export or upload.
Impact: A user's device identifiers and verbose hardware payloads become part of the log files shipped off the device, which the repository rules forbid.

<details><summary>Mechanism: `local` log type persists in production; callers pass unmasked identifiers</summary>

`hardwareLog` previously only did `console.log`. It now routes through `HardwareSDKScene.serviceEvent`, decorated with `@LogToLocal` (`packages/shared/src/logger/scopes/hardware/scenes/sdk.ts:113-116`). In `processEntry`, the `local` branch writes to the file-backed logger whenever `shouldLog` is true, and `shouldLog` returns `true` unconditionally in non-dev builds (`packages/shared/src/logger/config/loggerConfigManager.ts:86-95`, `packages/shared/src/logger/base/logFn.ts:88-100`).

Existing callers were not audited for this new persistence:
- `serviceHardwareUtils.hardwareLog('read legacy app features', connectId)` (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2591`)
- `serviceHardwareUtils.hardwareLog('call getDeviceState()', connectId)` (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:2721`)
- numerous `ServiceFirmwareUpdate` calls passing whole payload/error objects.

This contradicts the masking contract this same PR introduces and asserts in tests (`maskLogIdentifier` plus `expect(JSON.stringify(hardwareLogSpy.mock.calls)).not.toContain('PRO2_USB')`), and CLAUDE.md states "Never log sensitive data". Elsewhere in the PR the authors correctly wrap payloads in `devOnlyData(...)`, which redacts in production; the untouched call sites do not.

</details>

wabicai (approved):

PR #12892, opened 2026-08-15
fix: sync hardware connection status after reset(OK-60117)

originalix:
[OK-60117](https://onekeyhq.atlassian.net/browse/OK-60117)

----------

## Summary

- Track hardware identities reported by the SDK connection lifecycle.
- Merge background SDK identities with WebUSB enumeration when computing wallet connection indicators.
- Refresh the wallet connection indicator when the background hardware connection state changes.

## Intent & Context

Fix OK-60117 on Desktop 6.5.2. After resetting a OneKey Pro, re-importing the wallet, and successfully communicating with the device, the wallet selector could still omit the green connected indicator until macOS re-authorized and re-enumerated the USB device.

## Root Cause

The UI connection hook only used `navigator.usb.getDevices()` and matched its serial number against the wallet's stored hardware identity. A reset changes the device identity, while WebUSB can retain a stale or unavailable identity until a new macOS USB permission flow occurs. The background hardware service already had the current SDK connection identities after successful communication, but the UI did not consume that state.

## Design Decisions

- Keep WebUSB enumeration as a fallback for existing OneKey and Trezor behavior.
- Expose the SDK-tracked identity snapshot from the background service instead of duplicating hardware communication in the UI runtime.
- Emit a payload-free cross-runtime refresh event on SDK connect and disconnect; device identities remain available only through the background method.
- Match only known hardware identity keys and preserve hidden-wallet filtering.

## Changes Detail

- Add `HardwareConnectionStateUpdate` to the typed app event bus.
- Emit connection-state refresh events and expose connected hardware identity keys from `ServiceHardware`.
- Merge SDK identity keys with supported WebUSB serials in `useHardwareWalletConnectStatus`.
- Add regression coverage for a reset device with stale WebUSB identity and for connect/disconnect identity tracking.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Desktop and WebUSB-capable clients
- **Risk Areas**: Hardware identity matching and cross-runtime connection-state refresh timing

## Test plan

- [x] Reset a OneKey Pro, re-import the wallet, complete device communication, and verify the green indicator appears without another USB permission flow.
- [x] Verify the green indicator disappears after clearing the SDK connection state.
- [x] Run the focused hardware connection test suites (46 tests).
- [x] Run `yarn agent:check --profile commit` on the latest `hotfix/v6.5.2` base.
- [x] Verify the QA bundle on Desktop; bundle version 1282 fixed the reported issue.

## QA Evidence

- QA bundle run: https://github.com/OneKeyHQ/app-monorepo/actions/runs/31787864368
- Desktop QA bundle: https://bundle-test.onekey-asset.com/utility/js-bundle/6.5.2/966eba6c-1282/electron.zip
- The QA bundle contains the functional fix commit before rebasing onto the latest `hotfix/v6.5.2`; the rebased PR head was revalidated with the focused tests and commit checks.

[OK-60117]: https://onekeyhq.atlassian.net/browse/OK-60117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
+++ b/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
@@ -1603,6 +1603,10 @@ class ServiceHardware extends ServiceBase {
         this.recordLiveConnectIdEvidence(message.device?.connectId);
         const connectedIdentityKeys = this.trackConnectedDevice(message.device);
         if (connectedIdentityKeys.length > 0) {
+          appEventBus.emit(
+            EAppEventBusNames.HardwareConnectionStateUpdate,
+            undefined,
+          );
           void this.backgroundApi.serviceHardwarePortfolioSync
             ?.notifyHardwareDeviceConnected({
               identityKeys: connectedIdentityKeys,
@@ -1676,6 +1680,10 @@ class ServiceHardware extends ServiceBase {
           message.device,
         );
         if (disconnectedIdentityKeys.length > 0) {
+          appEventBus.emit(
+            EAppEventBusNames.HardwareConnectionStateUpdate,
+            undefined,
+          );
           void this.backgroundApi.serviceHardwarePortfolioSync
             ?.notifyHardwareDeviceDisconnected({
               identityKeys: disconnectedIdentityKeys,
@@ -1864,6 +1872,17 @@ class ServiceHardware extends ServiceBase {
   @backgroundMethod()
   async isDeviceSearchInProgress() {
     return this.deviceSearchInProgressCount > 0;
+  }
+
+  @backgroundMethod()
+  async getConnectedHardwareDeviceIdentityKeys() {
+    return [
+      ...new Set(
+        [...this.connectedDeviceIdentityKeysByConnection.values()].flatMap(
+          (identityKeys) => [...identityKeys],
+        ),
+      ),
+    ];
   }
 
   @backgroundMethod()
--- a/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
+++ b/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
@@ -861,16 +861,42 @@ class ServiceHardware extends ServiceBase {
     );
   }
 
-  private trackConnectedDevice(device: KnownDevice | undefined) {
+  private trackConnectedDevice(device: KnownDevice | undefined): {
+    identityKeys: string[];
+    identityKeysChanged: boolean;
+  } {
     const identityKeys = this.getConnectedDeviceIdentityKeys(device);
     const connectionKey = device?.connectId || identityKeys[0];
-    if (connectionKey && identityKeys.length > 0) {
+    if (!connectionKey || identityKeys.length === 0) {
+      return { identityKeys, identityKeysChanged: false };
+    }
+    const existingIdentityKeys =
+      this.connectedDeviceIdentityKeysByConnection.get(connectionKey);
+    const identityKeysChanged =
+      !existingIdentityKeys ||
+      existingIdentityKeys.size !== identityKeys.length ||
+      identityKeys.some((key) => !existingIdentityKeys.has(key));
+    if (identityKeysChanged) {
       this.connectedDeviceIdentityKeysByConnection.set(
         connectionKey,
         new Set(identityKeys),
       );
     }
-    return identityKeys;
+    return { identityKeys, identityKeysChanged };
+  }
+
+  private clearTrackedConnectedDevices() {
+    if (this.connectedDeviceIdentityKeysByConnection.size === 0) {
+      return;
+    }
+    // The identity map is a bg-runtime JS copy; UI runtimes cache their own
+    // snapshot, so every clear must broadcast or the green indicator goes
+    // stale after an SDK reset or transport switch.
+    this.connectedDeviceIdentityKeysByConnection.clear();
+    appEventBus.emit(
+      EAppEventBusNames.HardwareConnectionStateUpdate,
+      undefined,
+    );
   }
 
   private untrackConnectedDevice(device: KnownDevice | undefined) {
@@ -1007,6 +1033,7 @@ class ServiceHardware extends ServiceBase {
       refreshedFirmwareManifestKey !== this.loadedFirmwareManifestKey
     ) {
       await resetHardwareSDKInstance();
+      this.clearTrackedConnectedDevices();
       this.registeredEvents = false;
     }
 
@@ -1073,6 +1100,7 @@ class ServiceHardware extends ServiceBase {
       );
       this.resetHardwareUiEventQueue();
       await resetHardwareSDKInstance();
+      this.clearTrackedConnectedDevices();
       this.registeredEvents = false;
       this.activeHardwareSDKInstance = undefined;
       console.log('✅ TRANSPORT SWITCH: SDK reset completed');
@@ -1283,7 +1311,7 @@ class ServiceHardware extends ServiceBase {
   ) {
     if (this.registeredSdkEventsInstance !== instance) {
       this.resetHardwareUiEventQueue();
-      this.connectedDeviceIdentityKeysByConnection.clear();
+      this.clearTrackedConnectedDevices();
       this.registeredEvents = false;
     }
 
@@ -1585,6 +1613,19 @@ class ServiceHardware extends ServiceBase {
             return;
           }
 
+          // DEVICE.CONNECT can fire before features are complete, so the
+          // tracked identity may miss the raw deviceId; re-track once features
+          // arrive so deviceId-based consumers see the device as connected.
+          const { identityKeysChanged } = this.trackConnectedDevice(
+            (message.device ?? undefined),
+          );
+          if (identityKeysChanged) {
+            appEventBus.emit(
+              EAppEventBusNames.HardwareConnectionStateUpdate,
+              undefined,
+            );
+          }
+
           // TODO: save features to dbDevice
           // Full features dumps are dev-only; production logs keep the event
           // name without the device blob.
@@ -1601,7 +1642,8 @@ class ServiceHardware extends ServiceBase {
 
       instance.on(DEVICE.CONNECT, (message: { device: KnownDevice }) => {
         this.recordLiveConnectIdEvidence(message.device?.connectId);
-        const connectedIdentityKeys = this.trackConnectedDevice(message.device);
+        const { identityKeys: connectedIdentityKeys } =
+          this.trackConnectedDevice(message.device);
         if (connectedIdentityKeys.length > 0) {
           appEventBus.emit(
             EAppEventBusNames.HardwareConnectionStateUpdate,
@@ -1775,6 +1817,7 @@ class ServiceHardware extends ServiceBase {
     await this.backgroundApi.serviceHardwareUI.runExclusiveOneKeyOperation(() =>
       this.sdkInstanceMutex.runExclusive(async () => {
         this.resetHardwareUiEventQueue();
+        this.clearTrackedConnectedDevices();
         this.registeredEvents = false;
         await resetHardwareSDKInstance();
         this.activeHardwareSDKInstance = undefined;
--- a/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
+++ b/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
@@ -1617,7 +1617,7 @@ class ServiceHardware extends ServiceBase {
           // tracked identity may miss the raw deviceId; re-track once features
           // arrive so deviceId-based consumers see the device as connected.
           const { identityKeysChanged } = this.trackConnectedDevice(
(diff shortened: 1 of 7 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

@originalix

Reviewed commit `c31ee790ffbf`. This report always reflects the latest reviewed changes.

### Review summary

The PR makes the background hardware service the supplemental source of SDK-observed device identities and has the foreground merge serialized background snapshots with WebUSB identities. Connection, feature-completion, disconnection, SDK-reset, firmware-manifest-reset, and transport-switch transitions now invalidate the foreground snapshot through a typed cross-runtime event. This preserves the existing package and security boundaries, but overlapping foreground refreshes can commit snapshots out of order and restore stale connection state.

**What needs attention:** Verify that back-to-back connection and feature-completion invalidations always converge on the newest background identity snapshot.

### Issues to address

- **P2 · Overlapping refreshes can restore stale hardware identities**
  - **Impact:** DEVICE.CONNECT can start a refresh with a connectId-only background snapshot, followed by DEVICE.SUPPORT_FEATURES starting another refresh with the enriched deviceId. Because each refresh independently awaits WebUSB enumeration and the background RPC and then commits without sequencing, the older result can finish last and replace the enriched Set. A reset wallet that matches only the feature-derived deviceId can then remain displayed as disconnected until another event or manual refresh.
  - **Suggested action:** Add a monotonically increasing request generation to refreshDevices so only the latest started request can update the ref and state, then add a hook-level regression test that resolves two refreshes in reverse order.

### Validation gaps

- `git diff --check origin/hotfix/v6.5.2...HEAD` completed without errors.
- Targeted Jest and TypeScript checks could not run because this workspace has no installed dependencies or Yarn install-state file.
- The added tests cover background identity enrichment, lifecycle clearing, and pure identity merging, but not overlapping foreground refreshes or the complete event-to-RPC transition.
- Existing GitHub review and thread coverage could not be checked because API connectivity was unavailable.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1676,6 +1680,10 @@ class ServiceHardware extends ServiceBase {
           message.device,
         );
         if (disconnectedIdentityKeys.length > 0) {
+          appEventBus.emit(
+            EAppEventBusNames.HardwareConnectionStateUpdate,
+            undefined,
+          );
```
sidmorizon, line 1728: **Connection indicator can keep showing a device as connected after the hardware layer is restarted**

Severity: `non-severe`

The list of currently connected devices is wiped in the background (`connectedDeviceIdentityKeysByConnection.clear()` at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:1286`) without telling the screens that the connection state changed, so the wallet list can keep showing a device as connected when it is no longer tracked.
Impact: After the hardware layer restarts (e.g. transport switch or SDK reset), the green "connected" dot may stay on for a device that is no longer known to be connected until the next physical connect/disconnect event.

<details>
<summary>Why the UI snapshot goes stale when the tracked identity map is cleared</summary>

The new UI hook (`packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:33-46`) caches the merged identity-key set and only re-fetches it when `EAppEventBusNames.HardwareConnectionStateUpdate` is emitted or when a WebUSB connect/disconnect event fires. The PR emits that event only from `DEVICE.CONNECT` (`ServiceHardware.ts:1606-1609`) and `DEVICE.DISCONNECT` (`ServiceHardware.ts:1683-1686`).

`registerSdkEvents` clears the whole tracking map when the SDK instance changes (`ServiceHardware.ts:1284-1287`), and `resetHardwareSDK` (`ServiceHardware.ts:1766-1776`) tears the instance down; neither path emits `HardwareConnectionStateUpdate`. On Desktop there is no guaranteed WebUSB event in these flows, so the foreground keeps the pre-clear identity keys and the indicator remains inconsistent with the background state until a new connect/disconnect event arrives.

</details>

packages/kit/src/hooks/useHardwareWalletConnectStatus.ts
```diff
@@ -25,19 +30,18 @@ async function fetchConnectedDevices(): Promise<Set<string>> {
...
-    }
-  }
+  const [webUsbDevices, backgroundIdentityKeys] = await Promise.all([
+    usb && typeof usb.getDevices === 'function'
+      ? usb.getDevices()
+      : Promise.resolve([]),
+    backgroundApiProxy.serviceHardware
+      .getConnectedHardwareDeviceIdentityKeys()
+      .catch(() => []),
+  ]);
```
sidmorizon, line 40: **Fallback empty arrays are untyped, violating the repository's TypeScript precision rule**

Severity: `non-severe`

The two fallback empty arrays used when device enumeration is unavailable or the background call fails (`Promise.resolve([])` and `.catch(() => [])` at `packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:33-40`) are inferred as `never[]`, which the repository's agent instructions explicitly forbid.
Impact: The code conflicts with a mandatory repository coding rule, weakening type safety around the connected-device data flow.

<details>
<summary>Rule reference and inferred types</summary>

`CLAUDE.md` ("Core Rules") states: "TypeScript must stay precise: no fallback `never[]`, no avoidable `any`". Here `Promise.resolve([])` is inferred as `Promise<never[]>` and `.catch(() => [])` widens the result to `Promise<string[] | never[]>`. Both should carry explicit element types, e.g. `Promise.resolve<USBDevice[]>([])` and `.catch((): string[] => [])`.

</details>

Suggested fix:

```ts
  const [webUsbDevices, backgroundIdentityKeys] = await Promise.all([
    usb && typeof usb.getDevices === 'function'
      ? usb.getDevices()
      : Promise.resolve<USBDevice[]>([]),
    backgroundApiProxy.serviceHardware
      .getConnectedHardwareDeviceIdentityKeys()
      .catch((): string[] => []),
  ]);
```

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1866,6 +1874,17 @@ class ServiceHardware extends ServiceBase {
     return this.deviceSearchInProgressCount > 0;
   }
 
+  @backgroundMethod()
+  async getConnectedHardwareDeviceIdentityKeys() {
+    return [
+      ...new Set(
+        [...this.connectedDeviceIdentityKeysByConnection.values()].flatMap(
+          (identityKeys) => [...identityKeys],
+        ),
```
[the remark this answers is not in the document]
originalix: 已在 8b398df3f8 修复。新增幂等的 `clearTrackedConnectedDevices()`:清空身份 Map,且仅在原本非空时广播 `HardwareConnectionStateUpdate`;并接入全部 SDK 重置/实例替换路径——`registerSdkEvents` 的实例替换分支(原来只 clear 不广播)、固件清单变化重建、transport 切换重建、`resetHardwareSDK()`。已补充实例替换与 `resetHardwareSDK` 两个回归测试,验证清空后 `getConnectedHardwareDeviceIdentityKeys()` 返回空并收到广播。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1603,6 +1603,10 @@ class ServiceHardware extends ServiceBase {
         this.recordLiveConnectIdEvidence(message.device?.connectId);
         const connectedIdentityKeys = this.trackConnectedDevice(message.device);
         if (connectedIdentityKeys.length > 0) {
+          appEventBus.emit(
+            EAppEventBusNames.HardwareConnectionStateUpdate,
+            undefined,
+          );
```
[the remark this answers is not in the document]
originalix: 已在 8b398df3f8 修复。`trackConnectedDevice` 改为返回 `{ identityKeys, identityKeysChanged }` 做集合变更检测;`DEVICE.SUPPORT_FEATURES` 处理器在 features 与 raw deviceId 就绪后重新跟踪身份,身份集合真正变化时才广播一次 `HardwareConnectionStateUpdate`(相同身份的重复 features 事件不会重复广播)。补充回归测试:CONNECT 先到(无 features)、features 后到时补录 deviceId 并广播,重复事件不再广播。

packages/kit/src/hooks/useHardwareWalletConnectStatus.ts
```diff
@@ -25,19 +30,18 @@ async function fetchConnectedDevices(): Promise<Set<string>> {
...
-      deviceIds.add(key);
-    }
-  }
+  const [webUsbDevices, backgroundIdentityKeys] = await Promise.all([
+    usb && typeof usb.getDevices === 'function'
+      ? usb.getDevices()
+      : Promise.resolve([]),
+    backgroundApiProxy.serviceHardware
+      .getConnectedHardwareDeviceIdentityKeys()
+      .catch(() => []),
```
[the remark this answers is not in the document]
originalix: 已在 8b398df3f8 修复:两处 fallback 分别显式标注为 `Promise.resolve<USBDevice[]>([])` 与 `.catch((): string[] => [])`,不再产生 `never[]` 推断。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
originalix: Fixed in 8b398df3f8. Added an idempotent `clearTrackedConnectedDevices()` that clears the identity map and emits `HardwareConnectionStateUpdate` only when the map was non-empty, and wired it into every SDK reset / instance-replacement path: the instance-change branch of `registerSdkEvents` (which previously cleared without broadcasting), the firmware-manifest rebuild, the transport-switch rebuild, and `resetHardwareSDK()`. Regression tests cover instance replacement and `resetHardwareSDK` (map is emptied and the update event is emitted).

packages/kit/src/hooks/useHardwareWalletConnectStatus.ts
originalix: Fixed in 8b398df3f8 using the suggested explicit types: `Promise.resolve<USBDevice[]>([])` and `.catch((): string[] => [])`.

originalix (commented):
@originalix

Reviewed commit `8b398df3f82d`.

- **P2 · Overlapping refreshes can restore stale hardware identities**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12892#issuecomment-5299799720)

originalix (commented):
@originalix

Reviewed commit `c31ee790ffbf`.

- **P2 · Overlapping refreshes can restore stale hardware identities**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12892#issuecomment-5299799720)

packages/kit/src/hooks/useHardwareWalletConnectStatus.ts
sidmorizon, line 1: **Connected-device indicator can stay stuck on the wrong state when refreshes overlap**

Reported at `packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:58-64` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

Each refresh overwrites the stored connection snapshot with whatever it happens to return (`prevDeviceKeysRef.current = newDevices` at `packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:58-64`) with no ordering guard, so when two refreshes overlap the older, out-of-date result can win and the indicator keeps showing the wrong state until another event arrives.
Impact: A user can see a device shown as connected after unplugging it (or vice versa) until some further hardware event happens.

<details>
<summary>Overlapping refresh triggers and last-write-wins snapshot update</summary>

The PR adds a second refresh trigger: `HardwareConnectionStateUpdate` from the background (`packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:86-95`) in addition to the existing WebUSB `connect`/`disconnect` DOM listeners. Both can fire for the same physical unplug within milliseconds.

`fetchConnectedDevices` (`packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:27-47`) awaits `Promise.all` of `navigator.usb.getDevices()` and a cross-runtime background call `serviceHardware.getConnectedHardwareDeviceIdentityKeys()`. Because two in-flight invocations have independent latencies, the one started first (e.g. from the DOM disconnect event, before the background has processed the SDK `DEVICE.DISCONNECT` and cleared its identity map at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:1721-1734`) can resolve last and write its stale key set into `prevDeviceKeysRef`/state, leaving the indicator green with no further event to correct it.

A monotonically increasing request id captured before the await, with the result discarded when a newer request has started, would make the last-started refresh authoritative.

</details>

zhaono1 (approved):

wabicai (approved):

PR #12894, opened 2026-08-15
fix: use SDK dimensions for Pro2 images

wabicai:
## Summary

- use `@onekeyfe/hd-core` as the single source of Pro2/Neo NFT, thumbnail, and wallpaper dimensions
- surface duplicate Pro2 NFT uploads through the existing hardware error flow
- synchronize Protocol V2 device labels from the SDK's canonical device state
- make the extension offscreen hardware bridge recover after its background port disconnects
- stabilize the Protocol V2 firmware update UI across confirmation, progress events, device reconnects, and installation transitions
- update all App and CLI hardware SDK dependencies to `1.2.0-alpha.128`

## Root cause

The App duplicated Pro2/Neo image-size fallbacks that already belong to the SDK, which allowed the two release-coupled codebases to drift. Protocol V2 firmware installation can also replace the device connection identity while emitting separate tip and progress events; the App previously treated these as unrelated or stale state, leaving the UI on `Checking device`, losing `Confirm on device`, or dropping queued progress during the expected disconnect. The extension offscreen bridge could remain bound to a disconnected background port.

### NFT and wallpaper handling

- remove App-level Pro2/Neo NFT, thumbnail, and wallpaper dimension fallbacks
- keep the existing T1 default size and App-side resizing/cropping flow for bundled/default images
- map duplicate NFT uploads to the existing user-facing hardware error path before attempting redundant work

### Protocol V2 device state

- use canonical SDK state when synchronizing device labels
- preserve the confirmed protocol across ordinary reconnects
- keep device-management consumers synchronized without inferring connection state from unrelated devices

### Extension hardware bridge

- recreate the offscreen bridge after its background port disconnects
- guard and report failed event forwarding instead of leaving an unhandled bridge request

### Firmware update UI

- accept firmware status events when a Protocol V2 device reconnects with a new `connectId`
- preserve firmware tip and progress payloads when their events arrive independently
- retain queued progress through the expected installation disconnect
- transition out of the initial checking state when installation starts
- keep `Confirm on device` visible until real installation progress is observed
- present SafeOS as the primary Protocol V2 firmware version and expose component details only in developer mode

### SDK dependency

- upgrade App, extension, native, desktop, and CLI hardware SDK pins and lockfile entries from `1.2.0-alpha.127` to `1.2.0-alpha.128`

## Impact

Pro2/Neo image handling now follows one SDK-owned contract, duplicate NFT uploads produce a clear error, and device labels remain consistent. Firmware updates no longer appear stuck on checking or lose the device-confirmation state during reconnects. Extension hardware events also recover after an offscreen/background bridge disconnect.

## Validation

- image-size and NFT regression suite: 4 test suites passed, 37 tests passed
- firmware UI and offscreen bridge regression suite: 4 test suites passed, 75 tests passed
- focused Oxlint passed with 0 warnings and 0 errors
- `git diff --check` passed
- `@onekeyfe/hd-core`, `@onekeyfe/hd-web-sdk`, and `@onekeyfe/hd-ble-sdk` version `1.2.0-alpha.128` verified on npm with the `next` dist-tag


packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3311,29 +3311,17 @@ class ServiceHardware extends ServiceBase {
     if (isT1Model) {
       names = T1_HOME_SCREEN_DEFAULT_IMAGES;
     }
-    let size = getHomeScreenSize({
-      deviceType: device.deviceType,
-      homeScreenType,
-      thumbnail: false,
-    });
-    let thumbnailSize = getHomeScreenSize({
+    const size =
```
originalix, line 3411: P1: [Removing the Pro 2 size fallbacks reopens the undefined-config path]

This change drops the local Pro 2 fallback here, and the same assumption is repeated in `getDeviceNftConfig()` below, but this PR does not update the SDK dependency that actually supplies these size tables.

If the current `hotfix/v6.5.2` SDK still returns `undefined` for Pro 2, the failure propagates directly into the callers: wallpaper upload exits early on missing `config.size`, and NFT upload throws `Pro2 NFT thumbnail config is missing` when `thumbnailSize` is absent.

Please keep the local Pro 2 fallback until the SDK bump that guarantees both values is in this branch, or add an explicit guard plus service-level coverage that proves the current SDK always returns Pro 2 wallpaper and NFT dimensions.

originalix:

@wabicai

Reviewed commit `528db713d0ed`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the ownership of Pro 2 image-dimension compatibility inside `ServiceHardware`. It removes the local wallpaper/NFT size fallbacks and delegates those values entirely to `CoreSDKLoader`, while keeping the older Pro2-to-Pro server resource mapping and the T1 monochrome default path. Functionally, the PR is trying to simplify the size-selection path, but it does so without changing the SDK dependency or adding replacement service-level coverage for the Protocol V2 image flows.

**What needs attention:** Confirm that the SDK shipped on `hotfix/v6.5.2` always returns Pro 2 wallpaper and NFT dimensions, because the background service no longer has a local compatibility path.

### Issues to address

- **P1 · Removing the Pro 2 size fallbacks reopens the undefined-config path**
  - **Impact:** This PR deletes the only local compatibility path that supplied Pro 2 wallpaper and NFT dimensions when the SDK had no config, but the diff does not update the SDK dependency that produces those values. If `getHomeScreenSize()` or `getNftSize()` still returns `undefined` on the hotfix branch, the wallpaper flow stops at an early return and the NFT flow throws on missing `thumbnailSize`, so Protocol V2 image upload regresses for existing users.
  - **Suggested action:** Keep the local Pro 2 fallbacks until the SDK bump that guarantees both wallpaper and NFT sizes lands in this branch, or add explicit guards plus service-level tests proving the current SDK always provides those values.

### Validation gaps

- Existing PR comments / threads were checked and none were present.
- Targeted Jest execution could not be completed in this worktree. `yarn jest packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.test.ts packages/shared/src/utils/deviceHomeScreenUtils.test.ts --runInBand` failed because Yarn could not find the `node_modules` state file.
- I did not inspect unrelated repository areas beyond changed files and direct consumers of the returned config objects.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
```diff
@@ -39,33 +39,8 @@ function getHomeScreenServerDeviceType(deviceType: EDeviceType): EDeviceType {
   return isProtocolV2ProductType(deviceType) ? EDeviceType.Pro : deviceType;
 }
 
-function getPro2HomeScreenSizeFallback({
```
sidmorizon, line 42: **Bug-fix change does not record the required case study entry**

Severity: `non-severe`

The repository's mandatory self-testing rule requires every bug fix to append a case entry to `.claude/skills/1k-retrospective/references/case-studies.md`, but this fix (removal of the Pro 2 size fallbacks in `packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts:42`) adds no such entry, so the shared retrospective record stays incomplete.
Impact: The team loses the required written record of this fix, weakening future regression prevention.

<details>
<summary>Rule source and expected artifact</summary>

`.cursor/rules/self-testing.mdc` ("Bug Fix Case Recording") states: "After every bug fix, append to `.claude/skills/1k-retrospective/references/case-studies.md`" with Title/Date/Platforms/Symptom/Root Cause/Fix/Catchable-by fields. Recent bug-fix commits in this repo follow that convention (the file's tail contains entries dated up to 2026-08-13), but this PR touches only `ServiceHardware.ts`, `serviceHardwareUtils.ts`, and `serviceHardwareUtils.test.ts`.

</details>

packages/kit-bg/src/services/ServiceNFT.ts
```diff
@@ -93,21 +67,10 @@ class ServiceNFT extends ServiceBase {
...
-                nftUpdated: true,
-                message: PRO2_NFT_ALREADY_EXISTS_MESSAGE,
-              };
-            }
-            throw error;
-          }
+          return this.backgroundApi.serviceHardware.uploadPro2Nft({
+            connectId: device.connectId ?? '',
+            ...pro2UploadParams,
+          });
```
sidmorizon, line 73: **Adding an NFT that is already on the device now fails with an error instead of succeeding**

Severity: `non-severe`

The special handling that treated an "already on the device" response as a success was deleted (removal of the duplicate check in `packages/kit-bg/src/services/ServiceNFT.ts:39-63` and its use at the upload call), so re-adding an NFT that is already stored on the device now ends with an error instead of quietly succeeding.
Impact: Users who collect the same NFT to their device twice see a failure message even though the picture is already on the device.

<details>
<summary>How the duplicate-upload path changes</summary>

Before this PR, `uploadNFTImageToDevice` wrapped `serviceHardware.uploadPro2Nft` in a try/catch and mapped the device error whose message/payload contains `NFT already exists` to `{ nftUpdated: true, message: 'NFT already exists' }`. The new code calls `uploadPro2Nft` directly (`packages/kit-bg/src/services/ServiceNFT.ts:70-73`), so the raw hardware error propagates through `withHardwareProcessing` to the UI in `packages/kit/src/views/AssetDetails/pages/NFTDetails.tsx`, which shows an error toast.

The PR is described as "use SDK dimensions for Pro2 images" and only mentions removing size fallbacks; this idempotency removal is an unrelated behavioral change (test `ServiceNFT.pro2.test.ts` was rewritten to assert the rejection), and the repository rule in CLAUDE.md states existing code should not be removed unless the request explicitly requires it.

</details>

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: 可以忽略。sdk会同步变更。 并且后续应该再优化一个版本，直接在app中获取

packages/kit-bg/src/services/ServiceNFT.ts
wabicai: 符合预期。这里是对齐pro设备。

packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts
wabicai: 可以忽略，这里后续改为使用app中记录

originalix:

@wabicai

Reviewed commit `45a5d198d046`. This report always reflects the latest reviewed changes.

### Review summary

This PR upgrades the hardware SDK family to `1.2.0-alpha.128`, removes app-side Protocol V2 image-size fallbacks, and shifts more behavior to SDK-driven state in both `kit-bg` and the extension offscreen runtime. Alongside that ownership change, it also alters firmware-update UI state handling, device-label write-back, and Pro2 NFT duplicate semantics, so the hotfix now spans extension bridge recovery, firmware messaging, and device media flows in addition to the original size-configuration path. The result is a broader runtime surface where a few adjacent control paths now diverge from their previous contracts.

**What needs attention:** Re-check the Protocol V2 runtime paths that changed alongside the SDK bump: offscreen reconnect behavior, firmware update messaging, and duplicate NFT collection retries.

### Issues to address

- **P1 · Offscreen bridge reconnect leaves the port watchdog on a stale instance**
  - **Impact:** The new reconnect logic creates a replacement bridge only in `appGlobals`, but the offscreen runtime watchdog still polls the original `offscreenBridge` object. After the next background-port disconnect, the watchdog can still decide the port is dead and reload the offscreen page, dropping in-flight hardware events and negating the intended reconnect stabilization.
  - **Suggested action:** Keep a single mutable bridge reference for both reconnect and liveness checks, or switch the watchdog to read `appGlobals.extJsBridgeOffscreenToBg` instead of the captured startup bridge.
- **P2 · Protocol V2 update banner can announce the wrong version**
  - **Impact:** The new Protocol V2 banner branch runs before the BLE fallback and substitutes the current firmware version when `detectInfo.toVersion` is missing. On BLE-only upgrade detections, or on partial detect payloads without a firmware target version, users can be told that their current `SafeOS` version is the available update.
  - **Suggested action:** Only render the SafeOS version-specific copy when `detectInfo.toVersion` exists; otherwise fall back to the BLE-specific or generic firmware-update messaging.
- **P2 · Duplicate Pro2 NFT uploads now regress from idempotent success to user-visible failure**
  - **Impact:** The service no longer converts the SDK's duplicate-file error into a successful no-op, but the NFT collect UI still behaves like retries are safe and now shows the raw SDK error toast on an already-collected NFT. That turns a harmless retry into a visible failure without any device-state change.
  - **Suggested action:** Restore duplicate-file normalization at the service boundary, or update all callers to treat that specific SDK error code as an already-collected success path.

### Validation gaps

- I reviewed the changed files and directly related call sites only, per scope guidance.
- I did not run targeted Jest or app-level verification in this workspace.
- Existing PR comment/thread state was not incorporated into duplicate-comment checking, so inline comments remain draft suggestions only.

originalix (approved):

PR #12898, opened 2026-08-16
feat: add market list filters and client-side column sorting

erikzou:
## Summary

Implements PRD client-side column sorting and filter chips + Filters panel for the Market list, and unifies the row/header chrome across the four Market tabs so they stop drifting apart. Targets **6.6.0**.

Ships on by default — there is no feature flag.

## Filters

A chip bar plus a Filters panel over the trending list. `marketListFilterConfig.ts` is the single source of truth for every dimension, so the chips, the panel, the local filter and the request params cannot drift apart.

The split between server and client filtering is deliberate:

| Dimension | Where it runs |
|---|---|
| Market cap, liquidity, volume, txns, holders | Server (`filterParams` passthrough) |
| **Token age** | **Client only** — the API has no age field |

`marketListFilterPassthrough.test.ts` locks five invariants, including *every* server-side dimension emitting a param, so a silently dropped filter fails CI rather than quietly returning unfiltered rows.

Trending is the only list wired to filters, and `isMarketTrendingList` is the one predicate that decides it. Anything that *shows* a filter control gates on the same predicate that *consumes* it — the mobile toolbar previously used a lookalike condition, which would have put a live-looking Filters button on any future non-stock category that ignores it.

## Sorting

Header sorting is client-side across the spot, stock and perps tables. `useClientSortResult` intercepts the `sortBy`/`sortType` setters so a header press re-orders the rows already in hand instead of re-keying the SWR query and refetching. Favorites keeps its drag ordering and does not sort.

Measured 2026-08-16: the backend ignores `page`/`limit` and answers the first request with the whole pool (trending `list.length` 101 of `total` 101, stocks 106 of 106), so `canLoadMore` never opens and local sorting covers the complete set. This is observed behaviour rather than a documented contract — the client stays defensive (`pageSize: 20`, `maxPages: 5`), and if the backend starts honouring `limit`, sorting would cover one page only.

Token Age sorts through `getTokenAgeSortValue` rather than the raw timestamp: the cell renders an age (`now - firstTradeTime`), which runs opposite to the timestamp, so sorting the timestamp inverted the arrow.

## Shared list chrome

`marketListRedesignVisuals.tsx` holds the row height, star column, icon gap and header/sort rendering that the spot and perps tables both use. This closes a set of alignment bugs — perps rows were 60px vs 68px, mobile perps padding 16px vs 20px — that caused a visible jump when switching tabs.

The merged Name column gates the age on one `showTokenAge` flag shared by the header label and the cell, so a list that opts out (banner detail's `hideTokenAge`, the watchlist) cannot end up with a header reading "Name" above rows printing an age.

## Volume column naming

`Turnover` → `Volume`, and each header states its own window. Verified against the wire: `stock.assetAnalysis.volume24h` matched all four displayed values exactly, and the server has no turnover field at all, only `volume{window}`. Trending stays reactive to the selected time frame; Favorites, Banner detail and the detail-page token selector are pinned to `24h` because their time frame is not selectable. Header tooltips were removed from those three for the same reason — they promised a range the user cannot pick.

## Verification

- `yarn agent:check --profile commit` — all PASS
- `yarn jest packages/kit/src/views/Market` — 428 tests / 51 suites pass, 6 test files added or extended
- Filter passthrough proved end to end on web: filtered min market cap 512,085 vs 9,610 unfiltered
- Native bridge proved safe by reading `rpcProtocol.ts`: serialisation is plain `JSON.stringify` and deserialisation validates only `method` / `params`-is-array / `sync`, so no field is stripped
- Reactive trending header measured for cost: 47s of polling produced 102 page DOM mutations and **0** header mutations
- All four tabs re-checked in the browser at desktop and narrow widths after every behavioural change

## Notes for review

- **Not verified on a device.** Local CocoaPods is four major versions behind (`Manifest.lock` 14.2.0 vs `Podfile.lock` 18.21.0) and the build fails in `RNPurchases.m`. Everything was verified on web including narrow widths, but the `MobileLayout.native.tsx` path has not been seen on real hardware.
- **i18n blast radius.** `yarn i18n:pull` syncs the whole current Lokalise state, so the locale diff is larger than this feature: of the 70 new `en_US` keys only 14 are `market.*`. This matches how locale changes land in other PRs here (#12868, a one-line title fix, carried 2326 lines across 20 locale files).
- **Left alone deliberately**: `dexmarket_turnover` is still used in `views/Home/pages/PerpsContainer.tsx:682` and `views/UniversalSearch/components/MarketTableHeader.tsx:45`. Their data windows were not verified, so the label was not changed.
- The desktop `5m`/`1h`/`4h`/`24h` range labels remain untranslated literals in `TimeRangeSelector`, by request.
- **Known and not fixed here**: with the WebSocket feed on, `orderedData` is sorted before the live price overlay is applied and does not re-sort on a tick, so a price-sorted list shows fresh values in the previous poll's order until the next refresh. Handed to the team rather than patched, because `orderedData` also feeds the WS subscription range and reordering the two touches that calculation.


originalix:

@erikzou

Reviewed commit `f022cc2b291c`. This report always reflects the latest reviewed changes.

### Review summary

Before: Market lists relied mainly on server ordering, locally owned header state, and a compact set of category/time controls. After: the foreground now owns a shared filter-and-sort contract across chips, a modal, controlled Table headers, spot/perps/banner list comparators, and responsive list visuals; supported numeric filters are serialized through the existing Market request and cache boundary, while token age remains a local pass over fetched rows.

On iOS, Android, and extension, filter/sort UI state lives in main and request parameters are copied across the isolated main/bg JS heaps into the independently initialized bg service; no shared native resource or persistence ownership changes. Desktop and web remain single-runtime, with web retaining its direct light-API path. Watchlist drag ordering, authentication, signing, transaction verification, cryptography, and secret handling remain unchanged.

**What needs attention:** Confirm that final rendered values own ordering and pagination state, and that responsive banner-detail branches consume one sort contract.

### Issues to address

- **P1 · Live updates bypass the active client-side sort**
  - **Impact:** `MarketTokenListBase` sorts `rawData` before websocket and explicit live overrides are merged. When price, change, or another live field is the active comparator, displayed values can change while row order remains based on the older snapshot, contradicting the active sort indicator.
  - **Suggested action:** Merge live overrides into the dataset first, then apply the active client comparator and derive subscriptions from that final ordering.
- **P2 · Filter changes do not reset pagination exhaustion**
  - **Impact:** An empty load-more response sets `hasReachedEnd`, but the pagination-reset effect omits `filterParamsKey`, and the first-page result handler does not clear the flag. A filtered query started after an exhausted query can therefore remain unable to load additional pages.
  - **Suggested action:** Include `filterParamsKey` in the pagination-reset effect so every server-filter query clears `hasReachedEnd` and restarts from page one.
- **P2 · Banner-detail layouts use incompatible sort contracts**
  - **Impact:** Wide perps banners create private multi-column state with `usePerpsClientSort`, while compact banners consume the banner-level `change24h` state. Spot banners likewise expose broader sortable fields on wide layouts than compact layouts, so resizing or rotating can silently replace the visible ordering.
  - **Suggested action:** Drive both responsive branches from the banner-detail sort owner and restrict wide sorting to fields supported by the compact layout until both expose the same contract.

### Validation gaps

- Focused Jest tests could not start because the workspace lacks Yarn's `node_modules/.yarn-state.yml` installation state.
- Websocket-driven resorting, exhausted-query filter transitions, and responsive banner transitions were not verified in a running app.
- Existing GitHub reviews and comments could not be queried because the GitHub API was unreachable, so inline drafts require a duplicate-comment check before publication.

sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (commented):
@erikzou

Reviewed commit `7641a1e59ac7`.

- **P1 · Compact mobile filter bar exposes filters on tabs that do not consume them**
- **P1 · Spot tables now sort paginated categories on the loaded slice only**
- **P2 · Filter modal reset leaves the drafted time range behind**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenColumns/useColumnsDesktop.tsx
```diff
@@ -208,19 +439,38 @@ export const useColumnsDesktop = (
...
 
+          if (!record.perpsCoin) {
+            return renderRedesignTokenIdentity(
+              record,
+              intl,
+              gtXl,
+              copyFrom || ECopyFrom.Homepage,
+              showStockSubtitle ?? true,
+            );
+          }
```
sidmorizon, line 480: **Token creation age reappears in lists that asked for it to be hidden**

Severity: `non-severe`

The token age is always drawn in the redesigned name cell (`renderRedesignTokenIdentity` at `packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useColumnsDesktop.tsx:464-472`) even when the screen asked for it to be hidden, so lists such as the banner detail table now show an age they intentionally suppressed.
Impact: Screens that deliberately omit token age (banner detail) display it again, and the favorites table shows an age under a column still labelled just "Name".

<details>
<summary>How the hideTokenAge / header-label contract is broken</summary>

`useColumnsDesktop` still receives `hideTokenAge` and uses it to drop the standalone `tokenAge` column (`packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenColumns/useColumnsDesktop.tsx:742`), and it also uses `hideTokenAge` / `isWatchlistMode` / `useStockMetadataColumns` to decide whether the name header reads "Name/Token Age" or plain "Name" (same file, lines 445-452).

But the cell renderer no longer consults any of those flags: for every non-perps row it calls `renderRedesignTokenIdentity(...)`, which unconditionally computes `ageLabel = formatTokenAgeLabel(intl, record.firstTradeTime)` and renders it as the first element of the subtitle row. `MarketBannerDetail.tsx:220` passes `hideTokenAge`, and the watchlist passes `isWatchlistMode`; in both cases the header says "Name" while the rows now render the age, and the banner detail's explicit suppression is defeated whenever the API supplies `firstTradeTime`.

Threading the existing `hideTokenAge`/`isWatchlistMode` decision into `renderRedesignTokenIdentity` (e.g. a `showTokenAge` argument gated by the same condition that picks the header label) would keep the cell and the header consistent.

</details>

erikzou:
Thanks — went through every item. Four are fixed and pushed in `7259124c63`:

- **Token Age sorted the wrong way** (the real one). The cell renders an age (`now - firstTradeTime`), which runs opposite to the timestamp it was sorting on, so the arrow pointed one way and the rows went the other. Now sorted through `getTokenAgeSortValue`, with tests locking both directions and the missing-value sink.
- **`hideTokenAge` ignored in the merged Name cell.** Header and cell now read one `showTokenAge` flag, so they cannot disagree again.
- **Reset left the drafted time range behind.**
- **`sortChipClick` was declared but never emitted.**

The remaining three need a decision rather than a patch.

---

### `translations.ts` / locale JSON

These were not hand-edited. `packages/shared/src/locale/**` is the output of `yarn i18n:pull` after the keys were created in Lokalise, which is the source of record — nothing was typed into the generated files directly.

The part of the concern that does hold is blast radius. `i18n:pull` syncs the whole current Lokalise state, so of the 59 new `en_US` keys only 14 are `market.*`; the other 45 (`batch_psbt_*`, `perps_*`, `label_*`) belong to other teams and rode along. If the preferred flow is for a scheduled pull to carry those instead, I can reset this PR's locale files to `x` and keep only the source-side changes — happy to do whichever you prefer.

### Client-side sorting on paginated categories

The suggested action can't be applied as written, because the path it points at doesn't exist. On `x`, `MarketTokenListBase.handleHeaderRow` opens with:

```js
if (!isWatchlistMode && !clientSort) {
  return undefined;
}
```

and `MarketNormalTokenList` never passes `clientSort`. Spot category tables therefore had **no header sorting at all** before this PR — `clientSort` was used only by banner detail, for `change24h`. "Keeping them on the existing server-driven sort path" would remove the feature, not restore prior behaviour.

The substance of the finding is right, though, and a code comment of mine overstated the opposite (corrected in the same commit). Precisely:

- `trending` can lock pagination after page 1, but only when the first response already returned the whole pool (`hasReachedEndAfterFirstPage`), which is additionally gated on `HIDE_INJECTED_BTC_ROW`.
- Stock and any other spot category never lock, so their header sort covers the loaded rows only — exactly as described.

The trade-off behind the choice: the backend accepts three sort keys (`liquidity`, `mc`, `v24hUSD`), while the table exposes nine sortable columns, and stock rows sort on values read off `record.stock` — P/E has no backend ordering at all. Moving to server-side sorting would drop six columns outright and remove stock sorting entirely.

Options, and I'll take direction:

1. Keep as is, and document that sorting applies to loaded rows.
2. Stop auto-loading further pages while a client sort is active, so the set the user sorted stays the set they see.
3. Restrict client sort to `trending` and drop header sorting from stock and other spot categories.

### Compact filter trigger on tabs that don't consume filters

Condition mismatch confirmed:

- shown when `currentSpotCategoryId && !currentSpotCategoryHasStockData` (`MobileLayout.tsx:111`)
- consumed when `selectedCategory === 'trending' && !stockCategory` (`MobileMarketTokenFlatList.tsx:64`)

One correction on reproducibility: today's basic-config returns only `trending` plus stock categories as spot categories, so the two conditions currently coincide and the trigger renders on trending alone. I could not reproduce a tab where the badge changes and the rows don't — the "AI / Meme" categories in the report don't exist in the current config.

It is latent rather than live, and worth closing anyway since `spotCategories` is server-driven and a new category would surface it without a client release. I'd rather pass the category down and gate the trigger explicitly than rely on the two conditions continuing to coincide — let me know if you want that in this PR or a follow-up.

originalix (commented):
@erikzou

Reviewed commit `7259124c63c2`.

- **P1 · Client-side sort is enabled for spot categories that still paginate from the backend**
- **P2 · Live row updates do not re-run the active client-side sort**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

erikzou:
Correction to my previous comment — I conceded the pagination point without measuring it, and the measurement does not support it.

**The backend ignores `page`/`limit` and returns the whole pool on the first request.** Captured off the running app (XHR, desktop web):

| request | returned |
|---|---|
| `type=trending&page=1&limit=20` | `list.length: 101`, `total: 101` |
| `type=stocks&page=1&limit=20` | `list.length: 106`, `total: 106` |

Both lists therefore hold the complete set after one request, and `canLoadMore` is already `false` in both cases — for trending via `hasReachedEndAfterFirstPage`, and for stocks via `transformedData.length < totalCount` being false. No second page is ever fetched, so there is no "loaded slice" for the sort to diverge from, and nothing re-orders as the user scrolls.

So the specific failure described — *"the visible order diverges from the real backend pool and shifts again as more pages load"* — does not reproduce on either category today. My earlier "the substance is right" was wrong, and the code comment I changed to hedge it was accurate as originally written; I'll restore it.

Two things that remain true and are worth keeping on the record:

- The client code is defensive about pagination (`pageSize: 20`, `maxPages: 5`) because the response shape is backend behaviour, not a contract. If the backend starts honouring `limit`, client sort would cover 20 rows and the finding becomes real. Worth a backend-side confirmation that returning the full pool is intentional.
- Header sorting exists on desktop only. `MobileMarketTokenFlatList` has no sort path at all, and the three `MarketListColumnHeader` usages in `MobileLayout` pass no sort handler — so the mobile side of the report doesn't apply either.

Given that, I'd propose option 1 (keep as is). Happy to add the guard from option 2 if the backend intends to start paginating.

originalix (commented):
@erikzou

Reviewed commit `3f79d1dbe59c`.

- **P1 · Live market updates do not re-run the active client-side sort**
- **P2 · Mobile exposes filters on spot categories that do not consume them**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers.ts
```diff
@@ -4,6 +4,46 @@ import type { IMarketTokenListItem } from '@onekeyhq/shared/types/marketV2';
 import type { IMarketTimeRangeValue } from '../../../types';
 import type { IMarketToken } from '../MarketTokenData';
 
+// Temporarily hide the backend-injected native BTC row in the trending list.
+// Plan of record (PRD P0-1/P1-1): BTC moves to the curated Majors tab once it
+// ships. Flip to false at dev handoff to show the injected row again.
+export const HIDE_INJECTED_BTC_ROW = true;
```
originalix: P2: [将 Trending 的 BTC 可见性下沉到服务端]

这里用编译期常量过滤后端注入的 BTC，并为了修正过滤后与 `total` 的差异额外维护 `hasReachedEndAfterFirstPage`。这会把榜单内容策略、分页补丁和客户端发版绑定在一起：切换 BTC 是否展示需要重新发版，服务端分页或 `total` 语义变化时也容易出现重复拉取或提前终止。建议由 Market API 根据榜单配置决定是否返回 BTC，并保证 `list`/`total` 一致，前端只消费服务端结果，删除这组硬编码过滤和分页特判。

packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/MarketFiltersModal.tsx
```diff
@@ -0,0 +1,486 @@
...
+          variant="primary"
+          onPress={() => {
+            defaultLogger.dex.list.dexFilterChip({
+              action: 'popoverConfirm',
+              conditionCount: Object.keys(draft).length,
+            });
+            if (draftTimeRange !== initialTimeRange) {
+              onApplyTimeRange(draftTimeRange);
+            }
+            onApply(draft);
```
originalix, line 344: P2: [Confirm 无改动时会清空当前排序]

先点击 Top Volume 建立 `sortState`，再打开 Filters 并在不修改条件的情况下直接 Confirm，这里仍会调用 `onApply(draft)`；而 `applyConditions` 在未传 `options.sort` 时会把排序重置为 `EMPTY_SORT`。结果是一次无操作确认也会让排序 chip 消失、列表恢复默认顺序。建议先比较 `draft` 与 `initialConditions`，条件未变化时不要调用 `onApply`；仅修改 time frame 时也应保留现有排序。

packages/kit/src/views/Market/MarketHomeV2/components/marketListRedesignVisuals.tsx
```diff
@@ -0,0 +1,92 @@
...
+    return titleRow;
+  }
+
+  // Tooltip wraps the whole row: hovering anywhere on it explains the metric,
+  // and pressing anywhere on it sorts (the Tooltip trigger owns the press, so
+  // the handler must be forwarded here rather than left to the Column).
+  return (
+    <LazyTooltip
+      placement="top"
+      onPress={onSortPress}
```
originalix: P2: [Tooltip 表头会重复触发排序回调]

`HeaderColumn` 已经把同一个 `handleColumnPress` 绑定在外层 `Column`，这里又把它交给内层 `LazyTooltip.Trigger`。Web 点击 tooltip 表头时事件会继续冒泡到外层，因此同一次点击会执行两次 `onSortTypeChange` 及 `dexSort` 埋点。请只保留一个 press owner，例如不要在这里再次转发 `onSortPress`，让外层 `Column` 统一处理。

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers.ts
erikzou: Agreed, and it turns out the whole thing was unnecessary — product confirmed BTC does not need hiding at all. Removed in `adc8101f01` rather than moved server-side.

Gone: `HIDE_INJECTED_BTC_ROW`, `filterInjectedBtcRow`, `shouldHideInjectedBtcRowForType`, `hasReachedEndAfterFirstPage`, the two filter call sites in `useMarketTokenList`, and the now-dead `type` parameter on `transformMarketTokenListResponse`. Net -151 lines. None of it existed on `x`; this restores that behaviour.

Your point about the coupling was the right diagnosis: `hasReachedEndAfterFirstPage` only existed to patch the `list`/`total` mismatch the filtering created, so deleting the filter removed the pagination special case with it. Nothing is left binding list content policy to a client release.

Verified in the browser — BTC is back as the first trending row ($1.26T market cap, `--` for the fields the injected row does not carry), and 425 Market tests pass.

originalix (commented):
@erikzou

Reviewed commit `0068eaeaead6`.

- **P1 · Live token updates bypass the active client-side sort**
- **P2 · Banner detail now persists sort keys that some layouts cannot honor**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

originalix (commented):
@erikzou

Reviewed commit `adc8101f0182`.

- **P1 · Deselecting a filter tier leaves a ghost active condition behind**
- **P1 · Live token updates bypass the active client-side sort**
- **P2 · Banner detail now persists sort keys that the compact layout cannot honor**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

originalix (commented):
@erikzou

Reviewed commit `49df56ee1f93`.

- **P1 · Live token updates bypass the active client-side sort**
- **P2 · Re-selecting the current filter tier silently clears the sort**
- **P2 · Quick-chip matching ignores the chip's time-range anchor**
- **P2 · Banner detail stores sort keys that the compact layout cannot apply**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/MarketFiltersModal.tsx
erikzou: Confirmed and fixed in `49df56ee1f`.

Reproduced first: apply the Top Volume chip, open Filters, press Confirm without touching anything — the `成交量 由高到低` chip disappears and the list drops back to default order.

Took your suggestion: Confirm now compares `draft` against `initialConditions` and only calls `onApply` when they differ. Reused the `sameConditions` helper already in `marketListFilterConfig` (it was private, backing `findActiveMarketFilterChip`) rather than adding a second notion of "changed", and exported it with tests covering add / remove / retarget / key-order.

The time-frame case you raised is covered by the same shape: `onApplyTimeRange` is independent of `onApply`, so changing only the time frame no longer touches the sort.

Verified after the fix — unchanged Confirm keeps the sort chip and the row order is byte-identical before and after.

packages/kit/src/views/Market/MarketHomeV2/components/marketListRedesignVisuals.tsx
erikzou: Correct, and fixed in `49df56ee1f` the way you suggested — the forward is gone and the outer `Column` is the only press owner.

Worth recording why it bubbles, since it is not obvious from the call site: `LazyTooltip` does not wrap the trigger, it `cloneElement`s it and attaches `onPress` onto that same element (`LazyTooltip.tsx:118-127`), and once loaded the real `Tooltip` renders `TMTooltip.Trigger` inside the `Column` too (`Tooltip/index.tsx:195-203`). Neither stops propagation, so both handlers ran in both states. My original comment there claimed the Tooltip trigger owned the press — that was simply wrong, and the comment now records the real reason.

One correction to the impact, in case it matters for how you triage similar reports: the ordering itself was never wrong. Both invocations run off the same render’s closure, so `getNextSortOrder` reads the same `currentSortOrder` and returns the same value twice — the sort lands where the arrow says. What actually broke is exactly what you named: `dexSort` double-counts.

Verified after removing the forward that tooltip’d headers still sort on a single click (市值 desc → $1.26T … $287.51M) and the hover explainer still appears, so nothing depended on the forwarded handler.

originalix (commented):
@erikzou

Reviewed commit `b5e2b14cccc2`.

- **P1 · Live price updates bypass the active client-side sort**
- **P2 · Re-selecting the active filter tier clears the sort**
- **P2 · Preset matching ignores the chip's anchored time range**
- **P2 · Banner detail exposes sorts that the compact list cannot replay**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

originalix (commented):
@erikzou

Reviewed commit `bc30da45a7e7`.

- **P1 · Live updates bypass the active client-side sort**
- **P2 · Re-selecting the active filter tier clears the sort**
- **P2 · Preset matching ignores the chip time range**
- **P2 · Banner-detail layouts do not share one sort contract**
- **P3 · Deferred feature branches ship as dead production code**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

erikzou:
Latest round — two fixed in `91c3679e06`, one is a repeat, one is a deliberate no-change with the reachability worked out.

### Re-selecting the active filter tier clears the sort — fixed

Correct, and the guard belongs lower than the tier popover. `applyConditions` has four entry points (quick chip, tier popover, clear-all, Filters modal) and several of them re-apply a value that is already selected, so guarding one call site leaves the same bug behind the other three — which is exactly what happened: I had guarded the modal's Confirm in `49df56ee1f` and the popover still cleared the sort.

The no-op check now lives in `applyConditions` itself: resetting the sort belongs to an actual change of slice, so an unchanged apply returns early and leaves both states alone. A caller supplying its own sort (a chip that carries one) still wins unconditionally. The modal-level special case is gone, since it is redundant once the provider is right.

Verified: with `市值 $500K+` applied and a market-cap header sort active, re-selecting `500K+` in the popover keeps the sort chip and the row order ($1.28T → $775.26M, unchanged).

### Preset matching ignores the chip's anchored time range — fixed

Confirmed: `findActiveMarketFilterChip` compared conditions and sort only, so the sort-bearing chip stayed lit after the toolbar window moved off its `1h` anchor, and the preset-derived behaviour (icon, descending-only lock) kept applying. It now takes the window into account, with tests covering match-at-anchor, no-match-off-anchor, and the caller that passes no window.

### Live price updates bypass the active client-side sort — already tracked

Same finding Codex raised earlier in this PR; your report notes the cause under Validation gaps (`gh api` could not reach `api.github.com`, so comment dedupe did not run). Handed to the team rather than patched here: `orderedData` also feeds the websocket subscription range, so reordering sort-vs-overlay touches that calculation, and the product call — whether a price-sorted list should re-rank on every tick — is worth making before the code changes.

### Banner detail exposes sorts the compact list cannot replay — not changing, reachability below

The asymmetry is real: the sort lives in the persisted `marketBannerListSortAtom`, wide banner detail sorts through the shared client-sort surface, and `BannerDetailTokenFlatList` reorders by `change24h` only.

What makes it not worth restructuring is how it is reached. In `MarketBannerDetail.renderTokenList` the `!gtMd` branch is evaluated **before** the `platformEnv.isNative` branch, so phones always get the compact list and a non-`change24h` sort cannot be created there at all. Producing one requires the wide layout, and observing the disappear/reappear requires crossing the breakpoint twice — i.e. a web or desktop user resizing the window back and forth mid-task.

The same yardstick applies on the other side: the Market tabs avoid this only because `MarketListFilterProvider` is mounted inside each layout, so crossing the breakpoint destroys the sort *and the filter conditions* along with it. Neither behaviour is worth engineering around at that hit rate, and restricting banner detail to `change24h` would remove the wide-layout sorting that was just added to the perps section in `bc30da45a7` at product's request.

Happy to build the compact header out to the full sort set if you would rather close it properly — flagging the cost/benefit rather than declining outright.

originalix (commented):
@erikzou

Reviewed commit `91c3679e0681`.

- **P1 · Live updates bypass the active client-side sort**
- **P2 · Filter changes do not reset pagination exhaustion**
- **P2 · Banner-detail layouts use incompatible sort contracts**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/MarketFilterChipsBar.tsx
```diff
@@ -0,0 +1,735 @@
...
+              </XStack>
+              <XStack alignItems="center">
+                <SmallRoundIconButton
+                  icon="CrossedSmallOutline"
+                  onPress={() => {
+                    defaultLogger.dex.list.dexFilterChip({
+                      action: 'clearAll',
+                    });
+                    applyConditions({});
+                  }}
```
sidmorizon, line 714: **The clear button on the market filter bar does nothing when only a sort is active**

Severity: `non-severe`

Pressing clear-all only clears the filter conditions (`applyConditions({})` at `packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/MarketFilterChipsBar.tsx:713`) and leaves the active ordering untouched, so when nothing but an ordering is applied the button appears dead.
Impact: A user who sorted the trending table and then presses the clear button sees nothing change — the chip row stays expanded with the sort still applied.

<details><summary>Why the shared apply path silently no-ops</summary>

The chip row is rendered whenever `conditionEntries.length > 0 || hasSort` (`MarketFilterChipsBar.tsx:549`), so a header-driven sort alone expands the row and shows the clear-all `×`.

`applyConditions` in `packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/MarketListFilterContext.tsx:65-69` deliberately returns early when the incoming conditions equal the current ones, precisely to avoid dropping the sort on a no-op re-apply. With no conditions applied, `sameConditions({}, {})` is `true`, so the early return fires and neither `setFilterState` nor `setSortState` runs. Clear-all therefore only works when at least one filter condition exists.

The sort chip's own `×` calls `setSortState({})` directly (`MarketFilterChipsBar.tsx:677`) and does work, which makes the inconsistency more visible.

</details>

Suggested fix:

```ts
                  onPress={() => {
                    defaultLogger.dex.list.dexFilterChip({
                      action: 'clearAll',
                    });
                    applyConditions({});
                    setSortState({});
                  }}
```

originalix (commented):
@erikzou

Reviewed commit `eda5ecdb1e00`.

- **P1 · Live updates bypass the active client-side sort**
- **P2 · Filter changes do not reset pagination exhaustion**
- **P2 · Banner-detail layouts use incompatible sort contracts**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

sidmorizon:
@cursoragent review

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/MarketTokenListBase.tsx
```diff
@@ -551,23 +592,30 @@ function MarketTokenListBase({
 
   const handleHeaderRow = useCallback(
     (column: ITableColumn<IMarketToken>) => {
-      if (!isWatchlistMode && !clientSort) {
+      // The watchlist is ordered by drag, so a header sort would silently
+      // overwrite the arrangement the user built by hand. It never sorts.
+      if (isWatchlistMode || !clientSort) {
```
sidmorizon, line 597: # 🟠 P1 自选列表的表头排序被整体移除

## 问题

原来的守卫是 `if (!isWatchlistMode && !clientSort) return undefined;`，watchlist 会继续往下走并命中 `columnsMap = SORTABLE_COLUMNS`（`liquidity` / `mc` / `v24hUSD`）。本 PR 改成 `if (isWatchlistMode || !clientSort) return undefined;`，而 `MarketWatchlistTokenList` 从不传 `clientSort`，因此收藏页的每一列都不再返回 `onSortTypeChange`；`HeaderColumn` 里 `enableSortType = !!events?.onSortTypeChange` 恒为 false，排序箭头和点击响应一并消失。

注释里给出的理由（"watchlist 由拖拽排序，表头排序会覆盖手工顺序"）与既有实现不符：拖拽顺序只是 `sortBy` 为空时的默认序，`useMarketWatchlistTokenList` 在 `!sortBy || !sortType` 时按 `sortIndex` 排，第三次点击表头即回到该默认序，两者本来就是共存的。

## 影响

桌面「自选」标签上原本可点排序的 `24h Vol` 列（`turnover` → `v24hUSD` → `SORT_MAP` → `turnover`）现在变成纯文本，用户无法再按成交额排自选。连带 `useMarketWatchlistTokenList` 中 `sortBy/sortType` 的整个 `sortedData` 分支、`tokenListHelpers.SORT_MAP`，以及 `MarketTokenListBase` 里 `MarketWatchlistOnlyChanged` 为自选清空排序的逻辑全部变成死代码——没有任何入口能再把自选的 `sortBy` 置为非空。

## 建议

保留 watchlist 分支，只把 `clientSort` 分支新增到它旁边；如果确实要保护拖拽顺序，也应只在真正可拖拽时（`draggable`/`isDraggable` 为真）禁用，而不是整个 `isWatchlistMode`：

```ts
if (!isWatchlistMode && !clientSort) {
  return undefined;
}
if (!clientSort) {
  // 自选沿用原 SORTABLE_COLUMNS 分支
}
```

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -602,10 +637,9 @@ export function useMarketTokenList({
     // Update only rows whose visible fields changed so Table row memoization can
     // survive seed -> remote refresh and polling updates.
     setTransformedData((prev) =>
-      reuseStableMarketTokenRows({ prev, next: transformed }),
+      reuseStableMarketTokenRows({ prev, next: visibleTokens }),
```
sidmorizon, line 640: # 🟠 P1 筛选变化不复位分页，`hasReachedEnd` 会永久卡死列表

## 问题

本 PR 新增了 `filterParamsKey` 作为查询维度（已进入 `currentQueryKey` 和 `swrKeys.marketHomeTokenList`），但下面重置分页的 effect 依赖数组仍然是 `[networkId, sortBy, sortType, type, category, timeFrame]`，不含 `filterParamsKey`：

```ts
useEffect(() => {
  setCurrentPage(1);
  setIsLoadingMore(false);
  setHasReachedEnd(false);
}, [networkId, sortBy, sortType, type, category, timeFrame]);
```

同时本 PR 在紧接本行下方删除了 `setHasReachedEnd(false);`（原本每次新的 `apiResult` 到达都会兜底复位）。两处叠加后，`hasReachedEnd` 一旦置 true 就只能靠改网络/时间窗/分类来清除。

## 影响

`hasReachedEnd` 只在 `loadMore` 返回空列表时置 true，而分页只在选中了具体网络时开启（`hasNetworkId = Boolean(networkId)`），所以场景是：在某条链的 trending 上触底一次之后，用户再改任意筛选条件（新的 `filterParamsKey` → 新 query → 新 `apiResult`），`canLoadMore` 里的 `!hasReachedEnd` 恒为 false，筛选后的列表永远停在第 1 页 `20` 条。60s 轮询本身也会把 `transformedData` 截回第 1 页（`reuseStableMarketTokenRows` 在长度不同时直接返回 `next`），过去靠这里的 `setHasReachedEnd(false)` 才能重新翻页，现在同样恢复不了。

## 建议

把 `filterParamsKey` 加入重置 effect 的依赖，并保留新数据到达时的兜底复位：

```ts
}, [networkId, sortBy, sortType, type, category, timeFrame, filterParamsKey]);
```

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenColumns/useColumnsDesktop.tsx
```diff
@@ -167,6 +403,11 @@ export const useColumnsDesktop = (
     const watchlistNameWidth = gtLg ? 340 : 260;
     const shouldRenderRichCell = (index?: number) =>
       !shouldUseLightweightCell(index, deferRichRowAfterIndex);
+    // One gate for the merged column: the header only names Token Age where
+    // the cell actually prints one. Stocks and the watchlist never carry an
+    // age, and banner detail opts out via hideTokenAge.
+    const showTokenAge =
+      !useStockMetadataColumns && !hideTokenAge && !isWatchlistMode;
```
sidmorizon, line 410: # 🟡 P2 未启用 redesign 列表的 tab 会重复渲染代币年龄

## 问题

`showTokenAge` 只排除了 stock / watchlist / `hideTokenAge`，没有和「Name 与 Token Age 合并」这件事的真正开关 `redesignColumnOrderEnabled` 挂钩。而独立 `tokenAge` 列的生成条件是 `gtXl && !isWatchlistMode && !hideTokenAge`，它只会被 `REDESIGN_COLUMN_ORDER`（不含 `tokenAge`）过滤掉——也就是只有 trending 会被过滤。

`x_mentioned` 在 `SPOT_CATEGORIES_WITH_FULL_STATS` 里，所以 `shouldHideSpotExtendedStats` 为 false、`hiddenDesktopColumns` 为 `undefined`，`tokenAge` 不会被隐藏；同时它不是 trending，`redesignColumnOrderEnabled` 为 false，`showTokenAge` 却仍然为 true。

## 影响

在 `gtXl` 宽度下的 `x_mentioned` spot 标签上，Name 表头显示为 `Name / Token Age` 且单元格副行渲染 `ageLabel`，右侧同时还保留着独立的 `Token Age` 列，同一个数值在一行里出现两次，表头语义也与实际列集合不符。后续服务端再放出任何非 stock 的 full-stats 分类都会复现。

## 建议

把合并列的显示条件绑定到同一个开关上：

```ts
const showTokenAge =
  Boolean(redesignColumnOrderEnabled) &&
  !useStockMetadataColumns &&
  !hideTokenAge &&
  !isWatchlistMode;
```

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

sidmorizon (commented):
Review

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketFilterChipsBar/applyMarketListLocalFilter.ts
```diff
@@ -0,0 +1,49 @@
...
+      if (!dimension?.localField || !option) {
+        return true;
+      }
+      const raw = token[dimension.localField] as number | undefined;
+      if (raw === undefined || raw === null || Number.isNaN(raw)) {
+        return false;
+      }
+      // Age dimensions store a ms epoch timestamp (same clock as Date.now());
+      // convert to an age before comparing against the option range.
+      const value = dimension.isAge ? nowMs - raw : raw;
```
sidmorizon, line 39: **Future timestamps bypass age filters**

Severity: `non-severe`

For a future `firstTradeTime`, `nowMs - raw` is negative and satisfies every maximum-age tier. The row remains visible although its age cell renders no value.

Suggested fix:

```ts
      const value = dimension.isAge ? nowMs - raw : raw;
      if (dimension.isAge && value <= 0) {
        return false;
      }
```

originalix (commented):
@erikzou

Reviewed commit `f022cc2b291c`.

- **P1 · Live updates bypass the active client-side sort**
- **P2 · Filter changes do not reset pagination exhaustion**
- **P2 · Banner-detail layouts use incompatible sort contracts**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12898#issuecomment-5306111623)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/MarketTokenListBase.tsx
```diff
@@ -328,29 +352,55 @@ function MarketTokenListBase({
...
       return rawData;
     }
 
-    return [...rawData].toSorted((a, b) => {
-      const aVal = (a[field] as number) ?? 0;
-      const bVal = (b[field] as number) ?? 0;
-      return currentSortType === 'asc' ? aVal - bVal : bVal - aVal;
-    });
-  }, [clientSort, currentSortBy, currentSortType, rawData]);
+    return sortMarketTokensClient(rawData, field, currentSortType);
```
sidmorizon, line 387: 🟡 P2: Live price sorting becomes stale

Severity: `non-severe`

With WebSocket updates enabled, `orderedData` sorts pre-update prices. Displayed prices change without reordering, so the selected sort becomes incorrect until polling refreshes.

sidmorizon:
Claude review session: https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt

sidmorizon:
Devin Review has been triggered for this PR.

[View Devin Review ↗](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/12898>)

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/12898)

sidmorizon:
Cursor review failed.

Remote error: `cursor_request_failed: HTTP 503 Service Unavailable: Service Unavailable`

PR #12901, opened 2026-08-17
fix: sync stock token metadata for max amount(OK-58721)

zhaono1:
[OK-58721](https://onekeyhq.atlassian.net/browse/OK-58721)

----------

## Summary

- Synchronize the selected stock token's execution metadata with authoritative token details before percentage-based amount actions.
- Use the authoritative balance and decimals for Sell Max, and keep balance actions unavailable until the stock balance metadata is ready.
- Preserve Max behavior after a cold start and after reselecting the stock token.

## Intent & Context

Fixes OK-58721

The first visit to the Stocks Sell page after an app restart could leave Max unable to populate the amount. The same failure could recur after opening the network/token selector and reselecting the stock token.

QA verification issue: [OK-58833](https://onekeyhq.atlassian.net/browse/OK-58833) (closed).

## Root Cause

The selected/execution stock token could retain list or cached metadata while the balance token detail carried the authoritative decimals and stock identity. Max then calculated and validated the amount against stale execution metadata, so the input update could be rejected. Token reselection could recreate the same metadata mismatch.

## Design Decisions

- Treat the identity-matched balance token detail as the authoritative source for stock decimals and `isStock` metadata.
- Synchronize only execution metadata; do not change token identity, quote routing, or regular Swap behavior.
- Hide or disable Max and percentage actions until authoritative balance metadata is ready, avoiding writes from stale loading state.

## Changes Detail

- Added pure helpers and coverage for stock execution metadata resolution and balance-action readiness.
- Synchronized authoritative stock token details into selected and execution token state.
- Updated stock trade inputs to calculate and validate amounts using authoritative balance and decimals.
- Gated Max and percentage controls while balance metadata is initializing.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Mobile, Desktop, Web, and Extension through the shared Swap UI main runtime
- **Runtime Scope**: Main/UI runtime only; no background or native resource ownership changes
- **Risk Areas**: Async token-detail timing, Sell balance formatting, Max/percentage actions, and stock token reselection
- **Regression Boundary**: Regular Swap quote, history, and non-stock token paths are unchanged

## Test plan

- [x] Stock-focused Jest coverage: 6 suites and 77 tests passed
- [x] iOS Simulator (iPhone 17): cold start -> Trade -> Stocks -> Sell; Max was visible, enabled, and updated the amount field
- [x] iOS Simulator: open selector, search and reselect AAPL, then press Max successfully
- [x] No app error/fatal/exception log lines were observed around the verified Max interactions
- [ ] QA account with a nonzero AAPL balance: verify the exact nonzero amount after cold start and after token reselection

## Delivery Notes

- Target branch: `release/v6.5.0`
- The source fix already exists on `x`; no follow-up sync to `x` is required.
- No bundle was built as part of this PR creation task.

[OK-58833]: https://onekeyhq.atlassian.net/browse/OK-58833?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-58721]: https://onekeyhq.atlassian.net/browse/OK-58721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
+++ b/packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
@@ -33,10 +33,12 @@ import { buildSwapRateDifference } from '../utils/swapRateDifferenceUtils';
 
 import {
   type IStockBalanceSnapshot,
+  isStockBalanceActionReady,
   isStockBalanceInitializing,
   isStockPayTokenReadyForTradeInput,
   resolveStockBalanceSeed,
   resolveStockBalanceSnapshot,
+  resolveStockExecutionTokenMetadata,
   shouldRenderStockTradeInputSkeleton,
 } from './swapStockChannelUtils';
 import {
@@ -254,11 +256,13 @@ function useStockInputTokenBalance({
 
   const balanceReady =
     detailState.scope === balanceScope && detailState.balance !== undefined;
+  const authoritativeBalance = balanceReady ? detailState.balance : undefined;
+  const authoritativeTokenDetail = balanceReady
+    ? detailState.tokenDetail
+    : undefined;
   const balanceState = resolveStockBalanceSnapshot({
-    authoritativeBalance: balanceReady ? detailState.balance : undefined,
-    authoritativeTokenDetail: balanceReady
-      ? detailState.tokenDetail
-      : undefined,
+    authoritativeBalance,
+    authoritativeTokenDetail,
     ownerScope: balanceOwnerScope,
     previousSnapshot: lastValidBalanceStateRef.current,
     seededBalance: enabled ? seededBalance : undefined,
@@ -271,6 +275,8 @@ function useStockInputTokenBalance({
   const tokenDetail = balanceState?.tokenDetail;
 
   return {
+    authoritativeBalance,
+    authoritativeTokenDetail,
     balance,
     tokenDetail,
     loading: isStockBalanceInitializing({
@@ -476,6 +482,7 @@ export function useSwapStockAmountInputState({
     disableNativePayToken,
     selectPayToken,
     stockTokenStatus,
+    syncStockTokenDetail,
     tradeSide,
   } = stockChannel;
   const isBuySide = tradeSide === ESwapStockTradeSide.Buy;
@@ -496,6 +503,18 @@ export function useSwapStockAmountInputState({
   const stockInputTokenBalance = useStockInputTokenBalance({
     enabled: inputTokenReady,
     token: inputToken,
+  });
+  const authoritativeStockInputToken = isBuySide
+    ? undefined
+    : resolveStockExecutionTokenMetadata({
+        token: inputToken,
+        tokenDetail: stockInputTokenBalance.authoritativeTokenDetail,
+      });
+  const amountInputToken = authoritativeStockInputToken ?? inputToken;
+  const balanceActionsReady = isStockBalanceActionReady({
+    authoritativeBalance: stockInputTokenBalance.authoritativeBalance,
+    authoritativeStockToken: authoritativeStockInputToken,
+    isBuySide,
   });
   const resolvedInputTokenBalance =
     stockInputTokenBalance.balance ?? inputToken?.balanceParsed ?? '0';
@@ -543,7 +562,7 @@ export function useSwapStockAmountInputState({
     settingsPersistAtom.currencyInfo.symbol;
   const onAmountChange = useCallback(
     (value: string) => {
-      if (validateAmountInput(value, inputToken?.decimals)) {
+      if (validateAmountInput(value, amountInputToken?.decimals)) {
         setSwapAlerts({
           quoteId: '',
           states: [],
@@ -554,17 +573,17 @@ export function useSwapStockAmountInputState({
         });
       }
     },
-    [inputToken?.decimals, setFromTokenAmount, setSwapAlerts],
+    [amountInputToken?.decimals, setFromTokenAmount, setSwapAlerts],
   );
   const setInputAmount = useCallback(
     (amount: BigNumber) => {
-      if (!inputToken || !amount.isFinite() || amount.isNaN()) {
+      if (!amountInputToken || !amount.isFinite() || amount.isNaN()) {
         return;
       }
       const amountValue = amount
-        .decimalPlaces(Number(inputToken.decimals ?? 6), BigNumber.ROUND_DOWN)
+        .decimalPlaces(amountInputToken.decimals, BigNumber.ROUND_DOWN)
         .toFixed();
-      if (!validateAmountInput(amountValue, inputToken.decimals)) {
+      if (!validateAmountInput(amountValue, amountInputToken.decimals)) {
         return;
       }
       setSwapAlerts({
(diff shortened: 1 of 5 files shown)


originalix:

@zhaono1

Reviewed commit `d46434beecfd`. This report always reflects the latest reviewed changes.

### Review summary

This PR tightens Stock amount-entry ownership across the main/background boundary: the main Stock channel now accepts identity-matched token detail returned by the bg Swap service as the authoritative source for decimals, native status, and Stock metadata, then resynchronizes selected and execution tokens before Max or percentage calculations. It gates balance-derived actions until an authoritative balance and, on the sell side, authoritative Stock metadata are present, while retaining the existing quote/build protocol and account-scoped balance snapshots. The changes are localized and security-neutral, but the non-iOS keyboard percentage path is not fully gated.

**What needs attention:** Confirm that every Max and percentage entry point is hidden or disabled until the main runtime receives identity-matched balance and metadata from bg, especially the non-iOS keyboard path.

### Issues to address

- **P2 · Percentage controls remain visible but inert while balance data is pending**
  - **Impact:** When the Stock channel becomes quote-ready before its authoritative balance request settles, StockActionGate passes an undefined percentage handler. SwapActionsState nevertheless renders PercentageStageOnKeyboard on non-iOS platforms, and each badge silently optional-chains the missing handler. Users therefore see active-looking percentage controls that do nothing; the state can persist when token-detail loading fails.
  - **Suggested action:** Make SwapActionsState omit the non-iOS percentage toolbar when no percentage handler is provided, and add an integration test for quote-ready Stock state with balanceActionsReady=false.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` completed without whitespace errors.
- The focused command `yarn jest packages/kit/src/views/Swap/hooks/swapStockChannelUtils.test.ts --runInBand` could not execute because this checkout has no Yarn node_modules state file.
- No live UI validation was completed. The relevant reproduction is to focus the Stock amount input on Android or another non-iOS surface after quote readiness but before the balance-detail response. Merely seeing the percentage toolbar is not a pass; it must be absent or explicitly disabled until ready, then calculate from the fresh balance and decimals after readiness.
- Existing pull-request discussion could not be checked because GitHub API access failed, so the proposed inline wording should be checked for duplication before posting.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -497,6 +504,18 @@ export function useSwapStockAmountInputState({
...
+    : resolveStockExecutionTokenMetadata({
+        token: inputToken,
+        tokenDetail: stockInputTokenBalance.authoritativeTokenDetail,
+      });
+  const amountInputToken = authoritativeStockInputToken ?? inputToken;
+  const balanceActionsReady = isStockBalanceActionReady({
+    authoritativeBalance: stockInputTokenBalance.authoritativeBalance,
+    authoritativeStockToken: authoritativeStockInputToken,
+    isBuySide,
+  });
```
sidmorizon, line 518: **Max and percentage buttons stop working for stock trading when the balance lookup fails**

Severity: `non-severe`

Balance shortcuts are switched off (`isStockBalanceActionReady` at `packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:514-518`) whenever the balance lookup for the traded token errors out or comes back without a matching entry, so the Max and percentage buttons stay unusable even though a balance is displayed.
Impact: After a transient network hiccup on the stock trade screen, users see their balance but can no longer tap Max or 25/50/75%, until they switch tokens or accounts.

<details>
<summary>How the readiness gate can never flip to true</summary>

`authoritativeBalance`/`authoritativeTokenDetail` are only non-undefined when `balanceReady` holds (`packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:257-262`), which requires the `usePromiseResult` fetch of `fetchSwapTokenDetails` to have resolved for the current `balanceScope`. That promise is not wrapped in try/catch (`packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:191-206`), and `usePromiseResult` keeps the previous/`initResult` value on rejection, so a failed first request leaves `detailState.scope === ''` forever — there is no polling or retry, only `refreshKey`, account or token changes re-run it.

On the sell side the gate is even stricter: it also needs `authoritativeStockInputToken`, which is `undefined` when `details?.[0]` is missing (`markStockUsdPriceCurrency(undefined)` returns `undefined`) or when the returned detail does not identity-match the selected token.

All UI affordances are gated on this flag (`packages/kit/src/views/Swap/pages/components/SwapStockDesktopContainer.tsx:1106`, `:1127`, `:1178`, `:1180`), whereas previously Max operated on the displayed balance and always worked. A safer fallback would be to keep the previous displayed-balance behaviour when no authoritative detail is available (e.g. after the request settles, even unsuccessfully), rather than disabling the controls indefinitely.

</details>

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -554,17 +573,17 @@ export function useSwapStockAmountInputState({
...
   );
   const setInputAmount = useCallback(
     (amount: BigNumber) => {
-      if (!inputToken || !amount.isFinite() || amount.isNaN()) {
+      if (!amountInputToken || !amount.isFinite() || amount.isNaN()) {
         return;
       }
       const amountValue = amount
-        .decimalPlaces(Number(inputToken.decimals ?? 6), BigNumber.ROUND_DOWN)
+        .decimalPlaces(amountInputToken.decimals, BigNumber.ROUND_DOWN)
```
sidmorizon, line 584: **Max amount can produce a nonsense number when the token is missing its decimal precision**

Severity: `non-severe`

The rounding step now trusts the token's decimal precision without a default (`decimalPlaces(amountInputToken.decimals, …)` at `packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:584`), so if that precision is absent the rounding call returns a digit count instead of an amount and the input gets filled with a completely wrong number.
Impact: In the rare case where a stock token arrives without decimal precision, tapping Max or a percentage fills the amount field with a bogus value the user could then submit.

<details>
<summary>bignumber.js behaviour when dp is omitted</summary>

`BigNumber.prototype.decimalPlaces(dp, rm)` only rounds when `dp != null`; otherwise it ignores `rm` and returns the *number of decimal places* as a plain JS number. `.toFixed()` is then `Number.prototype.toFixed()`, producing e.g. `"18"`, which passes `validateAmountInput` and is written into the amount atom.

The previous code guarded this with `Number(inputToken.decimals ?? 6)`; the new code dropped both the `Number()` coercion and the fallback. The value can genuinely be missing at runtime because `resolveStockExecutionTokenMetadata` copies `decimals` straight from the fetched token detail (`packages/kit/src/views/Swap/hooks/swapStockChannelUtils.ts:160-165`), and that object is what becomes `amountInputToken`.

</details>

Suggested fix:

```ts
        .decimalPlaces(
          Number(amountInputToken.decimals ?? 6),
          BigNumber.ROUND_DOWN,
        )
```

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -497,6 +504,18 @@ export function useSwapStockAmountInputState({
     enabled: inputTokenReady,
     token: inputToken,
   });
+  const authoritativeStockInputToken = isBuySide
+    ? undefined
+    : resolveStockExecutionTokenMetadata({
+        token: inputToken,
+        tokenDetail: stockInputTokenBalance.authoritativeTokenDetail,
+      });
```
sidmorizon, line 512: **Stock trade input rebuilds a token object on every screen update, re-firing its synchronisation effect**

Severity: `non-severe`

A brand-new token object is built during rendering (`resolveStockExecutionTokenMetadata` at `packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:507-512`) and then used as the trigger for the metadata synchronisation step, so that step re-runs on every single screen update while the metadata still differs.
Impact: Extra repeated work and callback churn on the stock trade screen; the repo's React checklist forbids this pattern.

<details>
<summary>Rule violated and propagation of the unstable identity</summary>

`.cursor/rules/self-testing.mdc` section 2 requires: "No new object/array creation in render paths feeding dependency arrays". `authoritativeStockInputToken` is computed in the render body (no `useMemo`), and when the cached token metadata does not yet match the fetched detail, `resolveStockExecutionTokenMetadata` returns a fresh object literal (`packages/kit/src/views/Swap/hooks/swapStockChannelUtils.ts:160-165`).

That unstable reference feeds `useEffect([authoritativeStockInputToken, syncStockTokenDetail])` (`packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:600-605`), `setInputAmount`'s dependency array via `amountInputToken` (`:598`), and `onBalanceMaxPress`/`onSelectPercentageStage` (`:614-620`, `:632-638`). It is also returned as `inputToken` to the UI, so `StockAmountInput` receives a new token prop identity on each render. Wrapping the computation in `useMemo` keyed on the input token and the authoritative detail would keep the identity stable.

</details>

originalix (approved):

originalix (approved):

PR #12882, opened 2026-08-14
chore: sync release/v6.5.0 to x

originalix:
Sync bundle release changes from `release/v6.5.0` to `x` (after release #7).

Baseline: `bc5d83e3b1` (#12792), the last commit synced by #12812.

## Synced (6)

| Commit | PR |
|---|---|
| `ee380a24fa` | fix: prevent iPad dapp touch freeze (#12806) |
| `ca178118af` | fix: keep the perps first screen height stable on cold start (#12793) — **partial, see follow-up** |
| `4c0efaff64` | fix: correct iPad settings scrolling (#12813) |
| `794447804a` | fix: center desktop browser banner (#12826) |
| `fdd51b0607` | fix: restore perps chart on Android foldables (#12821) |
| `8cd90afe6e` | fix: self-heal unreadable SimpleDB records and support clearing Perps data (#12827) |

## Deliberately skipped

| Commit | Reason |
|---|---|
| `3f3f50f6af` Feat/spark bitway (#12765) | Already skipped by #12812 — backport of #12463, which is on `x`. Unchanged decision. |
| `54197b5967` / `1a75a3823f` chore: release #6 / #7 | Release-tracking only (`RELEASES.json`), must not sync to `x`. |

## ⚠️ Needs module owner (not in this PR)

These four carry real fixes `x` lacks, but `x` refactored the same code independently — a mechanical merge would regress `x`.

| PR | Conflict | Why it needs an owner |
|---|---|---|
| #12823 gas account funnel analytics | 10 hunks / 4 files | `x`'s #12837 (*improve Swap quote and sponsored fee states*, 08-12) rewrote the gas-account sponsorship path in `useSwapBuiltTx.ts` and `marketDirectSendTx.ts`. #12823 instruments the pre-#12837 shape. Needs someone across both. |
| #12834 track universal search source | 2 files | `x`'s `DappHeader.tsx` no longer renders `UniversalSearchInput`; applying the hunk would re-add UI `x` removed. `universalSearchSource.ts` is genuinely absent from `x`, so the tracking needs re-wiring to `x`'s entry points. |
| #12791 DeFi revamp for mobile 1.0 | 41 files | Earn/Borrow/Staking evolved independently on `x`. Largest single change in the release (93 files). |
| #12848 harden Swap recipient and warning flows | 7 files | `states/jotai/contexts/swap/actions.ts` was rewritten on `x` by #12862 (*stabilize Swap flows*) on the same day. Swap recipient resolution — not safe to merge mechanically. |

[...]

PR #12880, opened 2026-08-14
chore: release #7

originalix:
## Summary

- Record release #7 in `RELEASES.json`
- Commit: `fd071a7c71`
- PRs included: #12791, #12793, #12806, #12813, #12821, #12823, #12826, #12827, #12834, #12848

## Release Notes

Revamp the mobile DeFi experience (#12791); keep the Perps first screen height stable on cold start (#12793); prevent iPad DApp touch freeze (#12806); correct iPad settings scrolling (#12813); restore the Perps chart on Android foldables (#12821); add gas account funnel analytics (#12823); center the desktop browser banner (#12826); self-heal unreadable SimpleDB records and support clearing Perps data (#12827); track universal search source (#12834); harden Swap recipient and warning flows (#12848)

## Pre-release checks

- **Diff check**: 11 commits, 176 files (+10,621 / -1,125) since seq #6
- **Security audit**: 0 Critical, 0 High. No dependency, CI/CD, or native/build config changes — pure JS bundle against existing App Shell build `2026071325`.

PR #12903, opened 2026-08-17
chore: sync published release/v6.5.0 bundle #7 into hotfix/v6.5.2

originalix:
Follow-up to #12839, applying the same selection rule to the newly published bundle.

## Why now

#12839 synced `release/v6.5.0` up to `3f3f50f6af` (bundle #6) and stated the rule it used:

> This covers every commit **already shipped OTA** to 6.5.0 users via bundles #5 and #6, so upgrading 6.5.0 → 6.5.2 does not regress them. Commits after bundle #6 are deliberately excluded: **they have not shipped in any bundle, so omitting them regresses nobody.**

**Bundle #7 has since published** (`fd071a7c71`, recorded as seq 7 in `RELEASES.json` via #12880, merged 2026-08-14). Its 10 PRs are now on 6.5.0 users' devices, so the exclusion premise no longer holds — omitting them now *would* regress anyone upgrading 6.5.0 → 6.5.2.

Verified before this PR: all 10 were **absent** from `hotfix/v6.5.2` in every form. A per-file blob sweep of all 177 file-touches found **POST=0** — not one file was at the post-fix state, and the few that matched neither side differed only because *another* bundle-#7 PR was also missing.

## Synced (bundle #7, all 10 PRs)

| Commit | PR |
|---|---|
| `52506505c4` | fix: prevent iPad dapp touch freeze (#12806) |
| `75fa8a452b` | fix: keep the perps first screen height stable on cold start (#12793) |
| `b13dfd80a7` | fix: correct iPad settings scrolling (#12813) |
| `e767b76d69` | feat: add gas account funnel analytics (#12823) |
| `deed741a12` | fix: center desktop browser banner (#12826) |
| `fae5dc2b19` | feat: track universal search source (#12834) |
| `8b26f7b898` | fix: self-heal unreadable SimpleDB records and support clearing Perps data (#12827) |
| `960b9656b2` | Defi revamp for mobile 1.0 (#12791) |
| `71c4780885` | fix: restore perps chart on Android foldables (#12821) |
| `1f707f7aa4` | fix: harden Swap recipient and warning flows (#12848) |

Sequential `cherry-pick -x` in release order — **not** a branch merge. A merge is unusable here: `git merge-tree origin/hotfix/v6.5.2 fd071a7c71` reports **99 conflicted paths**, including files the hardware work touches (`ServiceStaking.ts`, `PerpTradingPanel.tsx`, the Earn `Universal*` pages). Those are phantom conflicts — #12839 was squash-merged, which destroyed the shared ancestry, so git re-diffs bundle #5/#6 content already present on both sides. Cherry-pick sidesteps it entirely.

## Deliberately excluded

| Commit | Reason |
|---|---|
| `54197b5967` / `1a75a3823f` chore: release #6 / #7 | `RELEASES.json` only. #12839 dropped that ledger from this line (it is keyed to `BUILD_NUMBER=2026071325`, which `chore: v6.5.2` removed). |
| `12a7b65696` fix: inject jpush key into native bundle builds (#12872) | Not part of bundle #7 — it landed *after* the seq-7 commit. `JPUSH_KEY` appears only in `release-native-bundle.yml`, which is invoked solely by the OTA bundle pipeline; the App Shell store workflows never reference it. Irrelevant to a 6.5.2 store build. **Note: it is also missing from `x`**, so the next release line cut from `x` inherits that gap. |

`.env.version` is untouched — this branch stays `VERSION=6.5.2` with no `BUILD_NUMBER`.

## Conflicts resolved

**10 locale JSON files** (`id, ko_KR, pt, pt_BR, th_TH, uk_UA, vi, zh_CN, zh_HK, zh_TW`) — resolved `--ours`, matching #12839. Verified lossless key-by-key: the hotfix side is a strict superset (5253 vs 4824 keys, 429 ours-only, **0 theirs-only**), and a sweep across *all* locale files confirms **0 keys** that #12791 needs are missing from this branch.

One genuine value difference: `earn_loans__action` was added independently by both sides — hotfix `借贷/借貸`, #12791 `借币/借幣` (en_US: `Loans`). Kept the hotfix wording, which comes from the newer translation generation. Per CLAUDE.md these files are generated and were not hand-edited.

**`.skillshare/skills/1k-retrospective/references/case-studies.md`** — both sides appended a different case study at the same position; git's auto-resolution dropped the hotfix one. Union-merged so both survive (BLE pairing dialog + Perps IndexedDB corruption).

**Three files git 3-way-merged silently** (blob differs from the source commit because this branch had its own edits). Each verified to carry the release-side content:
- `ServiceSetting.ts` — `values.perpsData` branch present (line 339)
- `appEventBus.ts` / `appEventBusNames.ts` — `EarnHomeBannerDragStateChanged` present

## Verification

- All 10 cherry-picks applied; 155 files changed, +10,334 / -1,125
- Per-file blob sweep after the backport: **151/177 at POST state**; all 26 remaining differences individually accounted for (locale `--ours`, the doc union, generated `translations.ts`, two files modified by a later cherry-pick in the series, three 3-way merges verified above)
- No conflict markers in any changed file
- No `RELEASES.json` / `.env.version` in the diff

⚠️ **Local checks were not run.** The worktree used for this backport has no `node_modules`, so `yarn agent:check --profile pr` cannot start (`Couldn't find the node_modules state file`). Unlike the `x` sync (#12882), no code conflict was hand-resolved here — the only resolutions were generated-locale `--ours` and a doc union — but CI is the gate. Worth particular attention: #12791 lands 27 new files on a branch whose hardware SDK is forked (`@onekeyfe/hd-core` `1.2.0-alpha.129` here vs `1.1.31` on release/v6.5.0), so type-level breakage is plausible and unverified.


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

@originalix

Reviewed commit `1f707f7aa4a0`. This report always reflects the latest reviewed changes.

### Review summary

This synchronization PR brings the hotfix branch in line with several upstream subsystem changes: a redesigned mobile Earn/Borrow navigation and catalog, stricter Swap recipient resolution, gas-account telemetry around existing transaction checks, Perps cache recovery, and universal-search attribution. The foreground UI continues to access privileged storage and transaction services through background boundaries, while the new behavior broadens native navigation and recovery paths without changing cryptographic or authorization rules. Two Earn error paths currently turn transient backend failures into authoritative empty or partial catalogs, so the new primary Earn surfaces can lose valid content until a later retry or cache expiry.

**What needs attention:** Confirm that refresh and fallback failures preserve the last known complete Earn catalog before merging the synchronization.

### Issues to address

- **P2 · Failed refreshes erase the last known Earn catalog**
  - **Impact:** When a pull-to-refresh, pending-transaction refresh, or foreground prefetch encounters a transient `getAvailableAssets` failure, the catch path converts the failure to an empty array and publishes it to the shared Earn state. Previously loaded Trending, Fixed Rate, or Staking content then disappears even though the server did not authoritatively return an empty catalog.
  - **Suggested action:** Only update each available-assets category after a successful request; on failure retain the existing category and clear its loading state in a `finally` block. Add a test covering a failed refresh with pre-existing data.
- **P2 · Partial protocol fan-out is cached as complete**
  - **Impact:** When the symbol-less endpoint is unavailable, the compatibility path fans out by symbol. `getProtocolList` calls without a request ID already translate transport failures into empty arrays, and the worker also suppresses rejected calls, so the aggregation resolves successfully with missing symbols and is cached for five minutes as if complete. Protocol rows and TVL-derived sorting can therefore remain incomplete after a transient failure.
  - **Suggested action:** Make the compatibility fan-out failure-aware: use scoped requests that propagate transport errors and reject the aggregate if any symbol fails, allowing the existing cache-failure path to clear the promise and retry.

### Validation gaps

- The checked HEAD and merge base matched the requested values after investigation.
- `yarn agent:check --profile commit` could not start because Yarn could not find the `node_modules` state file, so lint, TypeScript checks, and targeted tests were not executed in this workspace.
- The added tests cover several individual helpers, but no changed test covers retaining existing Earn data after refresh failure or rejecting an incomplete protocol fan-out.

wabicai (approved):

PR #12899, opened 2026-08-16
fix: sync Pro2 settings and send BLE cancel immediately

wabicai:
## What changed

- Wait for the canonical Protocol V2 `DeviceState` refresh after every Pro2 settings mutation, including wallpaper updates.
- Cover passphrase, label, brightness, language, auto-lock, auto-shutdown, haptic feedback, and wallpaper synchronization.
- Send user-initiated hardware dialog cancellation immediately while retaining debounce behavior for automatic cleanup.
- Recognize both `OneKey Pro2` and the official `OneKey Pro 2` QR wallet device-name prefixes as Pro2.
- Add focused regression coverage for settings synchronization, immediate Pro2 cancellation, and QR wallet device classification.

## Why

When Pro2 settings changed directly on the device, changing an unrelated setting in the App did not always reconcile stale passphrase or label state. The App also delayed manual Cancel through the generic cleanup debounce, which made BLE cancellation appear unresponsive.

Pro2 firmware exports QR wallet metadata with the product name. The spaced official name was previously classified as Pro1, causing QR-created wallets to use the wrong device type and Pro2-specific resources.

## Impact

Any App settings operation now waits for the refreshed Pro2 snapshot, explicit dialog cancellation reaches the SDK immediately without changing automatic cleanup behavior, and Pro2 QR wallets retain the correct device type and resources.

## Related SDK change

- https://github.com/OneKeyHQ/hardware-js-sdk/pull/881

## Validation

- Focused App tests: 80 passed.
- QR wallet device classification tests: 50 passed.
- Oxlint passed for all changed files.
- iOS Debug build succeeded with the existing manual provisioning profile.
- Installed and launched successfully on an iPhone 16 Pro running iOS 26.6.
- Verified the App consumed the exact local SDK build artifact by SHA-256.

PR #12908, opened 2026-08-17
fix: seed canonical device state before hidden wallet passphrase session (OK-59992)

originalix:
[OK-59992](https://onekeyhq.atlassian.net/browse/OK-59992)

----------

## Problem

OK-59992: creating a hidden wallet became noticeably slower. The app-side contributor: `createHWHiddenWallet` performed uncached live `getDeviceState` reads AFTER the hidden-wallet passphrase session was opened (V1 devices with no persisted snapshot). This costs extra device round trips, and a live V1 state read without the passphrase context restores the standard session — forcing the subsequent XFP derivation to re-establish the hidden session (the hazard already documented in the code comment inside `getFeaturesForHwWalletCreate`).

## Changes

- `createHWHiddenWallet`: when `deviceStateInfo` is missing (first run after upgrade), the vendor is not third-party, and the device is not known-V2, read the canonical device state ONCE BEFORE `getPassphraseState`, carry the snapshot in memory, and let every downstream step take its existing fast path — fewer device round trips and no hidden-session clobbering
- Replicate `resolveDeviceStateForHwWalletCreate`'s live-identity guard at the seed site (the seeded fast path would otherwise skip it; behavior stays equivalent to the old no-snapshot flow)
- `isAttachPinMode` now prefers the freshest persisted post-unlock device state (the seeded snapshot predates the PIN prompt, so using it directly would report a stale value in attach-PIN scenarios)
- Backfill `connectProtocol` from the seeded state when it was unknown
- Bump hardware SDK to `1.2.0-alpha.134`, which carries the companion SDK fixes (WebUSB protocol-probe caching + single-round-trip identity validation, see OneKeyHQ/hardware-js-sdk#884)

## Measurements (desktop, same OneKey Pro, same UX flow, silent-call totals)

| Phase | 6.5.0 baseline | Regression (alpha.129) | This PR + SDK fixes |
|---|---|---|---|
| Onboarding wallet creation | 14.0s | 25.6s | **14.7s** |
| Enable passphrase + create hidden wallet | 18.1s | 27.2s | **18.9s** |

Pro2 (V2): no regression on any metric (V2 is explicitly excluded from the seeding path by design).

## Notes

- 6 new unit tests cover seed ordering, fast paths, V2 exclusion, the identity guard, and attach-PIN value sourcing
- This PR no longer carries any SDK version change: the base branch already pins `1.2.0-alpha.142`, which was released from `onekey` after OneKeyHQ/hardware-js-sdk#884 merged and verifiably contains all of its changes (tarball-checked: probe cache, USBDevice identity binding, forced-detection bypass, session gate, Initialize identity merge). Only the three app-side fix commits remain
- Extension release requires deploying the matching iframe version to `jssdk.onekey.so` (`deploy-release-iframe-web`); desktop/TF uses the bundled `static/js-sdk` and is unaffected

[OK-59992]: https://onekeyhq.atlassian.net/browse/OK-59992?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
+++ b/packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
@@ -3606,10 +3606,61 @@ class ServiceAccount extends ServiceBase {
     // createHWHiddenWallet
     return this.backgroundApi.serviceHardwareUI.withHardwareProcessing(
       async () => {
+        // Seed the canonical device state BEFORE opening the hidden-wallet
+        // passphrase session: a live state read on Protocol V1 restores the
+        // standard session, so reading it after getPassphraseState would
+        // clobber the hidden session and cost extra device round trips to
+        // re-establish it (see getFeaturesForHwWalletCreate). Known-V2 devices
+        // are excluded — their flow always reads live state by design. An
+        // unknown-protocol device that turns out V2 pays one redundant read,
+        // but that class is practically empty: V2 device records have always
+        // stored connectProtocol since the protocol field was introduced.
+        let seededDbDevice = dbDevice;
+        let seededConnectProtocol = connectProtocol;
+        const hiddenWalletVendorProfile = getVendorProfile(
+          dbDevice.vendor ?? EHardwareVendor.onekey,
+        );
+        if (
+          !seededDbDevice.deviceStateInfo &&
+          !hiddenWalletVendorProfile.isThirdParty &&
+          seededConnectProtocol !== 'V2'
+        ) {
+          const seededState =
+            await this.backgroundApi.serviceHardware.getDeviceState({
+              connectId: compatibleConnectId,
+              params: {
+                scope: 'runtime',
+                ...(seededConnectProtocol
+                  ? { connectProtocol: seededConnectProtocol }
+                  : {}),
+              },
+            });
+          // Mirror the live-identity guard in resolveDeviceStateForHwWalletCreate:
+          // with a seeded snapshot its fast path returns early, so the guard
+          // must run here to keep the old no-snapshot fail-fast behavior.
+          if (
+            seededState.status.mode === 'normal' &&
+            !seededState.identity.deviceId
+          ) {
+            throw new OneKeyLocalError(
+              'Unable to resolve live hardware device identity',
+            );
+          }
+          // DEVICE.STATE persistence is async; carry the snapshot in memory so
+          // downstream steps take their fast paths deterministically.
+          seededDbDevice = { ...seededDbDevice, deviceStateInfo: seededState };
+          if (
+            !seededConnectProtocol &&
+            (seededState.protocol === 'V1' || seededState.protocol === 'V2')
+          ) {
+            seededConnectProtocol = seededState.protocol;
+          }
+        }
+
         const passphraseState = await getHwHiddenWalletPassphraseState({
           vendor: dbDevice.vendor,
           connectId: compatibleConnectId,
-          dbDevice,
+          dbDevice: seededDbDevice,
           serviceHardware: this.backgroundApi.serviceHardware,
           serviceThirdPartyHardware:
             this.backgroundApi.serviceThirdPartyHardware,
@@ -3630,14 +3681,14 @@ class ServiceAccount extends ServiceBase {
 
         // TODO save remember states
         const resolvedFeatures = await this.getFeaturesForHwWalletCreate({
-          dbDevice,
+          dbDevice: seededDbDevice,
           compatibleConnectId,
         });
         const dbWallet = await this.createHWWalletBase({
-          device: deviceUtils.dbDeviceToSearchDevice(dbDevice),
+          device: deviceUtils.dbDeviceToSearchDevice(seededDbDevice),
           features: resolvedFeatures,
-          connectProtocol,
-          deviceState: dbDevice.deviceStateInfo,
+          connectProtocol: seededConnectProtocol,
+          deviceState: seededDbDevice.deviceStateInfo,
           passphraseState,
           fillingXfpByCallingSdk: true,
         });
@@ -3666,9 +3717,28 @@ class ServiceAccount extends ServiceBase {
             await this.backgroundApi.serviceAccountProfile.isSoftwareWalletOnlyUser(),
         });
 
+        // The seeded snapshot predates the passphrase prompt, so its
+        // unlockedAttachPin can be stale when the device was unlocked with the
+        // attach PIN during this flow. Prefer the state persisted from the
+        // DEVICE.STATE events emitted by the passphrase/derivation calls above.
+        let isAttachPinMode = resolvedFeatures.unlockedAttachPin;
+        try {
+          const latestDbDevice =
+            await this.backgroundApi.serviceHardware.getDeviceByConnectId({
+              connectId,
+            });
+          const latestUnlockedAttachPin =
+            latestDbDevice?.deviceStateInfo?.status?.unlockedAttachPin;
+          if (typeof latestUnlockedAttachPin === 'boolean') {
+            isAttachPinMode = latestUnlockedAttachPin;
+          }
+        } catch {
+          // keep the resolved-features fallback
+        }
+
         return {
           ...dbWallet,
-          isAttachPinMode: resolvedFeatures.unlockedAttachPin,
+          isAttachPinMode,
         };
       },
       {
--- a/packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
+++ b/packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
@@ -3683,24 +3683,30 @@ class ServiceAccount extends ServiceBase {
         // unlock / attach-PIN status); wait for their persistence and prefer
         // the persisted post-unlock snapshot over the pre-unlock seed for
         // everything derived or stored below. Pure DB reads — no device I/O,
-        // so the hidden session established above is never touched.
-        await this.backgroundApi.serviceHardware.waitForDeviceStateSync({
-          connectIds: [
-            compatibleConnectId,
-            dbDevice.connectId,
-            dbDevice.deviceId,
-            seededDbDevice.deviceStateInfo?.identity.serialNo,
-          ],
-        });
-        const postUnlockDbDevice =
-          await this.backgroundApi.serviceHardware.getDeviceByConnectId({
-            connectId,
+        // so the hidden session established above is never touched. Gated like
+        // the seeding block: third-party vendors are excluded, and an empty
+        // connectId must not reach getDeviceByConnectId — getDeviceByQuery
+        // would degenerate to "first OneKey device" and overlay an unrelated
+        // device's state.
+        if (connectId && !hiddenWalletVendorProfile.isThirdParty) {
+          await this.backgroundApi.serviceHardware.waitForDeviceStateSync({
+            connectIds: [
+              compatibleConnectId,
+              dbDevice.connectId,
+              dbDevice.deviceId,
+              seededDbDevice.deviceStateInfo?.identity.serialNo,
+            ],
           });
-        if (postUnlockDbDevice?.deviceStateInfo) {
-          seededDbDevice = {
-            ...seededDbDevice,
-            deviceStateInfo: postUnlockDbDevice.deviceStateInfo,
-          };
+          const postUnlockDbDevice =
+            await this.backgroundApi.serviceHardware.getDeviceByConnectId({
+              connectId,
+            });
+          if (postUnlockDbDevice?.deviceStateInfo) {
+            seededDbDevice = {
+              ...seededDbDevice,
+              deviceStateInfo: postUnlockDbDevice.deviceStateInfo,
+            };
+          }
         }
 
         // TODO save remember states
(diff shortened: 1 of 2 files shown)


originalix:

@originalix

Reviewed commit `18bbc0b60588`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes hidden hardware-wallet creation in the background runtime so Protocol V1 OneKey devices without persisted canonical state are sampled before the passphrase session opens, preventing a later live read from restoring the standard session. The sampled state and inferred protocol are carried through feature projection and wallet creation, while the hardware SDK suite advances consistently to alpha.134 across app and CLI consumers. The implementation also tries to recover post-unlock attach-PIN status from the local device record, but it does not synchronize queued DEVICE.STATE persistence before reusing the pre-unlock snapshot.

**What needs attention:** Confirm that the background runtime waits for DEVICE.STATE persistence and passes the post-passphrase state into wallet creation instead of overwriting it with the pre-passphrase snapshot.

### Issues to address

- **P2 · Pre-unlock snapshot can overwrite the post-passphrase device state**
  - **Impact:** For a Protocol V1 hidden wallet unlocked with an attach PIN, seededDbDevice still contains the state captured before getPassphraseState. Passing that snapshot into createHWWalletBase causes LocalDb.createHwWallet to assign it back to the existing device record. If the asynchronous DEVICE.STATE task already persisted the newer unlocked state, wallet creation overwrites it, and the later database read can return stale unlockedAttachPin data. This leaves the canonical device record stale and selects the ordinary existing-wallet message instead of the same-PIN explanation.
  - **Suggested action:** After getPassphraseState, await ServiceHardware.waitForDeviceStateSync for both original and compatible connect IDs, reload the device record, and pass its post-passphrase deviceStateInfo into createHWWalletBase. Add a test that models DEVICE.STATE persistence completing before the wallet transaction.

### Validation gaps

- The targeted Jest test could not execute because the workspace has no installed dependency state; Yarn reported that the `node_modules` state file was missing.
- The alpha.134 package source, release notes, advisories, and install/runtime hooks could not be inspected because the packages are not installed and external package access is unavailable. The lockfile shows only the existing hardware package family moving versions, with no new package names.
- Existing GitHub reviews, inline comments, general comments, and thread state were unavailable, so the inline wording requires a duplicate-comment check before posting.
- No physical Protocol V1 attach-PIN flow was exercised. The current mocked test does not count as passing because it does not model the queued DEVICE.STATE write and the later LocalDb overwrite. Passing validation should demonstrate that the post-creation device record retains the post-passphrase attach-PIN state and that the same-PIN message is selected.
- `git diff --check` and JSON syntax checks passed; HEAD and merge base remained unchanged.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -3666,9 +3717,28 @@ class ServiceAccount extends ServiceBase {
...
+              connectId,
+            });
+          const latestUnlockedAttachPin =
+            latestDbDevice?.deviceStateInfo?.status?.unlockedAttachPin;
+          if (typeof latestUnlockedAttachPin === 'boolean') {
+            isAttachPinMode = latestUnlockedAttachPin;
+          }
+        } catch {
+          // keep the resolved-features fallback
+        }
```
sidmorizon: **Attach-PIN unlock indicator for a newly created hidden wallet can still report the outdated value**

Severity: `non-severe`

The freshest saved device status is looked up (`getDeviceByConnectId` at `packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts:3726-3729`) only after wallet creation has already written the older pre-passphrase snapshot back over the saved status, so the reported attach-PIN unlock flag is usually the outdated one this code was meant to correct.
Impact: After creating a hidden wallet by unlocking with the attach PIN, the user can be shown the wrong confirmation message.

<details>
<summary>How the saved device status is overwritten before it is read back</summary>

- The new seeding block reads the live state before the passphrase prompt and carries it in memory (`packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts:3628-3651`), then passes it as `deviceState` into `createHWWalletBase` (`ServiceAccount.ts:3691`).
- `createHWWalletBase` forwards it to `localDb.createHwWallet` (`ServiceAccount.ts:3946`), which sanitizes it into `initialDeviceState` (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:6274-6299`) and then unconditionally writes it onto the existing device record: `item.deviceState = deviceToAdd.deviceState ?? item.deviceState` (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:6397`). This bypasses the freshness/ordering checks that guard the DEVICE.STATE event path (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:5213-5241`).
- Consequently, any `unlockedAttachPin: true` persisted from events during `getPassphraseState`/XFP derivation is replaced by the pre-passphrase snapshot before line 3726 reads it back; only a DEVICE.STATE event that happens to land after the creation transaction would yield the fresh value, making the correction racy at best.
- A secondary effect: the stored device record temporarily regresses to a pre-passphrase state snapshot for other consumers.

</details>

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -3666,9 +3717,28 @@ class ServiceAccount extends ServiceBase {
...
+        // The seeded snapshot predates the passphrase prompt, so its
+        // unlockedAttachPin can be stale when the device was unlocked with the
+        // attach PIN during this flow. Prefer the state persisted from the
+        // DEVICE.STATE events emitted by the passphrase/derivation calls above.
+        let isAttachPinMode = resolvedFeatures.unlockedAttachPin;
+        try {
+          const latestDbDevice =
+            await this.backgroundApi.serviceHardware.getDeviceByConnectId({
+              connectId,
+            });
```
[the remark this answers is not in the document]
originalix: Fixed in c6f4f2c217. The final status read is now preceded by `serviceHardware.waitForDeviceStateSync(...)`, which drains the DEVICE.STATE persistence queues (including the yield that lets a split bg runtime register the pending task) before `getDeviceByConnectId` — so the attach-PIN status reflects every event received up to that point. Additionally, the same drain + re-read now happens right after `getPassphraseState`, so `resolvedFeatures` itself is already derived from the post-unlock snapshot; the final read only picks up events emitted later by the XFP/address derivation calls. Pure DB reads, no extra device I/O.

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -3630,14 +3681,14 @@ class ServiceAccount extends ServiceBase {
...
         });
         const dbWallet = await this.createHWWalletBase({
-          device: deviceUtils.dbDeviceToSearchDevice(dbDevice),
+          device: deviceUtils.dbDeviceToSearchDevice(seededDbDevice),
           features: resolvedFeatures,
-          connectProtocol,
-          deviceState: dbDevice.deviceStateInfo,
+          connectProtocol: seededConnectProtocol,
+          deviceState: seededDbDevice.deviceStateInfo,
           passphraseState,
```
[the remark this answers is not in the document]
originalix: Fixed in c6f4f2c217. After `getPassphraseState` succeeds, the flow now drains the DEVICE.STATE persistence queues (`waitForDeviceStateSync`) and re-reads the persisted device record; when a persisted snapshot exists it replaces the pre-unlock seed before `getFeaturesForHwWalletCreate` and `createHWWalletBase` run — so the state stored with the wallet is the post-unlock snapshot persisted from the passphrase call's own events. The pre-unlock seed remains only as the fallback when no persisted state is available yet, which matches the pre-change fast-path behavior for devices with existing snapshots. No extra device round trips introduced.

package.json
```diff
@@ -158,21 +158,21 @@
...
-    "@onekeyfe/hwk-ledger-connector-webhid": "1.2.0-alpha.133",
-    "@onekeyfe/hwk-trezor-adapter": "1.2.0-alpha.133",
-    "@onekeyfe/hwk-trezor-connector-electron-ble": "1.2.0-alpha.133",
-    "@onekeyfe/hwk-trezor-connector-rn-ble": "1.2.0-alpha.133",
-    "@onekeyfe/hwk-trezor-connector-webusb": "1.2.0-alpha.133",
+    "@onekeyfe/hd-ble-sdk": "1.2.0-alpha.134",
+    "@onekeyfe/hd-common-connect-sdk": "1.2.0-alpha.134",
+    "@onekeyfe/hd-core": "1.2.0-alpha.134",
+    "@onekeyfe/hd-shared": "1.2.0-alpha.134",
+    "@onekeyfe/hd-transport": "1.2.0-alpha.134",
```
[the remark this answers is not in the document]
originalix: Fixed in 529279a332. Agreed — alpha.134 was built before the SDK branch was rebased and would have dropped the alpha.130–133 changes that 2b476be1 (#12899) pairs with. `1.2.0-alpha.136` has now been published from the rebased SDK branch (onekey @ #881 + the two OK-59992 fixes, see OneKeyHQ/hardware-js-sdk#884), and this PR now pins alpha.136. Tarball content was verified after publish: the OK-59992 fixes (`staleProtocolPaths`, `assertExpectedDeviceIdentity`) and the #881 content (`refreshProtocolV2SettingsAfterMutation`) are all present in the published dist.

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
originalix: Fixed in c6f4f2c217 (same mechanism as the codex thread above): the flow now drains the DEVICE.STATE persistence queues via `waitForDeviceStateSync` at two points — right after `getPassphraseState` (so `resolvedFeatures` is derived from the persisted post-unlock snapshot) and again before the final `getDeviceByConnectId` read that sources `isAttachPinMode` (so status changes emitted during XFP/address derivation are also captured). The queue drain loops until the per-device queues are empty, covering the async persistence window you describe.

packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts
```diff
@@ -3628,16 +3679,40 @@ class ServiceAccount extends ServiceBase {
...
+        const postUnlockDbDevice =
+          await this.backgroundApi.serviceHardware.getDeviceByConnectId({
+            connectId,
+          });
+        if (postUnlockDbDevice?.deviceStateInfo) {
+          seededDbDevice = {
+            ...seededDbDevice,
+            deviceStateInfo: postUnlockDbDevice.deviceStateInfo,
+          };
+        }
```
sidmorizon, line 3704: **Hidden wallet can be saved with another device's state when the device record has no connect id**

Severity: `non-severe`

The freshest device snapshot is looked up purely by connect id (`getDeviceByConnectId({ connectId })` at `packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts:3695-3698`) without the vendor and without guarding against an empty connect id, so a hidden wallet can end up storing a completely different device's saved state.
Impact: For USB-connected third-party hardware whose stored record has no connect id, the newly created hidden wallet is saved with an unrelated device's identity/state, which can later cause wrong device matching and "device not the same" style failures.

<details>
<summary>Why an empty connect id matches an arbitrary device</summary>

`localDb.getDeviceByQuery` (`packages/kit-bg/src/dbs/local/LocalDbBase.ts:8716-8783`) only applies the connect-id predicate when `connectId` is truthy and defaults the vendor filter to `EHardwareVendor.onekey`. With `connectId === ''` the predicate degenerates to "first OneKey-vendor device", so `postUnlockDbDevice` is an unrelated record and its `deviceStateInfo` unconditionally replaces `seededDbDevice.deviceStateInfo` (`ServiceAccount.ts:3699-3704`). That value is then passed as `deviceState` into `createHWWalletBase`, where for third-party vendors `resolveDeviceStateForHwWalletCreate` returns the passed-in state as-is (`packages/kit-bg/src/services/ServiceAccount/deviceStateForHwWalletCreate.ts:52-55`) and it is persisted with the new wallet.

Empty connect ids for third-party USB devices are explicitly allowed — `createHWWalletBase` only rejects a missing connect id for *non-USB* third-party hardware (`ServiceAccount.ts:3870-3878`). Note also the seeding block above deliberately skips third-party vendors, while this refresh block does not; passing the vendor (and skipping the lookup when the connect id is empty / the vendor is third-party) would keep both blocks consistent.
</details>

originalix: Fixed in 289ef284f8. Both post-unlock lookups (the state refresh and the isAttachPinMode read) are now gated on `connectId && !hiddenWalletVendorProfile.isThirdParty` — matching the seeding block's existing gate, so an empty connectId can never reach `getDeviceByConnectId` (where `getDeviceByQuery` would degenerate to "first OneKey-vendor device"), and third-party records are never overlaid with another device's state. Regression test added: a device record with an empty connectId performs zero `getDeviceByConnectId` lookups and the seeded state is stored unchanged.

wabicai (dismissed):

wabicai (approved):

PR #12918, opened 2026-08-18
fix: keep the perps deeplink target over the cold-start restore (OK-60543)

Minnzen:
[OK-60543](https://onekeyhq.atlassian.net/browse/OK-60543)

----------

## Summary

- Record a short-lived instrument intent when a banner/push deeplink picks a Perps coin, and consume it in the cold-start symbol select before the persisted restore runs.
- The restore keeps winning when no intent is pending, so the pair flip #12680 was added to prevent stays fixed.
- Adds four regression cases to the existing decision-function suite.
- Collapses the four sequential background hops that ran before the symbol was decided into one, cutting up to ~470ms off the cold-start wait.

## Intent & Context

QA reported that tapping the home banner for a `para:*` market on iOS 6.5.0 lands on the Perps tab but keeps the previous symbol. Two separate defects were found while diagnosing it; this PR fixes the second, which is the one users actually hit.

The reproduction QA settled on is precise: switch to any symbol other than the banner target, kill the app, cold start, tap the banner — the first tap never switches, the second one does.

Despite being found on `para:*`, the defect has nothing to do with HIP-3 or dex resolution. It affects every symbol (BTC, HYPE, `xyz:*`, `para:*`) and every context-less entry point (home banner, push, desktop tray, universal search). It only surfaced on `para:*` because that is what the banner currently promotes.

Tracked as OK-60543.

## Root Cause

Introduced by #12680 (2026-08-07, shipped in release #5 on 2026-08-08). `x` carries the same code, so 6.6.0 is affected too.

1. The deeplink is handled by `NotificationHandlerContainer`, which has no perps context. It writes the background `perpsActiveAssetAtom`, then emits `PerpSwitchActiveInstrument`.
2. That event is consumed by a plain `appEventBus.on` inside `PerpsGlobalEffects`. On a cold start the Perp page has not mounted yet, so the event is dropped. (The sibling notification page-navigation processor is durable; this listener is not.)
3. The page mounts, `selectInitialSymbol` claims the process-wide latch, and passes `activeTradeInstrumentAtom` as `preferredInstrument`.
4. `activeTradeInstrumentAtom` is declared with `coldStartCache: true`, so on a cold start it holds the *previous session's* symbol.
5. `buildInitialTradeInstrumentSwitchParams` returns `preferredInstrument` unconditionally, without ever looking at the background atoms — forcing the market back to the stale symbol.

The comment on that branch states its premise: the optimistic instrument is "never the staler record" because the UI writes it synchronously at the start of a switch while the background atoms are written near the end. That holds within a session. A deeplink cold start inverts it — the background write happens during *this* boot, after the UI snapshot was persisted.

The second tap succeeds because the latch is already claimed (`preferredInstrument` is not passed on non-claiming runs) and the event listener is mounted by then.

Confirmed on-device (2026-08-18 10:59:21 cold start, bundle 19453079):

```
10:59:22  change_active_asset_end {assetId:180024, coin:"para:SMCI", hasUniverse:true}   <- background resolved it correctly
10:59:23  perpPageView {pageName:"Perp", source:"walletBanner"}                          <- home banner entry
10:59:23  initial_symbol_start {activeCoin:"BTC"}                                        <- cold-start cache
10:59:24  change_active_asset_start {coin:"BTC"}                                         <- overwritten
10:59:25  active_asset_ctx_atom_set_first {coin:"BTC", markPx:"64057.0"}                 <- page renders BTC
```

## Design Decisions

**Chosen: an explicit, short-lived intent that outranks the restore.**

The premise behind `preferredInstrument` is sound for ordinary cold starts, so the fix must not weaken it. What distinguishes the broken case is not freshness but category: "the user just asked for this market" is a different signal from "restore what was open". Modelling it as its own signal keeps the two orderings independent, so no timestamp comparison between two persisted values is needed.

The intent lives on the `ServiceHyperliquid` singleton, alongside the existing `tryClaimInitialSymbolSelect` latch — same lifetime, same reason for being there. It is deliberately in-memory: both the write (deeplink handler) and the read (symbol select) happen within one cold start, so persisting it would add a migration surface and a cross-restart staleness question for no benefit. A 30s TTL bounds an intent whose page never opened.

**Rejected: making `PerpSwitchActiveInstrument` durable.** It would buffer the event past mount, but the drain and `selectInitialSymbol`'s own `force: true` switch would still race — whichever lands second wins. That trades a deterministic bug for a nondeterministic one.

**Simplified after review.** The first shape carried a parallel `deeplinkIntent` parameter through the builder, a `claimed` gate at the call site, a helper to express that gate, and a wall-clock TTL. A Codex pass showed one mechanism covers all of it: the pending market reuses the existing `preferredInstrument` slot, and the setter no-ops once the latch is taken. After the claim the page is live and switches through the event bus, so anything recorded later could only override a market the user picked since — making "first mount only" a property of the store rather than a rule each caller upholds. `buildInitialTradeInstrumentSwitchParams` and its ten tests are back to their original state, so #12680's restore semantics are untouched rather than re-argued. The TTL is gone with its own defect: it discarded the target when app lock kept the user past five minutes.

**Rejected: comparing persisted timestamps on both sides.** Would require stamping both the background atom and the cold-start cache entry, and still leaves the ordering implicit rather than stated.

The intent is consumed on every `selectInitialSymbol` run so nothing lingers, but only the claiming run acts on it. Review caught the first version applying it on every run: a notification handled by an already-mounted page left the intent live for its whole TTL, and the next blur-to-focus forced that coin back over a market the user had picked since. A mounted page already switches through the event bus, and the non-claiming run resyncs from the background atoms when the two sides diverge, so it never needed the intent.

## Changes Detail

- `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts` — adds `setPendingInstrumentIntent` / `consumePendingInstrumentIntent` plus the 30s TTL constant. Pure addition; no existing code touched.
- `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx` — records the intent *before* `changeActiveAsset`, so no cold start can observe the coin without the intent that explains it.
- The same emit-before-mount pattern exists at every other context-less entry, and review confirmed each still lost its first tap. `useTrayDataProvider.desktop.ts`, `UniversalSearchPerpItem.tsx`, `Home/pages/PerpsContainer.tsx`, `PerpetualTradingBanner.tsx` and `Market/hooks/usePerpsNavigation.ts` now record the intent the same way; the Home card passes its own mode through, since it can open a spot pair.
- `packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx` — consumes the intent after the latch claim, feeds it to both the divergence check and the cold-start switch, and marks `initial_symbol_deeplink_intent` for log-based verification.
- `packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.ts` — `deeplinkIntent` takes precedence over `preferredInstrument`; behaviour is byte-for-byte unchanged when no intent is pending.
- `packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.test.ts` — four cases: intent beats the restore, intent pulls the mode from spot back to perp, the restore still wins with no intent, and an empty-coin intent does not shadow the restore.

Second commit (`perf:`), measured on the device trace from the first:

- `prepareInitialSymbolSelect` resolves the claim, the intent, the instrument target and the trading universe in one hop, in the same order as before. The universe read stays behind the claim flag, so a non-claiming run — which happens on every route focus — does no more work than it used to.
- On that trace the last three hops cost 24ms + 223ms + 223ms of a 1.3s wait; on an idle background the same three cost ~30ms. The win is the variance, and the variance peaks exactly during cold start.
- `initial_symbol_prepare` records the combined hop. It cannot be derived from the surrounding marks: `initial_symbol_start` and `initial_symbol_claimed` are emitted under different `sessionId`s, and `elapsed` is relative to each session's own origin.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS/Android), Desktop, Extension — every runtime where a context-less caller can pick the active instrument
- **Risk Areas**:
  - The precedence flip in `buildInitialTradeInstrumentSwitchParams` is the only behavioural change to an existing path. It is inert unless an intent is pending, and the nine pre-existing cases from #12680 still pass unchanged.
  - The intent is in-memory on the background singleton. On split-runtime targets (iOS/Android/extension) both the write and the read cross the background proxy into that one runtime, so there is no per-runtime copy to diverge.
  - There is exactly one claiming run per process, so a leftover intent is inert after it: it can only be discarded by the next run or by the TTL, never applied.
  - The `perf:` commit changes no behaviour: same four operations, same order, same claim gating. The only observable difference is that `initial_symbol_claimed` is now emitted after the combined hop rather than partway through it.

Not addressed here, deliberately: the first frame still renders the previously active symbol before the switch lands, because `activeTradeInstrumentAtom` hydrates synchronously from its cold-start cache. Seeding that first frame from the intent would remove the flash entirely, but it would also make the "coin known, assetId pending" state the common case on every deeplink cold start, and `views/Perp` has ~139 `assetId` references of which only a minority guard against undefined. On a leveraged trading screen that is not a change to bundle with a hotfix.

## Test plan

- [x] `npx jest` over Perp, Market, Home, UniversalSearch, hooks and the Hyperliquid service — 963 tests / 120 suites pass. `buildInitialTradeInstrumentSwitchParams` and its tests are unmodified, which is the check that the restore path still behaves as #12680 left it. The one-shot store itself has no unit test: `ServiceHyperliquid` cannot be constructed under jest — importing it fails on the module graph, the same reason #12680 recorded no isolated harness for this path.
- [x] `npx tsc -p ./tsconfig.json --noEmit` — no errors in the changed files
- [x] Device-verified on the `fix:` commit: cold start via banner switched to `para:SMCI`, with `initial_symbol_deeplink_intent` present and no follow-up `service_change_active_asset_start` for the previous symbol
- [ ] Re-verify on a build of the full branch — the `perf:` commit touches the same path, so the earlier device run does not carry over
- [ ] Cold-start repro on device, per OK-60543: disable the dev-settings test endpoint (it rolls the client back to the builtin bundle and masks this), switch to BTC in Perps, wait ~5s so the cold-start cache flushes, kill the app, cold start, tap the `para:*` banner
- [ ] Verify in logs: `initial_symbol_pending_instrument {coin:"<target>"}` appears, no `service_change_active_asset_start` for the previous symbol follows it, and `active_asset_ctx_atom_set_first` reports the target coin
- [ ] Confirm an ordinary cold start with no deeplink still restores the last symbol without a visible pair flip (the #12680 behaviour)
- [ ] Regression for the review catch: with the Perp page already open, tap a banner, switch to a different market by hand, then leave and re-enter the tab within 30s — the hand-picked market must stay
- [ ] Confirm Perp tab detach/remount still does not re-run the initial select (OK-53208 latch), and that switching symbols from the token selector, tray and universal search is unaffected
- [ ] First entry of a launch from each context-less caller opens the market that was tapped: desktop tray, universal search, Home perps card (both perp and spot), market banner, market list
- [ ] With app lock on, a cold start whose unlock takes any length of time still opens the tapped market — the value is consumed by the first mount, not by a clock
- [ ] Read `initial_symbol_prepare.elapsedMs` to quantify the hop, and check whether `hasCachedTradingUniverse` is now `true` — the ~1.1s universe wait seen on the first trace should be a one-off from the dex list growing

[OK-60543]: https://onekeyhq.atlassian.net/browse/OK-60543?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix:

@Minnzen

Reviewed commit `9450f31d7e52`. This report always reflects the latest reviewed changes.

### Review summary

The PR moves pre-navigation Perps market intent into the background Hyperliquid service and batches the first-symbol claim, authoritative target read, and trading-universe read so context-less entry points can survive the page-mount boundary. Tray, notification, Home, Market, banner, and Universal Search paths now record intent before switching tabs, while already-mounted pages continue through the event bus. This preserves existing signing and authorization boundaries and reduces bridge latency, but the resolver currently breaks the discriminated instrument type, the one-shot value lacks an abandonment lifecycle, and its precedence can overwrite a newer foreground choice.

**What needs attention:** Verify the pending intent's type contract, lifecycle, and precedence when navigation is abandoned or a second instrument is selected before initial metadata loading finishes.

### Issues to address

- **P1 · Resolver widens mode and is not assignable to the instrument union**
  - **Impact:** `resolveInitialPreferredInstrument()` returns one object whose `mode` is inferred as `'perp' | 'spot'`. The receiving `preferredInstrument` field expects `IActiveTradeInstrument`, whose `perp` and `spot` members are separately discriminated, so this value is not assignable and the TypeScript check fails at the new call site.
  - **Suggested action:** Branch on `pendingInstrument.mode` and return explicitly discriminated `perp` and `spot` objects, then run the package TypeScript check.
- **P2 · Pending market intent can outlive the navigation that created it**
  - **Impact:** `setPendingInitialTradeInstrument()` retains only `{ coin, mode }` until `prepareInitialSymbolSelect()` runs. If a notification, tray action, banner, Search result, or Market navigation records the value but never mounts Perps, a later unrelated first mount consumes the stale intent and opens a market the user did not select for that launch.
  - **Suggested action:** Bind the pending value to a navigation token or bounded expiry and explicitly cancel it when the initiating navigation does not reach the first Perps mount.
- **P2 · A newer market pick can be overwritten by the pending cold-start intent**
  - **Impact:** During a cache-miss cold start, initial selection awaits `refreshTradingMeta()`. A later instrument switch updates `activeTradeInstrumentRef.current` optimistically, but `resolveInitialPreferredInstrument()` still gives the earlier pending intent unconditional precedence; the subsequent forced switch returns to the older market and can clear the user's current order form.
  - **Suggested action:** Record the active instrument observed when preparation completes and apply the pending intent only if no newer foreground instrument has replaced that snapshot before switch parameters are built.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` passed.
- The focused Jest test could not run because Yarn reported a missing `node_modules` state file.
- Type checking could not run for the same dependency-state reason; the assignability defect is established directly from the declared return and parameter types.
- Existing PR reviews, inline comments, general comments, and thread state could not be fetched because GitHub API access was unavailable; inline candidates therefore require a human duplicate check.
- Runtime cold-start behavior was not exercised. Merely opening the Perps tab is not sufficient: the final pass condition is that the tapped coin matches the main-runtime active instrument, the background target, and fresh subscription data after an ordinary cold start, abandoned navigation, and a second selection during metadata refresh.

sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -963,6 +971,16 @@ function useHyperliquidSymbolSelect() {
...
+      // Consumed on every run, claiming or not, so an intent left behind by a
+      // deeplink that never reached this page cannot hijack a later init.
+      const deeplinkIntent =
+        await backgroundApiProxy.serviceHyperliquid.consumePendingInstrumentIntent();
+      if (deeplinkIntent?.coin) {
+        markPerpsColdStartPerf('initial_symbol_deeplink_intent', {
+          coin: deeplinkIntent.coin,
+          mode: deeplinkIntent.mode,
+        });
+      }
```
sidmorizon: **A recently opened notification can pull the trading page back to its coin after the user picks another market**

Severity: `non-severe`

The remembered deeplink coin is only cleared by the page's initial-symbol run (`consumePendingInstrumentIntent()` at `packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:977`) and not when the already-open trading page handles the switch directly, so for the next 30 seconds a tab re-focus can force the market back to that coin over the one the user just chose.
Impact: Shortly after opening a notification/banner while already on the trading page, changing to another market and returning to the tab silently reverts the chart and order form to the notification's market.

<details>
<summary>How the unconsumed intent survives the event-bus path and is replayed on the next focus</summary>

- `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx:153-158` records the intent, then emits `PerpSwitchActiveInstrument`.
- If the Perp page is already mounted, that event is handled by `useHyperliquidInstrumentSwitchRequest` (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1500-1521`), which switches the instrument but never clears the pending intent.
- If no tab focus transition happens (user was already on the Perp tab), `selectInitialSymbol` does not run, so the intent stays alive in the background singleton for the full 30s TTL (`PENDING_INSTRUMENT_INTENT_TTL_MS`, `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:225-227`).
- The user then picks another coin. Leaving the Perp tab and returning triggers `selectInitialSymbol` (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1145-1151`), which consumes the still-valid intent and passes it as `deeplinkIntent`; `buildInitialTradeInstrumentSwitchParams` gives it precedence over `preferredInstrument`, so the divergence check trips and the forced switch lands back on the notification's coin.

</details>

sidmorizon:
@cursoragent review

sidmorizon:
@cursoragent review

originalix (commented):
@Minnzen

Reviewed commit `41a22a4f6bdb`.

- **P2 · Mounted-page deeplink path never clears the pending intent**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Confirmed by tracing the path: mounted page handles the switch via the event bus, `selectInitialSymbol` never runs, intent survives the TTL. User then picks another market, and the next blur→focus runs with `claimed === false`, consumes the intent and forces the notification coin back.

Fixed in 0b29bf7b35. Consumption stays unconditional so nothing lingers; application is gated on the claiming run.

Minnzen: Right — the comment said consume-always while the code did apply-always. Fixed in 0b29bf7b35 exactly as described: still consumed on every run, fed into `buildSwitchParamsFromTarget` only when `claimed`.

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -1067,6 +1087,7 @@ function useHyperliquidSymbolSelect() {
         preferredInstrument: claimed
           ? activeTradeInstrumentRef.current
           : undefined,
+        deeplinkIntent,
```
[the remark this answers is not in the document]
Minnzen: Applied in 0b29bf7b35, at both call sites (the divergence check and the cold-start switch) via a single `appliedDeeplinkIntent = claimed ? deeplinkIntent : undefined`.

Added a regression case pinning the invariant your reasoning rests on — that the non-claiming run resyncs off the background atoms and does not need the intent.

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
```diff
@@ -147,6 +147,15 @@ function BaseNotificationHandlerContainer() {
 
       if (perpToken) {
         try {
+          // Recorded before the atom write so no cold start can observe the
+          // coin without the intent: on that path the Perp page restores its
+          // persisted instrument, which predates this tap and would win.
+          await backgroundApiProxy.serviceHyperliquid.setPendingInstrumentIntent(
```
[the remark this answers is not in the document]
Minnzen: Verified the scenario and fixed the outcome, though not with an ack in the listener.

Gating application on `claimed` (0b29bf7b35) closes the lifecycle: there is exactly one claiming run per process, so once it has happened no later run can act on a leftover intent — it is inert, then cleared by the next run or by the TTL.

Acking inside `useHyperliquidInstrumentSwitchRequest` would add a background round-trip to every `PerpSwitchActiveInstrument`, and the tray, universal search and market-page emitters never set an intent. Happy to add it if you still want the intent cleared eagerly rather than merely inert.

originalix (commented):
@Minnzen

Reviewed commit `0b29bf7b350d`.

- **P2 · Non-claiming remount can discard deeplink during mixed atom update**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -566,6 +579,61 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+  @backgroundMethod()
+  async consumePendingInstrumentIntent(): Promise<
+    { coin: string; mode: ITradingMode } | undefined
+  > {
+    const intent = this._pendingInstrumentIntent;
+    this._pendingInstrumentIntent = undefined;
+    if (!intent) {
+      return undefined;
+    }
+    if (Date.now() - intent.createdAt > PENDING_INSTRUMENT_INTENT_TTL_MS) {
```
sidmorizon: # 🟡 P2 应用锁场景下 30s TTL 会静默吞掉 deeplink

## 问题

`createdAt` 在 `setPendingInstrumentIntent` 里用点击瞬间的 `Date.now()` 记录，而真正消费它的 claiming run 要等 Perp 页面挂载后才发生。开启应用锁时这两者之间隔着整个解锁流程——本文件同一个 handler 里的注释就写明了这一点：

```
// Never while locked: resumeSubscriptions enables the handler
// unconditionally, and the user has to unlock before anything
// renders anyway
```

也就是说锁屏期间 Perp 页面根本不渲染，`prepareInitialSymbolSelect` 不会被调用，但 `PENDING_INSTRUMENT_INTENT_TTL_MS` 的 `30s` 计时照跑。

## 影响

冷启动 + 应用锁 + 解锁耗时超过 `30s`（密码输错一次、生物识别失败后回退密码）时，intent 被判定过期返回 `undefined`，`appliedDeeplinkIntent` 为空，`preferredInstrument`（持久化恢复的旧 instrument）重新胜出，退化回本 PR 要修的原始 bug。因为 intent 已被 consume，后续没有任何补偿路径，用户看到的是「点了 A 币的推送，打开却停在上次的 B 币」，且无任何日志区分「过期」与「本来就没有」。

## 建议

让 TTL 只统计「可消费的时间」，而不是墙钟时间。最小改动是在解锁完成时刷新 `createdAt`，或在 `consumePendingInstrumentIntent` 里对锁屏期间不计时；也可以区分过期与缺失的返回值，至少在 `markPerpsColdStartPerf` 里打出 `expired: true` 便于线上定位：

```ts
if (Date.now() - intent.createdAt > PENDING_INSTRUMENT_INTENT_TTL_MS) {
  markPerpsColdStartPerf('pending_instrument_intent_expired', {
    coin: intent.coin,
    ageMs: Date.now() - intent.createdAt,
  });
  return undefined;
}
```

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -957,21 +965,41 @@ function useHyperliquidSymbolSelect() {
...
+          applied: claimed,
+        });
+      }
       if (!claimed && activeTradeInstrumentRef.current?.coin) {
         // The latch is process-wide, so this skip doubles as the only
         // remount-time resync: skipping unconditionally would strand the page
         // on the previous coin when the event bus message was dropped.
-        const bgSwitchParams = buildSwitchParamsFromTarget(instrumentTarget);
+        const bgSwitchParams = buildSwitchParamsFromTarget(instrumentTarget, {
+          deeplinkIntent: appliedDeeplinkIntent,
```
sidmorizon: # 🟡 P2 该分支里 deeplinkIntent 恒为 undefined

## 问题

第 988 行 `const appliedDeeplinkIntent = claimed ? deeplinkIntent : undefined;`，而这段代码位于 `if (!claimed && activeTradeInstrumentRef.current?.coin)` 内部，所以此处传入的 `deeplinkIntent` 在结构上永远是 `undefined`，`buildSwitchParamsFromTarget` 拿到它没有任何作用。

这与上面新加的注释「Only the claiming run can act on it」表达的意图一致，但代码写法却暗示 resync 分支也参与 deeplink 判定，读者会误以为 divergence 检查会把 deeplink 目标算进去。

## 影响

纯可读性/可维护性问题，当前不产生行为差异。风险在于后续如果有人放宽 `appliedDeeplinkIntent` 的赋值条件（例如改成按 TTL 判断而非按 `claimed`），这一行会立刻把 intent 重新灌进 divergence 检查，静默复活 review 中已确认的「30s 内 sticky override」缺陷，而且没有任何测试能拦住。

## 建议

直接去掉这个恒为空的参数，让 resync 路径显式地不参与 deeplink：

```ts
const bgSwitchParams = buildSwitchParamsFromTarget(instrumentTarget);
```

claiming 分支第 1100 行保留 `deeplinkIntent: appliedDeeplinkIntent` 即可；此时 `claimed` 恒为真，也可以直接写 `deeplinkIntent`，把「只有 claiming run 生效」这条不变量收敛到 988 行一处表达。

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.test.ts
```diff
@@ -142,6 +142,85 @@ describe('buildInitialTradeInstrumentSwitchParams', () => {
...
+        allowPerpFallback: true,
+      }),
+    ).toMatchObject({ mode: 'perp', coin: 'BTC' });
+  });
+
+  // What a non-claiming remount/refocus passes: the caller withholds the
+  // intent there because a mounted page already switched via the event bus.
+  // Resyncing off the background atoms is what keeps a notification from
+  // dragging the user back after they picked another market.
+  it('resyncs from the background atoms when the intent is withheld', () => {
```
sidmorizon: # 🟡 P2 新增用例没有覆盖本次真正修复的逻辑

## 问题

这条用例的注释把它描述为「非 claiming run 撤回 intent 后靠 background atoms 重新同步」的回归防线，也是 review 讨论里用来背书 `0b29bf7b35` 的证据。但它实际只是给纯函数传了 `deeplinkIntent: undefined` + `preferredInstrument: undefined`，等价于「没有 deeplink 的普通恢复」，与 208 行的 `still restores the rendered instrument when no deeplink is pending` 覆盖同一条分支。

真正修复缺陷的那行判断 `const appliedDeeplinkIntent = claimed ? deeplinkIntent : undefined;` 在 `PerpsGlobalEffects.tsx` 组件里，不在 `buildInitialTradeInstrumentSwitchParams` 内。把 `claimed ?` 这个门删掉，本文件 5 条新用例全部照常通过。

## 影响

本 PR 引入的两块有状态逻辑——`claimed` 门控与 `consumePendingInstrumentIntent` 的 `30s` TTL / consume-once 语义——测试覆盖率为 `0`。这两块恰好是 review 中被指出过一次的地方，缺少测试意味着下一次重构可以无声地回退，而 CI 依然全绿。

## 建议

两处补测：一是给 `ServiceHyperliquid.consumePendingInstrumentIntent` 加单测，覆盖「未过期返回值」「超过 `PENDING_INSTRUMENT_INTENT_TTL_MS` 返回 `undefined`」「连续两次调用第二次为 `undefined`」；二是把门控抽成可测的纯函数再在组件里调用，例如：

```ts
export function resolveAppliedDeeplinkIntent({ claimed, deeplinkIntent }) {
  return claimed ? deeplinkIntent : undefined;
}
```

然后补一条 `claimed: false` + 非空 intent 断言结果为 `undefined` 的用例，这样删掉门控时测试会红。

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

sidmorizon (commented):
Review

---

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
```diff
@@ -147,6 +147,15 @@ function BaseNotificationHandlerContainer() {
...
+            {
+              coin: perpToken,
+              mode: 'perp',
+            },
+          );
```
sidmorizon, line 157: **Cold-start symbol hijack still happens for tray, search, home and market shortcuts**

Severity: `non-severe`

The user's requested market is only remembered for taps that come through the notification path (`setPendingInstrumentIntent` at `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx:153`), while the other places that jump into Perps without a page mounted record nothing, so those taps still land on the previously used market.
Impact: Opening a specific market from the desktop tray, universal search, the Home perps card or the market banner can still show the wrong market on the first tap.

<details>
<summary>Why the other context-less callers keep the pre-existing behaviour</summary>

The defect fixed here is: a context-less caller writes the background `perpsActiveAssetAtom` and then emits `PerpSwitchActiveInstrument`; if the Perp page has not mounted yet, the transient `appEventBus.on` listener in `packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1527` does not exist, and the first `selectInitialSymbol` run passes `activeTradeInstrumentAtom` (cold-start cache = previous session's coin) as `preferredInstrument`, which now still outranks the background atoms unless a `deeplinkIntent` is present (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.ts:44-52`).

The same emit-before-mount pattern exists in `packages/kit/src/hooks/useTrayDataProvider.desktop.ts:1126`, `packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchPerpItem.tsx:83`, `packages/kit/src/views/Home/pages/PerpsContainer.tsx:182` and `packages/kit/src/views/Market/MarketDetailV2/components/PerpetualTradingBanner/PerpetualTradingBanner.tsx:71` — each navigates to the Perp tab and then emits, and none of them call `setPendingInstrumentIntent`. When the Perp tab is mounted for the first time in the process (lazy tab mount, or after a detach/remount), the emit is dropped and the claiming `selectInitialSymbol` run forces the stale cached instrument, exactly as described for the notification case. The PR description itself states the defect affects "every context-less entry point (home banner, push, desktop tray, universal search)".

</details>

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Fixed in 0b29bf7b35, plus a5b805cba8 which adds the coverage this needed.

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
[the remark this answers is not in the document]
Minnzen: Gated in 0b29bf7b35; a5b805cba8 moves the gate into `resolveAppliedDeeplinkIntent` with a test that fails when it is removed (verified by deleting it).

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Confirmed and fixed in a5b805cba8. The bound is 5 minutes now, which clears a biometric failure falling back to a password, and expiry emits `pending_instrument_intent_expired` with the age so a report of "did not switch" can be told apart from "never set".

Kept it a plain wall-clock bound rather than pausing it while locked: only the one claiming run per process can act on the intent, so the bound exists for a single case — the tap never reached Perps and the user opens it much later — and coupling the service to lock state to sharpen that felt like the wrong trade here.

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
Minnzen: Removed in a5b805cba8 — the resync branch now calls `buildSwitchParamsFromTarget(instrumentTarget)` with no intent argument at all, so it reads as what it is.

packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.test.ts
Minnzen: You are right, and it undercuts what I claimed on the earlier thread — that case only exercised the `undefined` branch of the pure function, so deleting `claimed ?` left everything green.

a5b805cba8 extracts the gate as `resolveAppliedDeeplinkIntent` and covers it. Verified by mutation: replacing the body with `return deeplinkIntent` fails `withholds it from every later run`, and restoring it passes 27/27.

Not adding a `consumePendingInstrumentIntent` unit test: constructing `ServiceHyperliquid` pulls the whole background graph, the same reason #12680 recorded no isolated harness for this path. The TTL now at least reports itself through `pending_instrument_intent_expired`.

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
Minnzen: Verified — the same emit-before-mount pattern is in `useTrayDataProvider.desktop.ts`, `UniversalSearchPerpItem.tsx`, `PerpsContainer.tsx`, `PerpetualTradingBanner.tsx` and `usePerpsNavigation.ts`, and none of them record an intent, so each still loses its first tap when the Perp tab mounts for the first time in the process.

Leaving it open rather than folding it in: this PR is a hotfix for a reported production defect on the notification path, and those five sites reach into Market, Home, Search and the desktop tray. The change is mechanical — one `setPendingInstrumentIntent` before each emit, mirroring the mode already in the payload — but it widens what QA has to re-cover, and the call is the author's, not mine to make unilaterally.

The PR description overstates the scope today; correcting that either way once this is decided.

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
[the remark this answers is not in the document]
Minnzen: Closing this out. The outcome you flagged — a leftover intent replayed on a later focus, overriding a market the user picked since — is fixed by the `claimed` gate in 0b29bf7b35: only one run per process can act on an intent, so once the page has mounted a leftover is inert, not merely short-lived.

Note a5b805cba8 raised the bound from 30s to 5 minutes, because app lock put the whole unlock flow between the tap and the first Perp frame. That makes a leftover live longer but does not reopen this: inertness is what closes it, not the duration.

Still happy to add the eager ack if you would rather it be cleared than inert — reopen and say so.

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
Minnzen: Fixed in ffbd79eb0b — all five now record the intent before the atom write, so the first Perp mount of a launch restores the tapped market instead of the cold-start cache:

- `useTrayDataProvider.desktop.ts`
- `UniversalSearch/.../UniversalSearchPerpItem.tsx`
- `Home/pages/PerpsContainer.tsx` — passes its own mode through, since it can open a spot pair
- `Market/.../PerpetualTradingBanner.tsx`
- `Market/hooks/usePerpsNavigation.ts`

Verified the ordering at each site (intent recorded before the emit) and ran the touched areas: 880 tests / 110 suites pass. The PR description claimed this scope already, so it is now accurate rather than aspirational.

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
```diff
@@ -20,6 +20,12 @@ export function usePerpsNavigation(source?: EPerpPageEnterSource) {
         try {
           const { default: backgroundApiProxy } =
             await import('@onekeyhq/kit/src/background/instance/backgroundApiProxy');
+          // Recorded before the atom write so a Perp tab mounting for the
+          // first time this launch restores this coin instead of the one its
+          // cold-start cache holds; the event below has no listener yet then.
+          await backgroundApiProxy.serviceHyperliquid.setPendingInstrumentIntent(
```
sidmorizon: # 🟡 P2 intent 记录在触发挂载的导航之后

## 问题

本 PR 要关闭的窗口是「导航触发 Perp 首次挂载 → claiming run 读 intent」。但本站点、`PerpetualTradingBanner.tsx:71` 和 `useTrayDataProvider.desktop.ts:1120` 都把 `setPendingInstrumentIntent` 放在 `navigation.switchTab(ETabRoutes.Perp)` / `switchTabAsync(ETabRoutes.Perp)` **之后**，本站点还额外隔了一次 `await import('.../backgroundApiProxy')`。作为对照，此前 review 认可的 `NotificationHandlerContainer/index.tsx:153` 是先记录再导航，本次新增的 `PerpsContainer.tsx:161` 也是先记录再 `switchTab`。

行内注释写的「Recorded before the atom write」成立，但决定成败的是它是否早于 `prepareInitialSymbolSelect`，而这里没有任何结构性保证，只是靠 React 把挂载 effect 排到后一个宏任务、以及 `import()` 命中模块缓存。

## 影响

当前实现大概率能跑通（后台调用在同一 tick 同步派发，挂载 effect 在 commit 之后才跑），但正确性依赖一条未写明的调度器假设。若导航路径上出现同步 flush，或 `await import` 在 Web/扩展上命中未加载的分包（数十到数百 `ms`），claiming run 会先于 intent 落库，该入口静默退回本 PR 要修的原始缺陷。且线上无从分辨——`pending_instrument_intent_expired` 只覆盖「过期」，不覆盖「写得太晚」。

## 建议

把这三处 `setPendingInstrumentIntent` 提到 `switchTab` / `switchTabAsync` 调用之前（本文件把 `await import` 一并提前），与 `NotificationHandlerContainer` 和 `PerpsContainer` 保持同一顺序，使「intent 先于挂载」成为结构保证而非时序巧合。

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

packages/kit/src/views/Home/pages/PerpsContainer.tsx
```diff
@@ -154,6 +154,14 @@ function useOpenPerpAsset() {
           return;
         }
         try {
+          // Recorded before the atom writes so a Perp tab mounting for the
+          // first time this launch restores this pair instead of the one its
+          // cold-start cache holds; the event below has no listener yet then.
+          if (coin) {
+            await backgroundApiProxy.serviceHyperliquid.setPendingInstrumentIntent(
```
sidmorizon: # 🟡 P2 尽力而为的 intent 写入挡在关键路径前

## 问题

`setPendingInstrumentIntent` 是纯内存的辅助记账，成败都不该改变用户这次点击的结果。但五个新站点都把它 `await` 在 `changeActiveAsset` 之前的同一个 `try` 里。本站点后果最重：`catch { return; }` 位于 `navigation.switchTab(ETabRoutes.Perp)` 之前，一旦这次后台调用 reject，点击 Home 持仓后既不切币也不跳转。

```ts
try {
  if (coin) {
    await backgroundApiProxy.serviceHyperliquid.setPendingInstrumentIntent(
      { coin, mode },
    );
  }
  if (coin && mode === 'perp') { ... }
} catch {
  return;
}
```

## 影响

这引入了一个此前不存在的失败点，且排在关键调用之前。扩展 MV3 后台 service worker 冷启中或桥接超时时 proxy 调用可能 reject，此时原本仍可完成的 `changeActiveAsset` 与 tab 跳转被一并跳过，用户看到的是点击完全无响应；`UniversalSearchPerpItem.tsx:83`、`PerpetualTradingBanner.tsx:71`、`usePerpsNavigation.ts:26` 三处则会连 `PerpSwitchActiveInstrument` 的 emit 一起吞掉，已挂载的 Perp 页也不会切币。

## 建议

让它不阻塞关键路径，同时保留「先于 `changeActiveAsset` 派发」的顺序（本文件已有 `noop`）：

```ts
void backgroundApiProxy.serviceHyperliquid
  .setPendingInstrumentIntent({ coin, mode })
  .catch(noop);
```

或为它单独包一层 `try/catch`，吞掉异常后继续执行原有逻辑。

[Claude session ↗](https://claude.ai/code/session_01MkpLWw9pkfEVxChbKRG6Pt)

---

sidmorizon (commented):
Review

---

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
Minnzen: Fixed in 100ae3a37a. Four sites had it after the navigation, not three — `UniversalSearchPerpItem.tsx` was in the same shape. All five now record before the `switchTab` / `switchTabAsync` that mounts the tab, and `usePerpsNavigation.ts` hoists its `await import()` above the navigation for the reason you gave: a cold chunk there could push the write past the claiming run.

Verified the ordering per file rather than by eye — intent line number precedes the navigation line in each of the five.

packages/kit/src/views/Home/pages/PerpsContainer.tsx
Minnzen: Fixed in 100ae3a37a. Each write now has its own `try`/`catch` and the critical path continues regardless.

The Home card was the one that mattered: its `catch { return; }` sits above `switchTab`, so a rejected proxy call left the tap doing nothing at all. Went with a dedicated `try` over `void ... .catch(noop)` so the ordering guarantee from the other thread still holds — the write is awaited before navigation, it just can no longer abort it.

952 tests / 118 suites pass across Perp, Market, Home, hooks and the Hyperliquid service.

originalix (commented):
@Minnzen

Reviewed commit `ffbd79eb0bd8`.

- **P2 · Non-claiming remount can discard an in-flight target**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
sidmorizon: **A failed bookkeeping call can stop a notification tap from opening the tapped market**

Severity: `non-severe`

The new best-effort recording of the tapped market is placed inside the same try block as the real work (`setPendingInstrumentIntent` at `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx:153`) instead of being isolated like at every other entry point, so a failure of that one call skips switching to the tapped market entirely.
Impact: Tapping a perps push notification can land the user on the trading tab still showing the previous market, with no price reconnect warm-up.

<details><summary>Why the shared try block turns a best-effort call into a hard abort</summary>

Every other caller added in this PR wraps the intent write in its own `try { ... } catch { /* ignore */ }` precisely because "a missing intent only costs the first-mount restore, so this must not be able to abort the tap" (see `packages/kit/src/views/Market/hooks/usePerpsNavigation.ts:26-32`, `packages/kit/src/views/UniversalSearch/components/SearchResultItems/UniversalSearchPerpItem.tsx:78-89`, `packages/kit/src/views/Home/pages/PerpsContainer.tsx:159-167`, `packages/kit/src/hooks/useTrayDataProvider.desktop.ts:1119-1126`).

In the notification handler the call is the first statement of the existing try block (`packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx:149-193`). A rejection of the background proxy call — plausible during a cold start when the background runtime is still coming up, which is exactly the scenario this PR targets — jumps to the catch at line 190 and skips `changeActiveAsset`, `tradingModeAtom.set('perp')`, the `PerpSwitchActiveInstrument` emit and the subscription prewarm. Before this PR the coin switch was the first awaited call, so it could not be pre-empted this way.

</details>

Suggested fix:

```suggestion
          try {
            await backgroundApiProxy.serviceHyperliquid.setPendingInstrumentIntent(
              {
                coin: perpToken,
                mode: 'perp',
              },
            );
          } catch {
            // ignore: a missing intent only costs the first-mount restore
          }
```

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -566,6 +581,68 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+  }): Promise<void> {
+    if (!params.coin) {
+      return;
+    }
+    this._pendingInstrumentIntent = {
+      coin: params.coin,
+      mode: params.mode,
+      createdAt: Date.now(),
+    };
+  }
```
sidmorizon, line 592: **Bug-fix retrospective note required by the repository rules is missing**

Severity: `non-severe`

The repository requires every bug fix to append a case entry to the retrospective case-studies file, and this fix (introducing `setPendingInstrumentIntent` at `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:585`) ships without that entry, so the required post-fix record is absent.
Impact: The team loses the mandated written record of this defect and its root cause.

<details><summary>Rule source</summary>

`.cursor/rules/self-testing.mdc` ("Bug Fix Case Recording") mandates appending a `## Case:` block with Date/Platforms/Symptom/Root Cause/Fix/Catchable-by to `.claude/skills/1k-retrospective/references/case-studies.md` after every bug fix. No commit in this PR touches that file.

</details>

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
Minnzen: Already fixed — this review ran against 100ae3a37a; 3a38d106fd isolates it. The recording now sits in its own `try`/`catch` above the block that does `changeActiveAsset`, the mode write, the event and the prewarm, so a rejected proxy call no longer skips any of them.

It was the one entry left in that shape after the previous round; the reasoning you quote from the other callers is exactly why.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Added in the commit above. Appended a `## Case:` block to `.skillshare/skills/1k-retrospective/references/case-studies.md` (the tracked path behind the `.claude/skills` symlink) with Date, Platforms, Symptom, Root Cause, Fix and Catchable-by, following the entries already there. Pure append, no existing case touched.

Recorded the transferable lesson rather than the symptom: a value restored from a cold-start cache is not a recent write, so a precedence rule between two copies of the same state has to say which side a fresh cold start makes authoritative — and a transient event bus cannot carry intent across a mount boundary.

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
```diff
@@ -16,10 +16,22 @@ export function usePerpsNavigation(source?: EPerpPageEnterSource) {
...
+        // must not be able to abort the tap. Recorded before the navigation
+        // that mounts the Perp tab, so the claiming initial-select cannot
+        // run ahead of it; the import above is hoisted for the same reason.
+        try {
+          await backgroundApiProxy.serviceHyperliquid.setPendingInitialTradeInstrument(
+            { coin, mode: 'perp' },
+          );
+        } catch {
+          // ignore
+        }
```
sidmorizon: **Tapping a market row can silently do nothing if a lazily loaded module fails to load**

Severity: `non-severe`

The lazily loaded background module is now awaited outside any error handling (`await import(...)` at `packages/kit/src/views/Market/hooks/usePerpsNavigation.ts:19-21`) before the tab switch, so a single load failure swallows the whole tap instead of just the market restore.
Impact: On web/extension, a failed module load makes tapping a market list row do nothing at all — no navigation to the Perps tab and no error shown — where previously the tab still opened.

<details>
<summary>Why moving the dynamic import ahead of the try block changes the failure mode</summary>

Before this PR the `await import('@onekeyhq/kit/src/background/instance/backgroundApiProxy')` lived inside the `try { ... } catch (error) { console.error(...) }` block at `packages/kit/src/views/Market/hooks/usePerpsNavigation.ts:34-44`, and `navigation.switchTab(ETabRoutes.Perp)` ran before it. A chunk/module load rejection (stale deployment, offline, CSP) was therefore caught and logged after navigation had already happened.

Now the import is the first `await` in the `setTimeout(async () => { ... })` callback and is not wrapped, so a rejection:
1. skips `navigation.switchTab(ETabRoutes.Perp)` (line 33), `setPendingInitialTradeInstrument`, `changeActiveAsset` and the event-bus emit entirely;
2. escapes as an unhandled promise rejection because the async callback's result is not observed.

The inline comment at lines 22-25 states the intent recording "must not be able to abort the tap", but the hoisted import can. Wrapping the import in its own try/catch (falling back to plain navigation when it fails) restores the previous guarantee.
</details>

Suggested fix:

```ts
        // A missing intent only costs the first-mount restore, so neither the
        // module load nor the record itself may abort the tap.
        try {
          const { default: backgroundApiProxy } = await import(
            '@onekeyhq/kit/src/background/instance/backgroundApiProxy'
          );
          await backgroundApiProxy.serviceHyperliquid.setPendingInitialTradeInstrument(
            { coin, mode: 'perp' },
          );
        } catch {
          // ignore
        }
```

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
```diff
@@ -146,6 +146,18 @@ function BaseNotificationHandlerContainer() {
...
+          // coin without it: on that path the Perp page restores its persisted
+          // instrument, which predates this tap and would win. Isolated because
+          // losing it only costs that restore — it must not skip the writes and
+          // the event below.
+          await backgroundApiProxy.serviceHyperliquid.setPendingInitialTradeInstrument(
+            { coin: perpToken, mode: 'perp' },
+          );
+        } catch {
+          // ignore
+        }
```
sidmorizon, line 160: **A market picked from a notification can hijack the Perps tab much later in the session**

Severity: `non-severe`

The tapped market is remembered in the background with no expiry and no link to whether the screen ever opened (`setPendingInitialTradeInstrument` at `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx:155-157`), so if the jump that follows fails the value is applied to whenever the Perps screen is first opened later.
Impact: After a notification that failed to open the Perps screen, the next time the user opens Perps in that session it shows the notification's market instead of the market they last used.

<details>
<summary>Recorded-before-navigation ordering plus the removed expiry</summary>

In `packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx` the pending market is recorded first (lines 149-160), then `changeActiveAsset` / `tradingModeAtom` writes run, and only afterwards `navigateToNotificationDetailByLocalParams` (line 187) actually navigates. If that navigation throws (the `catch` at line 206 shows the fallback dialog) the Perps tab never mounts, but the pending market stays in `ServiceHyperliquid._pendingInitialTradeInstrument` (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:558-592`).

The value is only cleared by `prepareInitialSymbolSelect` (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:612-630`), which runs on the first Perps mount — and that first mount is exactly the claiming run that *applies* it as `preferredInstrument` (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:984-997,1091-1097`). Since `buildInitialTradeInstrumentSwitchParams` returns `preferredInstrument` unconditionally (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.ts:43-57`), the stale target wins over the restore whenever the user later opens Perps in that same process.

The same shape exists for the desktop tray caller (`packages/kit/src/hooks/useTrayDataProvider.desktop.ts:1119-1126`) if `switchTabAsync` rejects. A bound such as the previously considered TTL, or clearing the pending value when the navigation that was supposed to mount the tab fails, would close the window.
</details>

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -566,6 +573,62 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+    const tradingUniverse = claimed
+      ? await this.getTradingUniverse()
+      : undefined;
+    return {
+      claimed,
+      pendingInitialTradeInstrument,
+      instrumentTarget,
+      tradingUniverse,
+    };
+  }
```
sidmorizon, line 630: **New cold-start market-selection logic ships without any unit or regression test**

Severity: `non-severe`

The new one-shot market store and its combined preparation step were added with no accompanying test file (`prepareInitialSymbolSelect` at `packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:599-630`), which the repository's mandatory self-testing checklist requires for new logic and for every bug fix.
Impact: The fixed behaviour has no automated guard, so a future change can silently reintroduce the wrong market on a cold start.

<details>
<summary>Rule and diff evidence</summary>

`.cursor/rules/self-testing.mdc` section 6 requires "New/modified logic has corresponding `.test.ts` file" and "Bug fix includes a regression test" (only skippable for pure UI styling with no logic). The PR diff touches only source files: no `.test.ts` change is present, even though the description claims "four regression cases" were added to `PerpsGlobalEffects.utils.test.ts`.

The claim-and-consume semantics of `_pendingInitialTradeInstrument` (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:558-592`) — no-op after the latch is taken, consumed on every run, only applied on a claiming run — are pure in-memory state transitions and can be covered without constructing the full service (e.g. by extracting the store, or by testing the preference selection now inlined at `packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:984-997`).
</details>

originalix (commented):
@Minnzen

Reviewed commit `4fbab96fbf91`.

- **P2 · Pending market intent can outlive the navigation that created it**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
Minnzen: Fixed in 77bf24f5a6. Hoisting the import for the ordering guarantee took it outside any handler, which is a regression I introduced — before this PR the tab opened first and a load failure only cost the switch.

The import now sits inside the same `try` as the record it serves, `navigation.switchTab` runs either way, and only the `changeActiveAsset` + emit that genuinely need the proxy are skipped when it never resolved. That restores the previous failure mode and keeps the ordering.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
Minnzen: Fair, and it is the rule rather than a preference — added in 77bf24f5a6.

The pick-versus-restore decision now lives in `resolveInitialPreferredInstrument` with four cases: a pick wins, a spot pick stays spot, no pick restores the persisted instrument, an empty pick is ignored. Verified by mutation rather than by eye — reversing the precedence fails two of them.

Took your second suggestion (test the selection) over extracting the store: the store is two field assignments whose "first mount only" property is structural, while the selection is the decision a future change could silently invert. `ServiceHyperliquid` still cannot be constructed under jest — importing it fails on the module graph, which I confirmed with a throwaway spec.

Also note `buildInitialTradeInstrumentSwitchParams` and its ten cases are back to their original state in this PR, so #12680 restore path is still covered by the tests written for it.

packages/kit/src/provider/Container/NotificationHandlerContainer/index.tsx
Minnzen: Real, and a deliberate trade rather than an oversight — flagging it here instead of adding another mechanism.

An earlier revision of this PR bounded the value with a wall-clock TTL. It was removed because it broke the primary case: with app lock on, the unlock flow sits between the tap and the first Perp frame, so a failed biometric falling back to a password discarded the target and the original defect came back — silently, since expiry and "never recorded" look the same from the restore path.

What is left is narrower in both likelihood and cost. It needs `navigateToNotificationDetailByLocalParams` to throw, which surfaces a fallback dialog rather than passing silently, and the outcome is that the next Perps open shows the market the user asked for a moment earlier instead of their previous one. Closing it means a clear path plus a background method to drive it, on the API surface this round just cut from four mechanisms to one.

Happy to add `clearPendingInitialTradeInstrument` and call it from that catch if you would rather have it explicit — it is small, I just did not think it earned the surface against a rare and mild outcome. Leaving the thread open for that call.

originalix (commented):
@Minnzen

Reviewed commit `77bf24f5a65c`.

- **P2 · Pending market intent can outlive the navigation that created it**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx
```diff
@@ -957,16 +958,37 @@ function useHyperliquidSymbolSelect() {
...
         activeCoin: activeTradeInstrumentRef.current?.coin,
       });
-      // Resolved once and reused below: two independent reads can straddle the
-      // background's own two-step write, and a divergence seen only in that gap
-      // would force a switch onto the current pair, wiping the order form.
-      const instrumentTarget = await resolveActiveInstrumentTarget();
+      const preferredInstrument = resolveInitialPreferredInstrument({
+        pendingInstrument: pendingInitialTradeInstrument,
+        restoredInstrument: activeTradeInstrumentRef.current,
+      });
```
sidmorizon: **A market the user picks by hand right after launch can be replaced by the market they originally tapped**

Severity: `non-severe`

The market tapped in a banner/push/search is applied as the preferred pick (`resolveInitialPreferredInstrument` at `packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1091`) after a possibly long metadata wait, so a market the user selects by hand during that wait gets silently replaced.
Impact: On the first Perps entry after launch, a user who switches markets while the page is still loading can be thrown back onto the deeplink's market, discarding the pair they just chose.

<details>
<summary>Pending instrument is captured before the trading-meta await but applied after it, overriding the live context instrument</summary>

On the claiming run, `prepareInitialSymbolSelect` returns `pendingInitialTradeInstrument` at the very start (`packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts:612-630`). When there is no cached trading universe the code then does `await refreshTradingMeta()` (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1072-1074`), a network round-trip measured at ~1.1s in the PR's own device trace. During that await the page is interactive: a token-selector pick runs `switchTradeInstrument` and updates `activeTradeInstrumentAtom`, so `activeTradeInstrumentRef.current` becomes the hand-picked pair.

Afterwards `buildSwitchParamsFromTarget(..., { force: true, preferredInstrument: resolveInitialPreferredInstrument({ pendingInstrument, restoredInstrument: activeTradeInstrumentRef.current }) })` runs, and `resolveInitialPreferredInstrument` (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.utils.ts:39-47`) unconditionally prefers `pendingInstrument` over the restored/live one. The forced switch takes a newer request id, so it supersedes the user's in-flight switch. This contradicts the comment added directly above the call (`packages/kit/src/views/Perp/components/PerpsGlobalEffects.tsx:1087-1089`), which states the resolution is deliberately late so "a pair the user picks during that wait has to be the one restored" — that property only holds for the restored value, not for the pending one.

A fix would be to prefer the pending value only while the context instrument still equals what it was when the pending value was captured (or to resolve the pending value into the preferred slot before the meta await and re-check the context instrument afterwards).

</details>

Minnzen: Confirmed and fixed in the commit above. Extracting the decision into a helper moved the read of `activeTradeInstrumentRef` earlier — before the awaited `refreshTradingMeta` on the no-cache branch — which is a timing change I did not intend to make.

`resolveInitialPreferredInstrument` is now called at the `buildSwitchParamsFromTarget` site, so the restored instrument is read exactly where it was before this PR. The pending market keeps its earlier capture, since it is taken once at the claim and cannot change underneath.

Your note about the log is what made this checkable: `initial_symbol_build_switch_params_end` reads the fresh ref, so the disagreement it would have produced is now gone too.

originalix (commented):
@Minnzen

Reviewed commit `fd41ebb064b9`.

- **P2 · Pending market intent can outlive the navigation that created it**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

originalix (commented):
@Minnzen

Reviewed commit `9450f31d7e52`.

- **P1 · Resolver widens mode and is not assignable to the instrument union**
- **P2 · Pending market intent can outlive the navigation that created it**
- **P2 · A newer market pick can be overwritten by the pending cold-start intent**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12918#issuecomment-5323417398)

originalix (approved):

PR #12919, opened 2026-08-18
fix: improve perps confirmation and deposit UX

Kahnchan:
## Summary

- hide the skip-confirm option when an aggressive limit-price warning forces order confirmation
- keep both first-deposit buttons visually stable without shared loading or disabled state
- preload the Unifold deposit menu and open it before the background account-status refresh completes
- preserve account-scoped stale-result guards and the existing enable-trading fallback for other states

## Testing

- `yarn jest packages/kit/src/views/Perp/hooks/useEnableTradingWithDepositFallback.test.ts packages/kit/src/views/Perp/utils/enableTradingDialogConfirm.test.ts packages/kit/src/views/Perp/utils/perpsOrderPanelEnableTrading.test.ts --runInBand`
- Web hot-reload compilation and `http://127.0.0.1:3000` HTTP 200

## Jira Issue

- https://onekeyhq.atlassian.net/browse/OK-60323

### DIFF
--- a/packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx
+++ b/packages/kit/src/views/Perp/components/TradingPanel/TradingButtonGroup.tsx
@@ -90,7 +90,10 @@ import {
   useRequestEnableTradingWithDepositFallback,
 } from '../../hooks/useEnableTradingWithDepositFallback';
 import { useLiquidationPrice } from '../../hooks/useLiquidationPrice';
-import { useShowDepositWithdrawModal } from '../../hooks/useShowDepositWithdrawModal';
+import {
+  usePreloadPerpsUnifoldDepositModals,
+  useShowDepositWithdrawModal,
+} from '../../hooks/useShowDepositWithdrawModal';
 import { useTradingCalculationsForSide } from '../../hooks/useTradingCalculationsForSide';
 import { useTradingPrice } from '../../hooks/useTradingPrice';
 import { PerpTestIDs } from '../../testIDs';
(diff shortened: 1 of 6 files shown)


originalix:

@Kahnchan

Reviewed commit `e395977173d0`. This report always reflects the latest reviewed changes.

### Review summary

The Perps first-deposit flow now trusts the current status atom to open the deposit menu immediately, then refreshes account status in the background, and both deposit entry points preload the Unifold modal chunk. The order confirmation UI also prevents the skip-confirmation preference from being changed while an aggressive limit-price warning is shown. The status and modal ownership boundaries remain in the existing foreground hook and background service paths, but removing the prior loading guard leaves repeated taps able to start multiple menu presentations.

**What needs attention:** Verify that first-deposit actions remain single-flight from the initial tap through modal presentation on both TradingPanel entry points.

### Issues to address

- **P2 · First-deposit buttons can open duplicate deposit dialogs**
  - **Impact:** The previous loading/disabled guard was removed from both first-deposit buttons, while showDepositWithdrawModal can resolve immediately after creating a dialog. A second tap can therefore start another request after the account-scoped promise has settled and present overlapping deposit menus.
  - **Suggested action:** Keep a per-account action lock until the deposit menu is dismissed or add an equivalent guard around handleFirstDepositPress in both TradingButtonGroup and LimitOrderForm.

### Validation gaps

The targeted Jest test could not be executed because the workspace lacks the Yarn node_modules state file. `git diff --check` completed without errors.

sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #12930, opened 2026-08-19
fix: stop perps order books overwriting each other's tick option (OK-59102, OK-59100)

Minnzen:
[OK-59102](https://onekeyhq.atlassian.net/browse/OK-59102) [OK-59100](https://onekeyhq.atlassian.net/browse/OK-59100)

----------

## Summary
- Persist the order book tick option only to **seed** a symbol that has none, instead of writing back a derived correction whenever it differs from what is stored.
- Stop a tick-option list derived without `szDecimals` from being pinned as the cached one, so the real reference list can supersede it.
- Drop the synchronous `flushNow()` from the tick option cache write path.
- Remove the temporary OK-59100 / OK-59102 field diagnostics.

## Intent & Context
QA reported two iOS bugs that always appeared and disappeared together: the order book flickering as a whole block (OK-59102), and the chart page scrolling without following the finger while the two order buttons stopped responding (OK-59100). Repro: Perps home → BTC → chart page → back → switch to another coin → chart page.

Earlier attempts on this branch treated them as separate rendering defects and did not hold up. This change comes from instrumenting the actual write path and catching the loop in a reproduction.

## Root Cause
Two `PerpOrderBook` instances are alive whenever the chart page is pushed over the Perps tab (tab root + chart page). This is by design and is not a leak — an instance census confirmed the count peaks at 2 with correct mount/unmount pairing.

Each instance derived its tick-option list from **its own** book snapshot and wrote the result back into the **shared** persisted atom whenever it differed (`shouldPersistOrderBookTickOption` had no convergence condition). The two lists disagreed, so they overwrote each other indefinitely. One captured reproduction: **850 tick-option writes and 839 global L2 resubscribes**, ending in `Maximum update depth exceeded`.

Why the lists diverged: `changeActiveAsset` publishes an optimistic instrument with `universe: undefined` and hydrates a cached book behind it, so an order book can render a real price while `szDecimals` is still missing. Without `szDecimals`, `buildReferenceTickOptions` bails and the fallback builder runs with a different multiplier set — labelling the same `0.1` tick as `nSigFigs: 5` instead of `null`. That list was cached, and the reference list could not displace it because the hysteresis guard kept the cached one on **equal** `priceDecimals`.

Why both symptoms: each resubscribe made the incoming book fail `isL2BookForTarget` on whichever instance disagreed, dropping it to cold cache — that is the whole-book flicker. The render churn saturated the UI thread that `HeaderScrollGestureWrapper` drives `scrollTo` on (its pan callbacks are worklets, so this is not JS-thread starvation), and the synchronous per-write MMKV flush starved the JS thread the order buttons' `onPress` runs on.

## Design Decisions
- **Seed-only, not "seed then repair".** A repair branch was considered and rejected: it fires exactly in the state where an instance's list cannot resolve the shared entry, and it does not even cover the more common drift case where `byValue` matches but the `nSigFigs` meaning changed. Seed-only is safe by construction — writes happen only on the `undefined → defined` transition — rather than safe by enumeration of today's list shapes. Explicit user selection keeps its own unconditional path.
- **Authoritative seed check in the action.** The hook's existence check reads render-time state, so two instances can both pass it. `setOrderBookTickOption` re-checks against the live atom via `get()`.
- **`flushNow()` removal.** It is the app-background path; forcing it per write made every write a synchronous full-store read, merge, and MMKV rewrite. `simpleDb` holds the durable copy, so this cache only needs to be eventually consistent.

## Changes Detail
- `OrderBook/useTickOptions.ts` — seed-only persist, gated on stored preferences having loaded and on a list built with `szDecimals`; fallback-derived lists tagged `isFallback` so a reference list supersedes them and they never overwrite one.
- `states/jotai/contexts/hyperliquid/actions.ts` — `source: 'seed'` resolved against the live atom; the persisted map is built from the merged view so an early seed cannot drop symbols the module cache still holds.
- `shared/src/utils/perpsOrderBookTickOptionsCache.ts` — debounced write instead of `flushNow()`.
- `OrderBook/index.tsx`, `PerpOrderBook.tsx`, `PerpsGlobalEffects.tsx`, `PerpMarketFooter.tsx`, `MobilePerpMarket.tsx`, `TradingViewPerpsV2.tsx`, `utils/perpsFieldDiagnostics.ts` — field diagnostics removed.

Also retained from earlier on this branch (independent defects found while investigating, not the root cause): horizontal order book depth bars and ladder rows reading one snapshot, mobile skipping `useRafCoalesced`, and the TradingView interaction-overlay chart-ready gating.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (iOS verified), and the shared Perps order book path on all platforms
- **Risk Areas**: tick-option selection is what the L2 subscription parameters are derived from. The behavioural change is that an order book no longer rewrites a stored tick option it disagrees with — whichever value is persisted first now wins until the user changes it explicitly.

## Test plan
Verified on iOS simulator (iPhone 16 Pro / iOS 18.6) against instrumented builds:
- [x] Repro path (BTC → chart → back → switch coin → chart): tick-option writes **850 → 0**, global resubscribes **839 → 3** (all genuine coin changes), no `Maximum update depth exceeded`.
- [x] Seed path exercised on coins with no stored entry: exactly one write per symbol, every write over an empty entry, and the seeded list came from the reference builder (`0.01:null`, not `0.01:5`).
- [x] The optimistic `universe: undefined` window is skipped rather than seeded from — confirmed in the log as `szDecimals` missing → no seed, then `szDecimals` present → single seed.
- [x] Order book instance count peaks at 2 with matched mount/unmount.
- [x] App boots clean after diagnostics removal.
- [ ] Real device pass (bundle self-test + QA).

[OK-59102]: https://onekeyhq.atlassian.net/browse/OK-59102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-59100]: https://onekeyhq.atlassian.net/browse/OK-59100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx
```diff
@@ -610,6 +665,10 @@ function MobilePerpMarket() {
                       scrollEnabled={!isTradingViewInteractionOverlayOpen}
                       showsVerticalScrollIndicator={false}
                       contentContainerStyle={{ flexGrow: 0, minHeight: 0 }}
+                      onScrollBeginDrag={handleOrderbookScrollBeginDrag}
+                      onScroll={handleOrderbookScroll}
+                      scrollEventThrottle={16}
+                      onContentSizeChange={handleOrderbookContentSizeChange}
```
sidmorizon: **Scroll-position logging on the chart tab never records anything, so the exported log looks like the page never scrolled**

Severity: `non-severe`

The scroll listener attached to the tab scroller (`onScroll={handleOrderbookScroll}` at `packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx:669`) is discarded by the scroller itself, so the log lines that are meant to show whether the page moved are never written even when it scrolls normally.

Impact: The diagnostics gathered from a user's device will always be missing the scroll-offset lines, which can be misread as proof that the page is frozen and lead to the wrong root cause.

<details>
<summary>Library spreads caller props before installing its own scroll handler</summary>

`Tabs.ScrollView` resolves to `react-native-collapsible-tab-view`'s `ScrollView`, which renders `<ScrollViewMemo {...rest} ... onScroll={scrollHandler} onContentSizeChange={scrollContentSizeChangeHandlers} scrollEventThrottle={...} />` — visible in the repo's own patch context at `patches/react-native-collapsible-tab-view+8.0.1.patch:951-967`. Because `{...rest}` is spread *before* the explicit `onScroll` and `scrollEventThrottle`, the caller-supplied `onScroll` and `scrollEventThrottle={16}` (`packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx:669-670`) are silently overridden by the library's Reanimated handler and never invoked. `onContentSizeChange` *is* explicitly destructured and chained by the library (`useChainCallback`), so that probe works, and `onScrollBeginDrag` stays in `rest` and works too — only the offset probe, which is the signal the ticket investigation relies on ("offsetY pinned at 0"), is dead.

A working alternative is to read the tab scroll offset through the library's own API (e.g. `useCurrentTabScrollY` / `useFocusedTab`, already re-exported from `packages/components/src/composite/Tabs/index.native.tsx`) instead of an `onScroll` prop.

</details>

packages/kit/src/views/Perp/components/OrderBook/index.tsx
```diff
@@ -868,8 +877,18 @@ export function OrderBook({
...
-  const askPercents = useRafCoalesced(askPercentsRaw, depthEpoch);
+  const bidLadderCoalesced = useRafCoalesced(bidLadderRaw, depthEpoch);
+  const askLadderCoalesced = useRafCoalesced(askLadderRaw, depthEpoch);
+  // Mobile already renders a 200ms-throttled visual snapshot
+  // (`enableVisualSnapshot` in PerpOrderBook), so coalescing to the frame here
+  // can never merge two updates — it only costs a frame of latency and an extra
+  // render each time, which is what made the book read as slower when the rows
+  // were first bundled in. Desktop keeps it: its L2 feed has no such throttle,
+  // which is the case REACT-NATIVE-1JZ was written for.
+  const bidLadder = isMobileVariant ? bidLadderRaw : bidLadderCoalesced;
```
sidmorizon: **Order book flicker fix does not cover the desktop/tablet layout, where bars and numbers still come from different snapshots**

Severity: `non-severe`

The depth bars in the vertical order book are still fed from a delayed snapshot (`reversedAskPercents`/`bidPercents` at `packages/kit/src/views/Perp/components/OrderBook/index.tsx:880-895`) while the price/size rows drawn over them read the live data, so that layout keeps the exact bar/row mismatch this change set out to remove.

Impact: On desktop and split-view the order book can still show a depth block that doesn't match the numbers on top of it, i.e. the same flicker QA reported.

<details>
<summary>Only the horizontal branch was migrated to the bundled snapshot</summary>

The horizontal branch now renders rows from `bidLadder.levels` / `askLadder.levels` (`packages/kit/src/views/Perp/components/OrderBook/index.tsx:1053`, `:1094`), which are guaranteed to be the same snapshot as `bidPercents`/`askPercents`.

The vertical branch (`horizontal === false`, used by the `web` variant from `packages/kit/src/views/Perp/components/PerpOrderBook.tsx:1087-1104`) still renders `verticalAsks`/`verticalBids` computed directly from `aggregatedData` (`packages/kit/src/views/Perp/components/OrderBook/index.tsx:769-776`), while its bars use `reversedAskPercents` (frame-coalesced, `:892-895`) and `bidPercents` (coalesced because `isMobileVariant` is false for the `web` variant, `:888-891`). So on the platform that keeps the coalescer, bars and rows can still be one frame apart — and `reversedAskPercents` was not bundled into the new ladder object at all.

</details>

packages/kit/src/views/Perp/components/OrderBook/index.tsx
```diff
@@ -868,8 +877,18 @@ export function OrderBook({
         .map((item) => calculatePercentage(item.cumSize, askDepth)),
     [aggregatedData.asks, askDepth],
   );
-  const bidPercents = useRafCoalesced(bidPercentsRaw, depthEpoch);
-  const askPercents = useRafCoalesced(askPercentsRaw, depthEpoch);
+  const bidLadderCoalesced = useRafCoalesced(bidLadderRaw, depthEpoch);
+  const askLadderCoalesced = useRafCoalesced(askLadderRaw, depthEpoch);
```
sidmorizon: # 🟡 P2 移动端 bypass 并没有省掉那次额外渲染，coalescer 仍在空转

## 问题

`useRafCoalesced` 内部持有 `useState`，并在 rAF 回调里执行 `setEmitted(latestRef.current)`（`OrderBook/useRafCoalesced.ts:86-91`）。第 `880`/`881` 行这两个 hook 在移动端仍然被**无条件调用**，只是第 `888`/`889` 行把结果丢掉了。也就是说每次可视快照更新（`PERPS_ORDER_BOOK_MOBILE_VISUAL_FRAME_MS = 200`，约 `5` 次/秒），这两个 hook 依然各自注册一次 `requestAnimationFrame` 并提交一次 state，让 `OrderBook` 整棵树多渲染一次，而这次渲染产出的 `bidLadderCoalesced` / `askLadderCoalesced` 没有任何消费者。

紧挨着的注释写的是 bypass 可以省掉 `a frame of latency and an extra render each time`；实际省掉的只有那一帧延迟，`an extra render` 依旧存在。

## 影响

移动端 `horizontal` 分支每次数据更新仍然渲染 `2` 次而不是 `1` 次。该分支要重建 `13` 买 + `13` 卖共 `26` 个带内联 `onPress` / `style` 闭包的 `Pressable`（`index.tsx:1053-1133`），等于在正在排查卡顿的这块屏幕上，把 order book 的渲染量维持在两倍，同时每 `200ms` 还挂两个纯空转的 rAF 回调。

同理，`reversedAskPercents`（第 `892` 行）在 `horizontal` 分支里也没有任何消费者，属于同一类空转。

## 建议

把开关下沉到 hook 内部，让移动端直接短路返回，而不是"照常算完再丢弃"：

```ts
export function useRafCoalesced<T>(
  value: T,
  flushKey?: unknown,
  enabled = true,
): T {
  // enabled === false 时直接 return value，不注册 rAF、不 setEmitted
}

const bidLadder = useRafCoalesced(bidLadderRaw, depthEpoch, !isMobileVariant);
const askLadder = useRafCoalesced(askLadderRaw, depthEpoch, !isMobileVariant);
```

顺带把注释里 `an extra render each time` 的说法与实际行为对齐，避免后续读者误判该分支的开销。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

originalix:

@Minnzen

Reviewed commit `eb59d04da3f1`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the foreground Perps rendering and gesture boundary: the mobile horizontal order book now feeds depth bars and rows from the same visual snapshot, while the TradingView bridge rejects pre-readiness overlay-open events and resets overlay state when readiness is lost. It also adds release-build diagnostics around the main-runtime iOS scroll, footer, and order-book paths through the shared native file logger. The change stays within kit UI ownership and does not alter trading, signing, or account-state contracts, but the scroll handler currently bypasses its intended logging throttle.

**What needs attention:** Confirm that release-build iOS diagnostics cannot perturb the scroll gesture being investigated and that the overlay readiness guard still disables parent scrolling whenever a real chart overlay is active.

### Issues to address

- **P2 · Scroll diagnostics bypass their own throttle**
  - **Impact:** During a downward drag, each new offset exceeds maxOffsetY, so the combined condition permits a diagnostic write even when fewer than 500 ms have elapsed. With scrollEventThrottle set to 16, the main JS runtime can send records through the shared native file logger at approximately display-frame rate, potentially adding jank and distorting the release-device scroll investigation while flooding exported logs.
  - **Suggested action:** Make the 500 ms time gate independent of offset growth, update maxOffsetY separately, and emit no more than one scroll record per interval.
- **P3 · Mobile still schedules discarded RAF updates**
  - **Impact:** The two useRafCoalesced hooks run before the mobile branch selects the raw ladders. Every new mobile ladder therefore still schedules setEmitted for the next frame and causes a state-driven render even though neither coalesced result is displayed, retaining part of the overhead this change says it removes.
  - **Suggested action:** Allow useRafCoalesced to be disabled for mobile so it bypasses both frame scheduling and state updates while desktop retains coalescing.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` completed without whitespace errors.
- The targeted Jest command could not start because the workspace lacks Yarn's node_modules state file; no test result is claimed.
- Existing GitHub review and inline-comment coverage could not be fetched from the restricted network, so proposed comments require a duplicate check.
- No release-device interaction was performed here. Element or log existence alone does not count as passing.
- Final pass condition: on an iOS release build, a sustained downward drag should remain smooth and produce at most two scroll records per second; a real chart overlay should disable parent scrolling only while active; reload or render-process loss should restore scrolling; and rapid L2 updates should keep horizontal bars aligned with their displayed rows.

Minnzen:
Both findings addressed in `303cb33c15`.

**P2 — scroll diagnostics bypass their own throttle.** Confirmed and fixed. The guard combined the 500ms gate with a peak check, and during a downward drag every offset beats the previous peak, so it never short-circuited and wrote a record per scroll event. The time gate is now independent and the peak is tracked separately, so at most one record per interval is emitted.

Also lowered `scrollEventThrottle` from 16 to 100, which speaks to the "confirm diagnostics cannot perturb the scroll gesture" note: the sample only has to answer whether the offset moves at all, so there is no reason to take a JS callback every frame while measuring the gesture under investigation.

**P3 — mobile still schedules discarded RAF updates.** Confirmed and fixed. `useRafCoalesced` now takes an `enabled` flag; mobile passes `!isMobileVariant` so neither the frame nor the state update happens, instead of computing a coalesced ladder and throwing it away. Desktop is unchanged. The call site also reads better now — one line states the intent rather than deriving two values and discarding one.

**On the overlay readiness guard disabling parent scrolling for real overlays.** It does. The gate is `if (isOpen && !isChartContentReady) return`, and a user can only open a chart menu after the chart is live, so a genuine overlay-open passes through and disables scrolling as before. What the gate drops is only `open` arriving before readiness, which cannot correspond to a real menu — and today such an event can never be closed, because nothing resets the flag. `false` is always accepted, and any transition out of ready forces it closed.

**On the validation gaps.** Agreed that nothing here is device-verified — the two bugs do not reproduce on the simulator, which is what led an earlier attempt to be closed as "not reproducible". The diagnostics in this PR exist precisely to close that gap from a QA device, and they are to be removed before this merges.

packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.ts
```diff
@@ -37,19 +37,33 @@ const RAF_COALESCE_ENABLED = true;
...
+
   // Resolve the value this render returns. Default to whatever was last emitted.
   let resolved = emitted;
 
-  if (!RAF_COALESCE_ENABLED) {
+  if (!isActive) {
+    // Switched off: hand the value straight back without touching state, so no
+    // frame is scheduled and no render is queued for a result nobody reads.
+    // Keep the flush key current so re-enabling never sees a stale one.
+    prevFlushKeyRef.current = flushKey;
```
sidmorizon, line 80: # 🟡 P2 `enabled` 由 `false` 翻回 `true` 时会先吐出一帧挂载时的陈旧快照

## 问题

关闭期间（`isActive === false`）走第 `62-67` 行分支，只更新 `prevFlushKeyRef.current`；而 `emitted` 自 `useState<T>(value)` 的初值起再也没有被写过 —— 第 `82` 行的渲染期同步和第 `86-107` 行的 rAF 都被 `isActive` 短路了。

于是 `enabled` 翻回 `true` 的那一次渲染：

- `prevFlushKeyRef.current === flushKey`（上一帧刚同步过），第 `68` 行的 flush 分支不触发；
- `resolved = emitted`，即**挂载那一刻**的那份快照；
- `resolved !== emitted` 为 `false`，第 `82` 行的渲染期修正也不触发。

这一帧直接把陈旧值交回调用方，要等下一个 rAF（第 `101` 行）才纠正。第 `65` 行注释 `Keep the flush key current so re-enabling never sees a stale one` 的因果恰好是反的：正因为 `flushKey` 被同步了，唯一能同步兜底的 flush 分支才被跳过。

## 影响

`OrderBook` 传的是 `enabled = !isMobileVariant`，而 `variant` 会在**同一个 fiber** 上翻转：`PerpOrderBook.tsx:1086-1107` 的 `{gtMd ? <OrderBook variant="web"/> : mobileOrderBook}` 两个分支是同位置、同类型、无 `key` 的 `<OrderBook>`，且 `<PerpOrderBook entry="perpMobileMarket" />`（`MobilePerpMarket.tsx:561`）本身不随 `gtMd` 卸载。

所以在移动版行情页数据已加载的稳态下，拉宽窗口 / iPad 旋转或分屏跨过 `gtMd` 断点，就会命中 `disabled → enabled`。此时 `depthEpoch` 不变（`useOrderBookEpoch` 只跟 `coin` / `tick` / `isEmpty` 相关，`index.tsx:126-138`），flush 无法兜底，`bidPercents` / `askPercents` 会用**开页时那一份**深度数据画一帧柱状条，而行文字读的是实时 `aggregatedData` —— 正是本 PR 想消掉的那类条/行错配。

## 建议

记住上一次的 active 状态，在重新启用的第一次渲染强制同步发射：

```ts
const wasActiveRef = useRef(isActive);
// ...
} else if (!wasActiveRef.current || prevFlushKeyRef.current !== flushKey) {
  prevFlushKeyRef.current = flushKey;
  if (frameRef.current !== null) {
    cancelAnimationFrame(frameRef.current);
    frameRef.current = null;
  }
  resolved = value; // 交给第 82 行的 setEmitted 同步落地，不留陈旧帧
}
wasActiveRef.current = isActive;
```

顺带把第 `65` 行注释的因果表述一并修正。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.ts
```diff
@@ -37,19 +37,33 @@ const RAF_COALESCE_ENABLED = true;
...
+ * `enabled` (optional) turns the whole mechanism off. Callers already fed by a
+ * throttled upstream have nothing to coalesce, and leaving it on would still
+ * schedule a frame and a state update per change for a value they discard.
  */
-export function useRafCoalesced<T>(value: T, flushKey?: unknown): T {
+export function useRafCoalesced<T>(
+  value: T,
+  flushKey?: unknown,
+  enabled = true,
+): T {
```
sidmorizon, line 49: **Changed order book snapshot logic ships without any accompanying test**

Severity: `non-severe`

The order-book snapshot bundling and the new opt-out of the frame coalescer change real logic (`enabled = true` at `packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.ts:48`) with no new or updated test file, which the repository's mandatory self-testing checklist requires for modified logic and bug fixes.
Impact: A future change can silently re-introduce the order book flicker or the extra render without any check failing.

<details>

<summary>Rule reference and affected logic</summary>

`.cursor/rules/self-testing.mdc` section 6 (never skipped for non-styling logic) requires: "New/modified logic has corresponding `.test.ts` file" and "Bug fix includes a regression test". This PR changes the signature and branching of `useRafCoalesced` (`packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.ts:45-107`) and the snapshot pairing in `packages/kit/src/views/Perp/components/OrderBook/index.tsx:853-891`, but adds no test; the OrderBook folder only contains the pre-existing `OrderBook.empty.test.tsx`, `tickSizeUtils.test.ts` and `utils.test.ts`.

A focused unit test for `useRafCoalesced` (disabled pass-through, no state update / no scheduled frame when disabled, flush on `flushKey` change) would cover both halves.

</details>

originalix (commented):
@Minnzen

Reviewed commit `303cb33c157b`.

- **P2 · Re-enabling RAF coalescing can replay a stale ladder snapshot**
- **P3 · Scroll diagnostics keep the previous session's max offset**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12930#issuecomment-5337315913)

packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx
```diff
@@ -295,9 +308,84 @@ function MobilePerpMarket() {
...
+        (contentInset?.bottom ?? 0) -
+        layoutMeasurement.height;
+      perpsFieldDiagnostics('iosOrderbookTab.dragBegin', {
+        scrollEnabled: !isTradingViewInteractionOverlayOpen,
+        offsetY: Math.round(contentOffset.y),
+        contentHeight: Math.round(contentSize.height),
+        viewportHeight: Math.round(layoutMeasurement.height),
+        insetTop: Math.round(contentInset?.top ?? 0),
+        scrollableRange: Math.round(scrollableRange),
+        maxOffsetY: Math.round(orderbookScrollDiagRef.current.maxOffsetY),
```
sidmorizon: # 🟡 P2 `dragBegin` 上报的 `maxOffsetY` 结构上恒为 `0`，会给出"滚动器从未移动"的伪证

## 问题

`orderbookScrollDiagRef.current.maxOffsetY` 在本文件里只有**一处写入** —— 第 `362` 行的 `handleOrderbookScroll`。而该回调挂在 `onScroll` 上，这个 prop 既被 `Tabs.ScrollView` 的类型显式 `Omit`，运行时也被库自己的 handler 覆盖（已有另一条 review 线在跟进，这里不重复）。所以该字段从初始值 `0` 起，永远不会被写第二次。

但第 `347` 行的读取方 `handleOrderbookScrollBeginDrag` 挂的是 `onScrollBeginDrag` —— 它留在库的 `{...rest}` 里、没有被显式 prop 覆盖，是**真的会触发**的。于是产生了一条确实会落盘的记录，里面带着一个恒为 `0` 的 `maxOffsetY`。

## 影响

导出的日志里 `iosOrderbookTab.dragBegin` 会稳定长成这样：`{ scrollEnabled: true, offsetY: 0, scrollableRange: 1100, maxOffsetY: 0 }` —— 有可滚动区间、用户确实按下并拖动了、峰值却是 `0`。这恰好拼成"滚动器完全没动"的伪证，而这正是 OK-59100 要判定的那个结论。

这跟"scroll 行缺失"不是一回事：缺失的行只会让人觉得信息不全；而这里是一个**存在且看起来像实测值**的字段在说假话，误导代价高得多。

## 建议

在 `onScroll` 那条链路真正可用之前，不要在 `dragBegin` 里上报 `maxOffsetY`。该事件自身已带 `offsetY` 与 `scrollableRange`，足以回答"有没有可滚动空间"：

```ts
perpsFieldDiagnostics('iosOrderbookTab.dragBegin', {
  scrollEnabled: !isTradingViewInteractionOverlayOpen,
  offsetY: Math.round(contentOffset.y),
  contentHeight: Math.round(contentSize.height),
  viewportHeight: Math.round(layoutMeasurement.height),
  insetTop: Math.round(contentInset?.top ?? 0),
  scrollableRange: Math.round(scrollableRange),
  // maxOffsetY 移除：唯一的写入方 onScroll 不会执行
});
```

等偏移探针改成一条真正会跑的链路之后，再把峰值字段加回来。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

Minnzen:
All review findings are addressed as of `9acc0e0dd9`. Status per comment:

**First, on the diagnostics generally.** Everything touching `perpsFieldDiagnostics` is temporary. It exists because neither of these bugs reproduces on a simulator — an earlier attempt was closed as "not reproducible" on exactly that mistake, and QA has since reproduced both on a device. These records are how the remaining unknown (OK-59100) gets settled from a QA build. **They are removed before this merges**; the fix commits stand on their own.

That said, two of the findings against the diagnostics were real defects rather than cosmetic, and both are fixed:

- **`maxOffsetY` structurally always 0 → false evidence of "the scroller never moved"** (`MobilePerpMarket.tsx:347`) and **scroll-position logging never records anything** (`MobilePerpMarket.tsx:700`). Both trace to the same root: `Tabs.ScrollView` omits `onScroll` from its props type and react-native-collapsible-tab-view binds its own handler *after* spreading ours, so the callback never ran — and `scrollEventThrottle` and `onMomentumScrollEnd` are dropped the same way. This was worse than missing instrumentation: `dragBegin` does fire, and it was reporting a peak that could only ever be 0, which is precisely the false "never moved" conclusion OK-59100 is trying to establish. Replaced with a `dragBegin`/`dragEnd` pair, both of which survive in the spread; `dragEnd` reports how far the drag actually moved. `onContentSizeChange` was verified safe — the library chains it via `useChainCallback`.

- **Re-enabling RAF coalescing replays a stale snapshot** (`useRafCoalesced.ts:80`) — fixed: re-enabling now seeds from the live value, and disabling cancels any pending frame.

- **Mobile bypass still spins the coalescer** (`OrderBook/index.tsx:881`) — fixed: `useRafCoalesced` takes an `enabled` flag and mobile passes `!isMobileVariant`, so neither the frame nor the state update happens, instead of computing and discarding.

- **No accompanying test** (`useRafCoalesced.ts:49`) — added `useRafCoalesced.test.ts`: deferral, flush-key bypass, disabled pass-through, and the re-enable path.

**Not changed, deliberately: desktop/tablet vertical bars and numbers still come from different snapshots** (`OrderBook/index.tsx:888`). Confirmed pre-existing rather than introduced here. It stays within one frame there because no native easing stretches it — the iOS flicker came from the depth bar's `CADisplayLink` interpolation, which the desktop path does not have. Fixing it means touching the hover-range and reversed-asks logic for a defect nobody has reported, so it is left for a follow-up rather than widening a change that is already carrying diagnostics.

One related fix that *was* worth pulling in: the mobile vertical book resolved row taps against the live arrays while rendering the coalesced snapshot, so a tap could put a price into the order form that was never on the row the user pressed. That is a correctness issue rather than a visual one, and only `price` is consumed downstream, so it was a contained change.

packages/kit/src/views/Perp/components/OrderBook/index.tsx
sidmorizon: 这条只修复了一半，故保持 unresolved。

`bidLadder` / `askLadder`（`:888`、`:889`）确实已经通过 `!isMobileVariant` 关掉了，这部分没问题。但本条评论同时点名的 `reversedAskPercents` 至今未改，在 `9acc0e0dd9` 上仍是：

```ts
const reversedAskPercents = useRafCoalesced(
  reversedAskPercentsRaw,
  depthEpoch,
);
```

没有第三个参数，`enabled` 取默认值 `true`。而它唯一的消费者是 `:1192` 的 `percents={reversedAskPercents}`，位于 vertical 分支；在移动端 `horizontal` 分支上这个值无人读取，却照旧每次可视快照更新都注册一次 `requestAnimationFrame` 并 `setEmitted`，仍然多触发一次整棵 `OrderBook` 的渲染。

补一个参数即可，与上面两行保持一致：

```ts
const reversedAskPercents = useRafCoalesced(
  reversedAskPercentsRaw,
  depthEpoch,
  !isMobileVariant,
);
```

（该文件自 `eb59d04` 起这一段逐字节未变，已连续五个 head 未触及。）

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

packages/kit/src/views/Perp/components/OrderBook/index.tsx
```diff
@@ -1938,21 +1957,27 @@ export function OrderBookMobile({
...
   const bidLadderRaw = useMemo(
     () => ({
       percents: bidPercentsRaw,
       prices: bidPricesRaw,
       sizes: bidSizesRaw,
+      levels: aggregatedData.bids,
     }),
-    [bidPercentsRaw, bidPricesRaw, bidSizesRaw],
+    [aggregatedData.bids, bidPercentsRaw, bidPricesRaw, bidSizesRaw],
   );
```
sidmorizon, line 1981: **Mobile trade-tab order book keeps the extra frame of delay the change was meant to remove**

Severity: `non-severe`

The mobile vertical order book still routes its rows and bars through the frame delay (`useRafCoalesced(askLadderRaw, depthEpoch)` at `packages/kit/src/views/Perp/components/OrderBook/index.tsx:1979-1981`) even though its data source is already slowed to one update every 200ms, so it pays a delay that can never merge anything.
Impact: The order book on the mobile trade tab still updates a frame later than necessary — the same perceived slowness this change removed from the other mobile order book.

<details>
<summary>enableVisualSnapshot is on for every mobile surface, but only OrderBook (horizontal) got the enabled=false flag</summary>

`packages/kit/src/views/Perp/components/PerpOrderBook.tsx:544` sets `const enableVisualSnapshot = !gtMd;` — i.e. the 200ms visual throttle applies to *all* mobile rendering, including the `OrderBookMobile` path used for `variant="mobileVertical"` (`PerpOrderBook.tsx:994-1026`).

The PR's stated rationale ("Putting a 16ms frame coalescer downstream of a 200ms throttle can never merge two updates — it only adds a frame of latency and one extra render per update") therefore applies equally to `OrderBookMobile`, but the new `enabled` argument was only added in the `OrderBook` component (`index.tsx:888-889`). `OrderBookMobile` still calls `useRafCoalesced(askLadderRaw, depthEpoch)` and `useRafCoalesced(bidLadderRaw, depthEpoch)` with coalescing on.

This is an incomplete application of the transformation, not a data-consistency issue: `OrderBookMobile` already bundles percents/prices/sizes/levels, so its bars and rows stay mutually consistent.

</details>

originalix:

@Minnzen

Reviewed commit `9acc0e0dd93d`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the main-runtime Perps UI contract in two places: it tightens TradingView overlay readiness handling on the mobile market page, and it reworks order-book rendering so bars, ladder rows, and taps share one visual snapshot. It also adds a production diagnostics path through `NativeLogger` so release-device scroll and gesture failures can be exported from the field.

The order-book snapshot work stays within existing `@onekeyhq/kit` boundaries and improves visual consistency without changing background ownership, signing, or privileged trading flows. The main regression risk is that the new overlay gate still reuses the older `chartContentReady` state, which is marked ready on the earlier `READY` bridge event and therefore still admits pre-live overlay opens. The added diagnostics are also broader than the stated iOS-only troubleshooting target because shared Perps surfaces call a helper that logs on non-native platforms instead of stopping early.

**What needs attention:** Verify the TradingView readiness state machine first: the overlay-open guard must be driven only by true chart-content readiness, not by the earlier line-ready event.

### Issues to address

- **P1 · Overlay gating still trusts the pre-ready READY event**
  - **Impact:** The new guard is supposed to ignore `tradingview_interactionOverlay(open)` until the chart is actually ready, but `onChartLinesReady()` still marks `chartContentReadyWebviewKey` as ready on the earlier `READY` message. If the chart emits an overlay-open between `READY` and `CHART_READY`, the app will accept it during the same pre-live window this fix is trying to close, and the iOS page can end up with scrolling disabled again before the chart is truly interactive.
  - **Suggested action:** Separate line-ready from content-ready for overlay gating, or drive the guard from a dedicated ref/state that flips only in `onChartReady()`.
- **P2 · Temporary field diagnostics leak onto non-native Perps surfaces**
  - **Impact:** `perpsFieldDiagnostics()` falls back to `console.log` outside native, while this PR now calls it from shared `PerpOrderBook` and `TradingViewPerpsV2` paths that also run on desktop and web. That broadens a temporary iOS release-debugging path into unrelated browser/Electron sessions and makes the log stream noisier and less intentional.
  - **Suggested action:** Make the helper a no-op outside the intended native/iOS surface, or guard the shared call sites before invoking it.

### Validation gaps

- `git diff --check origin/release/v6.5.0...HEAD` completed without whitespace errors.
- Running `yarn jest packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.test.ts --runInBand` failed in this workspace because Yarn could not find the node_modules state file.
- No iOS release-build runtime validation was available here for the original scroll-lock and flicker scenarios.
- Existing PR comment coverage was not fetched in this environment, so inline-comment duplication could not be fully checked.

packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx
```diff
@@ -59,6 +61,9 @@ import {
   type IMobilePerpMarketTab,
   getMobilePerpMarketPageScrollState,
 } from '../utils/mobilePerpMarketScrollState';
+import { useTabsContext } from 'react-native-collapsible-tab-view';
```
sidmorizon: # 🟡 P2 共享文件里直接 value-import `react-native-collapsible-tab-view`，绕开了仓库既有的平台切分

## 问题

第 `64` 行的 `import { useTabsContext } from 'react-native-collapsible-tab-view'` 是一个**运行时**（value）导入，而 `MobilePerpMarket.tsx` 是不带 `.native` 后缀的共享文件。

仓库对这个库有一套一致执行的平台切分：全仓 20 处引用中，非 `.native.*` 文件一律只做 `import type`（`Tabs/context.ts:4-11`、`useCurrentTabScrollY.ts:3`、`Tabs/index.tsx:7`，以及 `packages/kit` 内的 `EarnMainTabs.tsx`、`MobileInformationTabs.tsx` 等），真正的 value 导入只出现在 `.native.*` 文件里（`Tabs/index.native.tsx:3`、`useFocusedTab.native.ts:1`、`CollapsibleTabContext.native.ts:1`）。本次新增是全仓**唯一**的例外。

这条切分不只是风格：`react-native-collapsible-tab-view` 仅在 `apps/mobile/package.json:143` 声明，web / desktop / ext 三个 app 都没有声明它。

## 影响

`MobilePerpMarket` 是注册在 `perpRouters` / `ModalPerpStack` 里的普通路由，三端的模块图都会经过它。由于是 value 导入，web/desktop/ext 打包时必须解析并纳入这个 native-only 库的运行时（连带 `react-native-pager-view` 等传递依赖）。

`nodeLinker: node-modules` 的提升让它大概率仍能解析成功，所以未必立刻红，但这三端会平白打进一份**既没声明、运行时也永远用不到**的原生库 —— 探针本身只在 `platformEnv.isNativeIOS` 分支下挂载。

## 建议

仓库已经为这件事提供了切分好的一等 API：`CollapsibleTabContext`，`.native.ts` 转发库的真 `Context`、`.ts` 转发 web 自己的 `TabsContext`，两端 index 均已导出（`Tabs/index.tsx:33`、`Tabs/index.native.tsx:80`）。`HeaderScrollGestureWrapper.native.tsx:41-45` 正是这个用法。

本文件第 `14-25` 行已从 `@onekeyhq/components` 引入 `HeaderScrollGestureWrapper` / `Tabs`，补一个名字即可：

```ts
import {
  CollapsibleTabContext,
  HeaderScrollGestureWrapper,
  Tabs,
  // ...
} from '@onekeyhq/components';

function OrderbookScrollMetricsProbe() {
  const tabsContext = useContext(CollapsibleTabContext);
  const headerHeight = tabsContext?.headerHeight;
  const containerHeight = tabsContext?.containerHeight;
  // 顺带获得 HeaderScrollGestureWrapper 里那种 `tabsContext?.` 空值保护，
  // 当前直接解构 useTabsContext() 在 Provider 缺失时会直接抛。
}
```

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

sidmorizon: 更正上面一处措辞，结论不变。

我写的"非 `.native.*` 文件一律只做 `import type`"不准确：`packages/components/src/composite/Tabs/TabsDraggableFlatList.tsx:5-13` 就是一个 value 导入，同样取了 `useTabsContext`，文件名也没有 `.native` 后缀。

但它不构成反例——该文件的导入方只有 `.native` 入口：

```
packages/components/src/composite/Tabs/index.native.tsx:6
packages/components/src/layouts/SortableListView/index.native.tsx:21
```

所以它永远进不了 web / desktop / ext 的模块图。

真正的判据不是文件名后缀，而是**能否从非 native 入口到达**。按这个判据，本条的问题依然成立，而且更清楚：`MobilePerpMarket.tsx` 经 `perpRouters` / `ModalPerpStack` 注册，三端模块图都会经过它，是目前唯一一个「可从非 native 入口到达、却 value 导入该库」的位置。

另外补一句：`lint (24.x)` 对这一行没有报错是预期的，不能作为本条已解决的依据。`tsc` 通过 hoist 后的 `node_modules` 解析，既不校验各 app 的依赖声明，也不关心最终打包内容；这条讲的是依赖声明与产物组成，不是类型。

建议不变：改用仓库已切分好的 `CollapsibleTabContext`。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

originalix (commented):
@Minnzen

Reviewed commit `856d40d210e4`.

- **P1 · Overlay gating still trusts the pre-ready READY event**
- **P2 · Temporary field diagnostics leak onto non-native Perps surfaces**
- **P3 · Tabs wrapper drops the order-book content-size diagnostic**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12930#issuecomment-5339505694)

packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx
Minnzen: Fixed. The root cause was the one you identified: `Tabs.ScrollView` omits `onScroll` from its props type and the library rebinds its own handler after spreading ours, so the callback never ran — same for `scrollEventThrottle` and `onMomentumScrollEnd`.

It was worse than silent, too: `maxOffsetY` was only written from that dead callback while `dragBegin` (which does fire) reported it, so every record would have read "scrollable range available, user dragged, peak 0" — precisely the false "never moved" conclusion OK-59100 is trying to establish.

Replaced with a `dragBegin`/`dragEnd` pair, both of which survive in the spread; `dragEnd` reports how far the drag actually moved. `onContentSizeChange` was verified safe (the library chains it via `useChainCallback`).

packages/kit/src/views/Perp/components/OrderBook/index.tsx
Minnzen: 另一半也补上了。`reversedAskPercents` 现在同样传 `!isMobileVariant` —— 它只服务 vertical 分支，而 vertical 在 mobile variant 下不渲染，所以在移动端属于纯空转。

`bidLadder` / `askLadder` 那部分按你的建议把开关下沉到了 hook 内部（`useRafCoalesced` 新增 `enabled` 参数），disabled 时不注册 rAF、不 setEmitted，而不是算完再丢弃。

packages/kit/src/views/Perp/components/OrderBook/useRafCoalesced.ts
Minnzen: Added `useRafCoalesced.test.ts` — four cases: the frame deferral, the flush-key synchronous bypass, the disabled pass-through (asserting no frame is scheduled), and the re-enable path returning the live value rather than a snapshot from before it was switched off.

packages/kit/src/views/Perp/components/OrderBook/index.tsx
Minnzen: Fixed. `OrderBookMobile` now passes `enabled: false`, for the same reason the horizontal book does: `enableVisualSnapshot = !gtMd` throttles every mobile surface to one update per 200ms, so a 16ms coalescer downstream can merge nothing and only carries the latency.

You were right that this was an incomplete application rather than a consistency problem — that component already bundles percents/prices/sizes/levels.

packages/kit/src/views/Perp/components/OrderBook/index.tsx
Minnzen: Confirmed pre-existing rather than introduced here, and left as-is deliberately.

The iOS flicker comes from the native depth bar's `CADisplayLink` interpolation stretching the one-frame skew into something visible. The desktop/tablet vertical path has no native easing, so the same skew stays within a frame there and has not been reported. Fixing it means touching the hover-range and reversed-asks logic, which is a wider change than the defect warrants — especially on a branch already carrying temporary diagnostics.

Tracking it as a follow-up rather than widening this PR.

packages/kit/src/views/Perp/pages/MobilePerpMarket.tsx
Minnzen: 已改。探针改为通过 `CollapsibleTabContext`（从 `@onekeyhq/components` 导入）读取，与 `HeaderScrollGestureWrapper.native.tsx` 的用法一致，`react-native-collapsible-tab-view` 的 value 导入已移除。

你说的影响是对的：这个探针只在 `platformEnv.isNativeIOS` 分支下挂载，却会让 web / desktop / ext 三端的模块图都解析进一个只在 `apps/mobile` 声明的原生库。仓库既然已经为此提供了切分好的一等 API，没有理由绕开。

packages/kit/src/views/Perp/components/OrderBook/index.tsx
```diff
@@ -704,7 +705,7 @@ export function OrderBook({
     undefined,
   );
   const [containerHeight, setContainerHeight] = useState(() =>
-    horizontal ? 0 : (initialContainerHeight ?? 0),
+    horizontal ? 0 : initialContainerHeight ?? 0,
```
sidmorizon: # 🟡 P2 顺手去掉的括号会触发 `prettier/prettier` 报错，且当前被 tsc 失败掩盖

## 问题

第 `708` 行把 `horizontal ? 0 : (initialContainerHeight ?? 0)` 改成了 `horizontal ? 0 : initialContainerHeight ?? 0`。

语义是等价的（三元的 else 分支本就会一直向右吃到底），但格式不符合本仓库 pin 的 prettier `3.8.1` 输出。两条证据：

- 这一行在 base `51ace8a8` 上就是**带括号**的（同文件 `:707`）；
- 仓库里同类写法一律带括号 —— `TokenIconView.tsx:87` 的 `: (networkId ?? '')`、`TxActionTransfer.tsx:339`、`useAccountData.ts:108`、`ProtocolLendingActionDialogContent.tsx:1084` 等。

而 `.eslintrc.js:74` 是 `'prettier/prettier': ['error', require('./.prettierrc.js')]` —— 这是 `error` 级规则，不是偏好。这一行本身跟本 PR 的目标无关，属于顺带改动。

## 影响

`lint:project` 的定义是 `yarn _tsc && yarn _lint && ...`（`package.json:92`），跑 eslint 的 `_lint` 被 `_tsc` 用 `&&` 短路。

当前 `_tsc` 仍在 `useRafCoalesced.test.ts` 上失败，所以这条 prettier 报错**现在根本看不到**；等那两行 `renderHook` 修好、tsc 放行之后，它会作为下一个 CI 失败冒出来，白白多一轮往返。

## 建议

还原括号即可：

```ts
const [containerHeight, setContainerHeight] = useState(() =>
  horizontal ? 0 : (initialContainerHeight ?? 0),
);
```

或直接跑 `yarn lint:all`（`package.json:94`，带 `--fix`）让 prettier 自己写回。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

originalix (commented):
@Minnzen

Reviewed commit `34eee1e05edc`.

- **P2 · Vertical order book still mixes coalesced bars with live rows**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12930#issuecomment-5339505694)

packages/kit/src/views/Perp/components/OrderBook/index.tsx
sidmorizon: **Reformatted conditional expression breaks the repository's formatting check**

Reported at `packages/kit/src/views/Perp/components/OrderBook/index.tsx:708` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

A conditional expression was rewritten without the parentheses the project's formatter requires around a nullish fallback (`horizontal ? 0 : initialContainerHeight ?? 0` at `packages/kit/src/views/Perp/components/OrderBook/index.tsx:708`), so the automated code-style check fails on this file.
Impact: The lint/format gate that must pass before commit fails, blocking CI on an unrelated cosmetic edit.

<details>
<summary>Prettier 3 requires parentheses for `??` inside a ternary</summary>

`package.json` pins `prettier@3.8.1`, and Prettier 3 formats a nullish-coalescing expression inside a conditional branch with explicit parentheses. Every other occurrence in the repo follows that shape, e.g. `packages/kit/src/components/TokenListView/TokenIconView.tsx:87` (`: (networkId ?? '')`) and `packages/kit/src/views/Swap/pages/modal/SwapKLineContent.tsx:130`. This line is now the only one without them, so `eslint-plugin-prettier` / `yarn lint:staged` reports a formatting error. CLAUDE.md explicitly forbids committing code that fails linting; the edit is also unrelated to either ticket in this PR.

</details>

Suggested fix:

```ts
    horizontal ? 0 : (initialContainerHeight ?? 0),
```

sidmorizon:
@cursoragent review

packages/kit/src/views/Perp/components/OrderBook/index.tsx
Minnzen: Outdated. `packages/kit/src/views/Perp/components/OrderBook/index.tsx` now passes `prettier --check` (and the repo's `lint-worktree-ts` / `lint-staged` / `tsc-staged` checks), so the formatting concern no longer applies to the current head.

The branch has also been rewritten since this comment: the field diagnostics it was reviewing are removed, and the change is now the OK-59102/OK-59100 root-cause fix (seed-only tick-option persist). Resolving.

Minnzen: Outdated. `packages/kit/src/views/Perp/components/OrderBook/index.tsx` now passes `prettier --check` (and the repo's `lint-worktree-ts` / `lint-staged` / `tsc-staged` checks), so the formatting concern no longer applies to the current head.

The branch has also been rewritten since this comment: the field diagnostics it was reviewing are removed, and the change is now the OK-59102/OK-59100 root-cause fix (seed-only tick-option persist). Resolving.

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
```diff
@@ -211,29 +233,36 @@ export function useTickOptions({
...
     };
 
-    if (
-      shouldPersistOrderBookTickOption({
-        isReady: Boolean(tickOptionsData),
-        persisted,
-        next: currentPersist,
-      })
-    ) {
+    if (!persisted) {
```
sidmorizon: # 🟡 P2 `shouldPersistOrderBookTickOption` 已成死代码，却仍导出且测试全绿

## 问题

改成 seed-only 之后，第 `254` 行的 `if (!persisted)` 取代了原来的 `shouldPersistOrderBookTickOption(...)` 判断，同时它也从本文件的 import 里被移除了。

它在生产代码里已经没有任何调用方。全仓搜索（`packages/` 与 `apps/`）只剩两处命中：

- 定义本身 `OrderBook/tickSizeUtils.ts:61-79`
- 测试 `OrderBook/tickSizeUtils.test.ts:152-196`

但它仍然是 `export function`，那 4 个用例也照常通过。

## 影响

这个函数编码的恰恰是本次提交要废弃的那条策略 —— **只要派生值与已存值不同就写回**。而这正是 OK-59102 里两个 order book 互相覆盖的成因：两份 instance-local 列表对同一个 `value` 给出不同的 `nSigFigs`，于是双方不停把对方写回的值再覆盖掉，永不收敛。

死代码和全绿测试在 CI 上都是绿的，从外部看它就是一个「在用、且有测试保障」的 API。后续有人把它接回写路径，就会把这个刚修好的 bug 原样带回来。

## 建议

连同它的测试一起删掉：

- `tickSizeUtils.ts:61-79` 的函数定义
- `tickSizeUtils.test.ts:152-196` 的 `describe('shouldPersistOrderBookTickOption', ...)`

如果打算留作日后复用，至少在函数上方注明它已被 seed-only 策略取代、不可直接接回写路径，避免后来者误用。

[Claude session ↗](https://claude.ai/code/session_01XFuiNxZjTJ9LPvpgVreuKv)

---

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
```diff
@@ -211,29 +233,36 @@ export function useTickOptions({
...
   useEffect(() => {
     if (!symbol) return;
-
-    const persisted = persistedTickOptions[symbol];
+    if (!tickOptionsData) return;
+    if (!hasLoadedPersistedTickOptions) return;
+    // Only derive a tick option from a list built with szDecimals: without it
+    // buildReferenceTickOptions bails and the fallback builder labels the same
+    // tick with a different nSigFigs, which is what the two books disagreed on.
+    if (!Number.isInteger(szDecimals)) return;
```
sidmorizon: **Order book can still lock in a wrong price-grouping default during a rare timing window**

Severity: `non-severe`

The order book's default price grouping is saved as authoritative (`setOrderBookTickOption({ source: 'seed' })` at `packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts:255`) from a temporarily-wrong grouping whenever live prices briefly vanish exactly as the coin's decimals become known, so the wrong grouping is what sticks until the user changes it by hand.
Impact: In an uncommon timing window the order book groups prices with the wrong precision and keeps doing so for the rest of the session.

<details>
<summary>How the fallback-derived list slips past the szDecimals gate</summary>

The seed effect only writes when `Number.isInteger(szDecimals)` (`useTickOptions.ts:241`), on the assumption that an integer `szDecimals` guarantees `tickOptionsData` is a reference-derived list rather than the fallback list (which mislabels e.g. `0.1` as `nSigFigs: 5` instead of `null` — the exact OK-59102 divergence).

However `tickOptionsData` can be a *cached fallback* even while `szDecimals` is an integer: if a fallback list was cached earlier (market price non-zero while `szDecimals` was still `undefined`, `useTickOptions.ts:167-176`), and then on the render where `szDecimals` becomes an integer the market price momentarily reads `'0'` (bids/asks empty), the memo takes the `marketPrice === '0'` transition branch and returns the cached fallback via `getTickOptionsDataDuringTransition` (`useTickOptions.ts:116-123`) without replacing the cache. `tickOptionsData.isFallback` is then still `true`.

At that point the seed effect passes all gates and seeds `selectedTickOption`, which for a symbol with no stored preference resolves to the fallback list's `defaultTickOption` — the mislabeled grouping. Because seeding is "first write wins", that wrong value persists until an explicit user change.

Gating the seed additionally on `!tickOptionsData.isFallback` (the field the PR already introduced for the cache-displacement logic) would close this window.

</details>

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
Minnzen: Agreed, and fixed in `cc25b7f557`.

You are right that leaving it exported with a green test suite is the actual hazard here — it reads as a supported API, and the rule it encodes (write the derived value back whenever it differs from what is stored) is precisely what let the two order books overwrite each other without ever converging. Wiring it back into the write path would reintroduce OK-59102 verbatim.

Removed rather than annotated:
- `tickSizeUtils.ts` — the function and its `IOrderBookTickOptionState` state type, which had no other user. `INSig` / `IMantissa` are kept; they are used elsewhere in the file.
- `tickSizeUtils.test.ts` — the `describe('shouldPersistOrderBookTickOption', ...)` block and the import.

Verified no remaining references across `packages/` and `apps/`. Perp suite is 269 passing (down 6, exactly the deleted cases), and `agent:check --profile commit` is green.

originalix (commented):
@Minnzen

Reviewed commit `cc25b7f557c9`.

- **P2 · Vertical order book still mixes coalesced bars with live rows**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12930#issuecomment-5339505694)

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
Minnzen: Confirmed and fixed in `5992fb1837`. Thanks — this is a real hole, and the diagnosis is exactly right.

`Number.isInteger(szDecimals)` was standing in for "this list came from `buildReferenceTickOptions`", and those are not equivalent. Once a fallback list is cached, a render where the market price momentarily reads `'0'` returns it through `getTickOptionsDataDuringTransition`, which consults neither `szDecimals` nor the cache-replacement rule. Since seeding is first-write-wins, the mislabeled grouping would then stick for the session.

Now the property is tested directly rather than inferred:

```ts
isFallbackList: Boolean(tickOptionsData?.isFallback),
```

I also moved the whole seed decision into `shouldSeedOrderBookTickOption` in `tickSizeUtils.ts`, so the invariants it rests on are pinned by tests instead of by a comment — never from a fallback list, never before stored options finish loading, never over an entry another order book already established. Five cases added in `tickSizeUtils.test.ts`.

Note on the residual: during that same window the *display* still derives from the fallback list, so the grouping label can be briefly wrong. That one self-corrects — as soon as a non-zero market price arrives, the reference branch runs and the `!cached.isFallback` rule replaces the cached list. Only the seed was permanent, and that is what this closes.

Perp suite 274 passing; `agent:check --profile commit` green.

originalix (commented):
@Minnzen

Reviewed commit `5992fb1837b6`.

- **P2 · Vertical order book still mixes coalesced bars with live rows**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12930#issuecomment-5339505694)

Minnzen:
Confirmed and fixed in `7db6c6f86b`.

The finding is accurate. The horizontal and mobile ladders already carry `levels` alongside `percents` precisely so bars and rows come from one frame, but the desktop vertical variant was still building its rows straight from live `aggregatedData` while its bars read the coalesced arrays — so it kept exactly the skew the rest of the PR removes, and the tap handler resolved against the live level rather than the one on screen.

Changes:
- `reversedAskLadderRaw` now carries `levels` next to `percents`, matching the other ladders, and is coalesced as one object.
- `verticalAsks` / `verticalBids` are taken from `reversedAskLadder.levels` / `bidLadder.levels` instead of `aggregatedData`. The empty-state placeholders are unchanged.
- The ask row's `originalIndex` is derived from `verticalAsks.length` rather than the live array, so it cannot point past the snapshot it was taken from.

Both sides were already coalesced with `enabled: !isMobileVariant`, and `depthEpoch` folds in `isEmpty`, so an empty/non-empty flip still flushes synchronously rather than showing a stale frame.

On the validation gaps you listed: `useRafCoalesced.test.ts` and `tickSizeUtils.test.ts` do run here — full Perp suite is 36 files / 274 tests passing, and `agent:check --profile commit` is green. There is still no render-level test asserting bar/row snapshot alignment; that would need a mounted-component test this PR does not otherwise have, so I have left it out rather than bolt one on.

Minnzen:
Superseded by #12957, which re-applies this fix onto `hotfix/v6.5.2`.

This branch was developed on `release/v6.5.0` and carries 240 commits of unrelated upstream work relative to 6.5.2, so retargeting the base here would have produced an unreviewable diff. The net fix is 9 files / +360 −129 against 6.5.2 and moved cleanly: the `OrderBook/` files and the tick-option cache were untouched on that base and were taken directly, while `actions.ts` and `TradingViewPerpsV2.tsx` had their own 6.5.2 changes, so only this fix's delta was three-way applied on top of them.

All review findings from this PR are carried over in the code. Closing in favour of #12957.

PR #12907, opened 2026-08-17
fix: align perp dex source labels(OK-59905)

Kahnchan:
[OK-59905](https://onekeyhq.atlassian.net/browse/OK-59905)

----------

## Issue

- [OK-59905](https://onekeyhq.atlassian.net/browse/OK-59905)

## Summary

- Preserve and display HIP-3 DEX source labels (`xyz` / `para`) in Wallet Home market favorites and perpetual categories.
- Reuse the existing perps display mapping so favorite and category rows render consistent source badges.
- Align the Native perps ticker badge line heights with the shared DEX badge.

## Intent & Context

[OK-59905](https://onekeyhq.atlassian.net/browse/OK-59905) reports that Wallet Home market favorites and perpetual lists do not show the `xyz` / `para` source labels. During Native verification, the DEX badge in the perps ticker was also visibly shorter than its adjacent badges.

## Root Cause

The Wallet Home perps mapper kept the display symbol but dropped the DEX prefix from the server token name. The Native ticker used implicit React Native text line heights for its local badges while the shared `PerpDexBadge` uses a fixed 16px line height.

## Design Decisions

- Parse the source through the existing `parseDexCoin` utility and render it with the existing `PerpDexBadge` component.
- Reuse the same perps display mapper for favorites and category data to keep both paths consistent.
- Normalize only the local Native ticker badge line heights instead of changing the shared badge sizing across Market and Home surfaces.

## Changes Detail

- Add `perpsDexLabel` to the Wallet Home market display model and forward it to the shared token identity row.
- Preserve `xyz` and `para` labels when mapping server perps tokens, with coverage for main-DEX tokens that should remain unlabeled.
- Set the Native perps ticker labels to a consistent 16px line height.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Mobile, Desktop, Web, Extension
- **Risk Areas**: Wallet Home perps row rendering and Native perps ticker badge layout.

## Test plan

- [x] Run focused Jest coverage for Wallet Home and Market perps token mapping (2 suites, 7 tests).
- [x] Run `yarn agent:check --profile commit`.
- [x] Run `yarn agent:check --profile pr`.
- [ ] Verify Wallet Home favorites and perpetual categories show `xyz` / `para` badges on Native.
- [ ] Verify the Native perps ticker badges have matching heights.

[...]

PR #12906, opened 2026-08-17
fix: Earn banner glass bar, shadow clipping and list copy (OK-60299, OK-60300, OK-60303)

ezailWang:
[OK-60299](https://onekeyhq.atlassian.net/browse/OK-60299) [OK-60300](https://onekeyhq.atlassian.net/browse/OK-60300) [OK-60303](https://onekeyhq.atlassian.net/browse/OK-60303)

----------

[OK-60299]: https://onekeyhq.atlassian.net/browse/OK-60299?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60300]: https://onekeyhq.atlassian.net/browse/OK-60300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60303]: https://onekeyhq.atlassian.net/browse/OK-60303?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12935, opened 2026-08-19
chore: sync release/v6.5.0 fixes after bundle #7 into hotfix/v6.5.2

originalix:
Syncs the fixes that landed on `release/v6.5.0` after bundle #7 into `hotfix/v6.5.2`.

## Included PRs (6)

| Commit | PR | Description |
|---|---|---|
| `7beb35ffa7` | #12872 | inject jpush key into native bundle builds |
| `880119c9e6` | #12901 | sync stock token metadata for max amount (OK-58721) |
| `f688b9537a` | #12907 | align perp dex source labels (OK-59905) |
| `9bd95eca97` | #12918 | keep the perps deeplink target over the cold-start restore (OK-60543) |
| `dff362df01` | #12887 | sync sidebar theme switching |
| `4763435128` | #12906 | Earn banner glass bar, shadow clipping and list copy (OK-60299/60300/60303) |

**Deliberately excluded:** `1a75a3823f` (#12880, `chore: release #7`). It only touches `RELEASES.json` and is release/v6.5.0 bundle bookkeeping. This branch has no `RELEASES.json` at all, so picking it would fabricate a manifest whose single `seq: 7` entry points at `fd071a7c71` — a commit that is not on this branch.

## Conflicts and resolution

- **First 5 commits: zero conflicts.** The only auto-merge was `.skillshare/skills/1k-retrospective/references/case-studies.md`, where the 3-way merge kept both branches' entries.
- **#12906 conflicted in 20 generated i18n files** (`translations.ts` plus 19 locale JSONs). The conflicts are adjacent-insertion artifacts — the `HEAD` side of every hunk is empty — not both branches editing the same key.
  - Before resolving, I verified that `release/v6.5.0`'s key set is a **strict superset** of this branch's: 0 keys unique to `hotfix/v6.5.2`, 92 keys added by release. Taking release's version of these generated artifacts wholesale is therefore lossless and equivalent to an `i18n:pull` result — no generated file was hand-edited, per the repo rule.
  - Post-resolution check: all 20 files are byte-identical to `51ace8a822`.

## Verification

- **File parity:** of the 68 unique files touched by the 6 commits, **67 are byte-identical to `release/v6.5.0`**. The single exception is the retrospective doc noted above.
- **Type check:** full `tsgo --noEmit` passes with exit 0 and zero errors. Compared in the same worktree across the 5-commit and 6-commit states — identical.
- **Tests:** 16 suites / 179 tests pass, covering PopularTrading, PerpsGlobalEffects, swapStockChannel, Earn and Discovery.
- **Missing-ancestor risk:** the branches diverged at `5cd0dbcf0c`, so these fixes originally sat on top of 23 commits this branch lacks. Verified that those 23 are absent **by SHA only — their content is already here**, folded in by #12839 and #12903. Both true semantic ancestors (#12680, #12792) are present.
- **Runtime i18n:** the translation keys these commits consume (`dexmarket.spot`, `global.max`, `perp.0_fee`) exist on this branch in both the enum and `en_US.json` — a missing JSON key would be a runtime-only failure that no type check catches.

## Before merging

1. **Confirm the `JPUSH_KEY` repository secret exists.** #12872 adds a hard gate (`exit 1` when empty) that also governs the `workflow_run` and `workflow_dispatch` entry points, which bypass the patched `secrets:` blocks. This cannot be verified from git.
2. **Re-test the perps deeplink fix on a native split-bundle build.** This branch runs `ENABLE_NATIVE_BACKGROUND_THREAD=true` locally where release does not. OK-60543's state lives on the bg-JS `ServiceHyperliquid` singleton while all recording call sites are main-JS, and this branch also added an awaited hardware migration to bg `initCritical`, making the cold-start window the fix optimizes busier.
3. **Spot-check chart theme switching on desktop and native** against the live `tradingview.onekey.so` deployment — #12887 removes the remount fallback on those platforms in favour of an injected `changeTheme` call.

Note: `LOCAL_DB_VERSION` is 20 on this branch vs 19 on release. This sync does not touch it, but it means v6.5.2 carries a one-way schema bump relative to v6.5.0.


sidmorizon:
@codex review

sidmorizon:
@codex security review

wabicai (approved):

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -497,6 +504,18 @@ export function useSwapStockAmountInputState({
...
+    : resolveStockExecutionTokenMetadata({
+        token: inputToken,
+        tokenDetail: stockInputTokenBalance.authoritativeTokenDetail,
+      });
+  const amountInputToken = authoritativeStockInputToken ?? inputToken;
+  const balanceActionsReady = isStockBalanceActionReady({
+    authoritativeBalance: stockInputTokenBalance.authoritativeBalance,
+    authoritativeStockToken: authoritativeStockInputToken,
+    isBuySide,
+  });
```
sidmorizon, line 518: **Max and percentage buttons stop working on the sell side when token details come back empty**

Severity: `non-severe`

The balance shortcuts are treated as unusable (`isStockBalanceActionReady` at `packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:514-518`) whenever the extra token record is missing, even though a balance is already shown, so sell orders lose their Max and percentage buttons.
Impact: A user selling a stock token can see their balance but cannot tap Max or 25/50/75%, and must type the amount by hand.

<details>
<summary>How a defined balance can coexist with a missing token detail</summary>

`useStockInputTokenBalance` builds its result at `packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts:194-204`: `balance` falls back to `seededBalance ?? '0'` while `tokenDetail` is `markStockUsdPriceCurrency(details?.[0])`, which is `undefined` when `fetchSwapTokenDetails` returns an empty array. `balanceReady` (`:257`) only checks `balance !== undefined`, so `authoritativeBalance` is defined while `authoritativeTokenDetail` stays `undefined`.

On the sell side `authoritativeStockInputToken` is `resolveStockExecutionTokenMetadata({ token, tokenDetail: undefined })`, which returns `undefined` by its first guard (`packages/kit/src/views/Swap/hooks/swapStockChannelUtils.ts:142-151`). `isStockBalanceActionReady` then returns false because `isBuySide` is false and `authoritativeStockToken` is missing.

That single flag now gates `onBalanceMaxPress`, `onSelectPercentageStage`, `enableMaxAmount`, `showPercentageInput`, the iOS `InputAccessoryView`, and `SwapActionsState.onSelectPercentageStage` in `packages/kit/src/views/Swap/pages/components/SwapStockDesktopContainer.tsx`, so all of them stay disabled for the whole session while the balance itself renders normally.

</details>

packages/shared/src/locale/enum/translations.ts
```diff
@@ -156,6 +156,7 @@ export enum ETranslations {
   address_type_tooltip_native_segwit__desc = 'address_type_tooltip_native_segwit__desc',
   address_type_tooltip_nested_segwit__desc = 'address_type_tooltip_nested_segwit__desc',
   address_type_tooltip_taproot__desc = 'address_type_tooltip_taproot__desc',
+  alerts_and_updates__action = 'alerts_and_updates__action',
```
sidmorizon, line 159: **Generated translation files are edited, which the repository rules forbid**

Severity: `non-severe`

Auto-generated translation artifacts are changed in this PR (new and edited keys in `packages/shared/src/locale/enum/translations.ts:159` and the locale JSON files), which the repository rules explicitly prohibit.
Impact: Hand-carried edits to generated i18n output can be silently reverted or conflict on the next generation run.

<details>
<summary>Rule text and affected files</summary>

`CLAUDE.md` Core Rules: "Never modify generated translations (`translations.ts`, locale JSON files)." `.cursor/rules/locale-handling.mdc` repeats it: "Do not modify the code in `@onekeyhq/shared/src/locale/enum/translations.ts` as these are automatically generated" and "Do not modify the locale JSON files in `@onekeyhq/shared/src/locale/json/*`".

The diff adds ~90 enum entries plus matching keys in `packages/shared/src/locale/json/en_US.json` and `packages/shared/src/locale/json/th_TH.json`, and also changes existing values (`protocol_tips__title` in en_US, `backup.file_permanently_deleted` in th_TH). The PR description states these came from the release branch rather than a hand edit; if that is the accepted workflow the rule text should be updated, otherwise these files should be regenerated via the i18n pull instead of merged.

</details>

PR #12937, opened 2026-08-19
feat: improve perps footer ticker marquee

Kahnchan:
## Summary

- Rework the Web Perps footer ticker into a smooth requestAnimationFrame marquee with hover pause and reduced-motion support.
- Use compact fixed-width numeric columns with tabular figures and no wrapping so live prices do not shift item widths.
- Align popular tickers with the server Hot tab without a local item limit, and move connection details into a status tooltip.

## Test Plan

- yarn agent:check --profile pr
- yarn jest packages/kit/src/views/Perp/components/FooterTicker/footerTickerUtils.test.ts packages/kit/src/views/Perp/utils/popularTickers.test.ts --runInBand
- Web Rspack dev compilation (passes with 52 existing third-party warnings)

### DIFF
--- a/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
+++ b/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
@@ -1,4 +1,4 @@
-import { memo, useCallback } from 'react';
+import { memo, useCallback, useState } from 'react';
 
 import { SizableText, XStack } from '@onekeyhq/components';
 
@@ -8,10 +8,8 @@ import { getFooterTickerDisplayText } from './footerTickerUtils';
 
 import type { IFooterTickerItemData } from './footerTickerUtils';
 
-const CHANGE_COLUMN_WIDTH = '8ch';
-const PRICE_COLUMN_WIDTH = '10ch';
-const DISPLAY_NAME_CHARACTER_WIDTH_PX = 8;
-const FIXED_ITEM_CONTENT_WIDTH_PX = 156;
+const ITEM_CHARACTER_WIDTH_PX = 7.75;
+const ITEM_INTERNAL_GAPS_WIDTH_PX = 12;
 
 interface IFooterTickerItemProps extends IFooterTickerItemData {
   isDuplicate?: boolean;
@@ -39,10 +37,13 @@ function FooterTickerItem({
     change24hPercent,
     markPrice,
   });
-  // Keep live price updates from changing the item's layout width.
-  const itemWidth =
-    displayName.length * DISPLAY_NAME_CHARACTER_WIDTH_PX +
-    FIXED_ITEM_CONTENT_WIDTH_PX;
+  // Freeze the initial text budget so live prices do not resize the item.
+  const [itemWidth] = useState(
+    () =>
+      (displayName.length + changeText.length + priceText.length) *
+        ITEM_CHARACTER_WIDTH_PX +
+      ITEM_INTERNAL_GAPS_WIDTH_PX,
+  );
   const handlePress = useCallback(() => {
     onPress({
       displayName,
@@ -89,9 +90,7 @@ function FooterTickerItem({
         size="$bodySmMedium"
         color={color}
         style={TABULAR_NUMS_STYLE}
-        width={CHANGE_COLUMN_WIDTH}
         flexShrink={0}
-        overflow="hidden"
         whiteSpace="nowrap"
       >
         {changeText}
@@ -101,9 +100,7 @@ function FooterTickerItem({
         color="$textSubdued"
         $group-hover={{ color: '$text' }}
         style={TABULAR_NUMS_STYLE}
-        width={PRICE_COLUMN_WIDTH}
         flexShrink={0}
-        overflow="hidden"
         whiteSpace="nowrap"
       >
         {priceText}
--- a/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
+++ b/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
@@ -6,15 +6,13 @@ import { TABULAR_NUMS_STYLE } from '../FavoritesBar/FavoriteTokenItem';
 
 import { getFooterTickerDisplayText } from './footerTickerUtils';
 
-import type {
-  IFooterTickerItemData,
-  IFooterTickerTextWidthBudget,
-} from './footerTickerUtils';
+import type { IFooterTickerItemData } from './footerTickerUtils';
+
+const CHANGE_COLUMN_WIDTH = '8ch';
+const PRICE_COLUMN_WIDTH = '10ch';
 
 interface IFooterTickerItemProps extends IFooterTickerItemData {
   isDuplicate?: boolean;
-  isMeasure?: boolean;
-  widthBudget?: IFooterTickerTextWidthBudget;
   onPress: (item: IFooterTickerItemData) => void;
 }
 
@@ -27,8 +25,6 @@ function FooterTickerItem({
   change24hPercent,
   markPrice,
   isDuplicate,
-  isMeasure,
-  widthBudget,
   onPress,
 }: IFooterTickerItemProps) {
   const color = change24hPercent >= 0 ? '$textSuccess' : '$textCritical';
@@ -64,18 +60,15 @@ function FooterTickerItem({
 
   return (
     <XStack
-      aria-hidden={isDuplicate || isMeasure || undefined}
-      tabIndex={isDuplicate || isMeasure ? -1 : undefined}
-      onPress={isMeasure ? undefined : handlePress}
+      aria-hidden={isDuplicate || undefined}
+      tabIndex={isDuplicate ? -1 : undefined}
+      onPress={handlePress}
       group
       userSelect="none"
       alignItems="center"
       gap="$1.5"
-      cursor={isMeasure ? 'default' : 'pointer'}
+      cursor="pointer"
       flexShrink={0}
-      style={{
-        width: isMeasure ? 'max-content' : widthBudget?.itemWidth,
-      }}
     >
       <SizableText
         size="$bodySmMedium"
@@ -89,7 +82,7 @@ function FooterTickerItem({
         size="$bodySmMedium"
         color={color}
         style={TABULAR_NUMS_STYLE}
-        width={isMeasure ? undefined : widthBudget?.changeWidth}
+        width={CHANGE_COLUMN_WIDTH}
         flexShrink={0}
         overflow="hidden"
         whiteSpace="nowrap"
@@ -101,7 +94,7 @@ function FooterTickerItem({
         color="$textSubdued"
         $group-hover={{ color: '$text' }}
         style={TABULAR_NUMS_STYLE}
-        width={isMeasure ? undefined : widthBudget?.priceWidth}
+        width={PRICE_COLUMN_WIDTH}
         flexShrink={0}
         overflow="hidden"
         whiteSpace="nowrap"
--- a/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
+++ b/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
@@ -10,7 +10,6 @@ import type { IFooterTickerItemData } from './footerTickerUtils';
 
 const CHANGE_COLUMN_WIDTH = '8ch';
 const PRICE_COLUMN_WIDTH = '10ch';
-const MIN_DISPLAY_NAME_WIDTH_PX = 32;
 const DISPLAY_NAME_CHARACTER_WIDTH_PX = 8;
 const FIXED_ITEM_CONTENT_WIDTH_PX = 156;
 
@@ -42,10 +41,8 @@ function FooterTickerItem({
   });
   // Keep live price updates from changing the item's layout width.
   const itemWidth =
-    Math.max(
-      MIN_DISPLAY_NAME_WIDTH_PX,
-      displayName.length * DISPLAY_NAME_CHARACTER_WIDTH_PX,
-    ) + FIXED_ITEM_CONTENT_WIDTH_PX;
+    displayName.length * DISPLAY_NAME_CHARACTER_WIDTH_PX +
+    FIXED_ITEM_CONTENT_WIDTH_PX;
   const handlePress = useCallback(() => {
     onPress({
       displayName,
--- a/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
+++ b/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
@@ -10,6 +10,9 @@ import type { IFooterTickerItemData } from './footerTickerUtils';
 
 const CHANGE_COLUMN_WIDTH = '8ch';
 const PRICE_COLUMN_WIDTH = '10ch';
+const MIN_DISPLAY_NAME_WIDTH_PX = 32;
+const DISPLAY_NAME_CHARACTER_WIDTH_PX = 8;
+const FIXED_ITEM_CONTENT_WIDTH_PX = 156;
 
 interface IFooterTickerItemProps extends IFooterTickerItemData {
   isDuplicate?: boolean;
@@ -37,6 +40,12 @@ function FooterTickerItem({
     change24hPercent,
     markPrice,
   });
+  // Keep live price updates from changing the item's layout width.
+  const itemWidth =
+    Math.max(
+      MIN_DISPLAY_NAME_WIDTH_PX,
+      displayName.length * DISPLAY_NAME_CHARACTER_WIDTH_PX,
+    ) + FIXED_ITEM_CONTENT_WIDTH_PX;
   const handlePress = useCallback(() => {
     onPress({
       displayName,
@@ -69,6 +78,7 @@ function FooterTickerItem({
       gap="$1.5"
       cursor="pointer"
       flexShrink={0}
+      width={itemWidth}
     >
       <SizableText
         size="$bodySmMedium"
--- a/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
+++ b/packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
@@ -1,15 +1,12 @@
 import { memo, useCallback, useState } from 'react';
 
-import { SizableText, XStack } from '@onekeyhq/components';
+import { SizableText } from '@onekeyhq/components';
 
 import { TABULAR_NUMS_STYLE } from '../FavoritesBar/FavoriteTokenItem';
 
 import { getFooterTickerDisplayText } from './footerTickerUtils';
 
 import type { IFooterTickerItemData } from './footerTickerUtils';
-
-const ITEM_CHARACTER_WIDTH_PX = 7.75;
-const ITEM_INTERNAL_GAPS_WIDTH_PX = 12;
 
 interface IFooterTickerItemProps extends IFooterTickerItemData {
   isDuplicate?: boolean;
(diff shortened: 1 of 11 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/views/Perp/hooks/usePopularTickers.ts
```diff
@@ -128,15 +131,20 @@ export function usePopularTickers(): IPopularTickerItem[] {
...
-        .slice(0, POPULAR_TICKER_COUNT)
-        .map(({ marketCap, ...item }) => item);
+      return scored.map(({ marketCap, ...item }) => {
+        const displayCtx = formatSpotPriceEntry(spotPriceMap[item.coinName]);
+        return {
+          ...item,
+          change24hPercent: displayCtx?.change24hPercent ?? 0,
+          markPrice: displayCtx?.markPrice,
+        };
+      });
```
sidmorizon, line 141: 🟠 P1: Footer price ticker can render the entire market list instead of a short highlight list

Severity: `severe`

The cap that kept the footer highlight list to ten entries was removed (`scored.map(...)` at `packages/kit/src/views/Perp/hooks/usePopularTickers.ts:134-141` and the un-sliced fallback in `packages/kit/src/views/Perp/utils/popularTickers.ts`), so in spot mode and whenever the server list is unavailable the strip is built from every tradable market, each rendered twice for the scrolling loop.
Impact: The bottom price strip can contain hundreds to thousands of live-updating entries, making the app sluggish and pushing most entries so far out of view that they effectively never scroll into sight.

<details>
<summary>Why the limit removal is unbounded outside the server Hot tab path</summary>

- Spot path: `packages/kit/src/views/Perp/hooks/usePopularTickers.ts:104-141` scores every entry of the full spot universe and now returns all of them (previously `.slice(0, POPULAR_TICKER_COUNT)`); there is no hot-tab filtering for spot.
- Perp fallback: `pickPopularPerpTickers` returns `items.filter(hotScore > 0).toSorted(...)` with no slice, i.e. the whole perp universe across all dexes, whenever `hotTabTokens` is empty or matches nothing.
- `packages/kit/src/views/Perp/components/FooterTicker/PerpFooterTicker.web.tsx:43-47` feeds this straight into `FooterTickerMarquee`, which renders `displayItems` twice (`renderLoop(false)` + `renderLoop(true)`), and every item re-renders on each price tick.
- At `SCROLL_SPEED_PX_PER_SEC = 20`, a loop of tens of thousands of pixels takes many minutes to a full cycle.

</details>

originalix:

@Kahnchan

Reviewed commit `0682d5741283`. This report always reflects the latest reviewed changes.

### Review summary

This PR turns the Perps web footer ticker into a web-only marquee controller that owns row rendering, motion, and quote display, and it adds a tooltip-based network-status summary to the same footer surface. That shifts ownership of list transitions and layout stability out of the previous item-level quote components and into a custom `requestAnimationFrame` state machine backed by precomputed ticker rows and structure-merge helpers. The presentation becomes richer, but correctness now depends on new footer-specific logic for list capping, structure handoff, and animation idling while the native footer path remains unchanged.

**What needs attention:** Confirm that popular-mode list size, structure swaps, and idle animation behavior converge correctly across mode switches, hover pause, and reduced-motion states.

### Issues to address

- **P1 · Popular footer ticker can expand to the full market universe**
  - **Impact:** The spot path and the perp fallback path both removed the old cap, so the footer can render hundreds or thousands of live rows and then duplicate that loop for scrolling. That makes the strip slow, pushes many symbols effectively out of reach, and turns a highlight surface into an unbounded feed.
  - **Suggested action:** Restore an explicit item cap or viewport-sized budget for popular-mode rows in both the spot ranking path and the perp fallback path, even if server hot-tab ordering is preserved.
- **P1 · Deferred structure swaps leave stale symbols visible and clickable**
  - **Impact:** When the popular ticker membership or order changes while the marquee is already scrolling, the code intentionally keeps the old structure and only merges live values. Users can keep seeing and clicking outdated symbols until the loop wraps, and hover pause can freeze that stale list indefinitely.
  - **Suggested action:** Apply list membership or ordering changes immediately, and reserve live-value merging for cases where the displayed symbol set is unchanged.
- **P2 · The marquee burns animation frames even when nothing is moving**
  - **Impact:** The `requestAnimationFrame` loop reschedules itself forever, including reduced-motion, no-scroll, and hover-paused states. That keeps rewriting the same transform and can repeatedly remeasure layout while the Perps screen is open, which is a concrete desktop/web CPU and battery regression.
  - **Suggested action:** Stop scheduling frames when scrolling is not needed or motion is paused, and restart the loop only when measurement or interaction shows that animation must resume.

### Validation gaps

- Targeted Jest execution for `packages/kit/src/views/Perp/components/FooterTicker/footerTickerUtils.test.ts` and `packages/kit/src/views/Perp/utils/popularTickers.test.ts` could not run because Yarn reported a missing `node_modules` state file.
- No browser runtime verification was available for hover pause, reduced-motion transitions, or live quote updates under the new marquee controller.

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
```diff
@@ -1,153 +1,106 @@
...
-          dexIndex={dexIndex}
-          assetId={assetId}
-          mode={mode}
-        />
-      )}
+      <SizableText
+        size="$bodySmMedium"
+        color={color}
+        style={TABULAR_NUMS_STYLE}
+        width={CHANGE_COLUMN_WIDTH}
```
sidmorizon: # 🟡 P2 固定 8ch 涨跌列会把 ≥1000% 的涨幅静默截断成错误数值

## 问题

涨跌幅列使用固定 `width={CHANGE_COLUMN_WIDTH}`（`8ch`）并叠加 `overflow="hidden"` + `whiteSpace="nowrap"`，且没有省略号。`getFooterTickerDisplayText` 生成 `+${change24hPercent.toFixed(2)}%`，当 24h 涨幅 `>= 1000%`（新上线代币在 spot 模式下并不罕见，且 `change24hPercent` 无上限封顶）时产出如 `+1234.56%` 共 9 个字符——tabular-nums 下每个数字恰为 `1ch`，加上 `+`、`.`、`%` 必然超出 `8ch`，右侧被硬裁切。

## 影响

用户看到的是被裁掉 `%` 甚至末位数字的字符串（如 `+1234.5`），呈现为一个貌似合法但错误的数值，出现在交易页 footer 上具有误导性。基线版本无固定宽度截断逻辑，该行为由本 PR 引入。

## 建议

把 `width` 改为 `minWidth`（常规值仍占满预算防抖动，极端值允许撑开），或对大数值换档格式化以保持在 `8ch` 预算内：

```ts
const changeText =
  Math.abs(item.change24hPercent) >= 1000
    ? `${sign}${item.change24hPercent.toFixed(0)}%`
    : `${sign}${item.change24hPercent.toFixed(2)}%`;
```

[Claude session ↗](https://claude.ai/code/session_01LTD9DXJAvbuYgNZFZj1bnN)

---

originalix (commented):
@Kahnchan

Reviewed commit `3a5678ee5492`.

- **P1 · Deferred marquee updates keep stale instruments visible after a mode switch**
- **P2 · Footer active symbol falls back to batch quote data**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12937#issuecomment-5340437203)

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
```diff
@@ -1,153 +1,110 @@
...
+    change24hPercent,
+    markPrice,
+  });
+  // Freeze the initial text budget so live prices do not resize the item.
+  const [itemWidth] = useState(
+    () =>
+      (displayName.length + changeText.length + priceText.length) *
+        ITEM_CHARACTER_WIDTH_PX +
+      ITEM_INTERNAL_GAPS_WIDTH_PX,
+  );
```
sidmorizon: 🟠 P1: Ticker entries in the footer get a permanently wrong width, so prices overlap neighbouring entries

Severity: `severe`

Each footer ticker entry locks its width to the text it happened to show on its very first frame (`useState` initializer at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx:41-46`), which on first render is usually the placeholder "-" because prices have not arrived yet, so the real price text no longer fits.
Impact: Footer ticker prices are squeezed/overlapping and stay broken for the whole session.

<details>
<summary>Why the frozen width is measured against placeholder text</summary>

`getFooterTickerDisplayText` (`packages/kit/src/views/Perp/components/FooterTicker/footerTickerUtils.ts:23-31`) returns `priceText: '-'` when `markPrice` is undefined and `changeText: '+0.00%'` when the change defaults to `0` (`packages/kit/src/views/Perp/components/FooterTicker/PerpFooterTicker.web.tsx:71-72` and `packages/kit/src/views/Perp/hooks/usePopularTickers.ts:137-139` both fall back to `change24hPercent: 0` / `markPrice: undefined`).

On the first render of `FavoritesTickerList`, the asset contexts are typically still empty — `markAllAssetCtxsRequired()` is only invoked in an effect after mount (`packages/kit/src/views/Perp/components/FooterTicker/PerpFooterTicker.web.tsx:102-110`) — so every item mounts with `markPrice === undefined`. The width budget therefore becomes `(len(displayName) + 6 + 1) * 7.75 + 12`, while the eventual rendered content is e.g. `BTC  +2.31%  43,251.5`. Because the item key (`getFooterTickerItemKey`) does not include the price, the component never remounts and the width is never recomputed. The same issue occurs later whenever a price gains digits (e.g. `9,999.9` → `10,000.2`).

A width that only grows (tracked in a ref/state via `Math.max`) or a `minWidth` instead of a fixed `width` would preserve the "no jitter" goal without clipping.

</details>

originalix (commented):
@Kahnchan

Reviewed commit `25bcd8356133`.

- **P1 · Popular footer ticker can render the full market universe**
- **P1 · Deferred marquee updates keep stale instruments visible after popular-list changes**
- **P2 · Footer items no longer use the dedicated active-market quote path**
- **P2 · Ticker item width is frozen before live quote text settles**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12937#issuecomment-5340437203)

packages/kit/src/views/Perp/components/FooterTicker/PerpFooterTicker.web.tsx
```diff
@@ -101,15 +119,13 @@ function PerpFooterTicker() {
...
-        ) : (
-          <FavoritesTickerList />
-        )}
-      </FooterTickerMarquee>
+      {footerMode.mode === 'popular' ? (
+        <PopularTickerList />
+      ) : (
+        <FavoritesTickerList />
+      )}
     </XStack>
```
sidmorizon, line 129: **Footer help links jump to the middle of the bar while the price strip has no data**

Severity: `non-severe`

The flexible spacer that used to always occupy the free footer width was removed (the marquee is now rendered conditionally at `packages/kit/src/views/Perp/components/FooterTicker/PerpFooterTicker.web.tsx:124-128`), so while the price list is empty or still loading the footer's right-hand links collapse to the left and then jump back.
Impact: The footer links visibly jump position each time the ticker data loads or when the user has no favorites.

<details>
<summary>Layout mechanism</summary>

Previously `<FooterTickerMarquee>` was always mounted around the list and carried `flex={1}` (see the container XStack at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:248-256`), so it always filled the space between the status badges and `PerpFooterActions` (which is `flexShrink={0}` with no flex, `packages/kit/src/components/Footer/Footer.tsx:93`).

Now `PopularTickerList`/`FavoritesTickerList` return `null` when their list is empty (`PerpFooterTicker.web.tsx:40` and `:87`), so no flex filler exists and the actions block sits directly next to the refresh button until data arrives.

</details>

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.web.tsx
```diff
@@ -1,154 +1,110 @@
...
+        flexShrink: 0,
+        gap: 6,
+        width: 'max-content',
+        padding: 0,
+        border: 0,
+        background: 'transparent',
+        cursor: 'pointer',
+        userSelect: 'none',
+        whiteSpace: 'nowrap',
+      }}
```
sidmorizon, line 84: **Web-only DOM markup added to a component file shared by all platforms**

Severity: `non-severe`

Raw browser markup and inline CSS were put into a file without a platform suffix (`<button style={{ appearance: ... }}>` at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx:60-80`), which the repository's cross-platform rules forbid for shared component files.
Impact: Any future non-web import of this component would break, and the code bypasses the shared UI component library.

<details>
<summary>Rule references</summary>

`.cursor/rules/cross-platform.mdc` requires platform-specific implementations to live in `.web.tsx`-style files and requires using Tamagui components from `@onekeyhq/components` for UI; `CLAUDE.md` states "Use platform-specific files or `Platform.select` for platform behavior".

The previous implementation used `XStack` with `hoverStyle`, which works on all platforms. The new implementation uses `button`, `onClick`, `onMouseEnter`/`onMouseLeave`, `appearance`, and `whiteSpace`, all web-only, in `FooterTickerItem.tsx` (no `.web` suffix). Currently only `FooterTickerMarquee.web.tsx` imports it, so it happens not to break native today.

</details>

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.web.tsx
```diff
@@ -1,154 +1,110 @@
...
+      onClick={handlePress}
+      onMouseEnter={() => setIsHovered(true)}
+      onMouseLeave={() => setIsHovered(false)}
+      style={{
+        appearance: 'none',
+        display: 'inline-flex',
+        alignItems: 'center',
+        flexShrink: 0,
+        gap: 6,
+        width: 'max-content',
```
sidmorizon, line 77: # 🟡 P2 宽度约束全部移除后，实时报价的文本长度波动会让整条 marquee 抖动

## 问题

本 commit 把项宽改为 `width: 'max-content'`，且价格/涨跌两列不再有任何 `width`/`minWidth` 约束。而 `priceText` 来自 `formatPriceToSignificantDigits`，会剥尾零，长度随每次报价在 `4–8` 字符间波动（如 `'1.05'` ↔ `'1.0525'`）；`changeText` 跨位数边界（`+9.99%` → `+10.12%`）也会 ±1 字符。仓库中 `FavoriteTokenItem.tsx` 的 `getStablePriceMinWidth` 注释已明确记载这一抖动来源，但此处未复用。

由于 item、loop、track 全为 `max-content`，任何 item 的文本变长/变短都会平移其后所有 item 并改变单圈宽度；rAF 的 `offsetRef` 是相对圈起点的像素偏移，宽度变化时 ResizeObserver 只重测 `loopWidthRef` 而不补偿 offset，可视区内容会瞬间横跳一个宽度差。

## 影响

Favorites 模式下每次 `allAssetCtxs`/`spotPriceMap` tick（约每 `1–5` 秒）都可能改变文本长度，footer 会持续出现字符级横向抖动和整条位移。这正是 PR body 宣称要解决的核心问题（"live price updates within measured width budgets to prevent layout shifts"）——前三版的 `8ch` 列宽、px 估算、useState 冻结均已移除，当前 head 上没有任何替代的防抖机制。

## 建议

给两列恢复**数据无关**的 `minWidth`（不是 `width`，避免此前版本的截断/叠字问题）：价格列直接复用现成的 `getStablePriceMinWidth(priceText)`，涨跌列用固定预算（如 `'-100.00%'` 对应的 `8ch`），配合 `textAlign: 'right'`。`minWidth` 允许极端值撑开（由 ResizeObserver 重测兜底），常规波动则完全不改变版面。

[Claude session ↗](https://claude.ai/code/session_01LTD9DXJAvbuYgNZFZj1bnN)

---

originalix (commented):
@Kahnchan

Reviewed commit `9434a892ac0d`.

- **P1 · Popular footer ticker can render the full market universe**
- **P1 · Deferred marquee updates keep stale instruments visible after popular-list changes**
- **P2 · Footer drops the dedicated active-market quote source**
- **P2 · Live quote text can still shift the marquee layout on every tick**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12937#issuecomment-5340437203)

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.web.tsx
Kahnchan: Fixed in 0682d57412: renamed the component to FooterTickerItem.web.tsx and updated the marquee to import the Web-specific module explicitly, so the raw DOM markup is isolated to Web.

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.web.tsx
Kahnchan: Fixed in 0682d57412: reused getStablePriceMinWidth for prices and added a small 7ch minimum budget for percentage changes. The min-widths absorb normal live-text length changes while still allowing extreme values to grow without clipping.

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerItem.tsx
Kahnchan: Fixed in 9434a892ac and refined in 0682d57412: removed the first-render frozen width, switched to intrinsic item sizing, and added stable min-width budgets for live numeric text. This avoids both overlap and normal quote-length jitter.

packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx
```diff
@@ -1,106 +1,253 @@
...
+          }
+        } else {
+          offsetRef.current = nextOffset;
+        }
+        track.style.transform = `translate3d(${-offsetRef.current}px, 0, 0)`;
+      }
+
+      frameId = requestAnimationFrame(step);
+    };
+    frameId = requestAnimationFrame(step);
```
sidmorizon, line 199: **Price strip keeps a per-frame animation loop running even when nothing is moving**

Severity: `non-severe`

The scrolling loop keeps requesting a new animation frame forever (`requestAnimationFrame(step)` at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:197`) even when the strip does not need to scroll or the user asked for reduced motion, so work happens every frame with nothing to show.
Impact: The desktop/web app burns CPU and battery continuously while the perps screen is open, even for users who disabled animations.

<details><summary>Loop has no exit condition</summary>

`step` unconditionally schedules the next frame at `packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:197`, and in the reduced-motion / no-scroll branch it just re-writes the same `translate3d(0, 0, 0)` transform (`packages/kit/src/views/Perp/components/FooterTicker/FooterTickerMarquee.web.tsx:176-178`). It also keeps running while hover-paused. Previously the marquee used a CSS animation, so no JS ran per frame. When `loopWidthRef.current` is 0 the loop additionally calls `syncLoopMetrics()` every frame, which reads `getBoundingClientRect()` / `clientWidth` and forces a layout each frame.

A guard that stops scheduling frames when `prefersReducedMotionRef.current || !needsScrollRef.current || pausedRef.current`, and restarts it from `syncLoopMetrics` / `handleMouseLeave`, would avoid the idle work.

</details>

originalix (commented):
@Kahnchan

Reviewed commit `0682d5741283`.

- **P1 · Popular footer ticker can expand to the full market universe**
- **P1 · Deferred structure swaps leave stale symbols visible and clickable**
- **P2 · The marquee burns animation frames even when nothing is moving**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12937#issuecomment-5340437203)

PR #12939, opened 2026-08-19
feat: add native chart sub-indicators

limichange:
## Summary

- add 13 TradingView Native sub-chart indicator algorithms, with one implementation and test file per indicator
- expose a canonical 17-indicator catalog (4 main + 13 sub) and connect it directly to the Native and V2 controller selectors
- add a settings-ready render bridge: definition → resolved instance → calculation → RenderPane → shared scene
- render panes through the shared Web Canvas and Native Skia adapters, including autoscale, bands, fills, palettes, legends, and crosshair values
- keep VOL sub-chart-only and preserve Native mobile pane limits and indicator settings state

The 13 sub-chart indicators are VOL, MACD, RSI, StochRSI, OBV, MFI, TRIX, EMV, WR, ROC, MTM, DMI, and CCI.

## Why

The Native Controller could expose sub-chart panes and values, but the algorithms, settings model, and cross-platform drawing pipeline were not connected. This adds a typed, extensible boundary so future per-indicator settings can change inputs and styles without coupling calculation code to Canvas or Skia.

The blank-pane rendering issue found during UI verification was caused by nested `'worklet'` functions capturing mutable local values by copy. That left auto-range at `Infinity/-Infinity`, so axis, plot, band, and fill commands exited while legends still rendered. The nested directives were removed while preserving the outer worklets, and a real Worklets Babel transform regression test now covers this behavior.

## User impact

- all 13 secondary indicators are selectable and rendered on Native chart surfaces
- Web/Desktop Canvas and iOS/Android Skia consume the same platform-neutral scene
- VOL renders only when explicitly selected as a sub-chart
- indicator inputs and styles can evolve independently from algorithms and platform renderers

## Validation

- `yarn tsc:only`
- 61 targeted Jest suites / 471 tests
- local Web Market UI: BTC with VOL, MACD, RSI, and StochRSI rendered together


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/components/TradingView/TradingViewNative/utils/chartIndicators/rsi.ts
```diff
@@ -0,0 +1,91 @@
...
+    const averageGain = averageGains[index];
+    const averageLoss = averageLosses[index];
+    if (
+      isTradingViewNativeFiniteValue(averageGain) &&
+      isTradingViewNativeFiniteValue(averageLoss)
+    ) {
+      if (isTradingViewNativeIndicatorZero(averageLoss)) {
+        rsi[index] = 100;
+      } else if (isTradingViewNativeIndicatorZero(averageGain)) {
+        rsi[index] = 0;
```
sidmorizon, line 74: **Momentum indicators read as fully overbought for tokens with extremely small prices**

Severity: `non-severe`

Average gains and losses below one ten-billionth are treated as exactly zero (`isTradingViewNativeIndicatorZero(averageLoss)` at `packages/kit/src/components/TradingView/TradingViewNative/utils/chartIndicators/rsi.ts:71`) before the ratio is taken, so for tokens whose prices move in extremely small steps the reading is pinned to 100 (or 0) instead of tracking the real trend.
Impact: On very-low-priced tokens the RSI and Stochastic RSI sub-charts show a flat, maxed-out line, giving traders a misleading signal.

<details>
<summary>Absolute epsilon breaks the scale-invariance of the gain/loss ratio</summary>

`isTradingViewNativeIndicatorZero` (`packages/kit/src/components/TradingView/TradingViewNative/utils/chartIndicators/indicatorMath.ts:15-17`) compares against a fixed absolute epsilon of `1e-10`. RSI itself is scale invariant, but the guard is not: for a token priced around `1e-9`, per-bar deltas are on the order of `1e-11`, so both `averageGain` and `averageLoss` fall under the epsilon. The first branch then returns `100` for every bar, and `calculateTradingViewNativeStochasticRsi` (`packages/kit/src/components/TradingView/TradingViewNative/utils/chartIndicators/stochasticRsi.ts:39-77`) inherits the flat series and degenerates to the `range === 0` carry-forward path.

Other indicators in this PR explicitly guard against this class of input (see the "tiny but non-zero" tests in `packages/kit/src/components/TradingView/TradingViewNative/utils/chartIndicators/cci.test.ts:54-72`, `dmi.test.ts:75-101`, `momentum.test.ts:34-39`), and market V2 does surface tokens with such prices (the axis-label logic handles ranges down to `1e-9`, e.g. `packages/kit/src/components/TradingView/TradingViewNative/utils/subIndicatorRender/coordinates.test.ts:154-182`). `SUB_CHART_INDICATORS.md` documents the `1e-10` tolerance as legacy-bundle parity, but the practical effect on small-cap tokens is an unusable indicator. A relative tolerance (scaled by the average of gain+loss, or by price magnitude) would keep legacy parity for normal magnitudes while staying correct here.

</details>

originalix (approved):

PR #12905, opened 2026-08-17
fix: align hardware errors, wallet state, and Pro2 workflows

wabicai:
## Summary

- preserve hardware-error classification across native runtime boundaries and surface failures consistently in onboarding and firmware workflows
- reconcile hardware wallets with live device state after reset while keeping deprecated wallets available for read-only access
- align Pro2 QR-wallet, settings, default-network, and firmware-update behavior
- keep filtered portfolio assets represented in the other-assets count

## Context

This PR combines the hardware workflow fixes required for the v6.5.2 hotfix. The original title only described onboarding error handling, but the final diff also covers reset-device reconciliation, deprecated-wallet availability, Pro2 capability alignment, firmware retry presentation, and portfolio counts.

## Root causes

- Native main and background runtimes do not share error prototypes, so hardware errors reconstructed after RPC must be identified from serialized metadata.
- A scanned device record can be stale after reset; live `DeviceState` identity must win when reconciling persisted wallets.
- Several capability checks still assumed legacy device behavior and did not account for Pro2 QR and settings support.
- Portfolio filtering removed tokens from the visible payload without preserving their count in `otherTokens`.

## Changes

- preserve `$isHardwareError` through native background RPC serialization and main-runtime rehydration, then use shared hardware-error guards in onboarding and firmware-related flows
- prefer live device identity in both legacy and V2 onboarding during wallet reconciliation, while preserving deprecated wallets as viewable but non-writable
- align Pro2 settings, QR-wallet, default-network, and firmware-update handling
- calculate the other-assets count from the total token count minus emitted payload tokens
- add focused regression tests for error classification, wallet selection, device identity, Pro2 settings, and portfolio payloads

## Risk

- **Level:** Medium
- **Areas:** Native RPC error reconstruction, hardware-wallet reset reconciliation, deprecated-wallet visibility, Pro2 feature gates, and portfolio display counts
- Existing signing and write guards for deprecated wallets remain unchanged.

## Validation

- unit test CI: passed
- CodeQL and security checks: passed
- targeted regression tests added for the changed behavior
- focused native RPC and device identity tests: 61 passed
- type-aware lint on the review-fix files: passed with zero warnings
- lint formatting issue in `portfolioPayload.ts` corrected in the latest update
- device-level iOS verification remains recommended for Bluetooth-disabled-while-USB-connected onboarding

PR #12940, opened 2026-08-19
fix: restore desktop usb routing(OK-60693)

originalix:
[OK-60693](https://onekeyhq.atlassian.net/browse/OK-60693)

----------

## Summary

- Restore device-independent desktop USB availability: any authorized OneKey WebUSB or Bridge device keeps hardware calls on USB.
- Preserve the existing firmware-update USB override and Protocol V2 transport selection.
- Update the focused hardware connection tests to cover the intended global USB-presence behavior.

## Intent & Context

OK-60693 reports that a OneKey Classic 1S connected by USB still used Bluetooth during BTC signing. The intended desktop policy is deliberately simple: when any usable OneKey USB device is present, hardware calls should prefer USB.

PR #12905 changed the shared USB availability probe to require the enumerated WebUSB serial or Bridge path to exactly match the current `connectId`. That review-driven change conflicted with the established transport policy and caused the signing path to fall back to Desktop BLE when the current identity did not match the USB transport identity.

## Root Cause

`detectWebUSBAvailability()` and `detectBridgeAvailability()` are generic transport-selection probes, not selected-device ownership validators. Exact `connectId` matching made their result device-specific even though callers and the two-second routing cache were designed around a global USB-presence signal.

For the reproduced Classic 1S flow, WebUSB discovery and pre-initialization succeeded with USB connect ID `CLA45F0023`, but the exact-match probe could still report USB unavailable and route the formal signing call through NobleBLE.

## Design Decisions

- Revert only the selected-device equality checks in the generic WebUSB and Bridge probes.
- Keep the OneKey VID/PID and non-empty serial filtering for WebUSB, so only authorized devices usable by the transport count as present.
- Keep the `UPDATE_FIRMWARE` USB-only branch and explicit Protocol V2 preflight parameters. The entire firmware-related change is not reverted.
- Keep the existing method signatures to avoid changing callers; the now-intentionally-unused `connectId` parameters are prefixed with `_`.
- Accept the multi-device behavior as intentional product policy: the presence of any OneKey USB device forces USB routing.

## Changes Detail

- `HardwareConnectionManager.ts`: restore global OneKey WebUSB/Bridge presence checks.
- `ServiceHardware.getCompatibleConnectId.test.ts`: assert that an unrelated selected `connectId` still resolves USB/Bridge as available when a OneKey device is enumerated.

## Risk Assessment

- **Risk Level**: Low
- **Risk Areas**: Desktop automatic transport selection when multiple hardware devices are nearby. The global USB preference is intentional; firmware update transport selection remains unchanged.

## Test plan

- [x] `yarn jest packages/kit-bg/src/services/ServiceHardware/ServiceHardware.getCompatibleConnectId.test.ts --runInBand` — 58 tests passed.
- [x] `yarn agent:check --profile commit` — worktree lint, staged lint, and TypeScript checks passed.
- [x] `release-app-bundles.yml` run `32247325784` completed successfully for commit `b46849499bffacc1c9dd5bf3f35fbf602940bad1`.
- [x] TF bundle `19913008`: pre-initialization and `btcSignTransaction` both used WebUSB connect ID `CLA45F0023`; no NobleBLE or Desktop BLE fallback occurred.
- [x] The device reached the signing confirmation step over WebUSB and returned only user cancellation (`803`), not a transport failure.

Jira: https://onekeyhq.atlassian.net/browse/OK-60693
Regression source: https://github.com/OneKeyHQ/app-monorepo/pull/12905

[OK-60693]: https://onekeyhq.atlassian.net/browse/OK-60693?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
+++ b/packages/kit-bg/src/services/ServiceHardware/HardwareConnectionManager.ts
@@ -139,7 +139,7 @@ export class HardwareConnectionManager {
   }
 
   // WebUSB detection
-  async detectWebUSBAvailability(connectId?: string): Promise<boolean> {
+  async detectWebUSBAvailability(_connectId?: string): Promise<boolean> {
     if (!platformEnv.isSupportDesktopBle) return true;
     try {
       const usb = globalThis?.navigator?.usb;
@@ -156,18 +156,13 @@ export class HardwareConnectionManager {
           dev.serialNumber.trim().length > 0;
         return isOneKey && hasSerialNumber;
       });
-      const expectedConnectId = connectId?.trim().toLowerCase();
-      return onekeyDevices.some(
-        (device) =>
-          !expectedConnectId ||
-          device.serialNumber?.trim().toLowerCase() === expectedConnectId,
-      );
+      return onekeyDevices.length > 0;
     } catch {
       return false;
     }
   }
 
-  async detectBridgeAvailability(connectId?: string): Promise<boolean> {
+  async detectBridgeAvailability(_connectId?: string): Promise<boolean> {
     if (!platformEnv.isSupportDesktopBle) {
       return true;
     }
@@ -185,14 +180,7 @@ export class HardwareConnectionManager {
       if (!Array.isArray(devices)) {
         return false;
       }
-      const expectedConnectId = connectId?.trim().toLowerCase();
-      return devices.some((device) => {
-        if (!expectedConnectId) return true;
-        return (
-          typeof device.path === 'string' &&
-          device.path.trim().toLowerCase() === expectedConnectId
-        );
-      });
+      return devices.length > 0;
     } catch (_error) {
       return false;
     }
--- a/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.getCompatibleConnectId.test.ts
+++ b/packages/kit-bg/src/services/ServiceHardware/ServiceHardware.getCompatibleConnectId.test.ts
@@ -1137,7 +1137,7 @@ describe('ServiceHardware.getCompatibleConnectId', () => {
     ]);
   });
 
-  it('uses USB only when the selected authorized OneKey WebUSB device is available', async () => {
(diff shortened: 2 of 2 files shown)


wabicai (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

@originalix

Reviewed commit `b46849499bff`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes the desktop hardware transport-selection boundary inside `HardwareConnectionManager`: WebUSB and Bridge availability are now treated as global signals that some OneKey USB device exists, instead of checks scoped to the selected `connectId`. The matching tests are updated to encode that broader rule. BLE permission handling, SDK reset behavior, and native/mobile transport paths remain unchanged, but the desktop per-device routing guarantee becomes looser and can now select the wrong transport when multiple devices are present.

**What needs attention:** Recheck the desktop transport-selection boundary before merge: USB availability must stay tied to the selected device identity before `getCompatibleConnectId()` commits WebUSB or Bridge over BLE.

### Issues to address

- **P1 · Global USB detection breaks per-device transport routing**
  - **Impact:** `getCompatibleConnectId()` still resolves transport for the selected device using that device's persisted USB identity, but this PR makes WebUSB and Bridge availability succeed when any OneKey USB device is present. In a desktop multi-device setup, a BLE-only target can therefore be routed back to USB because another device is connected, which can make the intended hardware call fail or hit the wrong transport.
  - **Suggested action:** Restore target-scoped matching for WebUSB and Bridge availability, then add a regression test where device A is selected over BLE while an unrelated device B is the only USB or Bridge device currently present.

### Validation gaps

- No regression test covers the mixed-device case where the selected device is BLE-only and a different device is the only USB or Bridge device currently present.
- This review was static; no desktop runtime verification was available in this run.

PR #12941, opened 2026-08-19
fix: prevent DotMap row number wrapping

huhuanming:
## Summary

- reserve a non-shrinking 40dp lane for DotMap row indices and the matching header
- reuse the original left padding so the matrix position and card width stay unchanged
- keep row indices on a single line

## Root cause

Android devices narrower than 400dp apply the global 0.9 UI scale. The original 24dp row-index lane became too narrow after scaling and font/pixel rounding could wrap two-digit indices such as 10–14.

## Impact

KeyTag recovery DotMap row indices remain single-line without shifting or shrinking the matrix. The shared layout remains unchanged on iOS.

## Validation

- `git diff --check`
- Android Pixel 6 emulator at 360dp, triggering the 0.9 scale: verified the actual 24-word KeyTag backup page, including rows 10–14 and 21–24
- iPhone 17 Pro simulator: verified the actual 24-word KeyTag backup page with rows 10–24 rendered on one line


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/kit/src/components/DotMap/index.tsx
```diff
@@ -80,8 +80,13 @@ const DotWord = ({ values }: IDotWordProps) => {
 type IDotMapRowProps = { data: IDotMapValues };
 const DotMapRow = ({ data }: IDotMapRowProps) => (
   <XStack alignItems="center">
-    <XStack width="$6" justifyContent="flex-end">
-      <SizableText size="$bodySm" mr="$2" color="$textSubdued">
+    <XStack width="$10" flexShrink={0} justifyContent="flex-end">
```
sidmorizon, line 83: **Android-only layout problem is fixed by changing the shared layout for every platform**

Severity: `non-severe`

The row-number lane and card padding are widened/reshuffled for all platforms (`width="$10"` / `pl={0}` at `packages/kit/src/components/DotMap/index.tsx:83-142`) even though the reported problem only happens on narrow Android devices, so a platform-specific issue is patched with a global style change instead of platform-scoped styling.
Impact: The backup grid layout of iOS, desktop, web and extension users is altered by a fix that was only needed on Android, and future tweaks to this shared style risk regressing all platforms at once.

<details>
<summary>Repository rule requiring platform-scoped styling for platform-specific visual bugs</summary>

`CLAUDE.md` (Core Rules) states: "Prefer platform-specific files or `Platform.select` for platform behavior. Do not apply global CSS/style fixes for platform-specific bugs."

The PR description identifies the root cause as "Android devices narrower than 400dp apply the global 0.9 UI scale", yet the change edits the shared `DotMap` component: `DotMapRow` lane goes from `width="$6"` to `width="$10" flexShrink={0}` (`packages/kit/src/components/DotMap/index.tsx:83-89`), the header lane likewise (`packages/kit/src/components/DotMap/index.tsx:108`), and container padding changes from `p="$4"` to `py="$4" pl={0} pr="$4"` (`packages/kit/src/components/DotMap/index.tsx:140-142`). This component is consumed by `packages/kit/src/views/KeyTag/pages/BackupDotMap/index.tsx:63` and the gallery story on all platforms. The total left offset happens to stay at 40 scaled units, but the shared component's internal layout contract (index lane no longer shrinks; card has zero left padding) is now different everywhere, which is what the rule aims to avoid.

</details>

packages/kit/src/components/DotMap/index.tsx
```diff
@@ -132,7 +137,9 @@ export const DotMapBase = ({ items }: { items: IDotMapValues[] }) => {
   return (
     <Stack
       backgroundColor="$bgSubdued"
-      p="$4"
+      py="$4"
+      pl={0}
+      pr="$4"
```
sidmorizon, line 142: **Bug-fix retrospective entry required by repo rules is missing**

Severity: `non-severe`

The required retrospective note for this bug fix was never added (no change to `.claude/skills/1k-retrospective/references/case-studies.md` alongside the fix at `packages/kit/src/components/DotMap/index.tsx:83-142`), so the team's record of fixed defects is incomplete.
Impact: The knowledge base used to prevent repeat regressions loses this case.

<details>
<summary>Mandatory "Bug Fix Case Recording" rule</summary>

`.cursor/rules/self-testing.mdc` (alwaysApply: true) states: "After every **bug fix**, append to `.claude/skills/1k-retrospective/references/case-studies.md`" with the Title/Date/Platforms/Symptom/Root Cause/Fix/Catchable-by template. The file is actively maintained (latest entries dated 2026-08-18), but this bug-fix PR only touches `packages/kit/src/components/DotMap/index.tsx`.

</details>

packages/kit/src/components/DotMap/index.tsx
huhuanming: Thanks for flagging this. We intentionally keep this as a unified DotMap layout invariant rather than an Android-only override. The total left allocation remains 40dp (`$4` padding + `$6` lane becomes `0` padding + `$10` lane), so the card width and matrix origin are unchanged on other platforms; the reallocated space only gives row labels a reliable non-shrinking lane. `numberOfLines={1}` also makes the no-wrap invariant explicit for every consumer. We verified the final 24-word layout on Android at 360dp with the 0.9 scale active and on an iPhone 17 Pro simulator, so we will keep the shared implementation.

packages/kit/src/components/DotMap/index.tsx
huhuanming: Agreed. I added the required retrospective case locally at `.skillshare/skills/1k-retrospective/references/case-studies.md` (the tracked path behind the `.claude` alias), covering the date, affected Android width, symptom, root cause, unified fix, and the narrow-mobile verification check that would catch it. It is ready for the follow-up commit and push.

sidmorizon:
@cursoragent review

originalix (approved):

PR #12942, opened 2026-08-19
fix: detect hardware disconnect on desktop USB and BLE (OK-60486)

originalix:
[OK-60486](https://onekeyhq.atlassian.net/browse/OK-60486)

----------

Fixes [OK-60486](https://onekeyhq.atlassian.net/browse/OK-60486) — after interacting
with a device, unplugging it left the wallet-list connected dot lit.

## Root cause

The dot reads the **union** of two sources in `useHardwareWalletConnectStatus`:
`navigator.usb.getDevices()`, which is live OS enumeration and self-clearing, and
`serviceHardware.getConnectedHardwareDeviceIdentityKeys()`, a background map.
`buildHardwareConnectedDeviceKeys` is purely additive, so the live source cannot
subtract a key the map added.

The map's only writers are the `DEVICE.SUPPORT_FEATURES` and `DEVICE.CONNECT`
handlers, both of which require an SDK call — which is exactly why plain
plug/unplug behaved correctly and only interaction-then-unplug did not. Its only
per-device eviction is `untrackConnectedDevice`, reachable solely from the
`DEVICE.DISCONNECT` handler; the four `clearTrackedConnectedDevices` call sites are
all SDK-teardown paths (firmware manifest change, transport switch, new SDK
instance, reset) and none fire on a plain unplug. On desktop `DEVICE.DISCONNECT`
never fired at all, so once populated the map was sticky.

This is a regression against the released app, where the dot was WebUSB enumeration
only and therefore inherently self-clearing. It was introduced when the background
map became a second source of truth.

## Changes

- **SDK 1.2.0-alpha.150 → 1.2.0-alpha.155**, which makes `DEVICE.DISCONNECT` fire on
  both desktop transports. The existing `DEVICE.DISCONNECT` handler then evicts the
  keys and broadcasts, with no further app-side wiring needed. Fixed in
  OneKeyHQ/hardware-js-sdk#891.
- Thread the new disconnect `reason` through the preload bridge and `IDesktopApi`
  types. The payload was already forwarded whole at runtime; this makes the types
  match.
- Guard `refreshDevices` with a request sequence. The `usb` disconnect event and
  `HardwareConnectionStateUpdate` can both fire and fetch concurrently, and the
  slower response could overwrite the newer one and re-light a device that had
  already gone. Independent of the main bug, but it would make the fix
  intermittently fail.
- Log the eviction path. It was previously silent end to end, which is why the
  original report could only prove the event's absence indirectly — a disconnect
  that never arrived and one that left no trace were indistinguishable in collected
  logs.

## Not changed, deliberately

Removing the background map from the union would fix the reported symptom on
desktop USB but re-open [OK-60117](https://onekeyhq.atlassian.net/browse/OK-60117)
(the map is what matches a device whose stored `usbConnectId` went stale after a
reset — `LocalDbBase` only writes that field when absent and never refreshes it),
drop the dot entirely for desktop BLE and native, break the Device Management list's
Pro 2 indicator, and leave `isHardwareDeviceConnected` — which reads the map
directly and short-circuits before the WebUSB check — still serving stale state to
portfolio sync. The stickiness was the bug, not the union.

## Verification

macOS desktop, OneKey Classic 1S:

| Case | Result |
| --- | --- |
| USB: interact, then unplug | dot clears |
| USB: plug/unplug with no interaction | dot clears (unchanged) |
| BLE: interact, then drop the device | dot clears |
| BLE: interact, then idle past the keep-alive window | dot clears |
| Reconnect after any of the above | dot returns |

Confirmed in the log by `device disconnected, untracked` with a non-zero removed-key
count. The same build before the SDK bump logged the keep-alive timeout with nothing
after it, leaving three identity keys resident.

`yarn agent:check --profile commit` passes.

[OK-60486]: https://onekeyhq.atlassian.net/browse/OK-60486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60117]: https://onekeyhq.atlassian.net/browse/OK-60117?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Desktop BLE: what the dot means

The dot tracks whether a BLE **link** is live, not whether the peripheral is
theoretically in range. The main process reclaims an idle link on its keep-alive
timer (`BLE_IDLE_DISCONNECT_MS`, 180s; 600s on the `keepSession` path), and that
is reported like any other drop — a link we closed ourselves is still a closed
link, and claiming a connection that no longer exists is the same class of
false report this PR set out to fix.

So on desktop BLE the dot goes dark roughly 180s after the last interaction and
returns on the next one. It settles once rather than oscillating: nothing
reconnects on its own, because background portfolio sync only piggybacks on an
already-open link and its `connect` is a no-op inside that scope
(`apps/desktop/app/preload.ts`). Recovery does not depend on `DEVICE.CONNECT` —
`trackConnectedDevice` is also called from the `DEVICE.SUPPORT_FEATURES` handler,
which fires on every SDK call.

The `reason` field carried on the disconnect event is **diagnostics only**; it
records whether the link dropped on its own or was reclaimed, and gates nothing.

This differs from USB, where the dot follows device presence and stays lit as
long as the device is plugged in.


sidmorizon:
@codex review

sidmorizon:
@codex security review

wabicai (dismissed):

package.json
```diff
@@ -158,21 +158,21 @@
...
+    "@onekeyfe/hd-transport-electron": "1.2.0-alpha.155",
+    "@onekeyfe/hd-web-sdk": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-adapter-core": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-ledger-adapter": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-ledger-connector-ble": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-ledger-connector-webhid": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-trezor-adapter": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-trezor-connector-electron-ble": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-trezor-connector-rn-ble": "1.2.0-alpha.155",
+    "@onekeyfe/hwk-trezor-connector-webusb": "1.2.0-alpha.155",
```
sidmorizon, line 175: **Hardware SDK version bump left one app behind, breaking the repo's version-consistency check**

Severity: `non-severe`

The hardware SDK packages are bumped only in the root manifest (`1.2.0-alpha.155` at `package.json:161-175`) while the command-line app still pins the old `1.2.0-alpha.150`, so the repository's mandatory version-consistency check now fails and the two apps ship mismatched hardware libraries.
Impact: The lint/CI gate required before commit fails, and the CLI ends up running a mix of old and new hardware SDK pieces.

<details>
<summary>Version-consistency lint and lockfile evidence</summary>

`development/lint/package-versions.js` compares every dependency version across all workspace `package.json` files and exits non-zero on any mismatch; it is wired into `yarn lint:project` via `_packageVersions` (`package.json:90-92`), which `yarn agent:check --profile commit` runs. Running it on this branch reports:

- `@onekeyfe/hd-core`: `1.2.0-alpha.155` (root) vs `1.2.0-alpha.150` (`apps/cli/package.json:48`)
- `@onekeyfe/hd-common-connect-sdk`: `1.2.0-alpha.155` (root) vs `1.2.0-alpha.150` (`apps/cli/package.json:47`)

Additionally, root `resolutions` pin `@onekeyfe/hd-core` / `hd-shared` / `hd-transport` to `alpha.155`, while `yarn.lock` still resolves `@onekeyfe/hd-common-connect-sdk@1.2.0-alpha.150` and `@onekeyfe/hd-transport-usb@1.2.0-alpha.150` for the CLI workspace — i.e. alpha.150 transports composed with alpha.155 core/shared in that workspace. `apps/cli/package.json:49` (`hd-transport-usb`) also needs the same bump.

</details>

packages/kit/src/hooks/useHardwareWalletConnectStatus.ts
```diff
@@ -50,11 +50,20 @@ async function fetchConnectedDevices(): Promise<Set<string>> {
...
   const refreshDevices = async () => {
+    // A USB event and a HardwareConnectionStateUpdate can land together and
+    // both fetch concurrently. Without a sequence guard the slower (older)
+    // response can overwrite the newer one and re-light a disconnected device.
+    requestSeqRef.current += 1;
+    const seq = requestSeqRef.current;
     const newDevices = await fetchConnectedDevices();
+    if (seq !== requestSeqRef.current) {
+      return;
+    }
```
sidmorizon, line 66: **Bug-fix retrospective record required by repo rules is missing**

Severity: `non-severe`

This bug fix changes the device-connection refresh logic (`refreshDevices` at `packages/kit/src/hooks/useHardwareWalletConnectStatus.ts:57-66`) without appending the required case record for the fix, so the repository's mandated retrospective log stays incomplete.
Impact: The team loses the required written record of this defect and its root cause.

<details>
<summary>Rule source and existing convention</summary>

`.cursor/rules/self-testing.mdc` (alwaysApply) states: "After every **bug fix**, append to `.claude/skills/1k-retrospective/references/case-studies.md`" with the Title / Date / Platforms / Symptom / Root Cause / Fix / Catchable-by template. That file exists and other recent fixes follow the convention (e.g. the OK-60543 and OK-59997 entries), but this PR adds no entry for OK-60486.

</details>

packages/shared/types/desktop.ts
```diff
@@ -37,8 +38,15 @@ export type INobleBleApi = {
...
   ) => () => void;
+  // `reason` tells a real peripheral drop apart from the main process
+  // reclaiming an idle link on its keep-alive timer. Optional: an older host
+  // bridge omits it, and consumers treat absent as a real drop.
   onDeviceDisconnected: (
-    callback: (device: { id: string; name: string }) => void,
+    callback: (device: {
+      id: string;
+      name: string;
+      reason?: EBleDisconnectReason;
```
sidmorizon, line 49: # 🟠 P1 desktop BLE 空闲保活回收会被当成真实断连

## 问题

`reason` 在本仓库没有任何消费方：全局搜索 `EBleDisconnectReason` 只有本文件和 `apps/desktop/app/preload.ts` 两处纯类型声明。SDK 侧的实际消费者 `@onekeyfe/hd-transport-web-device@1.2.0-alpha.155` 的 `electron-ble-transport.ts` 写得很明确：`` `reason` is carried for diagnostics only — behaviour is uniform ``，它对任何 reason 都无条件 `emitter.emit(TRANSPORT_EVENT.DEVICE_DISCONNECT, ...)`。本行上方注释所说的「区分真实掉线与主进程保活回收」在实现里并不存在。

更关键的是 alpha.150 → alpha.155 顺带改了断连订阅的作用域：150 里订阅建在 `acquire()` 内、`releaseLogical()` 通过 `cleanupDeviceState()` 摘掉它，所以保活回收观察不到；155 改成 transport 生命周期的 `subscribeHostDisconnects()` 后，主进程 `noble-ble-handler` 的 `BLE_IDLE_DISCONNECT_MS = 3 * 60_000` 定时器到点断链，会一路走到 hd-core 的 `DEVICE.DISCONNECT` → `ServiceHardware.untrackConnectedDevice()`。

## 影响

桌面端 BLE 硬件钱包在最后一次交互后约 `180s`（`keepSession` 路径为 `600s`）被主进程主动回收链路，钱包列表绿点随之熄灭，而设备其实还开着、还在范围内 —— 这正好是本 PR 想修的「拔掉设备绿点不消失」的反面误报。

而且 transport 层这条断连不经过 `DevicePool.reportDeviceChange()`，不会把设备移出 `DevicePool.current`，因此后续扫描的 `diff.connected` 里不会再出现这台设备，`DEVICE.CONNECT` 不一定会被重新发出，绿点未必能自行恢复。

## 建议

二选一，但需要显式确认：

1. 真的要区分：让 SDK 把 `reason` 带进 `TRANSPORT_EVENT.DEVICE_DISCONNECT` payload，`ServiceHardware` 的 `DEVICE.DISCONNECT` 分支对 `IdleKeepAlive` 不做 `untrackConnectedDevice()`；
2. 接受统一语义：把本行上方以及 `preload.ts` 的注释改成与实现一致（`reason` 仅用于诊断），并在 PR 描述里说明「BLE 空闲 3 分钟后绿点熄灭」是预期行为。

[Claude session ↗](https://claude.ai/code/session_014HeLzHYsH97ShCirxN34mY)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1808,6 +1809,13 @@ class ServiceHardware extends ServiceBase {
         const disconnectedIdentityKeys = this.untrackConnectedDevice(
           message.device,
         );
+        // The whole eviction path used to be silent, so a disconnect that
+        // never arrived and one that simply left no trace looked identical in
+        // collected logs (OK-60486).
+        serviceHardwareUtils.hardwareLog('device disconnected, untracked', {
+          connectId: message.device?.connectId,
```
sidmorizon: # 🟡 P2 新增日志把未脱敏的 connectId 写进持久化日志

## 问题

`serviceHardwareUtils.hardwareLog()` 最终落到 `defaultLogger.hardware.sdkLog.serviceEvent()`，而该方法带 `@LogToLocal({ level: 'info' })`（`packages/shared/src/logger/scopes/hardware/scenes/sdk.ts:113`），会写入本地日志文件并随用户反馈一起导出。

同一个工具文件里的 `serviceHardwareUtils.maskLogIdentifier()` 注释是硬性要求：`Device identifiers (serial numbers, connect ids) must never enter persisted logs in full`，`ServiceHardware.ts:1596` 也确实这么用了。本行直接写了原始 `message.device?.connectId`，绕过了这条约定。

## 影响

BLE 的 `connectId` 是设备 UUID/MAC，USB 的 `connectId` 是设备 path/序列号，都是稳定的设备标识符。一旦用户导出日志排查问题，这些完整标识符会随日志外流，且可用于跨会话、跨日志关联同一台硬件设备。

## 建议

复用现成的脱敏工具，保留后 4 位即可满足多设备关联排查的需要：

```ts
serviceHardwareUtils.hardwareLog('device disconnected, untracked', {
  connectId: serviceHardwareUtils.maskLogIdentifier(message.device?.connectId),
  removedIdentityKeyCount: disconnectedIdentityKeys.length,
});
```

[Claude session ↗](https://claude.ai/code/session_014HeLzHYsH97ShCirxN34mY)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -1808,6 +1809,13 @@ class ServiceHardware extends ServiceBase {
...
+          removedIdentityKeyCount: disconnectedIdentityKeys.length,
+        });
```
sidmorizon, line 1822: **Device identifier written unmasked into collected diagnostic logs when a hardware wallet disconnects**

Severity: `non-severe`

The device's full identifier is written into the persisted diagnostic log (`hardwareLog('device disconnected, untracked', ...)` at `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:1815-1818`) instead of the shortened form used everywhere else, so a user's device identifier ends up in logs that can be exported and uploaded.
Impact: Uploaded or exported log files expose a user's hardware device identifier in full.

<details><summary>Masking contract for persisted hardware logs</summary>

`serviceHardwareUtils.hardwareLog` forwards to `defaultLogger.hardware.sdkLog.serviceEvent`, i.e. a persisted logger (`packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts:8-23`). The same file documents the rule explicitly: "Device identifiers (serial numbers, connect ids) must never enter persisted logs in full; keep a short suffix for multi-device correlation" and provides `maskLogIdentifier` (`packages/kit-bg/src/services/ServiceHardware/serviceHardwareUtils.ts:25-34`). The sibling `DEVICE.STATE` handler complies (`packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts:1596-1599`), while the newly added disconnect log passes `message.device?.connectId` raw. On desktop USB the connectId is derived from the device serial number. CLAUDE.md's security section also forbids logging sensitive data.

</details>

Suggested fix:

```suggestion
        serviceHardwareUtils.hardwareLog('device disconnected, untracked', {
          connectId: serviceHardwareUtils.maskLogIdentifier(
            message.device?.connectId,
          ),
          removedIdentityKeyCount: disconnectedIdentityKeys.length,
        });
```

packages/shared/types/desktop.ts
originalix: 统一语义是刻意的设计选择，不是实现与注释脱节的疏漏 —— 采用你给的方案 2。

**为什么这么定**：绿点表达的是「当前有没有活的 BLE 链路」，不是「设备理论上在不在范围内」。主进程回收的链路同样是断开的链路，对使用者来说「设备连着但绿点灭了」和「设备断了但绿点亮着」一样是误报。这个问题在评审前讨论过，最初版本确实按 `IdleKeepAlive` 做了抑制，后来按上述语义改成一律上报（SDK 侧 commit `922cb4e29`，PR #891 内）。

**关于「绿点未必能自行恢复」—— 这一点不成立，已用日志证伪。**

恢复不依赖 `DEVICE.CONNECT`。`trackConnectedDevice()` 有两个调用点，另一个挂在 `DEVICE.SUPPORT_FEATURES` 上（`ServiceHardware.ts:1690` 注册，`:1707` 调用），而该事件在每次 SDK 调用的 `checkDeviceSupportFeature` 阶段都会触发，与 `DevicePool.current` 的 diff 结果无关。所以只要用户下一次操作设备，绿点就会重新点亮。

本机日志里有两次完整的熄灭→点亮循环：

```
21:24:46  device disconnected, untracked   {removedIdentityKeyCount:3}
23:09:24  features update                  ← 重新 track，绿点恢复
23:09:41  device disconnected, untracked
23:10:00  features update                  ← 重新 track，绿点恢复
```

实测重连约 1s（`Direct connect-by-id succeeded` → `Device ready for communication` 103ms，其余为配对与 PIN）。

**已改**：`packages/shared/types/desktop.ts` 与 `apps/desktop/app/preload.ts` 上方注释原本描述了一个实现里不存在的区分，这是你这条意见里确实站得住的部分，已改成与实现一致 —— `reason` 仅用于诊断（日志里可区分是自己回收还是设备掉线），不参与任何行为分支。PR 描述也已补上「BLE 空闲约 180s 后绿点熄灭、下次操作自动恢复」为预期行为。

另外补充一个背景：keep-alive 的 180s 与后台 portfolio sync 的租约窗口是对齐的（`DESKTOP_BLE_IDLE_DELAY_MS 30s + DESKTOP_BLE_REUSE_WINDOW_MS 150s`），而后台同步只蹭已有链路、其 `connect` 在该 scope 下是 no-op（`preload.ts:277`），所以不存在自动重连导致绿点反复闪烁的情况 —— 状态在最后一次交互后收敛一次。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
originalix: 成立，已按建议修复（`d70b9ae39`）。

```ts
serviceHardwareUtils.hardwareLog('device disconnected, untracked', {
  // Persisted logs ship with user feedback, so the identifier stays
  // masked; the suffix is enough to correlate a multi-device session.
  connectId: serviceHardwareUtils.maskLogIdentifier(
    message.device?.connectId,
  ),
  removedIdentityKeyCount: disconnectedIdentityKeys.length,
});
```

保留后 4 位仍足够做多设备关联排查，而这条日志的主要诊断价值本来就在 `removedIdentityKeyCount` —— 它区分「事件到了但没匹配到任何 key」和「确实驱逐了 N 个」，这正是本次定位 OK-60486 时缺失的信息。

originalix:

@originalix

Reviewed commit `93f70f9aa05c`. This report always reflects the latest reviewed changes.

### Review summary

This PR tightens the hardware-connectivity path by upgrading the internal `@onekeyfe/*` hardware SDK stack and extending the desktop BLE bridge so disconnect events can carry a disconnect reason into the transport layer. On the app side it also hardens the foreground connection-state hook against out-of-order refreshes and adds extra background logging around connection eviction so the UI can clear stale hardware-connected indicators more reliably. The scope stays focused on the desktop bridge, background hardware state ownership, and the UI-facing connected-device cache, with no broader architecture drift beyond the required lockfile updates.

**What needs attention:** Verify the hardware disconnect path clears UI state without persisting raw device identifiers into uploaded or shared logs.

### Issues to address

- **P2 · Disconnect telemetry logs raw hardware identifiers**
  - **Impact:** The new disconnect-path diagnostic writes `message.device?.connectId` into `hardwareLog`, even though `serviceHardwareUtils` explicitly documents that connect ids must be masked before they enter persisted logs. That turns a debugging improvement into an identifier-leak regression for hardware logs that may be retained or uploaded.
  - **Suggested action:** Mask the disconnect `connectId` with `serviceHardwareUtils.maskLogIdentifier(...)`, or remove the raw identifier and keep only non-identifying counters in the diagnostic payload.

### Validation gaps

- Targeted Jest validation did not run successfully in this workspace. `npx jest packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts packages/kit/src/hooks/useHardwareWalletConnectStatusUtils.test.ts --runInBand` failed immediately with a Jest setup resolution error for `./node_modules/react-native-gesture-handler/jestSetup.js`.
- I did not have an end-to-end desktop BLE environment here to exercise the new disconnect-reason bridge against real device disconnects.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
originalix: Already fixed in `d70b9ae390`, which landed after the `93f70f9aa` revision this comment was raised against. Same finding as the thread on `packages/shared/types/desktop.ts`, resolved there too.

Current code matches the suggestion exactly:

```ts
serviceHardwareUtils.hardwareLog('device disconnected, untracked', {
  // Persisted logs ship with user feedback, so the identifier stays
  // masked; the suffix is enough to correlate a multi-device session.
  connectId: serviceHardwareUtils.maskLogIdentifier(
    message.device?.connectId,
  ),
  removedIdentityKeyCount: disconnectedIdentityKeys.length,
});
```

The diagnostic value of this line is mostly in `removedIdentityKeyCount` anyway — it separates "the event arrived but matched nothing" from "N keys were actually evicted", which is exactly the signal that was missing while tracking down OK-60486.

wabicai (approved):

PR #12948, opened 2026-08-20
fix: prevent iOS address list liquid glass blur

originalix:
## Summary

- Hide the iOS 26 top scroll-edge effect for the Account address modal only.
- Reapply the edge-effect configuration after the asynchronously mounted address list becomes available.
- Keep Android, iOS versions below 26, and all other routes on their existing behavior.

## Intent & Context

Repeatedly opening the Account address modal from the Home copy-address action could cause the first network row to acquire an unintended Liquid Glass blur on iOS 26. The fix must remain narrowly scoped to this modal and must not affect the separate WebView navigation change in the same local workspace.

## Root Cause

The screen option was applied while the modal was rendering its loading state. The real FlashList-backed `RCTEnhancedScrollView` mounted later, after `react-native-screens` had already attempted to apply the edge-effect configuration. Its existing first-descendant-chain lookup also did not reach the list through the OneKey page wrappers, leaving the final scroll view with its top edge effect visible. Repeated modal presentations eventually exposed that system blur over the first row.

## Design Decisions

- Opt in from `WalletAddressModalRouter` only on iOS 26 or later.
- Preserve the existing fast first-descendant lookup for all screens, and use a bounded subtree fallback only for a screen that explicitly hides its top edge effect.
- Retry only while that explicitly opted-in screen is attached and its content scroll view has not mounted, with generation cancellation and a fixed attempt limit.
- Exclude nested screens and header configuration views from the fallback search to avoid adopting another screen's scroll view.

## Changes Detail

- `packages/kit/src/views/WalletAddress/router/index.ts`: sets `scrollEdgeEffects.top` to `hidden` only for the Wallet address route on iOS 26+.
- `patches/react-native-screens+4.23.0.patch`: adds the bounded scroll-view lookup and lifecycle reapplication needed for asynchronously mounted content.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Mobile iOS 26+
- **Risk Areas**: Native screen attachment timing and scroll-view discovery. The retry and fallback paths remain dormant unless a screen explicitly requests a hidden top edge effect.
- **Runtime Scope**: Main UI runtime only. The background JS runtime does not own this UIKit view hierarchy.

## Test plan

- [x] Built the signed Debug iOS app for an iPhone 17 Pro simulator running iOS 26.0.
- [x] Opened and closed the Account address modal 10 consecutive times; the first network row remained clear.
- [x] Inspected the final native list scroll view and confirmed its top edge effect was hidden.
- [x] Applied the regenerated `react-native-screens` patch to a pristine 4.23.0 package and matched the resulting native sources.
- [x] Passed focused formatting and type-aware Oxlint checks for the route change.
- [x] Passed `agent:check` worktree and staged lint checks.
- [ ] `agent:check` TypeScript check is blocked by four pre-existing base errors in unrelated hardware SDK and Electron updater files.

### DIFF
--- a/patches/react-native-screens+4.23.0.patch
+++ b/patches/react-native-screens+4.23.0.patch
@@ -37,10 +37,18 @@ index 7055a0d..81ee02b 100644
      }
  
 diff --git a/node_modules/react-native-screens/ios/RNSScreen.mm b/node_modules/react-native-screens/ios/RNSScreen.mm
-index 1c44762..3e8df39 100644
+index 1c44762..2771e08 100644
 --- a/node_modules/react-native-screens/ios/RNSScreen.mm
 +++ b/node_modules/react-native-screens/ios/RNSScreen.mm
-@@ -49,6 +49,25 @@
+@@ -9,6 +9,7 @@
+ #ifdef RCT_NEW_ARCH_ENABLED
+ #import <React/RCTConversions.h>
+ #import <React/RCTFabricComponentsPlugins.h>
++#import <React/RCTMountingTransactionObserving.h>
+ #import <React/RCTRootComponentView.h>
+ #import <React/RCTScrollViewComponentView.h>
+ #import <React/RCTSurfaceTouchHandler.h>
+@@ -49,6 +50,20 @@
  constexpr NSInteger SHEET_FIT_TO_CONTENTS = -1;
  constexpr NSInteger SHEET_LARGEST_UNDIMMED_DETENT_NONE = -1;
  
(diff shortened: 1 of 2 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

patches/react-native-screens+4.23.0.patch
```diff
@@ -57,11 +57,154 @@ index 1c44762..582527c 100644
...
+@@ -766,6 +792,66 @@ - (void)didMoveToWindow
+   } else {
+     [_touchHandler detachFromView:self];
+   }
++
++  // OneKey patch: only the iOS 26 screens that explicitly hide the top edge effect need a re-apply after attachment.
++#if RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)
++  if (@available(iOS 26.0, *)) {
++    if (self.window != nil && self.topScrollEdgeEffect == RNSScrollEdgeEffectHidden) {
++      [self updateContentScrollViewEdgeEffectsIfExists];
```
sidmorizon: # [🟡 P2] 三处 hook 都重复执行了一遍完整查找与应用

## 问题

`didMoveToWindow`、`finalizeUpdates`、`didSetProps` 三处都写成了 `updateContentScrollViewEdgeEffectsIfExists` + `scheduleHiddenTopScrollEdgeEffectApplication` 的组合，但这两步对 `topScrollEdgeEffect == RNSScrollEdgeEffectHidden` 的 screen 来说做的是同一件事。

`scheduleHiddenTopScrollEdgeEffectApplication` 只是 `generation += 1` 之后**同步**调用 `applyHiddenTopScrollEdgeEffectForGeneration:remainingAttempts:`，而它的第一次尝试同样是 `[self contentScrollViewForEdgeEffects]` 加 `[RNSScrollEdgeEffectApplicator applyToScrollView:withProvider:]` —— 和前一行 `updateContentScrollViewEdgeEffectsIfExists` 的实现完全一致。前一行不可能在后一行失败的地方成功，也不会得到不同的结果。

## 影响

对 `hidden` screen 而言，`contentScrollViewForEdgeEffects` 在首链查找返回 `nil` 后还会走 `findFirstScrollViewInSubtreeOf:` 的全子树 DFS（最多 `250` 次 `isKindOfClass:` 访问加数组进出栈）。这样每次 hook 触发都会**多跑一整轮子树遍历**，外加 4 次多余的 `UIScrollEdgeEffect` 属性写入。

`didSetProps` / `finalizeUpdates` 在四个 edge-effect prop 中任意一个变化时都会进来，`didMoveToWindow` 在祖先视图挂到 window 时也会对整棵子树逐级派发，所以这不是一次性开销。更重要的是，"立即执行一次 + 再排一条重试链"这种双路径写法会让后续维护者误以为两者职责不同，而这是一个每次升级 react-native-screens 都要手工 rebase 的 patch 文件。

## 建议

让 `scheduleHiddenTopScrollEdgeEffectApplication` 成为 `hidden` 场景的唯一入口：`didMoveToWindow` 新增分支里直接删掉 `updateContentScrollViewEdgeEffectsIfExists`；`finalizeUpdates` / `didSetProps` 里保留原有的 `updateContentScrollViewEdgeEffectsIfExists` 只服务非 `hidden` 分支，`hidden` 时只调 schedule。这样三处的语义统一为"非 hidden 走原逻辑，hidden 走重试链"，也省掉重复的子树扫描。

[Claude session ↗](https://claude.ai/code/session_01Bs26iziMsa12dhtFHhvrTR)

---

originalix:

@originalix

Reviewed commit `fe51c3665d37`. This report always reflects the latest reviewed changes.

### Review summary

This change adds an iOS 26-only WalletAddress navigation option for hiding the top scroll-edge effect and extends the patched react-native-screens native layer to discover content scroll views after Fabric mounting. The patch also changes shared iOS header appearance defaults, native stack retry behavior, SwiftUI controller containment, Android fragment removal, and transition diagnostics, so the effective system boundary is broader than the route-level feature. The main-runtime UI path is affected; no background runtime, wallet storage, or signing boundary is changed, but native window-attachment timing and shared navigation behavior now require regression validation.

**What needs attention:** Confirm that the iOS 26 visual behavior is explicitly scoped to WalletAddress and that detached native stacks and SwiftUI-hosted controllers always recover without stale or partial navigation state.

### Issues to address

- **P1 · Global iOS 26 header default changed instead of route-scoped opt-in**
  - **Impact:** The patched header builder uses configureWithDefaultBackground whenever an iOS 26 header has no explicit backgroundColor. That changes the shared default for unrelated screens, so a WalletAddress visual fix can silently make other navigation bars translucent or alter contrast and layout assumptions.
  - **Suggested action:** Keep the react-native-screens default branch unchanged and expose the Liquid Glass behavior through an explicit header or screen option enabled only by the target route.
- **P1 · Pending stack updates can get stuck permanently after retry exhaustion**
  - **Impact:** After 50 detached-window retries, the code leaves _pendingContainerUpdate set but refuses to schedule more work. Later didMoveToWindow, scene activation, and reattachment callbacks hit the exhausted counter, so a deferred push or modal update can remain unapplied indefinitely.
  - **Suggested action:** Reset or re-arm the retry budget on every recovery signal, or make the ready path flush a pending update regardless of the previous retry count.
- **P2 · SwiftUI containment fallback still calls didMove after an incomplete attach**
  - **Impact:** If addChildViewController succeeds and addSubview fails, or if addChildViewController itself throws, the catch path suppresses the hierarchy exception and execution still calls didMoveToParentViewController:. This can leave a controller partially contained or invoke the lifecycle callback with a nil or wrong parent during the crash-recovery path.
  - **Suggested action:** Track whether both containment and view attachment succeeded, call didMove only after both succeed, and remove the child from its parent when attachment fails.
- **P2 · Navigation diagnostics are emitted unconditionally in production**
  - **Impact:** RNSScreenStackLog is called from frequent layout, transition, and push/modal paths and always writes to the native logger plus NSLog. This creates production log volume and overhead in shared navigation code, and the logged UIView description can expose more UI structure than needed for a temporary diagnostic.
  - **Suggested action:** Remove the temporary diagnostics before shipping or guard the helper behind a debug-only build flag while retaining only bounded, non-sensitive counters when production telemetry is explicitly required.

### Validation gaps

- No iOS 26 run covered WalletAddress scroll-edge rendering, default headers on unrelated screens, or scroll-view replacement during loading.
- No detached-tab and scene-activation scenario verified that pending push/modal updates replay after the retry budget is exhausted.
- No SwiftUI bottom-tabs/split-view/stack containment run verified rollback after `UIViewControllerHierarchyInconsistency`.
- `git diff --check` reports whitespace in the patch file; this is not the primary correctness concern but should be cleaned before regeneration.

huhuanming (approved):

originalix (commented):
@originalix

Reviewed commit `fe51c3665d37`.

- **P1 · Global iOS 26 header default changed instead of route-scoped opt-in**
- **P1 · Pending stack updates can get stuck permanently after retry exhaustion**
- **P2 · SwiftUI containment fallback still calls didMove after an incomplete attach**
- **P2 · Navigation diagnostics are emitted unconditionally in production**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12948#issuecomment-5353955456)

PR #12950, opened 2026-08-20
fix: align Pro2 firmware and portfolio behavior

wabicai:
## Summary

- Surface the USB-priority hardware error during firmware update instead of treating it as a generic disconnect.
- Exclude zero-price and zero-value tokens from the Portfolio `Other` asset count while preserving valued tokens omitted from the hardware top-five list.
- Upgrade the complete Hardware SDK package set to `1.2.0-alpha.170` and refresh the lockfile.

## Hardware SDK integration

This PR consumes [OneKeyHQ/hardware-js-sdk#894](https://github.com/OneKeyHQ/hardware-js-sdk/pull/894) through `1.2.0-alpha.170`:

- Ignore stale Protocol V2 `FAILED` records until the corresponding target becomes live in the current install.
- Wait for `DeviceFirmwareUpdateRequest` to return `Success` before switching from `ConfirmOnDevice` to `installingFirmware`.
- Propagate device cancellation instead of treating a raw `Failure_ActionCancelled` response as a successful install trigger.
- Re-scan/re-enumerate, reacquire, and reconnect the expected BLE or USB device after an expected install-time transport release.
- Bypass generic Ping/Initialize only for install recovery and resume `DeviceFirmwareUpdateStatusGet` polling directly.
- Add Pro2 I8/LZ4 wallpaper compression and filter macOS metadata/non-`.okpkg` resource ZIP entries.

All direct Hardware SDK dependencies, CLI dependencies, transitive transport packages, and `yarn.lock` entries resolve to `1.2.0-alpha.170`; no `1.2.0-alpha.168` Hardware SDK references remain.

## User-visible behavior

- Pressing Cancel on the device no longer leaves the App in the installing state.
- The confirmation UI remains active until the device actually confirms the install request.
- Installation progress begins only after the confirmed response and then follows firmware status polling.
- Expected BLE transport release during installation enters the SDK recovery path instead of relying on a Ping that the install screen may not answer.
- USB-priority firmware errors retain their specific error instead of being collapsed into a generic disconnect.

## Compatibility and known limitation

- App-facing firmware update APIs are unchanged.
- The reconnect/polling behavior is implemented in the SDK and requires no App compatibility layer.
- The BLE pairing limitation during Bluetooth firmware installation is device-firmware behavior and still requires a Bluetooth firmware fix.

## Hardware validation

- Re-tested the Pro2 firmware flow on a physical iOS device.
- Verified the native background bundle contains the new typed install-confirmation call.
- The updated SDK behavior was accepted before publishing and pinning `1.2.0-alpha.170`.

## Test plan

- [x] `yarn install --mode=update-lockfile` resolves the Hardware SDK package set to `1.2.0-alpha.170`
- [x] `yarn install --immutable --mode=skip-build` validates the regenerated lockfile
- [x] Verify package manifests and `yarn.lock` contain no stale `.168` Hardware SDK references
- [x] `git diff --check -- package.json apps/cli/package.json yarn.lock`
- [x] Real-device iOS firmware flow verification
- [x] Linked SDK focused tests: 287 core tests and 12 WebUSB tests passed
- [ ] `yarn test packages/shared/src/utils/portfolioPayload.test.ts`
- [ ] Verify firmware update surfaces the USB-priority error instead of treating it as a generic disconnect on desktop


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix:

@wabicai

Reviewed commit `cf97fc486597`. This report always reflects the latest reviewed changes.

### Review summary

The firmware-update entry path now allows the specific USB-priority hardware error to continue into the firmware modal so the existing localized error view can render, while unrelated preflight failures still abort navigation. The portfolio payload path now subtracts zero-value tokens from the remainder count before serializing the hardware-facing summary, with matching test updates. The scope is narrow and does not alter secret handling or package ownership, but count eligibility is not fully aligned with the later payload-token filtering for incomplete fiat data.

**What needs attention:** Confirm that every token retained by the Home snapshot has a valid fiat record before computing the hardware portfolio remainder count, especially during partial aggregation and default-token retention.

### Issues to address

- **P2 · Invalid fiat entries still inflate otherTokens.count**
  - **Impact:** The new excluded-token count only recognizes finite non-positive price or fiat values. A retained token with missing or non-finite fiat data is later dropped by convertFiatStrictToDisplayCurrency and never appears in payload.tokens, but it is not subtracted from transferableTokenCount, so the hardware payload can advertise more remainder assets than actually exist.
  - **Suggested action:** Use the same eligibility rule for the count and payload selection, treating missing or non-finite fiat data as excluded (or deriving the remainder count from the filtered candidate set). Add a regression test for a retained default/custom token whose fiat record is unavailable.

### Validation gaps

- `git diff --check` completed successfully.
- `yarn test packages/shared/src/utils/portfolioPayload.test.ts --runInBand` could not start because the workspace has no Yarn `node_modules` state file; install/bootstrap validation is still required.

originalix (commented):
@wabicai

Reviewed commit `80ce0c1af249`.

- **P2 · Unusable fiat entries still inflate otherTokens.count**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/12950#issuecomment-5354752467)

zhaono1 (dismissed):

originalix (approved):

PR #12957, opened 2026-08-21
fix: stop perps order books overwriting each other's tick option (OK-59102, OK-59100)

Minnzen:
[OK-59102](https://onekeyhq.atlassian.net/browse/OK-59102) [OK-59100](https://onekeyhq.atlassian.net/browse/OK-59100)

----------

## Summary
- Persist the order book tick option only to **seed** a symbol that has none, instead of writing back a derived correction whenever it differs from what is stored.
- Stop a tick-option list derived without `szDecimals` from being pinned as the cached one, so the real reference list can supersede it.
- Drop the synchronous `flushNow()` from the tick option cache write path.
- Remove the temporary OK-59100 / OK-59102 field diagnostics.

## Intent & Context
QA reported two iOS bugs that always appeared and disappeared together: the order book flickering as a whole block (OK-59102), and the chart page scrolling without following the finger while the two order buttons stopped responding (OK-59100). Repro: Perps home → BTC → chart page → back → switch to another coin → chart page.

Earlier attempts on this branch treated them as separate rendering defects and did not hold up. This change comes from instrumenting the actual write path and catching the loop in a reproduction.

## Root Cause
Two `PerpOrderBook` instances are alive whenever the chart page is pushed over the Perps tab (tab root + chart page). This is by design and is not a leak — an instance census confirmed the count peaks at 2 with correct mount/unmount pairing.

Each instance derived its tick-option list from **its own** book snapshot and wrote the result back into the **shared** persisted atom whenever it differed (`shouldPersistOrderBookTickOption` had no convergence condition). The two lists disagreed, so they overwrote each other indefinitely. One captured reproduction: **850 tick-option writes and 839 global L2 resubscribes**, ending in `Maximum update depth exceeded`.

Why the lists diverged: `changeActiveAsset` publishes an optimistic instrument with `universe: undefined` and hydrates a cached book behind it, so an order book can render a real price while `szDecimals` is still missing. Without `szDecimals`, `buildReferenceTickOptions` bails and the fallback builder runs with a different multiplier set — labelling the same `0.1` tick as `nSigFigs: 5` instead of `null`. That list was cached, and the reference list could not displace it because the hysteresis guard kept the cached one on **equal** `priceDecimals`.

Why both symptoms: each resubscribe made the incoming book fail `isL2BookForTarget` on whichever instance disagreed, dropping it to cold cache — that is the whole-book flicker. The render churn saturated the UI thread that `HeaderScrollGestureWrapper` drives `scrollTo` on (its pan callbacks are worklets, so this is not JS-thread starvation), and the synchronous per-write MMKV flush starved the JS thread the order buttons' `onPress` runs on.

## Design Decisions
- **Seed-only, not "seed then repair".** A repair branch was considered and rejected: it fires exactly in the state where an instance's list cannot resolve the shared entry, and it does not even cover the more common drift case where `byValue` matches but the `nSigFigs` meaning changed. Seed-only is safe by construction — writes happen only on the `undefined → defined` transition — rather than safe by enumeration of today's list shapes. Explicit user selection keeps its own unconditional path.
- **Authoritative seed check in the action.** The hook's existence check reads render-time state, so two instances can both pass it. `setOrderBookTickOption` re-checks against the live atom via `get()`.
- **`flushNow()` removal.** It is the app-background path; forcing it per write made every write a synchronous full-store read, merge, and MMKV rewrite. `simpleDb` holds the durable copy, so this cache only needs to be eventually consistent.

## Changes Detail
- `OrderBook/useTickOptions.ts` — seed-only persist, gated on stored preferences having loaded and on a list built with `szDecimals`; fallback-derived lists tagged `isFallback` so a reference list supersedes them and they never overwrite one.
- `states/jotai/contexts/hyperliquid/actions.ts` — `source: 'seed'` resolved against the live atom; the persisted map is built from the merged view so an early seed cannot drop symbols the module cache still holds.
- `shared/src/utils/perpsOrderBookTickOptionsCache.ts` — debounced write instead of `flushNow()`.
- `OrderBook/index.tsx`, `PerpOrderBook.tsx`, `PerpsGlobalEffects.tsx`, `PerpMarketFooter.tsx`, `MobilePerpMarket.tsx`, `TradingViewPerpsV2.tsx`, `utils/perpsFieldDiagnostics.ts` — field diagnostics removed.

Also retained from earlier on this branch (independent defects found while investigating, not the root cause): horizontal order book depth bars and ladder rows reading one snapshot, mobile skipping `useRafCoalesced`, and the TradingView interaction-overlay chart-ready gating.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (iOS verified), and the shared Perps order book path on all platforms
- **Risk Areas**: tick-option selection is what the L2 subscription parameters are derived from. The behavioural change is that an order book no longer rewrites a stored tick option it disagrees with — whichever value is persisted first now wins until the user changes it explicitly.

## Test plan
Verified on iOS simulator (iPhone 16 Pro / iOS 18.6) against instrumented builds:
- [x] Repro path (BTC → chart → back → switch coin → chart): tick-option writes **850 → 0**, global resubscribes **839 → 3** (all genuine coin changes), no `Maximum update depth exceeded`.
- [x] Seed path exercised on coins with no stored entry: exactly one write per symbol, every write over an empty entry, and the seeded list came from the reference builder (`0.01:null`, not `0.01:5`).
- [x] The optimistic `universe: undefined` window is skipped rather than seeded from — confirmed in the log as `szDecimals` missing → no seed, then `szDecimals` present → single seed.
- [x] Order book instance count peaks at 2 with matched mount/unmount.
- [x] App boots clean after diagnostics removal.
- [ ] Real device pass (bundle self-test + QA).

## Rebase note
Developed on `release/v6.5.0` (PR #12930, now closed) and re-applied onto `hotfix/v6.5.2`. The `OrderBook/` files and the tick-option cache were untouched on 6.5.2 and were taken directly; `actions.ts` and `TradingViewPerpsV2.tsx` had their own 6.5.2 changes, so only this fix's delta was three-way applied on top of them — verified the net diff against 6.5.2 is exactly this fix and nothing else.

Pre-existing on this base and unrelated to this PR: `yarn tsc:staged` reports `isInstallerPath` missing on `AppUpdater` in `DesktopApiAppUpdate.ts:832`. Confirmed by running the same check on a clean `hotfix/v6.5.2` worktree with these changes stashed.

[OK-59102]: https://onekeyhq.atlassian.net/browse/OK-59102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-59100]: https://onekeyhq.atlassian.net/browse/OK-59100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
+++ b/packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useMemo, useRef } from 'react';
+import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 
 import BigNumber from 'bignumber.js';
 
@@ -12,7 +12,6 @@ import {
   resolveOrderBookSizeDecimals,
 } from '@onekeyhq/shared/src/utils/perpsUtils';
 import type { IBookLevel } from '@onekeyhq/shared/types/hyperliquid/sdk';
-import type { IPerpOrderBookTickOptionPersist } from '@onekeyhq/shared/types/hyperliquid/types';
 
 import {
   type ITickParam,
@@ -20,7 +19,7 @@ import {
   buildTickOptions,
   getDefaultTickOption,
   getTickOptionsDataDuringTransition,
-  shouldPersistOrderBookTickOption,
+  shouldSeedOrderBookTickOption,
 } from './tickSizeUtils';
 
 interface ITickOptionsResult {
@@ -63,6 +62,7 @@ export function useTickOptions({
     tickOptions: ITickParam[];
     defaultTickOption: ITickParam;
     priceDecimals: number;
+    isFallback?: boolean;
   } | null>(null);
 
   const [persistedTickOptions] = useOrderBookTickOptionsAtom();
@@ -72,8 +72,22 @@ export function useTickOptions({
   );
   const actions = useHyperliquidActions();
 
+  // Seeding makes the first write authoritative, so a seed that lands before
+  // the stored preferences finish loading would permanently replace the user's
+  // choice rather than transiently shadow it.
+  const [hasLoadedPersistedTickOptions, setHasLoadedPersistedTickOptions] =
+    useState(false);
   useEffect(() => {
-    void actions.current.ensureOrderBookTickOptionsLoaded();
+    let cancelled = false;
+    void (async () => {
+      await actions.current.ensureOrderBookTickOptionsLoaded();
+      if (!cancelled) {
+        setHasLoadedPersistedTickOptions(true);
+      }
+    })();
+    return () => {
+      cancelled = true;
+    };
   }, [actions]);
 
   const topBidPrice = bids[0]?.px;
@@ -117,6 +131,7 @@ export function useTickOptions({
     if (marketTickOptionsData) {
       if (
         cached &&
+        !cached.isFallback &&
         marketTickOptionsData.priceDecimals <= cached.priceDecimals
       ) {
         return cached;
@@ -127,7 +142,10 @@ export function useTickOptions({
 
     const priceDecimals = getDisplayPriceScaleDecimals(marketPrice);
 
-    if (cached && priceDecimals <= cached.priceDecimals) {
+    if (
+      cached &&
+      (!cached.isFallback || priceDecimals <= cached.priceDecimals)
+    ) {
       return cached;
     }
 
@@ -143,11 +161,15 @@ export function useTickOptions({
     const tickLabelDecimals =
       new BigNumber(defaultTickOption.label).decimalPlaces() ?? 0;
 
+    // Derived without szDecimals, so it uses a different multiplier set than
+    // buildReferenceTickOptions and can label the same tick with a different
+    // nSigFigs. Tagged so a later reference list can displace it (OK-59102).
     const result = {
       symbol,
       tickOptions,
       defaultTickOption,
       priceDecimals: Math.max(priceDecimals, tickLabelDecimals),
+      isFallback: true,
     };
 
     // Cache the result
@@ -212,28 +234,37 @@ export function useTickOptions({
   useEffect(() => {
     if (!symbol) return;
 
-    const persisted = persistedTickOptions[symbol];
-    const currentPersist: IPerpOrderBookTickOptionPersist = {
-      value: selectedTickOption.value,
-      nSigFigs: selectedTickOption.nSigFigs ?? null,
-      mantissa: selectedTickOption.mantissa ?? null,
-    };
-
+    // Seed-only writer. The persisted tick option is a user preference, so an
+    // order book adopts it and never writes a derived correction back: two
+    // books whose instance-local lists disagree on nSigFigs for the same value
+    // would otherwise overwrite each other without ever converging (OK-59102).
     if (
-      shouldPersistOrderBookTickOption({
+      !shouldSeedOrderBookTickOption({
         isReady: Boolean(tickOptionsData),
-        persisted,
-        next: currentPersist,
+        // Read off the data rather than inferred from szDecimals: the
+        // transition branch can hand back a cached fallback list after
+        // szDecimals has arrived.
+        isFallbackList: Boolean(tickOptionsData?.isFallback),
+        hasLoadedPersistedOptions: hasLoadedPersistedTickOptions,
+        hasPersistedOption: Boolean(persistedTickOptionsForRender[symbol]),
       })
     ) {
-      void actions.current.setOrderBookTickOption({
-        symbol,
-        option: currentPersist,
-      });
+      return;
     }
+
+    void actions.current.setOrderBookTickOption({
+      symbol,
+      option: {
+        value: selectedTickOption.value,
+        nSigFigs: selectedTickOption.nSigFigs ?? null,
+        mantissa: selectedTickOption.mantissa ?? null,
+      },
+      source: 'seed',
+    });
   }, [
     symbol,
-    persistedTickOptions,
+    hasLoadedPersistedTickOptions,
+    persistedTickOptionsForRender,
     selectedTickOption,
     tickOptionsData,
     actions,
--- a/packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
+++ b/packages/kit/src/components/TradingView/TradingViewPerpsV2/TradingViewPerpsV2.tsx
@@ -335,6 +335,30 @@ export function TradingViewPerpsV2(
   const isChartLinesReady = chartLinesReadyWebviewKey === _webviewKey;
   const isChartContentReady = chartContentReadyWebviewKey === _webviewKey;
 
+  // OK-59100: on iOS this flag gates both the header pan gesture and the tab
+  // scroller, and nothing ever reset it — a stray `open` arriving before the
+  // chart is live never receives its matching `close` and locks scrolling for
+  // the rest of the session (fail-closed). Accept `open` only once the chart
+  // reports ready; a reload invalidates any pending open state anyway.
+  const isChartContentReadyRef = useRef(isChartContentReady);
+  isChartContentReadyRef.current = isChartContentReady;
+  const guardedInteractionOverlayOpenChange = useCallback(
+    (isOpen: boolean) => {
+      if (isOpen && !isChartContentReadyRef.current) {
+        return;
+      }
+      onInteractionOverlayOpenChange?.(isOpen);
+    },
+    [onInteractionOverlayOpenChange],
+  );
+  // Any transition out of ready (reload, navigation, render-process loss)
+  // discards overlay state the chart can no longer close on its own.
+  useEffect(() => {
+    if (!isChartContentReady) {
+      onInteractionOverlayOpenChange?.(false);
+    }
+  }, [isChartContentReady, onInteractionOverlayOpenChange]);
+
   const prevWebviewKeyRef = useRef(_webviewKey);
   useEffect(() => {
     if (prevWebviewKeyRef.current !== _webviewKey) {
@@ -629,7 +653,7 @@ export function TradingViewPerpsV2(
     onOrderPriceUpdate,
     onChartOrderIntent,
     onTouchScroll,
-    onInteractionOverlayOpenChange,
+    onInteractionOverlayOpenChange: guardedInteractionOverlayOpenChange,
   });
 
   // Chart lines management (liquidation, position, orders)
--- a/packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
+++ b/packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts
@@ -1496,13 +1496,23 @@ class ContextJotaiActionsHyperliquid extends ContextJotaiActionsBase {
       payload: null | {
         symbol: string;
         option: IPerpOrderBookTickOptionPersist | null;
+        // 'seed' means "establish a default if this symbol has none"; the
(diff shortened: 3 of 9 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
```diff
@@ -212,28 +234,37 @@ export function useTickOptions({
...
     if (
-      shouldPersistOrderBookTickOption({
+      !shouldSeedOrderBookTickOption({
         isReady: Boolean(tickOptionsData),
-        persisted,
-        next: currentPersist,
+        // Read off the data rather than inferred from szDecimals: the
+        // transition branch can hand back a cached fallback list after
+        // szDecimals has arrived.
+        isFallbackList: Boolean(tickOptionsData?.isFallback),
```
sidmorizon, line 247: # [🟠 P1] 拒绝 fallback seed 会让订单簿在有实时数据时整体不渲染

## 问题

`shouldSeedOrderBookTickOption` 现在在 `isFallbackList` 为 `true` 时一律拒绝写入。但 `PerpOrderBook.tsx:531-588` 把"持久化 map 里是否存在该 coin"直接当成整个订单簿的渲染闸门：

```ts
const hasInitializedTickOption = Boolean(
  getPerpsOrderBookTickOptionWithCache({ coin, options: orderBookTickOptions }),
);
const visibleL2Book = hasInitializedTickOption ? candidateL2Book : null;
const hasRenderOrderBook = Boolean(visibleL2Book);
```

除用户手动选 tick 外，seed 是这个 map 从无到有的唯一入口。而 `szDecimals` 来自 `activeTradeInstrument.universe`：`changeActiveAsset` 先发 `universe: undefined` 的乐观 instrument（`actions.ts:1570-1572`），bg 侧 `getTradingUniverse()` 只读 `simpleDb`，全新安装/元数据尚未落库时返回 `[]`，`ServiceHyperliquid.changeActiveAsset` 随即 `perpsActiveAssetAtom.set({ universe: undefined })` 并提前 return，本身不做重试。

## 影响

在"首次打开某个 coin 且本地没有该 coin 持久化 tick"的窗口里，L2 数据（WS 实时或 `hydrateL2BookFromSwr` 冷缓存）已经进入 `candidateL2Book`，但 `hasInitializedTickOption` 为 `false`，桌面分支拿 `bids: []` / `asks: []` 渲染空表，移动竖版分支走 `loadingBody` 骨架。改动前 fallback 列表会 seed 一次把闸门打开（代价是 tick 标注可能不准），改动后变成完全不渲染。`universe` 只在下一次 `changeActiveAsset` 才可能补齐，冷启动元数据未就绪时这段空白可以一直持续到用户再次切币。PR 描述里验证的是"跳过乐观窗口不 seed"，但没有覆盖这个窗口同时也是订单簿渲染闸门这一耦合。

## 建议

把"渲染闸门"和"权威 tick 偏好"解耦，不要让 seed 同时承担两件事。可选方向之一：把 `hasInitializedTickOption` 改成"`tickOptionsData` 已解析出 `selectedTickOption`"，而不是"持久化 map 有条目"；或在 fallback 期允许一个只存在于内存、不写 `simpleDb` 的临时 tick 打开闸门，等 `szDecimals` 到达后再由 seed 写入权威值。

[Claude session ↗](https://claude.ai/code/session_01EKpFw6kwVAiPgzkniWAyWG)

---

originalix:

@Minnzen

Reviewed commit `aa62c5335080`. This report always reflects the latest reviewed changes.

### Review summary

This PR changes Perps order-book tick-option ownership in the main runtime from render-derived writeback to first-write seeding plus explicit user updates, with durable persistence delegated to the bg runtime. It also keeps rendered depth bars, rows, and tap targets on one visual snapshot, removes synchronous SWR-cache flushing, and gates TradingView interaction-overlay events on chart readiness. The design is directionally sound, but failed preference loads can still authorize destructive seeding, and chart readiness is not invalidated for every WebView restart path.

**What needs attention:** Confirm that main-runtime seeding only starts after a successful bg persistence read and that every WebView replacement lifecycle clears the iOS interaction lock before accepting new overlay events.

### Issues to address

- **P2 · Failed preference loads are treated as successful**
  - **Impact:** The main-runtime hook treats resolution of ensureOrderBookTickOptionsLoaded() as proof that bg persistence was read, but that action catches read failures and resolves while marking itself loaded. If the main-runtime atom and its per-runtime SWR JS copy have no entry, the seed effect can then send a derived default to bg simpleDb and replace the user's durable selection, causing an unexpected L2 aggregation change and resubscription.
  - **Suggested action:** Make the load action return or throw an explicit success result, keep hydration unconfirmed after failure, and permit seeding only after a successful authoritative read; add a test covering a rejected load followed by recovery.
- **P2 · Chart readiness survives untracked WebView restarts**
  - **Impact:** The new overlay guard depends on isChartContentReady becoming false, but readiness is currently cleared only for a WebView-key change or the network-restore branch. A load restart, navigation, or render-process recovery that reuses the key can therefore retain true readiness, accept an early overlay-open event from the replacement chart, and leave iOS header and tab scrolling disabled if its close event is lost.
  - **Suggested action:** Centralize WebView lifecycle invalidation and invoke it from load-start and process-termination/recovery callbacks so readiness keys and the parent overlay state are cleared before the replacement chart can emit events.

### Validation gaps

- The targeted Jest command for `tickSizeUtils.test.ts` and `useRafCoalesced.test.ts` could not start because the workspace lacks Yarn's `node_modules` state file.
- No runtime validation covered two simultaneous order books, rapid symbol changes, main/bg startup ordering, or iOS WebView initial-load and recovery behavior.
- Existing GitHub reviews, inline comments, general comments, and thread state could not be fetched because GitHub API connectivity failed, so inline wording should be checked before posting.

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
Minnzen: Thanks — the coupling you identified is real, but the severity does not hold on the cold-start path, so I am not changing this here.

`PerpsGlobalEffects.tsx:1073-1079` awaits the trading meta before the flow continues when nothing is cached:

```ts
if (!hasCachedTradingUniverse) {
  markPerpsColdStartPerf('initial_symbol_wait_trading_meta_no_cache');
  await refreshTradingMeta();
} else {
  deferTradingMetaRefresh(1200);
}
refreshSpotMeta(1800);
```

`refreshTradingMeta` persists through `simpleDb.perp.setTradingUniverse` (`ServiceHyperliquid.ts:1541`), and it is written once and kept. So on a fresh install the perp universe is resolved *before* `changeActiveAsset` publishes the instrument — `szDecimals` is present, the seed fires, and the render gate opens. The "universe stays undefined with no retry until the user switches coin again" premise does not apply to perp coins.

The residual window is narrower than described and specific to spot: `refreshSpotMeta(1800)` is deferred and not awaited, so `universe?.baseSzDecimals` can be missing for the first couple of seconds after a fresh install. Reaching a blank book needs a fresh install, the user switching to a spot coin with no stored tick inside that window, and L2 data already flowing. The default landing pair is a perp, so this is not on the first-run path.

Given that, I am not touching the render gate in this PR: the change would alter when the order book is allowed to render at all, which is a larger blast radius than the window it closes. The coupling itself is worth removing — gating on "a tick option has been resolved" rather than "the persisted map has an entry" is the right shape — and I have put it on the follow-up list rather than bundling it into a release-branch fix that is already in QA.

Resolving on that basis; happy to reopen if you think the spot window is reachable in practice.

packages/kit/src/views/Perp/components/OrderBook/useTickOptions.ts
```diff
@@ -212,28 +234,37 @@ export function useTickOptions({
...
+        hasLoadedPersistedOptions: hasLoadedPersistedTickOptions,
+        hasPersistedOption: Boolean(persistedTickOptionsForRender[symbol]),
       })
```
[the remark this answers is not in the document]
Minnzen: Correct that seed-only never re-converges this, and the display/subscription split is real: the selector resolves by `value` against the instance's current list while the subscription reads the persisted record, so a magnitude crossing can leave them on different `nSigFigs`.

Not fixing it here, deliberately. The fix would restore a corrective write, and that is the exact shape of the bug this PR removes — two order books deriving a tick option from their own snapshots and overwriting each other without converging (850 writes and 839 resubscribes in one reproduction, ending in `Maximum update depth exceeded`). Restricting the correction to reference-derived lists is the obvious guard, but two instances hold different book snapshots, and a magnitude boundary is precisely where their prices can land on opposite sides and produce different lists — so the guard is weakest exactly when the correction would fire. This PR already shipped one assumption of that kind that did not hold (`Number.isInteger(szDecimals)` was taken to imply a reference-derived list, and a cached fallback list slipped through it), so I would rather not add another on a release branch.

The safer direction is to remove the divergence instead of writing over it — have the subscription parameters derive from the same resolved option the selector displays, rather than from the raw persisted record. That is a change to the subscription path and does not belong in this fix.

User impact meanwhile: the grouping label can disagree with the actual aggregation after a 10x price move on a coin that already has a stored tick, and one tap on the tick selector rewrites it through `handleSelectTickOption` and restores agreement.

Tracking as follow-up. Resolving here.

Minnzen:
Both findings are accurate. Neither is being changed in this PR, and they are being tracked as one follow-up rather than two — they are the same design problem seen from two sides, and fixing either alone makes things worse.

**Failed preference loads are treated as successful** — confirmed at `actions.ts:1460-1472`: the read is wrapped in `try/catch` that logs and swallows, and `finally` sets `orderBookTickOptionsLoaded = true` regardless, so awaiting it is not proof the durable read succeeded. The `hasLoadedPersistedTickOptions` gate this PR adds was meant to prevent exactly the clobber you describe, and it does not achieve that.

Worth separating, though: the clobber itself is not introduced here. Before this change the persist effect wrote whenever the derived option differed from the persisted one, and a failed load leaves `persisted` undefined, so the same overwrite happened. What is new is an ineffective guard against it, not the hole.

**Chart readiness survives untracked WebView restarts** — the gap in readiness invalidation is real. It stays theoretical against the field data behind these tickets: across the instrumented reproductions the interaction overlay never latched, with all 18 `interactionOverlay.change` records reporting `isOpen: false`. The fix you suggest — centralising WebView lifecycle invalidation across load-start and process-recovery callbacks — is a change to the chart component's lifecycle handling, not to the guard this PR added.

**Why one follow-up and not two fixes here.** The persisted tick-option map is doing double duty: it is the user's preference *and* it is the order book's render gate (`PerpOrderBook.tsx:532` feeding `visibleL2Book` at `:590`). Seeding is the only thing that opens that gate. So tightening the load gate — refusing to seed until a read is confirmed authoritative — closes the clobber and simultaneously means a failed read leaves the order book blank for the session. That is strictly worse than today.

The coherent fix is to decouple the two roles in one change: gate rendering on "a tick option has been resolved" rather than "the persisted map has an entry", and make load failure explicit so seeding can wait for a confirmed read without holding the UI hostage. That is a self-contained refactor and does not belong in a release-branch fix that is already in QA.

Filed as follow-up; happy to take it as its own PR once this one lands.

originalix (approved):

PR #12952, opened 2026-08-20
fix: filter unmatched perps results for short search queries (OK-60751)

Minnzen:
[OK-60751](https://onekeyhq.atlassian.net/browse/OK-60751)

----------

## Summary

- Hold a short ASCII perps search query to a literal match on `name`, `subtitle` or an exact dex prefix, so `usdc` no longer lists 56 unrelated markets.
- Leave longer and non-ASCII queries to the endpoint's own ranking, so description hits (`nasdaq`, `bitcoin`) and localized aliases keep working.
- Additive change: 22 added lines in the service, nothing removed or restructured.

## Intent & Context

QA reported (OK-60751) that searching `usdc` in universal search fills the Perps section with markets that have nothing to do with the query — `CASHCAT`, `GRAM`, `AMAT`, `ARM`, `ASML` and so on — where the expected result is an empty Perps section.

The first suspicion was a regression from the HIP-3 work in 18e310d9 (OK-59307), which added `assetTypeVersion: 2` to this request. That turned out not to be the cause, so this PR adds an app side guard instead of reverting anything.

## Root Cause

Server side. `/wallet/v1/proxy/hyperliquid/perpsAsset` matches the query as a substring of the asset's collateral / description text, not only against the ticker or display name:

```
query=usdc&assetTypeVersion=2  -> 56 rows (perps 2 + xyz 38 + para 16)
query=usdc&assetTypeVersion=1  -> 40 rows (perps 2 + xyz 38)
query=usdc (no version param)  -> 40 rows
```

So the noise predates `assetTypeVersion: 2`; that param only widened the result set with the 16 `para` rows. Probing the endpoint confirms the substring behaviour, because every substring of `USDC` returns the same set while non substrings return nothing:

```
usdc -> 56    usd -> 56    sd -> 56    dc -> 57
usc  -> 0     udc -> 0
nasdaq -> 12 stocks    stock -> 24    shares -> 13
```

The app simply rendered what the endpoint returned: `universalSearchOfPerp` only filtered *out* delisted assets, and the UI does no relevance filtering of its own.

## Design Decisions

- **Substring match, not fuzzy scoring.** The endpoint already ranks results; the client only needs to reject rows that do not contain the query at all. A scorer would be more code and a second, competing notion of relevance.
[...]

PR #12962, opened 2026-08-21
fix: filter unmatched perps results for short search queries (OK-60751)

Minnzen:
[OK-60751](https://onekeyhq.atlassian.net/browse/OK-60751)

----------

## Summary

- Hold a short ASCII perps search query to a literal match on the row's `name` or `subtitle`, an exact dex prefix, or the market's server search aliases minus their pair notations — so `usdc` no longer lists 56 unrelated markets.
- Leave longer and non-ASCII queries to the endpoint's own ranking, so description hits (`nasdaq`, `bitcoin`) and localized aliases keep working.
- Additive change: nothing removed or restructured.

## Intent & Context

QA reported (OK-60751) that searching `usdc` in universal search fills the Perps section with markets that have nothing to do with the query — `CASHCAT`, `GRAM`, `AMAT`, `ARM`, `ASML` and so on — where the expected result is an empty Perps section.

The first suspicion was a regression from the HIP-3 work in 18e310d9 (OK-59307), which added `assetTypeVersion: 2` to this request. That turned out not to be the cause, so this PR adds an app side guard instead of reverting anything.

## Root Cause

Server side, and the mechanism is the per market alias list in `/utility/v1/perp-config` — the same map the perps token selector already searches with `findTokensByAlias`. Every market carries its pair notations as aliases:

```
"para:AAOI": { aliases: ["aaoi", "aaoi-usdc", "aaoi/usdc", "aaoi-usd", "aaoi/usd", "para:aaoi", ...] }
"BTC":       { aliases: ["btc", "bitcoin", "satoshi", "xbt", ...] }            <- no pair form
```

216 of the 3394 aliases in the live map are pair notations, and they sit exactly on the 54 stock rows. A `usdc` query matches all of them through text that never appears on screen, which is why the section fills with `AMAT` / `ARM` / `ASML`:

```
query=usdc&assetTypeVersion=2  -> 56 rows (perps 2 + xyz 38 + para 16)
query=usdc&assetTypeVersion=1  -> 40 rows (perps 2 + xyz 38)
query=usdc (no version param)  -> 40 rows
```

So the noise predates `assetTypeVersion: 2`; that param only widened the result set with the 16 `para` rows. Every substring of `USDC` returns the same set while non substrings return nothing, consistent with a substring match over that alias text:

```
usdc -> 56    usd -> 56    sd -> 56    dc -> 57
usc  -> 0     udc -> 0
nasdaq -> 12 stocks    stock -> 24    shares -> 13
```

The app simply rendered what the endpoint returned: `universalSearchOfPerp` only filtered *out* delisted assets, and the UI does no relevance filtering of its own.

## Design Decisions

- **Substring match, not fuzzy scoring.** The endpoint already ranks results; the client only needs to reject rows that match nothing a user can see. A scorer would be more code and a second, competing notion of relevance.
- **Match `name` / `subtitle` by substring, the dex by exact `asset.type`.** Matching against the dex prefixed coin (`xyz:AMAT`) instead would let a fragment such as `ar`, `pa` or `xy` wave through every row of that sub-dex — the exact noise this PR removes, on the most common short queries. An exact `asset.type` comparison still lets `xyz` or `para` browse a sub-dex.
- **Match the alias map too, minus the pair notations.** Part of the alias list is invisible on the row (`satoshi` and `digital gold` on BTC, `sandisk` on SNDK), so a ticker/subtitle-only rule silently killed those queries. Matching the list wholesale is not an option either: every stock row has an `*-usdc` alias, so `usdc` would go straight back to 56 rows. Skipping aliases that end in `-usd(c)` / `/usd(c)` holds both ends, and reuses the `findTokensByAlias` prefix rule for 1–2 character queries. The map is read only when the filter can drop something, and an absent or empty map degrades to the previous ticker/subtitle behaviour.
- **Only enforce it for ASCII queries of 4 characters or fewer.** Description search is a deliberate server feature and is valuable for longer queries: replaying live responses through an always-on literal filter would have dropped every result for `nasdaq` (12), `stock` (24) and `shares` (13), plus semantic hits such as `japan` -> `KIOXIA` / `SOFTBANK`. The ASCII guard matters because the endpoint answers localized aliases the cached map may not carry — on wallet.onekeytest.com the Chinese words for bitcoin, Japan and ethereum return BTC, EWJ and ETH — and those queries are 2–3 characters, so a length-only gate would have dropped all of them. Nothing in the client restricts the charset: the UI only trims the input and `universalSearch` passes it through verbatim.
- **Fix stays in the app.** The endpoint's relevance should still be fixed server side; this is a client side guard so 6.5.2 does not ship the broken list.

## Changes Detail

- `packages/shared/src/utils/perpsUtils.ts` — add `matchesTokenSearchAlias()`, a single market mirror of `findTokensByAlias` that skips the `*-usd(c)` pair notations.
- `packages/kit-bg/src/services/ServiceUniversalSearch.ts` — add `PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH`, read the cached alias map when the filter can drop rows, and add an early `return false` inside the existing filter in `universalSearchOfPerpCached`.
- `packages/kit-bg/src/services/ServiceUniversalSearch.perpsAssetCapability.test.ts` — one fixture plus a table of cases: `usdc` drops everything, `btc` / `nvda` keep their own row, `gold` survives on an alias alone, `xyz` browses the sub-dex, `ar` does not, `etf` survives on the subtitle alone, and `japan` / a Chinese alias stay on the endpoint's ranking. Both halves mutation checked: dropping the alias clause fails `gold`, dropping the pair-notation skip fails `usdc`.

## Trade-offs

Short queries still lose some genuinely semantic hits — `sol` -> `Bloom Energy` / `Nokia` survive only because their alias lists carry `sol...`, while rows outside the curated map are dropped. That is the price of removing the `usdc` / `usd` noise.

Replaying live endpoint responses through the final filter:

```
usdc  56 -> 0     usd  56 -> 1     sd  56 -> 0     dc  57 -> 4     ar  57 -> 22
btc    3 -> 3     eth   5 -> 5     sol  5 -> 5     ai  30 -> 21
sato   1 -> 1     xbt   2 -> 2     disk 3 -> 3     gold 1 -> 1     nvda 1 -> 1
etf   12 -> 12    xyz  46 -> 44    para 17 -> 17   japan 4 -> 4    bitcoin 6 -> 6
```

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Extension / Mobile / Desktop / Web (background runtime, all platforms)
- **Runtime scope**: bg runtime only — `ServiceUniversalSearch` runs in background and the filtered list is serialized to main; no shared native resource or persisted state is touched.
- **Risk Areas**: The 4 character threshold. A short ticker whose row carries no matching text in `name` / `subtitle` and no matching dex would be dropped; every ticker checked against live data (`btc`, `eth`, `sol`, `xrp`, `ada`, `bnb`, `sui`, `apt`, `arb`, `ena`, `ldo`, `trx`, `ton`, `nvda`, `tsla`, `aapl`, `amd`, `goog`, `meta`, `gold`) keeps its own row. Nothing outside perps search is touched.

## Test plan

- [ ] Universal search `usdc` — Perps section shows no results
- [ ] Universal search `usd` and `sd` — Perps section shows no results
- [ ] Universal search `btc`, `eth`, `sol`, `nvda`, `gold` — the matching market is still listed first
- [ ] Universal search `ar` — only markets whose ticker or name carries `ar`, no unrelated `para` rows
- [ ] Universal search `sato` and `xbt` — BTC is still listed (alias only match)
- [ ] Universal search `xyz` and `para` — sub-dex markets are still listed
- [ ] Universal search `nasdaq`, `japan`, `bitcoin`, `unitree` — description matches are unchanged
- [ ] Universal search with a Chinese alias (bitcoin / Japan) — results are unchanged
- [ ] Tap a perps result — it still opens the correct market, including sub-dex ones (`xyz:` / `para:`)
- [x] `yarn jest packages/kit-bg/src/services/ServiceUniversalSearch.perpsAssetCapability.test.ts` — 10 passed
- [x] `yarn agent:check --profile commit` — pass

[OK-60751]: https://onekeyhq.atlassian.net/browse/OK-60751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Rebase note
Re-applied onto `hotfix/v6.5.2` from `release/v6.5.0` (PR #12952, now closed). Both touched files are byte-identical between the two bases, so the PR patch applied cleanly and the resulting file contents are byte-identical to the original branch — verified by md5 per file. Net diff against 6.5.2 is exactly these two files, +86 lines.

Pre-existing on this base and unrelated to this PR: `yarn tsc:staged` reports `isInstallerPath` missing on `AppUpdater` in `DesktopApiAppUpdate.ts:832`, which also reproduces on a clean `hotfix/v6.5.2` worktree.

### DIFF
--- a/packages/kit-bg/src/services/ServiceUniversalSearch.ts
+++ b/packages/kit-bg/src/services/ServiceUniversalSearch.ts
@@ -21,6 +21,7 @@ import {
   buildTradablePerpMaxLeverageMap,
   isPerpsUniverseCacheComplete,
 } from '@onekeyhq/shared/src/utils/perpsDexUtils';
+import { matchesTokenSearchAlias } from '@onekeyhq/shared/src/utils/perpsUtils';
 import {
   PROMISE_CONCURRENCY_LIMIT,
   promiseAllSettledEnhanced,
@@ -56,10 +57,11 @@ const PERPS_UNIVERSE_SEARCH_MAX_AGE_MS = timerUtils.getTimeDurationMs({
   minute: 5,
 });
 const PERPS_ASSET_TYPE_VERSION = 2;
-// The search index also matches collateral text, so `usdc` returns every
-// USDC settled market. Only short ASCII tickers are held to a literal match:
-// a longer query is where the index's description hits such as `nasdaq` are
-// the point, and a non-ASCII one is where its localized aliases are.
+// Every market's search aliases include its pair notations (`btc-usdc`), so
+// `usdc` returns every USDC settled market. Only short ASCII tickers are held
+// to a literal match: a longer query is where the index's description hits
+// such as `nasdaq` are the point, and a non-ASCII one is where the localized
+// aliases the server may hold beyond our cached map are.
 const PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH = 4;
 
 @backgroundClass()
@@ -1301,10 +1303,19 @@ class ServiceUniversalSearch extends ServiceBase {
         const requiresLiteralMatch =
           /^[a-z0-9]+$/.test(normalizedQuery) &&
           normalizedQuery.length <= PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH;
+        // Read only when the filter can drop something: an alias is the only
+        // way a query reaches a market whose ticker and subtitle both miss it.
+        const tokenSearchAliases = requiresLiteralMatch
+          ? await this.backgroundApi.serviceHyperliquid.getTokenSearchAliases()
+          : undefined;
 
         const items: IUniversalSearchPerpResult['items'] =
           rawAssets
             ?.filter((asset) => {
+              const coin = buildCoinFromSearchAssetType({
+                assetType: asset.type,
+                name: asset.name,
+              });
               // `asset.type` is the dex prefix, so matching it exactly keeps
               // `xyz` browsing that sub-dex without a fragment such as `ar`
               // waving through every `para` row.
@@ -1313,7 +1324,13 @@ class ServiceUniversalSearch extends ServiceBase {
                 asset.type?.toLowerCase() !== normalizedQuery &&
                 ![asset.name, asset.subtitle].some((field) =>
                   field?.toLowerCase().includes(normalizedQuery),
-                )
+                ) &&
+                !matchesTokenSearchAlias({
+                  query: normalizedQuery,
+                  aliases: coin
+                    ? tokenSearchAliases?.[coin]?.aliases
+                    : undefined,
+                })
               ) {
                 return false;
               }
@@ -1323,10 +1340,6 @@ class ServiceUniversalSearch extends ServiceBase {
               if (!tradablePerpMaxLeverage) {
                 return true;
               }
-              const coin = buildCoinFromSearchAssetType({
-                assetType: asset.type,
-                name: asset.name,
-              });
               return Boolean(coin && tradablePerpMaxLeverage.has(coin));
             })
             .map((asset) => {
--- a/packages/kit-bg/src/services/ServiceUniversalSearch.ts
+++ b/packages/kit-bg/src/services/ServiceUniversalSearch.ts
@@ -56,6 +56,11 @@ const PERPS_UNIVERSE_SEARCH_MAX_AGE_MS = timerUtils.getTimeDurationMs({
   minute: 5,
 });
 const PERPS_ASSET_TYPE_VERSION = 2;
+// The search index also matches collateral text, so `usdc` returns every
+// USDC settled market. Only short ASCII tickers are held to a literal match:
+// a longer query is where the index's description hits such as `nasdaq` are
+// the point, and a non-ASCII one is where its localized aliases are.
+const PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH = 4;
 
 @backgroundClass()
 class ServiceUniversalSearch extends ServiceBase {
@@ -1292,9 +1297,26 @@ class ServiceUniversalSearch extends ServiceBase {
             ? await this.readTradablePerpMaxLeverageMap()
             : cachedPerpMaxLeverage;
 
+        const normalizedQuery = input.trim().toLowerCase();
+        const requiresLiteralMatch =
+          /^[a-z0-9]+$/.test(normalizedQuery) &&
+          normalizedQuery.length <= PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH;
+
         const items: IUniversalSearchPerpResult['items'] =
           rawAssets
             ?.filter((asset) => {
+              // `asset.type` is the dex prefix, so matching it exactly keeps
+              // `xyz` browsing that sub-dex without a fragment such as `ar`
+              // waving through every `para` row.
+              if (
+                requiresLiteralMatch &&
+                asset.type?.toLowerCase() !== normalizedQuery &&
+                ![asset.name, asset.subtitle].some((field) =>
+                  field?.toLowerCase().includes(normalizedQuery),
+                )
+              ) {
+                return false;
+              }
               // The search index still carries delisted assets — `xyz:UNITREE`
               // survives there after moving to `para`, so the same ticker would
               // appear twice with one dead row.
--- a/packages/shared/src/utils/perpsUtils.ts
+++ b/packages/shared/src/utils/perpsUtils.ts
@@ -1685,6 +1685,34 @@ export interface ITokenSearchAliasItem {
 
 export type ITokenSearchAliases = Record<string, ITokenSearchAliasItem>;
 
+// Every market also carries its pair notations (`btc-usdc`, `btc/usd`) as
+// aliases, so matching those would let a quote currency query pull in every
+// market that settles in it.
+const TOKEN_SEARCH_PAIR_ALIAS_REGEX = /[-/]usdc?$/;
+
+// Mirrors findTokensByAlias for a single market, minus the pair notations.
+export function matchesTokenSearchAlias({
+  query,
+  aliases,
+}: {
+  query: string;
+  aliases: string[] | undefined;
+}): boolean {
+  if (!query || !aliases?.length) {
+    return false;
+  }
+  const shouldMatchAliasPrefix = /^[a-z0-9]{1,2}$/.test(query);
+  return aliases.some((alias) => {
+    const normalizedAlias = alias.toLowerCase();
+    if (TOKEN_SEARCH_PAIR_ALIAS_REGEX.test(normalizedAlias)) {
+      return false;
+    }
+    return shouldMatchAliasPrefix
+      ? normalizedAlias.startsWith(query)
+      : normalizedAlias.includes(query);
+  });
+}
+
 /**
  * Find token symbols by search alias
  * @param query - Search query (already lowercased)
@@ -2232,6 +2260,7 @@ export default {
   getHyperliquidTokenImageUrl,
   getHyperliquidTokenImageUris,
   findTokensByAlias,
+  matchesTokenSearchAlias,
   getTokenSubtitle,
   mapTriggerOrderType,
   inferTpsl,
--- a/packages/kit-bg/src/services/ServiceUniversalSearch.perpsAssetCapability.test.ts
+++ b/packages/kit-bg/src/services/ServiceUniversalSearch.perpsAssetCapability.test.ts
@@ -80,4 +80,68 @@ describe('ServiceUniversalSearch perps asset capability', () => {
       }),
     ]);
   });
+  // Rows the endpoint returns for `usdc` because they are USDC settled, not
+  // because the query matches anything the row displays.
+  const searchAssets = [
+    { type: 'perps', name: 'BTC', subtitle: 'Bitcoin' },
+    { type: 'xyz', name: 'AMAT', subtitle: 'Applied Materials' },
+    { type: 'xyz', name: 'EWJ', subtitle: 'Japan ETF' },
+    { type: 'para', name: 'UNITREE', subtitle: 'Unitree' },
+  ].map((asset) => ({
+    ...asset,
+    logoUrl: 'https://example.com/logo.png',
+    maxLeverage: null,
+    midPx: '1',
+    dayNtlVlm: null,
+  }));
+
+  function buildSearchService() {
+    const backgroundApi = {
+      simpleDb: {
+        perp: {
+          getTradingUniverse: jest.fn().mockResolvedValue({
+            universesByDex: [
+              [{ name: 'BTC', maxLeverage: 40 }],
+              [
+                { name: 'xyz:AMAT', maxLeverage: 10 },
+                { name: 'xyz:EWJ', maxLeverage: 10 },
+              ],
+              [{ name: 'para:UNITREE', maxLeverage: 5 }],
+            ],
+            updatedAt: Date.now(),
+          }),
+        },
+      },
+      serviceHyperliquid: { refreshTradingMeta: jest.fn() },
+    };
+    const Ctor = ServiceUniversalSearch as unknown as new (args: {
+      backgroundApi: unknown;
+    }) => ServiceUniversalSearch;
+    const service = new Ctor({ backgroundApi });
+    (service as any).getClient = jest.fn().mockResolvedValue({
+      get: jest.fn().mockResolvedValue({ data: { data: searchAssets } }),
+    });
+    return service;
+  }
+
+  test.each<[string, string[]]>([
+    ['usdc', []],
+    ['btc', ['BTC']],
+    // An exact dex prefix browses that sub-dex, a fragment of one does not.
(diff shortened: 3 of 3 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit-bg/src/services/ServiceUniversalSearch.ts
```diff
@@ -1292,9 +1297,26 @@ class ServiceUniversalSearch extends ServiceBase {
...
         const items: IUniversalSearchPerpResult['items'] =
           rawAssets
             ?.filter((asset) => {
+              // `asset.type` is the dex prefix, so matching it exactly keeps
+              // `xyz` browsing that sub-dex without a fragment such as `ar`
+              // waving through every `para` row.
+              if (
+                requiresLiteralMatch &&
+                asset.type?.toLowerCase() !== normalizedQuery &&
+                ![asset.name, asset.subtitle].some((field) =>
```
sidmorizon, line 1325: # 🟠 P1 字面匹配只看 `name`/`subtitle`，短拉丁别名查询会被整体过滤为空

## 问题

新增的过滤只在 `asset.name` 与 `asset.subtitle` 两个字段上做 `includes`，但服务端搜索索引还有第三个维度——别名。仓库里已经有服务端下发的别名结构 `ITokenSearchAliases`（`packages/shared/src/utils/perpsUtils.ts`），例如 `BTC: { aliases: ['Bitcoin', 'Satoshi', '比特币'] }`、`SNDK: { aliases: ['SanDisk', '闪迪'] }`，注意 `ITokenSearchAliasItem.subtitle` 是可选的，`getTokenSubtitle()` 对没有 `subtitle` 的条目返回 `undefined`，也就是别名文本并不保证会出现在 `name` 或 `subtitle` 里。

本 PR 的注释已经承认索引里有别名（“a non-ASCII one is where its localized aliases are”），并因此把非 ASCII 查询豁免掉了；但同一批别名里的拉丁文本（`Satoshi` / `SanDisk` / `Zcash`）在 `length <= PERPS_SEARCH_LITERAL_MATCH_MAX_QUERY_LENGTH` 这一段里没有任何豁免，会被 `return false` 直接丢掉。

## 影响

具体回归：查 `sato`（4 字符）本应命中 `BTC`（别名 `Satoshi`），但 `'btc'.includes('sato')` 为 `false`、`subtitle` 又可能不存在，整个 Perp 分区返回 `0` 条；查 `disk` 命中 `SNDK`（别名 `SanDisk`）同理——仓库自带用例 `findTokensByAlias('disk', aliases)` 正是断言这条别名命中，而该别名条目并没有 `subtitle`。

结果是同一个资产用中文 `比特币` 能搜到、用 `sato` 搜不到，行为不对称。另外 `1`~`2` 字符查询这里用的是 `includes` 而不是 `findTokensByAlias` 里对 `/^[a-z0-9]{1,2}$/` 采用的 `startsWith`，短查询的噪声其实也没被真正收敛。

## 建议

让判定复用现有的 perps 搜索语义，而不是新写一套只看两个字段的规则：把服务端别名一并纳入匹配（或让后端在 `perpsAsset` 响应里补一个 `aliases` 字段再参与判定），并对极短查询采用与 `findTokensByAlias` 一致的前缀规则：

```ts
const matchesQuery = (field?: string) => {
  const value = field?.toLowerCase();
  if (!value) return false;
  return normalizedQuery.length <= 2
    ? value.startsWith(normalizedQuery)
    : value.includes(normalizedQuery);
};
if (
  requiresLiteralMatch &&
  asset.type?.toLowerCase() !== normalizedQuery &&
  [asset.name, asset.subtitle, ...(asset.aliases ?? [])].some(matchesQuery)
) {
  return false;
}
```

同时建议给 `sato` / `disk` 这类“别名命中但 `name`、`subtitle` 都不含查询词”的场景补一条测试，当前 `test.each` 的用例全部是 `name` 或 `subtitle` 字面命中，覆盖不到这条路径。

[Claude session ↗](https://claude.ai/code/session_013p4C8axFyxZX7H7vXyFYN5)

---

Minnzen: Fixed in f401c4b — and digging into it turned up the actual root cause of this ticket, which I had wrong in the description.

**The alias list is where the `usdc` noise comes from.** Pulled the live config (`/utility/v1/perp-config`, 222 entries) and every market carries its pair notations as aliases:

```
"para:AAOI": { aliases: ["aaoi", "aaoi-usdc", "aaoi/usdc", "aaoi-usd", "aaoi/usd", "para:aaoi", ...] }
"BTC":       { aliases: ["btc", "bitcoin", "satoshi", "xbt", ... ] }        <- no pair form
```

216 of the 3394 aliases are pair notations, and they sit exactly on the 54 stock rows. That is why `usdc` returns those 56 rows and `btc-usdc` style text never shows on screen. So it is not collateral or description text as the PR body said — I have corrected that section.

That also means the suggested fix cannot be applied as written: matching the alias list wholesale re-breaks the ticket. Simulated against the live endpoint plus the live alias map, `usdc` goes straight back to keeping all 56 rows, because every one of them has an `*-usdc` alias.

What landed instead: match the alias list but skip the pair notations (`/[-/]usdc?$/`), reusing the `findTokensByAlias` prefix rule for 1–2 character queries. New `matchesTokenSearchAlias()` in `perpsUtils.ts`, read from `serviceHyperliquid.getTokenSearchAliases()` and only when the filter can actually drop rows. An absent or empty map degrades to the previous ticker/subtitle behaviour.

Replaying the live endpoint through it:

```
query   rows  before  after
usdc     56      0      0     <- still clean
sato      1      0      1     <- BTC, alias `satoshi`
xbt       2      1      2     <- BTC
disk      3      1      3     <- SNDK, WDC, STX
btc       3      2      3     <- + IREN
gold      1      1      1
nvda      1      1      1
xyz      46     44     44
```

Two notes on the report itself:

- The snippet inverts the guard — `[...].some(matchesQuery)` inside the `if` that returns `false` drops the rows that match and keeps the ones that do not. Kept the negated form.
- `SNDK` does carry a subtitle on this endpoint (`SanDisk`), so `disk` was surviving on the subtitle substring already; `sato` on BTC was the real hole.

Tests: added `['gold', ['BTC']]` (alias-only path, the row shows neither `gold` nor an alias) and `['nvda', ['NVDA']]` for the 4 character boundary that the cursor review asked for. Mutation checked both halves — dropping the alias clause fails `gold`, dropping the pair-notation skip fails `usdc`. 10 passed, lint clean.

`tsc:staged` still reports `DesktopApiAppUpdate.ts:832 isInstallerPath`; that reproduces on a clean `hotfix/v6.5.2` tree (electron-updater drift), unrelated to this PR.

packages/shared/src/utils/perpsUtils.ts
```diff
@@ -1685,6 +1685,34 @@ export interface ITokenSearchAliasItem {
...
+  query: string;
+  aliases: string[] | undefined;
+}): boolean {
+  if (!query || !aliases?.length) {
+    return false;
+  }
+  const shouldMatchAliasPrefix = /^[a-z0-9]{1,2}$/.test(query);
+  return aliases.some((alias) => {
+    const normalizedAlias = alias.toLowerCase();
+    if (TOKEN_SEARCH_PAIR_ALIAS_REGEX.test(normalizedAlias)) {
```
sidmorizon, line 1707: # 🟠 P1 别名里的 dex 前缀写法没被跳过，`p` / `pa` / `par` / `ara` 会整片放行子 dex

## 问题

`matchesTokenSearchAlias()` 只跳过配对写法 `TOKEN_SEARCH_PAIR_ALIAS_REGEX = /[-/]usdc?$/`。但你上一条回复里贴的线上配置里，除了 `aaoi-usdc`、`aaoi/usd` 之外还有一条 `para:aaoi`——市场自身的「dex 前缀写法」。它既不被配对正则命中，也不是用户真正会去搜的词，却作为普通别名参与了 `includes` / `startsWith` 匹配。

于是 dex 名本身的片段会通过别名通道命中该 dex 的**全部**市场：3~4 字符走 `includes` 的 `par`、`ara` 命中每一条 `para:*`；1~2 字符走 `startsWith` 的 `p`、`pa` 同样命中；`x`、`xy` 对 `xyz:*` 同理。而上游 `asset.type?.toLowerCase() !== normalizedQuery` 那条精确 dex 判定的注释写的正是「a fragment such as `ar` 不能 waving through every `para` row」，别名子句把这个口子重新打开了。

用线上别名形状复算（`para:AAOI` 的别名取你贴的那组）：

```
query  当前命中                withFix
p      ["para:AAOI"]           []
pa     ["para:AAOI"]           []
par    ["para:AAOI"]           []
ara    ["para:AAOI"]           []
ar     []                      []      <- 2 字符前缀规则本来就挡住了
aaoi   ["para:AAOI"]           ["para:AAOI"]
usdc   []                      []
sato   ["BTC"]                 ["BTC"]
```

## 影响

用户逐字输入 `PLTR` 这类股票代码时，敲到 `p`、`pa`、`par` 这几步，Perp 分区返回的是整个 `para` 子 dex（线上 `54` 条股票行），而不是字面命中的那几条——和本 PR 要治的 `usdc` 噪声属于同一类，只是触发词换成了 dex 名的片段。`ara` 这种 3 字符查询同理。

新增用例挡不住这条：mock 的别名补了配对写法 `amat-usdc`、`amat/usd`，却没有线上同时存在的 `para:aaoi` / `xyz:amat` 形式，所以 `['ar', []]` 这条（`ar` 恰好被 1~2 字符前缀规则救下）通过了，`par` / `pa` 这类真正会退化的查询没有覆盖。

## 建议

把 dex 前缀写法和配对写法一起跳过，注意保持 `some` 回调里的极性——命中跳过条件时 `return false` 表示「这条别名不算匹配」：

```ts
const TOKEN_SEARCH_PAIR_ALIAS_REGEX = /[-/]usdc?$/;
const TOKEN_SEARCH_DEX_PREFIX_ALIAS_REGEX = /^[a-z0-9]+:/;

// ... inside aliases.some((alias) => { ... })
    if (
      TOKEN_SEARCH_PAIR_ALIAS_REGEX.test(normalizedAlias) ||
      TOKEN_SEARCH_DEX_PREFIX_ALIAS_REGEX.test(normalizedAlias)
    ) {
      return false;
    }
```

想更保守可以只跳过前缀命中 `SUB_DEX_LIST` 的别名（`perpsUtils.ts` 已经从 `perp.constants` 引入过常量）。精确 dex 浏览由 `asset.type === normalizedQuery` 单独负责，不依赖这条别名，`aaoi` / `sato` / `usdc` 的现有结论也都不变。同时建议把 `para:aaoi`、`xyz:amat` 补进测试 mock，并加一条 `['par', []]`，让这个维度真的能挡回归。

[Claude session ↗](https://claude.ai/code/session_013p4C8axFyxZX7H7vXyFYN5)

---

originalix:

@Minnzen

Reviewed commit `f401c4b95885`. This report always reflects the latest reviewed changes.

### Review summary

The background universal-search boundary now combines Perps endpoint rows with the persisted Hyperliquid alias map, preserving short alias-only matches while excluding settlement-pair aliases such as `btc-usdc`; longer and non-ASCII queries still retain endpoint ranking, and tradability filtering remains unchanged. The shared matcher and table-driven tests keep the implementation localized, with no signing, authorization, or sensitive-data changes. However, alias-map readiness has become a correctness prerequisite even though the configuration refresh remains asynchronous, so a first search can still discard the valid alias-only results this PR intends to restore.

**What needs attention:** Verify the first-search path when the background runtime has no persisted alias map and the Perps configuration refresh has not completed.

### Issues to address

- **P2 · Cold-start alias searches still lose valid Perps results**
  - **Impact:** In the `bg` runtime, `getTokenSearchAliases()` starts the server refresh without awaiting it and immediately reads SimpleDB. On a fresh install, after SimpleDB self-healing, or before bootstrap configuration loading finishes, that persisted value is undefined. The new predicate then drops endpoint rows matched only through aliases, such as BTC for `gold`, and the background memoizer retains the false-negative result for 30 seconds. SimpleDB's native backing resource may be process-shared, but the alias object and memoized result are per-runtime JS copies; `main` receives a separately serialized result and cannot repair the missing `bg` data. Because `main` and `bg` initialize independently, foreground startup does not guarantee alias readiness.
  - **Suggested action:** Make this search path await the Perps configuration refresh before reading the alias map and applying the alias-dependent rejection, then add a test where the persisted map starts undefined and becomes available through the refresh.

### Validation gaps

- The focused Jest invocation stopped before test execution because Yarn's node_modules state file is missing. CI still needs to run `packages/kit-bg/src/services/ServiceUniversalSearch.perpsAssetCapability.test.ts`.
- No live first-launch scenario was exercised. Runtime confirmation should start without persisted Perps aliases, immediately search `gold`, and verify that BTC appears without waiting for a second query.

zhaono1 (dismissed):

originalix (approved):

PR #12963, opened 2026-08-21
fix: restore extension firmware progress forwarding

wabicai:
## Summary

- restore the offscreen hardware event forwarder after the low-level SDK clears all listeners
- cover both removeAllListeners and dispose lifecycle paths
- add a focused regression test for firmware progress forwarding after SDK reset

## Root cause

The extension registers its hardware global event forwarder once in the offscreen runtime. SDK cleanup clears the low-level event emitter listeners, while the memoized offscreen SDK module remains alive, so the forwarder is not registered again. The SDK therefore continues producing progress logs in offscreen, but background atoms no longer receive those events. Workflow completion remains unaffected because it is written directly by ServiceFirmwareUpdate.

## Verification

- yarn jest packages/kit-bg/src/offscreens/instance/offscreenApi.test.ts apps/ext/src/offscreen/offscreenSetup.test.ts --runInBand
- yarn exec eslint packages/kit-bg/src/offscreens/instance/offscreenApi.ts packages/kit-bg/src/offscreens/instance/offscreenApi.test.ts --no-cache
- yarn app:ext:build

The extension production build completed successfully with existing dependency and bundle-size warnings.

### DIFF
--- a/packages/kit-bg/src/offscreens/instance/offscreenApi.test.ts
+++ b/packages/kit-bg/src/offscreens/instance/offscreenApi.test.ts
@@ -1,0 +1,71 @@
+import appGlobals from '@onekeyhq/shared/src/appGlobals';
+import { importHardwareSDKLowLevel } from '@onekeyhq/shared/src/hardware/sdk-loader';
+
+import { OFFSCREEN_API_MESSAGE_TYPE } from '../types';
+
+import offscreenApi from './offscreenApi';
+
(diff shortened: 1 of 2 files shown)


wabicai:
Superseded by #12966, which now includes this offscreen event-forwarding fix together with the TypeScript correction and the consolidated Extension firmware lifecycle validation.

PR #12964, opened 2026-08-21
fix: prevent iOS hardware dialog overlay conflicts

huhuanming:
## Summary

- serialize the iOS hardware loading Sheet mount before wallet creation state checks
- wait for the main-runtime close acknowledgement and native Sheet exit before opening the wallet-type dialog
- keep the hardware/BLE session active and leave non-iOS behavior unchanged

## Root cause

On iOS, the background runtime can finish clearing the hardware UI atom before the main runtime commits the React close and the native `FullWindowOverlay` Sheet finishes exiting. Mounting the wallet-type dialog during that window can leave the previous `Sheet.Overlay` above it and intercept touches.

## Side effects

- adds the existing 150 ms loading animation wait plus up to roughly 300 ms for a stable native Sheet exit on iOS
- fails through the existing error/retry path after a 5-second UI acknowledgement timeout instead of stacking another non-interactive overlay
- preserves the hardware loading UI, its delayed close action, and `skipDeviceCancel: true`

## Testing

- `yarn jest packages/kit/src/provider/Container/HardwareUiStateContainer/hardwareUiStateDialogLifecycle.test.ts --runInBand` (8 tests passed)
- `git diff --check`

## Device verification

- [ ] confirm the Standard/Hidden wallet-type options remain tappable on iOS 18.7.5
- [ ] confirm the subsequent PIN/passphrase dialogs render correctly
- [ ] confirm the BLE session remains connected throughout the flow


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix (approved):

PR #12968, opened 2026-08-21
fix: keep iOS tab bar hidden after rotation OK-60835

huhuanming:
[OK-60835](https://onekeyhq.atlassian.net/browse/OK-60835)

----------

## Summary
- use `UITabBarController.setTabBarHidden` on iOS 18+ so UIKit keeps its internal tab bar visibility state synchronized across rotation
- keep the direct `UITabBar.isHidden` fallback for iOS 17 and earlier
- remove all temporary JS, Fabric, Swift, and NativeLogger diagnostics after QA verification

## Scope
- target: iPadOS native shell and the main UI runtime tab view
- no JavaScript behavior change
- no background runtime change
- requires an App Shell build because the fix is native

## QA
- QA verified the fix on iPad mini with iOS 26 by repeatedly entering and leaving Market detail while rotating the device

## Validation
- patch applies cleanly to a pristine `@onekeyfe/react-native-tab-view` 3.0.78 package
- patch-package postinstall applies the patch successfully
- Xcode 26.6 Release build of the `react-native-tab-view` iOS Simulator target for arm64 and x86_64

Jira: https://onekeyhq.atlassian.net/browse/OK-60835

[OK-60835]: https://onekeyhq.atlassian.net/browse/OK-60835?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

patches/@onekeyfe+react-native-tab-view+3.0.78.patch
```diff
@@ -0,0 +1,284 @@
+diff --git a/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewComponentView.mm b/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewComponentView.mm
+index 6ccad7e..e829537 100644
+--- a/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewComponentView.mm
++++ b/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewComponentView.mm
+@@ -1,4 +1,5 @@
+ #import "RCTTabViewComponentView.h"
++#import "RCTTabViewLog.h"
```
sidmorizon: 🟠 P1: Patch references a logger class that does not exist

Severity: `severe`

The patch adds `#import "RCTTabViewLog.h"` and calls `RCTTabViewLog` in both the Objective-C++ and Swift files, but never adds that header or class, and no `RCTTabViewLog` exists anywhere in the repo. Applying the patch and building the iOS target fails to compile.

PR #12972, opened 2026-08-22
fix: stop stale device records from shadowing desktop BLE connectId resolution(OK-60878)

originalix:
[OK-60878](https://onekeyhq.atlassian.net/browse/OK-60878)

----------

## Summary
- Prefer the identity-qualified device lookup (`connectId` + `featuresDeviceId`) over the connectId-only lookup in `getCompatibleConnectId`, keeping the connectId-only match as a strict fallback
- Persist a verified BLE binding and de-alias stale sibling records atomically via the new `localDb.updateDeviceBleConnectIdAndCleanStaleAliases` (one transaction; nothing commits on failure, so both bind paths stay retryable)
- Key the pairing repair on the identity read over the live connection instead of the dialog's expected id, so a repair triggered through a stale record still binds (and heals) the live record
- Cover the semantics with a pure-predicate unit suite plus regression tests for the endless pairing-dialog loop and the stale-expected-id repair path

## Related Issues

- [OK-60878](https://onekeyhq.atlassian.net/browse/OK-60878) — QA handoff task: product behavior changes and post-fix verification cases (Chinese)
- [OK-60978](https://onekeyhq.atlassian.net/browse/OK-60978) — Bug / reproduction record: environment, preconditions, step-by-step repro on 6.5.2, the always-reproducible pre-recreation variant, and the poisoned-DB upgrade variant used to verify self-healing (Chinese)

## Intent & Context
A production v6.5.2 desktop (macOS) user wiped their OneKey Pro and re-onboarded it via desktop Bluetooth. Afterwards every hardware call — device details refresh, the batch account derivation right after wallet creation, settings reads — re-opened the "Pairing with your device…" dialog in an endless loop, even though each pairing repair reported success. Production logs showed `serviceHardware.repairBleConnectIdWithProgress` being called and resolving successfully every ~10 seconds while the dialog kept coming back, with the NobleBLE link itself healthy the whole time.

## Root Cause
A device wipe changes the raw `deviceId`, so wallet creation correctly produces a NEW Device record (`getExistingDevice` must not silently re-associate a new seed onto an old wallet). But the pre-wipe record survives, and both records share the same serial-based `connectId` (the USB serial, e.g. `PRB09B0058A`):

- `getCompatibleConnectId` resolved the device with a connectId-only `getDeviceByQuery`. `getAllDevices()` is unordered (dbDeviceId is a random UUID), and on the affected machine the stale pre-wipe record won the lookup. It has no `bleConnectId`, so the pairing dialog opened.
- The repair flow queries with `connectId` + `featuresDeviceId`, so it verified the live device and wrote the freshly bound `bleConnectId` to the NEW record.
- Reads kept resolving the stale record while writes went to the live one, so the loop never terminated. The silent-bind path (OK-60091) could not rescue it either: its usb-alias guard bails immediately when the incoming connectId equals the stale record's own `connectId`.

A local cross-review surfaced a residual P1 on top of the first two changes: on an already-poisoned database where the live record was bound **before** upgrading (exactly what the old app produced by repairing every 10 s), connectId-only callers (`getDeviceManagementSnapshot`, `getPro2OnboardingStatus`, `DeviceSettingsManager` operations) still resolved the stale record — and the repair those callers triggered could **never** succeed, because the dialog derives its expected id from the stale record's frozen features while the live device reports the post-wipe identity. No bind ⇒ no cleanup ⇒ the loop survived for those paths.

## Design Decisions
- **Align read and write predicates** instead of sorting or deduping lookups: `getCompatibleConnectId` tries `{connectId, featuresDeviceId}` first and falls back to the connectId-only match. The original guarantee — stale device info must never veto a valid transport-ID match — is preserved because the fallback keeps the legacy behavior whenever the combined query misses.
- **Bind + cleanup in one transaction** (review requirement): committing a binding without its cleanup would leave the stale sibling shadowing connectId-only lookups; a failure rolls back both writes so the pairing repair / silent bind stays fully retryable. The Device cache is invalidated again after commit so a concurrent pre-commit read cannot re-fill it with stale records.
- **The live-read identity is the authority in the repair** (cross-review P1 fix): after the repair's connect succeeds, the record is located by the identity the device just reported (`deviceId`/`device_id`), not by the dialog's possibly-stale expected id. This is the strongest identity proof available (synchronous read on the live link, stronger than DB shape or async state events), it adds no hardware round trip, and it makes healing bidirectional (restoring the old seed heals in the opposite direction automatically). If no record matches the live identity (wiped but not yet re-onboarded), the repair fails closed exactly as before.
- Cleanup only touches same-vendor records with a persisted and different `deviceId`; records without an identity are never treated as stale. Only colliding fields are cleared: `connectId` to `''` (required column) and `usbConnectId`/`bleConnectId` to `undefined`. No schema change, no `LOCAL_DB_VERSION` bump, no new hardware probes, no event-pipeline changes.
- The usb-alias guard in `getPersistedDesktopBleConnectId` is deliberately untouched — it protects against genuine serial-into-bleConnectId poisoning.
- Known accepted trade-offs: an already-poisoned install heals at its first pairing-dialog repair (one visible dialog) rather than invisibly; a harmless duplicate where both records carry the same usable BLE UUID is left as-is (any future dialog repair clears it). Deliberately deferred: wallet-create-time dedupe of stale siblings, and a dedicated "device was reset" UX for dead wallets.

## Changes Detail
- `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts` — identity-first lookup order in `getCompatibleConnectId`; `persistVerifiedBleConnectId` (non-swallowing) wired into the `repairBleConnectIdWithProgress` and `silentlyBindLiveDesktopBleConnectId` success paths; repair locates the record by the live-read identity (`deviceId || device_id`, covering V2 state projections).
- `packages/kit-bg/src/dbs/local/LocalDbBase.ts` — exported pure helpers `collectDeviceConnectIdAliases` / `isStaleDeviceConnectIdAliasRecord`; atomic `updateDeviceBleConnectIdAndCleanStaleAliases` (alias set pre-computed including the incoming bleConnectId, single transaction, post-commit cache invalidation, returns cleaned ids).
- `packages/kit-bg/src/dbs/local/LocalDbBase.deviceConnectIdAliases.test.ts` — predicate unit suite (9 cases) covering every guard branch.
- `packages/kit-bg/src/services/ServiceHardware/ServiceHardware.getCompatibleConnectId.test.ts` — regression tests: duplicate-record pairing loop, atomic bind+cleanup wiring for repair and silent bind, stale-expected-id repair healing, fail-closed when no record matches the live identity; five existing tests migrated from sequence mocks to predicate-honoring mocks so they genuinely exercise the "stale identity does not veto the connectId match" fallback.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop (primary — desktop BLE binding and repair); the lookup-order change is shared code, with native BLE (iOS/Android) behavior covered by updated tests
- **Risk Areas**: `getCompatibleConnectId` sits on every hardware call path — mitigated by the strict fallback (byte-for-byte legacy behavior whenever the combined query misses). The repair's identity-keyed lookup only runs inside the pairing dialog (an already-degraded state); healthy flows produce identical queries to the previous code. Alias cleanup writes to sibling records — gated on a just-read live identity plus a persisted-and-different `deviceId`, atomic with the binding, idempotent, and retried by any later dialog repair.

## Test plan
- [x] `yarn jest` on all ServiceHardware suites + LocalDbBase suites + related ServiceAccount/ServiceFirmwareUpdate/ServiceThirdPartyHardware suites — all pass (388 tests across hardware suites)
- [x] `yarn agent:check --profile commit` passes
- [ ] Manual (desktop): wipe a device that already has a wallet in the app, re-onboard it via desktop Bluetooth, and confirm batch account creation completes without repeated pairing dialogs; the stale record's aliases are cleared (sdkLog `persistVerifiedBleConnectId`)
- [ ] Manual (desktop): an install already showing the endless pairing dialog self-heals after its first pairing-dialog repair on this build, including on connectId-only paths (device settings / device label); device-management snapshot recovers afterwards

[OK-60878]: https://onekeyhq.atlassian.net/browse/OK-60878?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-60978]: https://onekeyhq.atlassian.net/browse/OK-60978?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -4465,6 +4474,34 @@ class ServiceHardware extends ServiceBase {
...
+        verifiedDeviceId,
+      });
+      if (cleanedIds.length) {
+        defaultLogger.hardware.sdkLog.log(
+          'cleanStaleDeviceConnectIdAliases: cleared stale sibling connect-id aliases',
+          JSON.stringify({ keepDbDeviceId, cleanedIds }),
+        );
+      }
+    } catch (error) {
+      console.error('cleanStaleDeviceConnectIdAliases failed:', error);
```
sidmorizon: # 🟠 P1 清理只有一次机会，失败后无法重试

## 问题

绑定与清理是两个独立事务：`updateDeviceConnectId()` 先提交 `bleConnectId`，随后 `cleanStaleDeviceConnectIdAliases()` 才另开一个 `withTransaction`。这里的 `catch` 只 `console.error` 就吞掉（`cleanStaleDeviceConnectIdAliases` 在 `getDeviceSafe` 取不到记录时还会静默 `return []`），所以清理是否真的发生对调用方完全不可见。

一旦保留记录 `L` 已写入 `bleConnectId`，`getCompatibleConnectId` 会在进入静默绑定分支之前直接短路：

```ts
if (targetTransportType === EHardwareTransportType.DesktopWebBle) {
  if (persistedDesktopBleConnectId) {
    return persistedDesktopBleConnectId;   // ServiceHardware.ts:4757
  }
  ...
  if (device && !persistedDesktopBleConnectId) {   // :4764 清理的唯一入口
```

更关键的是，重试在结构上不可能成功。只传 `connectId` 的调用会解析到残留记录 `S`，而 `silentlyBindLiveDesktopBleConnectId` 的 `expectedDeviceId` 正是从 `S` 自身派生（`featuresDeviceId || getRawDeviceId(dbDeviceToSearchDevice(device))`），探测拿到的却是 `D_live`，`probedDeviceId === expectedDeviceId` 恒不成立；接着配对对话框以 `deviceId = D_old` 调 `repairBleConnectIdWithProgress`，其 `connectResult.deviceId === expectedDeviceId` 同样恒不成立并抛 `DeviceNotFound`。清理代码永远到不了。

## 影响

触发条件是清理事务抛错（Realm 写冲突 / `resetUtils.checkNotInResetting()` / IndexedDB abort）或在两次 `withTransaction` 之间进程退出，窗口约 `1` 次 DB 往返——概率低，但一旦落入该状态就是**永久**的：`updateDeviceState()` 按 serialNo+deviceId 命中 `L`、不动 `S`；`_matchExistingDeviceRecord()` 按 `deviceId`/`uuid` 命中 `L`；`cleanDeviceConnectId()` 只被 dev-only 的 `clearAllBleConnectIdsForTesting` 调用；重启也没有相关迁移。

此时所有只传 `connectId` 的调用方退回本 PR 要修的配对循环：`getDeviceState`、`unlockDevice`、`getDeviceManagementSnapshot`、`uploadResource` / `uploadPro2Nft` / `uploadPortfolioPackage`，以及 `DeviceSettingsManager` 的 `_withDeviceProcessing`、`applySettingsToDevice`、`getDeviceLabel`。而带 `featuresDeviceId` 的调用方因身份优先查询仍然正常，故障从「必现」变成「时好时坏」，更难定位。

## 建议

合并为单个 LocalDb 方法、在一个事务里完成「写 `bleConnectId` + 清理兄弟别名」。可行性已确认：`updateDeviceConnectId` 与 `cleanStaleDeviceConnectIdAliases` 都用 `withTransaction(EIndexedDBBucketNames.account, ...)` 且都只写 `ELocalDBStoreNames.Device`，`createHwWallet` 已有同一事务内多次 `txUpdateRecords` 的先例。只需把别名集合改为事务前计算、不再依赖「绑定已提交」：

```ts
const aliases = [
  ...collectDeviceConnectIdAliases(keepDevice),
  bleConnectId.trim().toLowerCase(),
];
```

若不便合并，退而求其次：让 `:4757` 的早退不依赖「清理是否执行过」——早退前做一次廉价的同厂商别名冲突检测，发现冲突就补跑清理，使其保持可重试。

[Claude session ↗](https://claude.ai/code/session_01ARjbqDgutfi37w87LyTewp)

---

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -5479,6 +5533,75 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
+          }
+          if (collides(item.bleConnectId)) {
+            item.bleConnectId = undefined;
+          }
+          item.updatedAt = await this.timeNow();
+          return item;
+        },
+      });
+    });
+    return staleRecordIds;
```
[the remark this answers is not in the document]
originalix: 已修复(7bafaad01c)。清理逻辑已并入新的原子方法 `updateDeviceBleConnectIdAndCleanStaleAliases`,事务提交后调用 `clearStoreCachedDataIfMatch(ELocalDBStoreNames.Device)` 再次失效缓存,防止并发 `getAllDevices()` 在提交前用旧记录回填缓存导致 connectId-only 查询继续命中旧别名。

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
originalix: 已按建议修复(7bafaad01c),采用方案一:合并为单个 LocalDb 方法 `updateDeviceBleConnectIdAndCleanStaleAliases`,在同一个 `withTransaction(EIndexedDBBucketNames.account)` 里先写 `bleConnectId` 再清理兄弟别名。

关键点均已落实:

1. **别名集合在事务前计算**,并显式包含本次写入的 `bleConnectId`(绑定与清理同事务提交,无法从记录读回);
2. **ServiceHardware 侧不再吞错**:`persistVerifiedBleConnectId` 不做 try/catch,事务失败时绑定与清理一起回滚——repair 会以 DeviceNotFound 结束、下次弹窗重走完整流程,silent bind 返回 undefined 落回配对对话框,两条路径都保持可重试,不再存在"绑定已提交、清理永久丢失"的窗口;
3. 事务提交后再次失效 Device 缓存(同时覆盖 codex 提出的缓存回填问题)。

测试同步更新:silent bind / repair 的成功用例断言原子方法的入参,失败用例断言原子方法未被调用(未提交任何状态)。

originalix:
Follow-up commit 283966e3ac addresses a residual P1 found in local cross-review: on databases poisoned **before** this fix ships (live record already bound by the old app), connectId-only callers still resolved the stale record, and the repair they triggered could never succeed — its expected id came from the stale record's frozen features, so no bind and no cleanup ever happened.

Fix: the repair now locates the record by the identity read over the live connection (`deviceId || device_id`) instead of the dialog's expected id, then persists the binding + stale-alias cleanup through the existing atomic method. Healthy paths produce identical queries to before; wiped-but-not-re-onboarded devices still fail closed. One production-path change (~12 lines) + 2 regression tests; PR description updated accordingly.

wabicai (approved):

limichange (approved):

zhaono1 (approved):

sidmorizon (approved):

PR #13029, opened 2026-08-26
feat: register the Entropy hip-3 perps dex (OK-61073, OK-60689)

Minnzen:
[OK-61073](https://onekeyhq.atlassian.net/browse/OK-61073) [OK-60689](https://onekeyhq.atlassian.net/browse/OK-60689)

----------

## Summary

- Register the Entropy HIP-3 DEX (`io`, hyperliquid perp dex index `10`) in `SUB_DEX_LIST`.
- Stop a bare dex-prefix match from shadowing a real main-DEX symbol when the universe cache cannot confirm the guess (`?token=IOTA` was being read as `io:TA`).
- Show the `io` badge with its localized description on Market/Perps rows.
- Bump `assetTypeVersion` to `3` so the server can keep `io` out of responses to clients that predate it.
- Show the dex badge on Market watchlist perps rows, which never rendered one (OK-60689).

## Intent & Context

OK-61073 asks for Entropy to be reachable from universal search, the market list, the token selector, favorites and deep links, shipped through the 6.5 hot update. That is why this targets `release/v6.5.2` rather than `x`.

The multi-DEX plumbing already landed with Paragon (#12753) — asset-id offsets, positional universe cache, search `assetType` mapping, dex-scoped logos and deep-link encoding all read `SUB_DEX_LIST`. Registering a new dex is therefore a one-line change plus the surfaces that enumerate dexes by name.

Confirmed against the live `perpDexs` response that index `10` is `io` before baking the index in, since `SUB_DEX_LIST` is append-only and its array position is persisted in the `tradingUniverses` cache with no migration path.

## Root Cause — `io` shadowing `IOTA`

`io` is the first registered prefix that collides with an existing main-DEX symbol. `findDexPrefix` matches bare prefixes — that behaviour has to stay, because links minted before `encodeCoinForUrl` kept the separator look like `xyzNVDA`. Adding `io` makes it match `IOTA`, so `?token=IOTA` decodes to a guess of `io:TA`.

`resolvePerpCoinFromUrl` validates that guess against the universe and would recover `IOTA`. But when the cache is still incomplete — a cold-start deep link, which is precisely when this code path runs — it kept the unproven guess. `changeActiveAsset` then finds no such coin and silently leaves the previous market selected, so the user lands on Perps showing BTC with no error.

Checking every hyperliquid dex prefix against the main-DEX universe shows this is not a one-off: `cash` (index 7) shadows `CASHCAT` and would hit the same trap on registration.

## Root Cause — missing watchlist dex badge (OK-60689)

The Market watchlist renders perps rows through their own branch in the market columns (`record.perpsCoin ? ... : <TokenIdentityItem/>`). That branch carried `LeverageBadge` but never `PerpDexBadge`, so an `xyz` / `para` / `io` market was indistinguishable from a main-DEX one there, while every other perps surface shows it.

The watchlist has two renderers, and the branch above is only the wide one. Narrow web and native go through `MobileMarketWatchlistFlatList` → `TokenListItem` → `TokenIdentityItem` instead, which never touches the market columns at all.

`TokenIdentityItem` does render `PerpDexBadge`, and `TokenListItem` already forwards `maxLeverage` and `perpsSubtitle` — which is why those rows show the leverage badge and the company name but no dex tag. It just never passed `perpsDexLabel`, so the badge rendered with an undefined label and bailed out. Fixing only the columns therefore left every platform QA actually tested (mobile web, iOS, Android) unchanged.

## Design Decisions

**Kept bare-prefix matching, narrowed who is trusted without a universe.** Removing bare-prefix matching outright would break the `xyzNVDA` links that are actually out there. Instead, `MAIN_DEX_SHADOWED_DEX_PREFIXES` records which prefixes a real main-DEX symbol starts with — today only `io`. When the universe cannot settle the guess, a shadowed prefix falls back to the literal symbol (`IOTA`); every other prefix keeps the split guess, because its literal reading names no market at all.

When the universe is available (the normal path) nothing changes: every guess is still validated, so a hand-written `?token=paraUNITREE` still resolves.

`findDexPrefix` returns the narrowed `IPerpDexPrefix`, so the shadow list cannot drift from `SUB_DEX_LIST` without a type error. The main-DEX universe changes over time, so the constant carries a note to re-check a new prefix against it at registration — `cash` (dex 7) is already shadowed and would belong there.

**Alternatives rejected:**
- Always falling back to the literal token would undo `c04fd4e007`, which deliberately keeps the guess so `xyzNVDA` still resolves on a cold cache — the literal `XYZNVDA` is no market at all.
- Keying the fallback on *link history* (which prefixes ever shipped separator-free links, i.e. `xyz` alone) was the first attempt and is wrong: it agrees with shadowing on `xyz` and `io` but inverts `para`, turning an always-correct `para:UNITREE` into an always-broken `PARAUNITREE`. Caught in review; see 5df80fb67e.

**`assetTypeVersion` bumped per dex registration.** Search already drops an unregistered prefix on its own — `buildCoinFromSearchAssetType` returns undefined, the row is filtered out, and `UniversalSearchPerpItem` refuses to render a row it cannot open. But that guard does not cover every surface, so the version is what actually keeps a new dex away from older builds. `/utility/v2/market/perps/token-list` and `/utility/v1/perp-config` honour the same parameter, but neither request was sending it, so both were served the pre-`io` dataset while search — the only caller that sent it — could see `io`. That is what left the market list, the watchlist, the token selector and the pre-ipo tab empty. Both now send it, and the constant moved next to `SUB_DEX_LIST` so registering a dex and bumping the version are read together.

**Watchlist badge derived from `record.perpsCoin`**, which already holds the full prefixed coin (`symbol` holds the bare display name). That keeps it registry-agnostic — `xyz`, `para`, `io` and any future dex light up with no further change. Both renderers derive it the same way: the columns for the wide table, `TokenListItem` for the narrow/native list.

**URL codec extracted.** `encodeCoinForUrl` / `decodeCoinFromUrl` moved to a `.utils` sibling so they are unit-testable without pulling in `backgroundApiProxy` and react-navigation, matching the existing `PerpsGlobalEffects.utils` split in the same directory.

## Changes Detail

- `packages/shared/types/hyperliquid/perp.constants.ts` — append `{ prefix: 'io', hlDexIndex: 10 }`; add `IPerpDexPrefix` and `MAIN_DEX_SHADOWED_DEX_PREFIXES`.
- `packages/kit/src/views/Perp/components/usePerpTokenUrlSync.utils.ts` — new; hosts the URL codec, with `unverifiedFallbackCoin` carrying the per-prefix fallback.
- `packages/kit/src/views/Perp/components/usePerpTokenUrlSync.ts` — consume the extracted codec; use the fallback when the universe stays incomplete or the lookup throws.
- `packages/kit/src/views/Market/components/PerpsBadges.tsx` — `case 'io'` in `getPerpDexDescriptionId`.
- `packages/shared/types/hyperliquid/perp.constants.ts` — `PERPS_ASSET_TYPE_VERSION` 2 → 3, moved here from the search service so every caller shares it.
- `packages/kit-bg/src/services/ServiceUniversalSearch.ts` — consume the shared constant.
- `packages/kit-bg/src/services/ServiceMarketV2.ts` + `ServiceHyperLiquid/ServiceHyperliquid.ts` — send `assetTypeVersion` on `token-list` and `perp-config`, which were silently getting the pre-`io` dataset.
- `.../MarketTokenList/hooks/useMarketTokenColumns/useColumnsDesktop.tsx` + `useColumnsMobile.tsx` — add `PerpDexBadge` to the watchlist perps branch (wide table; `useColumnsMobile` is also what Swap Pro's token picker uses when narrow).
- `.../MarketTokenList/components/TokenListItem.tsx` — pass `perpsDexLabel` through to `TokenIdentityItem`, which covers the narrow-web and native watchlist.
- `packages/shared/src/locale/**` — `perp_io_market__desc` across all 19 locales (#13034, by Kahn).
- Tests — new coverage for the URL codec; existing dex tests updated for the fourth registry slot.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Web (URL codec is a no-op on Mobile/Desktop/Extension — those variants are empty stubs); all platforms for the registry and badge.
- **Risk Areas**:
  - `SUB_DEX_LIST` is append-only and positional; the new slot invalidates every persisted `tradingUniverses` cache, forcing one universe refresh on the next cold start. `isPerpsUniverseCacheComplete` already handles this by rejecting a short cache.
  - Each registered dex adds one `clearinghouseState` request plus `openOrders` and `twapStates` websocket subscriptions per user (3 dexes to 4).
  - Users who have not received the hot update cannot resolve `io:` links at all — the client-side registry is the only source of the dex list.

## Dependencies

- Server: `io` markets in `/utility/v2/market/perps/token-list`, `assetType: 'io'` in the universal search index, and `io` aliases in `/utility/v1/perp-config`.
- Assets: dex-scoped logos at `hyperliquid/io<SYMBOL>.png`. There is deliberately no bare-symbol fallback (it would render another dex's asset), so a missing file renders blank.
- ~~i18n: create `perp_io_market__desc` in Lokalise~~ — done, all 19 locales via #13034.

## Test plan

- [ ] Entropy markets (`io:ANTH`, `io:SNDK`, `io:NBIS`) appear in the market list and token selector once the server serves them.
  - `io:OAI`, `io:IONQ` and `io:EWY` are flagged `isDelisted` on Hyperliquid, and `buildTradablePerpMaxLeverageMap` drops delisted assets — them not appearing is correct, not a bug.
  - Blocked as of 2026-08-31: `/utility/v2/market/perps/token-list` returns zero `io:` rows on both test and prod, across every category, so no `io` market can render anywhere yet.
  - Also pending: `perp-config` carries no `io:` search aliases (so searching `io` only surfaces `IONQ`, which matches on its ticker), and `ioOAI.png` / `ioIONQ.png` / `ioNBIS.png` are 404 — there is no bare-symbol fallback, so those render blank.
- [ ] Universal search returns Entropy results and routes to the right market.
- [ ] Favoriting an `io:` market persists and reopens correctly.
- [ ] Deep link / URL `?token=io:ANTH` opens Entropy ANTH.
- [ ] Web: `?token=IOTA` opens main-DEX IOTA, including on a cold start with no cached universe.
- [ ] Web: a legacy `?token=xyzNVDA` link still opens `xyz:NVDA` on a cold start.
- [ ] Web: `?token=paraUNITREE` still opens `para:UNITREE` on a cold start.
- [ ] The `io` badge renders with its localized description (hover on web, tap on native), in a non-English locale too.
- [ ] Market watchlist (自选 → 合约): an `xyz` / `para` / `io` row shows its dex badge next to the leverage badge — check all four renderers: desktop web, narrow web, iOS and Android.

[OK-61073]: https://onekeyhq.atlassian.net/browse/OK-61073?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-60689]: https://onekeyhq.atlassian.net/browse/OK-60689?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit/src/views/Market/components/PerpsBadges.tsx
```diff
@@ -47,6 +47,13 @@ function getPerpDexDescriptionId(dexLabel?: string) {
   }
 }
 
+// TODO(OK-61073): replace with ETranslations.perp_io_market__desc once the key
+// is created in Lokalise. Hardcoded English ships the badge in the 6.5 hot
+// update, which cannot wait for a translation round.
+const UNTRANSLATED_PERP_DEX_DESCRIPTIONS: Record<string, string> = {
+  io: 'io identifies markets from the Entropy HIP-3 DEX on Hyperliquid, including pre-IPO AI companies and listed technology equities.',
+};
```
sidmorizon: **Hardcoded English DEX badge description**

Severity: `non-severe`

The `io` badge description is a hardcoded English string rendered directly in the tooltip, not a translation key via `intl.formatMessage`. The repository i18n rule requires translation keys instead of hardcoded strings, so non-English users see English regardless of locale.

packages/kit/src/views/Perp/components/usePerpTokenUrlSync.utils.ts
```diff
@@ -0,0 +1,91 @@
...
+  const dexPrefix = findDexPrefix(urlToken);
+  if (dexPrefix && urlToken.length > dexPrefix.length) {
+    const hasNoSeparator = !urlToken.includes(DEX_SEPARATOR);
+    const symbolStartIndex = hasNoSeparator
+      ? dexPrefix.length
+      : dexPrefix.length + DEX_SEPARATOR.length;
+    const symbol = urlToken.slice(symbolStartIndex);
+    const coin = `${dexPrefix}${DEX_SEPARATOR}${symbol.toUpperCase()}`;
+    const isTrustedWithoutUniverse =
+      !hasNoSeparator || LEGACY_SEPARATOR_FREE_DEX_PREFIXES.includes(dexPrefix);
```
sidmorizon: # 🟡 P2 `LEGACY_SEPARATOR_FREE_DEX_PREFIXES` 把 `para` 的兜底方向反转了

## 问题

`isTrustedWithoutUniverse` 用「该前缀历史上是否产出过无分隔符链接」(`LEGACY_SEPARATOR_FREE_DEX_PREFIXES = ['xyz']`) 来决定 universe 不可用时的兜底值。但这里真正需要判断的是另一件事：**字面读法是否可能命中一个真实的主 DEX 符号**。PR 描述自己已经把碰撞清单列全了 —— 只有 `io`(遮蔽 `IOTA`) 和 `cash`(遮蔽 `CASHCAT`) 存在碰撞，没有任何主 DEX 符号以 `para` 开头。

两个判据在 `xyz` / `io` 上恰好同解，在 `para` 上却相反：`unverifiedFallbackCoin` 从改前必定正确的 `para:UNITREE`，翻成了永远不存在的 `PARAUNITREE`。

```ts
// 改前：!universesByDex / catch 都 return coin  -> 'para:UNITREE'（可以打开）
// 改后：两条分支都 return unverifiedFallbackCoin -> 'PARAUNITREE'（无此市场）
```

## 影响

触发条件是「裸前缀链接 + universe cache 不完整」。而本 PR 把 `SUB_DEX_LIST` 扩到 4 槽后，`isPerpsUniverseCacheComplete` 会拒绝**所有**历史缓存，热更后每个用户的首次冷启动都正好落在这个窗口内；若此时 `refreshTradingMeta()` 再失败或抛错，就直接命中 `catch` 分支。

结果是 `changeActiveAsset('PARAUNITREE')` 找不到市场、静默停留在上一个（默认 `BTC`）行情 —— 正是本 PR 想修掉的那个失败形态，只是从 `io` 挪到了 `para` 上。范围限于手写/外部链接（`encodeCoinForUrl` 自 #12753 起对 `para` 一直带分隔符），所以定 P2 而非 P1。

另外 `usePerpTokenUrlSync.utils.test.ts:48-52` 把 `unverifiedFallbackCoin: 'PARAUNITREE'` 断言成了期望值，等于把这个行为锁进了测试。

## 建议

把判据换成「该前缀是否会被真实主 DEX 符号遮蔽」，只让 `io`（以及将来的 `cash`）退回字面读法：

```ts
// 只有会被主 DEX 符号遮蔽的前缀，字面读法才比猜测更可信
export const MAIN_DEX_SHADOWING_DEX_PREFIXES: readonly string[] = ['io'];

const isTrustedWithoutUniverse =
  !hasNoSeparator || !MAIN_DEX_SHADOWING_DEX_PREFIXES.includes(dexPrefix);
```

这样 `xyz` / `para` 保持猜测、`io` 退回 `IOTA`，三者都正确。顺带建议把该常量的元素类型约束成 `(typeof DEX_PREFIXES)[number]`，避免它和 `SUB_DEX_LIST` 脱钩后靠字符串默默失效；同时更新上述测试断言。

[Claude session ↗](https://claude.ai/code/session_01RG2M3PkaKtvPcGnKDsumSn)

---

Minnzen: Confirmed and fixed in 5df80fb67e — the criterion was wrong, and `para` was the case that exposed it.

The fallback has to answer "which reading could name a real market", and I keyed it on "did this prefix ever ship separator-free links". Those agree on `xyz` and `io` and disagree on `para`: nothing on the main DEX starts with `PARA`, so `PARAUNITREE` resolves to nothing while `para:UNITREE` resolves. I turned an always-right value into an always-wrong one.

Your point about the window is the part I had underweighted. I had filed this as "hand-written links only, and only on a cold cache", but this PR is what makes that cache cold: the fourth registry slot rejects every persisted `tradingUniverses`, so the first cold start after the hot update runs with an incomplete universe for every user.

Changed to `MAIN_DEX_SHADOWED_DEX_PREFIXES = ['io']` keyed on shadowing, so `xyz` and `para` keep the guess and only `io` falls back to the literal. Also took the typing suggestion — added `IPerpDexPrefix = (typeof DEX_PREFIXES)[number]` and narrowed `findDexPrefix`'s return, so the list cannot silently drift from `SUB_DEX_LIST`.

The test that locked in `'PARAUNITREE'` is now split into two cases: `IOTA` falls back to the literal, `paraUNITREE` keeps `para:UNITREE`.

Left a note on the constant to re-check a new prefix against the main-DEX universe when registering it — `cash` (dex 7) is already shadowed and would belong there.

packages/kit/src/views/Market/components/PerpsBadges.tsx
Minnzen: Intentional, not an oversight — flagged as `TODO(OK-61073)` on the constant.

The repo i18n rule holds, but the Lokalise round-trip would block this hot update, and `SUB_DEX_LIST` only reaches users through the bundle. The call was to ship the badge with English copy now and swap in `perp_io_market__desc` afterwards. `xyz` and `para` still go through `ETranslations`; only the new `io` entry sits in the hardcoded map, so removing it later is a one-line delete.

Tracked in the PR description under Dependencies.

Minnzen: Follow-up: the Lokalise key is being handed to a teammate rather than done here, so the hardcoded string stands as the reference copy for it. Leaving this thread open as the tracker.

For whoever picks it up — key `perp_io_market__desc`, `__desc` so sentence case, 19 locales:

- en_US: `io identifies markets from the Entropy HIP-3 DEX on Hyperliquid, including pre-IPO AI companies and listed technology equities.`
- zh_CN: `io 表示来自 Hyperliquid 上 Entropy HIP-3 DEX 的市场，涵盖 pre-IPO 的 AI 公司和已上市的科技股。`

Wording is derived from what `io` actually lists today (OAI, ANTH, SNDK, IONQ, NBIS — two pre-IPO AI names plus listed tech), matching how `perp_xyz_market__desc` and `perp_para_market__desc` describe their own universes. If there is official Entropy positioning copy, prefer that.

Once the key exists: delete `UNTRANSLATED_PERP_DEX_DESCRIPTIONS` and add `case 'io'` back to `getPerpDexDescriptionId` — one line each.

packages/kit-bg/src/services/ServiceUniversalSearch.ts
```diff
@@ -56,7 +56,11 @@ import ServiceBase from './ServiceBase';
 const PERPS_UNIVERSE_SEARCH_MAX_AGE_MS = timerUtils.getTimeDurationMs({
   minute: 5,
 });
-const PERPS_ASSET_TYPE_VERSION = 2;
+// Bumped per dex registration so the server can keep a new dex out of responses
+// to clients that predate it: an unregistered prefix is dropped by search, but
+// the market list renders whatever `token-list` returns and would navigate to a
+// coin this build cannot resolve. v3 adds `io` (Entropy).
+const PERPS_ASSET_TYPE_VERSION = 3;
```
sidmorizon: # 🟡 P2 版本号与 `SUB_DEX_LIST` 无任何绑定，且注释所述的 `token-list` 面并不携带该版本

## 问题

`PERPS_ASSET_TYPE_VERSION` 定义在 `packages/kit-bg/src/services/ServiceUniversalSearch.ts`，而 dex 注册表 `SUB_DEX_LIST` 在 `packages/shared/types/hyperliquid/perp.constants.ts`。两者必须同步递增（v2 对应 `xyz`/`para`，v3 才对应新增的 `io`），但没有任何类型、断言或测试把它们绑在一起；`SUB_DEX_LIST` 上方的 `APPEND ONLY` 注释也只提到位置索引被 `tradingUniverses` 缓存固化，完全没有提"注册新 dex 必须同时 bump 版本号"。

另外，全仓库 `assetTypeVersion` 只在一处发出（本文件第 `1294` 行的 `/wallet/v1/proxy/hyperliquid/perpsAsset`）。注释用来论证 bump 必要性的那个面——market 列表——走的是 `ServiceMarketV2.fetchMarketPerpsTokenList` / `fetchMarketBannerPerpsTokenList`（`/utility/v2/market/perps/token-list`、`/utility/v2/market/banner/perps-token-list`），这两个请求不带任何版本参数：

```ts
const response = await client.get<IMarketPerpsTokenListResponse>(
  '/utility/v2/market/perps/token-list',
  { params: params?.category ? { category: params.category } : undefined },
);
```

## 影响

对当前 PR 没有功能缺陷：新客户端本地已注册 `io`，返回什么都能解析。但注释把 `token-list` 面写成受该版本保护，实际这个常量到不了那个面（服务端只能靠 `X-Onekey-Request-JSBundle-Version` 之类的全局信号，服务端代码不在本仓库、无法验证）。后续维护者可能据此误判覆盖范围。

真正的风险在下一次注册 dex：只改 `SUB_DEX_LIST` 而忘记 bump 版本时，代码能编译、测试全绿（唯一断言在 `ServiceUniversalSearch.perpsAssetCapability.test.ts` 里硬编码了字面量 `3`，不会因新增 dex 而失败），旧版本号会让 search 面继续拿不到新 dex，而故障表现只是"搜不到"这种难以归因的静默行为。

## 建议

把版本号与注册表放在同一处并显式表达耦合，例如挪到 `perp.constants.ts` 并由 `SUB_DEX_LIST` 推导（当前 `PERPS_ASSET_TYPE_VERSION === SUB_DEX_LIST.length` 恰好成立），或至少保留字面量但加一条断言测试：

```ts
// perp.constants.ts
// APPEND ONLY: ... 追加新 dex 时必须同步 bump PERPS_ASSET_TYPE_VERSION。
export const PERPS_ASSET_TYPE_VERSION = SUB_DEX_LIST.length;
```

同时把本行注释里关于 `token-list` 的表述限定为"该面另有服务端灰度手段"，避免声称本参数覆盖了它。

[Claude session ↗](https://claude.ai/code/session_01RG2M3PkaKtvPcGnKDsumSn)

---

sidmorizon (commented):
Round 3 incremental review.

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenColumns/useColumnsMobile.tsx
```diff
@@ -84,6 +86,9 @@ export const useColumnsMobile = (
                     {record.maxLeverage ? (
                       <LeverageBadge leverage={record.maxLeverage} />
                     ) : null}
+                    <PerpDexBadge
+                      dexLabel={parseDexCoin(record.perpsCoin).dexLabel}
```
sidmorizon, line 90: # 🟠 P1 移动端/原生 Market 自选列表仍然看不到 dex badge

## 问题

commit message 说 “Native delegates to the mobile columns, so both platforms are covered”，但 Market 首页自选 tab 在移动端根本不走 `useColumnsMobile`。`MarketHomeV2.tsx` 在 `md || platformEnv.isNative` 时渲染 `MobileLayout`，其自选 tab 用的是 `MobileMarketWatchlistFlatList`（`MobileLayout.tsx:435` / `MobileLayout.native.tsx:817`），该列表逐行渲染 `TokenListItem` 而不是表格列。

`TokenListItem.tsx` 把 `maxLeverage`、`perpsSubtitle` 透传给了 `TokenIdentityItem`，唯独没有传 `perpsDexLabel`：

```tsx
// packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/components/TokenListItem.tsx:126
          maxLeverage={item.maxLeverage}
          perpsSubtitle={item.perpsSubtitle}
          // perpsDexLabel 缺失 -> TokenIdentityItem 内的 PerpDexBadge 恒为 null
```

`TokenIdentityItem` 内部 `<PerpDexBadge dexLabel={perpsDexLabel} />` 在 `dexLabel` 为 `undefined` 时直接 `return null`，所以杠杆 badge 显示、dex badge 不显示——正是本次要修的那个不对称。经过 `useMarketTokenColumns` 的 `MarketWatchlistTokenList` 只在 `DesktopLayout` 以及 `SwapProSelectTokenModal` / `MarketDetailV2` 的 `MobileTokenSelector` 这几个选择器里出现，不是首页自选 tab。

## 影响

iOS / Android 以及 `md` 宽度的 web、插件端，Market 自选列表里的 `io:OAI`、`xyz:NVDA`、`para:UNITREE` 依旧和主 DEX 行长得一模一样，commit message 描述的风险（“an xyz or para market was indistinguishable from a main-DEX one”）在移动端完全没有被消除。本次两个文件里 mobile 那一份改动实际只作用于代币选择器弹窗，覆盖不到用户使用频率最高的自选入口。

## 建议

在 `TokenListItem` 里补上透传，`MobileMarketTokenFlatList` / `BannerDetailTokenFlatList` 等复用方在 `perpsCoin` 为空时自然拿到 `undefined`，不受影响：

```tsx
perpsSubtitle={item.perpsSubtitle}
perpsDexLabel={
  item.perpsCoin ? parseDexCoin(item.perpsCoin).dexLabel : undefined
}
```

这样也顺带让 `TokenIdentityItem.perpsDexLabel` 这条已有链路在自选场景真正生效，与 `MarketPerpsTokenListItem` / `usePerpsColumnsMobile` 的取值口径（同样是 `parseDexCoin(name).dexLabel`）保持一致。

[Claude session ↗](https://claude.ai/code/session_01RG2M3PkaKtvPcGnKDsumSn)

---

sidmorizon (commented):
Round 4 incremental review.

---

packages/kit/src/views/Perp/components/TradingPanel/panels/LimitOrderForm.tsx
```diff
@@ -425,15 +427,19 @@ export function LimitOrderForm({
   }, [isSpot, spotHoldingBaseBN, spotUniverse?.baseName]);
   const sideStats = useMemo(() => {
     const buildStats = (targetSide: ITradeSide) => {
+      const sidePriceBN = resolvePriceForSide(targetSide).price;
+      const sideSizeBN = computeSizeBN(targetSide, sidePriceBN);
+      const sideOrderValueBN = sideSizeBN.multipliedBy(sidePriceBN);
```
sidmorizon, line 432: **BBO estimates freeze after first quote**

Severity: `non-severe`

With BBO pricing enabled, `sideStats` excludes live quotes from its memo dependencies. Button estimates remain stale while executable prices change.

originalix (approved):

PR #13035, opened 2026-08-27
fix: stabilize Market and Swap token state (OK-61075, OK-61322, OK-61334)

zhaono1:
[OK-61075](https://onekeyhq.atlassian.net/browse/OK-61075) [OK-61322](https://onekeyhq.atlassian.net/browse/OK-61322) [OK-61334](https://onekeyhq.atlassian.net/browse/OK-61334)

----------

## Issues
- Fixes OK-61075
- Fixes OK-61322
- Fixes OK-61334

## Summary
- scope the Market selected derive type to the network where it was chosen
- ignore stale BTC derive selections when resolving accounts for another network
- defer shared Swap default-token synchronization while Native Pro owns the token UI
- re-check the active owner after the asynchronous Home-account synchronization
- keep the Market payment-token list on a skeleton until current-account balances are ready and value sorting is final
- preserve the fixed server order used by Swap Pro and Stock token selectors
- add focused unit coverage for both state-ownership boundaries

### OK-61075
Market Detail stored the selected derive type without its network identity. After selecting BTC BIP84 and switching the page token to an EVM asset, the stale BTC derive type could be reused for EVM account resolution. The lookup then returned no account and the balance was cleared.

### OK-61322
Native Pro uses the `LIMIT` tab type but owns separate Pro token atoms. `useSwapGlobal` still validated the parked Swap From/To atoms against the active `LIMIT` channel. When a parked Swap token network did not advertise `supportLimit`, the default-token effect treated the valid Swap pair as unsupported and cleared both tokens. The effect timing made the issue intermittent during repeated Pro/Swap tab switches.

### OK-61334
The Market payment-token popover rendered its real rows before asynchronous balances and fiat values were authoritative. It first showed the server-config order, then visibly reordered the same rows by asset value after the requests settled. The list now stays on the existing token-selector skeleton until both the network account and token details belong to the current account and have finished loading.

## Test plan
- [x] `yarn agent:check --profile commit`
- [x] `yarn jest packages/kit/src/states/jotai/contexts/marketV2/marketDeriveType.test.ts --runInBand`
- [x] Desktop cold-run: select BTC Native SegWit, switch to Ethereum DOGE, open Sell, and confirm the DOGE balance remains visible
- [x] 155 focused Swap tests across token-cache, actions, and header suites
- [x] iOS cold-start regression with Robinhood ETH / Bitcoin Cash BCH: 8 rapid Pro/Swap cycles preserved both tokens
[...]

PR #13037, opened 2026-08-27
fix: preserve Swap tokens across Pro tab switches (OK-61322)

zhaono1:
## Summary

- defer shared Swap default-token synchronization while Native Pro owns the token UI
- re-check the active owner after the asynchronous Home-account synchronization
- cover Native Pro, desktop/web Limit, and Native Swap owner boundaries

## Root cause

Native Pro uses the `LIMIT` tab type but owns separate Pro token atoms. `useSwapGlobal` still validated the parked Swap From/To atoms against the active `LIMIT` channel. When a parked Swap token network did not advertise `supportLimit`, the default-token effect treated the valid Swap pair as unsupported and cleared both tokens. The effect timing made the issue intermittent during repeated Pro/Swap tab switches.

## Validation

- `yarn jest packages/kit/src/views/Swap/utils/swapColdStartTokenCacheUtils.test.ts --runInBand`
- `yarn jest packages/kit/src/states/jotai/contexts/swap/actions.test.tsx --runInBand`
- `yarn jest packages/kit/src/views/Swap/pages/components/SwapHeaderContainer.test.tsx --runInBand`
- iOS cold-start regression with Robinhood ETH / Bitcoin Cash BCH: 8 rapid Pro/Swap cycles preserved both tokens
- Hermes destructive-write breakpoint: 16 clear calls before the fix, 0 after the fix for the same run

Jira: https://onekeyhq.atlassian.net/browse/OK-61322

### DIFF
--- a/packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
+++ b/packages/kit/src/views/Swap/hooks/useSwapGlobal.ts
@@ -74,6 +74,7 @@ import {
   getSwapSelectedTokensHomeAccountSyncAction,
   isSwapColdStartAllNetworkContextNetworkId,
   isSwapSelectedTokensColdStartContextMatched,
+  shouldDeferSwapDefaultSelectedTokenSyncForNativePro,
   shouldMarkSwapInitialSelectedTokensSynced,
   shouldPreserveSwapUserInputAmountOnAccountSwitch,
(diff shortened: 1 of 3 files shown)


zhaono1:
Superseded by #13035. The OK-61322 fix was migrated to the existing Draft PR head as commit 3ab56d4316.

PR #13042, opened 2026-08-27
fix: scope prime transfer c2c rate-limit state per connection

sidmorizon:
## Summary
Part of a security review of the Prime Transfer E2EE flow (server-side changes live in the e2ee-server repo). This fixes a client-side memory leak in the client-to-client bridge&#39;s rate limiter.

## Change
`JsBridgeE2EEClientToClient` kept rate-limit state in a **module-level** `Map` keyed by `${socket.id}:${event}:${method}`, never pruned. socket.io assigns a fresh `socket.id` on every reconnect, so the map grew by one entry per method on every reconnect for the lifetime of the process.

Moved the state onto the bridge instance and dropped `socket.id` from the key, so it is bounded by method name, stable across reconnects, and freed with the instance. No socket `disconnect` listener is added (unlike the server): the client socket is long-lived and the bridge is re-created on it (e.g. QR refresh → joinRoom), where `setup()` already detaches the superseded instance&#39;s listener so it can be GC&#39;d — a disconnect listener would pin the old instance instead.

Follow-ups from review of that same function:

- **Per-instance bound on the map.** The key still ends in `req.method`, which arrives from the remote peer over the c2c channel, so a peer could grow the map without limit by varying the method name. Tracked methods are now capped at 64 per bridge instance (`RATE_LIMIT_MAX_TRACKED_METHODS`), mirroring the server side of this same bridge. On overflow only entries whose window has already expired are reclaimed (`pruneRateLimitState`); if every window is still live the call is treated as rate limited instead of tracking a new one. Pruning expired entries only is deliberate — wiping the map on a flood would let the flooder reset the windows of the calls it was just blocked on, turning the bound into a rate-limit bypass. `lastTime` is now read with an `undefined` check rather than `|| 0` so &quot;never seen&quot; is distinguishable from &quot;seen&quot; (behaviourally identical otherwise, since `Date.now()` is never 0).
- **Null-safe method extraction.** The function read `payload.data.method` directly, so a room peer emitting `e2ee-c2c-request` with missing or non-object `data` made the property read throw inside an async socket.io listener — an unhandled rejection on the client. The method is now extracted as `typeof req?.method === 'string' ? req.method : ''` and that local is used for both the whitelist check and the key.

Interval (3.5s), whitelist contents and rate-limit semantics for well-formed traffic are unchanged.

## Testing
Repo checks (`yarn agent:check --profile commit`) could not be run in this environment — `node_modules` is not installed there — so CI lint, typecheck and the unit-test matrix are the authoritative validation for this push.

What was verified locally:
- The changed file type-checks clean under an isolated `tsc --noEmit` harness using the real `@onekeyfe/cross-inpage-provider-*` and `socket.io` type definitions (from a sibling checkout) with the repo&#39;s compiler options; only the two `@onekeyhq/shared` locale imports were stubbed.
- A standalone Node script re-running the final rate-limit logic: a repeat call inside the window is limited and sends the error response; a call after the window elapses is not (and 1 ms before the window ends still is); a flood of 500 distinct method names keeps the map at ≤ 64 entries, throttles itself, and leaves the live window of a previously blocked method intact; once all windows expire, pruning frees the map and new methods are accepted again; nine malformed payload shapes (`data` missing/null/string/number/array, `method` missing/number/object/null) do not throw and collapse onto a single empty-method key; whitelisted methods are still never limited or tracked. All 20 checks passed.

Rate-limiter behavior is unchanged (per-connection, per-method, 3.5s window). No functional change to pairing or transfer.

https://claude.ai/code/session_01AuowqYp863fJaNNjEKvhPT

---
_Generated by [Claude Code](https://claude.ai/code/session_01AuowqYp863fJaNNjEKvhPT)_

### DIFF
--- a/packages/kit-bg/src/services/ServicePrimeTransfer/e2ee/JsBridgeE2EEClientToClient.ts
+++ b/packages/kit-bg/src/services/ServicePrimeTransfer/e2ee/JsBridgeE2EEClientToClient.ts
@@ -12,6 +12,11 @@ import type { Socket } from 'socket.io';
 
 const RATE_LIMIT_INTERVAL_MS = 3500;
 
+// Upper bound on distinct methods tracked per connection. `method` comes from
+// the remote peer, so without a cap a single peer could grow this map forever.
+// Well above the number of methods a real peer calls.
+const RATE_LIMIT_MAX_TRACKED_METHODS = 64;
+
 // Rate limiting whitelist - methods that are exempt from rate limiting
 const RATE_LIMIT_WHITELIST = new Set([
   'changeTransferDirection',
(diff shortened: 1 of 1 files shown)


sidmorizon:
CI note — the one red check, **Web startup graph budget (24.x)**, is a marginal timing flake, not a regression from this PR.

All 21 other checks pass (lint, typecheck, the full unit-test matrix, native/extension build budgets, CodeQL, and the security scans). The single hard failure is one web cold-start timing metric on the `root` scenario:

```
[perf:web:cold] root budgets
FAIL fcpMs: 1052 ms / 1000 ms (fail > 1050 ms)
```

First Contentful Paint measured **1052 ms vs the 1050 ms hard-fail threshold — 2 ms (0.2%) over**, and FCP varies with CI-runner load. The static startup-graph budget in the same job **passed** (`=== WEB STARTUP GRAPH BUDGET CHECK PASSED ===`, `largestModuleBytes 502.9 KiB / 800.0 KiB`), so the startup bundle did not grow.

This PR's only change is a single `kit-bg` file that moves a rate-limiter `Map` from module scope to instance scope with no import changes — it cannot affect web First Contentful Paint or the startup module graph. I've re-run the failed job once to confirm it clears.

---

PR #13059, opened 2026-08-28
fix: simplify Reanimated web patch OK-61366

huhuanming:
## Summary

- Follow up on #13056 by reducing the Reanimated Web patch to the minimal static-import change.
- Preserve the upstream mutable exports while initializing them from statically bundled React Native Web helpers.
- Keep the original dynamic fallback code in one documented block comment, as required by the repository patch-package workflow.

## Verification

- Patch applies cleanly to a pristine `react-native-reanimated@4.5.1` package.
- Patch contains no `android/build` artifacts.
- Rspack compiles and resolves both React Native Web helpers as bundled modules.
- Electron verification: transition container is `290px`, all 12 phrase inputs render, and the first input receives focus.

## Scope

No runtime behavior change from #13056; this only makes the dependency patch smaller and easier to review.


huhuanming:
Closing this follow-up: the cleanup is not required for the already merged OK-61366 runtime fix.

PR #13074, opened 2026-08-30
fix: keep WalletConnect namespace events an array on approve

### LINKED ISSUE #13073: [BUG]: WalletConnect approve fails — isValidApprove() rejects namespace, dApp sees "User rejected"
**Prerequisites**

- [x] I checked to make sure that this issue has not already been filed

**Describe the bug**

Approving a WalletConnect v2 session request fails inside the wallet. When the user taps **Approve** on the "Connection request" screen, OneKey calls the WalletConnect `approve()` with a namespace whose `events` field is not a valid array of strings. The SDK validator `isValidApprove()` rejects the call, the session is never established, and the dApp reports `User rejected the request` — even though the user actually approved it.

From the user's point of view: you tap Approve, the wallet returns to the home screen, and the dApp shows a rejection popup.

Captured from `adb logcat` at the exact moment of the tap (28 ms after the touch event):

```
15:00:44.451 22406 22406 I GestureDetector: obtain mCurrentDownEvent   <-- tap "Approve"
15:00:44.478 22406 22501 E ReactNativeJS: { context: 'client' }, 'approve() -> isValidApprove() failed'
15:00:44.479 22406 22501 E ReactNativeJS: { context: 'client' }, 'Unsupported events. approve(), namespace, events should be an array of strings or empty array for no events'
```

No crash occurs, the crash buffer stays empty, and no error is surfaced in the UI — the failure is silent.

**To Reproduce**

1. Open `https://lab.reown.com/appkit?name=siwx-default` (Reown's own AppKit Lab)
2. Click **Connect Wallet** under `EVM BUTTON` → **WalletConnect** → copy the `wc:` URI
3. Open that URI in OneKey (scan the QR, or `adb shell am start -a android.intent.action.VIEW -d 'wc:...' so.onekey.app.wallet`)
4. The "Connection request" screen appears — tap **Approve**
5. The wallet returns to the home screen; the dApp shows `User rejected the request` and falls back to the wallet-selection screen

**Control case (works):** the exact same flow on `https://lab.reown.com/appkit?name=ethers` connects successfully. Same AppKit build, same SDK version — the only difference is the requested namespace set, which isolates this to namespace handling rather than to connectivity or to a specific dApp.

**Expected behavior**

Tapping Approve should establish the session. The namespace passed to `approve()` should always carry `events` as an array of strings (an empty array when there are no events), so that `isValidApprove()` passes.

If approval genuinely cannot be completed, the wallet should surface the error to the user instead of failing silently and letting the dApp report a rejection the user never made.

**Related symptom (likely the same root cause)**

A dApp that requests One-Click Auth — a pairing URI containing `methods=wc_sessionAuthenticate` — shows **no approval screen at all**. The QR scanner closes, the app returns to the home screen, and nothing further happens. The pairing itself succeeds, so the URI is parsed correctly; the authenticate request is simply never surfaced.

**Smartphone (please complete the following information):**
 - Device: Samsung Galaxy S25 Ultra (SM-S938B)
 - OS: Android 16
 - App: OneKey Wallet 6.5.3 (versionCode 2026082287), installed from Google Play
 - Reproduces on at least 6.5.0 through 6.5.3

**Additional context**

Ruled out during debugging, so these can be skipped:

- **Network/DNS** — `relay.walletconnect.org` resolves correctly and answers `401 {"error":"JWT is missing"}` over TLS from the device; the app transfers megabytes normally
- **Crashes** — crash buffer is empty; the process is never killed
- **Camera/scanner** — reproduces identically when the URI is delivered as an `Intent` via ADB, bypassing the scanner entirely
- **Permissions** — CAMERA and INTERNET are both granted

A useful note for anyone reproducing this: the release build does emit JS errors to logcat under the `ReactNativeJS` tag, so `adb logcat | grep ReactNativeJS` is enough to observe the failure.

kjljxybr:
## Summary

- `buildWalletConnectNamespace` casts every key from the proposal to `INamespaceUnion`, but `supportEventsMap` only defines `eip155` and `algorand`. Any other namespace a dApp proposes (`solana`, `bip122`, `tron`) resolves to `undefined`.
- `approveSession()` then fails `isValidApprove()` with `Unsupported events`, so the session is never established.
- `onSessionProposal` reported that failure to the dApp as `USER_REJECTED` — which is how a user who tapped **Approve** ends up seeing "user rejected the request".
- `events` now falls back to `[]`, matching the `methods` line directly above it.
- A failure that happens after the user approved is now reported as `SESSION_SETTLEMENT_FAILED` instead of being attributed to the user.

Fixes #13073

## Runtime impact

- Background runtime only (`kit-bg`); no UI, JS bundle layout or native changes.
- Behaviour changes only on the path that previously threw: proposals carrying a namespace outside `eip155`/`algorand` now settle instead of failing. Proposals with a single `eip155` namespace are unaffected, since `supportEventsMap.eip155` was already defined.
- The `SESSION_SETTLEMENT_FAILED` branch is reachable only after the approval modal resolves, so a genuine rejection by the user still reports `USER_REJECTED`.

## Validation

Reproduced on a physical device against the released build, before the change:

- Samsung SM-S938B, Android 16, OneKey 6.5.3 (versionCode 2026082287) from Google Play
- `https://lab.reown.com/appkit?name=siwx-default` (multi-namespace proposal) → tap **Approve** → the dApp reports `user rejected the request`
- logcat at the moment of the tap, 27 ms after the touch event:
  ```
  E ReactNativeJS: 'approve() -> isValidApprove() failed'
  E ReactNativeJS: 'Unsupported events. approve(), namespace, events should be
                    an array of strings or empty array for no events'
  ```
- `https://lab.reown.com/appkit?name=ethers` (single `eip155` namespace) connects cleanly on the same build — the contrast between the two is what pointed at `supportEventsMap` coverage
- Ruled out separately: network and relay (`relay.walletconnect.org` answers `401 {"error":"JWT is missing"}` over TLS from the device), crashes (crash buffer empty), the QR scanner (identical symptom when the URI is delivered through `am start`), and permissions

**Not verified:** I have not built the app or run the test suite against this branch — the repro and the root cause were established on the released build and by reading the source. The patched binary still needs a run by someone with the toolchain set up.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -237,7 +243,9 @@ class ProviderApiWalletConnect {
       console.error('onSessionProposal error: ', e);
       await this.web3Wallet?.rejectSession({
         id: proposal.id,
-        reason: getSdkError('USER_REJECTED'),
+        reason: userApproved
+          ? getSdkError('SESSION_SETTLEMENT_FAILED')
+          : getSdkError('USER_REJECTED'),
```
sidmorizon: **Save failures leave unusable sessions active**

Severity: `non-severe`

When `saveConnectionSession` fails after approval, `sessionSettled` suppresses cleanup. The dApp stays connected, but later requests lack the saved account.

kjljxybr: Good catch — fixed in 9ab1e3e.

`approveSession()` consumes the proposal, so any failure after it (the `saveConnectionSession` call being the realistic one) was rejecting a proposal that no longer exists, while the session itself stayed live. The rejection is now gated on a `sessionSettled` flag set right after `approveSession()` resolves, so only a still-pending proposal is ever rejected; past that point the failure is logged and the settled session is left alone.

That leaves the three states distinct: the user declined (`USER_REJECTED`), approval happened but settlement failed (`SESSION_SETTLEMENT_FAILED`), and settlement succeeded but local persistence did not (logged, no rejection).

Worth noting the missing local connection record itself is still a separate gap — the session survives without one. I left that out of this PR to keep it focused on the `events` fix, but happy to address it here if you would prefer.

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
```diff
@@ -337,7 +337,7 @@ class ServiceWalletConnect extends ServiceBase {
         supportedNamespaces[namespace] = {
           chains: mergedChains,
           methods: supportMethodsMap[namespace] ?? [],
-          events: supportEventsMap[namespace],
+          events: supportEventsMap[namespace] ?? [],
```
sidmorizon, line 430: # 🟠 P1 不支持的 namespace 被当成空壳写进 approve 载荷

## 问题

`processNamespaces` 里的 `impl = namespaceToImplsMap[namespace]` 只用来找账户，从来没有用于过滤。对 `solana` / `bip122` 这类钱包不支持的 namespace，`filteredChains` 必定为空（它们的 chain 全部落在 `notSupportedChains` 里），所以把 `events` 从 `undefined` 兜底成 `[]` 之后，approve 载荷里不是「剔除该 namespace」，而是多出一个 `{ chains: [], methods: [], events: [], accounts: [] }` 的空壳。

## 影响

一是这个空壳会原样进入 `wc_sessionSettle`，dApp 侧读到的是「钱包支持 solana，但 0 个账户」，而不是「不支持」。二是当 dApp 用 CAIP-2 形式的 key 声明 namespace 时（`optionalNamespaces: { 'solana:5eykt…': {...} }`，`getNotSupportedChains` 里的 `key.includes(':')` 分支说明这种形态本就在预期内），`@walletconnect/utils@2.21.9` 的 `isConformingNamespaces` 会对每个含 `:` 的 approved key 校验 `getAccountsChains(accounts).includes(key)`，空 `accounts` 直接判 `NON_CONFORMING_NAMESPACES`，`approveSession()` 依然抛错 —— 用户点了 Approve 还是连不上，只是报错从 `UNSUPPORTED_EVENTS` 换成了 `NON_CONFORMING_NAMESPACES`。

## 建议

与其补空数组，不如在 `processNamespaces` 里直接跳过钱包无法服务的 namespace：

```ts
const impl = namespaceToImplsMap[namespace];
if (!impl) {
  continue; // 不声明无法服务的 namespace，而不是声明一个空壳
}
```

需要区分 required 与 optional：optional 里的不支持项直接不写入即可；required 里若出现不支持的 namespace（例如只给了 `methods`/`events` 而没有 `chains`，绕过了 `onSessionProposal` 开头的 `UNSUPPORTED_CHAINS` 早退），漏掉 key 会命中 `isConformingNamespaces` 的 "keys don't satisfy requiredNamespaces" 分支，应该在早退处一并拒绝。

[Claude session ↗](https://claude.ai/code/session_01VnfhdQqpFANS8CnriqXkkd)

---

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -235,10 +246,14 @@ class ProviderApiWalletConnect {
       });
     } catch (e) {
       console.error('onSessionProposal error: ', e);
-      await this.web3Wallet?.rejectSession({
-        id: proposal.id,
-        reason: getSdkError('USER_REJECTED'),
-      });
+      if (!sessionSettled) {
+        await this.web3Wallet?.rejectSession({
```
sidmorizon: # 🟡 P2 catch 里的 `rejectSession` 自身会抛，埋点拿不到

## 问题

`rejectSession` 仍是裸 `await`。`@walletconnect/sign-client@2.21.9` 的 `reject()` 先 `confirmOnlineStateOrThrow()`，再 `isValidReject()` → `checkRecentlyDeleted(id)` + `isValidProposalId(id)`，proposal 不存在或已过期都会直接抛。而 `wc_sessionPropose` 的 ttl 是 `FIVE_MINUTES`，到期后 expirer 会执行 `deleteProposal(id, true)` 并发 `proposal_expire`，proposal 记录就没了。

## 影响

审批弹窗停留超过 `5 分钟` 后用户再点 Approve（`approveSession` 先抛 `No matching key. proposal id doesn't exist`），或断网时点 Reject（`No internet connection`），进入 catch 后 `rejectSession` 会二次抛出。异常直接冲出 catch，紧随其后的 `defaultLogger.discovery.dapp.dappUse({ failReason })` 不会执行；在 iOS/Android/扩展上 `onSessionProposal` 只跑在 `bg` 运行时，最终表现为一条无人接管的 promise rejection。上面注释写的 “failures past this point are only logged” 在这些路径上并不成立。

## 建议

把 reject 单独包一层，保证埋点一定跑到：

```ts
if (!sessionSettled) {
  try {
    await this.web3Wallet?.rejectSession({ ... });
  } catch (rejectError) {
    console.error('onSessionProposal reject failed: ', rejectError);
  }
}
```

也可以先用 `getPendingSessionProposals()` 确认 proposal 仍在，再决定要不要发 reject。

[Claude session ↗](https://claude.ai/code/session_01VnfhdQqpFANS8CnriqXkkd)

---

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
kjljxybr: Fixed in cdfac42.

You are right that `?? []` only changed which error the SDK throws. `processNamespaces` now skips any namespace with no impl behind it instead of writing a shell, so `solana` / `bip122` / `tron` are simply absent from the approve payload rather than present with zero accounts.

Two things came out of following the CAIP-2 note:

- A namespace key can itself be a chain, and the SDK treats it that way — `getChainsFromNamespace` is `key.includes(':') ? [key] : value.chains || []`. The impl lookup, the requested-chain list and the approved key now follow the same rule through a small `parseNamespaceKey` helper shared by `processNamespaces`, `getSessionApprovalAccountInfo` and `getAvailableNetworkIdsForNamespace`. `eip155:1` as a key therefore resolves to the EVM impl and contributes `eip155:1` as a chain, instead of resolving to `undefined` and producing exactly the shell you describe.
- Entries whose chains were all filtered into `notSupportedChains`, and entries with no account behind them, are skipped for the same reason — both ended as the same empty shell.

On required vs optional: a required namespace the wallet cannot serve is now rejected in `onSessionProposal` before the modal opens, with `UNSUPPORTED_NAMESPACE_KEY`. `getNotSupportedChains` only inspects the chains an entry declares, so an entry naming an unsupported namespace without declaring any slipped past the early return and would now hit the "keys don't satisfy requiredNamespaces" branch after the user already tapped Approve.

`events: supportEventsMap[namespace] ?? []` stays, but only as a guard for a namespace present in `namespaceToImplsMap` and missing from `supportEventsMap` — matching the `methods` line above it. It is no longer what makes the unsupported-namespace case work.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
kjljxybr: Fixed in f8f5016.

`rejectSession` now runs in its own try/catch, so the `dappUse({ failReason })` call below it is always reached and the handler cannot end as an unhandled rejection in the `bg` runtime.

I kept the reject unconditional rather than gating it on `getPendingSessionProposals()`: that check still races the expirer between the read and the `reject()`, so the try/catch has to be there regardless, and adding the lookup would only narrow the window.

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
```diff
@@ -334,12 +375,18 @@ class ServiceWalletConnect extends ServiceBase {
...
 
-        supportedNamespaces[namespace] = {
+        // Same reasoning: an entry whose chains were all filtered out, or that
+        // has no account behind it, would be an empty namespace on the dApp
+        // side rather than an absent one.
+        if (mergedChains.length === 0 || !address) {
+          continue;
+        }
+
+        supportedNamespaces[key] = {
```
sidmorizon, line 427: # 🟠 P1 CAIP-2 key 写入 session 后，切换账户不再同步给 dApp

## 问题

这一行把写入键从 `namespace`（`eip155`）改成了 `key`，于是当 dApp 用 inline chain 形式声明命名空间时（`requiredNamespaces: { 'eip155:1': {...} }`），approve 后 settle 下来的 `session.namespaces` 键就是 `eip155:1`。但 `updateNamespaceAndSession`（本文件 452-457 行）仍然按裸命名空间反查 impl：

```ts
for (const [namespace, value] of Object.entries(session.namespaces)) {
  const matchAccount = accountsInfo.find(
    (account) =>
      account.networkImpl === namespaceToImplsMap[namespace as INamespaceUnion],
  );
```

`namespaceToImplsMap['eip155:1']` 是 `undefined`，而 `IConnectionAccountInfo.networkImpl` 永远是非空字符串，所以 `matchAccount` 恒为 `undefined`，该条目被整个跳过。这里没有复用新加的 `parseNamespaceKey`。

## 影响

用户在 OneKey 里对一个已连接的 WalletConnect 会话切换账户时，`ServiceDApp.updateWalletConnectSession` 会走到这里：`updatedNamespaces` 保持旧的 `accounts`，`updateSession` 变成空操作，紧接着的推送循环又用 `value.accounts?.[0]` 把**旧地址**当作 `accountsChanged` 发给 dApp。结果是钱包 UI 显示账户 B、dApp 仍然按账户 A 请求签名，且没有任何报错。

这条路径正是本 PR 新打开的：改动前 `getSessionApprovalAccountInfo` 遇到 `eip155:1` 会直接 `throw 'Namespace not supported'`，会话根本建立不起来，所以问题只在新支持的形态上暴露。

## 建议

在 `updateNamespaceAndSession` 里改用 `parseNamespaceKey(namespace).impl` 做 impl 反查，让它和 `processNamespaces` 对键的解析保持一致；顺带确认 `value.chains` 为空时（inline key 场景 chains 由 `chainsFromKey` 提供）重建 `accounts` 的逻辑仍然正确。

[Claude session ↗](https://claude.ai/code/session_01VnfhdQqpFANS8CnriqXkkd)

---

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
```diff
@@ -255,7 +287,7 @@ class ServiceWalletConnect extends ServiceBase {
     let index = 0;
     // Filter supported chains from requiredNamespace
     for (const namespace of Object.keys(requiredNamespaces)) {
-      const impl = namespaceToImplsMap[namespace as INamespaceUnion];
+      const { impl } = parseNamespaceKey(namespace);
```
sidmorizon, line 321: # 🟡 P2 required 循环缺少 impl 去重，多余的账户选择器会被静默丢弃

## 问题

`parseNamespaceKey` 让 `eip155:1` / `eip155:137` 这类 inline key 也能解析出 impl，但这个 required 循环是无条件 `supported.push(...)` 的，没有像下面 optional 循环那样用 `existImpl` 去重。于是 `requiredNamespaces: { 'eip155:1': {...}, 'eip155:137': {...} }` 会产出两条 `impl === IMPL_EVM` 的记录，`accountSelectorNum` 分别是 `1000` 和 `1001`。

`WCSessionProposalModal` 按 `sessionAccountsInfo` 的条数渲染选择器，所以用户要选两次；但 `buildWalletConnectNamespace` 只有 `accountsInfo.find((a) => a.networkImpl === impl)`，两个 key 都拿到数组里**第一个** EVM 账户。

## 影响

用户为 `eip155:137` 选的账户 B 不会进入 approve payload——两个 namespace 的 `accounts` 都是账户 A 的地址，而账户 B 又被 `saveConnectionSession` 持久化进 `connectionMap[1001]`，造成本地记录与 dApp 实际拿到的地址长期不一致。后续 `updateWalletConnectSession` 同样按 impl `find`，用户始终改不动第二个选择器。

同一份代码里 optional 循环去重、required 循环不去重，行为也不自洽。

## 建议

要么在 required 循环里同样按 impl 去重（同一 impl 只保留一个选择器，`networkIds` 合并），要么把 `buildWalletConnectNamespace` 的账户查找从 impl 改成按 `accountSelectorNum` / networkId 精确匹配，让每个 namespace key 对应到用户真正为它选的账户。二选一即可，但不能停在现在这种「多选一次、只用第一个」的中间状态。

[Claude session ↗](https://claude.ai/code/session_01VnfhdQqpFANS8CnriqXkkd)

---

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
kjljxybr: Fixed in 4525841.

Confirmed — `updateNamespaceAndSession` was the one place left resolving the impl straight off the key, and `IConnectionAccountInfo.networkImpl` is never empty, so `matchAccount` could only ever be `undefined` for an inline key. `updateSession` then went out unchanged and the push loop below it re-sent the old address as `accountsChanged`, which is worse than a visible failure.

It goes through `parseNamespaceKey` now, like `processNamespaces` does.

On the `value.chains` note: for a session settled by this build the field is populated — `processNamespaces` writes `chains: mergedChains`, and for an inline key `chainsFromKey` puts the key itself in there — but the rebuild now falls back to `chainsFromKey` when `chains` is empty, so a session settled by an older build cannot end up with an empty `accounts` array either.

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
kjljxybr: Fixed in edd8b433.

Took the first of the two options: the required loop now merges `networkIds` into the existing entry for that impl instead of pushing a second selector, so it matches the optional loop right below it.

Per-key account matching would be the bigger change and it would not stop at this function — `buildWalletConnectNamespace`, `updateNamespaceAndSession` and the `connectionMap` written by `saveConnectionSession` are all keyed by impl, so the model would have to change in all four places. One selector per impl is what the rest of the code already assumes, and the required loop was the only place breaking that assumption.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -146,33 +146,55 @@ class ProviderApiWalletConnect {
       // proposal,
       proposal?.params?.requiredNamespaces,
     );
+    // A required namespace the wallet has no impl for is left out of the
+    // approve payload, which then fails the SDK's requiredNamespaces
+    // conformance check after the user already tapped Approve. Catch it here:
+    // getNotSupportedChains only inspects the chains an entry declares.
+    const notSupportedNamespaces =
+      await serviceWalletConnect.getNotSupportedNamespaces(
```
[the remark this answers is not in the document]
kjljxybr: Fixed in cff985c8.

Confirmed against the SDK: `isValidNamespaces` in `@walletconnect/utils@2.21.9` requires `Object.keys(namespaces).length > 0` via `isValidObject`, so `approveSession({})` does throw `namespaces should be an object with data` — a required-empty, optional-only-unsupported proposal reproduces exactly as described, tapping Approve on a zero-account modal that then fails as `SESSION_SETTLEMENT_FAILED`.

`onSessionProposal` now runs `getNotSupportedNamespaces` against `optionalNamespaces` too, but only when `requiredNamespaces` is empty, and only rejects up front when *every* optional namespace comes back unsupported — a mixed proposal (optional `eip155` alongside `solana`) still opens the modal, matching your note that it isn't affected. `getNotSupportedNamespaces` also now guards its input with `?? {}` so an `undefined` `requiredNamespaces` can't throw on `Object.keys`.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -139,40 +139,87 @@ class ProviderApiWalletConnect {
...
+    // did nothing. Only reject on this path once every optional namespace is
+    // unsupported: a mixed proposal (e.g. optional eip155 + solana) still has
+    // something to approve.
+    const optionalNamespaceKeys = Object.keys(optionalNamespaces);
+    const notSupportedOptionalNamespaces =
+      Object.keys(requiredNamespaces).length === 0 &&
+      optionalNamespaceKeys.length > 0
+        ? await serviceWalletConnect.getNotSupportedNamespaces(
+            optionalNamespaces,
+          )
```
sidmorizon, line 170: **Unsupported optional chains reach approval**

Severity: `non-severe`

Optional-only proposals with a recognized namespace but no supported chains pass `allOptionalNamespacesUnsupported`. Approval then offers no selectable account and cannot complete.

kjljxybr: Fixed in 144d05c9.

Confirmed — the previous fix only checked namespace-family support (`impl`), which is exactly what let a recognized-but-chain-unsupported entry like `eip155:999999` through: `notSupportedOptionalNamespaces` didn't flag it, `allOptionalNamespacesUnsupported` stayed false, and it reached the same dead-end modal `buildWalletConnectNamespace` would later empty out.

`getNotSupportedNamespaces` now checks chain-level support too — a key is "not supported" if the wallet has no impl for its family, if it declares no chains at all (via the key or `chains`), or if none of its declared chains resolve through `getWcChainInfo`. That's the same criterion `buildWalletConnectNamespace` already uses to decide whether an entry survives, so a namespace flagged here is guaranteed to contribute nothing to the approve payload regardless of which account the user would have picked. No other call site changed; the required-namespaces check upstream just got more precise for free.

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
```diff
@@ -214,6 +230,42 @@ class ServiceWalletConnect extends ServiceBase {
...
+        continue;
+      }
+      const entryChains = [...chainsFromKey, ...(value.chains ?? [])];
+      if (entryChains.length === 0) {
+        notSupported.push(key);
+        continue;
+      }
+      const chainInfos = await Promise.all(
+        entryChains.map((chain) => this.getWcChainInfo(chain)),
+      );
```
sidmorizon, line 271: **Malformed chain proposals receive no rejection**

Severity: `non-severe`

When a chain lacks a colon, `getNotSupportedNamespaces` throws before the rejection handler starts. The dApp receives neither a session nor an error.

Suggested fix:

```ts
      const chainInfos = await Promise.all(
        entryChains.map((chain) =>
          chain.includes(':') ? this.getWcChainInfo(chain) : undefined,
        ),
      );
```

kjljxybr: Fixed in 3b1f842c.

Confirmed — `getWcChainInfo` throws `OneKeyLocalError` when a chain string has no CAIP-2 colon, and that call sat outside `onSessionProposal`'s `try/catch` (the not-supported checks all run before it), so a malformed `chains` entry in the proposal took the handler down with no `rejectSession` ever called and no error surfaced to the dApp.

`getNotSupportedNamespaces` now checks `chain.includes(':')` before calling `getWcChainInfo`, resolving a colonless entry straight to "unsupported" instead of throwing — same guard `getNotSupportedChains` already had for exactly this reason, which the new chain-level check in this method missed when it was added.

packages/kit-bg/src/services/ServiceWalletConnect/ServiceWalletConnect.ts
```diff
@@ -214,6 +230,49 @@ class ServiceWalletConnect extends ServiceBase {
...
+      }
+      // getWcChainInfo throws on a chain string without a CAIP-2 colon
+      // instead of returning undefined. A malformed `chains` entry -- the
+      // dApp's proposal payload, not validated before this point -- must
+      // still resolve to "not supported" rather than take down the handler
+      // this runs in before it reaches its own try/catch.
+      const chainInfos = await Promise.all(
+        entryChains.map((chain) =>
+          chain.includes(':') ? this.getWcChainInfo(chain) : undefined,
+        ),
```
sidmorizon, line 270: **Non-string chains abort proposals**

Severity: `non-severe`

A non-string chain makes `chain.includes` throw before rejection. The dApp receives no unsupported-namespace response.

Suggested fix:

```ts
        entryChains.map((chain) =>
          typeof chain === 'string' && chain.includes(':')
            ? this.getWcChainInfo(chain)
            : undefined,
        ),
```

kjljxybr: Fixed in `8aa2c7745c`.

Confirmed: `chains` is only TS-typed as `string[]` (via `ProposalTypes.RequiredNamespaces`), but the actual value comes straight from the dApp's wire payload and isn't runtime-validated before this point. A non-string entry would make `chain.includes(':')` throw before the guard added last round could help, and this call sits before `onSessionProposal`'s own try/catch, so it would crash the handler with no response to the dApp at all. Added `typeof chain === 'string' &&` ahead of the colon check.

sidmorizon:
Claude review session: https://claude.ai/code/session_01VnfhdQqpFANS8CnriqXkkd

kjljxybr:
Merged `x` into the branch (`595e6e0f3`) — it was only behind, no conflicts.

`x` has since renamed the WalletConnect listeners and put them behind the travel mode ingress wrapper (#13185), so this PR's changes now sit in `handleSessionProposal` instead of `onSessionProposal`. The logic itself is untouched: the net diff against `x` is still only `ProviderApiWalletConnect.ts` + `ServiceWalletConnect.ts`, and `ServiceWalletConnect.ts` is byte-identical to what you last reviewed.

All review threads on this PR are answered. Ready for another look.

PR #13075, opened 2026-08-30
feat: implement WalletConnect session_authenticate (One-Click Auth)

kjljxybr:
> **Draft.** This has not been built or run — I do not have the toolchain set up for this repo. The flow is written against the existing patterns in this file, but it needs a real device pass before it can be trusted. Opening it as a draft so the approach can be reviewed first.

## Summary

`session_authenticate` was registered as an event listener but the handler was empty:

```ts
onAuthRequest = (args: WalletKitTypes.SessionAuthenticate) => {
  console.log('onAuthRequest: ', args);
};
```

A dApp using One-Click Auth — a pairing URI carrying `methods=wc_sessionAuthenticate` — therefore got **no approval screen and no response at all**. From the user's side the QR scanner closes and the wallet returns to its home screen; from the dApp's side the request simply never gets answered.

This implements the flow for EVM chains:

- keep only the requested chains the wallet actually supports, rejecting with `UNSUPPORTED_CHAINS` when none remain
- run the payload through `populateAuthPayload` with the wallet's supported methods
- reuse the existing session proposal modal for account selection, by shaping the authenticate payload like a `SessionProposal` — no new screen, and `getSessionApprovalAccountInfo` keeps working unchanged
- build the CAIP-122 issuer (`did:pkh:<chain>:<address>`), format the message with `formatAuthMessage`, and sign it through the regular sign-message modal
- answer with `approveSessionAuthenticate` and persist the connection like `onSessionProposal` does
- reject explicitly on any failure instead of leaving the dApp waiting

Refs #13073

## Runtime impact

- Background runtime only (`kit-bg`); no new screens, routes or translation strings.
- Only affects `session_authenticate`, which previously did nothing at all — no existing path changes behaviour.
- Non-EVM authenticate requests are now rejected with a clear SDK error rather than being ignored.

## Validation

Reproduced the current behaviour on a physical device, before the change:

- Samsung SM-S938B, Android 16, OneKey 6.5.3 (versionCode 2026082287) from Google Play
- a dApp whose pairing URI contains `methods=wc_sessionAuthenticate`: the scanner closes, no approval screen appears, and the dApp never receives a response
- the same wallet connects normally when the dApp uses a plain `wc_sessionPropose`, which is what isolated this to the authenticate path

**Not validated:** I have not built the app or run it against this branch, so the implementation itself is unverified. Points I would specifically want checked on a device:

- whether reusing the session proposal modal reads acceptably for an authentication request, or whether this warrants its own screen
- the CACAO round-trip — that the signature produced through the sign-message modal verifies against the payload the dApp sent
- multi-chain authenticate requests: this signs a single CACAO for the first supported chain, which covers the common case but may not be what you want in general


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +351,170 @@ class ProviderApiWalletConnect {
...
+      });
+      return;
+    }
+
+    const methods = supportMethodsMap[evmNamespace] ?? [];
+    const authPayload = populateAuthPayload({
+      authPayload: params.authPayload,
+      chains: supportedChains,
+      methods,
+    });
```
sidmorizon: **Unsupported methods leave requests unanswered**

Severity: `non-severe`

When ReCap requests unsupported methods, `populateAuthPayload` throws outside the error handler. The dApp receives no rejection and waits until expiry.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +351,170 @@ class ProviderApiWalletConnect {
...
+        id,
+        auths: [auth],
+      });
+
+      await serviceDApp.saveConnectionSession({
+        origin,
+        accountsInfo: result.accountsInfo,
+        storageType: 'walletConnect',
+        walletConnectTopic: approved?.session?.topic,
+      });
```
sidmorizon: **Authentication-only approvals end in errors**

Severity: `non-severe`

Without ReCap methods, `approveSessionAuthenticate` returns no session. Persistence then throws after approval, and the error handler rejects an already-consumed request.

Suggested fix:

```ts
      if (approved?.session?.topic) {
        await serviceDApp.saveConnectionSession({
          origin,
          accountsInfo: result.accountsInfo,
          storageType: 'walletConnect',
          walletConnectTopic: approved.session.topic,
        });
      }
```

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
kjljxybr: Fixed in 6e24292.

Chain filtering, `populateAuthPayload` and the pseudo-proposal construction now all run inside the `try`, so a ReCap the wallet cannot satisfy — or a failing `getWcChainInfo` — is turned into an explicit `rejectSessionAuthenticate` instead of leaving the dApp waiting until expiry.

`rejectAuth` is also idempotent now and swallows its own errors, so the handler cannot end in an unhandled rejection when the request has already been consumed.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
kjljxybr: Fixed in 6e24292.

`saveConnectionSession` is now guarded on `approved?.session?.topic`, so an authentication-only approval no longer hits the `walletConnectTopic is required` throw in `ServiceDApp.saveConnectionSession`.

The second half of the problem is handled too: an `answered` flag is set as soon as `approveSessionAuthenticate` resolves, and `rejectAuth` returns early when it is set — so the catch can no longer reject a request that was already approved.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +351,170 @@ class ProviderApiWalletConnect {
...
+      const approved = await this.web3Wallet?.approveSessionAuthenticate({
+        id,
+        auths: [auth],
+      });
+
+      await serviceDApp.saveConnectionSession({
+        origin,
+        accountsInfo: result.accountsInfo,
+        storageType: 'walletConnect',
+        walletConnectTopic: approved?.session?.topic,
```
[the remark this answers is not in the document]
kjljxybr: Fixed in 6e24292 — the persistence call is now guarded on `approved?.session?.topic`, so an authentication-only request that establishes no session skips it instead of throwing `walletConnectTopic is required`.

The follow-on issue you point out is covered as well: an `answered` flag is set once `approveSessionAuthenticate` resolves, and `rejectAuth` returns early when it is set, so the catch no longer rejects an already-approved request.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +351,170 @@ class ProviderApiWalletConnect {
...
+            chains: supportedChains,
+            methods,
+            events: supportEventsMap[evmNamespace] ?? [],
+          },
+        },
+      },
+      verifyContext: args.verifyContext,
+    } as unknown as WalletKitTypes.SessionProposal;
+
+    try {
```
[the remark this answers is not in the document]
kjljxybr: Fixed in 6e24292.

The `try` now opens before the origin check and covers chain filtering, `populateAuthPayload` and the pseudo-proposal construction, so a failure in any of those steps replies with `rejectSessionAuthenticate` rather than leaving the dApp waiting for a response that never comes.

`rejectAuth` additionally guards against double-answering and swallows its own errors, so the async event handler cannot end in an unhandled rejection.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +351,191 @@ class ProviderApiWalletConnect {
...
+      // A request that only asks for authentication (no ReCap methods) is
+      // approved without a session being established, so there is no topic to
+      // persist and nothing to show in the connected-sites list.
+      if (approved?.session?.topic) {
+        await serviceDApp.saveConnectionSession({
+          origin,
+          accountsInfo: result.accountsInfo,
+          storageType: 'walletConnect',
+          walletConnectTopic: approved.session.topic,
+        });
```
sidmorizon: 🟠 P1: Approved sessions bypass wallet controls

Severity: `severe`

If `saveConnectionSession` fails after approval, the active session never appears in connected sites. Users cannot see or disconnect it through the wallet.

kjljxybr: Fixed in d53e3e2.

`saveConnectionSession` now runs in its own `try`. If it fails after the session was established, the session is torn down with `serviceWalletConnect.walletConnectDisconnect(topic)` — which emits `session_delete` to the dApp — and the original error is rethrown into the outer handler for logging. So the wallet no longer ends up with a live relay session it has no record of and the user cannot reach.

Two notes on the shape of the fix:

- `answered` is already `true` at that point, so the outer catch logs the failure without attempting to reject a request that was approved.
- The disconnect is itself wrapped, so a failing teardown surfaces the original save error rather than masking it.

For what it is worth, `onSessionProposal` has the same `approveSession` → `saveConnectionSession` ordering and the same exposure. I have left it alone to keep this PR scoped, but happy to fix it here or in a follow-up if you would prefer.

kjljxybr: Followed up on my own note above: `onSessionProposal` now gets the same treatment in 9daaab2.

- `saveConnectionSession` runs in its own `try`; if it fails after `approveSession` succeeded, the session is torn down with `walletConnectDisconnect(topic)` and the original error is rethrown for logging, so no live relay session is left without a stored connection.
- The outer catch only calls `rejectSession` when nothing was approved (`approvedTopic` is unset). Previously a post-approval failure tried to reject a proposal the SDK had already consumed, which both failed and masked the real error.
- `batchEmitNetworkChangedEvent` is skipped on that path, since the session no longer exists.

This does touch pre-existing code outside the original scope of the PR. Say the word if you would rather have it as a separate PR against `x` and I will pull it out of this branch.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +376,207 @@ class ProviderApiWalletConnect {
...
+        fullScreen: true,
+      })) as IWalletConnectSessionProposalResult;
+
+      const accountInfo = result.accountsInfo?.[0];
+      if (!accountInfo?.address || !accountInfo?.networkId) {
+        throw new OneKeyLocalError('No account selected for authentication');
+      }
+
+      // CAIP-122 issuer: did:pkh:<namespace>:<reference>:<address>
+      const iss = `did:pkh:${supportedChains[0]}:${accountInfo.address}`;
```
sidmorizon: # 🟠 P1 iss 固定用 `supportedChains[0]`，与用户在弹窗里实际选中的网络不一致

## 问题

`iss` 用的是 `supportedChains[0]`，但地址/网络来自 `result.accountsInfo[0]`，这两者不是同一个来源。传给弹窗的 `optionalNamespaces.eip155.chains = supportedChains`，`getSessionApprovalAccountInfo` 会把它展开成 `networkIds` 交给 `WalletConnectAccountTriggerList` 的网络选择器，用户可以在里面切换网络；而且 `DAppAccountListItem` 会优先复用 `discover` 场景下持久化的 `selectedAccount.networkId`（只要它在 `networkIds` 里），所以即使用户什么都不点，默认选中的网络也可能不是 `supportedChains[0]`。

结果就是：`accountInfo.networkId` 可能是 `evm--8453`，而 `iss` 仍然被写成 `did:pkh:eip155:1:0x…`。

## 影响

以 dApp 请求 `chains: ['eip155:1', 'eip155:8453']`、用户选中 Base 为例：

- `formatAuthMessage` 生成的 SIWE 正文里是 `Chain ID: 1`，但 `openSignMessageModal` 是用 `networkId: accountInfo.networkId`（Base）打开的，签名确认页展示的网络与正文声明的链互相矛盾。
- `buildAuthObject` 把同一个 `iss` 写进 CACAO，sign-client 的 `approveSessionAuthenticate` 用 `getNamespacedDidChainId(p.iss)` 建会话账户，按 `iss` 把登录绑定到 `eip155:1`；按认证链做作用域的 dApp 会拿到错误的链。
- 随后 `saveConnectionSession` 存的是 `evm--8453`，而这里又没有像 `onSessionProposal` 那样调用 `batchEmitNetworkChangedEvent`，dApp 侧的 active chain 停在 `eip155:1`，与「已连接站点」里显示的网络长期不一致。

## 建议

用用户实际选中的网络反推 wc chain，而不是取第一个受支持的链；并补上和 `onSessionProposal` 一致的 `chainChanged` 通知。

```ts
const wcChain = await serviceWalletConnect.getWcChainByNetworkId({
  networkId: accountInfo.networkId,
});
const iss = `did:pkh:${wcChain}:${accountInfo.address}`;
// ...approve 并保存连接之后
void serviceWalletConnect.batchEmitNetworkChangedEvent({
  topic,
  accountsInfo: result.accountsInfo,
});
```

`getWcChainByNetworkId` 返回的正是 `eip155:137` 这种格式，可直接拼进 `did:pkh:`。

[Claude session ↗](https://claude.ai/code/session_01HLXExJunJokiU1zzB9B5uU)

---

sidmorizon (commented):
补充一条关于 CACAO issuer 链选择的意见。

---

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
kjljxybr: Fixed in 1ed390d.

`iss` now comes from the account the user actually selected — `getWcChainByNetworkId({ networkId: accountInfo.networkId })` returns the `eip155:<reference>` form directly, so the SIWE `Chain ID` line, the network the sign-message screen opens on, the chain `approveSessionAuthenticate` scopes the session to, and the stored connection all agree.

`batchEmitNetworkChangedEvent` is sent after `saveConnectionSession` succeeds, the same as `onSessionProposal` does. It sits inside the `if (topic)` branch, since an authentication-only approval establishes no session and has nothing to emit on. The success-path `dappUse` logs the chain that was authenticated instead of the whole supported list; the catch still logs the list, as nothing was settled on there.

The multi-chain caveat from the description still stands: this signs a single CACAO, now for the selected chain rather than for the first requested one.

packages/kit-bg/src/providers/ProviderApiWalletConnect/ProviderApiWalletConnect.ts
```diff
@@ -339,8 +376,220 @@ class ProviderApiWalletConnect {
...
+      if (!authMessage) {
+        throw new OneKeyLocalError(
+          'Failed to build the authentication message',
+        );
+      }
+
+      const message = `0x${Buffer.from(authMessage, 'utf8').toString('hex')}`;
+      const signature = (await serviceDApp.openSignMessageModal({
+        request: {
+          scope: '$walletConnect',
```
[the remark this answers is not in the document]
kjljxybr: Fixed in a374bae1.

`openSignMessageModal`'s `request` now carries `data: { walletConnectVerifyContext: args.verifyContext }`, following the same convention `WalletConnectRequestProxy.request` already uses for `session_request`. `ServiceDApp.openModal` hoists that back onto `$sourceInfo`, and `MessageConfirm` already reads `sourceInfo?.walletConnectVerifyContext` into `useRiskDetection` — that wiring existed, it just wasn't fed for this flow. The SIWE confirmation screen now gets the same Verify-derived severity override as the session-proposal screen instead of falling back to backend-only reputation.

sidmorizon:
Claude review session: https://claude.ai/code/session_01HLXExJunJokiU1zzB9B5uU

kjljxybr:
Merged `x` into the branch (`c51aee8d7`). It was conflicting; one conflict, resolved as follows.

`x` renamed `onAuthRequest` to a private `handleAuthRequest` and now registers every WalletConnect listener through a `gatedHandle*` wrapper produced by `gateSessionEvent` (#13185, travel mode). This PR replaces the same handler's body, so the conflict was purely the signature line. I kept the upstream signature and this branch's `session_authenticate` implementation as its body — `session_authenticate` therefore stays registered as `gatedHandleAuthRequest`, gated like the other four handlers. Nothing else in the PR changed: the diff between the pre-merge head and the merge result is exactly the #13185 hunks.

One interaction worth your call, not changed here: `gateSessionEvent` passes `onBlocked: async () => {}`, so while travel mode is active a `session_authenticate` request is dropped with no response at all. That is deliberate for suppressed sessions ("must not start another outbound response"), but it is the one path where this PR's "always answer the request" rule does not hold. Leaving upstream's policy in place; say the word if you would rather it rejected with an explicit reason instead.

All review threads on this PR are answered. Ready for another look.

PR #13079, opened 2026-08-30
fix: sync Android navigation bar with app theme

huhuanming:
## Summary

- upgrade every integrated @onekeyfe app-module package from 3.0.94 to 3.0.95
- update iOS pod resolution for the new native module versions
- synchronize the Android system Navigation Bar background and icon appearance whenever the app theme changes
- keep the behavior Android-only; gesture navigation is handled by the native API without adding a visible three-button protection view

## Related

- OneKeyHQ/app-modules#85
- app-modules 3.0.95 release run: https://github.com/OneKeyHQ/app-modules/actions/runs/33327489194
- packages are published under the next dist-tag

## Verification

- yarn install
- yarn app:ios:pod-install
- yarn agent:check --profile pr
- Android Gradle :app:compileProdDebugSources
- full yarn app:android:legacy build, install, and launch on Android 15 / API 35
- three-button mode: light theme reported LIGHT_NAVIGATION_BARS and rendered a light navigation background with dark buttons; switching the system/app theme to dark cleared the light-navigation appearance flag


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/shared/src/modules3rdParty/rootview-background/index.native.tsx
```diff
@@ -14,6 +14,16 @@ export const updateRootViewBackgroundColor: IUpdateRootViewBackgroundColor = (
...
   ReactNativeDeviceUtils.changeBackgroundColor(r, g, b, Math.round(a * 255));
 
+  if (platformEnv.isNativeAndroid) {
+    ReactNativeDeviceUtils.setNavigationBarAppearance(
+      r,
+      g,
+      b,
+      Math.round(a * 255),
+      themeVariant === 'light',
+    );
```
sidmorizon, line 24: 🟠 P1: OTA bundle crashes older Android shells

Severity: `severe`

An OTA bundle running against native modules 3.0.94 calls missing `setNavigationBarAppearance` during startup. Android users can lose access before installing the new shell.

packages/shared/src/modules3rdParty/rootview-background/index.native.tsx
```diff
@@ -14,6 +14,16 @@ export const updateRootViewBackgroundColor: IUpdateRootViewBackgroundColor = (
   const { r, g, b, a } = parsedColor.toRgb();
   ReactNativeDeviceUtils.changeBackgroundColor(r, g, b, Math.round(a * 255));
 
+  if (platformEnv.isNativeAndroid) {
+    ReactNativeDeviceUtils.setNavigationBarAppearance(
```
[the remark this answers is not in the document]
huhuanming: Thanks. This rollout is native+JS coordinated: this bundle ships only with the 3.0.95 native shell and is not delivered as a JS-only OTA to 3.0.94 or earlier shells. Older-shell compatibility is explicitly out of scope for this release, so we are intentionally not adding a capability guard.

packages/shared/src/modules3rdParty/rootview-background/index.native.tsx
huhuanming: Thanks. This rollout is native+JS coordinated: this bundle ships only with the 3.0.95 native shell and is not delivered as a JS-only OTA to 3.0.94 or earlier shells. Older-shell compatibility is explicitly out of scope for this release, so we are intentionally not adding a capability guard.

huhuanming:
Regarding the double-writer note in the Cursor review: this shared responsibility is intentional in the current implementation. expo-navigation-bar continues to cover navigation-button appearance for the Activity and Expo extra windows, while setNavigationBarAppearance owns the current Activity background, Android 15 three-button protection view, contrast behavior, and resume reapplication. Both paths derive the same icon appearance: light theme uses dark icons, and dark theme uses light icons. Removing the Expo call would drop its extra-window coverage. Android 15 light/dark three-button behavior and three-button/gesture transitions were verified, so no change is needed for this PR.

originalix (approved):

PR #13080, opened 2026-08-31
fix: disable iOS tab content crossfade OK-61355

huhuanming:
[OK-61355](https://onekeyhq.atlassian.net/browse/OK-61355)

----------

[OK-61355](https://onekeyhq.atlassian.net/browse/OK-61355)

## Summary

- Keep the user-tap path on UIKit's immediate native tab selection.
- When `disablePageAnimations` is enabled, replace only the tab content transition with a zero-duration native animator.
- Preserve the system tab bar/Liquid Glass selection animation, haptics, reselection, native `preventsDefault`, and the existing programmatic-selection behavior.

## Root cause

On a tab tap, `UITabBarControllerDelegate.shouldSelect` returns synchronously, so UIKit selects the destination controller and starts its default content transition before the later JS navigation-state update. The existing `disablePageAnimations` handling only covered programmatic `selectedIndex` updates; by the time JS committed `selectedPage`, the selected index already matched and that path returned early.

The pages are not unloaded during warm switches: React's visited-scene set only grows, and the native container caches view controllers by tab key. Runtime traces reused the same view-controller identities with a stable four-controller cache. A native-only, single-variable A/B changed the visible result from a multi-frame crossfade to one frame boundary without changing JS bundles or page lifecycle code, isolating the delay and ghosting to UIKit's default content transition.

JS tracing also showed that `navigation.emit` and dispatch return take about 0.05 ms each; native click to JS event was about 14 ms median, React layout commit followed about 3.7 ms later, and the selected-page prop returned to native about 49 ms median. Moving the router to another runtime would not remove this scheduling/commit path.

Runtime scope: iOS main UI runtime. The `UITabBarController`, React navigation state, reconciliation, and Fabric commit belong to the main UI runtime. The independently initialized background Hermes runtime has an isolated JS heap and does not own this UI state. Main/background bundles remain version-locked; no bg behavior or shared native storage is changed.

## Verification

- Rebasing onto current `x` upgraded `@onekeyfe/react-native-tab-view` to 3.0.95. The patch was regenerated from the pristine Yarn archive and independently passed `git apply --check`; it contains no build artifacts.
- A current-SHA production Release build succeeded with Xcode whole-module Swift compilation and split main/background Hermes bundles using `development/scripts/ios-release-build-deploy.sh`.
- Split-bundle integrity passed across 2344 main, 270 bg, and 26 bg-shared segments with zero cross-segment sync violations.
- Same iOS 26.5 simulator, same warm tab scenes, 30 fps diagnostic capture:
  - UIKit default animator: blended content remained visible across about 6-10 frames, roughly 200-330 ms;
  - zero-duration animator: content changed at one captured frame boundary with no blended content frame in either direction.
- The bottom Liquid Glass selection feedback continued animating independently across multiple frames.
- Wallet, Trade, Perps, Discover, repeated switching, and reselection were exercised without a crash; accessibility assertions confirmed each destination selected state.
- Temporary native/JS timing logs were removed; the final Release contained no diagnostic markers.
- `yarn agent:check --profile commit` passed. The PR profile reports all local checks passing, remote CI 25 passed / 0 failed / 0 pending, and no unresolved review threads; approval is still required.

Not covered: physical-device timing validation; the measured A/B was on the same iOS 26.5 simulator and Release configuration.

[OK-61355]: https://onekeyhq.atlassian.net/browse/OK-61355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/patches/@onekeyfe+react-native-tab-view+3.0.95.patch
+++ b/patches/@onekeyfe+react-native-tab-view+3.0.95.patch
@@ -1,0 +1,52 @@
+diff --git a/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewContainerView.swift b/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewContainerView.swift
+index 302f239..386998c 100644
+--- a/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewContainerView.swift
++++ b/node_modules/@onekeyfe/react-native-tab-view/ios/RCTTabViewContainerView.swift
+@@ -731,6 +731,27 @@ class RCTTabViewContainerView: UIView {
+   }
+ }
+ 
++// OneKey patch: keep native tab selection immediate while suppressing UIKit's content crossfade.
++private final class ImmediateTabTransitionAnimator: NSObject, UIViewControllerAnimatedTransitioning {
++  static let shared = ImmediateTabTransitionAnimator()
++
++  func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
++    0
++  }
++
++  func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
++    guard let toViewController = transitionContext.viewController(forKey: .to),
++          let toView = transitionContext.view(forKey: .to) ?? toViewController.view else {
++      transitionContext.completeTransition(false)
++      return
++    }
++
++    toView.frame = transitionContext.finalFrame(for: toViewController)
++    transitionContext.containerView.addSubview(toView)
++    transitionContext.completeTransition(!transitionContext.transitionWasCancelled)
++  }
++}
++
+ // MARK: - UITabBarControllerDelegate proxy
+ 
+ class TabViewDelegateProxy: NSObject, UITabBarControllerDelegate {
+@@ -757,6 +778,19 @@ class TabViewDelegateProxy: NSObject, UITabBarControllerDelegate {
+ 
+     return false
+   }
++
++  func tabBarController(
++    _ tabBarController: UITabBarController,
++    animationControllerForTransitionFrom fromViewController: UIViewController,
++    to toViewController: UIViewController
++  ) -> UIViewControllerAnimatedTransitioning? {
++    guard let containerView = mapping[ObjectIdentifier(tabBarController)],
++          containerView.disablePageAnimations else {
++      return nil
++    }
++
++    return ImmediateTabTransitionAnimator.shared
++  }
+ }
+ 
+ // MARK: - Long press gesture handler


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix:

@huhuanming

Reviewed commit `148603399c32`. This report always reflects the latest reviewed changes.

### Review summary

The iOS main-UI selection boundary moves from immediate UIKit ownership to an asynchronous UIKit → main-JS navigation state → native prop round-trip whenever page animations are disabled. Before: a tab tap emitted `onPageSelected` and UIKit immediately selected the view controller. After: the delegate emits the event but returns `false`; React Navigation must dispatch and commit the new `selectedPage` before native `updateSelectedTab()` performs the non-animated selection.

The patch reuses the existing `disablePageAnimations` prop and changes only the iOS dependency implementation. Reselection, the animations-enabled branch, programmatic selection handling, and Android remain unchanged. Extension is also unchanged. Desktop/web are single-runtime and unchanged, with no new process-owned native resource. On iOS, the `UITabBarController` is owned per navigation view in the main UI runtime, not shared with `bg`; navigation state has one main-JS heap copy, and `bg` initialization or readiness is not involved.

**What needs attention:** Verify that the iOS UIKit-to-main-JS selection handoff removes the content crossfade without restoring delayed tab feedback when the main JS runtime is busy.

### Issues to address

- **P2 · Disabled-animation taps now wait for the JS round-trip**
  - **Impact:** `TabStackNavigator.native.tsx` enables `disablePageAnimations` for every native bottom-tab navigator. Returning `false` from the iOS delegate therefore vetoes UIKit's immediate selection on every non-reselection tap, and the icon, selection indicator, and content cannot move until `onPageSelected` reaches the main JS runtime, React Navigation dispatches, React commits, and `selectedPage` returns to native code. Under main-JS load this restores the delayed feedback that the existing native selected-image and early-return path was designed to avoid.
  - **Suggested action:** Keep the tap response in the native runtime by synchronously assigning the target `selectedIndex` through the existing no-animation native path before returning `false`, then emit the JS event only to reconcile React Navigation state. Verify the result on iOS with an intentionally busy JS thread.

### Validation gaps

- Patch syntax and build-artifact scope were checked; the patch also applies cleanly to the cached nearby 3.0.90 source.
- Exact installation against `@onekeyfe/react-native-tab-view@3.0.94` was not exercised.
- No iOS device/simulator interaction test was run for normal taps, reselection, busy-main-JS behavior, or content-crossfade suppression.
- No automated test covers the native/JS selection handoff.

huhuanming:
CI status after the final push: 21 passed, 4 failed, 0 pending.

All four red checks have the same base-branch cause rather than four independent failures:

- Native startup graph budget exits before analysis because module ID `23787` is assigned to `AddressInputWarnings.tsx`, `TradingViewChartLoadingMask.tsx`, and `usePrimeSubscriptionManagementTarget.ts`.
- Unit Tests 3/5 (`devVendor.test.js`) and 5/5 (`moduleIdRegistry.test.js`) fail while loading that same invalid registry; `unittest (24.x)` is only their aggregate failure.

Those modules and registry entries were merged into `x` after this branch was created and are unrelated to this Swift-only patch. The repository-level registry fix is tracked by #13081. The failed jobs should be rerun after that conflict is resolved on `x`.

originalix (approved):

PR #13093, opened 2026-08-31
fix: make debounced actions respond immediately OK-61507

huhuanming:
[OK-61507](https://onekeyhq.atlassian.net/browse/OK-61507)

----------

## Summary

- open imperative action-list and trigger interactions on the leading edge
- suppress duplicate interactions during the 350 ms window without replaying a trailing action
- align the same immediate-interaction debounce pattern in Bootstrap and Swap controls while preserving trailing debounce for state-coalescing paths

## Testing

- iOS simulator: Wallet Home More opens on the first tap
- iOS simulator: repeated taps within 350 ms produce only one More sheet

## Issue

- https://onekeyhq.atlassian.net/browse/OK-61507

[OK-61507]: https://onekeyhq.atlassian.net/browse/OK-61507?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/actions/ActionList/index.tsx
+++ b/packages/components/src/actions/ActionList/index.tsx
@@ -568,6 +568,7 @@ const showActionList = (
       </PageContext.Provider>
     </ModalNavigatorContext.Provider>,
   );
+  return ref;
 };
 function ActionListFrame(props: IActionListProps) {
   const isProcessing = useRef(false);
@@ -606,7 +607,12 @@ function ActionListFrame(props: IActionListProps) {
   );
 }
 
-const show = (props: IShowActionListParams) => showActionList(props, undefined);
+// Imperative action lists share one overlay slot; newer calls replace the active one.
+let imperativeActionListRef: ReturnType<typeof showActionList> | undefined;
+const show = (props: IShowActionListParams) => {
+  imperativeActionListRef?.destroy();
+  imperativeActionListRef = showActionList(props, undefined);
+};
 
 export const ActionList = withStaticProperties(ActionListFrame, {
   show,
--- a/packages/components/src/actions/ActionList/index.tsx
+++ b/packages/components/src/actions/ActionList/index.tsx
@@ -481,19 +481,26 @@ const showActionList = (
   // Use let so the destroy callback can reference it after assignment
   // eslint-disable-next-line prefer-const
   let ref: { destroy: () => void };
+  let isClosed = false;
 
   // eslint-disable-next-line react-perf/jsx-no-new-function-as-prop
   const handleOpenChange = (isOpen: boolean) => {
-    restProps.onOpenChange?.(isOpen);
-    if (!isOpen) {
-      setTimeout(() => {
-        restProps.onClose?.();
-      });
-      // delay the destruction of the reference to allow for the completion of the animation transition.
-      setTimeout(() => {
-        ref.destroy();
-      }, 500);
+    if (isOpen) {
+      restProps.onOpenChange?.(true);
+      return;
     }
+    if (isClosed) {
+      return;
+    }
+    isClosed = true;
+    restProps.onOpenChange?.(false);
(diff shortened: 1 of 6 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/components/src/actions/ActionList/index.tsx
```diff
@@ -573,6 +573,10 @@ const showActionList = (
 const debouncedShowActionList = debounce(
   showActionList,
   PROCESSING_RESET_DELAY,
+  {
+    leading: true,
+    trailing: false,
+  },
```
sidmorizon: **Unrelated menus suppress each other**

Reported at `packages/components/src/actions/ActionList/index.tsx:576-579` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

Calls to `ActionList.show` from different controls share one 350 ms cooldown, so the later menu is discarded. A quick second action opens no menu.

huhuanming: Fixed in `576ae51215`. `ActionList.show` now opens directly, so unrelated imperative menus no longer share a global 350 ms cooldown. The declarative ActionList keeps its per-instance guard. I also stabilized the Trigger debounce instance and preserved the default Button `debounce(0)` timing called out in the general review.

packages/components/src/actions/ActionList/index.tsx
```diff
@@ -612,8 +606,7 @@ function ActionListFrame(props: IActionListProps) {
   );
 }
 
-const show = (props: IShowActionListParams) =>
-  debouncedShowActionList(props, undefined);
+const show = (props: IShowActionListParams) => showActionList(props, undefined);
```
sidmorizon: **Imperative menus open multiple times**

Severity: `non-severe`

Rapid calls to `ActionList.show` bypass the 350 ms guard and create a portal for every call. Duplicate taps can stack multiple menus.

huhuanming: Fixed in `4be1abc349`. `ActionList.show` now keeps a single imperative portal: each new call destroys the previously active portal and opens immediately, so calls neither stack nor share a debounce cooldown. The declarative ActionList keeps its existing per-instance guard.

packages/components/src/actions/ActionList/index.tsx
```diff
@@ -612,8 +607,12 @@ function ActionListFrame(props: IActionListProps) {
   );
 }
 
-const show = (props: IShowActionListParams) =>
-  debouncedShowActionList(props, undefined);
+// Imperative action lists share one overlay slot; newer calls replace the active one.
+let imperativeActionListRef: ReturnType<typeof showActionList> | undefined;
+const show = (props: IShowActionListParams) => {
+  imperativeActionListRef?.destroy();
```
sidmorizon: **Replaced menus skip dismissal callbacks**

Reported at `packages/components/src/actions/ActionList/index.tsx:613` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

When another menu opens, `destroy()` bypasses the current menu's `onClose` and `onOpenChange(false)` callbacks. Callers can retain stale state or miss cleanup.

huhuanming: Fixed in `b9caee583c`. Imperative replacements now call the previous action list controller’s `close()` path before destroying its portal. That path emits `onOpenChange(false)` and schedules `onClose`, while an `isClosed` guard ensures the callbacks run only once if the menu was already closing.

originalix (approved):

originalix:

Review of commit `b9caee583c55` did not finish successfully.

PR #13095, opened 2026-08-31
fix: avoid redundant PIN prompt in advanced settings

wabicai:
## Summary

- replace the advanced-settings unconditional unlock with the shared conditional unlock flow
- use Protocol V2 runtime state to decide whether unlock is needed, then read settings state after unlock
- keep Protocol V1 settings behavior and reuse the existing hardware operation lease

## Related

- OneKeyHQ/hardware-js-sdk#905 fixes the same redundant Main PIN behavior during standard-wallet onboarding/address creation

## Test plan

- [x] `yarn jest packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.pro2.test.ts packages/kit-bg/src/services/ServiceHardware/ServiceHardware.getCompatibleConnectId.test.ts --runInBand`
- [x] targeted `oxfmt --check` and type-aware `oxlint`
- [x] `git diff --check`


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix:

@wabicai

Reviewed commit `e3a28077eaf4`. This report always reflects the latest reviewed changes.

### Review summary

The hardware advanced-settings read path changed from always calling unlockDevice and then reading settings to acquiring one operation lease and using getDeviceStateWithUnlock for a conditional unlock. Protocol V2 reads runtime state first, then rereads settings; support-feature and input-PIN reporting remain unchanged, and Trezor handling is unchanged. The new non-V2 branch also routes legacy devices through the DeviceState unlock helper, which is incompatible with the legacy Features response shape and must be corrected.

**What needs attention:** Verify that legacy Protocol V1 settings reads preserve the legacy Features response contract while Protocol V2 uses the leased runtime-then-settings sequence.

### Issues to address

- **P1 · Protocol V1 advanced-settings reads pass the wrong state shape**
  - **Impact:** For Touch and other non-Protocol-V2 devices, getDeviceState resolves the legacy Features payload, while getDeviceStateWithUnlock dereferences state.status. The new non-V2 branch therefore throws before returning advanced settings, breaking the legacy hardware settings screen.
  - **Suggested action:** Keep the existing unlockDevice plus getDeviceStateByWallet(settings) flow for Protocol V1, and use getDeviceStateWithUnlock only for Protocol V2, or add an explicit legacy-state normalization before invoking the helper.

### Validation gaps

- Targeted Jest execution was attempted but could not start because the workspace has no Yarn node_modules state file; runtime behavior still needs verification after dependencies are installed.

ByteZhang1024 (approved):

PR #13103, opened 2026-09-01
fix: stabilize wallet home asset refresh(OK-61576)

weatherstar:
[OK-61576](https://onekeyhq.atlassian.net/browse/OK-61576)

----------

## Summary
- Commit the authoritative Home asset snapshot before running non-critical wallet asset-status analytics.
- Serialize Home all-network refresh requests across debounce and in-flight windows so explicit refreshes are not dropped.
- Introduce per-key asset snapshot freshness metadata (`localSeq` + HTTP Date marker) and apply last-writer-wins admission at every persistence/state layer, so an out-of-order stale response can no longer overwrite newer data.
- Harden the failure paths surfaced by review: restore the last-good published result when an accepted refresh fails, guard against stale debounced runners from a previous owner, and self-heal the sequence counter after a device clock rollback.

## Intent & Context
Users reported that wallet balances and prices could remain stale or inconsistent after repeated refreshes, app background/foreground transitions, and switching between mobile and desktop. The investigation showed successful token-list API responses followed by delayed or missing UI publication.

## Root Cause
1. The Home refresh path awaited a slow asset-status analytics workflow before publishing the already-built authoritative snapshot. A logged `getAllHdHwQrWallets` call took about 93 seconds, and iOS suspension amplified the delay.
2. `useAllNetworkRequests` allowed manual/event refreshes to overlap its internal debounce and in-flight queue; queued must-run intent could be lost or filtered as a redundant run while the previous result remained visible.
3. With refreshes now firing more aggressively, out-of-order responses could overwrite newer persisted data (token-list cache, account values, overview worth) with stale payloads.

## Design Decisions
- Move the non-critical analytics work after the authoritative snapshot commit and refresh-state update so UI freshness does not depend on analytics latency.
- Route Home refreshes through the existing queue and preserve explicit refresh flags, without changing the generic `usePromiseResult` behavior or unrelated consumers. The Home-only flag is named `clearRetainedResultOnAcceptedRun` to avoid confusion with `usePromiseResult`'s same-named-but-different `undefinedResultIfReRun` option.
- Mint one monotonic `localSeq` per token fetch in the background runtime (the single sequencer) and attach it to responses as `assetSnapshotMeta`. Every write layer (`SimpleDbEntityLocalTokens`, `SimpleDbEntityAccountValue`, `activeAccountValueAtom` via `ServiceAccountProfile`, `accountWorthAtom` via accountOverview actions) admits a write only when it is newer than the stored marker; unversioned legacy writes may initialize a key but never clobber versioned data.
- A full-snapshot replacement (which may delete omitted keys) requires an aggregate marker proving every enabled network was observed by the same refresh; otherwise the write degrades to a per-network merge.
- The freshness admission helpers live in `@onekeyhq/shared/src/utils/assetSnapshotFreshness` and are shared by all layers, so the comparison semantics cannot drift.
- Clock-rollback resilience: comparing against a persisted marker lifts the in-memory sequence watermark, so after a backward clock jump the very next refresh mints a newer sequence instead of being rejected indefinitely.
- Keep the OTA change limited to shared JavaScript; no persistence schema (SimpleDB fields are additive/optional), native watchdog, dependency, or unrelated refactor changes are included.

## Changes Detail
Refresh scheduling and publication:
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx`: opt Home into retained-result handling, publish the authoritative snapshot before asset-status analytics, and drop the duplicate `AddDBAccountsToWallet` refresh (the hook-level wallet-scoped listener already forces one fan-out with an accounts-cache bypass; only the LP token list keeps a dedicated trigger).
- `packages/kit/src/hooks/useAllNetwork.ts`: reserve debounce windows, merge queued must-run configuration, distinguish accepted refreshes from redundant runs; restore the last-good published result when an accepted fan-out fails before publication; bail out stale runners whose closure belongs to a previous owner (they could otherwise clear the new owner's result and run a ghost fan-out).
- `packages/kit/src/hooks/allNetworkRunResultUtils.ts`: pure `resolveAllNetworkFailedRunRestore` decision + tests.

Snapshot freshness admission:
- `packages/shared/src/utils/assetSnapshotFreshness.ts` (+ `packages/shared/types/assetSnapshot.ts`): sequence minting, header parsing, tolerant comparison, shared admission helpers (`canApplyAssetSnapshotMeta`, `getNewestAssetSnapshotMeta`, …), clock-rollback watermark self-heal.
- `packages/kit-bg/src/services/ServiceToken.ts`: mint the marker before each fetch, thread it through responses and the local token cache; failed debounced cache flushes merge back and reschedule persistence.
- `packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityLocalTokens.ts` / `SimpleDbEntityAccountValue.ts`: versioned admission inside the entity mutex for token-list slices and account values (full-replace vs per-network merge).
- `packages/kit-bg/src/services/ServiceAccountProfile.ts`: mutex-serialized active-account-value atom updates with owner-aware freshness admission; the old "value can only go up" heuristic is replaced by the freshness gate (legitimate balance decreases now propagate).
- `packages/kit/src/states/jotai/contexts/accountOverview/actions.ts` / `atoms.ts`: worth-map admission with per-key and aggregate markers; create-at-network scalar recomputed from materialized absolute values.
- `packages/kit/src/views/Home/pages/HomeOverviewContainer.tsx`: persist per-network values by compound key with their markers; filtered TokenSelector responses no longer persist as canonical network totals (`packages/kit/src/views/AssetSelector/pages/TokenSelector.tsx`).

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile / Desktop / Web / Extension (shared Home, all-network, and persistence code paths)
- **Risk Areas**: Refresh scheduling and result publication during account/network switches, app lifecycle transitions, and concurrent manual/dependency triggers; freshness admission on the token-list cache and account-value persistence (stale writes are now rejected by design). Native runtime suspension can still delay JavaScript execution until resume.

## Test plan
- [x] `yarn jest packages/kit/src/views/Home/components/TokenListBlock packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAssetSnapshot.test.ts packages/kit-bg/src/services/ServiceAccountProfile.updateAllNetworkAccountValue.test.ts packages/kit/src/states/jotai/contexts/accountOverview/actions.test.tsx packages/shared/src/utils/assetSnapshotFreshness.test.ts packages/kit/src/hooks/allNetworkRunResultUtils.test.ts --runInBand` (12 suites, 96 tests passed)
- [x] `yarn agent:check --profile commit` (lint-worktree-ts, lint-staged, tsc-staged all passed)
- [ ] Verify repeated refresh and background/foreground recovery on iOS and desktop release builds
- [ ] Failure recovery: force `getAllNetworkAccounts` / `onStarted` to fail mid-refresh; the list must keep the last authoritative snapshot and later pull-to-refresh must still commit
- [ ] Switch accounts within 1s of a pull-to-refresh; the new account's data must not be cleared
- [ ] Clock rollback: set the device clock forward, refresh, set it back, restart; balances must keep updating
- [ ] Single-network balance decrease propagates after the removal of the "only increase" guard; OTA upgrade from a build with unversioned caches admits the first versioned write

[OK-61576]: https://onekeyhq.atlassian.net/browse/OK-61576?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
+++ b/packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
@@ -5,6 +5,15 @@ import {
 } from '@onekeyhq/shared/src/eventBus/appEventBus';
 import { defaultLogger } from '@onekeyhq/shared/src/logger/logger';
 import accountUtils from '@onekeyhq/shared/src/utils/accountUtils';
+import {
+  canApplyAssetSnapshotMeta as canApplySnapshotMeta,
+  getNewestAssetSnapshotMeta,
+  isAssetSnapshotNewer,
+  isAssetSnapshotSameOrNewer,
+  normalizeAssetSnapshotMeta as normalizeSnapshotMeta,
+  sameAssetSnapshotMeta as sameSnapshotMeta,
+} from '@onekeyhq/shared/src/utils/assetSnapshotFreshness';
+import type { IAssetSnapshotMeta } from '@onekeyhq/shared/types/assetSnapshot';
 
 import { SimpleDbEntityBase } from '../base/SimpleDbEntityBase';
 
@@ -14,6 +23,8 @@ import type ServiceAccount from '../../../services/ServiceAccount/ServiceAccount
 export interface IAccountValueEntry {
   value: string;
   currency: 'usd';
+  /** Freshness of the response that produced this single-network value. */
+  assetSnapshotMeta?: IAssetSnapshotMeta;
 }
 
 // Stored value entry for All Networks aggregate, keyed only by addressOrXpub.
@@ -22,6 +33,10 @@ export interface IAccountValueEntry {
 export interface IAllNetworkAccountValueEntry {
   value: Record<string, string>; // <networkId, value>
   currency: 'usd';
+  /** Freshness of the complete all-network snapshot, when known. */
+  assetSnapshotMeta?: IAssetSnapshotMeta;
+  /** Freshness for each network in the value map (partial writes need this). */
+  assetSnapshotMetaByNetwork?: Record<string, IAssetSnapshotMeta>;
 }
 
 export interface IAccountValueDb {
@@ -34,10 +49,22 @@ export interface IAccountValueDb {
 
   // Legacy fields preserved during the one-shot address-key migration so a rollback
   // PR can keep reading old data. Cleaned up in a later release.
-  _legacy_data?: Record<string, { value: string; currency: string }>;
+  _legacy_data?: Record<
+    string,
+    {
+      value: string;
+      currency: string;
+      assetSnapshotMeta?: IAssetSnapshotMeta;
+    }
+  >;
   _legacy_all?: Record<
     string,
-    { value: Record<string, string>; currency: string }
+    {
+      value: Record<string, string>;
+      currency: string;
+      assetSnapshotMeta?: IAssetSnapshotMeta;
+      assetSnapshotMetaByNetwork?: Record<string, IAssetSnapshotMeta>;
+    }
   >;
   _migratedAt?: number;
   // Migration version. Bumped when the migration logic itself is corrected so we
@@ -64,10 +91,77 @@ export interface IAccountValueAllWriteItem {
   xpub?: string;
   networkId: string;
   value: string;
+  assetSnapshotMeta?: IAssetSnapshotMeta;
+}
+
+// Freshness admission helpers live in
+// `@onekeyhq/shared/src/utils/assetSnapshotFreshness` so every persistence
+// layer applies the identical comparison. This adapter only bridges the
+// array-shaped call sites in this entity.
+function maxSnapshotMeta(
+  values: Array<IAssetSnapshotMeta | undefined>,
+): IAssetSnapshotMeta | undefined {
+  return getNewestAssetSnapshotMeta(...values);
+}
+
+function isSameStringMap(
+  left: Record<string, string>,
+  right: Record<string, string>,
+): boolean {
+  const leftKeys = Object.keys(left);
+  return (
+    leftKeys.length === Object.keys(right).length &&
+    leftKeys.every((key) => right[key] === left[key])
+  );
+}
+
+function isSameMetaMap(
+  left: Record<string, IAssetSnapshotMeta> | undefined,
+  right: Record<string, IAssetSnapshotMeta> | undefined,
+): boolean {
+  const leftKeys = Object.keys(left ?? {});
+  return (
+    leftKeys.length === Object.keys(right ?? {}).length &&
+    leftKeys.every(
+      (key) =>
+        Boolean(right?.[key]) && sameSnapshotMeta(left?.[key], right?.[key]),
+    )
+  );
+}
+
+// Value-level equality of a persisted all-network entry, used to skip
+// re-serializing the entity when a refresh repeats what is already stored.
+function isSameAllNetworkEntry(
+  left: IAllNetworkAccountValueEntry,
+  right: IAllNetworkAccountValueEntry,
+): boolean {
+  return (
+    left.currency === right.currency &&
+    isSameStringMap(left.value, right.value) &&
+    sameSnapshotMeta(left.assetSnapshotMeta, right.assetSnapshotMeta) &&
+    isSameMetaMap(
+      left.assetSnapshotMetaByNetwork,
+      right.assetSnapshotMetaByNetwork,
+    )
+  );
 }
 
 function emptyData(): IAccountValueDb {
   return { byAddress: {}, allByAddress: {} };
+}
+
+// A record persisted before the address-key migration (which runs deferred at
+// bootstrap) still has the legacy `data` / `all` shape and lacks the
+// address-keyed buckets. Writers must index normalized buckets rather than
+// `undefined`, otherwise every refresh throws until the migration lands.
+function withAddressBuckets(
+  raw: IAccountValueDb | null | undefined,
+): IAccountValueDb {
+  return {
+    ...raw,
+    byAddress: raw?.byAddress ?? {},
+    allByAddress: raw?.allByAddress ?? {},
+  };
 }
 
 export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValueDb> {
@@ -111,6 +205,9 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
       return {
         value: entry?.value,
         currency: entry?.currency,
+        ...(entry?.assetSnapshotMeta
+          ? { assetSnapshotMeta: entry.assetSnapshotMeta }
+          : {}),
       };
     });
   }
@@ -121,28 +218,53 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
     xpub,
     value,
     currency,
+    assetSnapshotMeta,
   }: {
     networkId: string;
     accountAddress?: string;
     xpub?: string;
     value: string;
     currency: 'usd';
+    assetSnapshotMeta?: IAssetSnapshotMeta;
   }) {
     const key = this.buildSingleKey({ networkId, accountAddress, xpub });
     if (!key) {
       return;
     }
-    const existing = (await this.getRawData())?.byAddress?.[key];
-    if (existing?.value === value && existing?.currency === currency) {
+    const normalizedMeta = normalizeSnapshotMeta(assetSnapshotMeta);
+    // Rejected (stale) or identical writes are no-ops. An unchanged value with
+    // a NEWER marker is not: the marker must be persisted, otherwise a later
+    // stale response would be admitted against the old one.
+    const isNoopAgainst = (existing: IAccountValueEntry | undefined) =>
+      !canApplySnapshotMeta(assetSnapshotMeta, existing?.assetSnapshotMeta) ||
+      (existing?.value === value &&
+        existing?.currency === currency &&
+        sameSnapshotMeta(assetSnapshotMeta, existing?.assetSnapshotMeta));
+    // Pre-check outside the mutex so a repeated refresh does not re-serialize
+    // the whole entity (see updateAllNetworkAccountValue).
+    if (
+      isNoopAgainst(withAddressBuckets(await this.getRawData()).byAddress[key])
+    ) {
       return;
     }
     await this.setRawData((rawData) => {
-      const base = rawData ?? emptyData();
+      const base = withAddressBuckets(rawData);
+      const existing = base.byAddress[key];
+      // The authoritative comparison happens inside the builder: setRawData
+      // serializes this entity's writes, whereas the pre-read above can be
+      // stale by the time the write acquires the mutex.
+      if (isNoopAgainst(existing)) {
+        return base;
+      }
       return {
         ...base,
         byAddress: {
           ...base.byAddress,
-          [key]: { value, currency },
+          [key]: {
+            value,
+            currency,
+            ...(normalizedMeta ? { assetSnapshotMeta: normalizedMeta } : {}),
+          },
         },
       };
     });
@@ -160,6 +282,12 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
       return {
         value: entry?.value,
         currency: entry?.currency,
+        ...(entry?.assetSnapshotMeta
+          ? { assetSnapshotMeta: entry.assetSnapshotMeta }
+          : {}),
+        ...(entry?.assetSnapshotMetaByNetwork
+          ? { assetSnapshotMetaByNetwork: entry.assetSnapshotMetaByNetwork }
+          : {}),
       };
     });
   }
@@ -168,6 +296,7 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
     items,
     currency,
     updateAll,
+    snapshotMeta,
   }: {
     items: IAccountValueAllWriteItem[];
     currency: 'usd';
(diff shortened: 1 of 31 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix:

@weatherstar

Reviewed commit `604901406225`. This report always reflects the latest reviewed changes.

### Review summary

The change adds freshness metadata to wallet asset snapshots across the background fetch, persistence, and foreground account-overview paths. Before, each settled response could replace shared worth/cache state without a durable ordering marker; after, background fetches mint request-order metadata, HTTP Date is retained as a tie-breaker, and the active atom plus SimpleDB admit or reject writes by snapshot freshness. Full all-network replacements additionally require a complete marker and preserve newer omitted networks. Cache migration tolerance, no-op writes, token-list propagation, and existing token rendering/filter behavior remain otherwise unchanged. The foreground and background runtimes each hold their own JavaScript copies while the background runtime persists through shared native storage, so marker propagation is the synchronization boundary.

**What needs attention:** Verify that a versioned merge:false single-network refresh replaces the scoped worth map while retaining only a stale-safe value for the requested key.

### Issues to address

- **P1 · Versioned single-network updates retain previous-network balances**
  - **Impact:** A merge:false single-network response carries only a per-key marker, but the replacement gate requires an aggregate marker. When the account ID is shared across networks, omitted entries from the previously selected network remain in accountWorth. Consumers such as useHomeBalanceState and HomeOverviewContainer aggregate all entries, so a zero-balance network can appear funded or exceed thresholds using stale data.
  - **Suggested action:** Handle merge:false single-network updates as scoped replacements: apply freshness only to supplied keys, preserve a newer value for the same key when necessary, and evict omitted network keys without requiring an all-network aggregate marker.

### Validation gaps

- Targeted Jest tests were not executed because the checkout has no installed node_modules state (`Couldn't find the node_modules state file`).
- GitHub review/comment surfaces could not be fetched from the current environment, so the inline candidate is marked for human review rather than automatic publication.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
```diff
@@ -46,39 +86,265 @@ class ContextJotaiActionsAccountOverview extends ContextJotaiActionsBase {
...
+        (hasCompleteIncomingSnapshotMeta &&
+          incomingKeysAreFresh &&
+          completeMetaIsFresh);
+      const preserveCurrentValues =
+        sameAccount && (!isCompleteSnapshot || !canReplaceFullSnapshot);
+      const nextWorth: Record<string, string> = preserveCurrentValues
+        ? { ...currentValue }
+        : {};
+      const nextMetaByKey: Record<string, IAssetSnapshotMeta> =
+        preserveCurrentValues ? { ...currentMetaByKey } : {};
```
sidmorizon, line 283: 🟠 P1: Single-network switches retain prior balances

Severity: `severe`

In the main-runtime Jotai copy, `preserveCurrentValues` retains the previous network when a versioned account switches networks. Merge-derived views sum both values and inflate the balance; no shared native resource or background copy participates.

packages/kit/src/hooks/useAllNetwork.ts
```diff
@@ -542,10 +596,24 @@ function useAllNetworkRequests<T>(params: {
...
+        }
         return;
       }
       lastRunSignatureRef.current = currentRunSignature;
 
+      if (undefinedResultIfReRun) {
+        // Keep a queued refresh from restoring the prior published result
+        // after the new run has already invalidated it.
+        clearResultRef.current?.();
+        lastPublishedResultRef.current = undefined;
```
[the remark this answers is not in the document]
weatherstar: Fixed in 65b860963a.

The runner now snapshots its owner key (`account|network|wallet|isAllNetworks`) when the closure is created and compares it against a render-updated ref (`liveRunOwnerKeyRef`) after the debounce wait. A stale runner from a previous owner bails out before the redundant-run gate, so it can no longer clear the new owner's published result through the direct `setResult` path, corrupt `lastRunSignatureRef`, or start a ghost fan-out (runtime scope: main).

Additionally, an accepted run that fails before publication now restores the invalidated last-good result (`resolveAllNetworkFailedRunRestore`, owner-checked), so the failure path cannot pin the consumer on `undefined` either.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
weatherstar: Analyzed with full codebase context — keeping this behavior as designed, no code change for now. Reasoning:

1. **The retention is the atom's intended cross-network-map semantics** (codified by the new tests, e.g. `retains omitted networks when an unversioned all-network cache is partial`). `worth` is keyed by `(accountId, networkId)`, so a retained entry is the *other* network's real balance under its own key — never a second copy of the current network's value, so per-key consumers cannot double count.
2. **The displayed Home total does not sum this map.** It reads `lastConfirmedOverviewBalance.byOwner[ownerKey]` (keyed per owner+network, `HomeOverviewContainer.tsx`), so the visible balance is unaffected by retained sibling-network keys.
3. The only consumers that sum the whole map are the walletStatus `hasValue` threshold (`HomeOverviewContainer.tsx` `allWorth`) and `useHomeBalanceState`'s funded-positivity bucket. Both are wallet/account-level signals where cross-network aggregation is semantically acceptable, and both already saw the same values whenever the all-network path had run.
4. HD accounts switching across impls (e.g. ETH -> BTC) change the payload `accountId`, so `sameAccount` is false and a full replacement clears the map. The retention window is limited to same-account-id families (EVM networks, Others accounts) — exactly where the values remain valid balances.

If we later want single-network views to drop sibling-network keys, that should be a deliberate change on the reset path in `HomeOverviewContainer`, not in the freshness admission logic. Leaving this thread open for your confirmation — happy to follow up if there is a concrete surface where the summed value is user-visible.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
```diff
@@ -46,39 +78,265 @@ class ContextJotaiActionsAccountOverview extends ContextJotaiActionsBase {
...
+          : {};
+      const incomingKeysAreFresh = Object.keys(payload.worth).every((key) => {
+        const incomingMeta = incomingMetaForKey(key);
+        const existingMeta = getNewestAssetSnapshotMeta(
+          currentMetaByKey[key],
+          currentAggregateMeta,
+        );
+        return (
+          !existingMeta || isAssetSnapshotNewer(incomingMeta, existingMeta)
+        );
```
sidmorizon, line 240: 🟠 P1: Full refreshes cannot remove stale networks

Severity: `severe`

After progressive updates admit a response, `incomingKeysAreFresh` rejects its full snapshot's equal markers. Main-JS, bg-JS, and shared SimpleDB retain omitted networks.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
```diff
@@ -46,39 +78,265 @@ class ContextJotaiActionsAccountOverview extends ContextJotaiActionsBase {
...
-          },
-          createAtNetworkWorth: new BigNumber(createAtNetworkWorth ?? '0')
-            .plus(payload.createAtNetworkWorth ?? '0')
-            .toFixed(),
+          worth: nextWorth,
+          createAtNetworkWorth: nextCreateAtNetworkWorth,
           initialized: payload.initialized,
           accountId: payload.accountId,
           updateAll: payload.updateAll,
           currency,
```
sidmorizon: **Rejected snapshots retag retained balances**

Severity: `non-severe`

When freshness rejects every incoming value, `currency` still adopts the rejected snapshot's tag. Main-JS then converts retained balances from the wrong currency.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
weatherstar: Confirmed and fixed in d39d97b670.

Root cause: the progressive per-network `merge: true` path admits each response as it settles, and the full snapshot built from the same round (`buildMergedAllNetworkSnapshot`) re-materializes those responses with EQUAL per-key markers. `incomingKeysAreFresh` used the strict `isAssetSnapshotNewer`, and `completeMetaIsFresh` required the aggregate (= oldest marker of the round) to be strictly newer than EVERY stored per-key marker — so after progressive updates a full snapshot could never be admitted and omitted networks were retained. The same strict gate existed in bg `ServiceAccountProfile.updateAllNetworkAccountValue` (activeAccountValueAtom) and `SimpleDbEntityAccountValue.updateAllNetworkAccountValue`.

Fix (all three layers):
- New `isAssetSnapshotSameOrNewer`: keys the full snapshot supplies are admitted when their marker is not older (equal included).
- The aggregate marker is compared only against the stored aggregate (same-or-newer) and against the markers of keys the snapshot OMITS (strictly newer, since those keys get evicted) — no longer against every stored per-key marker.
- An admitted full snapshot takes the incoming value for equal-marker keys too, and promotes its aggregate marker.

Tests added in `actions.test.tsx`, `ServiceAccountProfile.updateAllNetworkAccountValue.test.ts` and `SimpleDbEntityAssetSnapshot.test.ts`: progressive merges followed by the same-round full snapshot now evict the omitted network; a full snapshot whose omitted key carries a newer marker is still refused.

packages/kit/src/states/jotai/contexts/accountOverview/actions.ts
weatherstar: Fixed in d39d97b670.

Both the `merge` path and the replacement path now compute `nextCurrency`: when every incoming value was rejected as stale and the atom keeps only retained values for the same account, `current.currency` is preserved instead of adopting the rejected payload's tag. As soon as any incoming value is accepted (or the owner changes / nothing is retained), the payload tag applies as before. Regression test: `keeps the currency tag of retained values when a replacement rejects every incoming value` in `actions.test.tsx`.

packages/kit/src/hooks/useAllNetwork.ts
```diff
@@ -542,10 +644,27 @@ function useAllNetworkRequests<T>(params: {
...
       lastRunSignatureRef.current = currentRunSignature;
 
+      // Captured for the failure path: an accepted run invalidates the
+      // retained result below, and a failed fan-out must restore it.
+      const previousPublishedResult = lastPublishedResultRef.current;
+      if (clearRetainedResultOnAcceptedRun) {
+        // Keep a queued refresh from restoring the prior published result
+        // after the new run has already invalidated it.
+        applyResultRef.current?.(undefined);
+        lastPublishedResultRef.current = undefined;
```
sidmorizon, line 722: # [🟠 P1] 被抢占的成功结果被整体丢弃，后继失败时无快照可恢复

## 问题

accepted run 起跑时这里把 `lastPublishedResultRef.current` 置为 `undefined`。该轮 fan-out **成功**结束后，若期间排队了 must-run（finally 里 `hasQueuedRerun = scheduleQueuedRerun()` 返回 `true`），`resolveAllNetworkPublishedResult` 会走 superseded 分支：

```ts
if (hasQueuedRerun) {
  return {
    publishedResult:
      lastPublished?.runSignature === runSignature ? lastPublished.result : undefined,
    nextLastPublished: lastPublished, // 仍然是刚被清空的 undefined
  };
}
```

于是本轮刚拿到的 `completedResult` 既不发布、也不写回 `lastPublishedResultRef`，最近一次「好快照」被彻底丢失。

## 影响

紧接着的排队 run 若 fan-out 抛错，`catch` 里 `resolveAllNetworkFailedRunRestore` 拿到的 `previousPublished` 是 `undefined`，`shouldRestoreResult` 为 `false`，`allNetworksResult` 会一直停在 `undefined`。`updateAllNetworksTokenList` 首行 `if (!allNetworksResult?.length) return;` 直接返回，`commitAuthoritativeIngest` 与 `updateAccountWorth` 都不再执行——首页总资产与权威快照停更，直到下一次成功的 must-run。触发路径很常见：`下拉刷新` → 期间再触发一次显式刷新（`AddDBAccountsToWallet` / `EnabledNetworksChanged` / 手动刷新）→ 第二轮任一网络请求失败。改动前 `lastPublishedResultRef` 从不被清空，这个窗口里消费者始终能拿到上一次的完整结果，属于本 PR 新引入的回归。

## 建议

superseded 分支不要退回到已被清空的 `lastPublished`：把本轮的 `completedResult` 连同 `currentRunSignature` 写入 `nextLastPublished`（只是不作为 `publishedResult` 对外发布）。这样既保留「被抢占则不发布」的语义，后继 run 失败时 `resolveAllNetworkFailedRunRestore` 也才有真正可恢复的最近好快照。

[Claude session ↗](https://claude.ai/code/session_01GJHvviFfaN7xu1dXH27f57)

---

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
```diff
@@ -181,53 +248,178 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
...
-      }
-      return Object.entries(valueMap).every(
-        ([nId, v]) => prev.value[nId] === v,
-      );
-    });
-    if (isNoop) {
-      return;
-    }
-
     await this.setRawData((rawData) => {
```
sidmorizon: # [🟡 P2] 删掉写前 no-op 短路后，每次刷新都全量重写整个 accountValue 实体

## 问题

改动前 `updateAllNetworkAccountValue` 会先算 `isNoop`，无变化时直接 `return`，根本不进入 `setRawData`。现在这段短路被删除，任何调用都会执行这里的 `await this.setRawData(...)`。即使 builder 内所有 key 都被准入逻辑拒绝、`changed` 全为 `false`，返回值仍是新对象 `{ ...base, allByAddress: next }`，`SimpleDbEntityBase.setRawData` 依旧会对**整份实体**做一次 `JSON.stringify` + `appStorage.setItem`。同文件的 `updateAccountValue` 也把写前的 `existing?.value === value` 短路挪进了 builder（`return base`），同样无法避免落盘；而且它的 no-op 判定额外要求 `sameSnapshotMeta(...)`，一旦携带新 marker 就必然写盘，即使余额一字未变。

## 影响

运行时作用域：`bg`（iOS/Android/扩展为独立 JS 运行时，桌面/Web 与 main 同进程）。`HomeOverviewContainer` 的 effect 依赖 `accountWorth` 对象引用与 `accountWorth.worth`，全网刷新时每个网络 settle 都会生成新的 atom 快照并触发一次调用。`N` 个已启用网络 ≈ `N` 次「整份 `allByAddress` 序列化 + 存储写」，且全部串行排队在实体 `mutex` 上；改动前其中绝大多数被 `isNoop` 短路。网络数较多的移动端上这是可观的 IO 与序列化放大，也会拉长其他 accountValue 写入的排队时间。

## 建议

恢复一条便宜的短路：要么在进入 `setRawData` 前用 `getRawData()` 做一次预检（准入被全部拒绝 / 数据完全一致时 early-return，marker 本身也可一并比较），要么让 builder 在无变化时返回**原引用**并由 `setRawData` 检测 `data === rawData` 时跳过 `setItem`。后者可同时覆盖 `SimpleDbEntityLocalTokens` 里同样 `return base` 的拒绝分支。

[Claude session ↗](https://claude.ai/code/session_01GJHvviFfaN7xu1dXH27f57)

---

packages/kit/src/hooks/useAllNetwork.ts
weatherstar: Confirmed and fixed in 189ae1be86 — the analysis is right: with `clearRetainedResultOnAcceptedRun` the accepted run clears `lastPublishedResultRef` before its fan-out, so a superseded successful run hit the `hasQueuedRerun` branch with `lastPublished === undefined` and both results were dropped; the queued must-run then had nothing to restore on failure.

Fix follows the suggestion: `resolveAllNetworkPublishedResult` takes `retainSupersededResult` (passed as `clearRetainedResultOnAcceptedRun` from `useAllNetworkRequests`). In that mode the superseded branch records `{ result: completedResult, runSignature }` as `nextLastPublished` while still publishing nothing, so `resolveAllNetworkFailedRunRestore` in the queued run's `catch` has a real last-good snapshot. Legacy consumers (flag off) keep the previous behavior byte-for-byte.

Tests in `allNetworkRunResultUtils.test.ts`: superseded result is retained and restorable after a failed follow-up run; a newer superseded result wins over a retained one when retaining.

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
weatherstar: Fixed in 189ae1be86 with the first option (entity-local pre-check), keeping the in-builder comparison as the authoritative one.

- `updateAllNetworkAccountValue`: the builder body is extracted into `applyWrites(base) => { next, changed }`. It runs once on a `getRawData()` snapshot outside the mutex and returns early when nothing would change (all keys rejected, or values/markers identical — `canReplaceWhole` now also compares the replaced entry value-wise via `isSameAllNetworkEntry`), and again inside `setRawData` under the mutex. A stale pre-read can at most cost one redundant write, never a lost one.
- `updateAccountValue`: same pre-check with the shared `isNoopAgainst` predicate.
- I kept the `sameSnapshotMeta` requirement on purpose: an unchanged value with a NEWER marker must still persist the marker, otherwise a later response with a marker between the old and the new one would be admitted against the stale persisted marker and overwrite a fresher balance. Added a test asserting exactly that (value unchanged, marker newer → one write; identical/stale → no write).

I did not go with the base-class `data === rawData` skip: there are ~116 `setRawData` builders in kit-bg and some may mutate `rawData` in place and return it, which that change would silently stop persisting — too risky for a release branch without a full audit. `SimpleDbEntityLocalTokens`' `return base` branches only fire when a stale response is rejected (not on every refresh), so I left them as is; happy to add the same pre-check there if you'd prefer.

packages/kit/src/hooks/useAllNetwork.ts
```diff
@@ -912,6 +1031,25 @@ function useAllNetworkRequests<T>(params: {
...
+          // The accepted run cleared the retained result before fetching.
+          // Restore the last-good snapshot so a failed fan-out does not pin
+          // the consumer on `undefined` until the next successful must-run.
+          // The visible result is only restored while the owner is unchanged;
+          // the ref restore alone is safe because the skip path re-checks the
+          // run signature before serving it.
+          const { nextLastPublished, shouldRestoreResult } =
+            resolveAllNetworkFailedRunRestore({
+              previousPublished: previousPublishedResult,
+              ownerUnchanged: liveRunOwnerKeyRef.current === runnerOwnerKey,
```
sidmorizon, line 1105: # [🟡 P2] 失败恢复路径没有校验 `runSignature`，保留的上一 owner 结果会被回填到新 owner

## 问题

`resolveAllNetworkFailedRunRestore` 的 `ownerUnchanged` 只比较 `liveRunOwnerKeyRef.current === runnerOwnerKey`，即“本次失败的 run 自己的 owner 是否仍然是当前 owner”，并没有比较 `previousPublishedResult.runSignature` 与 `currentRunSignature`。而 `lastPublishedResultRef` 在账户/网络切换时不会被清空，本次增量又让被 supersede 的结果也写进这个 ref（`retainSupersededResult`），于是多了一条投喂路径：

```
owner X: run A 被排队的 must-run supersede
  → lastPublishedResultRef = { result: R_A, runSignature: sigX }
用户切到 owner Y（owner 变更重置 allNetworkDataInit，冗余门必然放行）
  → run B 接受，previousPublishedResult = { R_A, sigX }，清空 ref/result
  → B 的 fan-out 抛错 → ownerUnchanged = (Y === Y) = true
  → applyResultRef.current?.(R_A)   // 把 owner X 的快照 setResult 到 owner Y
```

skip 路径（`lastPublishedResultRef.current?.runSignature === currentRunSignature`）是做了这个签名校验的，恢复路径没有，两处语义不一致；注释里“the skip path re-checks the run signature”只覆盖了 ref 回写，覆盖不到 `applyResultRef` 这一步。

## 影响

当前不会造成数据污染：`TokenListBlock.updateAllNetworksTokenList` 会用 payload 内的 `ownerAccountId / ownerNetworkId` 再挡一次，foreign result 只让该 effect 空跑一轮。但 hook 对外暴露的 `result`，在 `切换账户后第一次刷新恰好失败` 这个窗口里确实持有上一个 owner 的完整快照，正确性只剩消费方一层兜底；一旦有新的消费方复用该 hook（或 DeFi/NFT 之后打开 `clearRetainedResultOnAcceptedRun`）而 payload 里没有 owner 字段，就会直接渲染他人账户的资产。

## 建议

把签名判断并入恢复决策，与 skip 路径共用同一条件：`ownerUnchanged: liveRunOwnerKeyRef.current === runnerOwnerKey && previousPublishedResult?.runSignature === currentRunSignature`；或给 `resolveAllNetworkFailedRunRestore` 增加 `currentRunSignature` 参数在纯函数内判定并补一条单测。ref 本身可以继续无条件回写（skip 路径已按签名过滤）。

[Claude session ↗](https://claude.ai/code/session_01GJHvviFfaN7xu1dXH27f57)

---

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
```diff
@@ -121,28 +204,51 @@ export class SimpleDbEntityAccountValue extends SimpleDbEntityBase<IAccountValue
...
+    // a NEWER marker is not: the marker must be persisted, otherwise a later
+    // stale response would be admitted against the old one.
+    const isNoopAgainst = (existing: IAccountValueEntry | undefined) =>
+      !canApplySnapshotMeta(assetSnapshotMeta, existing?.assetSnapshotMeta) ||
+      (existing?.value === value &&
+        existing?.currency === currency &&
+        sameSnapshotMeta(assetSnapshotMeta, existing?.assetSnapshotMeta));
+    // Pre-check outside the mutex so a repeated refresh does not re-serialize
+    // the whole entity (see updateAllNetworkAccountValue).
+    if (isNoopAgainst((await this.getRawData())?.byAddress[key])) {
```
sidmorizon: **Legacy balance refreshes crash**

Severity: `non-severe`

In the bg JS runtime, `updateAccountValue` indexes a missing legacy bucket. Refresh persistence fails before deferred migration completes.

Suggested fix:

```ts
    if (isNoopAgainst((await this.getRawData())?.byAddress?.[key])) {
```

packages/shared/src/utils/assetSnapshotFreshness.ts
```diff
@@ -0,0 +1,190 @@
...
+  }
+}
+
+export function createAssetSnapshotMeta(options?: {
+  serverDateMs?: number;
+  localSeq?: number;
+}): IAssetSnapshotMeta {
+  const localSeq = options?.localSeq ?? (nextLocalSeq += 1);
+  if (localSeq > nextLocalSeq) {
+    nextLocalSeq = localSeq;
```
sidmorizon, line 30: **Clock rollback drops first refresh**

Severity: `non-severe`

After a clock-rollback restart, `createAssetSnapshotMeta` mints below shared storage before either JS runtime observes its watermark. Every freshness gate rejects that refresh.

packages/kit/src/hooks/useAllNetwork.ts
weatherstar: 已修复：`resolveAllNetworkFailedRunRestore` 新增 `currentRunSignature` 参数，`shouldRestoreResult` 现在要求 `ownerUnchanged && previousPublished.runSignature === currentRunSignature`，与 skip 路径共用同一条签名判定。ref 回写保持无条件（skip / publish 路径已按签名过滤）。

补充单测 `keeps the ref but does not publish a snapshot minted for another run signature`，覆盖「切换 owner 后首个 run 失败」的场景：`ownerUnchanged: true` 但签名属于上一个 owner → 不再回填 `applyResultRef`，`lastPublishedResultRef` 仍保留。

packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAccountValue.ts
weatherstar: Fixed. This PR had regressed `?.byAddress?.[key]` to `?.byAddress[key]`, and the new `applyWrites` indexed `base.allByAddress[key]` the same way, so a record still in the legacy `data` / `all` shape (possible because `migrateFromAccountIdToAddressKey` runs deferred at bootstrap) threw `TypeError` on the first refresh in both writers.

Added `withAddressBuckets()` which normalizes `byAddress` / `allByAddress` to `{}` for both the pre-mutex check and the in-mutex builder of `updateAccountValue` and `updateAllNetworkAccountValue`. The legacy `data` / `all` fields are preserved so the deferred migration still consumes them.

Regression test: `writes into a pre-migration legacy record that lacks the address buckets` — fails on the previous code with `Cannot read properties of undefined (reading 'evm--1_0xalice')`.

packages/shared/src/utils/assetSnapshotFreshness.ts
weatherstar: Known and bounded, not changing it in this PR.

After a clock-rollback restart the first mint is indeed below the persisted watermark and that one refresh is rejected by every gate. The rejecting comparison itself lifts the counter (`observeLocalSeqWatermark` runs on both sides of every `compareAssetSnapshotMeta`), so the very next refresh (poller or pull-to-refresh) mints above the stored snapshot and wins. During that single cycle the UI keeps showing the persisted last-good snapshot rather than going empty. This is covered by the existing test `recovers the mint counter after observing a persisted future sequence (clock rollback)`.

Why not pre-observe before minting: the sequence has to be allocated synchronously before the first `await` in `fetchAccountTokens` so it captures request order, and `@onekeyhq/shared` has no access to persisted state to read the watermark ahead of that. A bootstrap-time warm-up from SimpleDb would not be airtight either (the first home refresh can run before deferred bootstrap tasks). Cost is one delayed refresh only after a backward clock jump across a restart, so leaving the self-heal as is.

weatherstar:
Status update after re-analysing the customer's iOS log (6.5.2, 2026-09-01):

The header/list mismatch in the screenshot is the `shouldHoldCurrentConfirmedBalance` gate pinning a persisted total until DeFi readiness arrives, and the cache-only DeFi hook that produces that readiness ran in only 2 of 8 cold starts that day. None of the freshness-marker work in this PR touches that gate.

To keep the 6.5.2 hotfix small, the evidence-backed parts of this PR (refresh queue race, analytics-after-commit, `deFiOverviewInitPlan`, selector/Others guards) were re-based onto a new branch together with a bounded header hold and a focus-independent DeFi readiness probe: #13124.

This PR keeps the full `IAssetSnapshotMeta` freshness system. Plan: leave it open, and once #13124 lands on `release/v6.5.2` and flows to `x`, rebase this PR onto `x` so it only carries the freshness delta and can soak there.

weatherstar:
Restacked: this branch was rewritten on top of `fix/wallet-home-header-hold-v6.5.2` (#13124) and the PR base was changed accordingly. The diff now contains only the `IAssetSnapshotMeta` freshness system (20 files); every other fix from the previous history lives in #13124, so the two PRs no longer duplicate code and this one can be rebased onto `x` without conflicts once the hotfix lands.

The previous head (7a9627501f, full history incl. review-round commits) is preserved at `backup/ios-wallet-home-asset-refresh-v6.5.2-full`.

originalix (commented):
@weatherstar

Reviewed commit `604901406225`.

- **P1 · Versioned single-network updates retain previous-network balances**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13103#issuecomment-5497302468)

PR #13111, opened 2026-09-01
fix: honor megafuel sponsorship for private send like normal send (OK-61643)

weatherstar:
[OK-61643](https://onekeyhq.atlassian.net/browse/OK-61643)

----------

## Summary

On BSC, private send never received megafuel gas sponsorship while a normal send from the same account did — even though the estimate response is identical in both flows (`megafuelEligible.sponsorable=true`, `payer="megafuel"`).

**Root cause**: a client-side gate in `TxFeeInfo` (`megafuelDisabledForPrivateSend = isPrivateSendTransfer`) unconditionally dropped `megafuelEligible` and restored `originalGasPrice` for private send, so the tx always signed user-paid. The gate dates from the private send launch (OK-55149, "user-paid by contract") and was kept in OK-59993 on the assumption that megafuel had no private-send contract on the backend. That premise no longer holds: the estimate runs on the actual private-send deposit `encodedTx` (with `scenario='privateSend'`) and the paymaster still reports it sponsorable, and the broadcast goes through the same `send-transaction` endpoint with only an extra `isPrivateSend` flag.

**Change**: private send now follows exactly the same sponsorship rules as a normal send.

- `TxFeeInfo`: remove the `megafuelDisabledForPrivateSend` gate and its strip/restore branches.
- `resolveSponsorPayerState`: remove the parameter and the megafuel-suppressed → gas-account-quote fallthrough logic that only existed to serve it (simplifies the function back to the shared rules).
- Tests: the private-send suppression block is dissolved; its non-duplicate cases (megafuel without an eligible quote, megafuel vs `gasAccountTemporarilyDisabled`, megafuel vs custom RPC) moved into the default-flow / global-suppression groups.

Unchanged suppressions: batch flows, external-wallet accounts, custom RPC, and the Gas Account scenario gates all still force user-paid as before.

## Test Plan

- [x] TDD: rewrote the payer-selection expectations first (5 failing on the old behavior), then removed the gate
- [x] 72 tests green across `gasAccountPayerSelection` / `gasAccountAnalytics` / `useSignatureConfirm` / `ServiceSend.broadcastDeadline`
- [x] `yarn agent:check --profile pr` pass
- [x] Manually verified on the web dev build: BSC private send now shows the BNB Chain sponsor badge with zero network fee, broadcast succeeds; normal send behavior unchanged

[OK-61643]: https://onekeyhq.atlassian.net/browse/OK-61643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx
+++ b/packages/kit/src/views/SignatureConfirm/components/TxFee/TxFeeInfo.tsx
@@ -160,7 +160,6 @@ function TxFeeInfo(props: IProps) {
     transferPayload,
     gasAccountScenario,
   } = props;
-  const isPrivateSendTransfer = transferPayload?.isPrivateSend === true;
   const gasAccountDisabledByScenario =
     isGasAccountDisabledScenario(gasAccountScenario);
   const intl = useIntl();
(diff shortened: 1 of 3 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix (approved):

PR #13119, opened 2026-09-02
feat: withdraw Perps USDC over Hyperliquid's CCTP rail(OK-61320)

Minnzen:
[OK-61320](https://onekeyhq.atlassian.net/browse/OK-61320)

----------

## Summary

- Perps USDC withdrawal follows the rail Hyperliquid serves from `info/usdcRouting` and sends `sendToEvmWithData` (Circle CCTP) instead of the deprecated `withdraw3` Arbitrum bridge, dropping the user-facing fee from **$1.00 to $0.20**.
- A CCTP withdrawal reaches the ledger as a `send` to the HyperEVM system address rather than a `withdraw`, so net deposits and the account history row are taught that shape.
- `sendToEvmWithData` requires an explicit source balance, so it mirrors `perpsComputedAccountValueAtom` — spot for unified / portfolio-margin accounts, perp clearinghouse otherwise.

## Intent & Context

Investigation started from a product question: our users pay $1 to withdraw while Hyperliquid's own app charges $0.20 — is the difference builder-vs-official, or something we need to change?

Neither. OneKey takes no cut (the `$1` in the modal was a hardcoded label; the amount is passed to Hyperliquid untouched). The $1 is the fee of Hyperliquid's **legacy Arbitrum bridge**, which their docs now mark deprecated: *"CCTP is the preferred method for minting USDC natively on the Hyperliquid L1… The legacy bridge is deprecated."* USDC moved to Circle CCTP, the legacy bridge holds under 10% of HyperCore's USDC supply, and the official app already switched. We had not.

`POST api.hyperliquid.xyz/info {"type":"usdcRouting"}` currently returns `{"depositRoute":"cctp","withdrawalRoute":"cctp"}`. That endpoint — rather than a constant in our bundle — is what this PR follows.

Circle's integration guide for this exact flow: https://developers.circle.com/cctp/howtos/withdraw-usdc-from-hypercore-to-evm

## Root Cause

Not a bug, but the same shape as one: the withdrawal rail was pinned in the bundle (`exchangeClient.withdraw3`) with no way to follow Hyperliquid's server-side migration, and no way to switch back without a release.

Three downstream assumptions were tied to the legacy rail and would have broken silently on the new one:

1. `usePerpPortfolioData` counted only `delta.type === 'withdraw'` for net deposits. CCTP withdrawals are `send`, so the "Net Deposits" stat would silently stop counting withdrawals the moment we switched rails. (Display only — PnL comes from Hyperliquid's own `portfolio` endpoint and never reads this value.)
2. `AccountRow` already had a `send` entry configured `isIncrease: true`. An outgoing withdrawal would have rendered as a green `+ $X` — money out shown as money in.
3. `withdraw3` let Hyperliquid pick the source balance; `sendToEvmWithData` does not. Unified-account users' withdrawable is the spot-side USDC balance, so a blind `sourceDex: ''` would have drawn from a balance the withdraw form never validated against.

## Design Decisions

**Follow `usdcRouting`, don't hardcode CCTP.** Hyperliquid kept a `bridge`/`cctp` switch, which implies they want the option to revert. Following it means their revert is our revert, with no release on our side. Rabby, the only other wallet found with a native Hyperliquid Perps module, hardcodes `withdraw3` and a literal `1` fee — it would break the day Hyperliquid pulls the bridge.

**Fall back to the confirmed rail, then to `bridge`.** A lookup failure teaches us nothing new, so it must not silently downgrade a user to a rail that is 5× more expensive. A rail already observed live wins; only with nothing observed does it fall back to `bridge`, which has worked in production all along.

**Constants over a live fee read.** The authoritative source is `calculateCrossChainWithdrawalFee(true, domain)` on CoreDepositWallet (HyperEVM `0x6B9E77…0A24`) — verified live: Ethereum 1.20, Solana 0.50, Arbitrum/Base/others 0.20. It is noted in the constants, but the value is display-only, only Arbitrum is reachable today, and 0.20 has held since launch. A network dependency inside a funds modal is not worth buying accuracy we do not yet need; it earns its place when multiple destinations make the per-chain values differ.

**Minimum withdrawal stays at 2 USDC.** The official app uses 2 × fee (0.40). Lowering it opens a failure class: if Circle raises the forwarding fee, a withdrawal under it reverts on the HyperEVM leg. 2 exceeds any plausible fee, and keeping it means zero change from today's behavior.

**`sourceDex` mirrors the atom the form validates against.** Deriving it independently would let the two drift; mirroring `perpsComputedAccountValueAtom` branch-for-branch makes "the balance we spend" and "the balance we validated" the same decision. Rabby reaches the same conclusion (`sourceDex: isUnified ? 'spot' : ''`).

**No optimistic pending-withdrawal row.** There is none today, so omitting it is not a regression. Adding one would need the SDK-internal nonce captured through a custom `nonceManager`, a new persisted store, and handling a write-after-ledger-event race — three new failure points for a new feature, inside a change that ships as a hot update.

## Changes Detail

| File | Purpose |
|---|---|
| `ServiceHyperLiquid/usdcWithdrawRoute.ts` (new) | Resolves the active rail from `info/usdcRouting`; 5-min cache, single in-flight request, prefers a confirmed rail over the fallback. Reports failures through `requestLoggedHyperLiquidTransport`, the same path as every other Hyperliquid request in this service. Uses the transport directly because `usdcRouting` is absent from the pinned SDK's info client — deliberately not bumping `@nktkas/hyperliquid` on a release branch. |
| `ServiceHyperLiquid/usdcWithdrawRoute.test.ts` (new) | 8 tests: following the served rail, both fallback paths, caching, in-flight sharing, failures not cached, confirmed rail retained across a failed lookup. |
| `ServiceHyperliquidExchange.ts` | `withdraw()` branches on the rail; adds `_resolveWithdrawSourceDex()`. Rail and address resolution sit inside the `try` so a failure there is logged like any other withdrawal failure. |
| `usePerpPortfolioData.ts` | The "Net Deposits" stat counts the outbound `send` to the HyperEVM system address, so withdrawals keep counting after the rail switch. Outbound only: inbound EVM → Core credits arrive as `spotTransfer`, and counting them would double up against the existing `accountClassTransfer` branch. |
| `AccountRow.tsx` | A `send` to another party renders as outgoing (`sendOut`); transfers between a user's own balances keep today's rendering, since neither direction describes them. |
| `DepositWithdrawModal.tsx` | Fee row reflects the active rail. Display only — amount validation still uses `MIN_WITHDRAW_AMOUNT`, which covers both rails, and the service re-resolves the rail at submit. |
| `perp.constants.ts` | CCTP domain / gas limit / hook data, the system address, and a per-domain fee map. |
| `logger/.../hyperliquid.ts` | The existing withdrawal analytics event carries `route` and `sourceDex` — the two new branch points, and the fields needed to diagnose a production failure. |

## Verified on mainnet

One real 4 USDC withdrawal from a unified account:

| | |
|---|---|
| Ledger delta | `{"type":"send","destination":"0x2000…0000","sourceDex":"spot","amount":"4.0","fee":"0.0","nativeTokenFee":"0.00002"}` |
| Arbitrum arrival | 3.80 USDC minted from `0x0` (CCTP burn-and-mint), block 500709023 |
| Fee | 4.00 − 3.80 = **$0.20**, exactly as quoted in the modal |
| Latency | ~10s (legacy bridge is 3–5 min) |
| `sourceDex` | Sent `spot`, echoed back `spot` — Hyperliquid honored it |

## Risk Assessment

- **Risk Level**: Medium — funds path, though the blast radius is confined to withdrawal plus two display surfaces.
- **Affected Platforms**: Extension / Mobile / Desktop / Web (JS only — no native code, no `LOCAL_DB_VERSION` bump, no DB schema change, nothing newly persisted; safe as a hot update and safe to roll back).
- **Risk Areas**:
  - `sourceDex` is the funds-critical field. Verified live for `unifiedAccount`; `disabled`, `dexAbstraction` and `portfolioMargin` still want a QA pass. A wrong value fails the withdrawal server-side rather than overspending — Hyperliquid enforces per-dex balances.
  - Net deposits is the widest surface: it changes the "Net Deposits" stat for anyone with `send` deltas to the system address, including HyperEVM users who never withdraw. Display-only — no funds impact, and PnL does not read it.
  - Two independent kill switches, neither needing a release: Hyperliquid flipping `usdcRouting` back to `bridge` returns every client to `withdraw3`, and a failed lookup cannot downgrade a user below a rail already confirmed live.

## Out of scope

- **Multi-chain destinations.** The rail supports Ethereum / Base / Solana; a destination selector plus Solana base58 address entry is a separate change. Fee constants are keyed by CCTP domain so it drops in later.
- **Sub-dex withdrawal.** Unchanged — main-dex USDC only, matching today's behavior and Rabby's.
- **Showing the $0.20 in account history.** The ledger records only the HyperCore-side gas (~$0.001, or 0 when paid in HYPE); the forwarding fee is charged after funds leave HyperCore. Hyperliquid's app papers over this with a hardcoded constant, but a CCTP withdrawal and a plain Core → HyperEVM transfer are indistinguishable in the ledger, so that override would stamp a fee on transfers that never paid one. The modal quotes $0.20 at the moment the user decides, which is the disclosure that matters.
- **Deposits.** `usdcRouting.depositRoute` is also `cctp` while our Arbitrum deposit still targets the legacy bridge address. Same sunset, tracked separately.

## Test plan

- [x] Real mainnet withdrawal, unified account: 4 USDC out, 3.80 USDC received on Arbitrum, $0.20 fee, ~10s
- [x] 8 rail-resolution tests; `ServiceHyperLiquid` + Perps atoms + portfolio utils suites (15 suites, 141 tests)
- [x] Full-repo `tsc`, eslint, prettier, cspell
- [ ] Withdraw from a `disabled` account — expect `sourceDex: ""`, funds leave the perp clearinghouse
- [ ] Withdraw from a `dexAbstraction` account — expect `sourceDex: ""`
- [ ] Withdraw from a `portfolioMargin` account — expect `sourceDex: "spot"`
- [ ] Account history shows the withdrawal as outgoing (red `−`), not incoming
- [ ] Portfolio net deposits / PnL show no phantom loss after withdrawing
- [ ] Fee row reads `$0.20` before submitting
- [ ] Withdrawal below 2 USDC is still rejected by the form

[OK-61320]: https://onekeyhq.atlassian.net/browse/OK-61320?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
@@ -31,6 +31,7 @@ import type {
 } from '@onekeyhq/shared/src/logger/scopes/perp/scenes/hyperliquid';
 import accountUtils from '@onekeyhq/shared/src/utils/accountUtils';
 import { convertHyperLiquidResponse } from '@onekeyhq/shared/src/utils/hyperLiquidErrorResolver';
+import { isUnifiedPortfolioMode } from '@onekeyhq/shared/src/utils/hyperliquidPortfolioUtils';
 import {
   assertValidScaleOrderLegs,
   buildScaleOrderLegs,
@@ -46,7 +47,12 @@ import {
   mapTriggerOrderType,
   parseSignatureToRSV,
 } from '@onekeyhq/shared/src/utils/perpsUtils';
-import { SPOT_ASSET_ID_OFFSET } from '@onekeyhq/shared/types/hyperliquid/perp.constants';
+import {
+  CCTP_DOMAIN_ARBITRUM,
+  CCTP_WITHDRAW_GAS_LIMIT,
+  CCTP_WITHDRAW_HOOK_DATA,
+  SPOT_ASSET_ID_OFFSET,
+} from '@onekeyhq/shared/types/hyperliquid/perp.constants';
 import type {
   IApiErrorResponse,
   IApiRequestResult,
@@ -85,11 +91,14 @@ import type { IHyperLiquidSignatureRSV } from '@onekeyhq/shared/types/hyperliqui
 import { ERookieTaskType } from '@onekeyhq/shared/types/rookieGuide';
 
 import {
+  perpsAbstractionModeAtom,
   perpsActiveAccountAtom,
   perpsActiveAccountStatusAtom,
 } from '../../states/jotai/atoms';
 import ServiceBase from '../ServiceBase';
 
+import { getUsdcWithdrawRoute } from './usdcWithdrawRoute';
+import type { IUsdcWithdrawRoute } from './usdcWithdrawRoute';
 import { createLoggedHyperLiquidClient } from './utils/logHyperLiquidApiFailure';
 
 import type {
@@ -1734,6 +1743,32 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
     );
   }
 
+  // The withdraw form needs the rail to show the right fee; the result is cached
+  // in the route module, so repeat opens do not re-request it.
+  @backgroundMethod()
+  async getUsdcWithdrawRoute(): Promise<IUsdcWithdrawRoute> {
+    return getUsdcWithdrawRoute();
+  }
+
+  // `sendToEvmWithData` requires an explicit source balance, where `withdraw3`
+  // let Hyperliquid pick one. Mirror perpsComputedAccountValueAtom: unified and
+  // portfolio-margin accounts report the spot-side USDC balance as withdrawable,
+  // every other mode reports the perp clearinghouse. Sourcing from a balance the
+  // withdraw form never validated against would let the action exceed what the
+  // user was shown as available.
+  private async _resolveWithdrawSourceDex(
+    userAddress: string | undefined,
+  ): Promise<'' | 'spot'> {
+    const modeData = await perpsAbstractionModeAtom.get();
+    const isModeForThisAccount =
+      Boolean(userAddress) &&
+      modeData?.accountAddress?.toLowerCase() === userAddress?.toLowerCase();
+    if (!isModeForThisAccount) {
+      return '';
+    }
+    return isUnifiedPortfolioMode(modeData?.mode) ? 'spot' : '';
+  }
+
   @backgroundMethod()
   async withdraw(params: IWithdrawParams): Promise<void> {
     await this.checkAccountCanTrade();
@@ -1749,17 +1784,46 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
       }),
     );
     const context = await this._buildLogContext();
+    // Resolved inside the try so a failure here is reported like any other
+    // withdrawal failure instead of escaping without a log.
+    let route: IUsdcWithdrawRoute | undefined;
+    // Which balance the action drew from is the field most worth having when a
+    // withdrawal misbehaves, and the one a signature prompt may not surface.
+    let sourceDex: '' | 'spot' | undefined;
     try {
-      await convertHyperLiquidResponse(() => exchangeClient.withdraw3(params));
+      const [resolvedRoute, userAddress] = await Promise.all([
+        getUsdcWithdrawRoute(),
+        wallet.getAddress(),
+      ]);
+      route = resolvedRoute;
+      await convertHyperLiquidResponse(async () => {
+        if (route === 'cctp') {
+          sourceDex = await this._resolveWithdrawSourceDex(userAddress);
+          return exchangeClient.sendToEvmWithData({
+            token: 'USDC',
+            amount: params.amount,
+            sourceDex,
+            destinationRecipient: params.destination,
+            addressEncoding: 'hex',
+            destinationChainId: CCTP_DOMAIN_ARBITRUM,
+            gasLimit: CCTP_WITHDRAW_GAS_LIMIT,
+            data: CCTP_WITHDRAW_HOOK_DATA,
+          });
+        }
+        return exchangeClient.withdraw3({
+          amount: params.amount,
+          destination: params.destination,
+        });
+      });
       defaultLogger.perp.hyperliquid.withdraw({
         ...context,
-        request: params,
+        request: { ...params, route, sourceDex },
         response: { success: true },
       });
     } catch (error) {
       defaultLogger.perp.hyperliquid.withdraw({
         ...context,
-        request: params,
+        request: { ...params, route, sourceDex },
         response: extractHyperLiquidErrorResponse<IApiErrorResponse>(error),
         error: serializeHyperLiquidError(error),
       });
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
@@ -1756,8 +1756,6 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
     );
   }
 
-  // The withdraw form needs the rail to show the right fee; the result is cached
-  // in the route module, so repeat opens do not re-request it.
   @backgroundMethod()
   async getUsdcWithdrawRoute(): Promise<IUsdcWithdrawRoute> {
     return getUsdcWithdrawRoute();
@@ -1786,12 +1784,8 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
     return getUsdcWithdrawFee(params.destinationId, this._callHyperEvmRpc);
   }
 
-  // `sendToEvmWithData` requires an explicit source balance, where `withdraw3`
-  // let Hyperliquid pick one. Mirror perpsComputedAccountValueAtom: unified and
-  // portfolio-margin accounts report the spot-side USDC balance as withdrawable,
-  // every other mode reports the perp clearinghouse. Sourcing from a balance the
-  // withdraw form never validated against would let the action exceed what the
-  // user was shown as available.
+  // Mirrors perpsComputedAccountValueAtom so the action spends the same balance
+  // the withdraw form validated the amount against.
   private async _resolveWithdrawSourceDex(
     userAddress: string | undefined,
   ): Promise<'' | 'spot'> {
@@ -1823,11 +1817,8 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
       }),
     );
     const context = await this._buildLogContext();
-    // Resolved inside the try so a failure here is reported like any other
-    // withdrawal failure instead of escaping without a log.
+    // Declared outside the try so a failure before the action still logs them.
     let route: IUsdcWithdrawRoute | undefined;
-    // Which balance the action drew from is the field most worth having when a
-    // withdrawal misbehaves, and the one a signature prompt may not surface.
     let sourceDex: '' | 'spot' | undefined;
     try {
       const [resolvedRoute, userAddress] = await Promise.all([
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
@@ -7,6 +7,7 @@ import {
   backgroundClass,
   backgroundMethod,
 } from '@onekeyhq/shared/src/background/backgroundDecorators';
+import { getNetworkIdsMap } from '@onekeyhq/shared/src/config/networkIds';
 import {
   DISABLE_PERPS_WALLET_BIND,
   type EHyperLiquidAgentName,
@@ -48,10 +49,12 @@ import {
   parseSignatureToRSV,
 } from '@onekeyhq/shared/src/utils/perpsUtils';
 import {
-  CCTP_DOMAIN_ARBITRUM,
-  CCTP_WITHDRAW_GAS_LIMIT,
-  CCTP_WITHDRAW_HOOK_DATA,
+  HYPEREVM_SYSTEM_ADDRESS,
   SPOT_ASSET_ID_OFFSET,
+} from '@onekeyhq/shared/types/hyperliquid/perp.constants';
+import type {
+  IUsdcWithdrawDestinationId,
+  IUsdcWithdrawFeeQuote,
 } from '@onekeyhq/shared/types/hyperliquid/perp.constants';
 import type {
   IApiErrorResponse,
@@ -97,14 +100,24 @@ import {
 } from '../../states/jotai/atoms';
 import ServiceBase from '../ServiceBase';
 
-import { getUsdcWithdrawRoute } from './usdcWithdrawRoute';
-import type { IUsdcWithdrawRoute } from './usdcWithdrawRoute';
+import {
+  buildCctpWithdrawDestination,
+  getLiveUsdcWithdrawFee,
+  getUsdcWithdrawFee,
+  requireUsdcWithdrawDestination,
+} from './cctpWithdraw';
+import {
+  getLiveUsdcWithdrawRoute,
+  getUsdcWithdrawRoute,
+} from './usdcWithdrawRoute';
 import { createLoggedHyperLiquidClient } from './utils/logHyperLiquidApiFailure';
 
+import type { IHyperEvmRpcCall } from './cctpWithdraw';
 import type {
   WalletHyperliquidOnekey,
   WalletHyperliquidProxy,
 } from './ServiceHyperliquidWallet';
+import type { IUsdcWithdrawRoute } from './usdcWithdrawRoute';
 import type { IBackgroundApi } from '../../apis/IBackgroundApi';
 
 interface IOrderLogOptions {
@@ -1750,6 +1763,29 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
     return getUsdcWithdrawRoute();
   }
 
+  private _callHyperEvmRpc: IHyperEvmRpcCall = async (method, params) => {
+    const [result] = await this.backgroundApi.serviceDApp.proxyRPCCall<unknown>(
+      {
+        networkId: getNetworkIdsMap().hyperevm,
+        request: {
+          jsonrpc: '2.0',
+          id: 0,
+          method,
+          params,
+        },
+        origin: 'onekey://perps',
+      },
+    );
+    return result;
+  };
+
+  @backgroundMethod()
+  async getUsdcWithdrawFee(params: {
+    destinationId: IUsdcWithdrawDestinationId;
+  }): Promise<IUsdcWithdrawFeeQuote> {
+    return getUsdcWithdrawFee(params.destinationId, this._callHyperEvmRpc);
+  }
+
   // `sendToEvmWithData` requires an explicit source balance, where `withdraw3`
   // let Hyperliquid pick one. Mirror perpsComputedAccountValueAtom: unified and
   // portfolio-margin accounts report the spot-side USDC balance as withdrawable,
@@ -1772,6 +1808,9 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
   @backgroundMethod()
   async withdraw(params: IWithdrawParams): Promise<void> {
     await this.checkAccountCanTrade();
+    const destinationConfig = requireUsdcWithdrawDestination(
+      params.destinationId,
+    );
     const wallet =
       await this.backgroundApi.serviceHyperliquidWallet.getOnekeyWallet({
         userAccountId: params.userAccountId,
@@ -1780,7 +1819,7 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
       new ExchangeClient({
         transport: new HttpTransport(),
         wallet,
-        signatureChainId: PERPS_EVM_CHAIN_ID_HEX,
+        signatureChainId: destinationConfig.signatureChainId,
       }),
     );
     const context = await this._buildLogContext();
@@ -1792,27 +1831,86 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
     let sourceDex: '' | 'spot' | undefined;
     try {
       const [resolvedRoute, userAddress] = await Promise.all([
-        getUsdcWithdrawRoute(),
+        destinationConfig.transferType === 'cctp'
+          ? getLiveUsdcWithdrawRoute()
+          : Promise.resolve(undefined),
         wallet.getAddress(),
       ]);
       route = resolvedRoute;
+      if (
+        destinationConfig.transferType === 'cctp' &&
+        (!params.expectedRoute || route !== params.expectedRoute)
+      ) {
+        throw new OneKeyLocalError(
+          'Withdrawal route changed. Review the updated fee and try again.',
+        );
+      }
+      if (
+        destinationConfig.transferType === 'cctp' &&
+        route === 'bridge' &&
+        !destinationConfig.supportsLegacyBridge
+      ) {
+        throw new OneKeyLocalError(
+          `${destinationConfig.name} withdrawals require the CCTP route`,
+        );
+      }
+      if (destinationConfig.transferType === 'cctp' && route === 'cctp') {
+        const liveFeeQuote = await getLiveUsdcWithdrawFee(
+          params.destinationId,
+          this._callHyperEvmRpc,
+        );
+        const cctpFee = liveFeeQuote.components.find(
+          (component) => component.kind === 'cctpForwarding',
+        );
+        if (!cctpFee?.amount) {
+          throw new OneKeyLocalError('Unable to quote CCTP withdrawal fee');
+        }
+        if (
+          !params.expectedCctpFee ||
+          !new BigNumber(params.expectedCctpFee).eq(cctpFee.amount)
+        ) {
+          throw new OneKeyLocalError(
+            'Withdrawal fee changed. Review the updated fee and try again.',
+          );
+        }
+        if (new BigNumber(params.amount).lte(cctpFee.amount)) {
+          throw new OneKeyLocalError(
+            'Withdrawal amount must exceed the CCTP fee',
+          );
+        }
+      }
       await convertHyperLiquidResponse(async () => {
+        if (destinationConfig.transferType === 'hyperEvm') {
+          sourceDex = await this._resolveWithdrawSourceDex(userAddress);
+          return exchangeClient.sendAsset({
+            destination: HYPEREVM_SYSTEM_ADDRESS,
+            sourceDex,
+            destinationDex: 'spot',
+            token: 'USDC',
+            amount: params.amount,
+            fromSubAccount: '',
+          });
+        }
         if (route === 'cctp') {
           sourceDex = await this._resolveWithdrawSourceDex(userAddress);
+          const destination = buildCctpWithdrawDestination({
+            destinationId: params.destinationId,
+            ownerAddress: userAddress,
+          });
           return exchangeClient.sendToEvmWithData({
             token: 'USDC',
             amount: params.amount,
             sourceDex,
-            destinationRecipient: params.destination,
-            addressEncoding: 'hex',
-            destinationChainId: CCTP_DOMAIN_ARBITRUM,
-            gasLimit: CCTP_WITHDRAW_GAS_LIMIT,
-            data: CCTP_WITHDRAW_HOOK_DATA,
+            destinationRecipient: destination.destinationRecipient,
+            addressEncoding: destination.addressEncoding,
+            destinationChainId: destination.destinationChainId,
+            gasLimit: destination.gasLimit,
+            data: destination.data,
           });
         }
         return exchangeClient.withdraw3({
           amount: params.amount,
-          destination: params.destination,
+          destination: userAddress,
         });
       });
       defaultLogger.perp.hyperliquid.withdraw({
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
@@ -174,8 +174,8 @@ export async function getUsdcWithdrawFee(
   };
 }
 
-// Submission must not rely on a cached or static CCTP fallback fee. The contract
-// can update per-domain USDC fees, while direct HyperEVM sends have no CCTP fee.
+// The contract can update per-domain fees, so submission must not quote a cached
+// or fallback value.
 export async function getLiveUsdcWithdrawFee(
   destinationId: IUsdcWithdrawDestinationId,
   rpcCall: IHyperEvmRpcCall,
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
@@ -1,0 +1,242 @@
+import BigNumber from 'bignumber.js';
+
+import { OneKeyLocalError } from '@onekeyhq/shared/src/errors';
+import {
+  CCTP_WITHDRAW_GAS_LIMIT,
+  CCTP_WITHDRAW_HOOK_DATA,
+  USDC_WITHDRAW_GAS_RESERVE,
+  getUsdcWithdrawDestination,
+} from '@onekeyhq/shared/types/hyperliquid/perp.constants';
+import type {
+  ICctpWithdrawDestinationConfig,
+  IUsdcWithdrawDestinationConfig,
+  IUsdcWithdrawDestinationId,
+  IUsdcWithdrawFeeComponent,
+  IUsdcWithdrawFeeQuote,
+} from '@onekeyhq/shared/types/hyperliquid/perp.constants';
+import type { IHex } from '@onekeyhq/shared/types/hyperliquid/sdk';
+
+const CORE_DEPOSIT_WALLET_ADDRESS =
+  '0x6B9E773128f453f5c2C60935Ee2DE2CBc5390A24';
+const CCTP_WITHDRAW_FEE_SELECTOR = 'e26f7d23';
+const CCTP_WITHDRAW_FEE_CACHE_TTL_MS = 5 * 60 * 1000;
+const USDC_DECIMALS = 6;
+
+export type IHyperEvmRpcCall = (
+  method: string,
+  params: unknown[],
+) => Promise<unknown>;
+
+interface ICctpWithdrawFeeCacheEntry {
+  fee: number;
+  fetchedAt: number;
+}
+
+export interface ICctpWithdrawDestinationActionFields {
+  config: ICctpWithdrawDestinationConfig;
+  destinationRecipient: string;
+  addressEncoding: 'hex';
+  destinationChainId: number;
+  gasLimit: number;
+  data: IHex;
+}
+
+const cachedFees = new Map<
+  IUsdcWithdrawDestinationId,
+  ICctpWithdrawFeeCacheEntry
+>();
+const inFlightFeeRequests = new Map<
+  IUsdcWithdrawDestinationId,
+  Promise<IUsdcWithdrawFeeComponent>
+>();
+
+export function requireUsdcWithdrawDestination(
+  destinationId: string,
+): IUsdcWithdrawDestinationConfig {
+  const config = getUsdcWithdrawDestination(destinationId);
+  if (!config) {
+    throw new OneKeyLocalError(
+      `Unsupported USDC withdrawal destination: ${destinationId}`,
+    );
+  }
+  return config;
+}
+
+function encodeWithdrawFeeCall(destinationDomain: number): IHex {
+  const useForwarding = '1'.padStart(64, '0');
+  const domain = destinationDomain.toString(16).padStart(64, '0');
+  return `0x${CCTP_WITHDRAW_FEE_SELECTOR}${useForwarding}${domain}` as IHex;
+}
+
+function parseWithdrawFee(result: unknown): number {
+  if (typeof result !== 'string' || !/^0x[0-9a-fA-F]+$/.test(result)) {
+    throw new OneKeyLocalError('Invalid CCTP withdrawal fee response');
+  }
+  const fee = new BigNumber(result.slice(2), 16)
+    .shiftedBy(-USDC_DECIMALS)
+    .toNumber();
+  if (!Number.isFinite(fee) || fee < 0) {
+    throw new OneKeyLocalError('Invalid CCTP withdrawal fee value');
+  }
+  return fee;
+}
+
+async function fetchCctpWithdrawFee(
+  config: ICctpWithdrawDestinationConfig,
+  rpcCall: IHyperEvmRpcCall,
+): Promise<number> {
+  const result = await rpcCall('eth_call', [
+    {
+      to: CORE_DEPOSIT_WALLET_ADDRESS,
+      data: encodeWithdrawFeeCall(config.domain),
+    },
+    'latest',
+  ]);
+  return parseWithdrawFee(result);
+}
+
+async function getCctpWithdrawFee(
+  destinationId: IUsdcWithdrawDestinationId,
+  rpcCall: IHyperEvmRpcCall,
+): Promise<IUsdcWithdrawFeeComponent> {
+  const config = requireUsdcWithdrawDestination(destinationId);
+  if (config.transferType !== 'cctp') {
+    throw new OneKeyLocalError(
+      `Destination does not have a CCTP fee: ${destinationId}`,
+    );
+  }
+  const cached = cachedFees.get(destinationId);
+  const now = Date.now();
+  if (cached && now - cached.fetchedAt < CCTP_WITHDRAW_FEE_CACHE_TTL_MS) {
+    return {
+      kind: 'cctpForwarding',
+      amount: cached.fee.toString(),
+      token: 'USDC',
+      debitedFrom: 'withdrawAmount',
+      isEstimate: false,
+    };
+  }
+
+  const inFlight = inFlightFeeRequests.get(destinationId);
+  if (inFlight) {
+    return inFlight;
+  }
+
+  const request = fetchCctpWithdrawFee(config, rpcCall)
+    .then((fee) => {
+      cachedFees.set(destinationId, { fee, fetchedAt: Date.now() });
+      return {
+        kind: 'cctpForwarding' as const,
+        amount: fee.toString(),
+        token: 'USDC' as const,
+        debitedFrom: 'withdrawAmount' as const,
+        isEstimate: false,
+      };
+    })
+    .catch(() => ({
+      kind: 'cctpForwarding' as const,
+      amount: config.fallbackFee.toString(),
+      token: 'USDC' as const,
+      debitedFrom: 'withdrawAmount' as const,
+      isEstimate: true,
+    }))
+    .finally(() => {
+      inFlightFeeRequests.delete(destinationId);
+    });
+  inFlightFeeRequests.set(destinationId, request);
+  return request;
+}
+
+export async function getUsdcWithdrawFee(
+  destinationId: IUsdcWithdrawDestinationId,
+  rpcCall: IHyperEvmRpcCall,
+): Promise<IUsdcWithdrawFeeQuote> {
+  const config = requireUsdcWithdrawDestination(destinationId);
+  if (config.transferType === 'hyperEvm') {
+    return {
+      components: [
+        {
+          kind: 'hyperEvmGas',
+          amount: USDC_WITHDRAW_GAS_RESERVE.toString(),
+          token: 'USDC',
+          debitedFrom: 'spotHypeOrSourceUsdc',
+          isEstimate: true,
+          displayAsLessThan: true,
+        },
+      ],
+      quotedAt: Date.now(),
+    };
+  }
+  const cctpFee = await getCctpWithdrawFee(destinationId, rpcCall);
+  return {
+    components: [cctpFee],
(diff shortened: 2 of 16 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1734,6 +1743,32 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+    userAddress: string | undefined,
+  ): Promise<'' | 'spot'> {
+    const modeData = await perpsAbstractionModeAtom.get();
+    const isModeForThisAccount =
+      Boolean(userAddress) &&
+      modeData?.accountAddress?.toLowerCase() === userAddress?.toLowerCase();
+    if (!isModeForThisAccount) {
+      return '';
+    }
+    return isUnifiedPortfolioMode(modeData?.mode) ? 'spot' : '';
```
sidmorizon, line 1799: **Unknown account mode blocks unified withdrawals**

Severity: `non-severe`

When `perpsAbstractionModeAtom` is unavailable or account selection changes, `_resolveWithdrawSourceDex` chooses the perp balance for unified accounts. Their withdrawals then fail despite a validated spot balance.

packages/kit/src/views/Perp/components/Portfolio/usePerpPortfolioData.ts
```diff
@@ -127,6 +128,22 @@ export function usePerpPortfolioData(
...
           return sum.minus(delta.usdc);
         }
+        // CCTP withdrawals leave HyperCore as a `send` to the HyperEVM system
+        // address, not as a `withdraw`. Without this they never reduce net
+        // deposits, so the balance drop reads as a trading loss. A plain
+        // Core -> HyperEVM transfer takes the same shape and is equally an
+        // outflow. Both legs are counted so a round trip nets to zero rather
+        // than showing phantom profit; the inbound leg is matched on the system
+        // address as the sender, mirroring the outbound shape.
+        if (delta.type === 'send' && delta.token === 'USDC' && delta.amount) {
```
sidmorizon: # 🟠 P1 send 分支忽略 sourceDex，与同一 reducer 的 accountClassTransfer 口径冲突

## 问题

这个 reducer 的记账口径是 **perp 账户**：`deposit` / `withdraw` 是 Arbitrum bridge 与 perp 之间的进出，而紧跟其后的 `accountClassTransfer` 分支用 `delta.toPerp ? plus : minus` 把 spot 明确当作「账户之外」。新增的 `send` 分支却只按 `HYPEREVM_SYSTEM_ADDRESS` 匹配，完全忽略了 `send` delta 自带的 `sourceDex` / `destinationDex` 字段（SDK 中定义为 `""` = 默认 USDC perp DEX，`"spot"` = 现货）。

于是任何**从 spot 发出**的 Core → HyperEVM 转账都会被再扣一次。典型序列：bridge 存入 `100`（`deposit`，`+100`）→ perp 转 spot（`accountClassTransfer` `toPerp:false`，`-100`）→ 从 spot 走 CCTP 提现 `100`（`send` `sourceDex:'spot'`、`destination` 为系统地址，本 PR 再 `-100`）。真实净入金是 `0`，页面显示 `-100`。入金方向同理：`delta.user === systemAddress` 且 `destinationDex === 'spot'` 时会凭空 `+`。

注意这不是理论路径：本 PR 的 `_resolveWithdrawSourceDex()` 在 `unifiedAccount` / `portfolioMargin` 下正是传 `sourceDex: 'spot'`，用户在 HL 官方 UI 做的 spot → HyperEVM 转账也是同一形态。改动前该 delta 记 `0`（净入金恰好正确），改动后反而变错。

## 影响

`PerpPortfolioContent` 的 Net Deposits 指标对上述用户会偏离真实值，偏差等于所有 spot 侧系统地址收发的总额；在「先 perp→spot 再提现」这一常见路径上误差是提现额的 `1` 倍（示例中 `-100` vs `0`），且会随每次往返累积。这是展示给用户的资金数字，错值会被直接理解为亏损或幻影收益。

## 建议

让 `send` 分支和 reducer 其余部分使用同一记账口径。若维持现有的 perp 口径，按 dex 字段收口：

```ts
if (delta.type === 'send' && delta.token === 'USDC' && delta.amount) {
  const systemAddress = HYPEREVM_SYSTEM_ADDRESS.toLowerCase();
  if (
    delta.destination?.toLowerCase() === systemAddress &&
    delta.sourceDex === ''
  ) {
    return sum.minus(delta.amount);
  }
  if (
    delta.user?.toLowerCase() === systemAddress &&
    delta.destinationDex === ''
  ) {
    return sum.plus(delta.amount);
  }
}
```

但这样 unified / portfolioMargin 下的 CCTP 提现（`sourceDex: 'spot'`）又会重新漏记。所以更根本的做法是先明确 Net Deposits 到底统计 perp 还是 perp+spot：若确定为 perp+spot，则应同步去掉 `accountClassTransfer` 分支（perp↔spot 属内部划转，不应计入净入金），两处口径统一后上述双扣与幻影加减都会消失。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

originalix:

Review of commit `799052abade0` did not finish successfully.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -1957,6 +2078,147 @@ function DepositWithdrawContent({
...
+  );
+
+  // The selector always needs the active external withdrawal rail, including
+  // when HyperEVM is selected. The background service serves its cached value.
+  useEffect(() => {
+    if (selectedAction !== 'withdraw') {
+      return;
+    }
+    let cancelled = false;
+    setWithdrawRoute(undefined);
```
sidmorizon, line 2179: # 🟠 P1 提现 rail 在弹窗内永不刷新，错误文案要求的「重试」无法成功

## 问题

`withdrawRoute` 只在这个 effect 里解析一次，依赖数组是 `[selectedAction]`，之后没有任何地方会再刷新它；新增的 `WITHDRAW_QUOTE_REFRESH_INTERVAL_MS` 定时器只调用 `getUsdcWithdrawFee`，不会重新解析 rail。而 UI 侧走的 `getUsdcWithdrawRoute()` 读的是 BG 的 5 分钟缓存，并且在请求失败时被 `.catch()` 吞掉、回落到 `lastResolvedRoute ?? 'bridge'`；提交时 BG 走的却是 `getLiveUsdcWithdrawRoute()`（无缓存、无回落），一旦 `route !== params.expectedRoute` 就抛 `Withdrawal route changed. Review the updated fee and try again.`。

## 影响

只要 `usdcRouting` 出现一次瞬时失败且 BG 缓存是冷的，UI 就会拿到回落值 `bridge`，而线上实际服务的是 `cctp`：用户每次点提现都命中同一个错误，`expectedRoute` 永远停在那个陈旧值，重试多少次都不可能成功，必须关闭并重开弹窗才能恢复。Hyperliquid 在弹窗打开期间切换 rail、或弹窗在切换后的 `5` 分钟缓存窗口内打开，也会进入同样的死循环。错误文案让用户「review the updated fee and try again」，但界面上没有任何值会随重试而更新。

## 建议

把 route 解析并入已有的 30 秒刷新回路，或在捕获到 route/fee 不匹配的错误后立即重新调用 `getUsdcWithdrawRoute()` 并清空 `withdrawFeeState`，让陈旧的 `withdrawRoute` 能自愈。fee 一侧已经具备这个回路（`getLiveUsdcWithdrawFee` 回写 `cachedFees` + UI 轮询，30 秒内自动收敛），route 一侧缺失对称处理。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -893,14 +981,14 @@ function DepositWithdrawContent({
...
-      amountBN.lt(MIN_WITHDRAW_AMOUNT);
+      amountBN.lt(minimumWithdrawAmountBN);
 
     if (selectedAction === 'deposit') {
       return tokenAmountBN.lte(availableBalanceBN) && !isBelowDepositMin;
     }
 
     if (selectedAction === 'withdraw') {
-      return amountBN.lte(availableBalanceBN) && !isBelowWithdrawMin;
+      return amountBN.lte(maximumWithdrawAmountBN) && !isBelowWithdrawMin;
```
sidmorizon, line 1045: # 🟡 P2 预留 gas 后的 0.01 USDC 区间里按钮静默禁用且无任何提示

## 问题

提现分支的校验改成了和 `maximumWithdrawAmountBN`（`availableBalanceBN` 减去 `USDC_WITHDRAW_GAS_RESERVE = 0.01`）比较，但 `isInsufficientBalance`（第 `1567` 行）仍然和 `availableBalanceBN` 比较，`errorMessage` 也只覆盖「低于最小额」这一种情况，两处上界不一致。

## 影响

当输入金额落在 `(maximumWithdrawAmountBN, availableBalanceBN]` 这个宽度为 `0.01` USDC 的区间时，`isValidAmount` 为 `false`，CTA 被 `withdrawSubmitDisabled` 禁用；但 `isInsufficientBalance` 为 `false`、`errorMessage` 为空、按钮文案仍是正常的提现文案，用户看到的是一个毫无解释的灰色按钮。界面上展示的可用余额正是 `availableBalance.balance`，用户照着这个数字手动输入就必然落进该区间；`validateAmountBeforeSubmit` 里那句余额不足的 Toast 也因为按钮已禁用而永远不会触发，形成没有出口的死状态。

## 建议

让提现分支的余额比较统一使用 `maximumWithdrawAmountBN`：

```ts
const isInsufficientBalance = useMemo(() => {
  const checkBN = selectedAction === 'deposit' ? tokenAmountBN : amountBN;
  const limitBN =
    selectedAction === 'deposit' ? availableBalanceBN : maximumWithdrawAmountBN;
  return checkBN.gt(limitBN) && checkBN.gt(0);
}, [amountBN, tokenAmountBN, availableBalanceBN, maximumWithdrawAmountBN, selectedAction]);
```

或在提现态直接把展示的可用余额改为 `maximumWithdrawAmountBN`，保证「显示多少就能输多少」。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1745,21 +1819,109 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+        wallet.getAddress(),
+      ]);
+      route = resolvedRoute;
+      if (
+        destinationConfig.transferType === 'cctp' &&
+        (!params.expectedRoute || route !== params.expectedRoute)
+      ) {
+        throw new OneKeyLocalError(
+          'Withdrawal route changed. Review the updated fee and try again.',
+        );
```
sidmorizon, line 1837: **Route-change retries fail again**

Severity: `non-severe`

After submission detects a route change, `setInterval` leaves the main-runtime route stale for 30 seconds after the background-runtime cache updates. Their JS heaps are separate; no native resource is involved, so immediate retries fail again.

packages/kit-bg/src/services/ServiceHyperLiquid/usdcWithdrawRoute.ts
```diff
@@ -0,0 +1,75 @@
...
+
+async function fetchWithdrawRoute(): Promise<IUsdcWithdrawRoute> {
+  const response =
+    await requestLoggedHyperLiquidTransport<IUsdcRoutingResponse>(
+      new HttpTransport(),
+      'info',
+      { type: 'usdcRouting' },
+      { action: 'usdcRouting' },
+    );
+  return parseRoute(response?.withdrawalRoute) ?? FALLBACK_ROUTE;
```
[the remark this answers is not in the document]
Minnzen: Fixed in `ac7e110db6`.

Confirmed: `requestLoggedHyperLiquidTransport` logs `{status:'err'}` bodies but still returns them, and a missing or unknown `withdrawalRoute` landed in the same place, so an unreadable answer was cached as `bridge` for 5 minutes and overwrote `lastResolvedRoute` — defeating the guard that exists to prevent exactly that.

`fetchWithdrawRoute` now resolves to `undefined` for anything unparsable. The cached path keeps the last confirmed rail and neither caches nor remembers the bad answer; the submit path throws instead. Added a test covering it.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -1957,6 +1959,46 @@ function DepositWithdrawContent({
...
+      .catch((error) => {
+        console.error(
+          '[DepositWithdrawModal] Failed to resolve withdraw route:',
+          error,
+        );
+      });
+    return () => {
+      cancelled = true;
+    };
+  }, [selectedAction]);
```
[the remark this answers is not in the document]
Minnzen: Already closed by #13125, which merged into this branch.

Submission now re-resolves the rail live (`getLiveUsdcWithdrawRoute`, no fallback) and the fee from chain, and hard-fails if either differs from what the form quoted via `expectedRoute` / `expectedCctpFee`. The modal also refreshes its quote every 30s. Showing $0.20 while charging $1.00 is no longer reachable.

The related gap you and another reviewer both raised — that the rail itself was only resolved once per form open, so the "try again" the error asks for could never succeed — is fixed in `fcbd6c6c0b`.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
Minnzen: Checked this one and I don't think it holds.

The withdrawable the form validates against and the `sourceDex` the action uses read the same `perpsAbstractionModeAtom` through the same account-match guard (`perps.ts:362-368` vs `ServiceHyperliquidExchange.ts:1798-1806`). When the atom is unset or belongs to another account, `perpsComputedAccountValueAtom` serves the perp-side withdrawable — which is exactly what `sourceDex: ''` spends. The two agree by construction rather than by coincidence.

Disagreement needs the atom to flip between render and submit, i.e. an account switch with the modal open. Hyperliquid then rejects the action server-side and the user sees a retryable error; funds can only ever move to the user's own address. Leaving as-is.

packages/kit/src/views/Perp/components/Portfolio/usePerpPortfolioData.ts
Minnzen: You're right that the branch conflicted with `accountClassTransfer`, and the sequence you gave is reachable. Fixed in `ac7e110db6`, though not the way you proposed — filtering on `sourceDex === ''` would have been worse.

Live data: the account we tested on has nine `send` deltas to the system address and **every one carries `sourceDex: 'spot'`**, because `_resolveWithdrawSourceDex` sends `spot` for unified and portfolio-margin accounts. Filtering on `''` would erase every withdrawal OneKey itself produces from this stat — far more common than the double count, which additionally needs `accountClassTransfer` deltas that only Hyperliquid's own UI emits.

The real defect was the other branch. This stat's tooltip reads "Total deposits minus total withdrawals during this period", and `accountClassTransfer` is neither — it moves USDC between the perp and spot balances of one account. Counting it was already wrong by that definition, and it became your double subtraction once a withdrawal could be sourced from spot. Dropping it makes your sequence net to 0 and the stat match its own description.

Also removed the inbound arm I had added: EVM → Core credits arrive as `spotTransfer` from the system address, never as a `send` from it (verified against the system address's own ledger — 86 such credits in 15 minutes, zero inbound sends), so it was dead code and the round-trip symmetry I claimed in the commit message did not hold.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed in `fcbd6c6c0b` — you're right that the retry the error asks for could never succeed.

The route now refreshes on the same 30s tick as the fee quote. The background service caches the rail for 5 minutes, so most ticks resolve without a network call.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
Minnzen: Same root cause as the rail-refresh thread; fixed in `fcbd6c6c0b`.

The modal now re-resolves the rail every 30s while the withdraw form is open, so after a rail change the quote and the submit-time check converge on their own instead of failing until the form is reopened.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed as suggested, in the commit above.

`isInsufficientBalance` now compares against `maximumWithdrawAmountBN` on the withdraw branch, so an amount inside the 0.01 reserve reports insufficient balance instead of silently greying out the button. Kept the displayed balance untrimmed rather than showing the reserved figure, since MAX already fills the correct amount and the reserve exists for a sub-cent gas leg that would otherwise read as an unexplained deduction.

packages/kit/src/views/Perp/components/Portfolio/usePerpPortfolioData.ts
```diff
@@ -127,13 +128,23 @@ export function usePerpPortfolioData(
...
-          delta.toPerp !== undefined
+          delta.type === 'send' &&
+          delta.token === 'USDC' &&
+          delta.destination?.toLowerCase() ===
+            HYPEREVM_SYSTEM_ADDRESS.toLowerCase() &&
+          delta.amount
         ) {
-          return delta.toPerp ? sum.plus(delta.usdc) : sum.minus(delta.usdc);
+          return sum.minus(delta.amount);
         }
```
sidmorizon: > <sub><sub>[P2 Badge](https://img.shields.io/badge/P2-yellow?style=flat)</sub></sub>  Core → HyperEVM 只记出账、不记入账，往返一次净入金永久少算

## 问题

移除 `accountClassTransfer` 之后口径确实统一到了「整账户（perp + spot）」，我原先报的双扣已经消失——这条独立于那个问题。

现在 reducer 只剩三个分支：`deposit` `+`、`withdraw` `-`、`send → HYPEREVM_SYSTEM_ADDRESS` `-`。按 `perp.constants.ts:66-70` 自己的注释，第三个分支匹配的是「USDC leaving HyperCore, whether it is a CCTP withdrawal or a plain transfer to HyperEVM」，即 **Core → HyperEVM 的全部出账**，其中包含 `USDC_WITHDRAW_DESTINATIONS` 里 `id: 'hyperevm'` 这个目的地。

但反向的 HyperEVM → Core 入账没有任何分支接住。按你在上一条回复里给出的实测（系统地址 15 分钟内 86 笔 credit，全部是 `spotTransfer`，没有 inbound `send`），这些入账的形态是 `delta.type === 'spotTransfer'` 且 `delta.user === HYPEREVM_SYSTEM_ADDRESS`（`AccountRow.tsx:216-218` 正是用 `delta.user` 判断 spotTransfer 的收/发方向，`:238` 用 `delta.usdcValue` 取金额）。`getPortfolioNetDeposits`（`ServiceHyperliquid.ts:4022-4034`）返回的是未过滤的 `userNonFundingLedgerUpdates`，所以这些 delta 确实在数组里，只是落到了末尾的 `return sum`。

## 影响

出账扣、入账不加，二者不对称。用户在 App 内提现到 HyperEVM（`-100`），之后把这笔钱从 HyperEVM 转回 Core（`spotTransfer`，`0`），一次往返后 Net Deposits 永久少 `100`，多往返几次可以把一个资金持平的账户显示成大额负数；从 HyperEVM 侧入金再走 CCTP 提现的用户同样会看到 `-` 而不是 `0`。

App 自身的入金走 `HYPERLIQUID_DEPOSIT_ADDRESS`（Arbitrum bridge，落在 `deposit` 分支）不受影响，所以这不影响主链路，但 `hyperevm` 是 App 提现目的地列表里的正式选项，往返路径完全在产品支持范围内。这个数字对用户就是资金数字，负值会被直接理解成亏损。

## 建议

补上与 `send` 分支对称的入账分支，让「整账户」口径两个方向都成立：

```ts
if (
  delta.type === 'spotTransfer' &&
  delta.token === 'USDC' &&
  delta.user?.toLowerCase() === HYPEREVM_SYSTEM_ADDRESS.toLowerCase() &&
  delta.usdcValue
) {
  return sum.plus(delta.usdcValue);
}
```

如果确认只想统计 Arbitrum bridge 两端、不把 HyperEVM 视为账户外部，那么对称的做法是把 `send` 分支也限制到 CCTP 提现本身，而不是匹配全部流向系统地址的出账；无论取哪一种，两个方向需要用同一条边界。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

originalix:

Review of commit `da2bd77e6905` did not finish successfully.

packages/kit/src/views/Perp/components/Portfolio/usePerpPortfolioData.ts
Minnzen: Right, and the reason I gave for leaving it out is one I removed myself — once `accountClassTransfer` was gone there was no double count left to avoid. Added in the commit above, matching the shape you described (`spotTransfer` with `user` = the system address).

The reducer is now symmetric under whole-account scope: `deposit` +, `withdraw` −, `send` → system −, `spotTransfer` ← system +. A Core → HyperEVM → Core round trip nets to zero, which is what the stat's own description asks for.

originalix (commented):
@Minnzen

Reviewed commit `2b01639d5bd2`.

- **P1 · Use Hyperliquid's fixed signing chain ID for every withdrawal**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13119#issuecomment-5506886545)

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -113,8 +128,93 @@ const PERP_DESKTOP_DEPOSIT_AMOUNT_INPUT_BLOCK_HEIGHT = 220;
...
+  }
+  if (route === 'cctp') {
+    return {
+      components: [
+        {
+          kind: 'cctpForwarding',
+          amount: destination.fallbackFee.toString(),
+          token: 'USDC',
+          debitedFrom: 'withdrawAmount',
+          isEstimate: false,
```
sidmorizon: ## 问题

CCTP 预览分支把常量 `destination.fallbackFee` 标成 `isEstimate: false`。`formatWithdrawFeeComponent` 只在 `isEstimate` 为真时加 `≈`，所以这一行会以「$1.20 / $0.20」这种确定值的形态渲染一个写死的兜底费。

但 `fallbackFee` 恰恰是本仓库自己声明不可信的那个数：`cctpWithdraw.ts` 中 `getLiveUsdcWithdrawFee` 上方的注释写着 "Submission must not rely on a cached or static CCTP fallback fee. The contract can update per-domain USDC fees"，而 `getCctpWithdrawFee` 在 RPC 失败回退到同一个 `fallbackFee` 时，标的是 `isEstimate: true`。同一个常量在 service 里是估算值，在预览里却成了确定值。

另外 fee 拉取失败时 `catch` 会把 `withdrawFeeState` 置为 `undefined`，于是 `withdrawFeeDisplayQuote` 长期回落到这个预览：改动前这里显示 `--`（费用未知），现在会持续显示一个确定的金额，同时 CTA 因 `isWithdrawDestinationReady` 为 false 而灰着，用户看不出这个数尚未确认。

## 影响

提交路径本身没问题：`cctpFeeComponent` / `isWithdrawDestinationReady` / `expectedCctpFee` 都取自按 `key` 匹配的已确认 quote，预览进不到提交，也不会放开按钮。所以这是纯展示问题——但发生在资金页，且在合约费率与 `fallbackFee` 不一致时，用户会先看到一个不带估算标记的错误金额（首次打开、bg 缓存未命中、移动端 RPC 慢，以及 fee 拉取失败后的持续状态）。

## 建议

预览分支统一用 `isEstimate: true`（bridge 分支的 `WITHDRAW_FEE` 是固定值可保持 `false`，CCTP 这条应改），让未确认的费用渲染成 `≈ $1.20`，与 service 对同一常量的处理保持一致，也让「未确认」在 UI 上可区分。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

packages/kit-bg/src/services/ServiceHyperLiquid/usdcWithdrawRoute.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!inFlightRequest) {
+    inFlightRequest = fetchWithdrawRoute()
+      .then((route) => {
+        if (!route) {
+          // Neither cache nor remember an unreadable answer, or one malformed
+          // response would pin the user to the fallback rail for the whole TTL.
+          return lastResolvedRoute ?? FALLBACK_ROUTE;
+        }
+        cachedRoute = { route, fetchedAt: Date.now() };
+        lastResolvedRoute = route;
```
sidmorizon, line 60: **Older lookup restores stale rail**

Severity: `non-severe`

When `getUsdcWithdrawRoute` overlaps `getLiveUsdcWithdrawRoute`, the older request can overwrite the newly confirmed rail. The form then retries with a stale route for five minutes.

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
```diff
@@ -0,0 +1,242 @@
...
+  }
+
+  const inFlight = inFlightFeeRequests.get(destinationId);
+  if (inFlight) {
+    return inFlight;
+  }
+
+  const request = fetchCctpWithdrawFee(config, rpcCall)
+    .then((fee) => {
+      cachedFees.set(destinationId, { fee, fetchedAt: Date.now() });
```
sidmorizon, line 127: **Older lookup restores stale fee**

Severity: `non-severe`

When an expired fee refresh overlaps `getLiveUsdcWithdrawFee`, the older RPC can replace the newly confirmed fee. Retries then fail against the stale quote for five minutes.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -1396,10 +1607,33 @@ function DepositWithdrawContent({
...
       } else {
+        // Submission stays bound to a confirmed fee; only the wait for it moved
+        // here from the button's disabled state.
+        let expectedCctpFee: string | undefined;
+        if (
+          selectedWithdrawDestination.transferType === 'cctp' &&
+          withdrawRoute === 'cctp'
+        ) {
+          expectedCctpFee = cctpFeeComponent?.amount;
+          if (!expectedCctpFee) {
```
sidmorizon: # 🔴 P1 目的地切换窗口内，提交会绑定一笔用户从未看到的手续费

## 问题

本次提交把 CTA 的门槛从已确认报价 `withdrawFeeQuote` 换成了 `withdrawFeeDisplayQuote`（466-470 行）。后者在确认报价缺失时回落到 `buildWithdrawFeePreviewQuote()` 返回的**静态** `destination.fallbackFee`（Ethereum = 1.2，其余 = 0.2）。

切换提现目的地时 `withdrawFeeKey` 立即变化，`withdrawFeeQuote` 变为 `undefined`；而 `amount` 不会被清空（`setAmount('')` 只在充值代币变化和原生键盘长按时触发），所以按钮保持可点——这正是本次提交想要的效果。

此时用户点击提现，走到这里的分支：`cctpFeeComponent?.amount` 为空 → 现取一次 `getUsdcWithdrawFee()` → 把取回的链上真实费用直接赋给 `expectedCctpFee` 并提交。这个新值**既没有写回 `withdrawFeeState`**（费用行不会刷新），**也没有任何二次确认**；`withdraw()` 成功后立即 `onClose?.()` 关闭弹窗。提现路径不像充值那样经过 `normalizeTxConfirm`，费用行是唯一的告知渠道。

## 影响

CCTP forwarding 费用是随目标链 gas 浮动的链上读数（`eth_call` 到 CoreDepositWallet），与静态 `fallbackFee` 可以相差很多，Ethereum 尤其明显；且该费用 `debitedFrom: 'withdrawAmount'`，直接从提现本金里扣。

具体后果：用户在费用行看到 `$1.20`（因 `isEstimate: false`，连 `≈` 都不带，读起来是确定值），切换目的地后立刻点提现，实际按 `$4.50` 签名发出，到账比预期少 `$3.30`，全程无任何提示。后台 `getLiveUsdcWithdrawFee` 的硬校验此时**不会拦截**——它校验的正是这个用户没看过的值，校验通过恰恰意味着这笔未披露的费用被确认执行。

`expectedCctpFee` 机制存在的意义就是"用户不会被收取他没看到的费用"，这条性质在切换目的地的窗口内被打破了。窗口在 bg 侧 5 分钟费用缓存未命中时是一次完整的 `eth_call` 往返（移动端几百毫秒到数秒），而用户切完目的地的下一个动作恰好就是点提现。

## 建议

保留按钮不闪烁的效果，但不要让"未展示过的费用"直接进入提交。在本分支拿到 `feeQuote` 后，先与当前展示值比较：

- 一致：照常提交；
- 不一致：`setWithdrawFeeState({ key: withdrawFeeKey, quote: feeQuote })` 刷新费用行并**中止本次提交**，让用户看到新费用后再点一次（或弹一个确认框）。

这样按钮在切换瞬间依然可点，而"提交的费用 == 用户看到的费用"的保证得以保留。

[Claude session ↗](https://claude.ai/code/session_018wae8qujCdGQpbuPt1q71D)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -113,8 +128,94 @@ const PERP_DESKTOP_DEPOSIT_AMOUNT_INPUT_BLOCK_HEIGHT = 220;
...
+// same number, and submission re-reads it and refuses on drift, so the marker
+// only ever blinks in and out without telling the user anything new.
+function formatWithdrawFeeComponent(
+  component: IUsdcWithdrawFeeQuote['components'][number],
+) {
+  const amount = new BigNumber(component.amount).toFixed(2);
+  if (component.kind === 'hyperEvmGas') {
+    return `< $${amount}`;
+  }
+  return `$${amount}`;
```
sidmorizon: **Estimated withdrawal fee appears exact**

Severity: `non-severe`

When the fee read fails after an on-chain update, `formatWithdrawFeeComponent` presents the stale fallback as exact. Users review an incorrect charge before submission rejects it.

Suggested fix:

```ts
  return `${component.isEstimate ? '≈ ' : ''}$${amount}`;
```

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
```diff
@@ -0,0 +1,242 @@
...
+    throw new OneKeyLocalError(
+      `Destination does not use CCTP: ${destinationId}`,
+    );
+  }
+
+  return {
+    config,
+    destinationRecipient: ownerAddress,
+    addressEncoding: config.addressEncoding,
+    destinationChainId: config.domain,
```
originalix, line 233: P1: [CCTP destinationChainId likely uses domain id instead of EVM chain id]

When building the CCTP withdraw action, `destinationChainId` is set from `config.domain` (Circle domain ids such as Arbitrum `3`).

`@nktkas/hyperliquid` types/docs describe this field as the target blockchain chain id, and the SDK example passes Arbitrum as `42161`. Fee quoting correctly uses the Circle domain in the eth_call payload, but that domain value appears to be reused here for the signed exchange action.

If Hyperliquid expects an EVM chain id, CCTP withdrawals can be rejected or routed incorrectly.

Please map `sendToEvmWithData.destinationChainId` to the real EVM chain id for the selected network, and keep Circle domain ids only for the fee contract call.

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
```diff
@@ -0,0 +1,242 @@
...
+  const domain = destinationDomain.toString(16).padStart(64, '0');
+  return `0x${CCTP_WITHDRAW_FEE_SELECTOR}${useForwarding}${domain}` as IHex;
+}
+
+function parseWithdrawFee(result: unknown): number {
+  if (typeof result !== 'string' || !/^0x[0-9a-fA-F]+$/.test(result)) {
+    throw new OneKeyLocalError('Invalid CCTP withdrawal fee response');
+  }
+  const fee = new BigNumber(result.slice(2), 16)
+    .shiftedBy(-USDC_DECIMALS)
```
originalix, line 76: P2: [CCTP fee parsing converts USDC amounts through JS number]

`parseWithdrawFee` does `BigNumber(...).shiftedBy(-USDC_DECIMALS).toNumber()` and later caches/compares `fee.toString()`.

UI quote and submit live-check both depend on that string. Any non-canonical decimal introduced by JS number conversion can make an unchanged chain fee look different and fail with "Withdrawal fee changed".

Please keep the fee as a decimal string from BigNumber end-to-end (for example `toFixed()`), and compare those strings/BigNumbers without `toNumber()`.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -1957,6 +2209,213 @@ function DepositWithdrawContent({
...
+    if (selectedAction !== 'withdraw') {
+      return;
+    }
+    let cancelled = false;
+    setWithdrawRoute(undefined);
+    const loadWithdrawRoute = () => {
+      void backgroundApiProxy.serviceHyperliquidExchange
+        .getUsdcWithdrawRoute()
+        .then((route) => {
+          if (!cancelled) {
```
sidmorizon, line 2184: **Route changes cause repeated failures**

Severity: `non-severe`

When Hyperliquid changes rails, `getUsdcWithdrawRoute()` leaves the form stale. The instructed retry fails again until the next 30-second poll.

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/AccountRow.tsx
```diff
@@ -188,6 +195,17 @@ const getDisplayType = (
...
+  // `send` carries both directions, and the shared config renders money coming
+  // in — so a withdrawal would show as a credit. Self-transfers are neither.
+  if (deltaType === 'send') {
+    const isSentToOthers =
+      'user' in delta &&
+      'destination' in delta &&
+      currentAddress &&
+      delta.user.toLowerCase() === currentAddress.toLowerCase() &&
+      delta.destination.toLowerCase() !== currentAddress.toLowerCase();
+    return isSentToOthers ? 'sendOut' : 'send';
```
sidmorizon, line 207: **Self-transfers appear as incoming funds**

Severity: `non-severe`

When a `send` has the same sender and recipient, `getDisplayType()` returns the green incoming type. Internal movements appear as received funds.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed — the preview is gone entirely as of `7705ab29db`.

You were right that it presented a constant the service itself treats as unreliable. Quoting every destination when the form opens closed the gap the preview existed to bridge, so on open it was only adding a second transition. The row now shows a skeleton until the confirmed quote lands, and quotes are kept per destination across modal opens so a reopen does not blank it again.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Confirmed and fixed in `da2bd77e69` — thank you, this was the right call to escalate.

I had loosened the CTA gate to stop the button greying out during the switch window, and traded a cosmetic problem for a funds-disclosure one. The button waits on the confirmed quote again and the at-press fetch is gone, so `expectedCctpFee` can only ever be the number the row showed.

The flicker that change was chasing is now fixed the way it should have been: every destination is quoted when the form opens (`c9d6edb19f`), so a switch lands on a confirmed quote instead of a gap.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Restored in `da2bd77e69` — the marker is back exactly as you suggested.

Removing it was my attempt to stop the prefix blinking in and out on destination change. That flicker is now gone at the source: quotes are fetched for every destination up front, so the row goes straight to the confirmed number and the estimate path is only reached when a chain read genuinely fails — which is precisely when the marker should show.

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
Minnzen: Checked this against mainnet rather than the SDK docs, and the domain id is correct — no change needed.

Our own builder account withdrew 115,000 USDC with `destinationChainId: 3` and it settled; a later 4 USDC test withdrawal did the same and landed 3.80 on Arbitrum (4.00 minus the 0.20 forwarding fee) in about 10 seconds. Circle's integration guide for this exact action is explicit: *"`destinationChainId`: The CCTP destination domain ID (for example, `3` for Arbitrum, `0` for Ethereum, `6` for Base)"* — https://developers.circle.com/cctp/howtos/withdraw-usdc-from-hypercore-to-evm

The EIP-712 field is `uint32`, which also fits a domain id rather than a chain id. The `42161` in the pinned SDK's docstring example is wrong; Hyperliquid's own docs leave the field undocumented, which is likely where that example came from.

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
Minnzen: Checked this one empirically and the round trip turns out to be lossless for every fee this can carry, so leaving it as is.

USDC is 6 decimals, and any 6-decimal value in this range survives a double exactly; `Number.prototype.toString` then emits the shortest decimal that round-trips, which is the same string `BigNumber.toFixed()` produces. Ran the real fees plus the boundaries — `200000`, `500000`, `1200000`, `1`, `999999`, `123456`, `333333`, `666667`, `10000000`, `1000000000` — and `toNumber().toString()` matched `toFixed()` on all of them.

The concern is right in principle; it would bite at magnitudes or precisions this field cannot reach. Not worth touching a funds path on a release branch for it.

packages/kit-bg/src/services/ServiceHyperLiquid/usdcWithdrawRoute.ts
Minnzen: Real, but leaving it: it needs two lookups to overlap, and the worst outcome is a stale rail for up to the cache window.

Submission does not trust that value. It re-resolves the rail with no fallback and refuses if it differs from what the form quoted, so a stale rail produces a rejected withdrawal, never a wrong one. Given this branch ships as a hot update and has already turned over several times chasing a display bug, an extra guard here costs more than the failure mode does.

packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts
Minnzen: Same call as the rail thread: real, but the exposure is bounded and submission does not trust the cached value.

The live fee is re-read at submit and compared against what the row showed; a stale quote fails the withdrawal rather than mispricing it. Not worth another change to a funds path on a release branch.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Accurate, and accepted as is.

Hyperliquid changing rails is a server-side event we have not seen happen, and the 30s refresh added in `fcbd6c6c0b` is the recovery — the form heals on its own rather than staying stuck. Making the retry immediate means routing the submit failure back into the route effect, which is more machinery than a rare event that already self-corrects justifies on a release branch.

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/AccountRow.tsx
Minnzen: Agreed that it reads wrong, but it predates this PR and I am deliberately not changing it here.

Before this branch, every `send` rendered with the incoming config. This change only reclassified sends to another party, which is what a CCTP or HyperEVM withdrawal is; a transfer between a user's own perp and spot balances renders exactly as it did before. Giving it a neutral treatment is the right fix, but it is a separate behaviour change on a display path, and this branch is a hot update that has already turned over enough times.

Worth its own ticket rather than folding in here.

originalix (approved):

PR #13097, opened 2026-08-31
fix: prevent iOS Perps layout shift on cold start

huhuanming:
## Summary

- keep the iOS native Perps fallback spacer aligned with the measured floating header height
- reuse the iOS header top-padding constant for both the initial fallback and rendered header
- leave Android and web layout behavior unchanged

## Root cause

On iOS cold start, the initial spacer used `safeAreaTop + 44`, while the floating header measured as `safeAreaTop + 50` because of the iOS-specific 26pt top padding and the -20pt offset. Updating the measured height moved the Perps body down by 6pt.

## Cross-platform impact

- iOS: fixes the 6pt cold-start layout shift
- Android: initial and measured heights already match at `safeAreaTop + 44`
- Web: does not render the native spacer or measured absolute header; `md` (<=767px) continues to use the mobile layout and `gtMd` (>=768px) the desktop layout

## Verification

- iOS simulator cold launch: tracked the Perps body across 109 consecutive frames after first render; Y position stayed constant with 0pt displacement
- `yarn agent:check --profile commit`
- `yarn agent:check --profile pr`
- `yarn jest packages/kit/src/views/Perp/layouts/PerpDesktopLayout.web.test.tsx --runInBand` (9/9 passed)

PR #13128, opened 2026-09-02
fix: Perps close-position price readiness, cold-start banner shift, favorites order sync and TWAP tab shadow flash(OK-61259 OK-61504 OK-61486 OK-61520)

Minnzen:
[OK-61259](https://onekeyhq.atlassian.net/browse/OK-61259) [OK-61504](https://onekeyhq.atlassian.net/browse/OK-61504) [OK-61486](https://onekeyhq.atlassian.net/browse/OK-61486) [OK-61520](https://onekeyhq.atlassian.net/browse/OK-61520)

----------

## Summary

Four Perps fixes for the 6.5.2 bundle. All four code paths are byte-identical on `x`, so this lands once here and follows the regular `release/v6.5.2 → x` sync.

- **OK-61259** (online, High): on iOS cold start the close-position dialog let the user confirm a market close before any mid price had reached the main runtime, and the tap failed with "Unable to get current market price". Confirm now stays disabled with a spinner until a valid price exists, and the submit re-resolves a fresh, dex-scoped price in the background. The dialog is also single-instance and guarded against double taps.
- **OK-61504** (High): ports #13097 from `x` so the iOS header spacer matches the measured header and the Perps body no longer drops 6pt after first paint on cold start.
- **OK-61486**: the footer ticker and the token selector's Favorites tab follow the drag order set in the favorites bar.
- **OK-61520**: switching TWAP sub-tabs no longer flashes the fixed-column shadow on a full-width table.

The two remaining tickets from the same batch (limit BBO toggle, iOS native chart) only exist on 6.6.0 code and are handled against `x` separately.

## Intent & Context

Working through the Perps Jira "doing" list for this sprint. Four of the six tickets were reported by QA on 6.6.0 test builds, but code inspection showed the affected paths are unchanged on the 6.5.2 hot-update line, and OK-61259 is an open High bug reported against production 6.5.0, so the batch is split by release line: these four land here and flow to `x` through the regular sync; the two 6.6-only tickets are handled against `x` separately.

OK-61259 was diagnosed from the reporter's log and screenshot (OneKey 6.6.0 build 2026082600, bundle 20484813, iOS 26.6).

### OK-61259 — "Unable to get current market price" when closing right after cold start

`main` and `bg` are separate JS runtimes with separate heaps. On cold start the main runtime restores positions, the order book and the trading buttons from its cold-start cache, but `perpsAllMidsAtom` has no cache and only fills once the background WebSocket delivers the first `allMids` and forwards it over the event bus. In the log the user opened the close dialog and tapped Confirm at 17:35:42, the first `allMids` reached `bg` at 17:35:44 (4.2s after launch), no `ordersClose` RPC followed, and the same action had succeeded in 885ms one launch earlier once prices were present.

The dialog made this reachable: `midPrice` collapsed a missing price to the string `'0'`, `isPriceValid` used `Boolean(midPrice)` (true for `'0'`), so Confirm was enabled, and only the submit path rejected `'0'` with the toast in the screenshot.

The same path also allowed double submits: `handleClosePosition` opened a new dialog on every tap (stacked dialogs with stale snapshots), and `handleSubmit` relied on a `disabled` state that lands one render late and was released before the dialog's 300ms exit animation finished.

### OK-61504 — trading area jumps after cold start (iOS)

Same defect #13097 fixed on `x`: on iOS the initial native header spacer used `safeAreaTop + 44` while the floating header measured `safeAreaTop + 50` (26pt top padding minus the 20pt offset), so the first `onLayout` moved the whole Perps body down by 6pt. `Perp.tsx` is identical on this line, so the commit is cherry-picked as-is; the later sync to `x` sees the same change on both sides.

### OK-61486 — favorites order not synced

`FavoritesBar.web.tsx` persists the drag result to `perpsFavoritesOrderPersistAtom` and reorders its own rows locally. `usePerpsFavorites` never read that atom, so the footer ticker rendered membership (add-time) order. The token selector did read it, but then re-sorted the Favorites tab whenever `sortField` was truthy, and the selector config ships `volume24h` as its default, so the drag order was always overwritten.

### OK-61520 — shadow divider flashes on TWAP sub-tab switch

`PerpTwapList.tsx` renders the three sub-tabs as separate unkeyed `CommonTableListView` instances, so a switch remounts the table, and `useFixedColumnShadow` seeds `showShadow` from `initialVisible: true`. The flash is the CSS transition rather than a wrong state: the fixed column is inserted with the seeded `box-shadow`, reading `scrollWidth` in the measurement effect forces a style flush that records that as the before-change style, and the switch to `none` is then animated over the 0.2s `transition` even when the measurement runs before paint.

## Design Decisions

**One price source for the gate, the submit and Close All.** `getValidPerpsPrice` turns missing, zero, negative or non-finite mids into `undefined`. The dialog prefers the main-runtime atom and otherwise polls `serviceHyperliquid.getMarketOrderReferencePrice(coin)` every 2s, which uses the background `allMids` cache when it is under 5s old and otherwise fetches a REST `allMids` snapshot; the submit re-resolves through the same method instead of trusting its React closure, and `closeAllPositions` prices every position through it as well, since the main-runtime atom carries no age marker. While no source has a price, market Confirm is disabled with a spinner; once the background source has also answered with nothing, the spinner stops and the existing toast shows, and the limit path stays editable (only the Mid shortcut dims) as the exit. Close All never submits a subset: a position still without a price aborts the action with a visible error.

**The REST fallback is queried per dex and never written to the shared cache.** The WebSocket `allMids` stream carries every dex (1221 coins in the log), but REST `allMids` returns one dex: verified live, the main-dex call returns 951 coins with no `xyz:` keys and `{dex: "xyz"}` returns 118 `xyz:` keys. The coin in the log is `xyz:NVDA`, so the dex is derived from the coin prefix, and the snapshot is not stored in `hyperLiquidCache.allMids` because a main-dex-only write would wipe the sub-dex prices other readers depend on.

**Close dialog is single-instance, guarded by `isExist()` plus a 1s mount grace.** `Dialog.show` mounts its portal asynchronously, so `isExist()` is still false for the first frames; the timestamp covers that window. The submit guard is a ref, not state, because state cannot stop a same-frame second press, and it stays armed after a successful submit because `onClose()` does not await the 300ms exit animation.

**Favorites tab sorts only after an explicit header click on that tab.** This mirrors how the category tabs already behave (`isPerpTokenSelectorDynamicTabUserSort`). The header indicator follows the same rule, so Volume no longer shows as active while the list is in drag order; clicking it cycles user desc → user asc → drag order. Ordering moved into `usePerpsFavorites` so every consumer inherits it.

**Shadow is measured in a layout effect and its transition stays off until the measured state has painted.** The hook now exposes `shadowTransitionEnabled`, which flips on two animation frames after the first measurement; `CommonTableListView` renders `transition: none` until then, so the first painted frame is already the measured state with nothing to animate, and later scroll-driven changes keep the fade. `InviteCodeListTable` and `PerpsRecordTable` use the same hook and are unchanged (they ignore the new field).

## Changes Detail

| File | Purpose |
|---|---|
| `shared/src/utils/perpsUtils.ts` | `getValidPerpsPrice`: finite and > 0 or `undefined` (+ test). |
| `kit-bg/.../ServiceHyperLiquid/utils/marketOrderReferencePrice.ts` | Background cache under 5s, else per-dex REST snapshot; falls through to REST when the fresh cache lacks the coin (+ 6 tests). |
| `kit-bg/.../ServiceHyperliquid.ts` | `getMarketOrderReferencePrice` background method; per-dex REST memo (1s, not written to the shared cache). |
| `OrderInfoPanel/ClosePositionModal.tsx` | Missing price is `undefined`; gate polls the background price source; Confirm disabled + spinner until ready, toast once when no source has a price; submit re-resolves the price in bg; limit input stays editable; single active dialog; `submittingRef` guard kept armed after success. |
| `TradingPanel/inputs/PriceInput.tsx` | `midPriceDisabled` dims the Mid shortcut without disabling the input. |
| `states/jotai/contexts/hyperliquid/actions.ts` | `closeAllPositions` prices every position through the background owner and aborts with an error instead of closing a subset. |
| `Perp/pages/Perp.tsx` | Cherry-pick of #13097: iOS header spacer matches the measured header. |
| `shared/src/utils/perpsTokenSelectorFavorites.ts` | `sortTokenSelectorFavoritesBySequence` (+ test). |
| `Perp/hooks/usePerpsFavorites.ts` | Items ordered by the persisted sequence. |
| `FavoritesBar/FavoritesBar.web.tsx` | Uses the shared sequence helper instead of its own merge. |
| `Perp/utils/tokenSelectorTabs.ts` | `isPerpTokenSelectorFavoritesTabUserSort`; header-active rule covers Favorites (+ test). |
| `TokenSelector/PerpTokenSelector.tsx`, `MoblieTokenSelector.tsx` | Favorites tab re-sorts only on user sort for that tab. |
| `kit/src/hooks/useFixedColumnShadow.ts` | Overflow measured in `useLayoutEffect`; `shadowTransitionEnabled` after the first measured paint. |
| `OrderInfoPanel/List/CommonTableListView.tsx` | Fixed-column `transition` gated on `shadowTransitionEnabled`. |

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop / Web / Mobile (iOS and Android)
- **Risk Areas**: market close now performs one background RPC before `ordersClose` (in-memory when the WebSocket cache is under 5s old; one HTTP round trip otherwise); a sub-dex coin whose price is missing from both the cache and the per-dex REST snapshot fails with the existing toast instead of sending a stale-price order; the close-dialog guard blocks a re-open for up to 1s after a dialog is shown or for 300ms after one closes; the favorites tab default order changes from volume-sorted to drag order; `useFixedColumnShadow` is shared with two referral tables, which only gain an earlier measurement.

## Test plan

**iOS real device, cold start (OK-61259, OK-61504)**

- [ ] With an open position, kill the app → launch → tap the Perps tab immediately → tap Close on the position → Confirm shows a spinner and is **not** tappable until the mark price appears in the dialog.
- [ ] Once Confirm enables, tap it once → exactly one `ordersClose` in the log, no "Unable to get current market price" toast, dialog closes.
- [ ] Same flow on a hip-3 position (an `xyz:` coin such as NVDA) → same result.
- [ ] During the same window, switch the dialog to Limit → the price input is editable and Mid is dimmed until the price arrives; Confirm enables once a valid limit price is typed.
- [ ] Close All right after cold start → either every position closes or one error toast appears with nothing submitted; never a partial close with a success toast.
- [ ] Kill the app → launch → Home → tap the Perps tab → after first paint the ticker / order book / trading form must **not** move down.
- [ ] Repeat the cold start twice more → layout identical each time.

**Desktop web / desktop app (wallet with an open position)**

- [ ] Positions list → click **Market** three times as fast as possible → exactly one close dialog opens; closing it reveals no second dialog underneath.
- [ ] In the close dialog, double-click **Confirm** → one order is sent; no second order during the dialog's fade-out.
- [ ] Drag a favorite to a new slot in the top bar → the footer ticker (favorites mode) shows the new order immediately.
- [ ] Open the token selector → Favorites tab shows the same order; the Volume header shows **no** active indicator.
- [ ] Click Volume → sorted desc with indicator; click again → asc; click again → back to drag order.
- [ ] Widen the window until the TWAP table needs no horizontal scroll → switch Active / History / Fills repeatedly → no shadow line flashes at the fixed column edge.
- [ ] Narrow the window until the table overflows → the shadow is present and stays across sub-tab switches.

**Mobile viewport (web ≤767px)**

- [ ] Repeat the two Close checks above on the position card's Close button.
- [ ] The mobile token selector's Favorites tab shows the drag order with no active Volume indicator.

[OK-61259]: https://onekeyhq.atlassian.net/browse/OK-61259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61504]: https://onekeyhq.atlassian.net/browse/OK-61504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61486]: https://onekeyhq.atlassian.net/browse/OK-61486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61520]: https://onekeyhq.atlassian.net/browse/OK-61520?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
@@ -150,6 +150,8 @@ const ClosePositionForm = memo(
         !priceUnavailableToastShownRef.current
       ) {
         priceUnavailableToastShownRef.current = true;
+        // TODO(i18n): pre-existing English literal shared with the submit
+        // path below and closeAllPositions; needs a Lokalise key.
         Toast.error({
           title: 'Unable to get current market price',
         });
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
@@ -27,6 +27,7 @@ import {
   calculateProfitLoss,
   formatHlSize,
   formatPriceToSignificantDigits,
+  getValidPerpsPrice,
   parseDexCoin,
   validateSizeInput,
 } from '@onekeyhq/shared/src/utils/perpsUtils';
@@ -81,7 +82,7 @@ const ClosePositionForm = memo(
     const assetId = tokenInfo?.assetId;
 
     const midPrice = useMemo(() => {
-      return allMids?.mids?.[position.coin] || '0';
+      return getValidPerpsPrice(allMids?.mids?.[position.coin]);
     }, [allMids?.mids, position.coin]);
 
     const tokenDisplayName = useMemo(() => {
@@ -219,7 +220,7 @@ const ClosePositionForm = memo(
 
     const handleUseMid = useCallback(() => {
       const latestMidPrice = midPrice;
-      if (latestMidPrice && latestMidPrice !== '0') {
+      if (latestMidPrice) {
         setFormData((prev) => ({
           ...prev,
           limitPrice: formatPriceToSignificantDigits(latestMidPrice),
@@ -263,8 +264,11 @@ const ClosePositionForm = memo(
         }
 
         if (formData.type === 'market') {
-          const latestMidPrice = midPrice;
-          if (!latestMidPrice || latestMidPrice === '0') {
+          const latestMidPrice =
+            await backgroundApiProxy.serviceHyperliquid.getMarketOrderReferencePrice(
+              position.coin,
+            );
+          if (!latestMidPrice) {
             Toast.error({
               title: 'Unable to get current market price',
             });
@@ -319,7 +323,7 @@ const ClosePositionForm = memo(
       formData.limitPrice,
       calculatedAmount,
       assetId,
-      midPrice,
+      position.coin,
       isLongPosition,
       hyperliquidActions,
       onClose,
@@ -532,7 +536,9 @@ const ClosePositionForm = memo(
             variant="primary"
             onPress={handleSubmit}
             disabled={!isFormValid || isSubmitting}
-            loading={isSubmitting}
+            loading={
+              isSubmitting || (formData.type === 'market' && !isPriceValid)
+            }
           >
             {intl.formatMessage({
               id: ETranslations.perp_confirm_order,
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
@@ -243,6 +243,7 @@ const ClosePositionForm = memo(
         return;
       }
       submittingRef.current = true;
+      let submitted = false;
       try {
         if (isNil(assetId) || Number.isNaN(assetId)) {
           return;
@@ -298,13 +299,18 @@ const ClosePositionForm = memo(
         }
 
         hyperliquidActions.current.resetTradingForm();
+        // The dialog stays mounted through its exit animation, so the guard
+        // must stay armed after success or a tap there places a second order.
+        submitted = true;
         if (isMountedRef.current) {
           onClose();
         }
       } finally {
-        submittingRef.current = false;
-        if (isMountedRef.current) {
-          setIsSubmitting(false);
+        if (!submitted) {
+          submittingRef.current = false;
+          if (isMountedRef.current) {
+            setIsSubmitting(false);
+          }
         }
       }
     }, [
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
@@ -15,6 +15,7 @@ import {
   YStack,
   useMedia,
 } from '@onekeyhq/components';
+import type { IDialogInstance } from '@onekeyhq/components';
 import backgroundApiProxy from '@onekeyhq/kit/src/background/instance/backgroundApiProxy';
 import { usePromiseResult } from '@onekeyhq/kit/src/hooks/usePromiseResult';
 import {
@@ -135,6 +136,7 @@ const ClosePositionForm = memo(
     }, []);
 
     const [isSubmitting, setIsSubmitting] = useState(false);
+    const submittingRef = useRef(false);
 
     const calculatedAmount = useMemo(() => {
       const percentage = Number.isNaN(formData.percentage)
@@ -235,6 +237,12 @@ const ClosePositionForm = memo(
     }, []);
 
     const handleSubmit = useCallback(async () => {
+      // The disabled state lands one render late, so same-frame double taps
+      // would otherwise send a second reduce-only order for the full size.
+      if (submittingRef.current) {
+        return;
+      }
+      submittingRef.current = true;
       try {
         if (isNil(assetId) || Number.isNaN(assetId)) {
           return;
@@ -294,6 +302,7 @@ const ClosePositionForm = memo(
           onClose();
         }
       } finally {
+        submittingRef.current = false;
         if (isMountedRef.current) {
           setIsSubmitting(false);
         }
@@ -531,16 +540,36 @@ const ClosePositionForm = memo(
 
 ClosePositionForm.displayName = 'ClosePositionForm';
 
+// Dialog.show mounts its portal asynchronously, so isExist() reports false
+// during the first frames; the grace window covers same-frame double taps.
+const CLOSE_DIALOG_MOUNT_GRACE_MS = 1000;
+let activeCloseDialog: IDialogInstance | undefined;
+let activeCloseDialogShownAt = 0;
+
 export function showClosePositionDialog({
   position,
   type,
   intl,
 }: IClosePositionParams & { intl: IntlShape }) {
+  // Rapid taps would stack dialogs, each holding its own position snapshot,
+  // so only one close dialog is kept alive at a time.
+  if (
+    activeCloseDialog &&
+    (activeCloseDialog.isExist() ||
+      Date.now() - activeCloseDialogShownAt < CLOSE_DIALOG_MOUNT_GRACE_MS)
+  ) {
+    return activeCloseDialog;
+  }
   const dialogInstance = Dialog.show({
     title: intl.formatMessage({
       id: ETranslations.perp_close_position_title,
     }),
     disableDrag: true,
+    onClose: () => {
+      if (activeCloseDialog === dialogInstance) {
+        activeCloseDialog = undefined;
+      }
+    },
     renderContent: (
       <PerpsProviderMirror>
         <ClosePositionForm
@@ -553,6 +582,8 @@ export function showClosePositionDialog({
     contentContainerProps: PERP_MOBILE_DIALOG_CONTENT_CONTAINER_PROPS,
     showFooter: false,
   });
+  activeCloseDialog = dialogInstance;
+  activeCloseDialogShownAt = Date.now();
 
   return dialogInstance;
 }
--- a/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
+++ b/packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
@@ -81,7 +81,7 @@ const ClosePositionForm = memo(
     const szDecimals = tokenInfo?.universe?.szDecimals ?? 2;
     const assetId = tokenInfo?.assetId;
 
-    const midPrice = useMemo(() => {
+    const atomMidPrice = useMemo(() => {
       return getValidPerpsPrice(allMids?.mids?.[position.coin]);
     }, [allMids?.mids, position.coin]);
 
@@ -114,9 +114,47 @@ const ClosePositionForm = memo(
     const initPriceRef = useRef(false);
     const isMountedRef = useRef(true);
 
+    const [referencePriceResolved, setReferencePriceResolved] = useState(false);
+    // The main runtime only sees mids the WebSocket has already forwarded,
+    // which lags by seconds on cold start, so the gate polls the same
+    // background price source the submit uses instead of waiting on the atom.
+    const { result: referencePrice } = usePromiseResult(
+      async () => {
+        if (atomMidPrice) {
+          return atomMidPrice;
+        }
+        const price =
+          await backgroundApiProxy.serviceHyperliquid.getMarketOrderReferencePrice(
+            position.coin,
+          );
+        if (isMountedRef.current) {
+          setReferencePriceResolved(true);
+        }
+        return price;
+      },
+      [atomMidPrice, position.coin],
+      { pollingInterval: 2000, checkIsFocused: false },
+    );
+    const midPrice = atomMidPrice ?? getValidPerpsPrice(referencePrice);
+    const isMarketPriceUnavailable = !midPrice && referencePriceResolved;
+    const priceUnavailableToastShownRef = useRef(false);
+
     useEffect(() => {
       setSliderPercentage(formData.percentage);
     }, [formData.percentage, setSliderPercentage]);
+
+    useEffect(() => {
+      if (
+        formData.type === 'market' &&
+        isMarketPriceUnavailable &&
+        !priceUnavailableToastShownRef.current
+      ) {
+        priceUnavailableToastShownRef.current = true;
+        Toast.error({
+          title: 'Unable to get current market price',
+        });
+      }
+    }, [formData.type, isMarketPriceUnavailable]);
 
     useEffect(() => {
       if (!midPrice) return;
@@ -484,7 +522,7 @@ const ClosePositionForm = memo(
             value={formData.limitPrice}
             onChange={handleLimitPriceChange}
             onUseMidPrice={handleUseMid}
-            disabled={!midPrice}
+            midPriceDisabled={!midPrice}
             szDecimals={szDecimals}
             ifOnDialog
           />
@@ -537,7 +575,10 @@ const ClosePositionForm = memo(
             onPress={handleSubmit}
             disabled={!isFormValid || isSubmitting}
             loading={
-              isSubmitting || (formData.type === 'market' && !isPriceValid)
+              isSubmitting ||
+              (formData.type === 'market' &&
+                !isPriceValid &&
+                !isMarketPriceUnavailable)
             }
           >
             {intl.formatMessage({
--- a/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
+++ b/packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
@@ -159,6 +159,7 @@ import {
 } from './userAbstractionCache';
 import { shouldPreserveConfirmedUserAbstractionMode } from './userAbstractionMode';
 import { buildDepositConfigFromTokensByNetwork } from './utils/depositConfigUtils';
+import { resolveMarketOrderReferencePrice } from './utils/marketOrderReferencePrice';
 import {
   mergePerpDexSlots,
   selectPerpMetasByDex,
@@ -1080,6 +1081,20 @@ export default class ServiceHyperliquid extends ServiceBase {
     {
       max: 1,
       maxAge: timerUtils.getTimeDurationMs({ seconds: 30 }),
+      promise: true,
+    },
+  );
+
+  // Per-dex REST snapshot that must not feed the shared cache: a main-dex-only
+  // result would wipe the sub-dex prices the WebSocket stream delivers.
+  _getMarketOrderAllMidsMemo = cacheUtils.memoizee(
+    async (dex: string) =>
+      dex
+        ? hyperLiquidApiClients.infoClient.allMids({ dex })
+        : hyperLiquidApiClients.infoClient.allMids(),
+    {
+      max: 8,
+      maxAge: timerUtils.getTimeDurationMs({ seconds: 1 }),
       promise: true,
     },
   );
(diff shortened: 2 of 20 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
```diff
@@ -294,6 +302,7 @@ const ClosePositionForm = memo(
           onClose();
         }
       } finally {
+        submittingRef.current = false;
```
sidmorizon: # 🟠 P1 提交成功后守卫过早释放，弹窗退场动画期间仍可二次下单

## 问题

`finally` 无差别执行 `submittingRef.current = false` 与 `setIsSubmitting(false)`。成功路径上 `onClose()` 只是 `dialogInstance.close()`，它不 await：`BaseDialogContainer.handleClose` 只做 `changeIsOpen(false)` 触发 `AnimatePresence` 退场，真正卸载由 `buildForwardOnClose` 里的 `setTimeout(..., 300)` 完成。所以 `finally` 跑到时组件仍挂载，`isMountedRef.current` 依然为 `true`。

结果是按钮的 `disabled={!isFormValid || isSubmitting}` 与 `loading` 一起复位，而 `formData.amount` 是本地 `useState`，`hyperliquidActions.current.resetTradingForm()` 只重置交易面板的 atom，不会清掉这里的数量。

## 影响

下单成功后存在约 `300ms` 的可点击窗口，且此时 `submittingRef` 已为 `false`，新增的守卫完全不拦截。用户在这个窗口再点一次 Confirm，会按 `formData.amount` 再发一笔 reduce-only 平仓单——部分平仓场景（例如 `50%`）会被平掉第二个 `50%`，正是 OK-61259 想消除的重复下单。

## 建议

成功路径不要释放守卫，让它随组件卸载一起消失；只有在早退和异常路径上复位：

```ts
try {
  // ...
  hyperliquidActions.current.resetTradingForm();
  submittedRef.current = true; // 标记已成功提交
  if (isMountedRef.current) onClose();
} finally {
  if (!submittedRef.current) {
    submittingRef.current = false;
    if (isMountedRef.current) setIsSubmitting(false);
  }
}
```

[Claude session ↗](https://claude.ai/code/session_01Q2G6AAGbbxWPwAAECsfAZb)

---

Minnzen: Agreed. `onClose()` does not await the 300ms exit animation, so the `finally` reset re-enabled Confirm while the dialog was still mounted. Fixed in 14fdcca091: the guard now resets only on early return or failure; after a successful submit it stays armed until the dialog unmounts.

packages/kit/src/views/Perp/components/PerpTips.tsx
```diff
@@ -1,31 +1,59 @@
...
+  const [frozenBannerConfig, setFrozenBannerConfig] =
+    useState(liveBannerConfig);
+
+  useEffect(() => {
+    liveBannerConfigRef.current = liveBannerConfig;
+  }, [liveBannerConfig]);
+
+  useFocusEffect(
+    useCallback(() => {
+      setFrozenBannerConfig(liveBannerConfigRef.current);
```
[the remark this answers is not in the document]
Minnzen: Confirmed against `preloadConfig.ts`: native preloads the Perp tab 2–3s after boot, so the background mount snapshot was `undefined` and the focus effect mounted the banner one render after the first visible frame. Fixed in 6aa12273c5: the snapshot now follows the live config whenever the screen is not focused (preloaded or blurred) and freezes only while it is focused, so the next visit paints with the banner already in place.

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
```diff
@@ -80,7 +82,7 @@ const ClosePositionForm = memo(
     const assetId = tokenInfo?.assetId;
 
     const midPrice = useMemo(() => {
-      return allMids?.mids?.[position.coin] || '0';
+      return getValidPerpsPrice(allMids?.mids?.[position.coin]);
```
sidmorizon, line 85: # 🟠 P1 限价关仓输入框在缺价时被锁死

## 问题

`midPrice` 从 `allMids?.mids?.[coin] || '0'` 改为 `getValidPerpsPrice(...)` 后，缺价时是 `undefined`，不再是恒真的 `'0'`。本文件 487 行 `PriceInput` 的 `disabled={!midPrice}` 此前永远为 `false`，现在会真正生效：`TradingFormInput` 把它透传给 `Input` 的 `disabled`，并且 `addOns={disabled && !showAddOnsWhenDisabled ? undefined : renderAddOns()}` 会连 `Mid` 快捷键一起隐藏。

同一时刻 `limitPrice` 仍是 `''`（初始化 effect 以 `if (!midPrice) return` 开头），`isPriceValid` 的 limit 分支因此也是 `false`，Confirm 同样 disabled。

## 影响

在本 PR 针对的 iOS 冷启动约 `4.2s` 窗口内，用户切到 Limit 后价格框不可编辑、无任何提示，market 与 limit 两条路径同时不可用；对任何在主进程 `perpsAllMidsAtom` 中长期缺失的 coin 则是永久锁死。改动前用户至少还能手动输入限价离场。

## 建议

把 `PriceInput` 的 `disabled` 与 mid 是否存在解耦：输入框保持可编辑，只让 `Mid` 这个 action 在 `!midPrice` 时禁用（`actions[].disabled`），而不是禁用整个输入。

[Claude session ↗](https://claude.ai/code/session_01Q2G6AAGbbxWPwAAECsfAZb)

---

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
```diff
@@ -517,7 +536,9 @@ const ClosePositionForm = memo(
             variant="primary"
             onPress={handleSubmit}
             disabled={!isFormValid || isSubmitting}
-            loading={isSubmitting}
+            loading={
+              isSubmitting || (formData.type === 'market' && !isPriceValid)
```
sidmorizon: # 🟡 P2 市价确认按钮的门禁比提交路径更严且无出口

## 问题

这里用 `formData.type === 'market' && !isPriceValid` 转圈，而 market 分支的 `isPriceValid` 仍是 `Boolean(midPrice)`，只认主进程 `perpsAllMidsAtom`。提交路径已经改走后台 `getMarketOrderReferencePrice`，缓存缺失时它会按 dex 发 REST `allMids` 兜底——门禁能拿到的价源严格窄于提交能拿到的价源。

按钮同时 `disabled={!isFormValid || isSubmitting}`，所以价格始终不到时是「永久 disabled + 永久转圈」，没有超时、没有 Toast、没有任何文案。

## 影响

WS 断线、该 coin 不在 `ALL_DEXS` 推送里、或主进程错过 emit 时，用户面对的是一个一直转圈的确认按钮且拿不到任何失败信息；改动前至少还会弹 `Unable to get current market price`，用户知道该重试。

## 建议

让门禁复用同一价源：用 `usePromiseResult` 调 `getMarketOrderReferencePrice(position.coin)` 驱动 market 分支的 `isPriceValid`；或给 loading 加超时，超时后停止转圈并给出可见错误，保留重试/切限价的出口。

[Claude session ↗](https://claude.ai/code/session_01Q2G6AAGbbxWPwAAECsfAZb)

---

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
```diff
@@ -254,8 +264,11 @@ const ClosePositionForm = memo(
         }
 
         if (formData.type === 'market') {
-          const latestMidPrice = midPrice;
-          if (!latestMidPrice || latestMidPrice === '0') {
+          const latestMidPrice =
+            await backgroundApiProxy.serviceHyperliquid.getMarketOrderReferencePrice(
```
sidmorizon, line 308: # 🟡 P2 一键全平未接新价源，缺价仓位被静默丢弃

## 问题

新的后台价源只接到了单仓关仓弹窗。`packages/kit/src/states/jotai/contexts/hyperliquid/actions.ts` 的 `closeAllPositions` 仍用 `getMidPrice`（直接读 `perpsAllMidsAtom`），拿不到 mid 的仓位在 `if (!tokenInfo || !midPrice)` 处 `console.warn` 后 `return null` 被 `.filter(Boolean)` 掉。

`closeAllPositions` 自身的 `withToast` 没有传 `actionType`，因此不产生任何提示；而内层 `ordersClose` 带 `EActionType.ORDERS_CLOSE`，会为实际提交的那部分弹成功 Toast。

## 影响

同样是本 PR 描述的冷启动缺价场景：`3` 个仓位缺 `1` 个价时只发出 `2` 笔 reduce-only 单，用户看到的却是一次成功提示，剩余仓位静默留仓；全部缺价时 `positionsToClose.length === 0`，只有一条 `console.warn`，对用户是彻底静默的 no-op。

## 建议

`closeAllPositions` 的 mid 解析同样改走 `serviceHyperliquid.getMarketOrderReferencePrice`，并在存在被丢弃仓位时显式报错（或至少给 `closeAllPositions` 补 `actionType`），不要静默过滤。

[Claude session ↗](https://claude.ai/code/session_01Q2G6AAGbbxWPwAAECsfAZb)

---

sidmorizon (commented):
Review

---

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
Minnzen: Agreed, that was a regression from making the mid `undefined`. Fixed in 90b68980cb: `PriceInput` gained `midPriceDisabled`, so the limit input stays editable during the cold-start window and only the Mid shortcut dims until a price arrives.

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
Minnzen: Agreed. Fixed in 90b68980cb: the dialog now polls `getMarketOrderReferencePrice(coin)` every 2s whenever the main-runtime atom has no valid mid, so the gate and the submit share one price source (background cache, then per-dex REST). If that source also returns nothing, the spinner stops and the existing "Unable to get current market price" toast shows once; Confirm stays disabled and the limit path (now editable) is the exit.

packages/kit/src/views/Perp/components/OrderInfoPanel/ClosePositionModal.tsx
Minnzen: Agreed. Fixed in 53727fa889: `closeAllPositions` falls back to `getMarketOrderReferencePrice` for any position the atom cannot price, and if any position is still unpriced it toasts and throws before submitting anything, so it never closes a subset behind a success toast.

originalix:

@Minnzen

Reviewed commit `b3b72a25fd61`. This report always reflects the latest reviewed changes.

### Review summary

Before: Perps close flows relied on the main-runtime ALL_MIDS atom, so cold-start gaps could disable an individual close and Close All could silently omit positions without a usable price. Rapid taps could also open or submit duplicate close dialogs. Favorites consumers did not consistently preserve the persisted cross-mode drag sequence, while the iOS header reservation and web fixed-column shadow could settle visibly after first paint.

After: market-close pricing is owned by a background per-dex resolver that accepts a recent background WebSocket snapshot or fetches a REST snapshot without overwriting the shared cache. Both individual market closes and every Close All item now resolve through that owner, and Close All aborts the batch when any price or symbol metadata is unavailable. Dialog and submission guards prevent duplicate orders. The persisted favorites sequence now drives the bar, ticker, and selectors until the user explicitly sorts the Favorites tab; iOS header fallback sizing and web shadow measurement are aligned before meaningful paint.

On native, dialogs, actions, selectors, and layout run in `main`, while `ServiceHyperliquid` and `hyperLiquidCache` run in `bg`. The background cache is a per-runtime JS instance, not a shared native singleton. ALL_MIDS and persisted atom data are deserialized independently into each runtime's JS heap, and `main`/`bg` initialize independently; the existing favorites persistence backend remains the shared native resource. Desktop and web use one JS runtime. Exchange order structure, reduce-only/slippage handling after reference-price selection, account/trading guards, signing, search, and non-Favorites sorting remain unchanged.

**What needs attention:** Confirm the localized failure UX and runtime behavior when the independently initialized native `main` and `bg` runtimes have no usable ALL_MIDS price during a close request.

### Issues to address

- **P2 · Close-price failure messages bypass localization**
  - **Impact:** When the background price resolver cannot provide a usable value during cold start or a disconnect, the newly reachable error branches pass the English literal `Unable to get current market price` directly to `Toast.error`; Close All duplicates the same literal. Non-English users therefore receive untranslated trading errors, and the copies can drift independently.
  - **Suggested action:** Add one approved close-price-unavailable translation key through the normal Lokalise workflow and format that key in all three new error call sites without editing generated locale files directly.

### Validation gaps

- The four targeted Jest files could not run because Yarn reported that the node_modules state file is missing. Static patch validation with `git diff --check` completed without errors.
- Existing GitHub review and inline-comment coverage could not be fetched from this workspace, so the proposed inline wording needs a human duplicate check.
- Native close-path verification should reproduce a cold start or ALL_MIDS disconnect where `main` has no usable price while `bg` initializes independently. Merely seeing the dialog or a cached price is not sufficient; passing means the background fallback supplies a current price or the whole close operation fails visibly without submitting any order.
- UI verification should also cover same-frame double taps, Favorites drag-order parity across mobile and desktop selectors, the iOS safe-area header on first layout, and a web table both with and without horizontal overflow. A settled shadow after interaction is not sufficient; passing means no incorrect first-paint shadow transition and correct shadow state after scrolling and resizing.

originalix (commented):
@Minnzen

Reviewed commit `137e197f4b3a`.

- **P1 · Close All bypasses the fresh-price gate**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13128#issuecomment-5520400014)

Minnzen:
Addressed the P1 in b3b72a25fd: `closeAllPositions` now prices every position through `getMarketOrderReferencePrice` (background snapshot under 5s, otherwise per-dex REST), the same freshness-checked owner the single close submit uses. The all-or-nothing abort for unpriced positions is unchanged.

originalix (commented):
@Minnzen

Reviewed commit `b3b72a25fd61`.

- **P2 · Close-price failure messages bypass localization**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13128#issuecomment-5520400014)

Minnzen:
On the P2: `Unable to get current market price` is a pre-existing literal in `ClosePositionModal` (alongside two other English toasts there); this PR only makes it reachable from two more places. Left as-is for the hot-update and marked with TODO(i18n) in 01ef996162; a dedicated key goes through Lokalise as a follow-up.

originalix (approved):

PR #13137, opened 2026-09-02
refactor: migrate native images away from expo-image

huhuanming:
## Summary

- Migrate the native ImageV2 implementation to the published `@onekeyfe/react-native-image@3.0.101` package behind native platform isolation.
- Upgrade every app-modules `@onekeyfe/*` dependency on the shared 3.x release line to 3.0.101; `@onekeyfe/react-native-ble-utils` remains on its independent 0.1.6 line.
- Upstream the existing `@onekeyfe/react-native-tab-view` patch plus the Android inactive-scene touch interception fix from `codex/android-tab-touch-diagnosis` to app-modules, then remove the monorepo patch.
- Preserve explicit native `resizeWidth` TOS optimization with raw-URL fallback, keep render/preload cache keys identical, preserve custom-header source identity, normalize React Native Web load events, and keep preload APIs boolean/non-rejecting across platforms.
- Preserve Android bundled-asset URI normalization and select responsive native sources by rendered size, DPR, and source scale.
- Restore the Spinner default icon color after the Tamagui 2 migration.
- Consume the app-modules fixes for rounded image/loading clipping and stable Android Glide memory-cache identity. Preserve blank-URI preload failure semantics through the app wrapper, and defer Android `onDisplay` while the native view is detached. No fixed 150 ms loading delay is used.

## Published packages

- app-modules PR: OneKeyHQ/app-modules#90
- Publish workflow: https://github.com/OneKeyHQ/app-modules/actions/runs/33715456651
- All 37 publishable app-modules packages are available as 3.0.101 under the `latest` dist-tag.

## Validation

- `yarn install` against the public 3.0.101 packages.
- `yarn agent:check --profile pr`: all local and latest GitHub checks passed; GitHub approval remains required.
- Native image JS Jest: 14/14; Android unit tests: 46/46.
- Android Example runtime: four identical image requests all reported memory-cache hits after unmount/remount.
- Android app `:app:installProdDebug`: passed and installed on the emulator.
- iOS CocoaPods install: passed; all app-modules pods resolve to 3.0.101 with no 3.0.99 lock entries.
- Spinner runtime verification: restored neutral icon color and native animation.
- iOS cache audit: managers share `SDImageCache.shared`; URL, header, size, and safety-generation cache-key inputs are value-stable, with no Android-equivalent identity regression.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

originalix:

@huhuanming

Reviewed commit `81fe5f2862ec`. This report always reflects the latest reviewed changes.

### Review summary

The image ownership boundary moves from expo-image and the JavaScript cache/useImage layer to a native OneKeyImage implementation on mobile, while web keeps a React Native Image adapter with equivalent optimization and fallback handling. Native cache preload/clear controls, loading placeholders, recycling keys, and image variants are added; token, DeFi, market, avatar, and cold-start callers are migrated. The old cache/useImage implementation and expo-image patch are removed, while unrelated transaction and wallet flows remain unchanged.

**What needs attention:** Verify that every supported build environment can resolve the native image packages and that the new native/web cache and fallback state machines preserve first-render, error, and preload behavior.

### Issues to address

- **P1 · Absolute portal dependencies make mobile installs non-reproducible**
  - **Impact:** apps/mobile/package.json and yarn.lock resolve @onekeyfe/react-native-image and @onekeyfe/react-native-skeleton from /Volumes/T7Shield/Project/app-modules-onekey-image. That path is machine-local, so a clean checkout cannot install or build the mobile workspace.
  - **Suggested action:** Replace the absolute portal locators with published package versions or a repository-relative, checked-in workspace dependency and regenerate the lockfile.
- **P2 · The image migration removes the existing loading test suite without replacement**
  - **Impact:** The 571-line useImage.test.tsx suite is deleted while the native implementation, cache behavior, callbacks, fallback handling, and preload paths are substantially rewritten. Regressions in first-frame loading, retries/fallbacks, or cache invalidation can therefore pass CI unnoticed.
  - **Suggested action:** Add focused platform-aware tests for the OneKeyImage adapters and preload/cache controls before removing the behavioral coverage.

### Validation gaps

No native iOS/Android install, build, or runtime image-cache validation is demonstrated. The absolute portal locators also prevent a reproducible clean-install check. The deleted useImage tests leave callback ordering, fallback transitions, preload success/failure, and cache clearing without equivalent automated coverage.

apps/mobile/package.json
```diff
@@ -81,6 +81,7 @@
     "@onekeyfe/react-native-check-biometric-auth-changed": "3.0.95",
     "@onekeyfe/react-native-cloud-kit-module": "3.0.95",
     "@onekeyfe/react-native-device-utils": "3.0.95",
+    "@onekeyfe/react-native-image": "portal:/Volumes/T7Shield/Project/app-modules-onekey-image/native-views/react-native-image",
```
sidmorizon: 🟠 P1: Machine-specific dependencies block clean builds

Severity: `severe`

Machines without the author's mounted drive cannot resolve the new `portal:` dependencies. Dependency installation and every clean mobile build stop before compilation.

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -126,33 +161,52 @@ export function ImageV2({
...
+      clearPlaceholderTimer();
       setHasError(false);
-      onLoad?.(event);
-      if (!isLoading) {
-        skeletonTimerRef.current = setTimeout(() => {
-          setIsLoading(true);
-        }, 150);
-      }
+      setIsPlaceholderVisible(false);
+      const { height, uri, width } = event.nativeEvent.source;
```
sidmorizon: # [🟠 P1] Web/Desktop 端 `onLoad` 解构 `event.nativeEvent.source` 会在每次加载成功时抛错

## 问题

`ImageV2.tsx` 是非 native 变体，web/desktop/ext 的 `react-native` 被 alias 到 `react-native-web@0.21.2`（`webpack.base.config.js:56`）。RNW 的 `ImageLoader.load` 传给 `onLoad` 的是 `{ nativeEvent: <DOM load Event> }`（`dist/modules/ImageLoader/index.js`），DOM Event 上没有 `source` 字段，这里解构 `undefined` 会直接抛 `TypeError`。RN 的 `ImageLoadEvent` 类型让它在编译期看不出来，仓库里的 `react-native-web+0.21.2.patch` 也只改了 `TextInput`。迁移前 web 走 expo-image，`onLoad` 事件带完整的 `source`。

```ts
const { height, uri, width } = event.nativeEvent.source; // source 在 RNW 下是 undefined
```

## 影响

- 每张图片加载成功都会在 `image.decode().then(onDecode, onDecode)` 里产生一个 unhandled `TypeError`。
- RNW 在同一回调里先调 `onLoad(e)` 再调 `onLoadEnd()`，异常后 `onLoad`、`onDisplay`、`onLoadEnd` 都不会传到业务层。本 PR 改造的 `TxConfirmHeaderRight.tsx` 依赖 `onLoad` 的 `source.width/height` 计算 MEV provider logo 宽度，desktop/web 上会永远停在 `DEFAULT_IMAGE_HEIGHT` 的正方形；`ShareContentRenderer`（Perp/Receive/RookieGuide）里等待 `onLoad` 的 ready 逻辑同样不会触发。

## 建议

web 变体不要假设 `nativeEvent.source` 存在，从 DOM 事件的 `target` 取自然尺寸并兜底：

```ts
const nativeEvent = event.nativeEvent as Partial<{ source: { width: number; height: number; uri: string } }> & {
  target?: HTMLImageElement;
};
const width = nativeEvent.source?.width ?? nativeEvent.target?.naturalWidth ?? 0;
const height = nativeEvent.source?.height ?? nativeEvent.target?.naturalHeight ?? 0;
const url = nativeEvent.source?.uri ?? nativeEvent.target?.currentSrc ?? resolvedUri ?? '';
```

并补一个 mock RNW 事件形状的 web `onLoad` 单测。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/preload.ts
```diff
@@ -39,47 +31,27 @@ function getPreloadUri(
...
-    return true;
-  }
-  const success = await Image.prefetch(missingUris, {
-    cachePolicy: DEFAULT_CACHE_POLICY,
-  });
-  if (success) {
-    await refreshCachedImagePaths(missingUris);
-  }
-  return success;
+  const results = await Promise.all(uris.map((uri) => Image.prefetch(uri)));
```
sidmorizon: # [🟠 P1] Web 端 `preloadImages` 契约变了：任一 URL 失败即 reject，成功时也永远返回 `false`

## 问题

迁移前 web 走 expo-image 的 `Image.prefetch`，它从不 reject，并以 boolean 返回结果。换成 `react-native-web` 的 `Image.prefetch` 后（`dist/modules/ImageLoader/index.js` 的 `prefetch`）：成功时 `resolve()` 不带值，所以 `results.every(Boolean)` 恒为 `false`；任意一个 URL 加载失败就 `reject`，`Promise.all` fail-fast 让整个 `preloadImages` 变成 reject。`IPreloadImagesFunc` 声明的 `Promise<boolean>` 契约因此在 web/desktop/ext 上被打破。

```ts
const results = await Promise.all(uris.map((uri) => Image.prefetch(uri))); // 成功 -> [undefined, ...]
return results.every(Boolean); // 永远 false
```

## 影响

- `DeFiContainer.tsx:174`（`void Image.preloadImages(fresh)`）、`tokenImagePrewarm.ts:64` 和 `marketDetailImagePreload.ts:53`（只有 `.then().finally()`）都没有 `.catch`，任一 logo `404` 就是 unhandled rejection。
- `tokenImagePrewarm` / `marketDetailImagePreload` 只有在返回 `true` 时才调用 `rememberPrewarmedUris`，现在永远不会执行，去重集合失效，每次进入页面都重复 prefetch 同一批 URL。

## 建议

把每个 prefetch 归一化为 boolean，避免 fail-fast：

```ts
const results = await Promise.all(
  uris.map((uri) =>
    Image.prefetch(uri).then(
      () => true,
      () => false,
    ),
  ),
);
return results.every(Boolean);
```

同时确认 `preload.native.ts` 里 `OneKeyImageCache.preload` 也遵守「不 reject、返回 boolean」的契约，否则同样的调用点在 native 上也会出现 unhandled rejection。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -1,36 +1,95 @@
...
+  if (typeof candidate === 'number') {
+    const resolved = ReactNativeImage.resolveAssetSource(candidate);
+    if (
+      Platform.OS === 'android' &&
+      resolved?.uri &&
+      !resolved.uri.includes(':')
+    ) {
+      return {
+        ...resolved,
+        uri: `android.resource://so.onekey.app.wallet/drawable/${resolved.uri}`,
```
sidmorizon: # [🟡 P2] Android 包名硬编码在 components 层的资源 URI 归一化里

## 问题

`normalizeSource` 在 Android release 下把 `resolveAssetSource()` 返回的无 scheme drawable 名拼成 `android.resource://so.onekey.app.wallet/drawable/...`，包名是字面量。`packages/shared/src/config/appConfig.ts` 已经导出 `ANDROID_PACKAGE_NAME`，`components` 允许依赖 `shared`，这里变成了第二份真源。另外 `@onekeyfe/react-native-image@3.0.99` 自身的 JS `normalizeSource` 也会走 `Image.resolveAssetSource()`，但不会补 scheme，所以直接使用 `OneKeyImage` 的调用方（如 `ImagePanel.native.tsx`）在 Android release 下 `require()` 图片仍会以无 scheme 的 `assets_x` 交给原生侧加载失败。

```ts
uri: `android.resource://so.onekey.app.wallet/drawable/${resolved.uri}`,
```

## 影响

`android.resource://<authority>` 由 `ContentResolver` 按已安装的 applicationId 解析。当前 `build.gradle` 没有 `applicationIdSuffix`，所以现在能工作；一旦某个 flavor 或白牌构建改了 applicationId，Android release 下所有 `require()` 图片会静默加载失败（Glide `onLoadFailed`，无 crash，很难排查），而且 `ANDROID_PACKAGE_NAME` 改了这里也不会跟着变。

## 建议

最小改动：改用 `ANDROID_PACKAGE_NAME`，并加一行注释说明为什么 release 下的 `resolveAssetSource` 只返回资源名：

```ts
import { ANDROID_PACKAGE_NAME } from '@onekeyhq/shared/src/config/appConfig';
// Release bundles loaded from assets:// resolve require() images to a bare drawable name
uri: `android.resource://${ANDROID_PACKAGE_NAME}/drawable/${resolved.uri}`,
```

更彻底的做法是把 scheme 补全下沉到 `@onekeyfe/react-native-image` Android 侧（`OneKeyImageModel.build` 检测无 scheme 时用 `context.packageName` 拼 `android.resource://`），JS 层不再感知包名，库的直接调用方也一并受益。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.tsx
sidmorizon: ## 补充：CI 已经复现这个问题

`4c545a0` 的 `Web startup graph budget (24.x)` 失败（[job 100292698007](https://github.com/OneKeyHQ/app-monorepo/actions/runs/33643638975/job/100292698007)），日志里每个带图片的 web 场景都出现 `invalid runs: pageErrors`：

| 场景 | 本 PR pageErrors | 同一 workflow 基线（[run 33643338034](https://github.com/OneKeyHQ/app-monorepo/actions/runs/33643338034/job/100291686407)） |
|---|---:|---:|
| market | `759` | `0` |
| defi | `72` | `0` |
| swap | `18` | `0` |
| perps | `3` | `0` |
| referFriends | `0` | `0` |

错误数量和页面上 token logo 的数量成正比，和「每张图 `onLoad` 成功时解构 `event.nativeEvent.source` 抛 `TypeError`」的路径一致；本 PR 在 web 端的运行时改动也只有 `ImageV2.tsx` / `preload.ts`。修掉这里之后 `market` 的 `resourceCount: 250 / 240` 是否仍超标再单独看。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/kit/src/views/Market/MarketDetailV2/utils/marketDetailImagePreload.ts
```diff
@@ -8,7 +8,7 @@ type IMarketTokenImageSource = {
 };
 
 const MARKET_TOKEN_IMAGE_PREWARM_LIMIT = 4;
-const MARKET_TOKEN_IMAGE_DECODE_TIMEOUT_MS = 350;
+const MARKET_TOKEN_IMAGE_PREWARM_RESIZE_WIDTH = getTokenImageResizeWidth('md');
```
sidmorizon: # [🟠 P1] 模块顶层调用 `getTokenImageResizeWidth('md')` 让 `useToMarketDetailPage.test.tsx` 整个 suite 无法加载

## 问题

这一行在 import 阶段就执行 `getTokenImageResizeWidth('md')`，把 `tokenSize.ts` 拉进了 `useToMarketDetailPage` 的依赖链，而 `tokenSize.ts` 在模块顶层对每个尺寸调用 `s()`。根 `jest.config.js` 的 `moduleNameMapper` 用的是非锚定的 `'@onekeyhq/components'`，会把 `@onekeyhq/components/src/utils/scale` 这类深路径也映射到同一个 `__mocks__/componentsMock.ts`；`useToMarketDetailPage.test.tsx` 又用 `jest.mock('@onekeyhq/components', factory)` 替换了它，factory 里没有 `s`，于是 `s(56)` 变成 `TypeError: (0, _scale.s) is not a function`。base `x` 上 `marketDetailImagePreload.ts` 不依赖 `tokenSize`，这条链是本 PR 引入的。

```ts
const MARKET_TOKEN_IMAGE_PREWARM_RESIZE_WIDTH = getTokenImageResizeWidth('md');
```

## 影响

`4c545a0` 的 `Unit Tests (3/5)` 红：`Test Suites: 1 failed, 263 passed`，失败的是 `useToMarketDetailPage.test.tsx`（[job 100292698759](https://github.com/OneKeyHQ/app-monorepo/actions/runs/33643639000/job/100292698759)），是 `Test suite failed to run`，不是断言失败，所以该文件里的所有用例都没跑。任何用同样方式 mock `@onekeyhq/components` 且间接引用 `tokenSize.ts` 的测试都会以同样方式挂掉。

## 建议

把求值延迟到调用时，避免 import 副作用（这也符合 `tokenImagePrewarm.ts` 里在函数内调用 `getTokenImageResizeWidth` 的写法）：

```ts
const MARKET_TOKEN_IMAGE_PREWARM_SIZE = 'md' as const;
// inside prewarmMarketTokenImages:
const resizeWidth = getTokenImageResizeWidth(MARKET_TOKEN_IMAGE_PREWARM_SIZE);
```

并在 `useToMarketDetailPage.test.tsx` 的 `jest.mock('@onekeyhq/components', ...)` factory 里补 `s: (value: number) => value`（或改为 `jest.requireActual` 扩展），确保 hook 内部真正调用到 `prewarmMarketTokenImages` 时也不会再抛。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -57,167 +116,93 @@ export function ImageV2({
...
     onLoadStart,
     onDisplay,
     autoplay,
-    resizeWidth,
-    ...imageProps
+    resizeMode,
+    contentFit,
+    cachePolicy,
+    recyclingKey,
+    resizeWidth: _resizeWidth,
```
sidmorizon: **Native resize hints are discarded**

Severity: `non-severe`

On iOS and Android main, `resizeWidth` is discarded, so images can fetch larger URLs than requested. Independent preloads can warm different shared-cache entries.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -1,36 +1,95 @@
...
+    return OneKeyImageContentFit.CONTAIN;
+  }
+  if (
+    contentFit === 'center' ||
+    resizeMode === 'center' ||
+    resizeMode === 'none'
+  ) {
+    return OneKeyImageContentFit.CENTER;
+  }
+  return OneKeyImageContentFit.COVER;
```
sidmorizon: **Explicit image fit loses precedence**

Severity: `non-severe`

On iOS and Android main, `getContentFit` lets legacy `resizeMode` override conflicting `contentFit`. The image then uses the wrong crop or aspect behavior.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -1,36 +1,95 @@
...
+}
+
+function normalizeSource(
+  source: IImageV2Props['source'] | undefined,
+): OneKeyImageProps['source'] {
+  if (typeof source === 'string') {
+    return { uri: source.trim() };
+  }
+  const candidate = Array.isArray(source) ? source[0] : source;
```
sidmorizon: **Responsive sources always use first image**

Severity: `non-severe`

On iOS and Android main, `normalizeSource` always takes the first source candidate. Responsive sets can display an unsuitable or low-resolution image.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -57,167 +102,93 @@ export function ImageV2({
...
     onLoadStart,
     onDisplay,
     autoplay,
-    resizeWidth,
-    ...imageProps
+    resizeMode,
+    contentFit,
+    cachePolicy,
+    recyclingKey,
+    resizeWidth: _resizeWidth,
```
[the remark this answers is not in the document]
huhuanming: Fixed in 22c945ee7d: native ImageV2 now applies the existing optimizer for resizeWidth or static dimensions, retries the raw URI on optimized-load failure, and disables native TOS rewriting for that raw retry.

packages/components/src/primitives/Image/ImageV2.tsx
huhuanming: Fixed in 22c945ee7d: the web handler now accepts both React Native source events and React Native Web DOM load events, using natural dimensions/currentSrc with the resolved URI as fallback.

apps/mobile/package.json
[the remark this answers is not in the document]
huhuanming: Fixed in 4c545a0fbf: both portal dependencies were replaced with published 3.0.99 packages and the lockfiles were regenerated.

huhuanming: Fixed in 4c545a0fbf: all machine-specific portal paths were removed and clean yarn install plus Android and iOS builds were verified against public 3.0.99 packages.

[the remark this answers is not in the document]
huhuanming: Fixed in 4c545a0fbf: image and skeleton now use published 3.0.99 versions with no local filesystem dependency.

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,37 @@
+import {
```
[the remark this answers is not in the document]
huhuanming: Fixed in 4c545a0fbf: the strict module-ID registry was regenerated. Validation reports 23,898 active modules, 0 tombstones, and both production main and background bundles succeed.

package.json
```diff
@@ -410,6 +410,7 @@
     "yaml": "^2.8.0"
   },
   "resolutions": {
+    "react-native-nitro-modules": "0.37.0",
```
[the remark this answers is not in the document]
huhuanming: Fixed in 4c545a0fbf: all app-modules packages on the shared 3.x line were published and upgraded to 3.0.99, aligning their Nitro 0.37 peer support. BLE remains on its independent 0.1.6 line.

packages/components/src/primitives/Image/preload.ts
huhuanming: Fixed in 22c945ee7d: web prefetch results are normalized to booleans without rejecting, native preload also converts bridge failures to false, and focused regression tests cover RNW success and failure behavior.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -161,63 +225,34 @@ export function ImageV2({
...
-      onLoad={onLoad}
+      placeholder={placeholderOverlay}
+      fallback={fallbackOverlay}
+      contentFit={getContentFit({ contentFit, resizeMode })}
+      cachePolicy={cachePolicy ? CACHE_POLICIES[cachePolicy] : undefined}
+      recyclingKey={recyclingKey}
+      autoplay={autoplay}
+      optimizeTos={!shouldUseRawSourceFallback}
+      loadingStrategy={OneKeyImageLoadingStrategy.SKELETON}
+      onError={onError ? handleError : undefined}
```
sidmorizon: # [🟠 P1] Native 端「优化 URL 失败回退原图」只在调用方传了 `onError` 时才生效

## 问题

`22c945e` 把 TOS 优化前置到 JS（第 160 行 `buildOptimizedImageSource`），并把「优化 URL 加载失败后回退到 `rawUri`」的逻辑放进了 `handleError`（第 220-227 行）。但这里写成 `onError={onError ? handleError : undefined}`，没传 `onError` 的调用方根本不会挂上 `handleError`。`Token.tsx` 的 `sharedImageProps` 就不传 `onError`，kit 内绝大多数 `<Image>` 也一样。

原生层自己的回退同时被绕过了：`@onekeyfe/react-native-image@3.0.99` 的 `OneKeyTosURL.swift` 把已带 `x-tos-` 参数的 URL 视为 protected 并原样返回，于是 `OneKeyImage.swift:287` 的 `mayFallbackToRaw = requestURL != rawURL` 为 `false`，原生不再重试原图；库的 JS 层也只在 `refs.current.onError != null` 或有 overlay 时才转发错误（`src/index.tsx:239`），即便有 `fallback` overlay，转发到的也不是这里的 `handleError`。

```tsx
onError={onError ? handleError : undefined}
```

## 影响

iOS/Android 上，只要 TOS 缩放变体加载失败（对象不支持 resize、变体 `4xx` 等），未传 `onError` 的图片会永久停在 fallback/错误态，而不是像上一版（原生构建 TOS URL 并自带回退）那样回退加载原图。这是本次提交相对 `4c545a0` 的行为回退。

## 建议

无条件挂载 `handleError`，在其内部再转发给调用方：

```tsx
onError={handleError}
// handleError 内部保持现有逻辑：优化失败先 setRawSourceFallbackUri，否则 onError?.(event)
```

`OneKeyImage` 的 `needsError` 会因为 `refs.current.onError != null` 自动开启原生错误回调，不需要再改库。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,37 @@
...
+        headers: source.headers,
+        cachePolicy: source.cachePolicy
+          ? CACHE_POLICIES[source.cachePolicy]
+          : OneKeyImageCachePolicy.MEMORY_DISK,
+        resizeWidth: source.resizeWidth ?? source.width,
+        resizeHeight: source.height,
+        pixelRatio:
+          source.pixelRatio ?? options?.pixelRatio ?? PixelRatio.get(),
+        overscan: source.overscan,
+        optimizeTos: source.optimize !== false,
```
sidmorizon: # [🟡 P2] iOS 预加载与渲染生成的 TOS URL 编码不一致，预热缓存命中不了

## 问题

`22c945e` 之后，native 渲染路径在 JS 里预先构建 TOS URL（`ImageV2.native.tsx:160` → `buildTosImageResizeUrl`），产物形如 `?x-tos-process=image%2Fresize%2Cw_N`（`optimization.test.ts:22` 也是这么断言的）。而这里的 `preload.native.ts` 仍把原始 `uri` 交给原生 `OneKeyImageCache.preload` 并让它 `optimizeTos`，iOS 用 `URLComponents.queryItems` 拼出的是 `?x-tos-process=image/resize,w_N`（`OneKeyTosURL.swift:38-40`，Foundation 不会对 `/` 和 `,` 做百分号编码）。SDWebImage 的缓存 key 就是 `url.absoluteString`（`OneKeyImageRequestContext.swift:758`），两者不相等。Android 的 `OneKeyTosUrl.kt:40` 固定输出 `%2F%2C`，与 JS 一致，所以只有 iOS 受影响。

```ts
optimizeTos: source.optimize !== false, // iOS 原生 builder 与 JS builder 编码不同
```

## 影响

iOS 上 `coldStartImagePreload` / `tokenImagePrewarm` / `marketDetailImagePreload` 预热进内存/磁盘缓存的条目，渲染时一律 miss，图片会再下载一次，CDN 侧也会产生两份变体。预热在 iOS 上实际失效，且比不预热多一次请求。`4c545a0` 时两条路径都走原生 builder，没有这个问题。

## 建议

让两条路径只用一种 builder。最小改动是在 `preload.native.ts` 里也用 JS 的 `buildTosImageResizeUrl` 算好 `uri`，再传 `optimizeTos: false`：

```ts
const optimized = buildTosImageResizeUrl({ uri: source.uri, resizeWidth, ... });
return { uri: optimized ?? source.uri, optimizeTos: false, ... };
```

或者反过来，渲染路径不做 JS 预优化，只把 `resizeWidth` 作为尺寸提示交给原生。关键是缓存 key 只能有一种生成方式。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in 9b13d00545. ImageV2 now uses the shared ANDROID_PACKAGE_NAME constant and documents why release require() assets need the Android resource URI normalization.

packages/kit/src/views/Market/MarketDetailV2/utils/marketDetailImagePreload.ts
huhuanming: Fixed in 9b13d00545. The resize width is evaluated only when there are URIs to prewarm, and the focused hook test mock now provides s(). The previously unloadable suite passes 8/8 locally.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed by 22c945ee7d. Native ImageV2 now applies resizeWidth/static display-size hints through buildOptimizedImageSource, with raw-URI fallback on optimized load failure.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in 9b13d00545. Explicit contentFit is now resolved first; legacy resizeMode is used only when contentFit is absent.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in 9b13d00545. Native ImageV2 now selects responsive candidates by source area against the rendered size and device DPR, then falls back to the closest scale candidate.

packages/components/src/primitives/Image/ImageV2.native.tsx
sidmorizon, line 1: **Optimized image failures skip fallback**

Reported at `packages/components/src/primitives/Image/ImageV2.native.tsx:251` (outside the diff of this PR, so this is a file-level comment)

Severity: `non-severe`

Without an `onError` callback, `OneKeyImage` receives no handler for failed pre-optimized URLs. Valid raw images then remain blank or show fallback content.

Suggested fix:

```ts
      onError={handleError}
```

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in 77a8be6719. ImageV2 now always supplies handleError, so optimized URL failures trigger raw-source fallback even when the caller did not provide onError.

packages/components/src/primitives/Image/preload.native.ts
huhuanming: Fixed in 77a8be6719. Native preload now builds the TOS URL with the shared JS builder and disables native URL rewriting, so iOS render/preload cache keys match. Added a focused regression assertion for the encoded URL.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Already fixed in 77a8be6719 on the current head: OneKeyImage now receives onError={handleError} unconditionally, preserving raw-source fallback without a caller callback.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -1,36 +1,171 @@
...
+    let bestSource: ImageURISource | undefined;
+    let bestPrecision = Number.POSITIVE_INFINITY;
+
+    sources.forEach((source) => {
+      const sourceWidth = getPositiveNumber(source.width);
+      const sourceHeight = getPositiveNumber(source.height);
+      if (!sourceWidth || !sourceHeight) {
+        return;
+      }
+      const precision = Math.abs(1 - (sourceWidth * sourceHeight) / targetArea);
```
sidmorizon: # [🟠 P1] 响应式候选源选择忽略了 `scale`，@1x/@2x/@3x 候选会恒选第一个

## 问题

`selectSourceCandidate` 里 `targetArea = displayWidth * displayHeight * pixelRatio²` 是像素面积，但 `sourceWidth * sourceHeight` 是 RN 语义下的 point 面积（`ImageURISource.width/height` 为 point，`scale` 为密度）。RN 原生 `RCTImageView.mm:274` 的同一算法算的是 `imgSize.width * imgSize.height * source.scale * source.scale`，这里漏掉了 `scale²`。对于典型输入 `[{width:100,height:100,scale:1}, {…,scale:2}, {…,scale:3}]`，三者 `precision` 完全相同，`<` 严格比较让 `bestSource` 始终停在第一个（最低清）候选。

```ts
const precision = Math.abs(1 - (sourceWidth * sourceHeight) / targetArea);
```

## 影响

只要候选带 `scale`，3x 设备也会请求 1x 资源，图片模糊；`scale` 越高的候选越不可能被选中，`9b13d00` 新增的响应式选择在最常见输入下失效。仓库内目前还没有数组 `source` 的调用方，属潜在缺陷，但这正是这段新代码的主用例。

## 建议

面积比较时计入候选 `scale`，与 RN 保持一致（无 `scale` 的像素尺寸候选行为不变）：

```ts
const scale = getPositiveNumber(source.scale) ?? 1;
const precision = Math.abs(
  1 - (sourceWidth * sourceHeight * scale * scale) / targetArea,
);
```

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,52 @@
...
+          uri: optimizedSource.uri ?? source.uri,
+          headers: source.headers,
+          cachePolicy: source.cachePolicy
+            ? CACHE_POLICIES[source.cachePolicy]
+            : OneKeyImageCachePolicy.MEMORY_DISK,
+          resizeWidth: source.resizeWidth ?? source.width,
+          resizeHeight: source.height,
+          pixelRatio,
+          overscan: source.overscan,
+          optimizeTos: false,
```
sidmorizon, line 102: # [🟡 P2] preload 固定 `optimizeTos: false` 后丢失了原生的原图回退

## 问题

`77a8be6` 把 JS 已改写的 TOS URL 作为 `uri`（第 36 行）交给原生并固定 `optimizeTos: false`。但 `@onekeyfe/react-native-image@3.0.99` 的原生 preload 只有在请求 URL 与原始 URL 不同时才回退原图：iOS `OneKeyImageCache.swift` 是 `if case .failure = result, requestURL != rawURL { retry rawURL }`，Android `OneKeyImageCache.kt` 在 `requestUrl == source.uri` 时直接返回 `false`。现在两者恒相等，TOS resize 失败时整个条目直接失败，不再尝试原图。上一版（`9b13d00`，原生自行 optimize）是有这层回退的。

```ts
optimizeTos: false, // uri 已是改写后的 TOS URL，原生 requestURL == rawURL，不再回退
```

## 影响

- 一批 `preloadImages` 里任一 URL 的 resize 变体失败（对象不支持 resize、变体 `4xx`）都会让整批返回 `false`；`tokenImagePrewarm` / `marketDetailImagePreload` 只在 `success` 时 `rememberPrewarmedUris`，同一批图片会在每次触发时反复重新请求。
- 渲染路径通过 `ImageV2.native.tsx` 的 `handleError` 回退到原图 URL，但该 URL 从未被预热，这类图片的预热失去意义。

## 建议

在 JS 侧补上和渲染路径一致的回退：`OneKeyImageCache.preload` 返回 `false` 且本批存在 `optimizedSource.optimized === true` 的条目时，用 `{ uri: source.uri, optimizeTos: false }` 对这些条目再 preload 一次；或者按条目拆分调用，让单个失败不影响整批的 `rememberPrewarmedUris`。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -161,50 +231,61 @@ export function ImageV2({
...
     ],
   );
 
-  const ImageComponent = useMemo<ComponentType<any>>(() => {
-    if (animated) {
-      return AnimatedExpoImage;
-    }
-    return ExpoImage;
-  }, [animated]);
+  const ImageComponent = ReactNativeImage as ComponentType<any>;
```
sidmorizon, line 326: # [🟠 P1] Web 端 Image 从 lazy `` 变为无条件 eager 加载，冷启动资源数超预算

## 问题

base 上 web 端走 expo-image，其 web 实现默认给 `` 加 `loading="lazy"`（`ExpoImage.web.tsx:95-96`：`loading ?? (responsivePolicy == null ? 'lazy' : undefined)`，仓库内没有任何调用方传 `loading` / `responsivePolicy`），未进入视口的 token / 网络 / banner 图片不会发起请求。这里改为 react-native-web 的 `Image` 后，RNW 在 `useEffect` 里对每个已挂载的 `source.uri` 无条件执行 `ImageLoader.load`（`new window.Image(); image.src = uri`，`dist/exports/Image/index.js:225-236`），并同时渲染隐藏 `` 与 CSS `background-image`，RNW 本身没有任何可视区或懒加载门控。

```ts
const ImageComponent = ReactNativeImage as ComponentType<any>; // RNW Image: 挂载即加载，无 lazy
```

## 影响

`22c945e` 修掉 `onLoad` 抛错后，`Web startup graph budget (24.x)` 仍红：root 与 market 场景的 `resourceCount` 从基线的 `202` / `201` 涨到 `250` / `250`（预算 `240`），scripts 数不变（`134`），增量全是图片；perps / swap / defi / referFriends 与基线完全一致。`250` 恰好是 `PerformanceResourceTiming` 的默认缓冲上限（`run-web-cold.js:624` 直接读 `getEntriesByType('resource')`，未调用 `setResourceTimingBufferSize`），所以真实增量 `≥48`。CI 的 unique 去重按 `origin+pathname` 归一化（`run-web-cold.js:220-223`），说明增量是不同的图片资源，不是 `x-tos-process` 桶位或 raw 回退造成的重复。Market 桌面页同时挂载 20 行 token logo + 网络角标、网络筛选栏 logo、分类图标和 banner，全部在挂载瞬间发起请求。

## 建议

1. web 端为 `ImageV2` 增加可视区门控（`IntersectionObserver` / 现有 `useVisible` 类 hook），只在接近视口时才把 `source` 交给 RNW `Image`；或在 web 上直接渲染 `` 而不经由 RNW `Image`。
2. 如果接受 eager 行为，需要同步在 `development/perf-ci/thresholds/web.cold.json` 明示调整 root / market 的 `resourceCount` 预算，并在采集处调用 `performance.setResourceTimingBufferSize`，否则 `250` 上限会掩盖真实资源数。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/kit/src/components/GlobalJotaiReady/GlobalJotaiReady.tsx
```diff
@@ -90,9 +89,7 @@ async function waitForJotaiReadyOnWebOrDesktop(): Promise<void> {
 }
 
 async function primeColdStartImagesBeforeRender(): Promise<void> {
-  await prewarmColdStartImagesFromSnapshot({
-    primeTimeoutMs: COLD_START_IMAGE_PRIME_TIMEOUT_MS,
-  });
+  await prewarmColdStartImagesFromSnapshot();
```
sidmorizon, line 92: 🟠 P1: Image preloads can block startup

Severity: `severe`

With cached remote images, `primeColdStartImagesBeforeRender` waits for every native preload without a timeout. A stalled request keeps the app blank indefinitely.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -112,47 +259,38 @@ export function ImageV2({
...
-  }, []);
-
-  useResetError(activeSource ?? null, hasError, onResetError);
-
+    [onLoad],
+  );
   const handleError = useCallback(
-    (event: ImageErrorEventData) => {
+    (event: OneKeyImageErrorEvent) => {
       if (
```
sidmorizon, line 341: **Failed images load twice**

Severity: `non-severe`

After OneKeyImage exhausts its built-in raw fallback, `handleError` starts the same raw request again. Broken images incur duplicate network requests and errors.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -57,46 +192,58 @@ export function ImageV2({
...
+    resizeMode,
+    contentFit,
+    cachePolicy,
+    recyclingKey,
     resizeWidth,
-    ...imageProps
+    blurRadius: _blurRadius,
+    capInsets: _capInsets,
+    defaultSource: _defaultSource,
+    fadeDuration: _fadeDuration,
```
sidmorizon, line 226: **Share captures lose fade suppression**

Severity: `non-severe`

On Android, `fadeDuration` is discarded before rendering. Receive-share captures can snapshot token, network, or logo images before their display transition finishes.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in `728c1a8034`: responsive candidate area now includes `scale²`, matching React Native source-selection semantics.

packages/components/src/primitives/Image/preload.native.ts
huhuanming: Fixed in `728c1a8034`: native preload first uses the JS-encoded optimized URLs, then retries the full batch with original URLs when the optimized batch fails. The boolean/non-rejecting contract is preserved, and a focused regression test covers the fallback.

packages/components/src/primitives/Image/ImageV2.tsx
huhuanming: Fixed in `7be7dd9589`: web ImageV2 now defers offscreen remote sources with `IntersectionObserver`. A local rerun of the exact web cold-start budget scenarios passed with root `resourceCount=194/240` and market `resourceCount=193/240` (FCP 224 ms for both).

packages/kit/src/components/GlobalJotaiReady/GlobalJotaiReady.tsx
huhuanming: This path does not await the network preload. `prewarmImageUris` starts `preloadImages(sources)`, attaches a rejection handler when `awaitPreload` is left at its default `false`, and immediately returns the source count. `primeColdStartImagesBeforeRender` therefore waits only for scheduling, not for native requests to settle, so a stalled image request cannot keep the placeholder mounted.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Clarified in `728c1a8034`: when JS has already produced an optimized URL, ImageV2 now explicitly passes `optimizeTos={false}`. The native optimizer therefore does not attempt or exhaust a second raw fallback; `handleError` performs the single raw-URL retry. For non-JS-optimized raw URLs, native optimization/fallback remains enabled.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: The replacement has no display fade to suppress. Android loads through a Glide `CustomViewTarget` and calls `hostView.setImageDrawable(resource)` directly without a transition factory/cross-fade; iOS uses SDWebImage render options without a forced transition. Therefore dropping the legacy `fadeDuration={0}` prop preserves the required immediate display behavior for receive-share captures.

packages/components/src/primitives/Image/ImageV2.tsx
sidmorizon: ## 复核 `7be7dd9`：web 场景已修，Desktop / Extension 仍是急加载

`IntersectionObserver` 门控只在 `platformEnv.isWeb` 时启用（`shouldLoadImage` 初始值 `!platformEnv.isWeb`，effect 里 `!platformEnv.isWeb` 直接 return）。`isWeb` 只在 `ONEKEY_PLATFORM === 'web'` 时为 true（`packages/shared/src/buildTimeEnv.js:9`），而 Electron 桌面端和浏览器插件走的是同一份 `ImageV2.tsx` + react-native-web，会以 `shouldLoadImage = true` 起步，所有已挂载图片仍在 mount 时立即 `ImageLoader.load`。基线的 expo-image 在所有 DOM 平台上默认 `loading="lazy"`，所以桌面 / 插件相对基线仍是回归，只是 CI 的 web cold-start 没有覆盖到。

建议把门控条件改成运行时判断（例如 `!platformEnv.isNative` 或 `platformEnv.isRuntimeBrowser`），让 desktop / ext / web-embed 与 web 走同一条延迟路径；IO 不存在时已有 `typeof IntersectionObserver === 'undefined'` 兜底。这条 thread 先保持 open，等这部分也覆盖后再关。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -30,12 +34,65 @@ const fullSizeStyle = {
...
+    }
+
+    const observer = new IntersectionObserver(
+      (entries) => {
+        if (entries.some((entry) => entry.isIntersecting)) {
+          setShouldLoadImage(true);
+          observer.disconnect();
+        }
+      },
+      { rootMargin: '200px 0px' },
```
sidmorizon: # [🟡 P2] `rootMargin: '200px 0px'` 水平方向为 0，横向滚动内容会在滑入时才开始请求

## 问题

`7be7dd9` 新增的 IntersectionObserver 只在纵向预留 `200px`，横向 margin 为 `0`。web 端 Carousel 是横向 `ScrollView` 且所有页同时挂载（`packages/components/src/composite/Carousel/pager.tsx`），下一页位于视口右侧、与 root 水平不相交，`source` 一直是 `undefined`；自动轮播开始滚动、露出第一个像素后才触发加载。Market 顶部横向的网络筛选列表同理。基线的 `loading="lazy"` 使用浏览器自己的距离阈值，两个方向都会提前加载。

```ts
{ rootMargin: '200px 0px' }
```

## 影响

首页 / Market banner 每次自动切页时，新页面在过渡动画里先是空白（或 `150ms` 后出现 skeleton），图片到达后再弹入；横向滚动网络筛选时 logo 逐个 pop-in。这是本次改动新引入的可见抖动，基线没有。

## 建议

`rootMargin` 四个方向都预留，例如 `'200px'` 或横向更大的 `'200px 400px'`。横向多预加载一屏对请求数几乎没有影响，但能保留轮播过渡时图片已就绪的体验。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

originalix (commented):
@huhuanming

Reviewed commit `7be7dd958979`.

- **P1 · Preload treats failed Image.prefetch results as success**
- **P2 · Behavioral image-loading tests were deleted without replacement**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13137#issuecomment-5510058839)

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,69 @@
...
+  const requests = preloadRequests.map(({ request }) => request);
+
+  return OneKeyImageCache.preload(requests)
+    .catch(() => false)
+    .then((success) => {
+      if (success || !preloadRequests.some(({ optimized }) => optimized)) {
+        return success;
+      }
+      return OneKeyImageCache.preload(
+        preloadRequests.map(({ request, rawUri }) => ({
```
sidmorizon: # [🟡 P2] preload 失败后整批按原图 URL 重试，已成功的条目也会被重新下载

## 问题

`728c1a8` 新增的回退逻辑里，native `OneKeyImageCache.preload` 只返回一个整体 boolean，这里一旦为 `false`，就把 `preloadRequests` 里的**全部**条目改成 `rawUri` 重新提交。已经成功的 TOS 优化条目会再按未裁剪的原图 URL 下载一遍（全尺寸字节，且渲染路径只在自身 fallback 时才用这个 key，缓存里基本是死数据）；`optimized: false` 的条目（非 TOS host / `optimize: false`）则用完全相同的 URL 原样重发。原生自身的 raw 重试是逐条只针对失败源的（`OneKeyImageCache.swift:132-138`、`OneKeyImageCache.kt:69-83`）。

```ts
preloadRequests.map(({ request, rawUri }) => ({ ...request, uri: rawUri }))
```

## 影响

冷启动批次上限 `96` 条（`coldStartImagePreload.ts:45`）、Perps 选择器 `72` 条、DeFi 首页无上限（`defiPositionUtils.ts:790+`）：任意一条失败（例如某个第三方协议 logo `404`）都会触发整批全尺寸原图重下，占用启动期带宽和磁盘缓存；对非 TOS 失败源该重试根本无效，最终仍返回 `false`。新增测试只覆盖单条批次，混合批次未覆盖。

## 建议

任选其一：

1. 让 native `preload` 返回逐条结果（`Promise<boolean[]>`），JS 只对 `optimized && failed` 的条目按 `rawUri` 重试。
2. 保持现有 native API，把 `optimized` 条目逐条（或小并发）调用 `OneKeyImageCache.preload([request])`，失败时只重试该条；非优化条目仍合并为一批。

最低限度也应把重试范围限制在 `optimized` 条目，并补一个混合批次测试，断言未优化 / 已成功条目不会被重发。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/ImageV2.tsx
huhuanming: Fixed in `f72a474445`: the observer now uses a four-direction `200px` root margin, so horizontally adjacent carousel/filter images begin loading before they enter the viewport.

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -30,12 +34,65 @@ const fullSizeStyle = {
...
+    }
+
+    const observer = new IntersectionObserver(
+      (entries) => {
+        if (entries.some((entry) => entry.isIntersecting)) {
+          setShouldLoadImage(true);
+          observer.disconnect();
+        }
+      },
+      { rootMargin: '200px' },
```
sidmorizon: # [🟡 P2] `rootMargin` 对嵌套 ScrollView 内的图片不生效，横向预取在实际布局里仍是 no-op

## 问题

`f72a474` 把 `rootMargin` 改成四向 `'200px'`，但 IntersectionObserver 的 `rootMargin` 只扩展 root（这里是 viewport）矩形；按规范的 intersection 算法，target 会先被所有 `overflow != visible` 的中间祖先裁剪，再与 root 求交。react-native-web 的 `ScrollView` 渲染为 `overflow-x: auto; overflow-y: hidden`（`horizontal`）或反向（`dist/exports/ScrollView/index.js:621-627`），位于滚动容器可视边缘之外的子元素交集恒为 0，`rootMargin` 完全不起作用。例如 `WalletBanner.tsx:387-405` 横向 `ScrollView` 里的 `<Image size={56} />`（`:121`）、Carousel pager 和 Market 网络筛选条，都只有滑入容器可视区后才开始加载，与改前 `'200px 0px'` 的行为一致。纵向在 `ScrollView` 内滚动的页面同样受影响（此前已存在）。

```ts
{ rootMargin: '200px' }, // 只扩展 viewport 矩形，不穿透 overflow 裁剪的滚动容器
```

## 影响

commit 声称的 "prefetch horizontal web images" 在实际布局中不成立：轮播自动切页 / 横向列表快速滑动时依然先空白或 skeleton，图片到达后再弹入；纵向 200px 预取在 web/desktop 上以 `ScrollView` 为滚动容器的页面同样无效。无回归，但修复未达预期。

## 建议

二选一：

1. observer options 里同时传 `scrollMargin: '200px'`（Chromium 133+ 支持，对嵌套滚动容器的裁剪矩形生效；不支持的浏览器会忽略该字段，无副作用）。
2. 观察时把 `root` 设为最近的滚动祖先（向上查找 `overflow-x/y` 为 `auto|scroll` 的元素），让 `rootMargin` 作用在真正的滚动容器上。

另建议补一行注释说明该限制，避免后续再把 `rootMargin` 当作对嵌套滚动容器有效。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

packages/components/src/primitives/Image/preload.native.ts
huhuanming: Fixed in `f8b44fa584`. Passthrough requests are issued once as their own batch. Optimized requests are batched on the success path; only after a batch failure are they bisected to isolate failing entries, and only a failing single entry is retried with its raw URL. Successful probes reuse the already populated optimized cache entry instead of downloading the raw asset. The new mixed-batch regression test covers two optimized entries plus one passthrough entry and verifies that only the failed optimized entry receives a raw retry.

packages/components/src/primitives/Image/ImageV2.tsx
huhuanming: Fixed in `12adef7a66`: the observer now sets both `rootMargin` and `scrollMargin` to `200px`. `scrollMargin` expands the clipping bounds of nested web ScrollViews, while browsers that do not recognize the dictionary member safely ignore it.

originalix (commented):
@huhuanming

Reviewed commit `f72a474445ff`.

- **P1 · Preload treats failed Image.prefetch results as success**
- **P2 · Behavioral image-loading tests were deleted without replacement**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13137#issuecomment-5510058839)

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,106 @@
...
+    return OneKeyImageCache.preload([
+      {
+        ...request,
+        uri: rawUri,
+      },
+    ]).catch(() => false);
+  }
+
+  const middleIndex = Math.ceil(preloadRequests.length / 2);
+  const results = await Promise.all([
```
sidmorizon: # [🟡 P2] 二分回退在失败路径上把网络请求放大约 log2(n) 倍，并绕开原生并发上限

## 问题

`f8b44fa` 的 `preloadWithRawFallback` 只拿到整批一个 boolean，于是对失败批次做二分：每一层都把包含失败项的子集整批再调一次 native preload。原生两端都不会对失败 URL 做负缓存：iOS 的 `preloadOptions` 显式带 `.retryFailed`（`OneKeyImageRequestContext.swift:668-670`，关闭了 SDWebImage 的 failedURLs 黑名单），Android Glide 本身没有失败缓存。因此失败的 optimized URL 在二分路径的每一层都会真实地再打一次网络；成功条目虽命中缓存，也要在每一层再走一次磁盘读 + 解码。另外两半和 passthrough 批次是并行的，叶子层会有接近 n 个并发 native 调用，iOS 的 `maximumConcurrentPreloads = 4`（`OneKeyImageCache.swift:47`）是按单次调用计数，实际被绕开。

```ts
const results = await Promise.all([
  preloadWithRawFallback(preloadRequests.slice(0, middleIndex)),
  preloadWithRawFallback(preloadRequests.slice(middleIndex)),
]);
```

以冷启动上限 96 条（`coldStartImagePreload.ts:45`）为例（96→48→24→12→6→3→2→1）：
- 1 条失败：native 调用 14-16 次（`f72a474` 为 2 次），该 URL 的 optimized 请求发 7-8 次 + raw 1 次，成功条目额外约 96 次缓存查找。
- 整批失败（离线冷启动 / TOS 或 CDN 故障 / 一组不可缩放资源）：约 191 次二分调用 + 96 次 raw ≈ 287 次 native 调用、约 860 次网络尝试（`f72a474` 为 192 次）。

## 影响

fallback 正是为"optimized URL 会失败"的场景设计的，而在这个场景下重试代价反而高于 `f72a474` 的整批重试：失败 URL 被反复请求、成功条目被反复查缓存解码，并发也不再受原生限制。离线或 CDN 故障时冷启动预热大列表尤为明显。功能上正确，只是代价被放大。

## 建议

任选其一，都能做到"每个 URL 最多 1 次 optimized + 1 次 raw，成功项不重复加载"：

1. 让原生 `preload` 返回 per-source 的 `boolean[]`（原生内部本来就是逐条得到结果的），JS 只对失败的 optimized 条目再发一次 raw 批次。
2. 纯 JS：首批失败后不二分，直接逐条调用（成功项命中缓存，失败项各 1 次 optimized + 1 次 raw），并在 JS 侧加一个小的并发上限（如 4）。

[Claude session ↗](https://claude.ai/code/session_01RbtMkaYa6cHr69boLGCDL1)

---

huhuanming: Fixed in `749907d5e2`: binary probing is removed. Every source now has one primary preload call under a JS concurrency cap of 4; only an optimized source whose own call fails receives one raw-URL retry. Passthrough sources are never retried. This keeps total active native calls bounded and limits each URL to one optimized attempt plus, when needed, one raw attempt.

huhuanming:
Addressed the findings from review https://github.com/OneKeyHQ/app-monorepo/pull/13137#pullrequestreview-5092220448 in `931afe5134`.

- `preloadImages` now preserves a fulfilled `false` result while continuing to treat React Native Web's fulfilled `undefined` as success and rejected prefetches as failure. A focused Web regression test covers the fulfilled-`false` case.
- No new Native `ImageV2` integration test is added in app-monorepo. The removed `useImage` implementation no longer exists, and the Native image implementation and behavior are owned and tested in `@onekeyfe/react-native-image` / app-modules#90 rather than duplicated here.

Validation: focused preload suite 6/6 passed; `yarn agent:check --profile commit` passed.

originalix:

@huhuanming

Reviewed commit `931afe513433`. This report always reflects the latest reviewed changes.

### Review summary

Before: native images used expo-image plus a JS loader, decoded-image cache, path cache, retries, and Android-specific autoplay/OOM safeguards. Prewarm paths populated that JS cache and prefetched through expo-image.

After: native rendering, preloading, and cache clearing move to @onekeyfe/react-native-image, while web/desktop use React Native Image with deferred offscreen loading and JS TOS optimization. Cold-start, token, and market prewarm now pass resize hints to platform preload APIs instead of priming the removed JS cache.

Supporting work includes the skeleton-to-placeholder API migration, removal of Image.loadImage sizing in the MEV header, SDWebImage pods, the 3.0.99 native-module/nitro upgrade, patch removal, and a native Image QA panel. Wallet signing flows and unrelated Expo image packages remain unchanged.

**What needs attention:** Verify native OneKeyImage defaults and source-identity handling on Android/iOS, especially animated images and authenticated image sources, before release.

### Issues to address

- **P2 · Android autoplay OOM guard no longer applied in ImageV2**
  - **Impact:** ImageV2.native now forwards autoplay without the previous Android-off default, while the expo-image patch that imposed a 16MB decode ceiling is removed. Dense Android GIF/WebP lists can therefore depend on unverified native defaults for memory safety.
  - **Suggested action:** Restore the wrapper default of autoplay disabled on Android (for example, autoplay={autoplay ?? Platform.OS !== 'android'}) and verify an equivalent native decode-memory bound.
- **P2 · Native TOS optimization bypasses custom source identity**
  - **Impact:** buildOptimizedImageSource deliberately marks sources with custom headers as unoptimized, but ImageV2.native then sets optimizeTos=true for every unoptimized source. The native layer can consequently rewrite a header-authenticated URL, invalidating request semantics or changing cache identity.
  - **Suggested action:** Disable optimizeTos whenever the source has custom headers (and apply the same guard in preload.native.ts) so those requests use the original URI unchanged.

### Validation gaps

- No native device run was available for Android animated-image lists, iOS/Android cache clearing, or first-frame prewarm timing.
- The installed workspace did not contain the `@onekeyfe/react-native-image` implementation, so its autoplay defaults, cache-key behavior, and decode limits could not be independently confirmed.
- The large `useImage` behavior suite was deleted; replacement coverage focuses on preload and optimization rather than the full render/error/fallback matrix.
- GitHub review-thread coverage was not available for duplicate-comment confirmation, so inline candidates are marked for human review.

packages/components/src/primitives/Image/preload.native.ts
```diff
@@ -0,0 +1,109 @@
...
+      const pixelRatio =
+        source.pixelRatio ?? options?.pixelRatio ?? PixelRatio.get();
+      const optimizedSource = buildTosImageResizeUrl({
+        uri: source.uri,
+        resizeWidth: source.resizeWidth,
+        displayWidth: source.width,
+        displayHeight: source.height,
+        pixelRatio,
+        enabled: source.optimize !== false,
+        overscanRatio: source.overscan,
```
sidmorizon, line 88: **Authenticated preloads miss rendered images**

Severity: `non-severe`

With request headers, `preloadImages` rewrites the native URL while rendering keeps it raw. The web path also drops headers, so preloading fails or warms the wrong cache entry.

huhuanming: Fixed in `77f248e8e0`.

- Native preload now disables JS-side TOS rewriting whenever the source has custom headers, so it sends the original URI together with the headers and warms the same cache identity used by rendering.
- Native rendering now also explicitly passes `optimizeTos={false}` for custom-identity sources.
- Web cannot attach headers through React Native `Image.prefetch`, so it skips those requests instead of issuing an unauthenticated preload, continues preloading public entries in a mixed batch, and returns `false` to report the incomplete preload.

Added a focused Web regression test; `yarn agent:check --profile commit` passes.

huhuanming:
Follow-up for the automated review of `931afe513433`:

- **Android autoplay / OOM guard:** no additional wrapper change is needed. The published `@onekeyfe/react-native-image@3.0.99` implementation already resolves `autoplay` as [`autoplay ?? Platform.OS !== "android"`](https://github.com/OneKeyHQ/app-modules/blob/7efde38b39f162b8e9cf698682061a956befae59/native-views/react-native-image/src/index.tsx#L321), so an omitted prop remains disabled on Android. The Android implementation also constrains a decoded ARGB bitmap to [16 MiB](https://github.com/OneKeyHQ/app-modules/blob/7efde38b39f162b8e9cf698682061a956befae59/native-views/react-native-image/android/src/main/java/com/margelo/nitro/onekeyimage/OneKeyImageDecodeDimensions.kt#L10-L27), animated encoded payloads to [16 MiB, canvas size, frame count, and duration limits](https://github.com/OneKeyHQ/app-modules/blob/7efde38b39f162b8e9cf698682061a956befae59/native-views/react-native-image/android/src/main/java/com/margelo/nitro/onekeyimage/OneKeyImageSafety.kt#L38-L47).
- **Custom source identity:** fixed in `77f248e8e0`. Rendering and native preload now explicitly disable TOS URL rewriting for sources with headers. Web skips authenticated sources instead of prefetching without headers, reports the batch as incomplete, and still prefetches public entries.

Validation: focused Web regression test passed; `yarn agent:check --profile commit` passed. Per the agreed scope, no new Native test or device run was added.

originalix (commented):
@huhuanming

Reviewed commit `77f248e8e0cb`.

- **P2 · Android autoplay OOM safeguard is no longer enforced**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13137#issuecomment-5517423127)

apps/mobile/package.json
```diff
@@ -71,32 +71,33 @@
...
+    "@onekeyfe/react-native-perp-depth-bar": "3.0.100",
+    "@onekeyfe/react-native-range-downloader": "3.0.100",
+    "@onekeyfe/react-native-scroll-guard": "3.0.100",
+    "@onekeyfe/react-native-segment-slider": "3.0.100",
+    "@onekeyfe/react-native-skeleton": "3.0.100",
+    "@onekeyfe/react-native-sni-connect": "3.0.100",
+    "@onekeyfe/react-native-splash-screen": "3.0.100",
+    "@onekeyfe/react-native-split-bundle-loader": "3.0.100",
+    "@onekeyfe/react-native-tab-view": "3.0.100",
+    "@onekeyfe/react-native-text-input": "3.0.100",
```
sidmorizon: **iOS dependency lock remains outdated**

Severity: `non-severe`

The `3.0.100` packages retain `3.0.99` entries in the CocoaPods lock. Direct iOS builds can pair new JavaScript with old native binaries.

huhuanming: Fixed in 218a38e0f4. Regenerated Podfile.lock from the published 3.0.100 packages; all app-modules pods now resolve to 3.0.100 and no 3.0.99 entries remain. Pod install completed successfully.

originalix (commented):
@huhuanming

Reviewed commit `8a487ccce152`.

- **P1 · iOS lockfile is stale after the 3.0.100 native-suite bump**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13137#issuecomment-5517423127)

huhuanming:
Resolved the finding from review https://github.com/OneKeyHQ/app-monorepo/pull/13137#pullrequestreview-5097622938.

The finding was valid for reviewed commit `8a487ccce152`: `apps/mobile/package.json` had moved the native suite to `3.0.100`, while `apps/mobile/ios/Podfile.lock` still resolved `OneKeyImage` and `react-native-tab-view` at `3.0.99`.

- `218a38e0f4` regenerated the iOS lockfile for `3.0.100`.
- `c6d29e1dc3` then moved the published `@onekeyfe/*` suite and iOS lockfile to `3.0.101`.
- On the current PR head, `OneKeyImage`, `Skeleton`, and `react-native-tab-view` all resolve to `3.0.101` in `Podfile.lock`.
- `yarn workspace @onekeyhq/mobile ios:pod-install` completes successfully (194 dependencies / 215 pods) and leaves the tracked worktree clean.

No additional code change is required for this stale-lockfile review. Per the agreed scope, this was dependency/lockfile validation rather than an iOS device test.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -57,46 +199,58 @@ export function ImageV2({
...
```
sidmorizon, line 226: **Share captures contain fading logos**

Severity: `non-severe`

On Android main UI, `fadeDuration={0}` is discarded. Share capture can save logos during the native fade instead of fully rendered.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -1,36 +1,178 @@
...
+      }
+      const sourceScale = getPositiveNumber(source.scale) ?? 1;
+      const precision = Math.abs(
+        1 -
+          (sourceWidth * sourceHeight * sourceScale * sourceScale) / targetArea,
+      );
+      if (precision < bestPrecision) {
+        bestPrecision = precision;
+        bestSource = source;
+      }
```
sidmorizon, line 126: **Responsive images select undersized assets**

Severity: `non-severe`

On iOS and Android main UI, `selectSourceCandidate` minimizes absolute area error. It can upscale blurry assets despite a suitable larger candidate.

huhuanming: This selector intentionally preserves React Native multi-source semantics. RN 0.86.2 performs this selection natively after layout on both platforms and uses the same absolute image-area error: RCTImageView.imageSourceForSize on iOS and MultiSourceHelper.getBestSourceForSize on Android. Changing it to always prefer a larger candidate would diverge from the ImageSourcePropType contract. I also found no multi-source array callsites in this repository, so there is no current product path affected. Moving candidate arrays into @onekeyfe/react-native-image can be considered as a separate API enhancement; it requires a structured Nitro prop plus Kotlin and Swift selection after layout, rather than a monorepo-only change.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: No code change is needed here. fadeDuration=0 was required by the React Native Android Fresco image, whose default hierarchy fade is 300 ms. OneKeyImage does not apply a fade on either platform: Android uses a Glide CustomViewTarget and immediately calls setImageDrawable without a transition, while iOS uses sd_setImage without a UIView transition. Ignoring this RN-only compatibility prop therefore already preserves the share-capture requirement: the image is displayed atomically rather than captured during a fade.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -161,63 +306,38 @@ export function ImageV2({
         setRawSourceFallbackUri(optimizedSourceResult.rawUri);
         return;
       }
```
sidmorizon, line 348: **Transient image failures become permanent**

Severity: `non-severe`

On iOS, Android, and web, `handleError` stops after the first raw-image failure. Temporary network errors leave images missing until remount.

packages/components/src/primitives/Image/ImageV2.native.tsx
```diff
@@ -57,46 +199,58 @@ export function ImageV2({
...
+    defaultSource: _defaultSource,
+    fadeDuration: _fadeDuration,
+    loadingIndicatorSource: _loadingIndicatorSource,
+    onPartialLoad: _onPartialLoad,
+    progressiveRenderingEnabled: _progressiveRenderingEnabled,
+    resizeMethod: _resizeMethod,
+    srcSet: _srcSet,
+    tintColor: _tintColor,
+    crossOrigin: _crossOrigin,
+    referrerPolicy: _referrerPolicy,
```
sidmorizon, line 234: **Native image effects are silently ignored**

Severity: `non-severe`

On iOS and Android, `tintColor` and other advertised image props are discarded. Tinted, blurred, inset, and default-image rendering silently disappears.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Fixed in a53b7e4a13. Restored bounded retries on native and web after the raw image request fails (default 1, configurable via canRetry/retryTimes). Optimized-source failures still fall back to the raw URL before consuming a retry.

packages/components/src/primitives/Image/ImageV2.native.tsx
huhuanming: Addressed in a53b7e4a13. ImageV2 intentionally does not support these native image-effect props: tintColor, blurRadius, and defaultSource are now explicitly deprecated/typed as unsupported, and the remaining unsupported native-only props are excluded from the public contract. This avoids advertising props that the native renderer discards; there are no current project call sites for them.

wabicai (approved):

sidmorizon (approved):

PR #10132, opened 2026-02-11
fix(android): patch expo-image SIGSEGV crash and webview bitmap allocation

huhuanming:
## Summary
- **expo-image**: Handle single `SourceMap` passed as a plain object instead of a list on Fabric (new architecture). On the new architecture, the bridge may serialize a single-element source array as a plain map object (`DynamicFromMap`). Without this handler, the `EitherOfThree` type converter throws `TypeCastException`, leaving the underlying `folly::dynamic` in an inconsistent state, leading to SIGSEGV in `folly::dynamic::destroy()` when the app backgrounds under memory pressure.
- **react-native-webview**: Use a lazy singleton `Bitmap` for `defaultVideoPoster` instead of allocating a new `Bitmap` on every `getDefaultVideoPoster()` call, reducing unnecessary memory allocations.

## Test plan
- [ ] Verify expo-image renders correctly on Android with Fabric (new architecture)
- [ ] Verify no SIGSEGV crash when backgrounding the app under memory pressure
- [ ] Verify webview video poster still displays correctly on Android
- [ ] Test on both old and new architecture builds

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/10132" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #13153, opened 2026-09-03
fix: Android first-frame flicker in bulk send, bulk copy addresses and buy crypto list, bulk send sender reseed, Cosmos base-unit fee(OK-61587, OK-61586, OK-61597, OK-61627, OK-61701)

weatherstar:
[OK-61587](https://onekeyhq.atlassian.net/browse/OK-61587) [OK-61586](https://onekeyhq.atlassian.net/browse/OK-61586) [OK-61597](https://onekeyhq.atlassian.net/browse/OK-61597) [OK-61627](https://onekeyhq.atlassian.net/browse/OK-61627) [OK-61701](https://onekeyhq.atlassian.net/browse/OK-61701)

----------

## Summary
- Android first-frame flicker batch: Bulk Send addresses page (OK-61587), Bulk Copy Addresses page (OK-61586) and the Buy Crypto token list (OK-61597) now paint a single, size-stable state per frame — no "Select token" / "No Results Found" / empty-selector intermediates, no icon or badge remounts, and re-entries render from an SWR snapshot on the first frame.
- Bulk Send keeps an explicitly picked sender when the address type changes in the recipient picker (OK-61627).
- Cosmos confirm page shows the amount and fee in on-chain base units (OK-61701).

## Intent & Context
QA filed a set of Android 6.6.0 (build 2026090141) issues from one screen recording: entering Bulk Send (One to Many / Many to One / Many to Many) filled the form in stages and briefly enabled Next and flashed "Sender address(es) is required" under an already-filled address; entering Bulk Copy Addresses flashed empty Wallet/Network selectors plus "No Results Found"; the Buy Crypto list flickered between skeleton, text, icons and network badges. Android modals open with `animation: 'none'`, so every main→bg round trip on a modal's first frame is visible (iOS hides it behind the slide animation). Each fix was diagnosed from frame-by-frame analysis of the QA recording (52 fps) and confirmed with a timing trace on a Pixel dev build, then re-verified on the same device.

## Root Cause
- **OK-61587 (Bulk Send)** — three independent defects in `BulkSendAddressesInput`: (1) the mount-time "mode reset" effect marked token details as initialized before the page was seeded, opening the Next gate and showing "Balance -"; (2) on a cold start the addressInput account-selector store hydrates ~100 ms late, so react-hook-form's mount validation ran with no registered fields and reported `isValid=true` (Next enabled for ~70 ms); (3) `useDebouncedValidation.cancel()` resolved a superseded pending validation with the last *completed* result of a different value — the mount validation of `''` completed as "required", the seeded address cancelled the follow-up `''` validation with that stale result, and RHF wrote the error under the filled address (a race between the 300 ms debounce and init, hence intermittent). Secondary: init needed 2–3 bg round trips and `Token` remounted on a networkId key, so name → icon → address → badge → balance arrived in stages.
- **OK-61586 (Bulk Copy Addresses)** — `usePromiseResult`'s `isLoading` is `undefined` until the runner starts, and the accounts runner early-returns an empty result while the wallet list is still resolving, so the page rendered the `Empty` "No Results Found" state for "not loaded yet"; the wallet trigger and the shared `ChainSelectorInput` each rendered blank boxes until their own fetches resolved.
- **OK-61597 (Buy Crypto)** — network metadata arrived after tokens, `Token` switched its root element when the network logo appeared (remounting the image), Android `useImage` has no synchronous cache so each remount re-decoded through a skeleton; the list also re-showed its 5-row skeleton on every open because the bg round trip is visible on Android.
- **OK-61627** — the Bulk Send init effect re-seeded the sender from the home scene whenever the home account id changed, including a derive-type-only change made from the recipient picker's address-type menu.
- **OK-61701** — Cosmos fee totals were fractional base units, so the confirm page preview and the fee display disagreed with the on-chain amount.

## Design Decisions
- **One-round-trip seed for Bulk Send**: new `serviceBulkSend.getAddressesInputSeed` resolves account/network fix-up, native token (with network name), network logo, the OneToMany sender address and wallet/account names together, degrading per lookup. The page consumes it through `usePromiseResult` + `swrKeys.bulkSendAddressesInputSeed`, moves `useForm` into the provider so token, network and sender address land in one commit, compares seeds by value (`isBulkSendSeedEqual`) to avoid re-applying identical snapshots, and gates Next on explicit `isInitializing` / `isSenderFieldMounted` flags (`computeBulkSendNextDisabled`) instead of trusting form validity alone. Size-stable skeletons (`BulkSendFormSkeleton`, asset row) replace the placeholder states.
- **Debounced validation semantics**: superseded callers now adopt the newest validation result (`cancel(promise)` supported for the synchronous multi-line path) rather than a stale completed result; `form.trigger()` after a selection change only re-validates non-empty fields, so an empty receiver never earns a "required" error on entry.
- **Bulk Copy Addresses**: the accounts result carries a `loaded` flag and `computeBulkCopyByAccountsViewState` renders the skeleton until `selectedWallet && loaded`, the empty state only after a completed empty load; wallets, compatible network ids and per-(wallet, network) account groups are SWR-snapshotted (`swrShouldPersist: loaded`), `walletsMap` is derived from the list so a cached list resolves the selection on frame 1, and the export gate uses "no form errors" instead of RHF's first-frame `isValid=false` (the form has no rules).
- **Shared `ChainSelectorInput`**: its network list is snapshotted (`swrKeys.chainSelectorInputNetworks`) and it shows a same-size placeholder while a value is set but the list has not resolved. Behaviour is otherwise unchanged; other screens using the selector (Send, address book, etc.) get the same first-frame benefit.
- **Buy Crypto**: `Token` always renders one `<Stack>` root; `serviceFiatCrypto.getTokensListWithNetworks` returns tokens + networks in one response; Android `useImage` serves a known-cached URL synchronously and `ImageV2.native` uses `memory-disk` cache policy (the decoded-ref cache stays iOS-only per #10132/#11975); the buy list is snapshotted via `swrKeys.fiatCryptoTokenList` and shows its skeleton only when there are no items.
- **OK-61627**: the home seed key omits the derive type (`homeAccountSeedUtils`), and the sender address-type chip resolves from the sender account (`getDBAccount` + `getDeriveTypeByDBAccount`) with the network default only as fallback.
- **OK-61701**: `chainValueUtils.floorNativeTokenAmount` for the amount preview and `calculateTotalFeeNative` rounding up (ROUND_CEIL) for the fee display, a no-op where totals are already integral.

## Changes Detail
- `packages/kit-bg/src/services/ServiceBulkSend.ts` (+ registration in `BackgroundApi*.ts` / `IBackgroundApi.ts`, tests): single-round-trip addresses-page seed.
- `packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/*`: provider-owned form + seed application, `Context` gates, `bulkSendSeedUtils`, `BulkSendFormSkeleton`, asset row skeleton / network logo from seed / no remount key, seeded sender badge, sender-field mount marker.
- `packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts` (+ tests): adopt-newest-result semantics.
- `packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx`, `utils/bulkCopyAddressesViewState.ts` (+ tests): `loaded` gating, selector skeleton, SWR snapshots, export gate.
- `packages/kit/src/components/ChainSelectorInput/index.tsx`: network-list snapshot + placeholder.
- `packages/kit/src/components/Token/Token.tsx`, `packages/kit/src/views/FiatCrypto/**`, `packages/kit-bg/src/services/ServiceFiatCrypto.ts`, `packages/components/src/primitives/Image/{useImage.ts,ImageV2.native.tsx}` (+ tests): Buy Crypto list stability.
- `packages/shared/src/utils/swrCacheUtils.ts`: new namespaces/keys for the snapshots above.
- `packages/shared/src/utils/{chainValueUtils,feeUtils}.ts` (+ tests), `usePreCheckTokenBalance.ts`: Cosmos base-unit amount and fee.
- `packages/shared/types/bulkSend.ts`, `packages/shared/types/fiatCrypto.ts`: seed / list response types.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (primary, Android-visible), Desktop / Web / Extension (shared components and hooks run everywhere)
- **Risk Areas**: `useDebouncedValidation` semantics change affects every Bulk Send validation path (sender/receiver, single- and multi-line); `ChainSelectorInput` is shared by Send, address book and other selectors; SWR snapshots persist wallet/account structures and token lists in MMKV, so stale-until-revalidated data is possible for one round trip after wallet or account edits; Android image cache policy change applies app-wide.

## Test plan
- [x] Unit tests: `useDebouncedValidation`, `ServiceBulkSend.getAddressesInputSeed`, `bulkSendSeedUtils`, `bulkCopyAddressesViewState`, `homeAccountSeedUtils`, `Token`, FiatCrypto `TokenList` / hooks, `ServiceFiatCrypto.getTokensListWithNetworks`, `useImage.android`, `ImageV2.native`, `swrCacheUtils`, `chainValueUtils`, `feeUtils`
- [x] Android device (Pixel, ProdDebug + Metro, frame-by-frame): Bulk Send One to Many ×3 + Many to One + Many to Many — skeleton → complete frame, address and badge on the same frame, Next never enabled early, no "required" flash; Bulk Copy Addresses cold and repeated entries — no "No Results Found" flash, complete first frame on re-entry; Buy Crypto ×5 (Ethereum) + ×5 (All Networks) — no skeleton regression, no icon/text/badge popping
- [ ] QA re-test on the release build: repeat the recorded steps for OK-61587 / OK-61586 / OK-61597 on Android, and the OK-61627 / OK-61701 steps on a BTC hardware wallet and a Cosmos account
- [ ] Spot-check other screens using `ChainSelectorInput` (Send, address book) for the selector's first-frame rendering

[OK-61587]: https://onekeyhq.atlassian.net/browse/OK-61587?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61586]: https://onekeyhq.atlassian.net/browse/OK-61586?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61597]: https://onekeyhq.atlassian.net/browse/OK-61597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61627]: https://onekeyhq.atlassian.net/browse/OK-61627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61701]: https://onekeyhq.atlassian.net/browse/OK-61701?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
@cursoragent review

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
```diff
@@ -799,6 +897,39 @@ function BulkSendAddressesInputProvider() {
...
+      bulkSendMode === EBulkSendMode.OneToMany ? seed.sender : undefined;
+    // Everything lands in one commit: the token, the network and the
+    // sender address paint together instead of in stages.
+    setSelectedAccountId(seed.accountId);
+    setSelectedNetworkId(seed.networkId);
+    setSelectedToken(seed.token);
+    setSelectedIndexedAccountId(seed.indexedAccountId);
+    setSeededNetwork(seed.network);
+    setSeededSender(sender);
+    setHasUserSelectedAsset(false);
```
sidmorizon: 🟠 P1: Fresh seed overwrites user selections

Severity: `severe`

When cached seed metadata changes during revalidation, `setSelectedAccountId` restores the entry sender and asset after the user changes them. The transfer can proceed with unintended selections.

packages/kit-bg/src/services/ServiceBulkSend.ts
```diff
@@ -0,0 +1,208 @@
...
+
+    const resolveAccountOnSelectedNetwork = async () => {
+      if (!selectedNetworkId || !indexedAccountId) {
+        return;
+      }
+      const networkAccounts =
+        await serviceAccount.getNetworkAccountsInSameIndexedAccountId({
+          networkIds: [selectedNetworkId],
+          indexedAccountId,
+        });
```
sidmorizon, line 62: **Account lookup failures freeze bulk send**

Severity: `non-severe`

For unsupported or All Networks selections, a rejected `getNetworkAccountsInSameIndexedAccountId` aborts the entire seed request. The address form remains permanently unavailable.

packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
```diff
@@ -18,20 +25,35 @@ export function useDebouncedValidation<T extends string>(
...
+  const settlePending = useCallback(
+    (result: IValidationResult | Promise<IValidationResult>) => {
+      const resolves = pendingResolvesRef.current;
+      pendingResolvesRef.current = [];
+      if (resolves.length === 0) {
+        return;
+      }
+      void Promise.resolve(result).then((settled) => {
+        resolves.forEach((resolve) => resolve(settled));
+      });
```
sidmorizon: **Rejected replacement stalls validation**

Severity: `non-severe`

When `cancel` receives a rejected replacement promise, `Promise.resolve(result).then` never settles earlier validations. The form remains validating and blocks the next step.

Suggested fix:

```ts
      void Promise.resolve(result).then(
        (settled) => resolves.forEach((resolve) => resolve(settled)),
        () => resolves.forEach((resolve) => resolve(false)),
      );
```

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
```diff
@@ -287,14 +329,44 @@ function BulkCopyAddresses({
...
+      // Snapshot per (wallet, network) so a re-entry renders the previous
+      // account groups on the first frame; only completed loads are kept.
+      swrKey:
+        selectedWalletId && selectedNetworkId
+          ? swrKeys.bulkCopyAddressesAccounts({
+              walletId: selectedWalletId,
+              networkId: selectedNetworkId,
+            })
+          : undefined,
+      swrShouldPersist: (result) => result.loaded,
```
sidmorizon: # 🟠 P1 新增 SWR 快照未接入钱包/账户失效链路，导出可能落在已删除的账户上

## 问题

`ServiceAccount` 构造函数里已经有一条明确的失效契约：`WalletUpdate` / `AccountRemove` / `AccountUpdate` / `WalletRename` / `AddDBAccountsToWallet` / `WalletClear` 触发时，按前缀丢弃 `walletListSideBar` 与 `accountSelectorList` 两个命名空间（`packages/kit-bg/src/services/ServiceAccount/ServiceAccount.ts:353-407`，注释原文就是 "avoids painting deleted wallets"）。

本 PR 新增的 `bulkCopyAddressesWallets` / `bulkCopyAddressesNetworkIds` / `bulkCopyAddressesAccounts` / `bulkSendAddressesInputSeed` 缓存的同样是钱包与账户结构（`networkAccounts` 里是完整 `INetworkAccount`，含 `address`，UTXO 链还含 `xpub`/`xpubSegwit`），但没有注册到上述任何一个 handler，MMKV 快照会跨进程重启一直留存。

## 影响

页面未挂载时（或上一次运行时）删除钱包 / 删除账户 / 批量新建账户后，下次冷进入 Bulk copy addresses：`availableWallets`、`networkAccountsByDeriveType` 首帧直接来自快照，而 `swrShouldPersist: (result) => result.loaded` 保证被持久化的条目 `loaded` 必为 `true`，于是 `accountsLoaded` 首帧就是 `true`；再叠加本 PR 把导出门槛从 `formState.isValid` 换成 `isEmpty(form.formState.errors)`（首帧即 `true`），`computeBulkCopyByAccountsViewState` 在第一帧就返回 `isExportDisabled: false`。

此时点击导出，`handleExportAddresses` 会把快照里的 `networkAccountsByDeriveType` 原样透传给 `ExportAddressesModal`（`BulkCopyAddresses.tsx:829-833`），导出的是已删除钱包的地址、或缺少新账户的旧集合。BulkSend 的 seed 同理会拿已删除账户预填 sender。

## 建议

把新命名空间纳入同一条失效链路，并让导出按钮只信任本次会话已完成的请求：

```ts
// ServiceAccount 构造函数，与 dropWalletListSwr 并列
const dropBulkAddressSwr = () => {
  swrCacheUtils.removeByPrefix(prefixOf(swrCacheNamespaces.bulkCopyAddressesWallets));
  swrCacheUtils.removeByPrefix(prefixOf(swrCacheNamespaces.bulkCopyAddressesNetworkIds));
  swrCacheUtils.removeByPrefix(prefixOf(swrCacheNamespaces.bulkCopyAddressesAccounts));
  swrCacheUtils.removeByPrefix(prefixOf(swrCacheNamespaces.bulkSendAddressesInputSeed));
};
```

快照只用于渲染骨架/占位，`isExportDisabled` 额外要求一个「本次挂载后请求已 resolve」的标志位。

[Claude session ↗](https://claude.ai/code/session_01Pu6QTwrYnbpopBtPiS5E2j)

---

originalix:

@weatherstar

Reviewed commit `2f9177c5f7e7`. This report always reflects the latest reviewed changes.

### Review summary

The change moves first-frame data assembly for Bulk Send and Fiat Crypto into background-owned composite responses, then uses persistent SWR snapshots and synchronous network/image metadata in the foreground. Before: screens independently resolved account, network, token, sender, and image state, producing staged empty or skeleton frames; fee previews could also retain fractional base units. After: Bulk Send applies one account/network/token/sender seed, Bulk Copy and Buy Crypto reuse scoped snapshots, Android reuses Glide-cached URLs and decoded bitmaps, and fee/max-send values are rounded to representable on-chain units.

Supporting work preserves a user-selected Bulk Send sender across derive-type changes, keeps Bulk Copy on a stable skeleton until account enumeration completes, and adds focused regression coverage. Sell lists and unsafe All Networks imported-account buy lists intentionally remain uncached, while signing and transaction construction remain unchanged apart from display and balance-unit normalization.

**What needs attention:** Confirm cross-runtime invalidation and rejection handling for persistent account-owned snapshots before accepting the first-frame optimizations.

### Issues to address

- **P1 · Seed remapping errors leave Bulk Send permanently initializing**
  - **Impact:** When OneToMany starts from All Networks or an unsupported network, getAddressesInputSeed must remap the indexed account. That lookup is the only seed lookup not protected by a fallback, so a rejection escapes the service. The foreground never receives a seed, appliedSeed never matches seedKey, the sender remains a skeleton, and Next stays disabled for the lifetime of the page.
  - **Suggested action:** Catch the account-remapping failure inside ServiceBulkSend and resolve a partial seed so the foreground can mount and let the user choose a valid account or asset; add a rejection regression test.
- **P1 · Account mutations do not invalidate Bulk Copy snapshots**
  - **Impact:** The new Bulk Copy SWR entries persist wallet objects and account groups without a TTL, but the existing ServiceAccount mutation handlers only invalidate walletListSideBar and accountSelectorList. If a wallet/account is removed, deactivated, or renamed while a snapshot exists, a later mount can still treat cached loaded data as actionable; when revalidation is delayed or fails, stale wallet/account data remains selectable and Export stays enabled. On iOS, Android, and extension, main and bg have separate JS cache copies over shared native storage, so invalidation must reach both runtime copies rather than assuming shared heap state.
  - **Suggested action:** Add a shared invalidator for the three Bulk Copy cache prefixes and invoke it from the existing wallet/account mutation listeners in both foreground and background runtimes, flushing the shared storage removal intents.
- **P2 · Account enumeration failures render an endless skeleton**
  - **Impact:** If any getNetworkAccountsInSameIndexedAccountIdWithDeriveTypes call rejects, Promise.all rejects and the result remains EMPTY_ACCOUNTS_RESULT with loaded=false. The new view-state helper interprets that value as an in-progress load forever, so the page shows no error or retry and both export actions remain disabled.
  - **Suggested action:** Model an explicit failed account-load state, catch the enumeration rejection, and render a recoverable error/retry state instead of leaving loaded false.

### Validation gaps

- The targeted Jest command could not start because the workspace lacks the Yarn `node_modules` state file; the added tests were inspected but not executed here.
- Native cold/warm visual behavior and wallet-mutation recovery were not exercised interactively.
- Existing GitHub reviews and comment threads could not be queried because network access to `api.github.com` failed, so inline drafts require a duplicate-comment check before publication.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/bulkSendSeedUtils.ts
```diff
@@ -0,0 +1,133 @@
...
+    tokenInfo: routeParams.tokenInfo,
+    bulkSendMode,
+  };
+}
+
+export function buildBulkSendSeedTokenKey(tokenInfo: IToken | undefined) {
+  if (!tokenInfo) {
+    return '';
+  }
+  return `${tokenInfo.networkId ?? ''}:${tokenInfo.address}`;
```
sidmorizon, line 51: **Token metadata updates remain cached**

Severity: `non-severe`

When token decimals or native status changes, `buildBulkSendSeedTokenKey` reuses the old snapshot and `isSameToken` accepts it. Bulk send then uses stale precision.

packages/kit-bg/src/services/ServiceBulkSend.ts
```diff
@@ -0,0 +1,208 @@
...
+    };
+
+    // Unsupported home network: fall back to the same indexed account on
+    // the corrected network. All Networks: pick the per-network account.
+    if (
+      (!isSupported && indexedAccountId) ||
+      (isAllNetwork &&
+        !accountUtils.isOthersAccount({ accountId: selectedAccountId ?? '' }))
+    ) {
+      await resolveAccountOnSelectedNetwork();
```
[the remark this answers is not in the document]
weatherstar: Fixed in 5099fcc194.

- `ServiceBulkSend.getAddressesInputSeed`: the account remap is now wrapped in try/catch. On failure the seed resolves without an account (the caller-provided one belongs to the uncorrected network / All Networks pseudo account, so it cannot seed token or sender lookups on the corrected network) and the rest of the seed still comes back. Regression test added (`resolves a partial seed when the account remap rejects`).
- Page side: the seed fetcher in `BulkSendAddressesInputProvider` now catches a rejected request and resolves a fallback seed built from the raw source (`buildBulkSendFallbackSeed`), so `appliedSeed` always lands and `isInitializing` clears. The fallback is excluded from the SWR snapshot via `swrShouldPersist`, so the next entry retries.

packages/kit-bg/src/services/ServiceBulkSend.ts
weatherstar: Fixed in 5099fcc194: the remap is caught, the seed resolves without an account (and therefore without token / sender) instead of rejecting, and the page additionally falls back to a raw-source seed if the whole request rejects. Regression test added in `ServiceBulkSend.getAddressesInputSeed.test.ts`.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
```diff
@@ -278,6 +319,7 @@ function BulkCopyAddresses({
       const resp = await Promise.all(accountsRequest ?? []);
```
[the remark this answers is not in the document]
weatherstar: Fixed in 5099fcc194.

- Enumeration now uses `Promise.allSettled`; any rejected indexed account yields `{ loaded: true, loadFailed: true }` rather than leaving `loaded: false`. A partial list is deliberately not returned so export can never forward an incomplete set.
- `computeBulkCopyByAccountsViewState` gained an explicit `showError` state: the list renders an `Empty` with a Retry button (re-runs the request), export stays disabled, and the failed result is never persisted (`swrShouldPersist` requires `!loadFailed`). Tests updated.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
weatherstar: Fixed in 5099fcc194. The apply step now goes through `resolveBulkSendSeedApplyPlan` (pure, unit-tested):

- If the user has moved off the previously applied seed's account / network and the incoming seed targets a different account (a revalidated snapshot whose metadata changed, or the re-keyed request after a mode switch), the seed is only recorded as applied; `selectedAccountId` / network / token and the sender field are left alone.
- A seed for the selection the user is on may refresh it, and an asset the user picked explicitly (`hasUserSelectedAsset`) is kept instead of being reset to the seed token.
- First mount and an untouched selection still apply in full as before.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
```diff
@@ -407,31 +339,70 @@ function BaseBulkSendAddressesInput() {
...
+      return;
     }
+    // The seed already wrote the address for this selection in the same
+    // commit as the token; only a selection made on the page (account /
+    // network / address type picker) needs the extra lookup.
+    if (
+      selectedAccountId === seededAccountId &&
+      selectedNetworkId === seededNetworkId
+    ) {
+      return;
```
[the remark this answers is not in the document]
weatherstar: Fixed in 5099fcc194. The skip no longer relies on the seed identity alone: the lookup is skipped only while the sender field still holds the seeded address for the seeded selection (`form.getValues('senderAddresses') === seededSender.address`). Switching to another account and back, or the mode reset that clears the field, now refreshes the address for the current selection. `fetchSelectedAccountAddress` also got a request-id guard so a slow lookup for a previous account cannot land over a newer pick, and a failed lookup keeps the field instead of throwing.

packages/kit/src/views/BulkSend/hooks/useDebouncedValidation.ts
weatherstar: Fixed in 5099fcc194 as suggested: a rejected replacement promise settles the pending callers with `false` (same as a validator that throws inside the hook), so the form leaves the validating state. Test added (`cancel(rejectedPromise) settles pending callers as invalid instead of hanging`).

packages/kit-bg/src/services/ServiceFiatCrypto.ts
```diff
@@ -117,6 +121,41 @@ class ServiceFiatCrypto extends ServiceBase {
...
 
+  @backgroundMethod()
+  public async getTokensListWithNetworks(
+    params: IGetTokensListParams,
+  ): Promise<IFiatCryptoTokenListWithNetworks> {
+    const tokens = await this.getTokensList(params);
+    const networkIds = uniq(tokens.map((token) => token.networkId));
+    const { serviceNetwork } = this.backgroundApi;
+    const [{ networks }, mergeDeriveFlags] = await Promise.all([
+      serviceNetwork.getNetworksByIds({ networkIds }),
```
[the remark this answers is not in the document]
weatherstar: Fixed in 5099fcc194: `getNetworksByIds` is caught and degrades to an empty `networksMap`, so the token list the fiat API already returned is still delivered (rows just lack name / icon, matching the previous separate-loader behaviour). Test added.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
weatherstar: Fixed in 5099fcc194, both parts:

1. **Invalidation chain** — `ServiceAccount` now has `dropBulkAddressSwr` next to `dropWalletListSwr`, dropping `bulkCopyAddressesWallets` / `bulkCopyAddressesNetworkIds` / `bulkCopyAddressesAccounts` / `bulkSendAddressesInputSeed` from every existing handler (WalletUpdate, AccountRemove, AccountUpdate, RenameDBAccounts, WalletRename, AddDBAccountsToWallet, WalletClear), followed by the same `flushNow`. It follows the existing contract for the sidebar / selector namespaces: the bg-owned drop is tombstoned as a prefix removal, so the other runtime's copy cannot revive it on merge.
2. **Export gate** — instead of greying the button on a cached entry (which would reintroduce the first-frame flip this PR removes), `handleExportAddresses` now checks whether the accounts for the current wallet / network were enumerated in this session (`freshAccountsScopeKeyRef`). If the page is still on the persisted snapshot it re-enumerates before navigating and forwards the fresh groups; a failed re-enumeration surfaces the retry state and aborts the export. The snapshot therefore only ever paints, never exports.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
```diff
@@ -287,14 +329,44 @@ function BulkCopyAddresses({
...
-        networkAccountsByDeriveType: {},
-      },
+      initResult: EMPTY_ACCOUNTS_RESULT,
+      // Snapshot per (wallet, network) so a re-entry renders the previous
+      // account groups on the first frame; only completed loads are kept.
+      swrKey:
+        selectedWalletId && selectedNetworkId
+          ? swrKeys.bulkCopyAddressesAccounts({
+              walletId: selectedWalletId,
+              networkId: selectedNetworkId,
```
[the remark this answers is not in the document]
weatherstar: Fixed in 5099fcc194: all four bulk copy / bulk send SWR namespaces are now purged from the existing wallet / account mutation handlers in `ServiceAccount` (including WalletClear and WalletRename), and export re-enumerates the accounts for the current wallet / network before forwarding anything when the list is still the persisted snapshot, so removed wallets or accounts cannot be exported from cache.

weatherstar:
@originalix Addressed the automated review report in 5099fcc194 (on top of the `x` merge 088aec51e9):

- **P1 · Seed remapping errors** — the account remap in `ServiceBulkSend.getAddressesInputSeed` is caught and resolves a partial seed (no account / token / sender) instead of rejecting; regression test added. The page also falls back to a raw-source seed if the whole request rejects, so `isInitializing` always clears.
- **P1 · Account mutations do not invalidate Bulk Copy snapshots** — `ServiceAccount` now drops the four bulk copy / bulk send namespaces from every existing wallet / account mutation handler (with the same `flushNow`), using the existing prefix-tombstone contract so the other runtime's copy cannot revive them on merge. Export additionally re-enumerates the accounts for the current wallet / network before forwarding anything when the list is still the persisted snapshot.
- **P2 · Account enumeration failures render an endless skeleton** — enumeration uses `Promise.allSettled`; a rejected indexed account yields an explicit `loadFailed` state rendered as an error with a Retry button, never persisted and never exported.

Also from the Cursor report: the mode switch no longer re-applies the home seed over a user-picked sender (seed apply now goes through `resolveBulkSendSeedApplyPlan`), and `fetchSelectedAccountAddress` has an in-flight guard. The `chainSelectorInputNetworks` key note (`undefined` vs `[]`) is left as is: both resolve to the full network list today, so the key is behaviour-preserving.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/bulkSendSeedUtils.ts
```diff
@@ -0,0 +1,209 @@
...
+    appliedSeed !== undefined &&
+    ((selectedAccountId ?? '') !== (appliedSeed.seed.accountId ?? '') ||
+      (selectedNetworkId ?? '') !== (appliedSeed.seed.networkId ?? ''));
+  const isSeedForCurrentSelection =
+    (seed.accountId ?? '') === (selectedAccountId ?? '') &&
+    (seed.networkId ?? '') === (selectedNetworkId ?? '');
+  if (hasUserChangedSelection && !isSeedForCurrentSelection) {
+    return { action: 'record' };
+  }
+  return { action: 'apply', keepUserToken: hasUserSelectedAsset };
```
sidmorizon: # [🟠 P1] keepUserToken 未限定「seed 指向当前选择」，会让 token 与 network/account 脱钩

## 问题

本函数的 JSDoc 写的是「A seed for the selection the user is on may refresh it, keeping an asset the user picked explicitly」，但实现里 `keepUserToken` 只取决于 `hasUserSelectedAsset`，并没有要求 `isSeedForCurrentSelection`。

`hasUserChangedSelection` 是拿「当前选择」和 `appliedSeed.seed` 比，而 `isSeedForCurrentSelection` 是拿「新 seed」和「当前选择」比。当用户**没有**动过 account/network（`hasUserChangedSelection === false`）、但新 seed 因为 seed source 变化（home 活跃账户切到别的 account/network、路由参数变化）指向了另一个 account/network 时，会走到这一行返回 `{ action: 'apply', keepUserToken: true }`。

回到 `BulkSendAddressesInput.tsx` 的 apply 分支：

```ts
setSelectedAccountId(seed.accountId);
setSelectedNetworkId(seed.networkId);
if (!plan.keepUserToken) {
  setSelectedToken(seed.token);
  setHasUserSelectedAsset(false);
}
```

account / network / `seededNetwork` 全部换成了新 seed 的值，唯独 `selectedToken` 保留了用户在**旧网络**上挑的那个 token。round 1 的旧代码是无条件 `setSelectedToken(seed.token)`，三者始终同步，这个错位是本次修复新引入的。

触发前置条件都是可达的：`handleSelectAsset` 打开 TokenSelector 时用的就是当前 `selectedAccountId/selectedNetworkId`，`buildTokenSelectHandler` 里 `nextNetworkId = token.networkId ?? networkId`，所以「在当前 account/network 上换个 token」会把 `hasUserSelectedAsset` 置 true 而不改变选择，正好满足 `hasUserChangedSelection === false`。

## 影响

`selectedToken.networkId !== selectedNetworkId`：

- OneToMany：`fetchTokensDetails` 拿新 network + 旧 network 的合约地址去查，`isBulkSendTokenDetailsMatched` 判否 → `selectedTokenDetail` 为 undefined，但 `tokenDetailsState.initialized` 已为 true，`computeBulkSendNextDisabled` 不再拦截；`navigateToNextStep` 里 `if (isOneToMany && (!selectedAccountId || !selectedTokenDetail)) return;` 静默返回 → Next 可点但毫无反应，页面卡死。
- ManyToMany / ManyToOne：`isSubmitDisabled` 不检查 token，直接 push 到 `BulkSendAmountsInput`，navParams 为 `networkId`（新网络）+ `tokenInfo`（旧网络的 token）。AmountsInput 用 `networkId` + `tokenInfo.address` / `tokenInfo.decimals` 构造 transfers，属于跨链错配的资金相关路径。

UI 上 `displayNetworkId = selectedToken?.networkId ?? selectedNetworkId`，资产行仍显示旧网络徽标，用户更难察觉。

## 建议

把 `keepUserToken` 收敛到函数注释描述的语义，只在 seed 就是当前选择时才保留用户选的资产：

```ts
return {
  action: 'apply',
  keepUserToken: hasUserSelectedAsset && isSeedForCurrentSelection,
};
```

这样「seed 换了 account/network」时会照旧写入 `seed.token` 并把 `hasUserSelectedAsset` 复位，token 与 network/account 始终来自同一次 apply；建议同时补一条单测覆盖「用户选过资产 + seed 换到另一个 network」的分支。

[Claude session ↗](https://claude.ai/code/session_01Pu6QTwrYnbpopBtPiS5E2j)

---

sidmorizon (commented):
Round 2 incremental review (5099fcc).

---

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
```diff
@@ -737,14 +865,26 @@ function BulkCopyAddresses({
...
+          const fresh = await loadAccounts();
+          if (fresh.loadFailed) {
+            void runAccounts();
+            return;
+          }
+          enumeratedAccounts = fresh.networkAccountsByDeriveType;
+        }
+        let accountsData = enumeratedAccounts;
         if (isHwWallet && !exportWithoutDevice) {
           accountsData = await handleGenerateAddressesByAccounts();
```
sidmorizon: # [🟡 P2] 硬件钱包导出：派生用旧快照、过滤用新枚举，两半对「当前账户集合」的认知不一致

## 问题

`:869-879` 已经在「仍停留在持久化快照」时重新枚举了一次，得到 `fresh` / `enumeratedAccounts`。但这一行调用的 `handleGenerateAddressesByAccounts`（`:557-560`）内部是从 React state 取数的：

```ts
const { customNetworks, indexes, addressCount } =
  buildBulkCopyByAccountsFlowParams({
    networkAccounts: networkAccounts ?? [],
  });
```

`networkAccounts` 就是可能仍为旧快照的 state，而不是刚刚 `loadAccounts()` 返回的 `fresh.networkAccounts`。随后 `:883-895` 只把 `enumeratedAccounts` 当作**过滤器**盖在 `accountsData` 上。

于是导出的两半使用了两个不同的账户集合作为事实来源：派生用旧的，过滤用新的。

## 影响

仅硬件钱包路径（`isHwWallet && !exportWithoutDevice`）：

- 快照之后**被删除**的账户 → 在过滤那一步被剔除，这半边是对的（本 PR 已解决的 SWR 失效 / 导出已删除账户问题不受影响）。
- 快照之后**新建**的账户 → `indexes` / `customNetworks` 里根本没有它，设备不会为它派生地址；过滤只能做减法，无法把它补回来。

结果是导出的地址列表静默缺项：用户拿到一份不完整的导出，没有任何报错或提示。非硬件路径直接透传 `enumeratedAccounts`，不受影响。这是完整性缺口，不是泄露。

## 建议

让派生和过滤共用同一份枚举结果：把账户分组作为参数传进 `handleGenerateAddressesByAccounts`，重新枚举发生时传 `fresh.networkAccounts`，否则传当前 state，例如

```ts
const handleGenerateAddressesByAccounts = useCallback(
  async (accountsForFlow: IBulkCopyNetworkAccounts[]) => {
    // ...
    buildBulkCopyByAccountsFlowParams({ networkAccounts: accountsForFlow });
  },
  [/* 去掉 networkAccounts 依赖 */],
);
```

调用处相应传入 `fresh.networkAccounts ?? networkAccounts ?? []`，保证 `indexes` 与最终过滤基于同一次枚举。

[Claude session ↗](https://claude.ai/code/session_01Pu6QTwrYnbpopBtPiS5E2j)

---

packages/kit-bg/src/services/ServiceBulkSend.ts
```diff
@@ -0,0 +1,219 @@
...
+      }
+      const networkAccounts =
+        await serviceAccount.getNetworkAccountsInSameIndexedAccountId({
+          networkIds: [selectedNetworkId],
+          indexedAccountId,
+        });
+      const resolvedAccountId = networkAccounts?.[0]?.account?.id;
+      if (resolvedAccountId) {
+        selectedAccountId = resolvedAccountId;
+      }
```
sidmorizon: **Missing remap keeps invalid sender**

Severity: `non-severe`

When no fallback-network account exists, `resolveAccountOnSelectedNetwork` retains the all-networks account. The form then resolves its asset and sender against the wrong network.

Suggested fix:

```ts
      const resolvedAccountId = networkAccounts?.[0]?.account?.id;
      selectedAccountId = resolvedAccountId || undefined;
```

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
```diff
@@ -395,43 +329,102 @@ function BaseBulkSendAddressesInput() {
...
+      cancelled = true;
+    };
+  }, [selectedAccountId, selectedNetworkId, setSelectedDeriveType]);
 
   useEffect(() => {
-    if (isOneToMany && selectedAccountId && selectedNetworkId) {
-      void fetchSelectedAccountAddress();
+    if (!isOneToMany || !selectedAccountId || !selectedNetworkId) {
+      return;
+    }
```
sidmorizon, line 406: **Mode switch restores stale sender**

Severity: `non-severe`

Changing modes during `fetchSelectedAccountAddress` does not invalidate that request. Its late response refills the cleared sender field with the previous account.

Suggested fix:

```ts
  useEffect(() => {
    senderAddressRequestIdRef.current += 1;
    if (!isOneToMany || !selectedAccountId || !selectedNetworkId) {
      return;
    }
```

packages/kit/src/components/ChainSelectorInput/index.tsx
```diff
@@ -47,13 +48,24 @@ export const ChainSelectorInput: FC<IChainSelectorInputProps> = ({
...
-    { initResult: [] },
+    {
+      initResult: [],
+      // Snapshot the list so the selected network name paints on the first
+      // frame of later visits instead of an empty box (OK-61586).
+      swrKey: swrKeys.chainSelectorInputNetworks({
+        excludeAllNetworkItem,
+        networkIds,
+      }),
+    },
```
sidmorizon, line 69: **Cached networks overwrite new selection**

Severity: `non-severe`

When `chainSelectorInputNetworks` omits a newly added network, the fallback effect replaces it with the first cached network before revalidation completes.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
```diff
@@ -737,14 +865,26 @@ function BulkCopyAddresses({
...
+          // Still on the persisted snapshot: re-enumerate before exporting
+          // so a wallet / account removed or renamed since the snapshot was
+          // taken is never forwarded to the export modal.
+          const fresh = await loadAccounts();
+          if (fresh.loadFailed) {
+            void runAccounts();
+            return;
+          }
+          enumeratedAccounts = fresh.networkAccountsByDeriveType;
+        }
```
sidmorizon, line 889: **Hardware export omits new accounts**

Severity: `non-severe`

After `loadAccounts` refreshes a stale snapshot, hardware generation still uses old `networkAccounts`. Newly discovered accounts are omitted from the export.

originalix (commented):
@weatherstar

Reviewed commit `5099fcc1945a`.

- **P1 · BulkSend seed keeps incompatible accountId when remap finds no account**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13153#issuecomment-5523674071)

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/bulkSendSeedUtils.ts
weatherstar: Fixed in 046e11385a exactly as suggested: `keepUserToken: hasUserSelectedAsset && isSeedForCurrentSelection`. A seed that moves the account / network brings its own token and resets `hasUserSelectedAsset`, so token and network always come from the same apply. Test added (`brings the seed token along when the seed moves to another network`).

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
```diff
@@ -756,33 +781,145 @@ function BulkSendAddressesInputProvider() {
...
+            source,
+          );
+        return { key, seed, isFallback: false };
+      } catch {
+        // A rejected request must still settle the initializing gate, or
+        // the sender stays a skeleton and Next stays disabled for the life
+        // of the page. Mount on the raw source and let the user pick.
+        return {
+          key,
+          seed: buildBulkSendFallbackSeed(source),
```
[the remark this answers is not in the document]
weatherstar: Fixed in 046e11385a. Seed results are now tagged `isDegraded` (request fallback, or `source.accountId` present while the resolved seed has none — this covers the remap catch and the empty-remap path too). `resolveBulkSendSeedApplyPlan` records a degraded seed instead of applying it whenever an `appliedSeed` exists (and, independent of the flag, whenever the incoming seed lacks an account the applied one had), so a painted snapshot or a successfully applied seed is never blanked; a cold start without a snapshot still mounts on the fallback. Degraded results are excluded from the SWR snapshot via `swrShouldPersist`. Test added (`records, without applying, a degraded seed over an already applied one`).

packages/kit-bg/src/services/ServiceBulkSend.ts
weatherstar: Fixed in 046e11385a as suggested: `selectedAccountId = networkAccounts?.[0]?.account?.id || undefined`, so an empty remap drops the All Networks / unsupported-network account instead of resolving token and sender against the wrong network. Test added (`drops the account when the corrected network has none yet`). The page treats such a result as degraded: it mounts an empty sender on a cold start but never overwrites an already applied seed and is not snapshotted.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/BulkSendAddressesInput.tsx
weatherstar: Fixed in 046e11385a as suggested: the effect bumps `senderAddressRequestIdRef` on every run (selection or mode change), so a lookup still in flight when the mode reset clears the field is discarded when it lands.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
weatherstar: Fixed in 046e11385a: `handleGenerateAddressesByAccounts` now takes the enumeration as a parameter (`accountsForFlow`) and no longer reads `networkAccounts` state. The export passes `fresh.networkAccounts` when it re-enumerated and the current state otherwise, so device derivation (indexes / customNetworks) and the final filter share one account set — accounts added after the snapshot are derived and exported too.

packages/kit/src/views/BulkCopyAddresses/pages/BulkCopyAddresses.tsx
weatherstar: Fixed in 046e11385a together with the thread above: hardware generation receives the same enumeration the filter uses (the fresh re-enumeration when the page was still on the snapshot), so newly discovered accounts are derived rather than omitted.

packages/kit/src/components/ChainSelectorInput/index.tsx
weatherstar: Fixed in 046e11385a: the fallback only runs once a list for the current key has been resolved in this session (`freshListKeyRef` set by the fetcher, `watchLoading` so the effect re-runs when the request settles). The persisted snapshot still paints the selected network's name on the first frame, but it can no longer decide that a value it does not know (e.g. a network added since the snapshot) should be replaced by its first entry.

packages/kit/src/views/BulkSend/pages/BulkSendAddressesInput/bulkSendSeedUtils.ts
weatherstar: Tightened in 046e11385a: `isSameToken` now also compares `decimals` and `isNative`, so a revalidated seed whose token metadata differs re-applies instead of being treated as equal. For context, the seeded token is either the caller-provided `tokenInfo` (passed through verbatim, not cached) or the chain's native token, whose precision does not change, so the practical exposure was nil — but the comparison no longer relies on that.

originalix (approved):

PR #13167, opened 2026-09-04
fix: require Prime crypto payment broadcast

yikZero:
## Summary

- fail closed when a Prime Infini payment would only sign instead of broadcasting
- honor the Dev mode master switch for alwaysSignOnlySendTx and guard explicit sign-only and external-account paths
- route Prime batch sign-only requests through the shared broadcast guard
- add privacy-safe primeBroadcastDiagV1 logs for path decisions, session persistence, real vault broadcast attempts, and broadcast results

## Root cause

ServiceSend previously returned a signed transaction as a successful result when broadcasting was skipped. The Prime caller treated that resolved result as a completed broadcast and entered payment polling even though no send-transaction request was made.

## Diagnostics

The added fields are restricted to booleans and stable classifications. Payment identifiers continue to be hashed by getPrimeInfiniPaymentSafeLogParams; wallet addresses, account identifiers, raw transactions, signatures, and credentials are not logged.

## Test plan

- yarn jest packages/kit-bg/src/services/ServiceSend.broadcastDeadline.test.ts packages/shared/src/utils/primeInfiniPaymentDiagnostics.test.ts --runInBand --silent
- git diff --check


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

sidmorizon (approved):

PR #13180, opened 2026-09-04
ci: stabilize mobile shell OCI cache keys

huhuanming:
## Summary

- derive iOS and Android dev-shell keys from installed JS-to-native ABI contracts after yarn install, without Pod, Gradle, Xcode, or local shell builds
- keep the existing v3 shell tag namespace while making non-ABI source, package version, lock checksum, and WebEmbed changes key-stable
- derive the WebEmbed OCI key from its reachable dependency graph and serve the current WebEmbed build from Metro during native DevSession
- resolve exact public OCI artifacts before build jobs and skip builds when the stable key already exists
- retry transient manifest, timeout, and token-service failures; preserve cache misses as exists=false

## Validation

- focused Jest: 5 suites, 139 tests passed
- two independent key calculations produced identical Android, iOS, and WebEmbed keys
- all configured native dependencies matched autolinking and produced non-empty ABI descriptors

No local native build was run.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-100ade61-d7aa-5a21-8901-a9aff897581a)

.github/workflows/mobile-dev-shell-android.yml
```diff
@@ -42,8 +42,48 @@ concurrency:
...
+          node-version: 24.x
+
+      - name: Resolve exact OCI shell
+        id: shell
+        run: |
+          set -euo pipefail
+          status_file="$RUNNER_TEMP/mobile-dev-shell-android.json"
+          node apps/mobile/scripts/native-dev-shell.js resolve \
+            --platform android \
+            --output "$status_file"
```
sidmorizon, line 79: 🟠 P1: Fresh runners cannot resolve shells

Severity: `severe`

On a fresh runner, `resolve` requires installed transitive packages through [`getNativeContractDescriptor`](apps/mobile/plugins/devVendor.js:601), but this job never installs dependencies. The iOS job fails identically, so neither workflow can build or skip a shell.

apps/mobile/scripts/mobile-dev-shell-resource.js
```diff
@@ -333,7 +338,51 @@ async function resolveOciShell({ compatibility, fetchImpl, locator, tag }) {
...
+      if (error?.code === 'SHELL_LOCATOR_NOT_FOUND') {
+        return {
+          exists: false,
+          ociDigest: null,
+          sourceCommit: null,
+          tag: compatibility.exactTag,
+        };
+      }
+      if (attempt === maxAttempts || !isRetryableOciError(error)) {
+        throw error;
```
sidmorizon, line 374: **Token service failures bypass retries**

Severity: `non-severe`

When GHCR's token request returns 429 or 5xx, its error lacks `retryable`, so `resolveExactMobileDevShell` aborts on the first attempt.

.github/workflows/mobile-dev-shell-android.yml
```diff
@@ -42,8 +42,48 @@ concurrency:
...
+          exists="$(jq -er '.exists' "$status_file")"
```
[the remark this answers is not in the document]
huhuanming: Fixed in a8d0a4aef1: exists is parsed with an explicit boolean type check without jq -e, so a valid false value reaches the build-job output.

.github/workflows/mobile-dev-shell-android.yml
huhuanming: Fixed in a8d0a4aef1: fresh resolve runners install the dependency graph in skip-builds mode before resolving Android or iOS shell keys.

apps/mobile/scripts/mobile-dev-shell-resource.js
huhuanming: Fixed in a8d0a4aef1: token-service 408, 425, 429, and 5xx responses are retryable, covered by a 503 authentication-challenge regression test.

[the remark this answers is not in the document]
huhuanming: Fixed in a8d0a4aef1: AbortError and TimeoutError are now classified as transient OCI errors, with a focused timeout retry regression test.

.github/workflows/mobile-dev-shell-android.yml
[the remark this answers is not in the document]
huhuanming: Fixed in a8d0a4aef1: both native resolve jobs now run yarn install --immutable --mode=skip-builds before computing the installed ABI contract.

originalix:

@huhuanming

Reviewed commit `3074ba11070a`. This report always reflects the latest reviewed changes.

### Review summary

Before: mobile development shells were keyed by broad native inputs plus the web-embed input, so web changes and many non-ABI native changes produced new shells; the web embed was packaged into the shell. After: shell identity is derived from a synthesized native ABI descriptor, CI checks the exact public OCI tag before building, and each development session restores or builds the web embed separately, serves it through Metro with no-store caching, and lets the native main UI runtime discover that URL from its bundle URL. The web-embed prebundle key now attempts to trace the resolved module graph and installed package records. OCI digest, provenance, and attestation verification remain in place, while production, desktop/web single-runtime behavior, and the isolated mobile background runtime are intentionally unchanged.

**What needs attention:** Confirm that both synthesized cache keys cover the exact resolver and native bridge contracts consumed by the artifacts they can cause CI to reuse.

### Issues to address

- **P1 · Web prebundle key ignores bundler fallbacks**
  - **Impact:** The tracer removes Node built-ins before resolution, but the real Rspack configuration replaces `crypto` and `stream` with browser implementations. The reachable web-embed bootstrap imports `crypto` through `packages/shared/src/modules3rdParty/cross-crypto/verify.ts`. A change to the actual fallback implementation can therefore leave the input key and OCI tag unchanged; CI then finds the old tag and skips rebuilding, serving stale code despite the triggering source change.
  - **Suggested action:** Build the input graph with the same exported resolve aliases and fallbacks as the web-embed Rspack build, resolving enabled fallbacks before discarding built-ins, and add a test proving that a fallback implementation change alters `getInputKey()`.
- **P1 · Native shell ABI key drops bridge parameter types**
  - **Impact:** Legacy Android and iOS bridge methods are represented only as `methodName/argumentCount`; package versions and source digests are intentionally excluded. Changing an exposed parameter from one native type to another while keeping the method name and arity leaves the native contract and exact OCI tag unchanged. CI can then skip the native build and pair new JavaScript with an old native method signature, causing bridge conversion failures at runtime.
  - **Suggested action:** Canonicalize and hash the full exported legacy bridge signatures, including parameter types and promise/callback roles, and add Android and iOS tests where a type-only signature change invalidates the native contract key.

### Validation gaps

- `git diff --check` and syntax checks for the changed JavaScript files passed.
- `yarn agent:check --profile pr` could not start because the workspace lacks the Yarn node-modules state file; repository tests, lint, and type checks were therefore not executed here.
- GitHub review and comment state could not be fetched because network access to `api.github.com` failed, so the inline drafts require a human duplicate-comment check.

originalix (commented):
@huhuanming

Reviewed commit `3074ba11070a`.

- **P1 · Web prebundle key ignores bundler fallbacks**
- **P1 · Native shell ABI key drops bridge parameter types**

[View the full review report](https://github.com/OneKeyHQ/app-monorepo/pull/13180#issuecomment-5539077510)

huhuanming:
Addressed both blocking findings in b8868cee99:

- Web prebundle dependency tracing now shares the Rspack resolve aliases and fallbacks, and resolves enabled Node built-in fallbacks before pruning disabled built-ins. A focused test proves fallback implementation changes invalidate the input key.
- Native shell ABI descriptors now include canonical Android @ReactMethod and iOS RCT method signatures with parameter types and Promise/Callback roles. Android and iOS type-only signature changes are covered by regression tests.

Validation: focused Web prebundle tests (11/11), Dev Vendor tests (28/28), and the local commit gate passed.

sidmorizon (approved):

originalix (approved):

PR #13186, opened 2026-09-04
fix: recover Market list after request failure

huhuanming:
## Summary

- Keep the Market network-switch skeleton active until a successful token-list response has been transformed and committed.
- Reset failed latest requests, clear rows from the previous scope, and then release the switching state so the page cannot remain on its skeleton indefinitely or expose stale tokens.
- Add real-hook regression coverage for both the successful transition and a failed network switch.

## Reproduction

1. Launch Desktop with a fresh profile.
2. Fail the first `/utility/v2/market/token/list` request.
3. Before this change, the Market page remains on the skeleton after the request has settled and network access is restored.

## Verification

- Full Electron build and isolated-profile runtime test.
- Successful cold request: no settled empty-state frame before transformed rows render.
- Injected Robinhood token-list failure after Hot data loaded: empty state renders without exposing the previous rows.
- Restored requests: Robinhood rows render without restarting the App.
- `yarn tsc:only`
- `yarn tsc:staged`
- 4 focused Jest suites, 34 tests passed.
- Type-aware Oxlint: 0 warnings, 0 errors.
- Prettier and `git diff --check` passed.

## Environment note

`yarn agent:check --profile commit` cannot spawn its Windows child processes in this environment (`exitCode: null` with no code diagnostics). The underlying type, lint, format, and focused test commands were run directly and passed. The repository `lint:staged` shell pipeline also requires Unix `xargs`/`grep`, which are unavailable in this PowerShell environment.

### DIFF
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
@@ -461,11 +461,7 @@ export function useMarketTokenList({
     {
       checkIsFocused: !platformEnv.isWeb,
       watchLoading: hasNetworkId,
-      onIsLoadingChange: (nextIsLoading) => {
-        if (!nextIsLoading) {
-          setIsNetworkSwitching(false);
-        }
-      },
+      undefinedResultIfError: true,
       pollingInterval,
       revalidateOnFocus: true,
       revalidateOnReconnect: true,
@@ -492,6 +488,15 @@ export function useMarketTokenList({
       timeRange: timeRangeRef.current,
     }),
   );
+
+  useEffect(() => {
+    // Successful results keep the switching skeleton until the transform effect
+    // commits their rows. A settled error must instead drop the previous scope.
+    if (hasNetworkId && isLoading === false && apiResult === undefined) {
+      setTransformedData([]);
+      setIsNetworkSwitching(false);
+    }
+  }, [apiResult, hasNetworkId, isLoading]);
 
   const effectiveIsLoading = hasNetworkId ? isLoading : false;
   const isSeedResult = Boolean(apiResult?.__fromSeed);
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
@@ -461,6 +461,11 @@ export function useMarketTokenList({
     {
       checkIsFocused: !platformEnv.isWeb,
       watchLoading: hasNetworkId,
+      onIsLoadingChange: (nextIsLoading) => {
+        if (!nextIsLoading) {
+          setIsNetworkSwitching(false);
+        }
+      },
       pollingInterval,
       revalidateOnFocus: true,
       revalidateOnReconnect: true,
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
@@ -488,15 +488,29 @@ export function useMarketTokenList({
       timeRange: timeRangeRef.current,
     }),
   );
+  const previousIsLoadingRef = useRef(isLoading);
 
   useEffect(() => {
+    const wasLoading = previousIsLoadingRef.current;
+    previousIsLoadingRef.current = isLoading;
+
+    if (isLoading === true && transformedData.length === 0) {
+      setIsNetworkSwitching(true);
+      return;
+    }
+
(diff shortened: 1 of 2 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-650c2d8a-6c09-56dc-845d-199f2b3bbc26)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -461,6 +461,11 @@ export function useMarketTokenList({
     {
       checkIsFocused: !platformEnv.isWeb,
       watchLoading: hasNetworkId,
+      onIsLoadingChange: (nextIsLoading) => {
+        if (!nextIsLoading) {
+          setIsNetworkSwitching(false);
+        }
```
sidmorizon: **Failed switches reveal stale tokens**

Severity: `non-severe`

When a network switch request fails, `onIsLoadingChange` clears the switch state without removing the previous network’s tokens. Mobile users then see tokens from the network they left.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -461,6 +461,11 @@ export function useMarketTokenList({
     {
       checkIsFocused: !platformEnv.isWeb,
       watchLoading: hasNetworkId,
+      onIsLoadingChange: (nextIsLoading) => {
+        if (!nextIsLoading) {
+          setIsNetworkSwitching(false);
```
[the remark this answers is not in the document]
huhuanming: Fixed in e725444b51. The unconditional onIsLoadingChange callback is removed, so successful requests keep isNetworkSwitching=true until the existing transform effect commits the new rows. I also replaced the mocked hook test with a real deferred successful request and assert that no settled empty-state frame occurs.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
huhuanming: Fixed in e725444b51. Failed latest requests now reset apiResult to undefined; once loading settles, the hook clears transformedData before releasing isNetworkSwitching, so tokens from the previous network are not exposed. Added a real network-switch failure regression test and verified the same path in Electron with request interception and recovery.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -488,6 +489,15 @@ export function useMarketTokenList({
     }),
   );
 
+  useEffect(() => {
+    // Successful results keep the switching skeleton until the transform effect
+    // commits their rows. A settled error must instead drop the previous scope.
+    if (hasNetworkId && isLoading === false && apiResult === undefined) {
+      setTransformedData([]);
+      setIsNetworkSwitching(false);
```
sidmorizon, line 511: **Recovered lists flash empty state**

Severity: `non-severe`

After an initial failure, a successful retry leaves `isNetworkSwitching` false while `isLoading` clears before transformed rows commit. Users briefly see an empty Market list between the retry skeleton and recovered tokens.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -488,6 +489,15 @@ export function useMarketTokenList({
     }),
   );
 
+  useEffect(() => {
+    // Successful results keep the switching skeleton until the transform effect
+    // commits their rows. A settled error must instead drop the previous scope.
+    if (hasNetworkId && isLoading === false && apiResult === undefined) {
```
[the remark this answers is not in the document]
huhuanming: Fixed in 95fe6b0e4c. The failure effect now requires the current loading cycle to transition from true to false before treating undefined apiResult as a settled failure. The network-switch test now also asserts that the previous rows remain present while the new request is still pending.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
huhuanming: Fixed in 95fe6b0e4c. When a request starts with no transformed rows, the hook restores isNetworkSwitching until either failure settles or the successful response is transformed. Added a failure-then-successful-retry render-sequence test that rejects any settled empty-state frame.

sidmorizon:
Claude review session: https://claude.ai/code/session_01EJTqNFSnV85YSjA1pf58MN

originalix (approved):

PR #13136, opened 2026-09-02
feat: integrate the DeviceStage hardware interaction stage (OK-59934, OK-61489, OK-59933, OK-61489, OK-59932)

originalix:
[OK-59934](https://onekeyhq.atlassian.net/browse/OK-59934) [OK-61489](https://onekeyhq.atlassian.net/browse/OK-61489) [OK-59933](https://onekeyhq.atlassian.net/browse/OK-59933) [OK-59932](https://onekeyhq.atlassian.net/browse/OK-59932)

----------

## Overview

Integrates the DeviceStage hardware-interaction surface (OK-59934), replacing four dialog-based hardware UIs: the OneKey track's `HardwareUiStateContainer`, the third-party track's `ThirdPartyHardwareUiStateContainer`, the authenticity check's `FirmwareVerifyDialog`, and the air-gap `AirGapQrcodeDialogContainer` (two-way flow only). The firmware-update page, full-page onboarding, permission dialogs and Trezor BLE pairing stay as they are, per the spec.

Rides along with OK-61489: button devices now default to entering the PIN on the device, with a "switch back to app entry" affordance inside the stage. Franco's three component commits are cherry-picked.

## Architecture

- **Burst scope** (`kit-bg/ServiceHardwareUI/DeviceStageBurst.ts`): one run of consecutive hardware requests is one entrance and one exit for the stage. `off` has exactly three sources — the 600 ms grace window after a burst ends, a user close, and a closed error outcome. Every call-end close inside a burst only swaps the step. `begin`/`end` bracket `withHardwareProcessing`; flows that span several wrappers (onboarding, the authenticity runner, the teaching card) hold an explicit join/leave token.
- **Legacy containers are silenced, not deleted**: `isLegacyHardwareUiActive()` is a constant switch, so the legacy code still compiles and the surface can be restored in one line. A cleanup PR follows after merge.
- **Failures land on the stage**: SDK errors map to an `errorReason` (rejected / pinInvalid / disconnected / busy / generic); generic repeats the localized message the hardware error already carries. The wrapper suppresses toasts for hardware errors. Disconnection is decided by a bounded connection probe rather than by enumerating error codes.
- **deviceNotFound**: a device that is absent when the call starts gets a "Device not connected" card with the troubleshooting-article and Intercom buttons, worded identically to the shipped dialog. A mid-flow unplug keeps the disconnected notice. Legacy dialog events are suppressed at the constructor level, which removes the double-dialog race.
[...]

PR #13214, opened 2026-09-07
feat: play the wallet-type fork on the DeviceStage and rework its exit policy (OK-59934)

franco-chan:
[OK-59934](https://onekeyhq.atlassian.net/browse/OK-59934)

----------

## Summary

Three commits on the DeviceStage (OK-59934), the hardware interaction overlay that replaced the legacy hardware dialogs in #13136.

**1. The wallet-type fork plays on the stage** (`91ac4aeffb`)
Onboarding's "Standard or Hidden wallet" question, previously the `SelectAddWalletTypeDialog`, is now a `selectWalletType` step on the stage: the same two option cards, one callback, and the app-side flow awaits the answer over the event bus. The legacy dialog stays as the fallback while the stage is disabled. Onboarding's fork deliberately does not show the passphrase teach card — that card belongs to the account selector's Add-hidden-wallet entry only.

**2. Exit policy reworked around who the stage waits on** (`1f77fe5ef4`)
Design hard rule #3, revised, and the fix for OK-62065 / OK-62073:
- Ask cards (confirm, PIN, passphrase, the third-party "do this on the device" beats) open every exit — close button, drag, Escape, Android back — one second after the stage appears, for the rest of that appearance.
- Machine waits (connecting, processing, searching, an app install) show the close only once stalled: the device silent for 10s, or the wait past a 30s cap. The connecting capsule swaps its device name for a transport-specific hint once stalled (two new Lokalise keys, 19 languages).
- Escape and back on the wait that follows an answered card also follow the stall clock, so a habitual back press cannot throw away a confirmation just given on the device.
- Outcomes, decisions, the teach card, the authenticity run and the air-gap QR pair open at once; the done beat never closes.
- The policy and its clocks live in `shared` (`resolveDeviceStageExitGrant`, `resolveDeviceStageWaitStall`) with tests; bg tracks device activity per burst (`activitySeq`); two analytics events (`deviceStageClosed`, `deviceStageWaitEnded`) for tuning the thresholds.

**3. The vanishing fork card, plus review fixes** (`df558e9737`)
The fork card sometimes disappeared before the person chose, and onboarding then sat on "Connecting your device". Root cause: hd-core interrupts a still-running call on the same device with `DeviceInterruptedFromOutside`, and the burst treated that code as the person's own cancel, forcing the stage off by a route that announced nothing. Fixed on three sides: only genuine user-cancel codes close the stage from a call's end; a late call-end close no longer paints over an app-authored card; every stage exit now emits `DeviceStageOff`, which the answer wait listens on. Regression tests for all three.

Also from the review pass on this branch:
- A wait is one stretch on one wait step: moving to another wait step restarts the clocks and clears a fired stall, so a stall earned by the previous call cannot open the next flow's connecting with the "device is silent" hint.
- The teach card's Continue counts as an answer, so a stray Escape/back right after it cannot cancel the hidden-wallet call.
- The overlay shell's edge is the dark theme's neutral3 on every app theme (it was reading the ambient theme, which showed as a pale halo in light mode).
- The Lazy fallback tests follow the close call's new shape.

## Notes for reviewers

- `apps/mobile/bundle-registry/module-id-registry.json` registers the two new source files (`useDeviceStageExitPolicy.ts`, `waitForDeviceStageAnswer.ts`).
- Locale JSON / `translations.ts` changes come from a Lokalise pull; the pull also brought five pending `transaction_security_*` keys that are not this PR's.
- Two latent items are documented but intentionally left out: the v1 onboarding page (WebDapp mode) paints the fork at burst depth 0 with no card protection (the right fix is for that page to hold a burst like v2), and a bystander SDK ask could in theory still replace the fork card (no live bystander exists at that point today).

## Test plan

- [x] jest: `deviceStageOwnership`, `DeviceStageBurst`, `Lazy.fallback`, `ServiceHardwareUI.sendUiResponse` suites pass
- [ ] Desktop + iOS, passphrase-enabled device: run onboarding to the fork several times; the card stays until answered
- [ ] Ask cards (PIN, confirm) closable ~1s after appearing; connecting stalled >10s shows the X and the transport hint; after answering a card, the following processing swallows Esc/back until stalled
- [ ] Chain two flows: after a stalled connecting, the next flow's connecting opens without the hint
- [ ] Add hidden wallet from the account selector: teach card → Continue → Esc/back during the wait does not cancel the call
- [ ] Light mode, desktop and phone: shell edge reads as part of the dark shell
- [ ] Storybook Console / Passphrase / Flow scenes

[OK-59934]: https://onekeyhq.atlassian.net/browse/OK-59934?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix:

Review of commit `df558e9737d0` did not finish successfully.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-b3d61447-e519-52ae-8e61-c187e6b27308)

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
```diff
@@ -85,6 +87,24 @@ export function useSelectAddWalletTypeDialog() {
...
+    // selectWalletType card — the stage is already up (onboarding holds
+    // its burst across the whole creation), so the card morphs in place
+    // and the answer rides back through the driver; closing the stage
+    // cancels, the way this dialog's close did. A silenced stage (the
+    // firmware workflow) does not paint the card, and this dialog, with
+    // the iOS layering hack below, remains for the legacy surface until
+    // the cleanup pass.
+    if (
+      !isLegacyHardwareUiActive() &&
+      (await backgroundApiProxy.serviceHardwareUI.deviceStageShowSelectWalletType())
```
sidmorizon: 🟠 P1: Legacy onboarding loses wallet choice

Severity: `severe`

Legacy onboarding calls `deviceStageShowSelectWalletType` without a burst, so a trailing SDK close can dismiss the wallet-choice card. Unlike [the v2 page](packages/kit/src/views/Onboardingv2/pages/ConnectYourDevice.tsx:1029), [the legacy flow](packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/ConnectYourDevice.tsx:1295) never owns `useDeviceStageBurst`. The flow then waits forever for a choice users cannot submit.

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
```diff
@@ -652,6 +655,34 @@ class ServiceHardwareUI extends ServiceBase {
...
+   */
+  @backgroundMethod()
+  async deviceStageShowSelectWalletType(params: { connectId?: string } = {}) {
+    if (!(await this.deviceStageBurst.isEnabled())) {
+      return false;
+    }
+    await this.deviceStageBurst.noteStep('selectWalletType', {
+      connectId: params.connectId,
+    });
+    return true;
```
sidmorizon: **Firmware takeover strands wallet selection**

Severity: `non-severe`

If firmware starts after `isEnabled`, `noteStep` skips the card but this method still returns `true`. [`showSelectAddWalletTypeDialog`](packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx:87) then waits for an answer after the close notification can already be lost.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
franco-chan: Fixed in 272c04c222: the legacy ConnectYourDevice page now holds one DeviceStage burst across the whole run (device state → fork → wallet creation), released in a finally with the error, the same way the v2 page does. Defence in depth in the background too — an unanswered app-authored card now outranks a call-end close at any depth, not only while a burst is held, with a regression test for the depth-0 path.

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
franco-chan: Fixed in 272c04c222: noteStep now reports whether it actually painted, and deviceStageShowSelectWalletType returns that instead of answering from an isEnabled() check taken before the paint, so the firmware workflow raising its flag in between makes the caller fall back to the legacy dialog rather than wait. Same treatment for deviceStageShowPassphraseIntro, whose caller had the identical hang, plus a regression test.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
```diff
@@ -85,6 +87,24 @@ export function useSelectAddWalletTypeDialog() {
...
+    ) {
```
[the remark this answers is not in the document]
franco-chan: Already fixed in 272c04c222, pushed before this review landed (it was against c4031cbcab). The legacy ConnectYourDevice page now holds one burst across the whole run — device state, fork, wallet creation — released in a finally with the error, the same shape as the v2 page. The background also no longer relies on a burst being held: an unanswered app-authored card outranks a call-end close at any depth, with a regression test for the depth-0 path.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
```diff
@@ -85,6 +87,24 @@ export function useSelectAddWalletTypeDialog() {
...
+      const fork = await waitForDeviceStageAnswer(
```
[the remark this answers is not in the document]
franco-chan: Good catch — fixed in 8f89728a39. New watchForDeviceStageAnswer registers the listeners synchronously and returns the promise plus a release, so both callers now listen first, ask the background to paint second, and release when the card never landed. Covers the split-runtime window on iOS/Android/extension and the await-yield window on desktop/web. Tests added for an exit announced before the caller awaits, the normal answer, and the release path.

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
```diff
@@ -917,6 +947,24 @@ export class DeviceStageBurstScope {
         // wait the person can make nothing of.
         return;
       }
+      if (current && APP_AUTHORED_ASK_STEPS.has(current.step)) {
```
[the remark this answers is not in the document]
franco-chan: Confirmed and fixed in 60f02fbbd5 — you are right that the rewritten PIN close never reaches the close-branch guard, and the askCompleted beat skipped the progress gate. The protection now sits ahead of every repaint gate on the SDK-event path, so ticks, the narrative re-assert and the askCompleted beat all leave an unanswered app-authored card alone. Kept the exception you implied: another ASK still outranks the card, since that means the device itself is waiting on the person and a request nobody can answer would strand the call that made it. Regression test covers ProcessLoading + askCompleted over selectWalletType, plus the REQUEST_PIN case that must still get through.

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
sidmorizon, line 1: 🟠 P1: Firmware takeover can resurrect wallet prompt

Severity: `severe`

If firmware takeover begins after the `dismissSeq` check, the asynchronous write can restore the wallet prompt over the update page. On split-runtime targets, the flow then waits for an inaccessible answer while the prompt blocks the update UI.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
```diff
@@ -85,6 +87,32 @@ export function useSelectAddWalletTypeDialog() {
...
+    if (!isLegacyHardwareUiActive()) {
+      // Listening starts before the card is asked for: the paint is an
+      // RPC, and an exit the background announces while it is in flight
+      // would otherwise reach a main runtime that is not listening yet.
+      const fork = watchForDeviceStageAnswer(
+        EAppEventBusNames.DeviceStageWalletTypeSelected,
+        ({ walletType }) => (walletType === 'hidden' ? 'Hidden' : 'Standard'),
+      );
+      const painted =
+        await backgroundApiProxy.serviceHardwareUI.deviceStageShowSelectWalletType();
```
sidmorizon: **Rejected card requests leak event listeners**

Severity: `non-severe`

If `deviceStageShowSelectWalletType` rejects, the preinstalled `fork` listeners remain because cleanup only follows a false result. The same leak affects [`deviceStageShowPassphraseIntro`](packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/hooks/useAddHiddenWallet.tsx:229-243). Repeated bridge failures retain abandoned flows until another stage event arrives.

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
franco-chan: Confirmed and fixed in 5527d97bb3. noteStep was the one write path that never sampled dismissSeq — the SDK-event path, the third-party feed and the vendor close all do — so a takeover landing across either of its awaits got the card written back over the update page, and the true return left the caller waiting on it. noteStep and noteCardAnswered now sample on entry and re-check before the write, and noteStep reports the refusal through the "did the card land" contract added earlier in this branch, so the fork and the teach card fall back instead of waiting. Regression test drives the takeover from inside the enablement read.

packages/kit/src/views/Onboarding/pages/ConnectHardwareWallet/SelectAddWalletTypeDialog.tsx
franco-chan: Confirmed and fixed in ade1b3b3a6. The release now sits in a finally at both call sites, so it covers the throw as well as the false result — previously only a false result reached cancel(), leaving the abandoned run subscribed and waking on the next stage event. Added a test for the failing-call path that asserts the bus is empty afterwards.

originalix:

Review of commit `60f02fbbd54f` did not finish successfully.

originalix (approved):

originalix:

Review of commit `3d71da0600a4` did not finish successfully.

PR #12533, opened 2026-07-17
fix: purge delisted networks from aggregate token cache and refresh stale wallet config OK-58112

weatherstar:
[OK-58112](https://onekeyhq.atlassian.net/browse/OK-58112)

----------

## Problem

With All Networks enabled, opening an aggregate token's details page (e.g. USDT) and then the "Contract Address" popover shows blank rows — only the copy/explorer buttons render, with no network icon or name.

## Root cause

- `/wallet/v1/wallet/config` responses are persisted into `simpleDb.aggregateToken.allAggregateTokenMap`. On app versions before #10373 (removed 21 low-usage built-in EVM networks, 2026-02), networks like Gnosis (`evm--100`), B² (`evm--223`), Polygon zkEVM (`evm--1101`) and zkLink Nova (`evm--810180`) were LISTED presets, so USDT entries on them were cached.
- After upgrading, those networks no longer exist in the bundle, so `getNetworkSafe` resolves to `undefined` → empty network name and a `null` `NetworkAvatar`, while the address-driven buttons still render → blank rows.
- The cache never heals: `syncWalletConfig` was only triggered when `aggregateTokenConfigMap` was completely missing, so existing users never re-synced and the stale map survived indefinitely (server-side cleanup alone cannot fix affected users either).

## Changes

Symptom fixes (render-side tolerance):
- `TokenDetails`: skip aggregate tokens whose network cannot be resolved when building the contract-address list.
- `ServiceToken.getAllAggregateTokenInfo`: filter cached aggregate tokens against the current bundled LISTED preset networks, healing all consumers of this accessor (receive flow, universal search, home token list).

Root fix (cache refresh):
- Record `configSyncMeta` (`appVersion` + `syncedAt`) on every successful `syncWalletConfig`.
- New `ServiceSetting.syncWalletConfigIfNeeded`: re-syncs when the config was never synced, the app version changed (bundled preset network list may differ), or the cache is older than 24h; concurrent calls are deduped.
- `TokenListBlock` fires it in the background on all-networks refresh, so the full-replace sync (`merge=false`) purges delisted networks from the persisted maps.

## Verification

- Live prod `/wallet/v1/wallet/config` still returns 10 networkIds absent from current presets (4 of them removed by #10373 carry USDT — matching the 4 blank rows in the report).
- New unit tests: `ServiceToken.getAllAggregateTokenInfo` filtering and `ServiceSetting.syncWalletConfigIfNeeded` decision matrix (9 tests, all passing).
[...]

PR #13219, opened 2026-09-07
feat: Robinhood network defaults and stale wallet config refresh on v6.5.2(OK-62300 OK-58112)

weatherstar:
[OK-62300](https://onekeyhq.atlassian.net/browse/OK-62300) [OK-58112](https://onekeyhq.atlassian.net/browse/OK-58112)

----------

## Summary
- All Networks now enables the server-delivered Robinhood chain (`evm--4663`) by default, via a new id-based default-enabled helper.
- The single-network selector's default top list pins Robinhood above Bitcoin.
- Adds unit tests for the All Networks default-enabled set and the Robinhood opt-out.
- Aggregate-token config sync (`/wallet/v1/wallet/config`) now keeps members on server-delivered networks: the eligibility filter uses the merged network registry instead of the preset-only listed map, so Robinhood ETH/USDG show up inside the ETH/USDG aggregate tokens once the server config lists them.
- The token selector's aggregate network list (Receive → ETH → choose network) now orders zero-balance members by the server `order` like the aggregate detail page does: `buildAggregateTokenListData` stamps `order` / `commonSymbol` / `networkName` / `logoURI` on every folded member instead of only the first one.

- Backports the stale wallet-config refresh from #12533 (OK-58112) to this hot-update line and extends it: `/wallet/v1/wallet/config` is re-fetched when the cache is older than 1 day, when the app version changes, or when the hot-update bundle version changes (`platformEnv.bundleVersion`), via `configSyncMeta` on `simpleDb.aggregateToken` and a single-flight `serviceSetting.syncWalletConfigIfNeeded()` kicked off in the background from the home token list. Without it, existing 6.5.2 users would never pick up the Robinhood entries the aggregate fix above depends on.

## Intent & Context
Jira OK-62300 ("Robinhood 网络调整") asks for two product changes on the 6.5.2 hot-update line: (1) All Networks should include Robinhood by default, and (2) in single-network mode Robinhood should sit at the top of the default network list, above Bitcoin. The user asked for the work to be based on `release/v6.5.2`.

Before this change the client only default-enabled the static preset list in `getDefaultEnabledNetworksInAllNetworks()`, and the pinned top list was the hardcoded `defaultPinnedNetworkIds` in `ServiceNetwork`. Robinhood is not a preset network (it is delivered by `/wallet/v1/network/list`), so it could not participate in either list even though the server already reports `defaultEnabled: true` for it — the client never consumed that server flag.

Wallet config lifecycle: on `release/v6.5.2` the config was fetched exactly once (the home All Networks run, gated on the cache being absent) and never refreshed until an app reset, because the branch forked from `x` on 2026-07-13 before #12533 landed. Hot-update bundles all carry `VERSION=6.5.2`, so even #12533's app-version check would not fire across hot updates; only `BUNDLE_VERSION` changes.

## Design Decisions
- **Aggregate members are gated on `serviceNetwork.getAllNetworks({ excludeCustomNetwork, excludeAllNetworkItem })`, not `getListedNetworkMap()`.** The preset-only map silently dropped every non-preset member (`evm--4663`, and also 34443 / 48900 / 1101 / 7777777 on the test server), which is why the ETH aggregate never showed Robinhood even though the app's own client received it. The merged registry keeps preset status for presets, server status for server-delivered chains, already drops TRASH networks, and excludes user custom RPC networks; the server list refreshes hourly, which bounds the stale-cache window #12533 worried about. Note for the x sync: `ServiceToken.getAllAggregateTokenInfo` on `x` (#12533) applies the same preset-only filter at read time and needs the same change.
- **Reference Robinhood by id, do not promote it to `presetNetworks`.** `getAllNetworks` merges presets first and is first-wins, so a preset copy would permanently mask server-side flag updates such as `backendIndex`. This follows the pattern already used on `x` for the scam-history and approval switches (#12718).
- **New `getDefaultEnabledNetworkIdsInAllNetworks()` helper instead of consuming the server `defaultEnabled` flag.** `isEnabledNetworksInAllNetworks` is called from ~20 sites with only a network id; switching them to the server flag would be a much larger change and would also silently alter behavior for every other server network. The id helper keeps the existing preset list intact and appends Robinhood.
- **Opt-out semantics unchanged.** Robinhood behaves like the other default-enabled ids: `disabledNetworks['evm--4663']` still wins, so users who previously turned it off stay off.
- **`ROBINHOOD_NETWORK_ID` is exported and declared at the same position `x` uses for its private const.** The release→x sync will surface a visible merge conflict at that spot rather than a silent duplicate declaration; resolve by keeping a single exported constant.
- **Pinned list falls back gracefully.** The pinned ids resolve through `getNetworksByIds` → `getAllNetworks`, so on a cold install Robinhood appears at the top only once the server network list has been fetched; before that the list is simply the previous order.
- **Only the TTL/refresh slice of #12533 is backported.** #12533 also degraded stale aggregate-token groups in `TokenDetails` / `TokenSelector` / `ServiceToken.getAllAggregateTokenInfo` for delisted networks. Those hunks conflict with the 6.5.2 `TokenDetails` page and are a separate concern, so they are intentionally left out; `syncWalletConfig` on 6.5.2 already filters incoming tokens through `getListedNetworkMap()`, so a refreshed config is written clean.
- **Same semantics as `x`, plus bundle awareness.** Re-sync when `aggregateTokenConfigMap` is missing, when `configSyncMeta` is missing (existing 6.5.2 caches), when `appVersion` differs, when `bundleVersion` differs, or when `syncedAt` is more than 1 day old. Concurrent callers share one promise.
- **Why `bundleVersion`.** Hot-update bundles on this line are all built with `VERSION=6.5.2`; only `BUNDLE_VERSION` (`platformEnv.bundleVersion`) changes. Without it, a hot update that ships a changed preset network list would keep the stale config for up to 24 h. `x` has the same gap; a cache without `bundleVersion` is treated as stale only when the running build has one, so web/desktop builds without a bundle version do not refetch spuriously.
- **Non-blocking on the hot path.** The first fetch (no cache) still awaits so the home list renders with config; the stale-cache refresh is fire-and-forget and the next home refresh picks up the new data. A failed background refresh is swallowed and retried on the next run.
- **No schema migration needed.** `configSyncMeta` is an optional field on an existing simpleDb entity; caches without it are treated as stale and refreshed once.

## Changes Detail
- `packages/shared/src/utils/tokenUtils.ts`: `buildAggregateTokenListData` enriches later members of an aggregate group with the config metadata. The home path folds each network into a fresh map (every member was "first"), but the selector self-fetch (`buildSelectorTokenListFromResponses`) accumulates one map across responses, so only the first-arriving member carried `order` and `sortTokensByOrder` left the rest in response-arrival order — Robinhood, enabled last, landed last.
- `packages/shared/src/utils/tokenUtils.buildAggregateTokenListData.test.ts`: two networks folded into one map both carry order/commonSymbol/networkName/logoURI/accountId.
- `packages/shared/src/config/presetNetworks.ts`: export `ROBINHOOD_NETWORK_ID`; add `getDefaultEnabledNetworkIdsInAllNetworks()` (preset defaults + Robinhood).
- `packages/shared/src/utils/networkUtils.ts`: build `defaultEnabledNetworkIds` from the new id helper.
- `packages/kit-bg/src/services/ServiceNetwork/ServiceNetwork.ts`: prepend `ROBINHOOD_NETWORK_ID` to `defaultPinnedNetworkIds` (Robinhood → Bitcoin → Lightning → Ethereum → Tron → Solana → BNB Chain → Polygon → TON).
- `packages/shared/src/utils/networkUtils.allNetworksDefaults.test.ts`: new tests — helper composition, Robinhood default-enabled, explicit opt-out, non-default mainnets stay opt-in, testnets never default.
- `development/spellCheckerSkipWords.txt`: add `robinhood`.
- `packages/kit-bg/src/services/ServiceSetting.ts`: `syncWalletConfig` filters aggregate members through the merged network registry (`keyBy(getAllNetworks(...).networks, 'id')`).
- `packages/kit-bg/src/services/ServiceSetting.syncWalletConfig.test.ts`: server-only member kept, unknown member dropped, single survivor is not an aggregate.
- `packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityAggregateToken.ts`: add `configSyncMeta { appVersion, bundleVersion?, syncedAt }` to the struct and to `updateAllAggregateInfo` (kept when not supplied).
- `packages/kit-bg/src/services/ServiceSetting.ts`: `syncWalletConfig` stamps `configSyncMeta` (app + bundle version); new `syncWalletConfigIfNeeded()` with the staleness rule and single-flight promise.
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx`: when the cached config exists, kick off `syncWalletConfigIfNeeded()` in the background.
- `packages/kit-bg/src/services/ServiceSetting.syncWalletConfigIfNeeded.test.ts`: never synced / no meta / app version changed / bundle version changed / meta without bundle version / TTL expired / fresh / concurrent dedupe.

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**:
  - All Networks aggregation now includes one more EVM chain per EVM account (extra `evm--4663` requests in home balance/token/history fan-out). The server entry currently reports `backendIndex: true`, `status: LISTED`.
  - The single-network top section keeps its existing value-based override (`sortChainSelectorNetworksByValue`, >$1): for funded accounts the top section is sorted by value and Robinhood only appears there when it has value. The new default order applies to zero-value accounts (the case shown in the Jira screenshot).
  - Existing users who explicitly enabled Robinhood are unaffected; users who explicitly disabled it remain disabled.
  - One extra `/wallet/v1/wallet/config` request per install on the first All Networks home run after this hot update (existing caches have no `configSyncMeta`), then at most one per day / per new bundle. `syncWalletConfig` replaces the aggregate maps wholesale, the same write path used for the first fetch.
  - Production `wallet.onekeycn.com` does not list `evm--4663` in the wallet config yet (the test server does); the aggregate change is inert until the backend ships it. On a cold install where the server network list fails to load, that sync drops server-only members until the next config sync.

## Test plan
- [x] `yarn jest packages/shared/src/utils/networkUtils.allNetworksDefaults.test.ts` — 5/5 pass; `packages/shared/src/utils` + `ServiceNetwork` suites — 63 suites / 1328 tests pass.
- [x] `yarn agent:check --profile commit` — lint-worktree-ts / lint-staged / tsc-staged PASS.
- [x] Desktop dev over CDP: `serviceNetwork.getNetworkSelectorPinnedNetworks({ useDefaultPinnedNetworks: true })` returns `evm--4663:Robinhood` first, then `btc--0:Bitcoin`; `getChainSelectorNetworksCompatibleWithAccountId().frequentlyUsedItems` matches.
- [x] Desktop UI: All Networks tab shows Robinhood checked with an empty persisted state (19 selected = 18 presets + Robinhood); Single network tab for a zero-value HD account lists Robinhood above Bitcoin.
- [ ] Fresh install: after the server network list loads, Robinhood is the first item in the single-network top list; All Networks includes Robinhood without any user action.
- [ ] Turn Robinhood off in All Networks, restart the app: it stays off.
- [ ] Funded account: the single-network top section is still value-sorted; Robinhood appears there only with balance.
- [ ] Mobile (iOS/Android) and extension: same selector order and All Networks default.
- [x] `yarn jest packages/kit-bg/src/services/ServiceSetting.syncWalletConfig.test.ts` — 2/2 pass.
- [ ] Against the test endpoint: ETH aggregate token detail lists Robinhood as a member; with Robinhood enabled in All Networks its ETH balance is included.
- [x] Desktop dev, zero-balance account: Receive → 接收转账 → ETH → network list is Ethereum, Robinhood, Base, Arbitrum, … (server order), matching the aggregate detail page; funded account keeps value-first ordering.
- [x] `yarn jest packages/kit-bg/src/services/ServiceSetting.syncWalletConfigIfNeeded.test.ts` — 8/8 pass.
- [x] Local desktop dev (working tree hot-reloaded onto this branch): first All Networks home run stamped `configSyncMeta` once (08:21), no further request on later reloads.
- [x] `yarn agent:check --profile commit` — lint-worktree-ts / lint-staged / tsc-staged PASS.
- [ ] Upgrade from a 6.5.2 build with a cached config: first All Networks home load issues one `/wallet/v1/wallet/config` request in the background, `simpleDb.aggregateToken.configSyncMeta` gets stamped.
- [ ] Reopen within 24 h: no additional config request; after 24 h, after a native upgrade, or after installing a new hot-update bundle: one request.
- [ ] Fresh install: first All Networks load still awaits the config before rendering aggregate tokens.

[OK-62300]: https://onekeyhq.atlassian.net/browse/OK-62300?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-58112]: https://onekeyhq.atlassian.net/browse/OK-58112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceSetting.ts
+++ b/packages/kit-bg/src/services/ServiceSetting.ts
@@ -43,6 +43,7 @@ import {
   buildAggregateTokenListMapKeyForTokenList,
   buildAggregateTokenMapKeyForAggregateConfig,
   buildHomeDefaultTokenMapKey,
+  sortTokensByOrder,
 } from '@onekeyhq/shared/src/utils/tokenUtils';
 import type {
   EHardwareTransportType,
@@ -991,6 +992,14 @@ class ServiceSetting extends ServiceBase {
         }
       },
     );
+
+    // The server array is not guaranteed to follow `order` (prod USDG arrives
+    // as Ethereum, Robinhood, X Layer, Solana with orders 3, 1, 2, 4) and the
+    // Receive search flatten keeps array order, so sort members once here.
+    // Members without `order` sink to the end.
+    Object.values(allAggregateTokenMap).forEach((group) => {
+      group.tokens = sortTokensByOrder({ tokens: group.tokens });
+    });
 
     const allAggregateTokens: IAccountToken[] = Object.keys(
       allAggregateTokenMap,
--- a/packages/kit-bg/src/services/ServiceSetting.ts
+++ b/packages/kit-bg/src/services/ServiceSetting.ts
@@ -1008,6 +1008,10 @@ class ServiceSetting extends ServiceBase {
         homeDefaultTokenMap,
         allAggregateTokenMap,
         aggregateTokenSymbolMap,
+        configSyncMeta: {
+          appVersion: platformEnv.version ?? '',
+          syncedAt: Date.now(),
+        },
       }),
       this.backgroundApi.simpleDb.approval.updateApprovalResurfaceDaysConfig({
         approvalResurfaceDays,
@@ -1016,6 +1020,37 @@ class ServiceSetting extends ServiceBase {
     ]);
 
     return aggregateTokenConfigMap;
+  }
+
+  _syncWalletConfigIfNeededPromise: Promise<void> | undefined;
+
+  @backgroundMethod()
+  public async syncWalletConfigIfNeeded() {
+    if (this._syncWalletConfigIfNeededPromise) {
+      return this._syncWalletConfigIfNeededPromise;
+    }
+    this._syncWalletConfigIfNeededPromise = (async () => {
+      const rawData =
+        await this.backgroundApi.simpleDb.aggregateToken.getRawData();
+      const configSyncMeta = rawData?.configSyncMeta;
+      const appVersion = platformEnv.version ?? '';
+      // Re-sync when the config has never been synced, when the app version
+      // changed (the bundled preset network list may differ, leaving stale
+      // networks in the cached aggregate-token maps), or when the cache is
+      // older than the TTL.
+      const shouldSync =
+        !rawData?.aggregateTokenConfigMap ||
+        !configSyncMeta ||
+        configSyncMeta.appVersion !== appVersion ||
+        Date.now() - configSyncMeta.syncedAt >
+          timerUtils.getTimeDurationMs({ day: 1 });
+      if (shouldSync) {
+        await this.syncWalletConfig();
+      }
+    })().finally(() => {
+      this._syncWalletConfigIfNeededPromise = undefined;
+    });
+    return this._syncWalletConfigIfNeededPromise;
   }
 
   @backgroundMethod()
--- a/packages/kit-bg/src/services/ServiceSetting.ts
+++ b/packages/kit-bg/src/services/ServiceSetting.ts
@@ -1,5 +1,5 @@
 import { consts } from '@onekeyfe/cross-inpage-provider-core';
-import { flatten, groupBy, isEqual, uniqBy } from 'lodash';
+import { flatten, groupBy, isEqual, keyBy, uniqBy } from 'lodash';
 import semver from 'semver';
 
 import {
@@ -12,10 +12,7 @@ import {
   backgroundMethod,
   toastIfError,
 } from '@onekeyhq/shared/src/background/backgroundDecorators';
-import {
-  getListedNetworkMap,
-  getNetworkIdsMap,
-} from '@onekeyhq/shared/src/config/networkIds';
+import { getNetworkIdsMap } from '@onekeyhq/shared/src/config/networkIds';
 import {
   IMPL_BTC,
   IMPL_EVM,
@@ -900,7 +897,17 @@ class ServiceSetting extends ServiceBase {
     const aggregateTokenConfigMap: Record<string, IAggregateToken> = {};
     const homeDefaultTokenMap: Record<string, IHomeDefaultToken> = {};
     const aggregateTokenSymbolMap: Record<string, boolean> = {};
-    const listedNetworkMap = getListedNetworkMap();
+    // Aggregate members may live on server-delivered chains (e.g. Robinhood)
+    // that are not part of presetNetworks, so gate on the merged network
+    // registry instead of the preset-only listed map: presets keep their
+    // preset status, server networks their server status, delisted (TRASH)
+    // networks are already dropped and user custom RPC networks are excluded.
+    const { networks: eligibleNetworks } =
+      await this.backgroundApi.serviceNetwork.getAllNetworks({
+        excludeCustomNetwork: true,
+        excludeAllNetworkItem: true,
+      });
+    const eligibleNetworkMap = keyBy(eligibleNetworks, 'id');
     homeDefaults.forEach((homeDefault) => {
       homeDefaultTokenMap[
         buildHomeDefaultTokenMapKey({
@@ -912,7 +919,7 @@ class ServiceSetting extends ServiceBase {
     Object.entries(tokens).forEach(
       ([commonSymbol, { data, logoURI, name }]) => {
         const filteredData = uniqBy(
-          data.filter((token) => !!listedNetworkMap[token.networkId]),
+          data.filter((token) => !!eligibleNetworkMap[token.networkId]),
           (token) => token.networkId,
         );
 
--- a/packages/kit-bg/src/services/ServiceSetting.ts
+++ b/packages/kit-bg/src/services/ServiceSetting.ts
@@ -1010,6 +1010,7 @@ class ServiceSetting extends ServiceBase {
         aggregateTokenSymbolMap,
         configSyncMeta: {
           appVersion: platformEnv.version ?? '',
+          bundleVersion: platformEnv.bundleVersion ?? '',
           syncedAt: Date.now(),
         },
       }),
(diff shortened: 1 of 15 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-0ed687c0-10c1-59fe-805c-c4e153af2906)

originalix:

Review of commit `594d82ee9a3d` did not finish successfully.

packages/kit-bg/src/services/ServiceSetting.ts
```diff
@@ -900,7 +897,17 @@ class ServiceSetting extends ServiceBase {
...
+    // that are not part of presetNetworks, so gate on the merged network
+    // registry instead of the preset-only listed map: presets keep their
+    // preset status, server networks their server status, delisted (TRASH)
+    // networks are already dropped and user custom RPC networks are excluded.
+    const { networks: eligibleNetworks } =
+      await this.backgroundApi.serviceNetwork.getAllNetworks({
+        excludeCustomNetwork: true,
+        excludeAllNetworkItem: true,
+      });
+    const eligibleNetworkMap = keyBy(eligibleNetworks, 'id');
```
sidmorizon, line 920: **Stale network cleanup gets overwritten**

Severity: `non-severe`

With an hour-old server cache, `ensureServerNetworksFetched` returns while `getAllNetworks` starts an asynchronous refresh. In bg, the sync can persist the old JS-cache result after refresh clears shared native SimpleDB metadata. Removed aggregate members then remain visible for another day.

originalix:

Review of commit `46093153f089` did not finish successfully.

packages/kit-bg/src/services/ServiceSetting.ts
weatherstar: Fixed in 9f2e4dc536.

- `syncWalletConfig` now awaits `serviceCustomRpc.ensureServerNetworksFetched()` (alongside the config fetch) before building `eligibleNetworkMap`, so on a fresh install the first sync sees the server-delivered members instead of the empty cache.
- `fetchNetworkFromServer` is single-flight, so this wait shares the refresh already started by `getServerNetworks` instead of issuing a second `/network/list` request.
- When the server-network set actually changes, `fetchNetworkFromServer` clears `configSyncMeta`, so the next `syncWalletConfigIfNeeded` rebuilds the aggregate maps instead of waiting for the 1-day TTL. This also covers the case where the first fetch failed and the hourly refresh lands later.

Existing installs receiving the hot bundle already have the cache filled and are unaffected. Tests: `ServiceSetting.syncWalletConfig.test.ts` (wait ordering) and new `ServiceCustomRpc.serverNetworks.test.ts` (ensure/skip/swallow, single-flight, invalidation).

originalix:

Review of commit `9f2e4dc53635` did not finish successfully.

originalix (approved):

PR #13237, opened 2026-09-08
fix: market list and detail UI/UX walkthrough

erikzou:
UI/UX walkthrough fixes for the Market list and detail pages. Every change is scoped to Market (plus the one Swap header branch that only the embedded market swap reaches); no Market-external surface changes behaviour.

## Perps banner & top coins

- Restore the Perps banner on the top coins detail page by deriving `hlTicker` client-side from the Hyperliquid main-DEX universe, since the asset payload carries no such field. A cache miss no longer clears what `changeActiveToken` supplied.
- Show the full asset name as the first column's subtitle, matching the stock list.
- Use the 5m series for the top coins simple chart, and fill the gaps the backend leaves between traded buckets so the line stays continuous. Market only — the trade panel is untouched.

## Stock detail

- Let the price toggle hug its labels instead of carrying widths read off the English boxes.
- Move the chart mode switch below the chart, into a footer slot outside the resizable box so the drag handle keeps its own band.
- Make the market status badge's outer label say what its popover says, and add the next-open countdown.
- Merge the price and its info glyph into one button.
- Read an unheld variant as a zero balance rather than `--`.
- Put the expand chevron in the event's title row, per Figma.
- Give the analyst rating bars width floors so a narrow panel cannot collapse the track, and wrap the block instead of crushing it.

## Tooltips

Binds the copy from the tooltip work order: the stock detail's Key Data cells, header price and price-mode buttons, the list page's Price header, and the list page's tab titles.

Two mechanics worth flagging for review:

- A Tooltip renders its trigger as a pressable and consumes the click. A sortable header therefore builds its tooltip in `HeaderColumn` and hands the sort to the Tooltip's own `onPress`.
- The list tab bar is `Tabs.TabBar`, not the `CategorySelector` the work order names, so the tab tooltips go through `TabBar`'s `renderItem` hook.

## Consistency pass

- Every metric column now labels itself through the trending table's ranged phrases (`market.volume_in_range` / `market.change_in_range`), so the same column no longer reads "24h Volume" on one tab and "24h volume" on the next.
- Every price change display under Market goes through `priceChangeCapped`, which clamps at ±99,999% with a `>` marker. The activity range row had its own 999.99% clamp with no marker; it now shares the ceiling.
- The Pro chart's "Original" mode label picks up `market.chart_original`, which was already translated but unused.

## Layout

- The shared content frame caps at 1440px with a 20px gutter, down from 1480/62.
- The market swap panel's header actions match the stock trade panel beside them. Note the icon size has to be a size token: `Icon` resolves `size` through the token table, so the raw number the container derived was silently falling back to the 24px default.

## Verification

`yarn agent:check --profile commit` clean. Market suites pass (92 suites / 653 tests). The swap header icon sizing was measured in the running app rather than inferred.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-b1312c4b-6a5d-551c-86bf-536b829150ae)

packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
```diff
@@ -202,6 +213,42 @@ export function DesktopLayout({
...
+    if (!tooltip) {
+      return tabItem;
+    }
+    return (
+      <Tooltip
+        placement="top"
+        renderTrigger={tabItem}
+        renderContent={<SizableText size="$bodySm">{tooltip}</SizableText>}
+      />
+    );
```
sidmorizon, line 249: 🟠 P1: Tooltip tabs stop switching

Severity: `severe`

Clicking a tab with help text activates the wrapper instead of `TabBarItem`. The [`Tooltip` trigger](packages/components/src/actions/Tooltip/index.tsx:198-205) receives no `onPress`, so Trending, Stocks, Top Coins, and Robinhood labels cannot switch tabs.

packages/kit/src/views/Market/MarketDetailV2/components/StockTokenInfo/StockTokenInfoPopover.tsx
```diff
@@ -237,14 +258,37 @@ export function StockTokenInfoPopover() {
...
+              testID="stock-token-info-trigger"
+              iconAfter="InfoCircleOutline"
+              size="small"
+              variant="tertiary"
+              childrenAsText={false}
+            >
+              {label}
+            </Button>
+          }
         />
```
sidmorizon, line 293: **Stock token details stay closed**

Severity: `non-severe`

Clicking the combined price button activates the nested `Tooltip` instead of the details action. The [`Popover` trigger](packages/components/src/actions/Popover/index.tsx) opens only from its own press handler, so users cannot view stock-token details.

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/StockTokenVariantSelector.tsx
```diff
@@ -91,7 +91,14 @@ function findVariantBalance({
...
     );
   });
-  const balance = new BigNumber(position?.amount ?? '');
+  // No position for this variant means the account holds none of it, which is
+  // a zero balance, not missing information. An amount that fails to parse
+  // still falls back — that is corrupt data, and reading it as zero would
+  // state something the payload never said.
+  if (!position) {
+    return '0';
+  }
```
sidmorizon: **Unknown balances appear as zero**

Severity: `non-severe`

Before portfolio loading completes, `findVariantBalance` converts every missing position to zero. [`useStockPortfolioData`](packages/kit/src/views/Market/MarketDetailV2/hooks/useStockPortfolioData.ts:235-266) defaults unresolved and uncached failed requests to an empty list, so users see false zero holdings.

packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
erikzou: Checked this against the running app rather than the code alone, because a Tooltip trigger swallowing a press is a real trap in this codebase — it is exactly why `HeaderColumn` has to hand the sort to the Tooltip's own `onPress` (68d7386).

It does not apply here. In `HeaderColumn` the press handler sits on an **ancestor** of the trigger, so the trigger consumes the click before it arrives. Here `TabBarItem` is a **descendant** of the trigger and keeps its own `onPress`, so the press lands on it directly.

Verified with real clicks on all four tabs that carry copy, checking the active weight and the rendered list after each:

- 熱門 → active (`font-weight: 600`), list swapped
- 股票 → active, list swapped
- Robinhood → active, list swapped
- 主流幣 → active, list swapped

Same pairing as the price/info button below, which also still opens its popover.

packages/kit/src/views/Market/MarketDetailV2/components/StockTokenInfo/StockTokenInfoPopover.tsx
erikzou: Verified on the AAPL detail page: clicking the combined button opens the details popover (Token issuer / underlying / shares per token / trading hours / network / contract address all render).

`Popover` injects its press handler into the trigger it is given, and `Tooltip` accepts `onPress` and forwards it to `TMTooltip.Trigger`, so the press reaches the popover. Same pairing `ThemeButton` uses.

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/StockTokenVariantSelector.tsx
erikzou: Confirmed and fixed in 90ee20d.

You are right that the payload cannot carry this on its own: an empty list is also what arrives with no account, while the first fetch is in flight, and when a request fails with nothing cached.

`fetchStockPortfolioData` now reports which variants it could not establish — no resolvable account for their network, or a failed request with no cached row — and `useStockPortfolioData` treats "no account" and "not yet settled" as all variants unresolved. The selector renders those as unavailable. A variant the lookup did reach and did not find still reads as `0`, which is what the change was for.

The selector also defaults to unavailable when it is handed no resolution data at all, so a caller that forgets to pass it cannot silently claim a zero.

Covered by new cases in `useStockPortfolioData.test.ts` and `StockTokenVariantSelector.test.tsx`.

packages/shared/src/locale/enum/translations.ts
```diff
@@ -2471,6 +2471,7 @@ export enum ETranslations {
   market_chart_indicator_zero__label = 'market.chart_indicator_zero__label',
   market_chart_indicator_zero_line__label = 'market.chart_indicator_zero_line__label',
   market_chart_mode_simple = 'market.chart_mode_simple',
+  market_chart_original = 'market.chart_original',
```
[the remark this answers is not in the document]
erikzou: These files are not hand-edited. They are the output of `yarn i18n:pull`, the repo's own localization flow, and they are isolated in three commits that do nothing else: 9114423, 9bd6509, 010b549.

The unrelated copy in the diff is inherent to that flow — a pull brings down whatever Lokalise currently holds for every key, not just the ones this branch touches.

packages/shared/src/utils/tradingHoursUtils.ts
```diff
@@ -680,3 +680,61 @@ export function resolveUSTradingHoursActiveRow({
...
+      totalMinutes = Math.ceil((openAt - now) / MINUTE_MS);
+    }
+  }
+
+  if (
+    totalMinutes === undefined &&
+    typeof nextOpenMinutes === 'number' &&
+    Number.isFinite(nextOpenMinutes)
+  ) {
+    totalMinutes = Math.ceil(nextOpenMinutes);
```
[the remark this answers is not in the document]
erikzou: Correct, and fixed in 9e92b80.

The minute count is a snapshot, so it needs an anchor to decay from. `getUSMarketNextOpenCountdown` now takes `nextOpenMinutesObservedAt` and subtracts the elapsed time; `useNextOpenCountdownText` records when the payload landed. The countdown drops away once the snapshot runs out instead of holding a stale value through a stalled poll.

The `nextOpenTime` path is unchanged and still wins over the fallback. Two cases added to `tradingHoursUtils.test.ts`.

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/StockTokenVariantSelector.tsx
```diff
@@ -91,7 +91,14 @@ function findVariantBalance({
       equalsIgnoreCase(item.tokenAddress, variant.contractAddress)
     );
   });
-  const balance = new BigNumber(position?.amount ?? '');
+  // No position for this variant means the account holds none of it, which is
+  // a zero balance, not missing information. An amount that fails to parse
+  // still falls back — that is corrupt data, and reading it as zero would
+  // state something the payload never said.
+  if (!position) {
+    return '0';
```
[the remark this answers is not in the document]
erikzou: Confirmed and fixed in 90ee20d — same finding as the review comment above, addressed together.

The fetch now reports which variants it could not establish (no resolvable account for their network, or a failed request with nothing cached), and no-account / first-load-pending mark every variant unresolved. Those render as unavailable; a variant that was actually looked up and not found still reads as `0`.

packages/kit/src/views/Market/MarketDetailV2/hooks/useStockPortfolioData.ts
```diff
@@ -265,8 +286,19 @@ export function useStockPortfolioData() {
     },
   );
 
+  const portfolioData = portfolioResult?.items ?? [];
+  // Until the first run settles nothing has been established, so every variant
+  // counts as unresolved rather than as a zero holding.
+  const unresolvedVariantKeys = useMemo(
+    () =>
+      portfolioResult?.unresolvedVariantKeys ??
+      tokenVariants.map(getStockPortfolioVariantKey),
```
sidmorizon: **Stock switches show false zero balances**

Severity: `non-severe`

When a mounted detail page changes stocks, `portfolioResult` still describes the previous stock until the new request completes. New variants absent from that old unresolved set display zero instead of loading.

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -212,4 +217,17 @@ export function MarketDesktopChartContainer({
       ) : null}
     </Stack>
   );
+
+  // Fullscreen drops the handle and the footer with it, so the box stands on
+  // its own there and keeps carrying the fullscreen frame styles itself.
+  if (isFullscreen || !footer) {
```
originalix: P2: [全屏切换会重新挂载图表，丢失缩放和历史位置]

Desktop/Web 的 Pro 图表在普通模式下返回 `<YStack>{box}{footer}</YStack>`，进入全屏后却直接返回 `box`。父节点结构变化会让整棵图表子树卸载并重新挂载；用户缩放或拖到历史行情后进入/退出全屏，会丢失当前视口，并重新加载图表数据。`TradingViewNativeChart` 的 pan offset 和 zoom scale 就保存在组件内部的 state/ref 中。

已用当前容器做针对性测试：切换全屏后子组件状态被重置，同一用例在基线版本通过。建议始终保留外层 `YStack` 和 `box` 的层级，仅通过 `isFullscreen ? null : footer` 隐藏工具栏；这个调整在测试副本中能保留子组件状态。

packages/kit/src/views/Market/MarketDetailV2/hooks/useStockPortfolioData.ts
originalix: P2: [切换账户后仍沿用旧账户的“余额已查询”标记]

补充同一处的账户切换场景：账户 A 查询成功且无持仓后，保持股票详情页挂载并切换到账户 B，`usePromiseResult` 在 B 请求期间仍保留 A 的 `portfolioResult`，这里便继续返回旧的 `unresolvedVariantKeys: []`。选择器将 B 尚未查询的余额显示为 `0`，因此只处理首次加载和请求失败还不够。

已使用真实 `useStockPortfolioData` 和 `usePromiseResult`、延迟返回 B 的请求复现：`hasAccount=true`、`isRefreshing=true` 时，`unresolvedVariantKeys` 仍为空。

建议让返回结果携带账户、派生类型及股票/变体的查询身份，在渲染时校验是否属于当前身份；不匹配时将余额标为未知。同一身份的周期刷新可以继续保留缓存，避免每 15 秒清空一次显示。

erikzou: Good catch — real gap in the previous fix, addressed in ff8c9e0.

Reporting failures could only speak about variants the run had seen. A stock switch leaves the previous stock's result standing until the new request lands, and the incoming variants are absent from that old failure list, so they read as resolved and printed a zero again.

The fetch now reports `resolvedVariantKeys` — what it did establish — instead. Anything outside that set is unknown, so another stock's variants, and any variant added since the standing payload was built, both fall back. That also subsumes the "nothing has settled yet" case, so the separate fallback for it is gone.

Pinned by a new case in `StockTokenVariantSelector.test.tsx` that renders the previous stock's resolution against the new stock's variants.

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
erikzou: Confirmed and fixed in e873e1a — thanks, this was mine and the diagnosis is exactly right.

Your suggested shape is what it does now: the `YStack` wrapper is unconditional and fullscreen drops the footer instead of the wrapper.

One thing worth adding: the `!footer` branch went down the same bare path, and `TokenDetailChart` sets its toolbar to `undefined` in fullscreen — so that caller was hitting the remount through both conditions at once. Both collapse into the single wrapper now.

Pinned by a case in `MarketDesktopChartContainer.test.tsx` that counts child mounts across a fullscreen round trip with the footer appearing and disappearing. I checked it fails against the previous code before keeping it.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1650,6 +1651,33 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+      return undefined;
+    }
+    const findMainDexAsset = (universesByDex: IPerpsUniverse[][]) =>
+      universesByDex[0]?.find((item) => !item.isDelisted && item.name === coin);
+
+    let { universesByDex } = await this.getTradingUniverse();
+    if (!universesByDex[0]?.length) {
+      // Market is reachable without ever mounting Perps, so the positional
+      // universe cache can still be empty here on a cold start.
+      await this.refreshTradingMeta();
```
sidmorizon: **Stale perps metadata mislabels markets**

Severity: `non-severe`

A nonempty persisted universe skips `refreshTradingMeta`, even when outdated. Market detail loads without [PerpsGlobalEffects](packages/kit/src/views/Perp/pages/Perp.tsx:44), so new markets stay missing and delisted markets remain advertised.

erikzou: Confirmed and fixed in 7575534.

The cache is persisted (`simpleDb.perp`), so "non-empty" said nothing about age — a universe written on the last Perps visit was served however stale, and as you note nothing on the Market path refreshes it.

The guard is now empty **or** older than an hour. `getTradingUniverse` already returned `tradingUniversesUpdatedAt`, so the check costs nothing extra, and `refreshTradingMeta` dedupes concurrent callers.

Two things I added beyond the report:

- A failed refresh falls back to the cached universe instead of propagating. Losing a mostly-correct answer to a network blip would be worse than the staleness this guards against, and the caller reads a rejection as "no perps entry" either way.
- The staleness decision is extracted to `marketPerpsUniverse.ts` with unit tests, since importing the service into jest pulls the whole background graph.

packages/kit-bg/src/services/ServiceHyperLiquid/marketPerpsUniverse.ts
```diff
@@ -0,0 +1,37 @@
...
+  updatedAt?: number;
+  now?: number;
+}) {
+  if (!universesByDex[0]?.length) {
+    return true;
+  }
+  if (!updatedAt) {
+    return true;
+  }
+  return now - updatedAt > MARKET_PERPS_UNIVERSE_MAX_AGE_MS;
```
sidmorizon, line 44: **Clock rollback freezes market metadata**

Severity: `non-severe`

After a device clock rollback, `now - updatedAt` stays negative and treats the cache as fresh. Market listings remain stale until the clock passes the persisted timestamp plus one hour.

Suggested fix:

```ts
  return (
    updatedAt > now ||
    now - updatedAt > MARKET_PERPS_UNIVERSE_MAX_AGE_MS
  );
```

originalix:

Review of commit `6dde2009aa29` finished, but some comments could not be published automatically.

packages/kit-bg/src/services/ServiceHyperLiquid/marketPerpsUniverse.ts
erikzou: Right, and fixed in 1d83959 — I took your suggested shape.

A timestamp ahead of `now` means the age cannot be measured at all, so it refreshes like an absent one rather than being read as a negative difference. Covered by a case in `marketPerpsUniverse.test.ts`.

erikzou:
Heads-up on a deliberate divergence from #13271, so it does not read as a bad merge.

@zy2000-star Merging `x` into this branch conflicted on the stock price-mode toggle in `StockDesktopLayout.tsx`. #13271 restored `minWidth` on the container (191) and on both buttons (94 / 95); this branch had removed those widths earlier, per Figma 25476:89067 — they were measured off the English labels, so the shorter CJK ones sat in half-empty pills.

I kept this branch's version, and I believe it also covers the case your comment is protecting:

- The buttons carry no minimum and `flexShrink={0}`, so a long Spanish or Italian label grows the button to its text rather than truncating it.
- Your own change in the same commit — `flexWrap="wrap-reverse"` on the header, which merged in cleanly — moves the pair onto its own line when the row runs out of room.

So nothing should truncate, and the CJK padding is gone. The reasoning is in a comment at the call site referencing both the Figma node and #13271, so the next person to touch it does not restore the minimums assuming a merge was dropped.

If the widths were protecting something I have missed, say so and I will put them back — the design side of it is Figma 25476:89067.

sidmorizon:
Claude review session: https://claude.ai/code/session_018g6DW8hft3L8HC3gkYHQVJ

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquid.ts
```diff
@@ -1674,6 +1676,46 @@ export default class ServiceHyperliquid extends ServiceBase {
...
+    symbol,
+  }: {
+    symbol: string;
+  }): Promise<IMarketPerpsInfo | undefined> {
+    const coin = symbol?.trim().toUpperCase();
+    if (!coin) {
+      return undefined;
+    }
+    const findMainDexAsset = (universesByDex: IPerpsUniverse[][]) =>
+      universesByDex[0]?.find((item) => !item.isDelisted && item.name === coin);
```
sidmorizon, line 1696: **Multiplier-prefixed perpetuals never match**

`findMainDexAsset` compares plain asset symbols literally against Hyperliquid tickers. PEPE cannot match `kPEPE`, so its Top Coins page omits a valid perpetuals banner.

originalix (approved):

PR #13239, opened 2026-09-08
fix: align EVM addresses in signature confirm display with EIP-55 checksum | OK-62341

weatherstar:
## Summary
- Signature confirm (transaction and message parsing) now shows EVM addresses in EIP-55 checksum form, matching what the hardware device renders.
- Applies to `Address` components (`address`) and `Approve` components (`spender`) on the server display, the local-fallback display, and the parsed-message display. Non-EVM networks and non-address values (ENS names, placeholders) are left untouched.

## Intent & Context
OK-62341: on the tx confirm page the "接收地址" / "交互合约" rows showed plain lowercase hex (`0x8b844f...`) while the OneKey Pro showed the checksum form (`0x8B844f...`). Hardware users compared the two and doubted whether they were the same address.

## Root Cause
`/wallet/v1/account/parse-transaction` and `/wallet/v1/account/parse-signature` return normalized lowercase addresses. `ServiceSignatureConfirm.buildDecodedTx` copies `display.components` into `decodedTx.txDisplay` and `SignatureConfirmComponents/Address.tsx` prints `component.address` verbatim, so the UI showed the server's canonical form instead of the display form. The local fallback (`convertDecodedTxActionsToSignatureConfirmTxDisplayComponents`) has the same gap: it uses `action.to` as supplied by the dApp.

## Design Decisions
- **Client-side, at display-build time, not on the server.** Lowercase is the right canonical form for data and every client comparison already lowercases (`getAddressKey`, address book `_findItemByConditions`, localDb EVM index keys). Only the display should change, and the server only covers one of the three display paths (server display / local fallback / message display).
- **One hook point in the background.** `checksumDisplayComponentAddresses` runs once at the end of `buildDecodedTx` (after the Approve backfill and private-send recipient rewrite, so their substitutions are covered too) and once before `parseMessage` returns. UI components need no change, and copy / explorer links also get the checksum form.
- **Reuses the existing validator.** `validateEvmAddress` (`@ethersproject/address` `getAddress`) already produces `displayAddress`; this is the same path `serviceAccountProfile.queryAddress` and `BulkSendApprovalCard` use.
- **Component `networkId` wins over the tx `networkId`** so bridge destinations on non-EVM chains are not touched.
- **Identity-preserving no-op.** The helper returns the same array instance when nothing changed, so a display with no EVM addresses keeps the original server object.

## Changes Detail
- `packages/kit-bg/src/services/utils/displayComponentAddressUtils.ts` (new): `checksumDisplayComponentAddresses({ networkId, components })`.
- `packages/kit-bg/src/services/ServiceSignatureConfirm.ts`: apply the helper at the end of `buildDecodedTx` and before `parseMessage` returns.
- `packages/kit-bg/src/services/utils/displayComponentAddressUtils.test.ts` (new): 7 unit tests (EIP-55 vectors, Approve spender, non-EVM passthrough, component networkId override, invalid values, identity no-op, other component types by reference).
- `packages/kit-bg/src/services/ServiceSignatureConfirm.test.ts`: 3 integration tests (server display, local fallback display, parsed message display).
- `apps/mobile/bundle-registry/module-id-registry.json`: register the new module.

## Risk Assessment
- Low. Display-only change in the background; addresses are only rewritten when `validateEvmAddress` accepts them, otherwise the original string is kept. All local lookups that consume `component.address` (account name, address book, contract tags, explorer URL) are case-insensitive or format-agnostic.
- Not covered by design: `Default` component values and `txABI.args` are untyped strings, so the client cannot know which of them are addresses. If product wants those checksummed too, the server needs to emit `Address` components or a type hint (follow-up).

## Test plan
- [x] `yarn jest packages/kit-bg/src/services/utils/displayComponentAddressUtils.test.ts packages/kit-bg/src/services/ServiceSignatureConfirm.test.ts` (25 passed)
- [x] `yarn agent:check --profile commit` and `--profile pr` green
- [ ] Device QA: EVM send / swap / approve confirm page and EIP-712 sign page show checksum addresses; compare with the device screen; non-EVM (SOL/BTC) unchanged

### DIFF
--- a/apps/mobile/bundle-registry/module-id-registry.json
+++ b/apps/mobile/bundle-registry/module-id-registry.json
@@ -19047,6 +19047,7 @@
     "packages/kit-bg/src/services/utils/borrowActionServiceUtils.ts": 18125,
     "packages/kit-bg/src/services/utils/buildSpeedSwapTxParams.ts": 8003,
     "packages/kit-bg/src/services/utils/currencySymbolSyncUtils.ts": 7359,
+    "packages/kit-bg/src/services/utils/displayComponentAddressUtils.ts": 2904,
     "packages/kit-bg/src/services/utils/marketStockUtils.ts": 10760,
     "packages/kit-bg/src/services/utils/marketTokenDetailUtils.ts": 19353,
     "packages/kit-bg/src/services/utils/permit2SignatureConfirmUtils.ts": 14635,
--- a/packages/kit-bg/src/services/ServiceSignatureConfirm.test.ts
+++ b/packages/kit-bg/src/services/ServiceSignatureConfirm.test.ts
@@ -49,6 +49,7 @@ import {
   EParseTxComponentType,
   EParseTxType,
   type IDisplayComponent,
+  type IDisplayComponentApprove,
   type IDisplayComponentSimulation,
   type IParseTransactionResp,
 } from '@onekeyhq/shared/types/signatureConfirm';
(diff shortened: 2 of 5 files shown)


weatherstar:
Closed: opened by mistake, the author will handle PRs in a batch later.

PR #13246, opened 2026-09-08
fix: EIP-55 checksum on signature confirm, recent-selection cache on network switch, network-less account selection after wallet creation, MEV logo first open(OK-62341, OK-62330, OK-62137, OK-62097)

weatherstar:
[OK-62341](https://onekeyhq.atlassian.net/browse/OK-62341) [OK-62330](https://onekeyhq.atlassian.net/browse/OK-62330) [OK-62137](https://onekeyhq.atlassian.net/browse/OK-62137) [OK-62097](https://onekeyhq.atlassian.net/browse/OK-62097)

----------

## Summary
- Signature confirm now renders EVM addresses in EIP-55 checksum form so the app matches what the hardware device shows (OK-62341).
- The account selector's 5-minute "recent selection" cold-start cache is refreshed on every later in-memory change, so a network switch survives an app restart (OK-62330).
- An account selection can no longer exist without a network: it is filled in memory, refused by storage, repaired on load, and the background falls back to All Networks instead of reporting "no address" (OK-62137).
- The tx-confirm MEV popover renders the provider logo in a fixed contain box instead of resizing from an in-popover `onLoad`, so the first open is no longer squeezed (OK-62097).

## Intent & Context
All four are App-6.6.0 QA bugs on the `x` line, batched on `fix/wallet-issues` per the current workflow (commit per ticket, one PR at the end).

- **OK-62341**: server-parsed display components carry lowercase EVM addresses while OneKey devices render EIP-55 checksum addresses, so the confirm page and the device disagreed visually.
- **OK-62330** (desktop): pick a wallet in the account selector, switch the home network ETH → SOL, change language (which restarts the app); after the restart the network was back on ETH and simpleDb had been overwritten with ETH.
- **OK-62137** (extension, hardware wallet onboarding): after connecting a hardware wallet the expand tab home showed 「无地址 / 创建地址以启用网络」 with a blank network selector; a page refresh showed the account normally.
- **OK-62097** (desktop): first open of the "MEV 保护" popover on the tx-confirm page showed the Blink logo tiny and left-aligned; every later open was normal.

## Root Cause
- **OK-62341**: no checksum step existed in `ServiceSignatureConfirm`; addresses were shown exactly as the server / local fallback produced them.
- **OK-62330**: `onekey_account_selector_recent_selection` (written only by `confirmAccountSelect`, valid for `RECENT_ACCOUNT_SWITCH_COLD_START_MS`) outranks simpleDb in `initFromStorage`. A later `updateSelectedAccountNetwork` updated the atoms and simpleDb but not that cache, so the next cold start restored the pre-switch selection and `saveToStorage` wrote it back over the newer record.
- **OK-62137**: the exported extension log (last 1000 background lines) showed the persisted home selection as `{ walletId, indexedAccountId, focusedWallet }` with no `networkId`; `buildActiveAccountInfoFromSelectedAccount` treats a missing network as the single-network "no account" case (`canCreateAddress = true`, `network` undefined), which is exactly the empty state and blank selector in the screenshot. Structurally nothing guaranteed a network on an identity selection: `defaultSelectedAccount()` has `networkId: undefined`, the only initializer was the one-shot `useAutoSelectNetwork` effect (a `void` call with no retry), `autoSelectToCreatedWallet` spreads whatever the store had, `saveToStorage` only checked wallet/account identity, and `buildMergedSelectedAccount` let an explicit `networkId: undefined` in the target override a defined source network (which is why the swap scene healed itself but home did not). The exact trigger in the onboarding runtime (a failed/hung background call before `useAutoSelectNetwork` completed, or a store recreated right before creation) could not be proven from the truncated log.
- **OK-62097**: `#13137` (2026-09-03) replaced the header-mount `Image.loadImage` size prefetch + skeleton with an `Image` whose width starts at `DEFAULT_IMAGE_HEIGHT` (40) and is recomputed from the popover image's `onLoad`. The 2504×960 Blink logo is `contain`-fitted into a 40×40 box until that state lands, and the recording shows the first open never re-rendering with the new width. The CDN was verified innocent (`x-tos-process=image/resize,w_96` returns 96×37 with the ratio kept).

## Design Decisions
- **OK-62341**: a display-only rewrite (`checksumDisplayComponentAddresses`) applied as the last step for both tx and message display so server, local-fallback and private-send paths are covered; non-EVM networks and non-address values (ENS names, placeholders) are returned untouched, and the input array instance is returned when nothing changed.
- **OK-62330**: `refreshRecentAccountSelectorSelectionCacheIfActive` at the end of `updateSelectedAccount` only rewrites an entry that already exists and is still valid, so no new cache entries are created outside the window and there is no extra churn.
- **OK-62137**: fix the invariant at every layer instead of adding another UI-side heal: `fillMissingNetworkIdForAccountSelection` runs inside `updateSelectedAccount` and on the `confirmAccountSelect` fast path (scene `defaultNetworkId` → scene `networkIds[0]` → All Networks; discover scenes never get All Networks, mirroring `useAutoSelectNetwork`); `saveToStorage` refuses an identity without a network; `buildMergedSelectedAccount` spreads `omitBy(data, isUndefined)`; background `resolveSelectedAccountNetworkId` repairs persisted records in `fixDeriveTypesForInitAccountSelectorMap` (existing corrupted installs self-heal on next load) and falls back to All Networks in `buildActiveAccountInfoFromSelectedAccount`; `useAutoSelectNetwork` retries up to 3× at 1 s.
- **OK-62097**: size the box, not the image: `width="100%"`, `height={40}`, `resizeMode="contain"`, `resizeWidth={160}` keeps the CDN resize and centers the logo on web (RNW base style `backgroundPosition: center`) and native alike, matching the ticket's expectation of a centered, normal-size logo. Same pattern as `ExternalWalletList.tsx`.

## Changes Detail
- `packages/kit-bg/src/services/utils/displayComponentAddressUtils.ts` (+ test): EIP-55 rewrite of `Address` and `Approve.spender` display components for EVM networks.
- `packages/kit-bg/src/services/ServiceSignatureConfirm.ts` (+ test): apply the rewrite to tx display components and parsed message components.
- `packages/kit/src/states/jotai/contexts/accountSelector/actions.tsx` (+ tests): recent-selection cache refresh (OK-62330); `fillMissingNetworkIdForAccountSelection` wired into `updateSelectedAccount` and `confirmAccountSelect`; `saveToStorage` network guard (OK-62137).
- `packages/shared/src/utils/accountSelectorUtils.ts` (+ test): `buildMergedSelectedAccount` ignores explicit `undefined` keys in the target.
- `packages/kit-bg/src/services/ServiceAccountSelector.ts` (+ tests): `resolveSelectedAccountNetworkId` used for load-time repair and the active-account fallback.
- `packages/kit/src/components/AccountSelector/hooks/useAutoSelectNetwork.tsx` (+ test): bounded retry of the network auto-select.
- `packages/kit/src/views/SignatureConfirm/components/SignatureConfirmHeader/TxConfirmHeaderRight.tsx` (+ `TxConfirmHeaderRight.logo.test.tsx`): fixed contain box for the MEV provider logo; the `onLoad` sizing state is removed.
- `apps/mobile/bundle-registry/module-id-registry.json`: registers the new kit-bg util module.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**:
  - OK-62137 touches the central account-selector write path: every identity selection without a network now gets a network filled in (scene default or All Networks). Scenes that pass `availableNetworksMap` keep their own default; discover scenes are exempt from the All Networks fallback. The background fallback also applies to imported/watching accounts (All Networks with an others-wallet account is already a supported state).
  - OK-62341 is display-only; signing payloads are untouched.
  - OK-62097 loads the raw 2504×960 logo only if the CDN resize is unavailable; `resizeWidth` keeps the 384px rendition on 2× screens.

## Test plan
- [ ] Unit: `yarn jest` on `ServiceSignatureConfirm.test.ts`, `displayComponentAddressUtils.test.ts`, `actions.test.tsx` (60), `ServiceAccountSelector.test.ts` (9), `accountSelectorUtils.test.ts` (3), `useAutoSelectNetwork.test.tsx` (2), `TxConfirmHeaderRight.logo.test.tsx` (1) — all green locally; `yarn agent:check --profile commit` passed per commit.
- [ ] OK-62341: open an EVM transfer / approve / typed-data confirm; every address on the confirm page is in EIP-55 form and matches the device screen.
- [ ] OK-62330 (desktop): select an account in the account selector, switch the home network, change the app language; after the restart the switched network is kept.
- [ ] OK-62137 (extension): fresh install → connect a hardware wallet in the expand tab → home shows the account on All Networks with a populated network selector on the very first render; an install that already has the corrupted record shows the account after upgrade without a refresh.
- [ ] OK-62097 (desktop, light and dark theme): send on BSC or Ethereum, open the MEV popover on the first try; the Blink logo is centered and the same size as on later opens.

[OK-62341]: https://onekeyhq.atlassian.net/browse/OK-62341?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62330]: https://onekeyhq.atlassian.net/browse/OK-62330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62137]: https://onekeyhq.atlassian.net/browse/OK-62137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62097]: https://onekeyhq.atlassian.net/browse/OK-62097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5be46c63-9b2d-5a08-813a-492acb53b0a1)

packages/shared/src/utils/accountSelectorUtils.ts
```diff
@@ -78,9 +80,11 @@ function buildMergedSelectedAccount({
...
   mergedByData: IAccountSelectorSelectedAccount;
 }): IAccountSelectorSelectedAccount {
+  // Explicit `undefined` keys (e.g. a freshly created default selection)
+  // must not erase a defined network context coming from the source.
   const result: IAccountSelectorSelectedAccount = {
     ...mergedByData,
-    ...data,
+    ...omitBy(data, isUndefined),
     walletId: mergedByData.walletId,
     indexedAccountId: mergedByData.indexedAccountId,
```
sidmorizon, line 102: **Merge restores incompatible account derivation**

Severity: `non-severe`

When a target network clears `deriveType`, `omitBy` removes that reset. The merge inherits the source network’s derivation and can select the wrong account variant.

Suggested fix:

```ts
    ...data,
    networkId: data?.networkId ?? mergedByData.networkId,
```

packages/kit/src/components/AccountSelector/hooks/useAutoSelectNetwork.tsx
```diff
@@ -74,19 +84,47 @@ export function useAutoSelectNetwork({ num }: { num: number }) {
...
+            retryCountRef.current = 0;
+          })
+          .catch(() => {
+            if (
+              cancelled ||
+              retryCountRef.current >= AUTO_SELECT_NETWORK_MAX_RETRY
+            ) {
+              return;
+            }
+            retryCountRef.current += 1;
```
sidmorizon, line 101: **Later network selection loses retries**

Severity: `non-severe`

After exhausted retries, `retryCountRef` stays capped when another path selects a valid network. A later failed auto-selection gets no retries and remains unselected.

originalix:

Review of commit `03254a46780c` did not finish successfully.

packages/shared/src/utils/accountSelectorUtils.ts
weatherstar: Fixed in 8e55d51569. Confirmed the scenario: `SimpleDbEntityAccountSelector.cloneAndFixSelectedAccount` reads an All Networks selection back with `deriveType: undefined`, and `actions.tsx` merges that home selection as `data`, so with `omitBy` the source's derivation (e.g. BTC `BIP86`) leaked onto the target's All Networks context.

Went one step further than `networkId: data?.networkId ?? mergedByData.networkId`: the network context now travels as a `(networkId, deriveType)` pair. A target that already has a network keeps its own `deriveType` as-is (a cleared one stays cleared); a target without a network inherits both from the source. Protecting only `networkId` would leave a fresh selection with an inherited single network but `deriveType: undefined`, which `buildActiveAccountInfoFromSelectedAccount` treats as "cannot query the indexed account" and reproduces the "no address" symptom OK-62137 fixes. `lodash` import dropped; tests added for both directions.

packages/kit/src/components/AccountSelector/hooks/useAutoSelectNetwork.tsx
weatherstar: Fixed in 8e55d51569. The effect now resets `retryCountRef` whenever it observes a valid network (the `else` branch of the network check), so any path that lands on a valid network restores the retry budget, not only this hook's own successful call. Dropped the `.then` reset since the effect re-runs with the new `networkId` anyway. Added a test: exhaust retries → an external path selects a valid network → the network is lost again → the auto-select retries again.

packages/kit-bg/src/services/ServiceAccountSelector.ts
```diff
@@ -580,6 +606,10 @@ class ServiceAccountSelector extends ServiceBase {
         async (item: [string, IAccountSelectorSelectedAccount | undefined]) => {
           // TODO add whitelist
           const [num, v] = item;
+          if (v && !v.networkId) {
+            // Repair a persisted account selection that lost its network.
+            v.networkId = resolveSelectedAccountNetworkId(v);
```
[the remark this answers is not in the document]
weatherstar: Fixed in 8e55d51569. `resolveSelectedAccountNetworkId` now takes `sceneName` and returns `undefined` for discover scenes, so neither the load repair (`fixDeriveTypesForInitAccountSelectorMap`) nor `buildActiveAccountInfoFromSelectedAccount` (which now receives the scene from `reloadActiveAccountInfo`) writes All Networks for a dApp connection. The repair is left to the scene's available-network auto-select, matching the main-side guard in `fillMissingNetworkIdForAccountSelection`. The `ServiceDApp` caller always passes an explicit `networkId`, so it is unaffected. Tests added for both background paths.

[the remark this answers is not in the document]
weatherstar: Same finding as the sibling thread on this line; fixed in 8e55d51569 (discover scenes skip the All Networks repair in both background paths). See the reply there for details.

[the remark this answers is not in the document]
weatherstar: Same finding as the sibling thread on this line; fixed in 8e55d51569 (discover scenes skip the All Networks repair in both background paths). See the reply there for details.

originalix (approved):

sidmorizon (approved):

PR #13247, opened 2026-09-08
feat: render Earn tokens by displaySymbol so relabels need no symbol …

ezailWang:
…rename

The server can now relabel a token with an optional `displaySymbol` alongside `symbol`. `symbol` stays the key it always was — the request param, the /earn/:network/:symbol/:provider segment, and what supportedSymbols and the staking config match on — so none of those change and older clients keep rendering `symbol` unaffected.

earnUtils.getDisplaySymbol resolves the label; matchesSymbolKeyword searches both, so a relabelled token is still found by the symbol users know. Applied to identity surfaces only: the available-assets lists, search, protocol token rows, recommendations and the two detail page titles. Amount units keep the short symbol, since a relabel can be far longer than a symbol and would break "0.00 <symbol>" layouts. Those rows grew a single-line clamp so a long label truncates instead of pushing the APR off the edge.

Katana vbUSDC is the first token to use it, reading as "USDC for Katana (vbUSDC)".

### DIFF
--- a/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
+++ b/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
@@ -69,7 +69,7 @@ function TokenDetailsFooter(props: { networkId: string }) {
           if (tokenMetadata?.coingeckoId) {
             navigation.push(EModalAssetDetailRoutes.MarketDetail, {
               token: tokenMetadata.coingeckoId,
-              networkId,
+              preferredNetworkId: networkId,
             });
           }
         }}
--- a/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
+++ b/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
@@ -69,6 +69,7 @@ function TokenDetailsFooter(props: { networkId: string }) {
           if (tokenMetadata?.coingeckoId) {
             navigation.push(EModalAssetDetailRoutes.MarketDetail, {
               token: tokenMetadata.coingeckoId,
+              networkId,
             });
           }
         }}
--- a/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
+++ b/packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
@@ -24,8 +24,13 @@ import { AssetDetailsTestIDs } from '../../testIDs';
 
 import { useTokenDetailsContext } from './TokenDetailsContext';
 
-function TokenDetailsFooter(props: { networkId: string }) {
-  const { networkId } = props;
+function TokenDetailsFooter(props: {
+  networkId: string;
+  tokenAddress: string;
+  isNative?: boolean;
+  isAggregateToken?: boolean;
+}) {
+  const { networkId, tokenAddress, isNative, isAggregateToken } = props;
   const intl = useIntl();
   const { bottom } = useSafeAreaInsets();
   const { tokenMetadata } = useTokenDetailsContext();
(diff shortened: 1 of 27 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-77ec47bc-4bd4-55d7-8bec-3e5558dde2f0)

packages/kit/src/views/Earn/components/AvailableAssetsTabViewList.tsx
```diff
@@ -272,7 +273,9 @@ export function AvailableAssetsTabViewList({
               tokenImageUri={asset.logoURI}
               borderRadius="$full"
             />
-            <SizableText size="$bodyLgMedium">{asset.symbol}</SizableText>
+            <SizableText size="$bodyLgMedium" numberOfLines={1} flexShrink={1}>
+              {earnUtils.getDisplaySymbol(asset)}
+            </SizableText>
```
sidmorizon, line 278: **Displayed asset names sort incorrectly**

Severity: `non-severe`

When a relabeled asset column is sorted, `render` shows `displaySymbol` while [`comparator`](packages/kit/src/views/Earn/components/AvailableAssetsTabViewList.tsx:268) orders by `symbol`. Katana vbUSDC therefore appears under V despite its displayed name beginning with USDC.

packages/kit/src/views/Earn/hooks/useEarnAllProtocols.ts
```diff
@@ -112,20 +123,30 @@ function mergeItemIntoProviderMap(
...
-        .map((asset) => asset.symbol),
-    ),
+    new Set(normalAssets.map((asset) => asset.symbol)),
   );
+  // The protocol rows this path merges have no relabel of their own, so keep
+  // the one the v2 assets carry, keyed by the symbol they were fetched under.
+  const displaySymbolBySymbol = new Map<string, string>();
+  for (const asset of normalAssets) {
+    if (asset.displaySymbol && !displaySymbolBySymbol.has(asset.symbol)) {
+      displaySymbolBySymbol.set(asset.symbol, asset.displaySymbol);
```
sidmorizon, line 135: **Relabels leak across matching symbols**

Severity: `non-severe`

When normal assets share `symbol` but carry different relabels, `displaySymbolBySymbol` keeps only the first relabel. Every provider row for that symbol then displays the same name, regardless of network or vault.

packages/kit/src/views/Market/components/tradeHook.utils.ts
```diff
@@ -0,0 +1,56 @@
...
+  const values = detailPlatforms ? Object.values(detailPlatforms) : [];
+
+  if (preferredNetworkId) {
+    const preferred = values.find(
+      (platform) => platform.onekeyNetworkId === preferredNetworkId,
+    );
+    if (preferred) {
+      return preferred;
+    }
+    return { contract_address: '', onekeyNetworkId: preferredNetworkId };
```
originalix: P1: [Unmapped network hint fabricates a tradable token]

When Market data has no entry for the preferred network—the Katana case covered by this helper's test—this fallback creates a non-native platform with no `tokenAddress`. `useMarketTradeActions` then uses that value as `realContractAddress` when building `importFromToken` and `marketPresetToken`, so Swap is opened without the contract address of the wallet asset that launched Market.

Carry the selected asset address and native flag with the preferred network context and build this fallback from that complete identity; otherwise keep contract-dependent Market actions disabled for an unmapped platform.

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
```diff
@@ -67,8 +72,19 @@ function TokenDetailsFooter(props: { networkId: string }) {
...
+            // non-native token without an address would send Swap a token
+            // it cannot resolve; those fall back to market data as before.
+            const hasConcreteIdentity =
+              !isAggregateToken &&
+              !networkUtils.isAllNetwork({ networkId }) &&
+              (isNative || Boolean(tokenAddress));
             navigation.push(EModalAssetDetailRoutes.MarketDetail, {
               token: tokenMetadata.coingeckoId,
+              preferredToken: hasConcreteIdentity
+                ? { networkId, tokenAddress, isNative }
```
sidmorizon, line 86: **Aggregate trades ignore selected network**

For an aggregate token, `hasConcreteIdentity` discards the active tab's concrete network and address. [`activeTabIndex`](packages/kit/src/views/AssetDetails/pages/TokenDetails/index.tsx:512-525) already selects that tab's market metadata. Trade and Earn actions can target another network or fail on unmapped networks such as Katana.

sidmorizon:
Claude review session: https://claude.ai/code/session_01FfqhrBbh1eWnqZeHq1sPGq

originalix (approved):

PR #13254, opened 2026-09-08
fix: handle Discovery IME Enter confirmation OK-62394

yikZero:
[OK-62394](<https://onekeyhq.atlassian.net/browse/OK-62394>)

----------

Chinese IME confirmation in Discovery's search field and address bar now only commits text. Pressing Enter again performs one search. Normal Enter, arrow navigation, and Escape remain available outside composition.

Keep the composition lock through the confirming Enter and preserve focus in the Discovery inputs. Non-native SearchBar retains focus by default and honors an explicit caller `blurOnSubmit` setting. Reuse the existing patched RN-web ESM composition prop forwarding, which preserves React events and handles callbacks added after mount.

Register the new IME module for native bundles. Regenerate the translation enum with the existing repository script to remove an identical duplicate inherited from `x`, which was blocking lint and a CLI unit test; translation mappings are unchanged.

Validation:
- 27 tests passed across five suites, including the patched RN-web ESM event path, late-added callbacks, `nativeEvent`, focus retention, second Enter, SearchBar submit locking and explicit blur settings, and the previously failing CLI subprocess test.
- `yarn agent:check --profile commit` passed, including full TypeScript validation.
- Full repository Oxlint and formatting checks passed in CI mode (`CI=true yarn _lint`).
- Native union build, main/background startup budgets, bundle architecture, and registry validation against the generated module map passed locally.
- The requester manually tested the initial desktop IME fix; review follow-ups were verified with automated tests.

[OK-62394]: https://onekeyhq.atlassian.net/browse/OK-62394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1cdeaa9a-b72a-55c2-81ad-03d121f42a78)

packages/components/src/forms/Input/index.tsx
```diff
@@ -249,6 +250,39 @@ export const useOnWebPaste = platformEnv.isNative
...
+        }
+        return attachImeCompositionListeners(node, {
+          onStart: (event) => {
+            handlersRef.current.onCompositionStart?.(event as any);
+          },
+          onEnd: (event) => {
+            handlersRef.current.onCompositionEnd?.(event as any);
+          },
+        });
+      }, [inputRef]);
```
sidmorizon: **Late composition handlers never activate**

Severity: `non-severe`

When composition callbacks are added after mount, `useOnWebImeComposition` has already skipped listener registration. Its effect only reruns when `inputRef` changes.

packages/components/src/forms/Input/index.tsx
```diff
@@ -249,6 +250,39 @@ export const useOnWebPaste = platformEnv.isNative
...
+        const { onCompositionStart, onCompositionEnd } = handlersRef.current;
+        if (!onCompositionStart && !onCompositionEnd) {
+          return;
+        }
+        return attachImeCompositionListeners(node, {
+          onStart: (event) => {
+            handlersRef.current.onCompositionStart?.(event as any);
+          },
+          onEnd: (event) => {
+            handlersRef.current.onCompositionEnd?.(event as any);
```
sidmorizon: **Composition callbacks receive wrong event type**

Severity: `non-severe`

Web consumers receive a native `Event` despite `onCompositionStart` and `onCompositionEnd` promising React composition events. Synthetic-event fields such as `nativeEvent` are absent.

originalix (dismissed):

packages/components/src/forms/Input/index.tsx
yikZero: Fixed in 3cfac8bdc8. Removed the redundant DOM-listener hook and restored composition prop forwarding through Input. The repository already patches the production RN-web ESM TextInput to forward these React callbacks, so callbacks added after mount are handled on the next render. Added a regression test against that patched ESM entry for late-added callbacks and exactly-once delivery. All 25 scoped tests and the commit checks, including TypeScript, pass locally.

packages/components/src/forms/Input/index.tsx
yikZero: Fixed in 3cfac8bdc8 by removing the native DOM-event adapter and using the existing RN-web ESM composition prop patch. Callers now receive React CompositionEvent objects rather than raw Event values cast with `as any`. The regression fixture now imports the actual patched ESM TextInput and verifies nativeEvent is a CompositionEvent and each callback fires once. The earlier CJS fixture did not exercise the production patch; that gap is corrected.

packages/components/src/layouts/SearchBar/index.tsx
```diff
@@ -163,8 +175,9 @@ export function SearchBar({
         !rest.addOns?.length && {
           addOns: clearAddOns,
         })}
-      onCompositionStart={handleCompositionStart}
+      onCompositionStart={compositionLock.start}
       onCompositionEnd={handleCompositionEnd}
+      onSubmitEditing={handleSubmitEditing}
```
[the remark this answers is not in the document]
yikZero: Fixed in dce70b6c1e and retained in 3cfac8bdc8: non-native SearchBar sets blurOnSubmit={false} after the caller props, so an ignored IME confirmation cannot trigger RN-web's automatic blur. Native callers retain their existing submit/blur policy. The SearchBar test asserts the non-native blur setting and submit lock; the updated real RN-web ESM regression verifies focus is retained through confirmation and the next Enter searches once. All 25 scoped tests pass locally.

packages/components/src/layouts/SearchBar/index.tsx
```diff
@@ -163,8 +175,10 @@ export function SearchBar({
...
+      {...(!platformEnv.isNative ? { blurOnSubmit: false } : null)}
```
sidmorizon: **Web search bars ignore blur requests**

Severity: `non-severe`

When a caller passes `blurOnSubmit={true}`, `SearchBar` overwrites it with `false` on every browser target. Enter leaves the field focused instead of honoring the caller's dismissal contract.

yikZero: Fixed in 3e4dbbce14. Non-native SearchBar now uses `rest.blurOnSubmit ?? false`: the IME-safe default retains focus, while an explicit caller true/false is preserved. Native behavior is unchanged, and the two Discovery inputs continue to explicitly disable auto-blur. Added regression coverage for both explicit values; the true case failed before the fix. All 27 scoped tests and the commit checks, including TypeScript, pass locally.

sidmorizon:
Claude review session: https://claude.ai/code/session_01Hr1MxV1TxatikinbAQt2i1

sidmorizon (approved):

originalix (approved):

PR #13260, opened 2026-09-08
fix: OK-62355 OK-62418 reduce selector stalls and duplicate add-account

huhuanming:
[OK-62355](<https://onekeyhq.atlassian.net/browse/OK-62355>) [OK-62418](<https://onekeyhq.atlassian.net/browse/OK-62418>)

----------

## Issues

- [OK-62355](https://onekeyhq.atlassian.net/browse/OK-62355) — switching accounts stalls while selector values are calculated.
- [OK-62418](https://onekeyhq.atlassian.net/browse/OK-62418) — a repeated add-account interaction starts the same workflow more than once and overlaps shared batch progress.

## Runtime scope

The reported and reproduced target is iOS. Main and BG are isolated JS runtimes with separate Hermes heaps; the account-value work runs in BG, while Main waits for RPC results and updates the selector. Native storage remains process/device owned. This PR changes JS/TS only.

### OK-62355

Confirmed from logs and source call chain:

- The original supplied log contains a selector-value RPC lasting `2,383ms`.
- A historical pre-fix session retained in the physical-device native log, using the same QA wallet data, contains overlapping selector RPCs lasting `6,871ms` and `3,998ms`; an old account-value batch lasted `7,005ms`.
- The dataset contains `2,926` persisted DB account records. Before this fix, one 50-row selector batch ran `Promise.all(rows.map(...getAccountsInSameIndexedAccountId))`.
- That call reaches `getAllAccounts()` and then the synchronous persistence `runSync + cloneDeep(allItemsInCache)` path before filtering by `indexedAccountId`. `async`/`Promise.all` does not make this clone asynchronous or yield BG to another native/RPC task.
- Therefore one 50-row batch could clone/filter the 2,926-record account table about 50 times; two overlapping equivalent loaders could multiply that to about 100 full scans/clones (~292,600 account-object visits/clones).

Physical-iPhone regression with temporary NativeLogger timing (diagnostics removed before commit):

- Three 100-account “Large Data” wallets loaded in two 50-row batches.
- First batch: `297–342ms`; second batch: `69–75ms`; complete Main loader: `400–459ms`.
- Each batch now takes one account-table snapshot and lookup-map build. The first 50 results are published before the second batch completes.
- A single-account wallet still paid `254–303ms` for the one 2,926-record snapshot, which independently confirms that full-table materialization/cloning is the dominant local cost rather than the balance DB lookup.

Conclusion: the long BG occupancy is strongly attributed to repeated synchronous full-table cloning/filtering. Realm materialization, allocation, clone cost, and Hermes GC are not separately apportioned because there is no pre-fix Time Profiler capture for that exact historical session.

### OK-62418

Confirmed:

- The original log recorded the full UI add-account chain three times: `generateWalletsMissingMetaWithUserInteraction → addHDNextIndexedAccount → addDefaultNetworkAccounts`.
- Account counts changed `10 → 12 → 13`, proving real repeated creation rather than duplicate logging.
- Source inspection finds only user-event entry points for this path (footer add row, search action, and native-list add action); there is no timer/effect that automatically starts add-account.
- Those entry points can mount separate `useAddAccount` instances, so their local 300ms debounce state is not a cross-instance lock.
- On the physical iPhone, a rapid double press on the footer produced one Main add-account invocation and one BG all-network request with the shared in-flight guard.

Not confirmed: the missing original video/log ZIP prevents attributing the historical duplicate to one exact UI control versus a remounted/second entry point. The evidence supports a repeated UI workflow, not an automatic scheduler.

### OK-62355

- Equivalent selector-value requests share only the same in-flight Promise, keyed by their stable semantic payload.
- The entry is removed after success or failure; this is request coalescing, not a persistent result cache.
- V1 and V2 selector loaders use the same coalescing path.
- Each service batch reads the Account table once and builds maps by DB account ID and `indexedAccountId`, replacing per-row full-table clones/filters.

### OK-62418

- The complete UI add-account workflow uses one in-flight guard keyed by `walletId`, shared across hook instances.
- Repeated calls for the same wallet share the same task and cannot allocate the next indexed account twice.
- Different wallets retain independent behavior.
- Latest `x` already provides BG serialization through the existing `serviceHardwareUI.withHardwareProcessing` lease and its existing cancellation/request-id handling; this PR does not modify `ServiceBatchCreateAccount`.

## Minimal change boundary

- JS/TS only, plus generated mobile module-registry entries for the two small helpers.
- No native code, schema, dependency, route, public API, or product-level backup/export change.
- No new semaphore, queue, `operationId`, `flowId`, cancellation generation, or cancel semantics.
- No BatchCreate Preview submit guard: that path was removed from this PR because the original 62418 call chain does not provide strong evidence for it.
- Deduplication is limited to equivalent in-flight selector requests and the same wallet's in-flight add-account workflow.
- Temporary NativeLogger diagnostics and the local QA backup exporter were removed before commit.

## Validation

- Merged latest `origin/x` at `32e7ca870a` (merge commit `2b268862a3`).
- Focused Jest: 5 suites, 22 tests passed.
  - one Account-table snapshot expands a batch
  - equivalent selector requests share only in-flight work and rerun after settlement
  - V2 scheduling/cancellation remains isolated
  - same-wallet add-account calls share one task
  - different-wallet calls remain independent and rejection releases the guard
  - existing BG normal/all-network lease and cancellation tests remain green
- `yarn agent:check --profile commit` passed on the final scoped commit.
- Built, installed, and launched the merged branch on an iPhone 17 Pro; QA wallets/data were preserved and the account selector remained usable during the measured runs.
- Hosted CI passed on the final head: all 5/5 unit-test shards and their aggregate, lint, security scans, CodeQL, and startup-budget checks are green.

[OK-62355]: https://onekeyhq.atlassian.net/browse/OK-62355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62418]: https://onekeyhq.atlassian.net/browse/OK-62418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62355]: https://onekeyhq.atlassian.net/browse/OK-62355?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62418]: https://onekeyhq.atlassian.net/browse/OK-62418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-8159e543-d085-50b6-80d3-f1069ab7c1ac)

packages/kit-bg/src/services/ServiceBatchCreateAccount/ServiceBatchCreateAccount.ts
```diff
@@ -381,6 +384,24 @@ class ServiceBatchCreateAccount extends ServiceBase {
           saveToCache?: boolean;
           params: IBatchBuildAccountsNormalFlowParams;
         },
+  ) {
+    return this.batchCreateFlowMutex.runExclusive(() =>
```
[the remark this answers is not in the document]
huhuanming: 已在 2b95f1b18e 修复。两个 start 入口仍共享 semaphore 串行排队，但现在会在入队时捕获 cancellation generation；cancel 会递增 generation，旧的排队任务拿到锁后会先校验 generation，失配则在 beforeStartFlow 之前直接按取消失败，因此不会再把取消状态重置后继续创建。cancel 之后新提交的合法任务会捕获新 generation，可继续正常执行。Preview 的确认入口也增加了同步 isSubmittingRef 防重入，阻止连点产生第二个 BG 请求。另外补了回归测试，覆盖旧排队任务被 cancel 淘汰、其内部流程未执行，以及 cancel 后新任务仍可运行；相关 4 个 suite 共 16 个 test 通过，module-id registry 校验和 agent commit gate 均通过。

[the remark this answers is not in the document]
huhuanming: 补充审计后，这个 generation 修复存在新的取消作用域问题，应在合入前调整。`batchCreateFlowCancellationGeneration` 是整个 ServiceBatchCreateAccount 实例共享的，两个 start 入口及所有账户、钱包、调用场景共用同一 semaphore；而未改动的 ProcessingDialog 不仅在用户 Cancel 时调用 cancel，正常 Done 和 onClose 也会调用。于是任意弹窗正常结束都可能递增全局 generation，使此前已入队的另一个合法任务在开始前失败。合法调用来源包括 ServicePrimeTransfer 的导入恢复、ServiceAccount 建账户、addDefaultNetworkAccounts、BulkCopy 和 BatchCreateAccount，它们没有共同的 operation 标识。现有回归测试也明确验证的是“cancel 后 queuedFlow 一律不执行”，没有按账户、类型或来源过滤。建议保留 semaphore 串行化，但将取消绑定到 operationId/flowId，并区分显式 Cancel 与正常 Done/onClose；否则 PR 描述中的“preserving subsequent distinct work by queueing”并不总成立。

packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/hooks/addAccountFlowGuard.ts
```diff
@@ -0,0 +1,19 @@
+let inFlightAddAccountFlow: Promise<void> | undefined;
+
+export function runAddAccountFlowOnce(
+  flow: () => Promise<void>,
+): Promise<void> {
+  if (inFlightAddAccountFlow) {
+    return inFlightAddAccountFlow;
```
sidmorizon: **Different-wallet additions are silently dropped**

Severity: `non-severe`

While one addition runs, `runAddAccountFlowOnce` reuses it for every wallet. A second wallet reports completion without creating its requested account.

packages/kit-bg/src/services/ServiceBatchCreateAccount/ServiceBatchCreateAccount.ts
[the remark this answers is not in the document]
huhuanming: 已按审计结果在 c5e0651b30 收敛修复范围：删除全局 cancellation generation 及其“cancel 淘汰所有旧排队任务”的测试，不引入 operationId/flowId 或新的取消语义。原问题按两层处理：短时间重复创建由共享 in-flight guard 和 Preview isSubmittingRef 阻止；不同合法批量任务如果并发到达，则由现有 PR 内的同一 semaphore 串行执行，避免共享 progressInfo 被并发覆盖，但排队任务不会因另一个流程的 Done/onClose 被丢弃。cancel 恢复原有行为。聚焦回归 4 suites / 15 tests 通过，agent commit gate 通过。

packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/hooks/addAccountFlowGuard.ts
huhuanming: Fixed in cfcbf37c0c. The in-flight guard is now keyed by walletId: concurrent calls for the same wallet share one task and cannot allocate the next indexed account twice, while calls for different wallets get independent Promises and both execute. Added regression coverage proving same-wallet dedupe, different-wallet independence, and guard release after rejection. Focused result: 4 suites / 16 tests passed; agent commit gate passed.

originalix:

Review of commit `2b268862a35a` did not finish successfully.

sidmorizon:
Claude review session: https://claude.ai/code/session_01UBBYZnJNd98ugScoC7W4bi

packages/kit-bg/src/services/ServiceAccountProfile.ts
```diff
@@ -1329,62 +1329,73 @@ class ServiceAccountProfile extends ServiceBase {
...
-              });
-            }
-            return;
+    for (const account of accounts) {
+      if (account.accountId) {
+        if (accountUtils.isOthersAccount({ accountId: account.accountId })) {
+          const dbAccount = dbAccountById.get(account.accountId);
+          const accountAddress = account.accountAddress ?? dbAccount?.address;
+          const xpub =
+            account.xpub ?? accountUtils.pickXpubFromDBAccount(dbAccount);
```
sidmorizon: **Empty fields hide account balances**

Empty address fields make `getAllNetworkAccountsValueByAccountIdBatch` preserve empty strings instead of using the account snapshot. The [query builder](packages/kit-bg/src/services/ServiceAccountSelector.ts:978-1003) emits empty strings, so affected accounts lose their persisted balances.

Suggested fix:

```
          const dbAccount = dbAccountById.get(account.accountId);
          const accountAddress = account.accountAddress || dbAccount?.address;
          const xpub =
            account.xpub || accountUtils.pickXpubFromDBAccount(dbAccount);
```

huhuanming: Fixed in 5c8d28a797. Empty accountAddress/xpub values now fall back to the single account-table snapshot, preserving the pre-optimization truthy fallback semantics. Added focused coverage for both DB address and DB xpub fallback paths. The targeted suite passes 7/7 tests, and the agent commit gate passes.

patches/@onekeyfe+react-native-image+3.0.117.patch
```diff
@@ -0,0 +1,1203 @@
...
++    guard let image = memoryCachedImage(for: requestURL)
++      ?? (requestURL == rawURL ? nil : memoryCachedImage(for: rawURL)) else {
++      return false
++    }
++    displayState = .image
++    requestActive = requestIsActive
++    skeletonIndicator.stopAnimatingIndicator()
++    fallbackLayer.isHidden = true
++    resetImageTransition()
++    hostView.image = image
```
originalix: P1: [Memory-cache fast path never completes reusable image callbacks]

When this memory-cache branch is hit, it assigns `hostView.image` and returns without calling `onLoad` or `onLoadEnd`. The new native-list `hideUntilLoaded` callers use `onLoad` to unhide overlays and secondary images, so preloaded cached badges can remain invisible indefinitely on iOS.

Route this branch through the same successful-load completion path as the normal SDWebImage callback, including the memory cache type and completion callbacks, before returning.

originalix:

Review of commit `3fd6314e0e3d` finished, but some comments could not be published automatically.

packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/WalletDetailsV2.tsx
```diff
@@ -498,21 +520,145 @@ function WalletDetailsViewV2({ num }: IWalletDetailsProps) {
...
+      candidateList.hasResolved &&
+      candidateList.identity === presentedList.identity
+    ) {
+      setPresentedList(candidateList);
+    }
+  }, [candidateList, presentedList.identity]);
+  useEffect(() => {
+    if (
+      !candidateList.hasResolved ||
+      candidateList.identity === presentedList.identity
```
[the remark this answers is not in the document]
huhuanming: Fixed in d621e48514. Search text no longer participates in the presentation preload scope: rows for the current wallet/network/derive scope are presented immediately, so filtering does not retain the stale list or disable pointer events. The preload hold remains only for actual scope switches.

patches/@onekeyfe+react-native-image+3.0.117.patch
huhuanming: Fixed in d621e48514. The iOS memory-cache fast path now claims the active request terminal, emits onLoad(width, height, .memory) followed by onLoadEnd, and schedules onDisplay behind the existing generation guard. NativeList hideUntilLoaded images therefore become visible on memory hits as well.

originalix (approved):

PR #13280, opened 2026-09-09
fix: persist market chart height and selector favorites ｜OK-62519 & OK-62549

zy2000-star:
[OK-62519](<https://onekeyhq.atlassian.net/browse/OK-62519>) [OK-62549](<https://onekeyhq.atlassian.net/browse/OK-62549>)

----------

## Summary

Fixes OK-62519 and OK-62549.

### OK-62519 — Persist Market desktop chart height

Market chart height was stored only in component state, so reopening the page or restarting the app reset it to the default 456px.

- Add a persisted Market desktop layout atom shared by Token, Top Coins, and Stock detail charts.
- Keep drag updates local and save the height when dragging ends. Keyboard adjustments are also saved.
- Apply the saved height when persisted state becomes available, including delayed hydration.
- Validate saved values and clamp the displayed height to the current viewport.
- Preserve the saved preference when the window temporarily shrinks or the chart enters fullscreen.

### OK-62549 — Enable favorites in the detail-page asset selector

The stock selector rendered a static star icon because favorites were not enabled in its column configuration. Top Coins rows used the regular token favorite action with empty chain IDs and addresses.

- Enable interactive favorite buttons in the stock selector.
- Resolve Top Coins favorite identities through `MarketListingStar`, using the asset ID to obtain the appropriate network and token address.
- Reuse existing loading, unavailable, and retry behavior for identity resolution.
- Record stock selector favorite actions with the `Search` source instead of `Homepage`.
- Keep the selector open and prevent row navigation when clicking a listing’s favorite button.

## Validation

- Commit checks passed, including lint, formatting, and TypeScript.
- Targeted tests cover height restoration on remount, delayed hydration, invalid saved values, viewport clamping, drag-end persistence, fullscreen behavior, and selector favorite wiring.
- Existing identity-resolution tests cover native assets, stock variants, address normalization, unavailable identities, request sharing, and retries.
- Electron verification confirmed that the adjusted chart height was restored to 480px after reload and the chart rendered successfully.
- Electron verification confirmed stock and Top Coins favorite toggling without closing the search popover. Test favorite changes were restored afterward.

## Verification limits

- Full application process restart has not been verified.
- Cross-device favorite synchronization was not tested; the development environment had cloud sync disabled.

[OK-62519]: https://onekeyhq.atlassian.net/browse/OK-62519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62549]: https://onekeyhq.atlassian.net/browse/OK-62549?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-68ca4fcc-ed1e-59f3-8e01-1c403a5d7e49)

sidmorizon:
Claude review session: https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -130,9 +148,9 @@ export function MarketDesktopChartContainer({
         return;
       }
       event.preventDefault();
-      setChartHeight(clampChartHeight(nextHeight, maxHeight));
+      saveHeight(clampChartHeight(nextHeight, maxHeight));
```
sidmorizon: # 🟠 P1 Keyboard resize silently discards the saved chart height on a short viewport

## Problem

`chartHeight` (lines 48-54) is the **display** value — already clamped to `maxHeight`. The drag path deliberately avoids writing a clamped no-op: `finishPointerDrag` only saves when `dragState.currentHeight !== dragState.startHeight` (line 120), so dragging against the clamp ceiling leaves the stored preference intact. `handleKeyDown` has no equivalent guard — it derives `nextHeight` from the already-clamped `chartHeight`, re-clamps, and calls `saveHeight(...)` unconditionally.

Concrete trace with `MARKET_DESKTOP_CHART_MIN_HEIGHT = 456` and `MARKET_DESKTOP_CHART_VIEWPORT_GUTTER = 160`:

1. Stored `chartHeight = 900` (set on a 1200px-tall window).
2. User shrinks the window to 800px → `maxHeight = 640`, displayed `chartHeight = 640`, stored value still `900`. This is correct and is exactly what the new test at `MarketDesktopChartContainer.test.tsx:99-105` asserts.
3. User focuses the handle and presses <kbd>ArrowDown</kbd> to grow the chart → `nextHeight = 640 + 24 = 664` → `clampChartHeight(664, 640) = 640` → `saveHeight(640)` overwrites the stored `900`.

Nothing moves on screen, but `marketDesktopLayoutAtom` (`packages/kit-bg/src/states/jotai/atoms/market.ts`) is now permanently `640`.

Before this PR the same line was `setChartHeight(...)` against local `useState`, so a clamped no-op had no lasting consequence. Now it writes through to persisted storage, which is what makes this a regression rather than a pre-existing quirk.

## Impact

A keypress that produces no visible change silently destroys the user's saved chart height. Restoring the window no longer restores the previous height, and the preference cannot be recovered. This contradicts the "preserve the saved preference when the window temporarily shrinks" behavior the PR implements for the drag and resize paths. The added tests cover keyboard adjustment and viewport clamping separately, but never the two combined, so the gap is not caught.

## Recommendation

Mirror the drag-path guard in `handleKeyDown` (lines 137-154) so a clamped no-op never reaches storage:

```ts
const next = clampChartHeight(nextHeight, maxHeight);
if (next === chartHeight) {
  return;
}
saveHeight(next);
```

A regression test combining a shrunk viewport with <kbd>ArrowDown</kbd>, asserting the persisted value is unchanged, would lock this in.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -103,7 +117,11 @@ export function MarketDesktopChartContainer({
       if (!dragState || dragState.pointerId !== event.pointerId) {
         return;
       }
+      if (dragState.currentHeight !== dragState.startHeight) {
+        saveHeight(clampChartHeight(dragState.currentHeight, maxHeight));
+      }
       dragStateRef.current = undefined;
+      setDragHeight(undefined);
```
sidmorizon, line 330: # 🟡 P2 Drag-end clears the local override before the shared atom commits, so the chart snaps back on the extension expand tab

## Problem

`finishPointerDrag` calls `saveHeight(...)` (line 121) and then synchronously clears the local override with `setDragHeight(undefined)` (line 124). The rendered height then falls back to `layoutState.chartHeight` (lines 48-54). That is only correct if the atom write lands synchronously.

- **Desktop / web — fine.** `wrapAtomPro` (`packages/kit-bg/src/states/jotai/utils/wrapAtomPro.ts:36-59`) resolves the updater and reaches `set(baseAtom, payload)` before its first real suspension, so the new value is readable in the same commit.
- **Extension UI — not fine.** `shouldProxyStateUpdateFromUiToBg = platformEnv.isExtensionUi` (`wrapAtomPro.ts:27-30`), so `doSet` takes the `await appGlobals.$jotaiBgSync?.proxyStateUpdateActionFromUiToBg(...)` branch and **returns without touching the local atom** (`wrapAtomPro.ts:45-51`). The UI store only updates when the background broadcasts back and re-enters with `$$isFromBgStatesSyncBroadcast` (`wrapAtomPro.ts:91-97`). Meanwhile `setDragHeight(undefined)` has already flushed in the same React batch, so the component renders the stale `savedHeight`.

This layout is reachable on the extension: `MarketDetailV2.tsx:200` gates on `media.gtLg && !platformEnv.isNative` (extension UI is not native), and the Market tab is shown everywhere except popup and side panel — `shouldShowMarketTab = !(platformEnv.isExtensionUiPopup || platformEnv.isExtensionUiSidePanel)` (`packages/kit/src/routes/Tab/router.ts:108-113`). The expand tab (`platformEnv.isExtensionUiExpandTab`) is a full-width browser tab, so it clears `gtLg`.

Before this PR the height was plain `useState`, so the commit was always synchronous and this could not happen on any platform.

## Impact

On the extension expand tab, releasing the drag makes the chart snap back to its pre-drag height and then jump to the new one once the background round trip completes — potentially several hundred ms under MV3 when the service worker has to wake. If the broadcast never arrives, the drag result is dropped entirely and the resize appears to have failed. The new test's mock (`MarketDesktopChartContainer.test.tsx:14-30`) applies the setter synchronously, which is why it does not surface this.

## Recommendation

Do not drop the local override until the shared value catches up. Keep the committed value in state rather than clearing it at line 124:

```ts
setDragHeight(dragState.currentHeight);
```

then clear it from an effect once the atom agrees:

```ts
useEffect(() => {
  if (dragHeight !== undefined && savedHeight === dragHeight) {
    setDragHeight(undefined);
  }
}, [dragHeight, savedHeight]);
```

Any equivalent that keeps the last committed height rendered while the write is in flight works; the point is that line 124 must not be the same synchronous batch as the `saveHeight` call on a platform where that write is asynchronous.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +42,46 @@ export function MarketDesktopChartContainer({
...
-    setChartHeight((height) => clampChartHeight(height, maxHeight));
-  }, [maxHeight]);
+    if (pendingHeight !== undefined && savedHeight === pendingHeight) {
+      setPendingHeight(undefined);
+    }
+  }, [pendingHeight, savedHeight]);
+
+  const saveHeight = useCallback(
+    (height: number) => {
+      setPendingHeight(height);
```
sidmorizon: # 🟠 P1 A resize back to the value the stale UI mirror still holds is never sent to bg, and the earlier in-flight write then reverts it

Thanks for the `pendingHeight` fix — it does remove the drag-release snap-back I flagged earlier. But making `pendingHeight` the interaction baseline introduces a new failure in the same async path.

## Problem

`saveHeight` (lines 76-84) sets `pendingHeight` and then writes `prev.chartHeight === height ? prev : { ...prev, chartHeight: height }`. On split-runtime targets, `wrapAtomPro` evaluates that updater against the **UI's own mirror** and bails out entirely when the result is reference-identical:

```ts
nextValue = update(prevValue);
if (prevValue === nextValue) {
  return;   // wrapAtomPro.ts:77-79 — before doSet, so nothing is proxied to bg
}
```

Nothing is sent, so no broadcast is ever produced. Previously this was harmless: the interaction baseline *was* the mirror, so a computed target could only equal the mirror when the user had genuinely not moved. Now the baseline is `pendingHeight` (line 51), which can be **ahead** of the mirror while a write is in flight — so a real user action can compute a target equal to the stale mirror value and be silently swallowed. The clear effect (lines 70-74) then matches that stale value, drops the override, and the still-in-flight earlier write lands and overwrites the display.

Keyboard repro — only two keypresses inside one bg round trip, starting from an acked 456:

```
uiMirror = 456, pending = undefined                      display 456
ArrowDown  -> clamped 480 != 456, saveHeight(480)
              pending = 480, {chartHeight:480} proxied   display 480   (in flight)
ArrowUp    -> baseline is pending 480, so 480-24 = 456
              saveHeight(456): updater returns prev
              -> identity -> EARLY RETURN, nothing sent
              effect: savedHeight 456 === pending 456 -> override cleared
                                                        display 456
ack for 480 arrives -> uiMirror = 480                    display 480, persisted 480
```

The user's last input (456) is lost and visibly reverted to 480. The drag variant is the same shape: drag out, drag back to the pre-drag value, release.

Scope: extension UI and native `main` with `enableNativeBackgroundThread` (`wrapAtomPro.ts:27-30`). Desktop/web commit synchronously, so the baseline never diverges and they are unaffected.

The same root cause — the override is discharged **only** by an exact-value acknowledgement that is not guaranteed to exist — has a second consequence: `setLayoutState` is fire-and-forget from a React handler and `proxyStateUpdateActionFromUiToBg` is awaited inside the async atom writer (`wrapAtomPro.ts:45-50`). If that round trip ever fails, no broadcast arrives, `pendingHeight` never clears, and the component is pinned to that height for its lifetime, ignoring every later atom update.

## Impact

On extension and native-bg-thread builds, a user who nudges the chart and nudges it back loses their final adjustment and watches the chart jump to the value they abandoned — the same symptom class this commit set out to remove, relocated into the correction path. A dropped round trip additionally freezes the chart height permanently.

## Recommendation

Guarantee that every commit produces an acknowledgement, and stop discharging the override on exact-value equality alone.

Minimal repair — drop the identity short-circuit so a new object is always written, which guarantees bg broadcasts the committed value back and the effect at lines 70-74 always fires:

```ts
setLayoutState((prev) => ({ ...prev, chartHeight: height }));
```

Then discharge the override defensively rather than by value match — e.g. record the dispatched value in a ref and clear `pendingHeight` when `savedHeight` changes to *any* value after that dispatch — so a dropped or failed round trip cannot pin the chart.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.test.tsx
```diff
@@ -3,8 +3,39 @@ import { useEffect } from 'react';
...
+  const React = jest.requireActual<typeof import('react')>('react');
+  return {
+    useMarketDesktopLayoutAtom: () => {
+      const [, refresh] = React.useState(0);
+      const set = React.useCallback(
+        (update: (prev: IMarketDesktopLayout) => IMarketDesktopLayout) => {
+          const next = update(mockSavedLayout);
+          mockPersist(next);
+          if (mockDelayWrites) {
+            mockPendingWrites.push(next);
```
sidmorizon, line 30: # 🟡 P2 The delayed-write test double cannot fail on the swallowed-write path, making the ack protocol look total

## Problem

The mock models `wrapAtomPro` only partially. It always computes `next = update(mockSavedLayout)` and always pushes it onto `mockPendingWrites` (line 26) — including the case where the updater returns `prev` unchanged, which is exactly the case where the real `wrapAtomPro` returns early and sends nothing (`packages/kit-bg/src/states/jotai/utils/wrapAtomPro.ts:77-79`).

The two async tests then hand-advance `mockSavedLayout` strictly in write order (lines 131, 134, 162, 167), so every pending value is guaranteed a matching acknowledgement. Under that harness the clear effect at `MarketDesktopChartContainer.tsx:70-74` can never miss, and the swallowed-write path cannot be expressed at all.

## Impact

"keeps the latest keyboard height until delayed writes catch up" and "does not snap back on release or let an earlier acknowledgement interrupt a new drag" read as proof that the pending/ack protocol is sound on the extension path, but neither can fail on its actual failure mode. They directly mask the P1 I filed on `MarketDesktopChartContainer.tsx:78` — a reviewer reading these tests would reasonably conclude the async path is covered.

I'd normally leave test-shape feedback out of a review at this level; flagging it only because it conceals a live defect.

## Recommendation

Mirror `wrapAtomPro`'s contract in the mock — when the updater returns the same reference, send nothing:

```ts
const next = update(mockSavedLayout);
if (next === mockSavedLayout) {
  return;
}
mockPersist(next);
```

Then add a case that commits a value and, before delivering its acknowledgement, commits back to the pre-drag value; assert the display and the persisted value both end on the user's last input. That test should fail against the current implementation and pass once the identity short-circuit in `saveHeight` is removed.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.test.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +43,71 @@ export function MarketDesktopChartContainer({
...
+      const request = saveRequestRef.current;
+      setPendingHeight(height);
+      const finish = () => {
+        if (saveRequestRef.current === request) {
+          setPendingHeight(undefined);
+        }
+      };
+      // Serialize bridge writes so an older save cannot overwrite the last
+      // input. Always send a new object, even if the UI mirror still matches.
+      saveQueueRef.current = saveQueueRef.current
```
sidmorizon: # 🟠 P1 The watchdog releases the UI override but nothing releases the write queue, so one stalled write blocks every later save

The serialization does close both halves of my previous finding — the identity short-circuit is gone at line 104, and `.then(finish, finish)` plus the watchdog discharge the override without needing an exact-value match. But the two mechanisms were given different escape hatches, and the queue got none.

## Problem

`saveQueueRef.current` is advanced **only** by the settlement of the `setLayoutState(...)` promise (lines 101-107). There is no timeout, no abort, no bypass. Meanwhile the commit added a bounded escape hatch for the *display* override — `MARKET_DESKTOP_CHART_SAVE_TIMEOUT` at line 18, applied by the watchdog at lines 71-80. The override is bounded at 5 s; the queue is unbounded.

On split-runtime targets that promise is a bg RPC: `proxyStateUpdateActionFromUiToBg` → `backgroundApi.setAtomValue` (`jotaiBgSync.ts:147`) → `callBackground` (`BackgroundApiProxyBase.ts:417-419`). There is no application-level timeout anywhere on that path — the only `timeout` in `BackgroundApiProxyBase.ts` is commented out (line 601). So from this component's perspective the settle time is genuinely unbounded.

## Impact

If a single write stalls or never settles — MV3 service-worker eviction/wake under load, a port disconnect that doesn't reject the pending callback, a wedged native bg thread — **every subsequent resize is queued behind it and never dispatched**, for the remaining lifetime of the component. What the user sees: the chart snaps back to the stale mirror 5 s after each resize (the watchdog fires on schedule), and nothing they do is ever persisted. Recovery requires unmounting the layout.

Note this is a regression in kind: the pre-serialization code dispatched every write immediately, so a stalled write could not block later ones. Even in the merely-slow case the asymmetry misleads — at t=5 s the display reverts to the stale height, and any drag or arrow key the user then performs takes that stale value as its baseline, silently discarding their earlier intent.

The new test at `MarketDesktopChartContainer.test.tsx:260-282` ("releases the local override when a bridge response never arrives") sets up exactly this state — `mockWriteResults[0]` is never settled — but asserts only that `aria-valuenow` recovers to `600`. It never fires another `keyDown` afterwards, so it passes with the queue permanently wedged.

## Recommendation

Give the queue the same bound as the override, and drive both from one settlement rather than two independent ones:

```ts
saveQueueRef.current = saveQueueRef.current
  .then(() =>
    Promise.race([
      Promise.resolve(setLayoutState((prev) => ({ ...prev, chartHeight: height }))),
      new Promise<void>((resolve) =>
        setTimeout(resolve, MARKET_DESKTOP_CHART_SAVE_TIMEOUT),
      ),
    ]),
  )
  .then(finish, finish);
```

That way the queue slot always settles and the watchdog at 71-80 becomes a backstop rather than the only bound. Extend the test at 260-282 with a second `keyDown` after the timer advance, asserting the new value reaches `mockPendingWrites`.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +43,71 @@ export function MarketDesktopChartContainer({
...
+        .then(() =>
```
sidmorizon: # 🟡 P2 Superseded queued writes are never dropped, so a keyboard burst costs one serialized round trip and one persisted write per keystroke

## Problem

The dequeued closure captures `height` and dispatches it unconditionally — it never checks whether `request` is still `saveRequestRef.current`. The `request` id computed at line 92 is used *only* inside `finish` (lines 94-98) to guard the override clear; it does not suppress the write itself. For a layout preference only the last value in a burst matters, but every intermediate one is now sent, and sent **serially**.

Arrow-key auto-repeat is the realistic trigger. `handleKeyDown` fires per repeat, React 18 flushes discrete events synchronously so `chartHeight` advances between repeats (via `pendingHeight`, line 54), and the `clampedHeight === chartHeight` guard therefore does not coalesce them. Holding <kbd>ArrowDown</kbd> walks from 456 to `maxHeight` in 24 px steps — roughly 12-20 writes on a typical desktop viewport.

## Impact

Before serialization these were dispatched concurrently; now each waits for the previous round trip, so persistence latency for the value the user actually wants becomes N × RTT instead of one RTT. A resize followed by a quick navigation away or window close can lose a final value that would previously have been in flight already. It also multiplies bridge and storage traffic for a single gesture — the pattern `jotaiBgSync.ts:20-30` documents as the cause of a prior main-thread freeze ("2395 setAtomValue … saturated the main JS thread").

It also makes the watchdog reachable with no bg fault at all: a 15-deep backlog at ~400 ms RTT drains in ~6 s, so the 5 s override expires mid-drain and the chart visibly crawls upward through the intermediate values as the remaining acks land.

## Recommendation

Drop superseded entries at dequeue — `finish` already has the matching guard, so this is one line with no semantic loss:

```ts
saveQueueRef.current = saveQueueRef.current
  .then(() => {
    if (saveRequestRef.current !== request) {
      return undefined;
    }
    return Promise.resolve(
      setLayoutState((prev) => ({ ...prev, chartHeight: height })),
    );
  })
  .then(finish, finish);
```

Only the newest pending value is ever worth sending.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +45,73 @@ export function MarketDesktopChartContainer({
...
+              // Always send a new object, even if the UI mirror still matches.
+              await Promise.resolve(
+                setLayoutState((prev) => ({ ...prev, chartHeight: height })),
+              );
+            },
+            timeout: MARKET_DESKTOP_CHART_SAVE_TIMEOUT,
+            timeoutRejectError: new OneKeyLocalError(
+              'Chart height save timed out',
+            ),
+          })(undefined);
```
sidmorizon: **Timed-out saves overwrite newer preferences**

`makeTimeoutPromise` starts the next save after five seconds without cancelling the timed-out write. A later completion can overwrite the newer persisted chart height.

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +45,73 @@ export function MarketDesktopChartContainer({
...
+    };
+  }, []);
+
+  const saveHeight = useCallback(
+    (height: number) => {
+      saveRequestRef.current += 1;
+      const request = saveRequestRef.current;
+      setPendingHeight(height);
+      const finish = () => {
+        if (isMountedRef.current && saveRequestRef.current === request) {
```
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +46,131 @@ export function MarketDesktopChartContainer({
...
+                chartHeight: height,
+                chartHeightUpdateId: request.id,
+              })),
+            );
+            // A timeout releases the queue, not the original write. Repair a
+            // late write even after unmount so the latest preference wins.
+            const latest = latestSaveRef.current;
+            if (timedOut && latest && latest !== request) {
+              enqueueSave(latest.height);
+            }
```
sidmorizon: **Unmounted save overwrites newer height**

A delayed save from an unmounted chart can finish after another chart saves and overwrite its persisted height. Component-local `latestSaveRef` cannot detect the newer write, so restart restores the stale height.

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +46,131 @@ export function MarketDesktopChartContainer({
...
+              setLayoutState((prev) => ({
+                ...prev,
+                chartHeight: height,
+                chartHeightUpdateId: request.id,
+              })),
+            );
+            // A timeout releases the queue, not the original write. Repair a
+            // late write even after unmount so the latest preference wins.
+            const latest = latestSaveRef.current;
+            if (timedOut && latest && latest !== request) {
```
sidmorizon: # 🟠 P1 The late-write repair is identity-guarded but allocates a new identity, so on a slow bridge it becomes a permanent 5-second write loop that outlives unmount

The acknowledgement rework does fix the mirror-vs-settlement problem — the override now clears only on a real id+height match against the atom mirror (lines 100-114), which is the right condition. This is about the repair path added alongside it.

*(Distinct from the line-145 thread about a stale cross-instance overwrite: that one is about **which** height a repair writes; this is about the repair **re-triggering itself indefinitely**. Different mechanism, different fix.)*

## Problem

The guard is object identity — `latest !== request` (line 143) — but every compensation goes through `enqueueSave` (line 144), which allocates a brand-new `request = { id: generateUUID(), height }` and assigns it to `latestSaveRef.current` (lines 118-119). So each compensation installs a `latest` that is identity-distinct from every in-flight request, **including the one whose landing comes next**. One compensation therefore guarantees the next landing also compensates.

Trace with bridge latency L = 8 s against `MARKET_DESKTOP_CHART_SAVE_TIMEOUT` = 5 s, from two keypresses (heights A then B):

```
t=0   keypress A -> req1; cb1 dispatches write1
t=1   keypress B -> req2; latestSaveRef = req2; cb2 chained
t=5   write1 times out -> queue released -> cb2 dispatches write2
t=8   write1 LANDS; latest = req2 != req1  -> enqueueSave(B) -> req3
t=10  write2 times out -> cb3 dispatches write3
t=13  write2 LANDS; latest = req3 != req2  -> enqueueSave(B) -> req4
t=18  write3 LANDS; latest = req4 != req3  -> req5
...   steady state: one new write every 5 s, forever
```

The height stops changing after B, but the identity never does. The queue chain doesn't damp it either: the `.then(ok, err)` at lines 154-169 swallows the timeout rejection, so every slot settles and the next compensation always dispatches. It terminates only if latency drops under 5 s (a write completing with `timedOut === false`) or the app restarts.

## Impact

A single resize on a degraded bridge leaves a perpetual background loop: every 5 s a `setLayoutState` → `proxyStateUpdateActionFromUiToBg` → bg `setAtomValue` → persist → `broadcastStateUpdateFromBgToUi`. Because each payload carries a fresh UUID (line 137), the value genuinely differs every iteration, so `wrapAtomPro`'s `prevValue === nextValue` short-circuit never trips and **every iteration writes to persistent storage**. On the extension this repeatedly wakes the MV3 service worker; on native dual-thread it feeds the bg→main broadcast path that `jotaiBgSync.ts:20-30` documents as a prior main-thread freeze cause.

It also survives navigating away: line 122 guards only `setPendingSave`, not the queue chaining, and the comment at lines 140-141 deliberately keeps repair alive after unmount. So the loop runs for the rest of the app's lifetime with the chart long gone and no user involved.

## Recommendation

Make the repair value-based, so a compensation that writes the same height cannot seed another:

```ts
const latest = latestSaveRef.current;
if (timedOut && latest && latest !== request && latest.height !== height) {
  enqueueSave(latest.height);
}
```

In the trace above: write1 wrote A while latest is B → repair fires (correct); write2 wrote B and latest is B → no repair → loop terminates. Tagging compensation-originated requests and refusing to compensate from one would work equally well.

Worth a test that keeps the *compensating* write stalled — the new case at `MarketDesktopChartContainer.test.tsx:182-242` stops one step short by resolving `mockWriteResults[2]` normally (lines 226-232), so the loop never gets a chance to appear and the `expect(jest.getTimerCount()).toBe(0)` at line 237 still passes.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +46,131 @@ export function MarketDesktopChartContainer({
...
+          ),
+        })(undefined).then(
+          () => {
+            // RPC completion can precede the UI broadcast. Keep the overlay
+            // until that acknowledgement, with a bound for a lost broadcast.
+            if (
+              isMountedRef.current &&
+              pendingSaveRef.current?.id === request.id
+            ) {
+              acknowledgementTimerRef.current = setTimeout(
```
sidmorizon: # 🟡 P2 The lost-broadcast fallback discards the override on the one path where it is guaranteed to render the wrong height

## Problem

The acknowledgement timer at line 162 is armed only in the **fulfilled** branch of the write — that is, after bg has accepted and persisted the new height. It then fires 5 s later only if the match effect at lines 100-114 has *not* matched. Since the RPC succeeded, that combination means precisely one thing: the mirror is stale because the broadcast never arrived, or was coalesced away (`jotaiBgSync.flushBroadcastMicroBatch` dedupes same-atom payloads last-write-wins within a microtask, so an earlier request's id can be dropped before it reaches the UI).

At that moment `clearPendingSave` runs and `chartHeight` (lines 62-69) falls back to `savedHeight` — the value the code already knows bg has replaced.

## Impact

The chart snaps back to the old height and stays there for the rest of the session while storage holds the new one, so display and storage disagree persistently. A reload or remount then shows the new value, which presents to users as a flaky, non-reproducible bug. This is the original snap-back symptom reintroduced on a narrower path — delayed 5 s and requiring a dropped or coalesced broadcast — by newly added code.

## Recommendation

On acknowledgement timeout the RPC result is authoritative, so resynchronise rather than discard. Either re-dispatch the same payload once as a self-healing broadcast retry, or force-set the local atom through the existing `$$isForceSetAtomWithoutProxy` escape hatch in `wrapAtomPro` (lines 98-105 there), and only then clear `pendingSave`:

```ts
acknowledgementTimerRef.current = setTimeout(() => {
  // RPC succeeded, so the mirror is stale rather than the write lost.
  // Re-assert the committed value instead of reverting the display.
  ...
  clearPendingSave(request.id);
}, MARKET_DESKTOP_CHART_SAVE_TIMEOUT);
```

If you keep the plain clear, at minimum log the divergence so a lost broadcast is diagnosable instead of silent.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -41,22 +46,131 @@ export function MarketDesktopChartContainer({
...
+                () => clearPendingSave(request.id),
+                MARKET_DESKTOP_CHART_SAVE_TIMEOUT,
+              );
+            }
+          },
+          () => clearPendingSave(request.id),
```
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -9,11 +9,44 @@ import type {
...
+  for (;;) {
+    current.started = true;
+    await current.write();
+    const latest = latestChartHeightSave;
+    if (!latest || latest === current) {
+      return;
+    }
+    // Repair even a previously acknowledged write. Reuse the user request so
+    // slow repairs cannot create new generations and sustain a write loop.
+    repairListeners.forEach((listener) => listener(latest));
```
sidmorizon, line 50: # 🟠 P1 Override adoption is cross-instance but override release is instance-local, so an adopted override can be pinned forever

The module-level coordination genuinely fixes the four earlier findings — the repair loop is gone (compensation reuses the request object instead of minting one), the timeout is queue-release only, the watchdog re-dispatches instead of discarding, and `latestChartHeightSave` really is shared across navigation. This is about an asymmetry the new design introduces.

## Problem

**Adoption is global.** Line 46 pushes another instance's request into *every* mounted chart:

```ts
repairListeners.forEach((listener) => listener(latest));
```

and each listener (lines 121-124) sets that instance's `pendingSaveRef.current` and `pendingSave`.

**Release is not.** Both release mechanisms stay bound to the instance whose `saveHeight` built the request:

1. **Failure release, lines 166-169.** `request.write`'s catch calls `clearPendingSave(request.id)` — but that `clearPendingSave` is the `useCallback` captured from the *originating* instance, so it only clears that instance's `pendingSaveRef`.
2. **Watchdog, lines 178-193.** `startAcknowledgementWatchdog` is reachable only from the originating instance's own `saveQueueRef` continuation, and is gated on that instance's `isMountedRef.current`.

There *is* one global release — the effect at lines 133-141, which keys off module-level `latestChartHeightSave` and so fires on every mounted instance. But it requires the broadcast to land (`layoutState.chartHeightUpdateId === latestChartHeightSave?.id && savedHeight === latestChartHeightSave?.height`). An instance that adopted a request therefore has **no timeout net and no failure notification** — only the happy path.

Concrete failure, two instances (`MarketDetailV2` keeps unfocused screens alive via `useIsFocused` / `shouldKeepEmbeddedSwapMounted`, so a chart screen pushed over a chart screen gives exactly this):

1. Instance **A** resizes → `R1`; `latestChartHeightSave = R1`; write `W1` in flight.
2. Instance **B** mounts, resizes → `R2`; `latestChartHeightSave = R2`; `W2` in flight.
3. User pops back; **B unmounts** — listener removed, `isMountedRef = false`. A is the only live instance, holding `R1`.
4. `W2` rejects → line 167 runs **B's** `clearPendingSave` on a dead instance. A is untouched.
5. `W1` resolves → A's `persistChartHeight` loop reaches line 40, sees `latest === R2 !== R1`, fires line 46 → **A adopts `R2`** and renders `R2.height` → line 47 `current = R2` → `await R2.write()` rejects again → line 167 again clears only dead B → loop exits by throw.
6. A is pinned at `R2.height` indefinitely. Storage holds `R1.height`. No watchdog was ever armed for `R2` on A, and the effect at 133 can never match because `R2`'s broadcast will never exist.

## Impact

The surviving chart renders a height that was never persisted and simultaneously ignores every later change to `marketDesktopLayoutAtom` — line 91 is `dragHeight ?? pendingSave?.height ?? savedHeight`, so a set `pendingSave` masks `savedHeight` unconditionally. Recovery needs a fresh user resize or a remount. A milder and more reachable variant needs no rejection at all: a *successful* repair write whose broadcast is dropped or micro-batch-coalesced leaves the adopting instance stuck with no retry, because the one-shot watchdog belongs to the unmounted originator.

## Recommendation

Make release symmetric with adoption. Either:

- **(a)** add a module-level release fan-out — a `releaseListeners` set (or reuse `repairListeners` with a nullable payload) invoked from the catch at line 167 and from `clearPendingSave`, so every instance holding that id drops it; or
- **(b)** mark the request itself — set `request.failed = true` in the catch at 167 and extend the effect at 133-141 to also clear when `pendingSave.id === latestChartHeightSave?.id && latestChartHeightSave.failed` — and arm the watchdog inside `onRepair` (guarded by `isMountedRef` and the request id) so an adopted override inherits the same bounded retry the originator gets.

## Test gap

The new cases do exercise the cross-instance and late-write paths honestly, but every one of them serialises the instances (`first.unmount()` at line 256, `second.unmount()` at 282, `unmount()` at 329). `repairListeners` is never driven with more than one live listener, and no test rejects a write on the repair leg at line 47 — which is exactly the region this defect occupies.

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -49,22 +84,168 @@ export function MarketDesktopChartContainer({
...
+                ...prev,
+                chartHeight: height,
+                chartHeightUpdateId: request.id,
+              })),
+            );
+            request.accepted = true;
+          } catch (error) {
+            if (!request.accepted) {
+              releaseListeners.forEach((listener) => listener(request.id));
+            }
```
sidmorizon: **Failed retry freezes chart height**

When an accepted save loses its broadcast, a failed retry leaves `pendingSave` active because `request.accepted` remains true. Later persisted heights cannot replace the frozen pending height.

zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketDesktopChartContainer.tsx
```diff
@@ -49,22 +86,197 @@ export function MarketDesktopChartContainer({
...
+        acknowledgementTimerRef.current = setTimeout(() => {
+          if (
+            latestChartHeightSave === request &&
+            pendingSaveRef.current?.id === request.id &&
+            !request.acknowledgementRetried
+          ) {
+            // Adopting instances can retry after the originator unmounts, but
+            // all instances share one retry allowance for this user request.
+            request.acknowledgementRetried = true;
+            void persistChartHeight(request).catch(() => undefined);
```
sidmorizon, line 159: # 🟡 P2 The watchdog retry is unbounded, so a retry that hangs rather than rejects still pins the override indefinitely

The `releasedSave` mirror fallback does fix the reported freeze — the fan-out at lines 232-238 is now unconditional and passes `request.accepted ? request.height : undefined`, and gating the fallback on atom-object identity (line 111) means any save from any runtime invalidates it and `savedHeight` wins. That closes the cross-runtime case properly. This is about the one remaining path it can't reach.

## Problem

The *initial* dispatch is bounded — `persistChartHeight(request)` runs inside `makeTimeoutPromise({ timeout: MARKET_DESKTOP_CHART_SAVE_TIMEOUT, … })` (lines 259-268). The retry is not:

```ts
request.acknowledgementRetried = true;
void persistChartHeight(request).catch(() => undefined);   // line 159
```

`makeTimeoutPromise` never cancels `asyncFunc` (`packages/shared/src/background/backgroundUtils.ts:179-224`), so every release path — both the new `releasedSave` fallback and the older `!accepted` release — lives inside `request.write()`'s `catch` (lines 232-240) and fires only if the write promise **rejects**. A write that hangs instead never reaches it.

At that point nothing else can release the override:
- `acknowledgementRetried` is already `true`, so `startAcknowledgementWatchdog` (lines 143-155) will not re-arm.
- The broadcast-match effect (lines 198-206) requires `layoutState.chartHeightUpdateId === latestChartHeightSave?.id`, which a foreign runtime's save can never satisfy.
- `pendingSave` is still set, and it takes precedence at line 110 over both `releasedSave` and `savedHeight`.

## Impact

On the split-runtime targets this machinery exists for — iOS, Android, browser extension (desktop/web are single-runtime and clear immediately via line 198) — if the bg bridge hangs rather than rejecting, `pendingSave` stays set for the lifetime of the mounted instance. `chartHeight` keeps returning the pending height and masks every subsequent `savedHeight`, including one persisted by another extension tab. That is the same symptom the `releasedSave` fix was written to remove, surviving on the one path it doesn't cover.

Storage is unaffected and it self-heals on unmount/remount or the next local resize, which is why this is P2 rather than P1. But the trigger isn't hypothetical: a hanging bridge write is precisely what `makeTimeoutPromise` was added to the initial path to defend against, so the asymmetry is the thing to fix.

## Recommendation

Bound the retry the same way as the initial write, and route its timeout through the same release fan-out:

```ts
request.acknowledgementRetried = true;
void makeTimeoutPromise({
  asyncFunc: () => persistChartHeight(request),
  timeout: MARKET_DESKTOP_CHART_SAVE_TIMEOUT,
  onTimeout: () => {
    releaseListeners.forEach((listener) =>
      listener(request.id, request.accepted ? request.height : undefined),
    );
  },
  timeoutRejectError: new OneKeyLocalError('Chart height retry timed out'),
})(undefined).catch(() => undefined);
```

That gives every armed override a guaranteed terminal release without touching the in-flight semantics the existing tests pin (the `480`-held-against-a-foreign-`600` cases).

[Claude session ↗](https://claude.ai/code/session_01CdCqKJkVFuX3yk4bo2pPgs)

---

originalix:

Review of commit `8a1437c023d0` finished, but some comments could not be published automatically.

originalix (approved):

PR #13218, opened 2026-09-07
fix: aggregate token tab auto-enable, network sheet sizing, uncreated-address value, cold-start home images, custom token keyboard dismiss, tooltip close(OK-61863, OK-61860, OK-61879, OK-61505, OK-61527, OK-61417, OK-61524)

weatherstar:
[OK-61863](https://onekeyhq.atlassian.net/browse/OK-61863) [OK-61860](https://onekeyhq.atlassian.net/browse/OK-61860) [OK-61879](https://onekeyhq.atlassian.net/browse/OK-61879) [OK-61505](https://onekeyhq.atlassian.net/browse/OK-61505) [OK-61527](https://onekeyhq.atlassian.net/browse/OK-61527) [OK-61417](https://onekeyhq.atlassian.net/browse/OK-61417) [OK-61524](https://onekeyhq.atlassian.net/browse/OK-61524)

----------

## Summary
- Aggregate token details: jumping to a member tab now enables only the landed network and toasts once; the "Select network" and "Contract address" sheets size to their content instead of a fixed 92% snap; the aggregate network selector no longer shows an asset value next to "Create address" rows.
- Home cold start: wallet tab support and the banner image no longer jitter or flash a skeleton on the first frame (swrKey snapshot + image prewarm).
- Add custom token page: a blank tap or drag dismisses the keyboard.
- Tooltip (web/desktop): closes on trigger press and before a modal is pushed, so it never lingers above a new modal.
- Tests: lock background RPC error rehydration against a getter-only `constructorName` (regression guard for the "loading forever" RPC hang).

## Intent & Context
QA reported a batch of App-6.6.0 wallet issues, most of them in the aggregated token flows on Android:
- OK-61863: with only Arbitrum enabled in All Networks, tapping the 4th network in the USDC overview enabled every network in between and stacked 4 "network enabled" toasts; returning to Overview toasted again.
- OK-61860: the "select more networks" sheet on Android had a large blank band under the last row and the sheet moved oddly when scrolling.
- OK-61879: in the aggregate network selector (Receive / Buy), networks without a created address showed "$0.00" next to the "Create" button.
- OK-61505: home tabs jittered and the banner showed a skeleton on cold start.
- OK-61527: the keyboard could not be dismissed by tapping blank space on the custom token page.
- Tooltip: a hover tooltip stayed open above a freshly pushed modal (desktop home "Copy address" → WalletAddress modal).
[...]

PR #13281, opened 2026-09-09
fix: native cold-start home first paint and home tab pane freeze timing(OK-61505, OK-62565)

weatherstar:
[OK-61505](<https://onekeyhq.atlassian.net/browse/OK-61505>) [OK-62565](<https://onekeyhq.atlassian.net/browse/OK-62565>)

----------

## Summary
- Native cold start: read the contextAtom snapshot after the travel-mode runtime launch is acknowledged, prime the first-paint home images (banner cards + header network logos) before the app mounts, and snapshot the fiat buy/sell support flag so the home entry paints enabled on the first frame (OK-61505).
- Native home header: resolve the All Networks trigger statically instead of through a lazy chunk (OK-61505).
- Home tabs: keep the outgoing pane rendered across the pager slide and thaw the pressed target on the tab press, so switching tabs no longer blanks the content area or snaps the collapsible header (OK-62565).

## Intent & Context
QA reopened OK-61505 on the 6.6 TestFlight build: after #13218 the tab row no longer reflowed, but on every cold start the banner image still showed a skeleton behind already-rendered text, the Buy/Sell entry painted later than Send/Receive, and the header network selector painted last. The guiding rule is that on a second launch every element already has a snapshot and must paint from it on the first frame.

OK-62565 reported slow in-page tab switches (especially to History) with a fully blank content area for up to ~2 s; the same mechanism also makes the collapsible header jump between collapsed and expanded when switching Spot/Perps.

## Root Cause
- **Snapshot never read at bootstrap.** Since the mobile travel mode landed (#13185), every synchronous storage read is masked (returns `undefined`) until `completeTravelModeRuntimeLaunchAcknowledgement` runs. `bootstrapNativeStorage` read the contextAtom snapshot before that step, so `__ONEKEY_CTX_ATOM_SNAPSHOT__`, `warmCriticalIcons`, and the banner prewarm from #13218 silently never ran on native. Provider hydration still worked through its storage fallback, which hid the breakage. Confirmed on the iOS simulator: the bootstrap log line never appeared and the image prime reported 0 images.
- **Image prime was fire-and-forget and gated in the wrong place.** `GlobalJotaiReady` only runs its prime when jotai is not yet ready; on native the app mounts after jotai hydration, so that gate never executed. `prewarmColdStartImagesFromSnapshot` also defaulted to `awaitPreload: false`. The native `OneKeyImage` shows the skeleton unless the SDWebImage memory cache already holds the key, so the images must be decoded before the first layout.
- **Buy/Sell entry.** `useSupportNetworkId` used `initResult: false` + a 100 ms debounce + an HTTP `fiat-pay/list` round trip with no swr snapshot, and the entry is fail-closed on it, so every cold start painted it disabled until the response landed.
- **Network selector.** In All Networks mode the header renders `LazyAllNetworksManagerTrigger` (dynamic import, `fallback = null`); the compat avatars are remote CDN logos that were not part of the cold-start prewarm.
- **Tab switch blank + header jump.** `FreezeInactiveHomeTab` freezes inactive panes with react-freeze (renders nothing). The collapsible tab view flips `focusedTab` half-way through the pager slide, so the outgoing pane vanished mid-animation while the incoming pane was still frozen; a frozen target list also ignores the tab view's scroll-offset sync, so the header snapped to the target's stale collapse state after it thawed. The half-way flip is upstream library behavior at every patch revision, not a #13218 regression.

## Design Decisions
- Snapshot hydration moved into `hydrateColdStartSnapshotAfterRuntimeLaunch`, called from `NativeStorageBootstrapRoot` between the runtime-launch acknowledgement and jotai init, with `[StartupTiming]` file-log lines for both the hydration and the image prime so device runs can be verified from `app-latest.log`.
- `startColdStartImagePrewarm` splits the prewarm into a critical batch (banner images at 56, header network logos at `s(24)` — matching the `$6` avatar size and therefore the same TOS resize URL and decode thumbnail) that is awaitable, and a fire-and-forget remainder. `waitForColdStartCriticalImagesBeforeMount` waits at most 300 ms; on a warm disk cache the batch finishes during jotai init (measured 0–1 ms of added wait), and a first-ever launch has no snapshot and therefore nothing to wait for.
- Header logos in All Networks mode are read from the trigger's own `allNetCompat` swr entry using the same key inputs as `AllNetworksManagerTrigger`, so the prewarmed URLs are exactly what the first frame requests.
- `LazyAllNetworksManagerTrigger.native.tsx` re-exports the static component; native ships a single bundle so the lazy chunk only added a Suspense fallback frame. Web keeps the split chunk. The module-id registry was updated for the new file.
- `useSupportNetworkId` gets `swrKeys.fiatCryptoNetworkSupport` keyed by network + type (the bg check takes no account input); the live check still runs and corrects the snapshot.
- `homeTabFreeze.ts`: the pressed target (already tracked as `activeTabName` on the tab-bar press) counts as active immediately, and blur freezes only after 500 ms, covering the ~300 ms slide plus the tab view's ~250 ms scroll-sync window. Swipe gestures still thaw the target at the half-way flip; that path is unchanged.

## Changes Detail
- `apps/mobile/src/backgroundThread/bootstrapNativeStorage.ts`, `NativeStorageBootstrapRoot.tsx` (+ tests): snapshot hydration after runtime launch, bounded critical-image wait before mount, file logging.
- `packages/kit/src/utils/coldStartImagePreload.ts` (+ test): header network logo collector, critical/remainder split, `startColdStartImagePrewarm` / `waitForColdStartCriticalImages`.
- `packages/kit/src/views/FiatCrypto/hooks/index.ts`, `packages/shared/src/utils/swrCacheUtils.ts` (+ test): fiat network-support swr key.
- `packages/kit/src/components/AccountSelector/LazyAllNetworksManagerTrigger.native.tsx`, `apps/mobile/bundle-registry/module-id-registry.json`: static native trigger.
- `packages/kit/src/views/Home/pages/HomePageView.tsx`, `homeTabFreeze.ts` (+ test): press-aware delayed freeze for home tab panes.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile (iOS/Android); the fiat swr key and tab freeze timing also apply to Desktop / Web / Extension home.
- **Risk Areas**: native bootstrap ordering (snapshot read now sits between runtime-launch acknowledgement and jotai init; a failure there is caught and logged, never fatal); the 300 ms mount wait on launches with a snapshot but a cold image disk cache; inactive home panes now stay mounted 500 ms longer after blur.

## Test plan
- [x] Unit tests: cold-start prewarm split and header logo collection, fiat swr key identity, bootstrap ordering, home tab freeze timing.
- [x] iOS simulator (iPhone 17 Pro) cold start, single-network and All Networks: banner images, enabled Buy/Sell entry, header network avatars and all five tabs present on the first painted frame; `app-latest.log` shows `snapshot hydrated: 14 keys` and `critical images done: 9–10 images, waited ≤1ms`.
- [x] iOS simulator tab switches Spot ⇄ Perps with expanded and collapsed header: both panes visible during the slide, no blank frame, header position unchanged, Spot list keeps its scroll offset.
- [ ] Device (iOS TestFlight) re-test of OK-61505 and the OK-62565 History switch on a slow JS thread.

[OK-61505]: https://onekeyhq.atlassian.net/browse/OK-61505?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62565]: https://onekeyhq.atlassian.net/browse/OK-62565?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/utils/coldStartImagePreload.ts
+++ b/packages/kit/src/utils/coldStartImagePreload.ts
@@ -9,6 +9,7 @@ import {
   swrCacheUtils,
   swrKeys,
 } from '@onekeyhq/shared/src/utils/swrCacheUtils';
+import { EAccountSelectorSceneName } from '@onekeyhq/shared/types';
 
 import {
   type ITokenSize,
@@ -58,6 +59,11 @@ const WALLET_BANNER_IMAGE_LIMIT = 8;
 // resize URL + decode thumbnail as the first paint.
 export const HEADER_NETWORK_LOGO_SIZE = s(24);
 const HEADER_NETWORK_LOGO_LIMIT = 6;
+// Cold-start scope key of the home account-selector store (see
+// jotaiContextStore.buildJotaiContextStoreId: `store:accountSelector@<scene>`),
+// the same key SplashProvider / HomeOverviewContainer read.
+const HOME_ACCOUNT_SELECTOR_COLD_START_SCOPE_KEY = `store:accountSelector@${EAccountSelectorSceneName.home}`;
+const HOME_ACCOUNT_SELECTOR_NUM = '0';
 // AllNetworksManagerTrigger shows at most this many network avatars.
 const HEADER_ALL_NETWORKS_AVATAR_LIMIT = 2;
 const WALLET_TOKEN_OWNER_LIMIT = 2;
@@ -327,20 +333,30 @@ function collectHeaderNetworkImageItems({
   snapshot: IColdStartSnapshot;
 }) {
   const seen = new Set<string>();
-  const add = (uri: unknown) =>
-    addPreloadItem({ items, seen, uri, resizeWidth: HEADER_NETWORK_LOGO_SIZE });
-  for (const value of getSnapshotValuesByColdStartKey({
-    snapshot,
-    coldStartCacheKey: CONTEXT_ATOM_COLD_START_CACHE_KEYS.activeAccountsAtom,
-  })) {
-    const activeAccounts = isRecord(value) ? Object.values(value) : [];
-    for (const activeAccount of activeAccounts) {
-      if (seen.size >= HEADER_NETWORK_LOGO_LIMIT) {
-        return;
-      }
-      collectHeaderNetworkImageItemsFromActiveAccount({ activeAccount, add });
+  const add = (uri: unknown) => {
+    if (seen.size < HEADER_NETWORK_LOGO_LIMIT) {
+      addPreloadItem({
+        items,
+        seen,
+        uri,
+        resizeWidth: HEADER_NETWORK_LOGO_SIZE,
+      });
     }
+  };
+  // Only the home scene's slot 0 paints in the header on the first frame;
+  // other account-selector scenes (swap, perps, ...) also persist this atom
+  // but must not spend the critical prewarm budget.
+  const value =
+    snapshot[
+      `${HOME_ACCOUNT_SELECTOR_COLD_START_SCOPE_KEY}::${CONTEXT_ATOM_COLD_START_CACHE_KEYS.activeAccountsAtom}`
+    ];
+  if (!isRecord(value)) {
+    return;
   }
+  collectHeaderNetworkImageItemsFromActiveAccount({
+    activeAccount: value[HOME_ACCOUNT_SELECTOR_NUM],
+    add,
+  });
 }
 
 function collectWalletTokenImageUris({
--- a/packages/kit/src/utils/coldStartImagePreload.ts
+++ b/packages/kit/src/utils/coldStartImagePreload.ts
@@ -2,8 +2,13 @@ import { preloadImages } from '@onekeyhq/components/src/primitives/Image/preload
 import { s } from '@onekeyhq/components/src/utils/scale';
 import { CONTEXT_ATOM_COLD_START_CACHE_KEYS } from '@onekeyhq/shared/src/consts/jotaiConsts';
 import platformEnv from '@onekeyhq/shared/src/platformEnv';
+import accountUtils from '@onekeyhq/shared/src/utils/accountUtils';
 import networkUtils from '@onekeyhq/shared/src/utils/networkUtils';
 import { getHyperliquidTokenImageUris } from '@onekeyhq/shared/src/utils/perpsUtils';
+import {
+  swrCacheUtils,
+  swrKeys,
+} from '@onekeyhq/shared/src/utils/swrCacheUtils';
 
 import {
   type ITokenSize,
@@ -48,6 +53,13 @@ const COLD_START_IMAGE_PRELOAD_LIMIT = 96;
 // same resize URL + decode thumbnail as the first paint (cache-key match).
 export const WALLET_BANNER_IMAGE_SIZE = 56;
 const WALLET_BANNER_IMAGE_LIMIT = 8;
+// Logical size of the network avatar in the home header trigger (`$6`).
+// NetworkAvatar renders at this size, so the prewarm resolves to the same
+// resize URL + decode thumbnail as the first paint.
+export const HEADER_NETWORK_LOGO_SIZE = s(24);
+const HEADER_NETWORK_LOGO_LIMIT = 6;
+// AllNetworksManagerTrigger shows at most this many network avatars.
+const HEADER_ALL_NETWORKS_AVATAR_LIMIT = 2;
 const WALLET_TOKEN_OWNER_LIMIT = 2;
 const WALLET_TOKEN_LIMIT_PER_OWNER = 24;
 const SWAP_POSITION_OWNER_LIMIT = 3;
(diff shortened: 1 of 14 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-f49ebf0c-34c0-5a85-888b-ddaf095d40c4)

sidmorizon:
Claude review session: https://claude.ai/code/session_01KyuguAme3EBdXSQ9pmzeor

packages/kit/src/utils/coldStartImagePreload.ts
```diff
@@ -221,6 +233,116 @@ function collectWalletBannerImageItems({
...
+}: {
+  items: IImagePreloadItem[];
+  snapshot: IColdStartSnapshot;
+}) {
+  const seen = new Set<string>();
+  const add = (uri: unknown) =>
+    addPreloadItem({ items, seen, uri, resizeWidth: HEADER_NETWORK_LOGO_SIZE });
+  for (const value of getSnapshotValuesByColdStartKey({
+    snapshot,
+    coldStartCacheKey: CONTEXT_ATOM_COLD_START_CACHE_KEYS.activeAccountsAtom,
```
sidmorizon: # 🟡 P2 Header logo prewarm is not scoped to the home scene

## Problem

`collectHeaderNetworkImageItems` selects its snapshot entries through `getSnapshotValuesByColdStartKey({ snapshot, coldStartCacheKey: CONTEXT_ATOM_COLD_START_CACHE_KEYS.activeAccountsAtom })` (`coldStartImagePreload.ts:332-335`). That helper matches on the `::ctx:activeAccountsAtom` suffix only (`coldStartImagePreload.ts:190-192`), so it matches **every** account-selector scene, not just home.

Snapshot keys are `store:accountSelector@<sceneId>::ctx:activeAccountsAtom` (`accountSelector/actions.tsx:489-536`), and `flushCurrentAccountSelectorColdStartSnapshot` runs from `AccountSelectorEffects.tsx:145` for every mounted scene except `discover` (`actions.tsx:496-497`). Every other consumer of this atom in the snapshot scopes itself explicitly — `SplashProvider.tsx:97`, `HomeOverviewContainer.tsx:245`, `useSwapColdStartDisplayTokens.ts:28`, `JotaiContextStoreMirrorTracker.tsx:41` all key off `accountSelector@home` and read slot `0`. This collector does neither: it iterates `Object.values(value)` for **all** `num` slots of **all** scenes (`:336-337`) under one shared `HEADER_NETWORK_LOGO_LIMIT = 6` cap, and `return`s outright the moment the cap is reached (`:338-340`).

## Impact

This is the *critical* batch. `startColdStartImagePrewarm` awaits it, and `waitForColdStartCriticalImagesBeforeMount` blocks the native app mount on it for up to 300 ms (`apps/mobile/src/backgroundThread/bootstrapNativeStorage.ts:25-28`). Two consequences:

1. The 300 ms mount block and the 4-slot native preload queue are partly spent decoding logos for scenes that are not on the first paint — e.g. the swap scene's from/to networks, two `num` slots of their own.
2. Iteration order is `Object.entries(snapshot)` order from `JSON.parse(raw)`. A non-home scene appearing earlier in the snapshot can consume the whole 6-URI cap and trigger the early `return` before the home scene is ever reached — so the home header logo, the sole target of OK-61505 here, is silently dropped while the mount still pays the wait.

The new test only ever supplies a single `store:accountSelector@home::…` key (`coldStartImagePreload.test.ts:99`), so neither case is covered.

## Recommendation

Scope the scan the same way the four existing consumers do: filter the snapshot keys with `accountSelector@home` (or thread a scope key through `getSnapshotValuesByColdStartKey`) and read `value['0']` rather than `Object.values(value)`, so the 6-URI cap acts as a safety net rather than as the selector. Add a test case with a non-home scene ordered ahead of the home scene to pin the behaviour.

[Claude session ↗](https://claude.ai/code/session_01KyuguAme3EBdXSQ9pmzeor)

---

packages/kit/src/views/Home/pages/homeTabFreeze.ts
```diff
@@ -0,0 +1,40 @@
...
+  if (!focusedTab) {
+    return true;
+  }
+  return focusedTab === tabName || pressedTabName === tabName;
+}
+
+// `true` while the pane should be frozen. Unfreezing is immediate; freezing
+// waits HOME_TAB_FREEZE_DELAY_MS so the pane survives the pager animation.
+export function useHomeTabFreeze(isActive: boolean) {
+  const [frozen, setFrozen] = useState(false);
```
sidmorizon: # 🟡 P2 Inactive home panes now stay live 500 ms after mount

## Problem

`useHomeTabFreeze` initialises `frozen` to `false` unconditionally (`homeTabFreeze.ts:28`) and only flips it to `true` after `HOME_TAB_FREEZE_DELAY_MS` inside the effect (`:29-38`). The delay is documented as covering the **pager slide on blur** (`homeTabFreeze.ts:3-7`, `HomePageView.tsx:199-204`), but it is applied to the mount case too, where there is no blur and no animation to cover.

Before this change `FreezeInactiveHomeTab` passed `freeze={focusedTab ? focusedTab !== tabName : undefined}` straight to `DelayedFreeze`. `focusedTab` is already non-empty on the first render — `Tabs/Container.tsx:313-315` seeds the shared value with `initialTabName || tabNames[0] || ''`, and `useFocusedTab` reads that value into its initial state (`useFocusedTab.ts:13,30-33`; the native re-export at `useFocusedTab.native.ts:1` is the unpatched library hook, and `patches/react-native-collapsible-tab-view+8.0.1.patch` does not touch it). So inactive panes got `freeze === true` on their first render and `DelayedFreeze` committed the freeze on the next `setImmediate`. They now get `false` and stay fully live for 500 ms.

## Impact

On native `HomePageView.tsx:935-942` renders `tab.component` for every pane unconditionally (`platformEnv.isNative || …`). So on every home mount — cold start, and every `Tabs.Container` remount on wallet/account switch — `PerpsContainer`, `DeFiContainerWithProvider`, `NFTListContainerWithProvider` and `TxHistoryListContainerWithProvider` (`HomePageView.tsx:526-585`) keep running their effects, subscriptions and re-renders for 500 ms before being frozen. That is exactly the reconciliation cost the `FreezeInactiveHomeTab` comment says the wrapper exists to avoid (`HomePageView.tsx:191-197`), and it lands inside the cold-start window this PR is otherwise trying to shorten for OK-61505. Web/desktop/extension are affected too (the file is not platform-scoped), bounded there by the `mountedHomeTabIds` gate.

## Recommendation

Seed the state from the current activity so only *transitions* are delayed and mount freezes immediately:

```ts
const [frozen, setFrozen] = useState(() => !isActive); // homeTabFreeze.ts:28
```

The effect needs no change — it still re-sets `true` after the delay on blur and clears to `false` on reactivation — and the render-time `return isActive ? false : frozen` (`:39`) still guarantees an immediate thaw. Both existing cases in `homeTabFreeze.test.tsx` start from `isActive: true` and keep passing; add one starting from `isActive: false` asserting `result.current === true` at t=0.

[Claude session ↗](https://claude.ai/code/session_01KyuguAme3EBdXSQ9pmzeor)

---

packages/kit/src/views/Home/pages/homeTabFreeze.ts
```diff
@@ -0,0 +1,40 @@
...
+  pressedTabName,
+}: {
+  tabName: string;
+  focusedTab: string | undefined;
+  pressedTabName: string | undefined;
+}) {
+  if (!focusedTab) {
+    return true;
+  }
+  return focusedTab === tabName || pressedTabName === tabName;
```
sidmorizon, line 22: **Programmatic tab targets stay frozen**

When `jumpToTab` starts a switch, `isHomeTabActive` leaves the target frozen until the pager changes focus. The [recent-history shortcut](packages/kit/src/views/Home/pages/HomePageView.tsx:965-976) never updates `pressedTabName`, so the transition initially shows a blank pane.

packages/kit/src/views/Home/pages/homeTabFreeze.ts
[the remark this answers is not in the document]
weatherstar: Fixed in 2ef1fed850: `useHomeTabFreeze` now seeds `frozen` from `!isActive`, so panes that mount inactive freeze immediately and the 500 ms delay only covers active → inactive transitions. Added a test that starts from `isActive: false` and asserts `frozen === true` at t=0.

weatherstar: Fixed in 2ef1fed850: `useState(() => !isActive)` as suggested; mount freezes immediately, the effect still delays only the blur transition and thaws immediately on reactivation. Test added for the inactive-mount case.

packages/kit/src/views/Home/pages/HomePageView.tsx
```diff
@@ -919,7 +928,10 @@ export function HomePageView({
       >
         {pagerTabConfigs.map((tab) => (
           <Tabs.Tab key={tab.name} name={tab.name}>
-            <FreezeInactiveHomeTab tabName={tab.name}>
+            <FreezeInactiveHomeTab
+              tabName={tab.name}
+              pressedTabName={activeTabName}
```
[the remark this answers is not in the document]
weatherstar: Fixed in 2ef1fed850: `handleSwitchWalletHomeTab` now applies the same press-ahead update as the tab bar (`setActiveTabName` / `setActiveTabId` / `lastDisplayableTabNameRef`) before `jumpToTab`, so `SwitchWalletHomeTab` targets (RecentHistory, desktop tray) thaw before the pager starts sliding.

packages/kit/src/views/Home/pages/homeTabFreeze.ts
weatherstar: Fixed in 2ef1fed850: the programmatic switch (`handleSwitchWalletHomeTab`) now updates `activeTabName` before `jumpToTab`, so the target pane thaws ahead of the slide instead of at the halfway flip.

packages/kit/src/utils/coldStartImagePreload.ts
weatherstar: Fixed in 2ef1fed850: the collector now reads only `store:accountSelector@home::ctx:activeAccountsAtom` slot `0` (same scope the other snapshot consumers use); the 6-URI cap is now a safety net inside `add` rather than the selector. Added a test with a `swap` scene ordered ahead of the home scene, asserting only the home slot-0 logo is prewarmed.

originalix (approved):

PR #13282, opened 2026-09-09
fix: iPad onboarding v2 real-width breakpoints and recycled view autoresizing (OK-62575)

franco-chan:
[OK-62575](<https://onekeyhq.atlassian.net/browse/OK-62575>)

----------

## Summary

Onboarding v2 on iPad rendered the phone (`md`) layout in every orientation, and once the desktop-style layout was enabled a set of native-only rendering bugs surfaced (overlapping list items in dialogs, page-sized coloured slabs in the confetti, CTA / step rows filling their container). This PR makes onboarding v2 evaluate breakpoints against the real window width on iPad and fixes the native issues that path exposed.

## Root causes

1. **iPad never reaches `gtMd`.** `patches/@tamagui+web+1.108.0.patch` clamps tablets to `MAX_OVERRIDE_WIDTH = 767` (landscape: `width / 2`, portrait: `min(width, 767)`) because tab pages live in a half-width split-view pane. Onboarding is full-screen, so the clamp is wrong there. `useMedia().gtMd` and every `$gtMd` prop were `false` on iPad regardless of orientation.
2. **Native centered `Dialog` path was dead code on native.** It sits in an absolute-fill `Stack`, so Yoga measures the subtree in AtMost mode and `ListItem`'s `Pressable` (`flex: 1`) collapses to zero height; rows overlap ("Pick your device" dialog).
3. **Recycled Fabric views kept a flexible `autoresizingMask`.** `@onekeyfe/react-native-tab-view` stamps `autoresizingMask = [.flexibleWidth, .flexibleHeight]` on its React child views. `RCTViewComponentView.prepareForRecycle` does not reset the mask, so after the home (split view + tabs) re-lays out under onboarding, those views enter the recycle pool and are reused by unrelated `<View>`s, which UIKit then autoresizes to whatever superview they land in. Confirmed with an lldb `recursiveDescription` dump: exactly the oversized views carried `autoresize = W+H`; Yoga sizes were correct.

## Changes

- `patches/@tamagui+web+1.108.0.patch`: adds `globalThis.$$onekeyNativeMedia.useRealWidth` bypass (dual-screen spanning branch unchanged) and a `refresh()` hook that re-evaluates every media key.
- `packages/components/src/hooks/nativeTabletRealWidthMedia{.native,}.ts`: ref-counted `acquire/releaseNativeTabletRealWidthMedia()`, active only on `isNativeIOSPad` (Android tablets intentionally excluded). Global type in `@types/globals.d.ts`.
- `packages/kit/src/routes/Modal/Navigator.tsx`: `OnboardingNavigator` acquires it in a layout effect, so the first painted frame of every onboarding v2 page is already `gtMd` (covers all entry points; several v2 pages use no shared shell).
- `packages/components/src/composite/Dialog/index.tsx`: native centered dialog wraps its content in a `ScrollView` (unconstrained measurement, and tall content can scroll). Native + `gtMd` only.
- `patches/@onekeyfe+react-native-tab-view+3.0.78.patch` (new): clear the child's `autoresizingMask` in `removeChild(atIndex:)`. Needs to be upstreamed to the package.
- `patches/react-native+0.81.5.patch`: `RCTViewComponentView.prepareForRecycle` resets `autoresizingMask` as defense in depth.

Native rebuild required (two native patches).

## Verification

iPad Pro 11-inch (M5) simulator, iOS 26.5, Debug build:
- GetStarted shows the three-card layout, Terms left-aligned, language selector with label.
- "Pick your device" → Other Brands dialog renders two distinct rows.
- Check & Update: genuine + firmware success, confetti, Continue — restarted the app 3 times × 3 full runs, no oversized views.
- iPhone behaviour unchanged (the bypass is a no-op below 768pt).

## Follow-ups

- Upstream the `removeChild` mask reset to `@onekeyfe/react-native-tab-view`.
- When porting to `x` (tamagui 2.7.7) the tablet clamp lives in `patches/@tamagui+react-native-media-driver+2.7.7.patch` (`getMediaDimensions()`); the bypass has to be applied there.
- The onboarding-time media switch still re-lays out the hidden home underneath; acceptable for now, revisit if it shows up in profiling.

[OK-62575]: https://onekeyhq.atlassian.net/browse/OK-62575?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/components/src/hooks/nativeTabletRealWidthMedia.native.ts
+++ b/packages/components/src/hooks/nativeTabletRealWidthMedia.native.ts
@@ -1,0 +1,39 @@
+import platformEnv from '@onekeyhq/shared/src/platformEnv';
+
+// The patched Tamagui native media driver (patches/@tamagui+web+*.patch) clamps
+// tablets to phone breakpoints because tab pages live in a half-width
+// split-view pane. Full-screen surfaces such as onboarding v2 are not inside a
+// pane, so they hold real window-width breakpoints for their lifetime.
+// Ref-counted so overlapping holders never release early.
+let holders = 0;
+
+function syncNativeMedia(useRealWidth: boolean) {
+  const control =
+    globalThis.$$onekeyNativeMedia ??
+    (globalThis.$$onekeyNativeMedia = { useRealWidth: false });
+  if (control.useRealWidth === useRealWidth) {
+    return;
+  }
+  control.useRealWidth = useRealWidth;
+  control.refresh?.();
+}
+
+export function acquireNativeTabletRealWidthMedia() {
+  if (!platformEnv.isNativeIOSPad) {
+    return;
+  }
+  holders += 1;
+  if (holders === 1) {
+    syncNativeMedia(true);
+  }
+}
+
+export function releaseNativeTabletRealWidthMedia() {
+  if (!platformEnv.isNativeIOSPad) {
+    return;
+  }
+  holders = Math.max(0, holders - 1);
+  if (holders === 0) {
+    syncNativeMedia(false);
+  }
+}
--- a/packages/kit/src/routes/Modal/Navigator.tsx
+++ b/packages/kit/src/routes/Modal/Navigator.tsx
@@ -1,10 +1,12 @@
-import { useEffect, useRef } from 'react';
+import { useEffect, useLayoutEffect, useRef } from 'react';
 
 import { useIsFocused } from '@react-navigation/native';
 
 import {
   EPageType,
   Theme,
+  acquireNativeTabletRealWidthMedia,
+  releaseNativeTabletRealWidthMedia,
   setGlassHeaderUIStyle,
   useThemeName,
 } from '@onekeyhq/components';
@@ -45,6 +47,17 @@ export function FullScreenPushNavigator() {
 }
 
 export function OnboardingNavigator() {
+  // Onboarding is full-screen (never inside a split-view pane), so iPad
+  // evaluates breakpoints against the real window width for as long as this
+  // root route lives. Layout effect, not passive: the switch must land before
+  // the first paint, otherwise iPad paints the phone layout and re-lays out
+  // everything mid-transition, which leaves animated views with stale frames.
+  useLayoutEffect(() => {
+    acquireNativeTabletRealWidthMedia();
+    return () => {
+      releaseNativeTabletRealWidthMedia();
+    };
+  }, []);
   // Onboarding forces a dark Theme for its content, so the iOS 26 glass header
   // bar must use the dark variant while onboarding is the foreground route —
   // otherwise it flashes the light variant (the app theme is usually light).
--- a/packages/kit/src/routes/Modal/Navigator.tsx
+++ b/packages/kit/src/routes/Modal/Navigator.tsx
@@ -47,17 +47,24 @@ export function FullScreenPushNavigator() {
 }
 
 export function OnboardingNavigator() {
+  const isFocused = useIsFocused();
   // Onboarding is full-screen (never inside a split-view pane), so iPad
-  // evaluates breakpoints against the real window width for as long as this
-  // root route lives. Layout effect, not passive: the switch must land before
-  // the first paint, otherwise iPad paints the phone layout and re-lays out
-  // everything mid-transition, which leaves animated views with stale frames.
+  // evaluates breakpoints against the real window width while this root route
+  // is in the foreground. The flag is global to the UI runtime, so it is held
+  // for the focused lifetime only: a V1 modal pushed on top, or the main tab
+  // tree after onboarding is replaced, must get the clamped breakpoints their
+  // narrower containers were designed for. Layout effect, not passive: the
+  // switch must land before the first paint, otherwise iPad paints the phone
+  // layout and re-lays out everything mid-transition.
   useLayoutEffect(() => {
+    if (!isFocused) {
+      return undefined;
+    }
     acquireNativeTabletRealWidthMedia();
     return () => {
       releaseNativeTabletRealWidthMedia();
     };
-  }, []);
+  }, [isFocused]);
   // Onboarding forces a dark Theme for its content, so the iOS 26 glass header
   // bar must use the dark variant while onboarding is the foreground route —
   // otherwise it flashes the light variant (the app theme is usually light).
@@ -71,7 +78,6 @@ export function OnboardingNavigator() {
   // tracks whoever is actually foreground; the unmount cleanup covers the
   // onboarding-replaced-by-main case where we never blur first.
   const appThemeName = useThemeName();
-  const isFocused = useIsFocused();
   const appGlassStyle = appThemeName === 'dark' ? 'dark' : 'light';
   if (platformEnv.isNativeIOS26Plus) {
     setGlassHeaderUIStyle(isFocused ? 'dark' : appGlassStyle);
--- a/patches/@tamagui+web+1.108.0.patch
+++ b/patches/@tamagui+web+1.108.0.patch
@@ -2,7 +2,7 @@ diff --git a/node_modules/@tamagui/web/dist/cjs/hooks/useMedia.native.js b/node_
 index 47addac..110de63 100644
 --- a/node_modules/@tamagui/web/dist/cjs/hooks/useMedia.native.js
 +++ b/node_modules/@tamagui/web/dist/cjs/hooks/useMedia.native.js
-@@ -160,11 +160,98 @@ function unlisten() {
+@@ -160,22 +160,125 @@
    }), dispose.clear();
  }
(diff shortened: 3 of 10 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-b13fa867-e728-541e-8537-9f06247e5531)

sidmorizon:
Claude review session: https://claude.ai/code/session_0194r7tpakXw1M2V2pcaMmJp

patches/@tamagui+web+1.108.0.patch
```diff
@@ -12,6 +12,18 @@ index 47addac..110de63 100644
...
++// OneKey: full-screen surfaces (onboarding v2) opt out of the tablet clamp so
++// iPad evaluates breakpoints against the real window width. App code flips
++// `useRealWidth` and calls `refresh()`; see
++// packages/components/src/hooks/nativeTabletRealWidthMedia.native.ts.
++const nativeMediaControl =
++  globalThis.$$onekeyNativeMedia ||
++  (globalThis.$$onekeyNativeMedia = { useRealWidth: false });
++let mediaUpdaters = [];
++nativeMediaControl.refresh = function () {
++  mediaUpdaters.forEach((update) => update());
```
sidmorizon, line 23: **Stale orientation checks restore phone layout**

When onboarding starts in the main runtime during a pending native orientation query, `refresh` briefly applies real-width breakpoints. The per-runtime `useRealWidth` flag is never rechecked on completion, so the callback restores the 767-point clamp. iPad onboarding remains in phone layout until another media event.

packages/kit/src/routes/Modal/Navigator.tsx
```diff
@@ -45,6 +47,17 @@ export function FullScreenPushNavigator() {
...
+  // evaluates breakpoints against the real window width for as long as this
+  // root route lives. Layout effect, not passive: the switch must land before
+  // the first paint, otherwise iPad paints the phone layout and re-lays out
+  // everything mid-transition, which leaves animated views with stale frames.
+  useLayoutEffect(() => {
+    acquireNativeTabletRealWidthMedia();
+    return () => {
+      releaseNativeTabletRealWidthMedia();
+    };
+  }, []);
```
sidmorizon: # 🟠 P1 Real-width override is held for onboarding's mount lifetime, not its foreground lifetime

## Problem

Target: iPadOS, `main` (UI) JS runtime. `$$onekeyNativeMedia` is a per-runtime heap copy, so `bg` is not involved; the flag is nonetheless process-global *within* the UI runtime, and every `useMedia()` consumer in the app reads it.

This effect holds `useRealWidth = true` for as long as `OnboardingNavigator` is **mounted**, with no focus tracking. Onboarding is a root-stack route, so other trees are mounted concurrently:

- `ERootRoutes.Main` — the tab tree that on iPad lives in the half-width split-view pane the 767 clamp exists for — stays mounted underneath.
- `ERootRoutes.Modal` can be pushed **on top** of onboarding. The repo documents this at `NavigationContainer.tsx:426-431`: *"V2 can push V1 on top of itself (Connect external wallet reuses V1 ConnectWalletSelectNetworks)"*.

The sibling `setGlassHeaderUIStyle` block immediately below (lines 61-78) exists precisely because a global pinned to onboarding's mount lifetime is wrong, and already subscribes to `useIsFocused()` to relinquish on blur. The new acquire sits five lines above it and reintroduces the same shape without that mechanism.

This is a distinct root cause from the two async-orientation findings already on this PR (stale in-flight callback; non-synchronous release). Those concern *when* the flag takes effect; this concerns *which trees* it applies to while it is held.

## Impact

While onboarding is mounted, unrelated surfaces evaluate `gtMd === true` even though their actual container is far narrower than 768pt:

- the V1 onboarding modal pushed above V2 lays out for a ≥768pt viewport inside a much narrower UIKit sheet;
- `Dialog` inside it takes the centered branch instead of the bottom `Sheet` — the same branch this PR changes in `Dialog/index.tsx`;
- `RootStackNavigator.tsx:66` (`const isVerticalLayout = useMedia().md;`) flips, invalidating the `useMemo` at line 94 and re-creating every root screen's options.

Concrete scenario: iPad landscape, signed-in user hits a mid-session onboarding entry (`useKeylessWallet.tsx:369/394/407/441` all call `navigate(ERootRoutes.Onboarding, …)`). Onboarding slides in, so for the duration of the transition the Main split-view panes are still on screen and visibly reflow from the `md` layout to the `gtMd` layout. Pushing V1 on top then renders it at desktop width inside the sheet.

## Recommendation

Scope the override to onboarding's **foreground** lifetime rather than its mount lifetime. `useIsFocused()` is already computed in this component (line 74) for exactly this reason; hoist it above and gate the effect on it. The existing ref-counting in `nativeTabletRealWidthMedia.native.ts` already tolerates the extra acquire/release churn, and keeping it a layout effect preserves the before-first-paint property:

```tsx
const isFocused = useIsFocused();
useLayoutEffect(() => {
  if (!isFocused) {
    return undefined;
  }
  acquireNativeTabletRealWidthMedia();
  return () => {
    releaseNativeTabletRealWidthMedia();
  };
}, [isFocused]);
```

Note the tradeoff this inherits from the glass-header block: onboarding may still be partially visible behind a pushed modal, so releasing on blur reverts V2 to clamped breakpoints while it is backgrounded. If that is unacceptable, the more correct fix is to stop using a `globalThis` flag and resolve the override from a React context/provider consumed during media resolution, so each tree gets breakpoints matching its own container regardless of what else is mounted.

[Claude session ↗](https://claude.ai/code/session_0194r7tpakXw1M2V2pcaMmJp)

---

patches/@tamagui+web+1.108.0.patch
```diff
@@ -55,6 +67,10 @@ index 47addac..110de63 100644
 +  if (isDualScreenDevice() && ReactNativeDeviceUtils.isSpanning()) {
 +    // DualScreen device in spanning mode: use half screen width for media queries
 +    callback({ shouldOverride: true, overrideWidth: getHalfScreenWidth() });
++  } else if (nativeMediaControl.useRealWidth) {
++    // Full-screen surface opted out of the tablet/Android clamp: keep the raw
++    // matchMedia result computed from the real window width.
++    callback({ shouldOverride: false, overrideWidth: null });
 +  } else if (ExpoDevice.deviceType === ExpoDevice.DeviceType.TABLET) {
 +    // Tablet: check if landscape (spanning layout)
 +    isLandscapeOrientation((isLandscape) => {
```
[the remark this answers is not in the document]
franco-chan: Fixed in 82cb3aa2a0: the tablet branch of `getMediaOverride` no longer uses the async `getOrientationAsync()`; landscape is derived synchronously from `Dimensions.get('window')` (same approach as the 2.7.7 media-driver patch on `x`), so there is no in-flight callback left to overwrite a mode switch.

patches/@tamagui+web+1.108.0.patch
franco-chan: Fixed in 82cb3aa2a0: the orientation probe is now synchronous (window width vs height), so a release/acquire settles `mediaState` in the same `refresh()` call and no stale callback can restore the clamp afterwards.

packages/components/src/hooks/nativeTabletRealWidthMedia.native.ts
```diff
@@ -0,0 +1,39 @@
...
+
+function syncNativeMedia(useRealWidth: boolean) {
+  const control =
+    globalThis.$$onekeyNativeMedia ??
+    (globalThis.$$onekeyNativeMedia = { useRealWidth: false });
+  if (control.useRealWidth === useRealWidth) {
+    return;
+  }
+  control.useRealWidth = useRealWidth;
+  control.refresh?.();
```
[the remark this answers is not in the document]
franco-chan: Fixed in 82cb3aa2a0: with the tablet branch now synchronous, `syncNativeMedia(false)` → `refresh()` writes the clamped `md`/`gtMd` values before the next paint; the unhandled `getOrientationAsync` promise is gone entirely.

packages/kit/src/routes/Modal/Navigator.tsx
franco-chan: Fixed in 82cb3aa2a0: the acquire is now gated on `useIsFocused()` (hoisted above and shared with the glass-header block), so the override is held for onboarding's foreground lifetime only and released when a V1 modal is pushed on top or Main replaces it. Kept as a layout effect to preserve the before-first-paint switch.

originalix (approved):

PR #13283, opened 2026-09-09
fix: prevent dismissed device stage from reappearing (OK-62157)

huhuanming:
[OK-62157](<https://onekeyhq.atlassian.net/browse/OK-62157>)

----------

## Summary

- merge the complete stale Device Stage gesture protection from #13276 into the desktop QR wallet fix
- prevent late pan updates, completion, cancellation, and queued dismissals from restoring or dismissing a hidden or reopened stage
- add 12 focused regression cases for dismissal races, rapid reopening, queued dismissal, and normal dismissal behavior

## Root cause

The pan gesture and the exit animation both write `presence`. Hiding the stage disables the active recognizer, but its late update, end, and finalize callbacks can still run. Previously they could overwrite the exit animation with a value near or equal to `1`, leaving a visible shell whose logical pose was already hidden. Reusing one eligibility value could also re-enable callbacks retained by an old recognizer after a rapid hidden-to-card transition.

## Design

Each enabled drag period owns its own Reanimated shared eligibility value. Layout-effect cleanup retires that value before the exit animation is aimed, and reopening creates a new value instead of re-enabling the old one. Update, end, and finalize callbacks retain their own period's value, and queued `runOnJS` dismissal checks it again before invoking `onDismiss`.

## Reproduction and verification

- deterministic regression: capture the pan gesture created while the card is visible, hide the stage, then deliver late update, end, cancel, and dismissal callbacks, including after reopening
- before the fix, stale callbacks can restore `presence` or dismiss the reopened stage; after the fix they cannot affect it
- `yarn jest --runInBand --watch=false --runTestsByPath packages/components/src/composite/MorphOverlay/index.dismiss.test.tsx` — 12 tests passed
- `yarn agent:check --profile commit` — passed

## Related

- Issue: OK-62157
- Source implementation: #13276 (OK-62095, OK-62163)

[OK-62157]: https://onekeyhq.atlassian.net/browse/OK-62157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9cd35643-9b99-51fe-8031-e83b766452f0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Gd8BsL9vSciyjuuozSTWVT

packages/components/src/composite/MorphOverlay/index.tsx
```diff
@@ -870,6 +872,10 @@ export function MorphOverlay<T>({
   // re-render instead of one frame behind it — and the mount lands its
   // pose targets before the estimates ever paint.
   useLayoutEffect(() => {
+    // Disabling an active recognizer can deliver its final callbacks after
+    // the exit starts. Share the live grant with those old worklets before
+    // aiming presence, so they cannot pull a hidden shell back on screen.
+    dragAllowed.value = dragEnabled;
```
originalix: P2: [A stale drag can dismiss a newly reopened stage]

If a card is hidden while a Pan recognizer is active and the stage becomes `card` again before that recognizer's late terminal callback arrives, this assignment turns `dragAllowed` back to `true`. The old worklet then passes the guard, can write `presence`, and can call the current `onDismiss`, closing the reopened stage.

Track a monotonic drag-lifecycle token: capture it when the recognizer becomes active, invalidate it on hide, and require token equality in every callback. Please add a regression test that retains the original gesture through `card → hidden → card`, then replays its late terminal event and asserts that dismissal is not called.

packages/components/src/composite/MorphOverlay/index.tsx
```diff
@@ -850,6 +850,8 @@ export function MorphOverlay<T>({
     : Math.min(screenWidth - CARD.margin * 2, CARD.maxWidth);
   const cardHeight = CARD.padTop + cardInnerHeight + CARD.bottomPad;
   const dismissible = Boolean(onDismiss);
+  const dragEnabled = dismissible && pose === 'card';
+  const dragAllowed = useSharedValue(dragEnabled);
```
sidmorizon: **Reopened stages accept stale gestures**

When a card reopens before the old gesture finalizes, `dragAllowed` becomes true for every gesture instance. The old callback can move or dismiss the new card. A dismissing callback also invokes the current `onDismiss`, cancelling the new operation.

PR #10404, opened 2026-02-28
fix(password): allow biometric unlock during security cooldown OK-50493

ezailWang:
---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/10404" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #10208, opened 2026-02-12
fix: remove redundant Stack wrapper in DatePickerTrigger causing height collapse(OK-50339)

yikZero:
## Summary

- Remove unnecessary `<Stack position="relative" flex={1}>` wrapper from `DatePickerTrigger`, which caused the Input height to collapse when rendered inside a vertical layout (`YStack`) on mobile
- Remove unused `Stack` import

## Root Cause

`DatePickerTrigger` wrapped `Input` in `<Stack position="relative" flex={1}>` (copied from Select's trigger pattern). Unlike Select, DatePicker doesn't need `position: relative` for absolute-positioned icons since it uses Input's `addOns`. The `flex: 1` in a column layout parent (`YStack width={240}`) with no explicit height caused the wrapper to not properly size from its content, clipping the Input's top border on mobile.

## Affected Pages

All three referral detail pages share the same `DatePickerTrigger` component:
- PerpsReward (合约)
- EarnReward (链上奖励)
- HardwareSalesReward (硬件销售)

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/10208" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #9987, opened 2026-02-03
fix: limit rendered browser tabs to prevent Android ANR on low-end devices

huhuanming:
## Summary
- Only render active tab + 2 recently used tabs (max 3) in the native view tree instead of all open tabs (up to 99)
- Make `collapsable={false}` conditional on active tab only, allowing Android to flatten inactive tab view hierarchies

## Root Cause
All browser tabs were rendered into the native view tree simultaneously via `tabs.map()` in `Browser.native.tsx`. `react-freeze` only prevents React re-renders but does NOT remove native views from the Android draw traversal. Each tab adds 7+ native view layers (Freeze → ViewShot → Stack → WebContent → WebView → RefreshControl → NativeWebView).

On low-end devices (e.g. Moto G31, device.class=1), traversing 30+ nested ViewGroups with WebView GPU rendering caused the main thread to block for >5 seconds during the draw phase, triggering an ANR at `BackgroundStyleApplicator.clipToPaddingBox`.

## Changes
- **Browser.native.tsx**: Track recently active tab IDs with `useRef` and filter `tabs.map()` to only render tabs in the recent set (max 3)
- **MobileBrowserContent.tsx**: Apply `collapsable={false}` only when the tab is active, allowing Android view flattening for non-active tabs

## Trade-offs
- Switching to a tab not in the recent-3 set will reload the WebView from the saved URL (standard mobile browser behavior on resource-constrained devices)
- URL and metadata are always preserved in Jotai state

## Test plan
- [ ] Open 5+ browser tabs on Android, verify no ANR
- [ ] Switch between tabs, confirm active tab renders correctly
- [ ] Verify tab switcher shows thumbnails
- [ ] Test on low-end Android device
- [ ] Verify iOS is not affected (collapsable change is Android-only)

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/9987">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #9614, opened 2026-01-09
feat(perp): add search support, favorites toggle, and improve account validation OK-46430 OK-48555 OK-48628 OK-48660

Minnzen:
### 🔍 Perp Search in UniversalSearch
- Add Perp (perpetual contracts) search type to UniversalSearch
- Support searching across multiple DEX perpetual contracts
- Filter and display Perp items in search results with dedicated section
- Enhance type definitions for Perp search results

### ⭐ Favorites Display Mode Toggle
- Add price/percent display mode toggle ($ | %) in favorites bar
- Support switching between price and 24h change percentage display
- Show token icons alongside favorite items
- Add loading skeleton for price data
- Persist user's display mode preference

### 🔒 Read-Only Account Validation
- Detect and disable trading for read-only/watch-only accounts
- Add "Account Unsupported" UI state for invalid accounts
- Display loading spinner during account selection
- Prevent trading operations on unsupported account types

### 📊 Trading Size Calculation
- Replace `availableToTrade` with `maxTradeSzs` in trading context
- Update trading form and related hooks to use new data structure
- Improve logic for available margin and max position size calculations
- Ensure consistency across trading panel and calculation utilities

PR #13288, opened 2026-09-10
fix: upgrade reanimated to 4.5.5 for stuck device stage(OK-62195)

originalix:
[OK-62195](<https://onekeyhq.atlassian.net/browse/OK-62195>)

----------

## Summary

- Upgrade `react-native-reanimated` from 4.5.1 to 4.5.5 to fix native views freezing at stale animated values.
- Drop the local `topSvgLayout` C++ patch hunk, superseded by the upstream fix in 4.5.4; the remaining eight hunks apply unchanged.
- Regenerate `apps/mobile/ios/Podfile.lock` with CocoaPods 1.17.0 (the version CI activates).

## Intent & Context

OK-62195 was reported as "switching to a single-chain network pops up a Bluetooth interaction dialog". The user reproduced it on an iPhone 17 Pro with a OneKey Pro over BLE and supplied a full log plus two screen recordings, which showed two distinct symptoms:

1. After account creation finishes, the DeviceStage "Processing… / Pro 76D1" capsule never leaves the screen.
2. In one recording the stage is additionally deformed: a card-sized black box occupying the lower half of the screen, with the capsule row drawn at its top-left and a thumbnail-scale device replica still playing the PIN-entry scene.

The question to answer first was whether the DeviceStage burst state machine (`kit-bg`) was leaking, since that is where "one burst = one entrance/exit" is enforced. It is not.

## Root Cause

The two symptoms are one bug, and it is in Reanimated, not in our state machine. The evidence chain:

1. **Background reaches `off`.** Replaying the exact event sequences from the log (the 15:03:26 and 14:53:09 flows) against `DeviceStageBurstScope`, at both the shipped commit and current `x`, lands `off` 250 ms after `end()` and writes it exactly once. The straggler close that follows `sdk.cancel` does not rewrite it.
2. **The main runtime received `off`.** In both recordings the user operates the whole app underneath the stage. The stage is modal, and its blocking wall is a plain React conditional (`blocking && pose !== 'hidden'`). The wall being gone proves React committed `pose === 'hidden'`, i.e. `step === 'off'` crossed the bridge.
3. **Therefore only prop application is left.** The layout effect unconditionally starts `presence -> withSpring(0)` on the hidden pose, so the exit animation ran; the shell simply never moved.
4. **The deformed frame proves it at pixel level.** In a single frame the shell and the replica layer sit at card-pose values while their own children read the same shared values and sit at capsule-pose values. Measured against the build's constants: box width 386pt (`402 - 2 * CARD.margin`), miniature centre x 155.8pt and top 58.7pt, badge centre (43.8, 37.3)pt. That combination is only reachable if two specific native views stopped receiving updates.

The mechanism is Reanimated 4.5.1's `FORCE_REACT_RENDER_FOR_SETTLED_ANIMATIONS` (enabled by default). `PropsRegistryGarbageCollector` polls every 500 ms, copies settled animated props into the component's React state, renders them appended last in the style array, and never invalidates them (the source still carries `TODO(future): revert changes when animated styles are detached`). A React commit landing during a re-animation re-applies that stale snapshot, and the view freezes at its previous pose.

Upstream tracks this as [#9478](https://github.com/software-mansion/react-native-reanimated/issues/9478) (iOS bottom sheet, same mechanism), [#9574](https://github.com/software-mansion/react-native-reanimated/issues/9574) and [#9614](https://github.com/software-mansion/react-native-reanimated/issues/9614).

## Design Decisions

**Upgrade rather than disable the flag.** The documented workaround is `reanimated.staticFeatureFlags.FORCE_REACT_RENDER_FOR_SETTLED_ANIMATIONS: false` in `apps/mobile/package.json`. Both options need a native rebuild, so the upgrade was preferred: it keeps the performance work the flag exists for, and it is the path upstream supports.

**4.5.5 rather than 4.6.0.** The actual fix is [PR #9987](https://github.com/software-mansion/react-native-reanimated/pull/9987) in **4.5.3**, which invalidates the `settledProps` snapshot as soon as a fresh worklet update arrives for an already-synced tag, and fixes view-counter drift for nested animated components — which is exactly what the overlay shell is (`GestureDetector` wrapping an `Animated.View`, a case the registry explicitly special-cases). It also states that no JSI bindings change, so it is OTA-safe in both directions. 4.5.5 is the newest on that line and keeps `react-native 0.83 - 0.86` and `react-native-worklets 0.10.x - 0.11.x`, so **worklets stays at 0.10.1**. 4.6.0 would force worklets to 0.12.x, i.e. swapping a second native library for no additional benefit.

**Local patch shrinks.** Of the nine files the patch touches, eight are byte-identical between 4.5.1 and 4.5.5 and apply unchanged. The ninth, `ReanimatedModuleProxy.cpp`, is where our `topSvgLayout` guard lived; 4.5.4's [#10208](https://github.com/software-mansion/react-native-reanimated/pull/10208) fixes the same crash generically for RN < 0.87 by null-checking `rawEvent.shadowNodeFamily.lock()` before taking the tag, so our hunk is dropped.

**Podfile.lock kept minimal.** A local `pod install` also rewrites the `hermes-engine` spec checksum, which is environment-dependent and unrelated to this change. Since CI runs `pod _1.17.0_ install --deployment`, that line was restored to the committed value so the lockfile diff contains only RNReanimated.

## Changes Detail

- `package.json`, `apps/mobile/package.json` — three version pins (dependencies, resolutions, mobile workspace) moved to 4.5.5.
- `yarn.lock` — reanimated entries only; the worklets peer range widens to `0.10.x - 0.11.x` and the installed worklets version is untouched.
- `patches/react-native-reanimated+4.5.1.patch` renamed to `+4.5.5.patch`, minus the 33-line C++ hunk.
- `apps/mobile/ios/Podfile.lock` — RNReanimated version and spec checksum only.

## Risk Assessment

- **Risk Level**: Medium. A native dependency bump that cannot be validated on a device from this machine.
- **Affected Platforms**: Mobile (iOS/Android) primarily; Web/Desktop/Extension consume the same JS package, and the web-specific patch hunks are unchanged.
- **Risk Areas**: layout animations. 4.5.4's [#10404](https://github.com/software-mansion/react-native-reanimated/pull/10404) stops Reanimated taking over `UIManagerAnimationDelegate`, which changes how `entering`/`exiting` are plumbed. Five files use those props: `composite/MorphOverlay`, `composite/DeviceStage/AppInputs`, and three Onboardingv2 screens.
- Local native validation was not possible: this machine's Xcode 26.0 / Swift 6.2 cannot compile `expo-modules-core`'s `weak let`, and the prebuilt dev-shell no longer matches after a native dependency change.

## Verification already done

- `yarn install` with patch-package: `react-native-reanimated@4.5.5` applied cleanly (8 hunks).
- Full-monorepo `yarn tsc:only`: passes with no diagnostics, covering all 99 reanimated import sites.
- DeviceStage-related Jest suites: 10 suites, 165 tests, all pass.
- `yarn lint:podfile`: passes (CocoaPods 1.17.0).
- `yarn agent:check --profile commit`: all checks pass, no user notices.
- Confirmed in the installed package that both the settled-updates fix (`collectSettledUpdates`) and the `shadowNodeFamily.lock()` crash guard are present.

## Test plan

- [ ] Produce a CI build from this branch (local native build is blocked by the toolchain issue above).
- [ ] Reproduce the OK-62195 path on a real OneKey Pro over BLE: switch to a single-chain network, create an address, and confirm the stage leaves when the flow ends.
- [ ] Repeat with the aggravating sequence from the recordings: close the stage manually, then immediately start a new create-address flow.
- [ ] Confirm the deformed card-sized box no longer appears during the card-to-capsule morph.
- [ ] Regression-check layout animations: MorphOverlay capsule row swap, its close button and corner badge, DeviceStage app inputs, and the three Onboardingv2 screens.
- [ ] Smoke-test SVG-heavy screens (charts, market) for the crash the removed patch hunk used to guard.

[OK-62195]: https://onekeyhq.atlassian.net/browse/OK-62195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


weatherstar (approved):

wabicai (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-cec0ba2e-5053-52ee-87bd-aa38b11c66e7)

sidmorizon (approved):

PR #8221, opened 2025-08-29
feat: sign and verify message

originalix:
## Summary by CodeRabbit

- New Features
  - Sign & Verify Messages: new modal and wallet action for BTC, EVM, and SOL with address auto-detection, hex message input, BTC format choices (BIP137/BIP322), grouped address selector, and one-tap copy of signed results.
- Improvements
  - Radio: horizontal/vertical orientation, adjustable gap, and per-option disabled.
  - Input add-on: supports custom render content.
  - Select: improved trigger layout for icon alignment.
- Chores
  - Updated OneKey HD SDK packages to 1.1.11.
  - Added “electrum” to spell-check skip list.

PR #13289, opened 2026-09-10
fix: use account address as internal sign message signer for external wallets(OK-62523)

weatherstar:
[OK-62523](<https://onekeyhq.atlassian.net/browse/OK-62523>)

----------

## Summary
- Sign & Verify Message now resolves the account address once and uses it as the signer in the `personal_sign` and Solana sign-message payloads, instead of the internal `accountId`.
- Adds a regression test that drives `signInternalMessage` through a mocked keyring and asserts the payload signer slot.

## Intent & Context
QA reported (OK-62523, desktop, test env) that Sign & Verify Message with an external wallet account fails: the external wallet (OneKey iOS via WalletConnect) shows the sign prompt, the user confirms, and the desktop app still shows a "rejected" error. Regular sends from the same external account worked, so the WalletConnect session and address matching were fine and only the message-signing path was broken.

## Root Cause
`ServiceInternalSignAndVerify.signInternalMessage` built the unsigned message with `payload: [message, accountId]`. HD/hardware keyrings ignore `payload`, so this never mattered for them. External keyrings, however, hand the payload straight to `evmConnectorUtils.parseSignMessageParams`, which forwards it verbatim as the RPC params. The peer wallet therefore received `personal_sign(["0x…", "external--wc--<topic>--evm"])`.

On a OneKey peer, `ProviderApiEthereum.personal_sign` takes `messages[1]` as the address without validation and opens the confirm sheet. When the user confirms, `validateSignMessageData` rejects the non-hex `from` address, `dappApprove.reject` fires, and `ProviderApiWalletConnect.handleSessionRequest` wraps every error as `USER_REJECTED` (5000 "User rejected."). The desktop side surfaces that message via `@toastIfError`, which is exactly the "confirmed but rejected" symptom in the report. Peers such as MetaMask reject earlier with invalid params. The desktop log confirms the WalletConnect peer ("OneKey iOS"), a successful send from the same account minutes before, and two sign attempts each ending ~12 s after the confirm click.

The bug has existed since the feature landed in #8221; the dApp path (`ProviderApiEthereum.personal_sign`) always used `[message, address]`.

## Design Decisions
- Resolve the address in the background service via `serviceAccount.getAccount({ accountId, networkId })` rather than trusting an address sent from the UI, so the signer slot is always derived from the account the user selected.
- Apply the same change to the Solana branch even though external Solana signing is not wired up yet, so the payload contract is consistent.
- Reuse the single account lookup in the BTC BIP-137 normalization branch, which previously fetched the account a second time.
- No change to keyrings or connector parsing: the payload contract (`[message, address]`) matches the dApp path, so fixing the producer is the minimal and correct fix.

## Changes Detail
- `packages/kit-bg/src/services/ServiceInternalSignAndVerify.ts`: fetch the account once before building the unsigned message; use `account.address` in the EVM and Solana payloads; drop the duplicate account fetch in the BTC branch.
- `packages/kit-bg/src/services/ServiceInternalSignAndVerify.test.ts`: new test that mocks the background decorators and `vaultFactory`, calls `signInternalMessage` for an external EVM account and a Solana account, and asserts the payload carries the address (it failed with the account id before the fix).

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Web / Extension / Mobile (all platforms that expose Sign & Verify Message for external accounts)
- **Risk Areas**: The extra `getAccount` call runs for every internal sign; it is the same lookup the BTC branch already performed and is cheap. HD/hardware/imported flows are unaffected because their keyrings do not read `payload`.

## Test plan
- [x] `yarn jest packages/kit-bg/src/services/ServiceInternalSignAndVerify.test.ts` fails before the fix (payload signer = account id) and passes after
- [x] `yarn agent:check --profile commit` passes
- [ ] Desktop connected to OneKey iOS via WalletConnect: Wallet > More > Sign & Verify Message > Sign; confirm on the phone; desktop shows the signature instead of "rejected"
- [ ] Verify tab: paste the produced signature and address; verification succeeds
- [ ] Regression: Sign & Verify Message with an HD account (ETH / BTC bip137 / bip322 / SOL) still produces a verifiable signature

[OK-62523]: https://onekeyhq.atlassian.net/browse/OK-62523?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-747870b0-4eba-523c-85b6-27b673bfc553)

originalix (approved):

PR #13291, opened 2026-09-10
fix: release desktop shortcuts when inactive or quitting

wabicai:
OneKey Desktop can retain system-wide shortcuts such as Cmd+W, Cmd+T, and Cmd+R after losing focus, hiding, or starting to quit, affecting keyboard shortcuts in other macOS apps.

Register shortcuts synchronously only while the main window is visible, focused, and not quitting. Release them before renderer notifications on blur, and on hide, close, and the first before-quit event, including while asynchronous BLE cleanup is pending. This closes the deferred-registration and missing-release gaps without changing the BLE shutdown sequence.

Reported in [Slack](https://onekeygroup.slack.com/archives/C01SZCKLJ0P/p1788955179706199).

Validation:

- 10 shortcut lifecycle regression tests pass. The same tests against the original code produce 8 failures and 2 passes.
- Both existing real Electron macOS exit regressions pass: bundle restart waits for BLE cleanup, and native quit completes when cleanup resolves immediately.
- Lint, formatting, and the other local agent checks pass.
- Full TypeScript validation remains blocked by missing local dependencies (including `@onekeyfe/react-native-native-list` and `@onekeyfe/react-native-text`) and type errors in unchanged modules. No TypeScript errors were reported in the modified files.

Draft: verification in a complete macOS Desktop build is still pending, particularly for the reported intermittent shortcut capture during normal focus switching. The tests validate the identified lifecycle gaps, rather than proving the exact event sequence behind every report.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-fdee038e-c47d-57fc-8bd8-d2fda2467660)

sidmorizon:
Claude review session: https://claude.ai/code/session_01HU6g36WJsRSBG8nyfhKx4g

apps/desktop/scripts/shortcuts-lifecycle.test.js
```diff
@@ -0,0 +1,207 @@
...
+        event === 'before-quit' &&
+        !hasQuitHandler
+      ) {
+        hasQuitHandler = true;
+        evaluate(node.getText(appSource));
+      }
+    }
+    ts.forEachChild(node, visit);
+  };
+  visit(appSource);
```
sidmorizon, line 133: # 🟡 P2 Handler discovery can silently find nothing, leaving tests green

## Problem

`loadShortcuts()` discovers the handlers under test by scanning the `app.ts` AST for `browserWindow.on('<event>', …)` / `app.on('before-quit', …)` (lines 102-119) and `evaluate()`-ing the matched source text. Nothing asserts that the discovery actually matched anything.

Because `registered` starts as an empty `Map` (line 21) and `emit()` on an `EventEmitter` with no listener is a silent no-op, every test whose expectation is `expect(registered.size).toBe(0)` passes just as well when the handler was never installed as when it was installed and behaved correctly.

Concretely, if the `enter-full-screen` handler in `app.ts:1235` is removed, renamed, or moved into a helper (so `target === 'browserWindow.on'` at line 106 no longer matches), the whole suite still passes:

- line 156 `test.each(['focus', 'enter-full-screen'])` — the `enter-full-screen` case asserts `0` from an empty registry;
- line 186 `'quit releases shortcuts …'` — its `browserWindow.emit('enter-full-screen')` at line 195 becomes a no-op and the trailing `expect(registered.size).toBe(0)` still holds.

No other test in the file exercises `enter-full-screen`, so nothing fails. (A rename of *all* window handlers would be caught by line 128's `toBe(3)`, so the exposure is specifically a per-handler change.)

Related, secondary: the vm context (lines 58-79) does not define `store`, `getSafelyMainWindow`, `destroyTrayManager`, `systemIdleInterval`, `disposeContextMenu`, `clearTimeout`, or `setImmediate`, all of which the extracted `before-quit` handler references (`app.ts:1944-1969`). This only works today because the fixture pins `isMac: true` + `bleQuitReady: false` and `nobleBleInitialization` never settles, so control always returns at `app.ts:1945` before reaching them. An edit that hoists work above that `return` would surface as a `ReferenceError` rather than a meaningful assertion failure.

## Impact

The suite is this PR's stated evidence of correctness ("10 shortcut lifecycle regression tests pass"), and it is a regression lock on global-hotkey release. In its current form it can quietly stop locking the `enter-full-screen` path while still reporting green, which is the failure mode a regression test exists to prevent.

## Recommendation

Make discovery failure loud. In `visit()`, collect matched event names, then assert completeness before returning — e.g. accumulate into `const found = new Set()` on the `browserWindow.on` branch (line 106) and, after `visit(appSource)` on line 119, assert `found` equals `windowEvents` and that `hasQuitHandler === true`. A handler rename in `app.ts` then fails the suite explicitly instead of silently emptying it.

Additionally, stub the remaining `before-quit` identifiers listed above in the context at lines 58-79 so a future change to that handler fails on an assertion rather than a `ReferenceError`.

[Claude session ↗](https://claude.ai/code/session_01HU6g36WJsRSBG8nyfhKx4g)

---

wabicai: Fixed in b07bf9f598.

`loadShortcuts()` now asserts that every `windowEvents` handler and the first `before-quit` handler were actually found before returning. A rename/move of `enter-full-screen` (or any other scanned handler) fails the suite instead of leaving `registered.size === 0` tests green.

Also stubbed the remaining `before-quit` identifiers (`store`, `getSafelyMainWindow`, `destroyTrayManager`, `systemIdleInterval`, `disposeContextMenu`, `clearTimeout`, `clearInterval`, `setImmediate`, `app.quit`) so a later edit in that handler fails on an assertion rather than a `ReferenceError`. Added a positive `enter-full-screen` registration case as well.

originalix (approved):

PR #13293, opened 2026-09-10
fix: honor AnimatePresence initial false

huhuanming:
## Summary

- preserve Tamagui's existing enterStyle selectors and normal animation behavior
- latch Web initial={false} for the initial child instance because Presence only exposes it on the first render
- skip only that instance's enterStyle atomic class, including variant-expanded enterStyle, so an entering Popover ancestor cannot activate it
- document the account-menu root cause directly in the dependency patch

## Testing

- generated the patch with patch-package and verified git apply --check against pristine @tamagui/web@2.7.7
- built Web with yarn app:web:no-proxy
- verified first open and close/reopen have no enterStyle class and remain at transform: none
- verified forward and back account-panel navigation still produce non-zero translateX transitions
- yarn agent:check --profile commit passed
- yarn agent:check --profile pr passed all local checks; remote status requires review


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-eb1b8a4c-fe36-5134-867c-c02d625b01b0)

originalix (approved):

PR #11937, opened 2026-06-06
fix: refresh stale wallet token list & DeFi overview on off-tab network switch (OK-55826, OK-55705)

weatherstar:
## Summary
- Fix the always-visible header worth (wallet token list **and** DeFi overview) getting stuck on the previous network when the user switches network while on a non-wallet home tab (History / DeFi / NFT).
- Align the new off-tab imperative refresh with the closure-bound `run` path: forward a fresh `indexedAccountId`, abort superseded in-flight fetches, sync the local token directory, and cover single-address Others accounts.
- Null-safety crash fix in the send → swap token conversion when the token has no price.
- Dedupe the triplicated DeFi overview currency-conversion math and reuse the shared event-payload type.

## Intent & Context
QA filed **OK-55826** and **OK-55705**. Two problems:
1. When the user switches network while NOT on the wallet/Portfolio tab, the inactive home tab is frozen, so the token-list and DeFi-overview blocks keep showing the previous network's worth in the always-visible header until the user navigates back to the wallet tab.
2. Opening the send → swap amount conversion for a token that has no price crashed.

## Root Cause
- **Frozen tab → stale closures:** the inactive home tab is suspended via `FreezeInactiveHomeTab` (react-freeze), so the component function never re-runs while frozen. `usePromiseResult`'s "deps changed on blur" recovery is never armed, and the closure-bound `run` keeps pointing at the previous network. The always-visible header gates the single-network worth on both token and DeFi readiness, so it falls back to a stale/zero placeholder.
- **Send → swap crash:** `convertTokenToSwapToken` did `tokenDetails.price.toString()`, assuming `price` is always present.

## Design Decisions
- **Drive the refresh from the always-mounted `HomePageView`:** on an off-tab network change it emits `RefreshTokenList` with `refreshByProvidedAccounts` and an explicit `{ accountId, networkId, indexedAccountId }`. The frozen blocks resolve the request from the **payload** (fresh values) instead of their stale closures.
- **Out-of-order / race safety:** a per-block `seq` ref + `isLatest()` guard drops superseded responses; the explicit path now also calls `abortFetchAccountTokens` before fetching, mirroring `run`.
- **Parity with `run`:** the explicit token path forwards `indexedAccountId` (so aggregate hidden/custom tokens resolve correctly) and calls `updateLocalTokens` to keep the broader local token directory in sync.
[...]

PR #13300, opened 2026-09-10
fix: refresh wallet token list on off-tab account switch(OK-62678)

weatherstar:
[OK-62678](<https://onekeyhq.atlassian.net/browse/OK-62678>)

----------

## Summary
- Switching account while a non-spot home tab (History / NFT / DeFi / Perps) is active left the always-visible header total on a skeleton until the user returned to the spot tab.
- The wallet token list now refreshes its freshly mounted owner itself when it mounts (or re-renders a new owner) while another home tab is active.
- Adds a small pure resolver (`offTabRefresh.ts`) with unit tests.

## Intent & Context
Jira OK-62678 (reported today, priority Highest): "在非现货 tab 下切换账户，总资产一直 loading 不加载，切回去就能恢复加载". Reproduced on desktop dev via CDP: on the History/NFT tab, switching to an account whose worth is not cached left the header skeleton for 10 s+ with zero `serviceToken.fetchAccountTokens` calls; pressing the spot tab fired the fetch ~1.2 s later and the header rendered. Accounts that already have a confirmed/cached balance show that value instead of a skeleton, which is why QA only sees it on accounts without cache.

## Root Cause
- An account switch changes the `Tabs.Container` key (wallet + indexedAccountId), so the whole tab pane tree remounts and `TokenListBlock` mounts inside a non-focused `Tabs.Tab`.
- Its token fetch is a `usePromiseResult` with `overrideIsFocused: isPageFocused && isFocused` (inner-tab focus). The runner sees `triggerByDeps && !isFocused`, flags `isDepsChangedOnBlur` and skips; the run only happens when the spot tab regains focus. `HomeOverviewContainer` has no fetch of its own and only reads `accountWorthAtom`, so nothing writes the new owner's worth.
- The off-tab *network* switch was already special-cased in `HomePageView` (#11937, OK-55826) via `RefreshTokenList` + `refreshByProvidedAccounts`, but account switches were not.
- Not a regression from #13281 (`FreezeInactiveHomeTab` timing): `DelayedFreeze` still commits the first render, and the same focus gate / remount key exist on `release/v6.5.2`.

## Design Decisions
- First attempt extended the `HomePageView` network-switch effect to account changes. It failed live: on an account switch the old listeners unsubscribe at +250 ms, the emit lands at +256 ms, and the remounted `TokenListBlock` / `DeFiListBlock` subscribe only at ~+430 ms, so the event is lost. A fire-and-forget bus emit cannot reach a component that is being remounted in the same interaction.
- Chosen approach: `TokenListBlock` reads `activeTabId` from `HomeStickyHeaderContext` and, in an effect keyed on `account.id` / `network.id`, calls the existing `refreshSingleNetworkTokenListByTarget` when another tab is active. `DelayedFreeze` allows one committed render before freezing, so the mount effect runs; the result is written through the context action into `accountWorthAtom`, which the header reads regardless of tab focus. The DeFi cache-only hook already bypasses the focus gate (`shouldAlwaysFetch`).
- All Networks is excluded on purpose: the fan-out hook is gated on route focus only and already runs off-tab.
- The effect is owner-keyed, not tab-keyed, so tab switches never refetch. The on-tab path is untouched (`activeTabId === Portfolio` → no-op). `refreshSingleNetworkTokenListByTarget`'s sequence guard dedups any overlap with the network-switch event path.

## Changes Detail
- `packages/kit/src/views/Home/components/TokenListBlock/offTabRefresh.ts`: `resolveOffTabTokenListRefreshOnMount` — returns the single-network refresh target when the active home tab is not the spot tab, the owner is complete, and the network is not All Networks.
- `packages/kit/src/views/Home/components/TokenListBlock/offTabRefresh.test.ts`: 6 unit tests (off-tab owner, Others account without indexed account, spot tab, unknown tab, All Networks, incomplete owner).
- `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx`: owner-keyed effect that runs the resolver against `HomeStickyHeaderContext.activeTabId` and calls `refreshSingleNetworkTokenListByTarget`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Web / Extension / Mobile (same home tab container on all)
- **Risk Areas**: one extra single-network token fetch when the list mounts off-tab; merge-derive networks (BTC etc. on HD accounts) still bail inside `refreshSingleNetworkTokenListByTarget`, the same limitation as the existing network-switch path.

## Test plan
- [x] `yarn jest packages/kit/src/views/Home/components/TokenListBlock/offTabRefresh.test.ts` (6/6) and Home pages/TokenListBlock suites (18 suites, 157 tests)
- [x] Desktop dev (CDP): History tab → switch to an account without cached worth → `fetchAccountTokens` fires at ~0.4 s, header settles by ~0.9 s (before: skeleton 10 s+, no fetch)
- [x] Desktop dev: off-tab network switch still refreshes via the existing event path (single fetch); on-tab account switch unchanged (single fetch)
- [ ] Device QA: iOS/Android — on History/NFT/DeFi tab switch to a newly created account, header total loads without returning to 现货

[OK-62678]: https://onekeyhq.atlassian.net/browse/OK-62678?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
+++ b/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
@@ -1,5 +1,6 @@
 import {
   useCallback,
+  useContext,
   useEffect,
   useLayoutEffect,
   useMemo,
@@ -127,6 +128,7 @@ import type {
   ITokenFiat,
 } from '@onekeyhq/shared/types/token';
 
+import { HomeStickyHeaderContext } from '../HomeStickyHeaderContext';
 import { RichBlock } from '../RichBlock/RichBlock';
 
 import {
@@ -144,6 +146,7 @@ import {
   shouldReportWalletAssetStatusSnapshot,
 } from './assetStatusAnalytics';
 import { buildHomeTokenListCacheIngestRound } from './buildHomeTokenListCacheIngestRound';
+import { resolveOffTabTokenListRefreshOnMount } from './offTabRefresh';
 import { PortfolioSyncButton } from './PortfolioSyncButton';
 import {
   countFundedHardwarePortfolioTokens,
@@ -3245,6 +3248,29 @@ function TokenListBlock({
     refreshSingleNetworkTokenListByTarget,
     showLpTokensOnly,
   ]);
+
+  // The fetch above is gated on this tab being focused, and an account switch
+  // remounts the whole home tab container. When that happens while another
+  // home tab is active, nothing fetches the new owner until the user returns,
+  // leaving the always-visible header worth on a skeleton. The off-tab
+  // network-switch path (RefreshTokenList with refreshByProvidedAccounts,
+  // emitted by HomePageView) cannot cover this: it fires before the remounted
+  // list has subscribed. Refresh the mounted owner explicitly instead.
+  const activeHomeTabId = useContext(HomeStickyHeaderContext)?.activeTabId;
+  useEffect(() => {
+    const target = resolveOffTabTokenListRefreshOnMount({
+      accountId: account?.id,
+      networkId: network?.id,
+      indexedAccountId: indexedAccount?.id,
+      activeTabId: activeHomeTabId,
+    });
+    if (target) {
+      void refreshSingleNetworkTokenListByTarget(target);
+    }
+    // Owner-keyed on purpose: re-running on tab changes would refetch on
+    // every tab switch.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [account?.id, network?.id]);
 
   useEffect(() => {
     if (isEmptyAccount) {
--- a/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
+++ b/packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
@@ -86,6 +86,7 @@ import {
   POLLING_INTERVAL_FOR_TOKEN,
 } from '@onekeyhq/shared/src/consts/walletConsts';
 import errorToastUtils from '@onekeyhq/shared/src/errors/utils/errorToastUtils';
+import { isRequestCanceledError } from '@onekeyhq/shared/src/errors/utils/errorUtils';
 import {
   EAppEventBusNames,
   type IAppEventBusPayload,
@@ -3095,22 +3096,19 @@ function TokenListBlock({
 
       let emittedRefreshing = false;
       try {
-        // Multi-derive (merge-derive) HD accounts need the parallel
-        // per-derivation fetch that only `run` performs; the single-account
-        // fast path below would apply partial data, so skip them. Others
-        // accounts (imported/watch-only) are always single-address even on
-        // merge-derive networks, so they can safely use the fast path here.
+        // Multi-derive (merge-derive) HD accounts aggregate every derivation
+        // of the indexed account, mirroring the parallel fetch in `run`.
+        // Others accounts (imported/watch-only) are always single-address
+        // even on merge-derive networks, so they take the single fetch.
         const targetVaultSettings =
           await backgroundApiProxy.serviceNetwork.getVaultSettings({
             networkId,
           });
-        if (
-          targetVaultSettings?.mergeDeriveAssetsEnabled &&
-          !accountUtils.isOthersAccount({ accountId })
-        ) {
-          return;
-        }
         if (!isLatest()) return;
+        const mergeDeriveTarget =
+          !!targetVaultSettings?.mergeDeriveAssetsEnabled &&
+          !accountUtils.isOthersAccount({ accountId }) &&
+          !!indexedAccountId;
 
         appEventBus.emit(EAppEventBusNames.TabListStateUpdate, {
           isRefreshing: true,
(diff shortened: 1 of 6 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-cefc0789-4ac8-5469-8991-f29674ecbad8)

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -3246,6 +3249,29 @@ function TokenListBlock({
...
+  const activeHomeTabId = useContext(HomeStickyHeaderContext)?.activeTabId;
+  useEffect(() => {
+    const target = resolveOffTabTokenListRefreshOnMount({
+      accountId: account?.id,
+      networkId: network?.id,
+      indexedAccountId: indexedAccount?.id,
+      activeTabId: activeHomeTabId,
+    });
+    if (target) {
+      void refreshSingleNetworkTokenListByTarget(target);
```
sidmorizon, line 3307: **Merge-derived account refresh remains blocked**

Switching an HD account off-tab on a merge-derived network makes `refreshSingleNetworkTokenListByTarget` exit before fetching. The header remains loading until Portfolio gains focus.

sidmorizon, line 3307: **Remounts defeat stale refresh protection**

Rapid off-tab account switches start `refreshSingleNetworkTokenListByTarget` with separate sequence refs. An older remounted instance can abort the newest fetch, leaving the current header loading.

originalix:

Review of commit `1ddfd87f669a` did not finish successfully.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
weatherstar: Fixed in ef5f7aabb7: `refreshSingleNetworkTokenListByTarget` no longer bails on merge-derive HD accounts. It now resolves the derivations via `getNetworkAccountsInSameIndexedAccountIdWithDeriveTypes`, fetches them in parallel and writes the per-derivation worth map keyed by the indexed account, mirroring the merge-derive branch of `run`. Others accounts keep the single fetch.

weatherstar: Fixed in ef5f7aabb7: the owner-keyed mount effect now returns a cleanup that bumps `explicitRefreshSeqRef`. Because an account switch remounts the block, the old instance is unmounted before the new owner fetches, so every later stage of its in-flight refresh (`abortFetchAccountTokens`, worth write) fails the `isLatest()` check and cannot abort or overwrite the successor. An extra `isLatest()` check was also added after the derive-account lookup.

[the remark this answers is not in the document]
weatherstar: Fixed in ef5f7aabb7: added `isRequestCanceledError` (shared `errorUtils`) that recognizes cancels by the serialized `name`/`code` (`CanceledError` / `ERR_CANCELED`) as well as the live axios flag, with tests for the RPC-rebuilt form. The explicit refresh now uses it and no longer rethrows: non-cancel errors are logged via `console.error`, so the fire-and-forget callers can never produce an unhandled rejection.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -3246,6 +3249,29 @@ function TokenListBlock({
...
+    }
+    // Owner-keyed on purpose: re-running on tab changes would refetch on
+    // every tab switch.
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [account?.id, network?.id]);
```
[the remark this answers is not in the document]
weatherstar: Fixed in ef5f7aabb7: the mount effect cleanup invalidates the instance sequence on owner change / unmount, so a late `getVaultSettings` from the previous owner instance fails `isLatest()` before it reaches `abortFetchAccountTokens` or the `merge: false` `accountWorthAtom` write. The stale instance therefore cannot cancel or replace the current owner's header worth.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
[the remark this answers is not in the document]
weatherstar: Fixed in ef5f7aabb7: the explicit header refresh no longer calls `updateTokenListState({ initialized: true })`. Since it does not ingest rows into the owner cells, the list stays uninitialized (skeleton) until the focused `run` fills it, so `EmptyToken` is not shown on return.

originalix:

Review of commit `0cc177684675` did not finish successfully.

originalix (approved):

PR #13304, opened 2026-09-10
fix: prevent iOS image crash and simulator startup failure ｜OK-62654

zy2000-star:
[OK-62654](<https://onekeyhq.atlassian.net/browse/OK-62654>)

----------

## Summary

Fixes the iOS image crash reported in OK-62654 when searching for and selecting a stock token from a stock detail page. Also fixes a simulator startup check that incorrectly reported the app as terminated on newer iOS simulator runtimes.

### Stock detail image crash

The reported flow opens the AAPL detail page, searches for AAPLX, and selects the result. During the transition, the image source can become empty.

The native image wrapper forwards the missing URI as `undefined`. When React Native Fabric clears the existing native property, it sends `null`. Nitro’s optional string converter does not accept that value, causing:

`Exception in HostFunction: OneKeyImage.sourceUri: Value is null, expected a String`

The exception occurs while updating the native image props, before the normal image error handler or fallback can handle it.

### Simulator startup detection

The development launcher checked whether the iOS app survived startup by executing `/bin/kill -0 <pid>` inside the simulator through `simctl spawn`.

On the installed simulator runtime, that command failed even though the app process was still running. The launcher interpreted the command failure as an app crash and shut down Metro, leaving the app unable to load its JavaScript bundle.

### Handle empty image sources on iOS

- Render a regular `View` with the existing fallback overlay when the active image URI is missing, empty, or whitespace-only.
- Unmount the native image instead of clearing its source property, preventing the invalid `null` update.
- Preserve the image container’s layout and view props.
- Mount the native image again when a valid source becomes available.
- Keep Android’s existing empty-source handling unchanged.

### Check simulator processes on the host

- Run `/bin/kill -0 <pid>` on macOS, where simulator app processes reside.
- Preserve the existing startup grace period and process-exit detection.
- Update the startup tests to verify the host-side process check.

## Validation

- **67 tests passed** across the image component and native development launcher suites.
- Image coverage includes:
  - Missing, empty, and whitespace-only source URIs.
  - Transition from a valid image to a fallback and back to a new image.
  - Fallback container dimensions.
  - Unchanged Android empty-source behavior.
  - Existing resize hints and custom request identity handling.
- Launcher coverage includes successful startup and detection of a process that exits during the startup grace period.
- `yarn agent:check --profile commit` passed, including lint, formatting, and TypeScript checks.
- The app successfully launched on an **iPhone 16 Pro simulator running iOS 26.5**, and a market detail page rendered.

## Remaining verification

The complete AAPL → search AAPLX → select result flow still requires manual regression testing. Successful startup and detail-page rendering do not establish that the full reported flow has passed.

RevenueCat anonymous-user logout errors and a separate debug error notification were observed during the simulator session; these are not resolved by this change.

## Related issue

[OK-62654](https://onekeyhq.atlassian.net/browse/OK-62654)

[OK-62654]: https://onekeyhq.atlassian.net/browse/OK-62654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/mobile/scripts/__tests__/native-dev-shell.test.js
+++ b/apps/mobile/scripts/__tests__/native-dev-shell.test.js
@@ -593,14 +593,7 @@ describe('native-dev-shell', () => {
         wait,
       }),
     ).resolves.toBeUndefined();
-    expect(iosOutput).toHaveBeenCalledWith('xcrun', [
-      'simctl',
-      'spawn',
-      'SIMULATOR-A',
-      '/bin/kill',
-      '-0',
-      '4321',
-    ]);
+    expect(iosOutput).toHaveBeenCalledWith('/bin/kill', ['-0', '4321']);
     expect(wait).toHaveBeenCalledTimes(15);
     expect(wait).toHaveBeenCalledWith(1000);
   });
--- a/apps/mobile/scripts/native-dev-shell.js
+++ b/apps/mobile/scripts/native-dev-shell.js
@@ -2209,14 +2209,8 @@ async function waitForNativeAppStartup({
         elapsed += 1000
       ) {
         await wait(1000);
-        runForOutputCommand('xcrun', [
-          'simctl',
-          'spawn',
-          deviceId,
-          '/bin/kill',
-          '-0',
-          String(launch.processId),
-        ]);
+        // Simulator app PIDs belong to the host; newer runtimes omit /bin/kill.
+        runForOutputCommand('/bin/kill', ['-0', String(launch.processId)]);
       }
     }
   } catch (error) {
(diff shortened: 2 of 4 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-fd5a0860-9785-5982-8745-766ec248be7e)

originalix:

Review of commit `a4fe158d468b` did not finish successfully.

originalix (approved):

PR #13316, opened 2026-09-10
fix: Borrow and trading flows (OK-62680, OK-62699, OK-62728, OK-62730, OK-62738, OK-62739, OK-62747, OK-62754)

zhaono1:
[OK-62680](<https://onekeyhq.atlassian.net/browse/OK-62680>) [OK-62699](<https://onekeyhq.atlassian.net/browse/OK-62699>) [OK-62728](<https://onekeyhq.atlassian.net/browse/OK-62728>) [OK-62730](<https://onekeyhq.atlassian.net/browse/OK-62730>) [OK-62738](<https://onekeyhq.atlassian.net/browse/OK-62738>) [OK-62739](<https://onekeyhq.atlassian.net/browse/OK-62739>) [OK-62747](<https://onekeyhq.atlassian.net/browse/OK-62747>) [OK-62754](<https://onekeyhq.atlassian.net/browse/OK-62754>)

----------

## Summary

Fix eight Borrow, stock-detail, and trading-position regressions. Each issue has a separate commit; the two Borrow selector commits share the same component and are ordered so selection persistence lands before async selection handling.

| Issue | Change |
| --- | --- |
| [OK-62680](https://onekeyhq.atlassian.net/browse/OK-62680) | Preserve stock identity on global-search re-entry and synchronize manual token-variant selection with the route. |
| [OK-62699](https://onekeyhq.atlassian.net/browse/OK-62699) | Keep the background stock action stable and inert while order review is opening/open, including the fallback branch. |
| [OK-62728](https://onekeyhq.atlassian.net/browse/OK-62728) | Close the Borrow market selector before data refresh and honor the latest selection through retained callbacks and pending requests. |
| [OK-62730](https://onekeyhq.atlassian.net/browse/OK-62730) | Prefetch Swap/Stocks/Pro positions independently according to configuration readiness while retaining activation freshness checks. |
| [OK-62738](https://onekeyhq.atlassian.net/browse/OK-62738) | Reapply closed Popover styles when adaptive layout replaces its content node. |
| [OK-62739](https://onekeyhq.atlassian.net/browse/OK-62739) | Remember the chosen Borrow market before async settlement so navigation does not lose selection intent. |
| [OK-62747](https://onekeyhq.atlassian.net/browse/OK-62747) | Filter supply assets against authoritative reserve eligibility; show retry on validation failure instead of admitting unverified assets. |
| [OK-62754](https://onekeyhq.atlassian.net/browse/OK-62754) | Use the existing press-based Switch for this iOS collateral control, preserving business guards, disabled state, and accessibility semantics. |

## Issues

Fixes OK-62680
Fixes OK-62699
Fixes OK-62728
Fixes OK-62730
Fixes OK-62738
Fixes OK-62739
Fixes OK-62747
Fixes OK-62754

### After merging x at ea59373575

- Resolved the navigation helper and test conflicts while retaining stock identity preparation and incoming unresolved-asset preview/routing behavior.
- Original regression set: 9 suites / 158 tests passed; incoming selector/layout/asset-route regressions: 3 suites / 15 tests passed (173 total).
- Commit profile and all local PR-profile checks passed, including TypeScript. One independent read-only review found no new regression in the merged logic or adjacent callers.
- Confirmed all 24 original PR paths not touched by incoming x are unchanged from the previously validated head; original eight issue commits remain intact, followed by one merge commit.
- Post-merge Desktop/iOS navigation runtime acceptance is incomplete: the Desktop renderer stopped responding to CDP page commands during the check, and iOS UI automation was slow/unreliable. Neither observation is classified as a merge-caused product defect without causal evidence. The earlier runtime results below are pre-merge evidence, not proof for the integrated build.
- CI must be checked against the new merge commit after push.

### Before merging x

- Targeted regression suite: 9 suites / 151 tests passed.
- Commit profile run before each issue commit; PR profile run before publication.
- Desktop: actual global-search stock re-entry, manual token variant switching, live quote review/cancel/reopen, responsive Popover open/closed behavior, and Borrow asset-detail back navigation.
- Android: live first market selection; controlled A→B→A and A→B→C with a delayed real Arbitrum response; positions across Swap/Stocks/Pro and independent configuration failure.
- iOS: live expired-supply-asset exclusion and repeated collateral-toggle preview/cancel interactions.
- API fixtures follow local backend contracts, are explicitly controlled evidence, and were removed after validation. A cold restart confirmed no mock positions remained.
- No transaction was signed or submitted.

## Scope and limitations

- Open for review at the author's request; this status is not a production or merge-readiness claim. The post-merge runtime acceptance gap remains documented above.
- Merged base `x` at `ea59373575` in merge commit `3ceedb693c`; the original eight issue commits were not rewritten. The two content conflicts are resolved. Complete combined-navigation runtime acceptance before merging.
- No generated translation changes, backend changes, wallet-data changes, or temporary mocks/logging are included.
- Production builds, full VoiceOver device testing, hardware signing, and remote CI completion are not claimed.
- Local environment interruptions (Android emulator low-memory termination and development-session initialization issues) are not treated as product-fix evidence. Successful repeat runs are the acceptance evidence; this PR does not claim to fix those environment problems.

[OK-62680]: https://onekeyhq.atlassian.net/browse/OK-62680?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62699]: https://onekeyhq.atlassian.net/browse/OK-62699?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62728]: https://onekeyhq.atlassian.net/browse/OK-62728?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62730]: https://onekeyhq.atlassian.net/browse/OK-62730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62738]: https://onekeyhq.atlassian.net/browse/OK-62738?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-62739]: https://onekeyhq.atlassian.net/browse/OK-62739?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62747]: https://onekeyhq.atlassian.net/browse/OK-62747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62754]: https://onekeyhq.atlassian.net/browse/OK-62754?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-d9cc5c32-7469-5ee8-8e72-9a01b058ed60)

packages/kit/src/states/jotai/contexts/marketV2/actions.ts
```diff
@@ -179,6 +179,46 @@ class ContextJotaiActionsMarketV2 extends ContextJotaiActionsBase {
...
 
+  prepareStockTokenDetail = contextAtomMethod(
+    (
+      get,
+      set,
+      target: { tokenAddress: string; networkId: string; isNative?: boolean },
+    ) => {
+      // A stock listing without a token leaves selection to the stock provider.
+      if (!target.networkId || (!target.tokenAddress && !target.isNative)) {
+        return;
```
sidmorizon, line 198: **Tokenless stocks retain stale details**

For a tokenless stock, `prepareStockTokenDetail` preserves the previous detail and request. A failed variant lookup then leaves another stock’s token data visible.

sidmorizon:
Claude review session: https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN

packages/kit/src/views/Swap/utils/swapPositionPrefetchUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+      networkList: stockNetworkList,
+      stockOnly: true,
+    },
+    {
+      key: 'pro',
+      networkList: proNetworkList,
+      stockOnly: false,
+    },
+  ];
+  return scopes.filter(
```
sidmorizon: # 🟡 P2 Active-tab prefetch priority dropped while the loader still treats scope[0] as primary

## Problem

The previous implementation sorted the three scopes so the **active** one came first (`SwapMainLand.tsx` at the merge base: `activeKey = 'pro'` when `focusSwapPro`, `'stock'` when `swapTypeSwitch === ESwapTabSwitchType.STOCK`, then `scopes.toSorted(...)`). The new helper returns a fixed `swap → stock → pro` order and drops the sort entirely.

The consumer, however, still treats index 0 as the priority scope. `SwapMainLand.tsx:1343` destructures `const [primaryScope, ...additionalScopes] = positionPrefetchScopes`, and `swapProPositionsLoader.ts:240-246` sorts the per-network fetch queue by `scopesToLoad[0]?.networkIds` before draining it with a concurrency of 3 (`swapProPositionsLoader.ts:339`). `uniqueScopes` preserves `[primary, ...additional]` order (`swapProPositionsLoader.ts:109-111`), so scope order still determines fetch order.

Separately, the effect's `focusSwapPro` and `swapTypeSwitch` dependencies (`SwapMainLand.tsx:1357,1360`) are no longer read anywhere in the effect body. The re-run they force now passes byte-identical arguments, so a tab switch can no longer re-prioritize anything.

## Impact

Native only. `positionSupportNetworkLists.stock` is `swapNetworks.filter(supportStock)` — a handful of chains — while `.swap` is every single/cross-chain swap network (`SwapMainLand.tsx:1292-1301`). With `swap` pinned at index 0, the sort now hoists nearly all networks to the front, leaving the stock networks at an arbitrary position in a long queue drained 3 at a time. A user cold-starting on the Stocks tab (or Pro, now last) sees positions stream in measurably later than before, since `publishNetworkPositions` publishes per network as each task completes.

## Recommendation

Restore active-scope priority inside the helper, which keeps the memo stable while making the ordering explicit:

```ts
// swapPositionPrefetchUtils.ts:40
return scopes
  .filter(
    (scope) =>
      (scope.key === 'pro' ? proNetworksReady : swapNetworksReady) &&
      scope.networkList.length > 0,
  )
  .toSorted((left, right) => {
    if (left.key === activeKey) return -1;
    if (right.key === activeKey) return 1;
    return 0;
  });
```

Add `activeKey: 'swap' | 'stock' | 'pro'` to the parameter object and thread it from `SwapMainLand.tsx:1327-1341`, deriving it as `focusSwapPro ? 'pro' : swapTypeSwitch === ESwapTabSwitchType.STOCK ? 'stock' : 'swap'` with `focusSwapPro`/`swapTypeSwitch` added to the `useMemo` deps. That also makes the effect deps at `SwapMainLand.tsx:1357,1360` genuinely used rather than dead.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/components/CollateralSwitchCell.tsx
```diff
@@ -654,27 +655,51 @@ export function CollateralSwitchCell({
...
     <Stack
       position="relative"
       ai="center"
       jc="center"
-      onPress={(e) => {
-        e.preventDefault();
-        e.stopPropagation();
-      }}
+      onPress={
+        platformEnv.isNative
```
sidmorizon: # 🟡 P2 Row-press guard removed on all native platforms, but the switch swap is iOS-only

## Problem

Two sibling conditions in the same JSX use different platform predicates:

- Line 683: `native={!platformEnv.isNativeIOS}` — only **iOS** swaps to the press-based Tamagui switch.
- Lines 669-675: `onPress={platformEnv.isNative ? undefined : ...}` — **every** native platform loses the wrapper's `stopPropagation` guard.

Before this PR the guard applied unconditionally. On Android the component still renders the RN native `Switch` (`native === true`), which is the configuration that never had the iOS responder-stealing problem this commit targets — yet it loses the guard anyway.

The wrapper `Stack` (line 665) has no width/height, so as a flex child of the table's column `Stack` (`packages/kit/src/components/ListView/TableList.tsx:473`) it stretches to fill the whole collateral cell, which is far wider than the 32 px `extraSmall` track (`packages/components/src/forms/Switch/index.tsx:19-23`).

## Impact

`CollateralSwitchCell` is rendered inside pressable rows: `SuppliedCard.tsx:225,282` feeds `BorrowTableList` → `TableList`, whose rows are `ListItem onPress={handlePress}` (`TableList.tsx:466-469`). On Android, a tap landing inside the collateral cell but off the switch track now activates the row instead of being swallowed.

Because `SuppliedCard` only renders on `gtMd` paths (`BorrowHome.tsx:405` and the `!isPhone` fallback at `:451`, with `isPhone = !gtMd` at `:242`), that row press resolves to `BorrowNavigation.pushToBorrowReserveDetails` (`SuppliedCard.tsx:128-139`). So the practical effect is confined to native tablet layouts: a near-miss on the collateral toggle navigates the user off the Borrow home page unexpectedly. Not a fund-movement path, but a behavior regression that iOS does not incur and Android did not need.

## Recommendation

Scope the guard removal to the platform that needs it, mirroring line 683:

```tsx
// CollateralSwitchCell.tsx:669-675
onPress={
  platformEnv.isNativeIOS
    ? undefined
    : (e) => {
        e.stopPropagation();
      }
}
```

Android then keeps both the RN switch and the guard; iOS keeps the new press-based switch unobstructed. The accompanying test `'does not add a competing press responder around the native switch'` (`CollateralSwitchCell.test.tsx:420-430`) should set `isNativeIOS` rather than `isNative` to match.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/modal/BorrowTokenSelectModal.tsx
```diff
@@ -99,37 +108,63 @@ export default function BorrowTokenSelectModal() {
...
+            }
+          : result;
+      borrowAssetsListCache.set(cacheKey, nextResult);
+      return nextResult;
     },
     [accountId, networkId, provider, marketAddress, action, cacheKey],
     {
       initResult: initialAssetsList,
       watchLoading: true,
+      undefinedResultIfError: true,
```
sidmorizon, line 155: # 🟡 P2 New retry branch is unreachable per the declared types (`result` typed non-nullable)

## Problem

Passing `initResult` (line 153) selects the `usePromiseResult` overload that returns `IUsePromiseResultReturnWithInitValue<T>`, i.e. `result: T` with `undefined` **excluded** (`packages/kit/src/hooks/usePromiseResult.ts:67-76`). But `undefinedResultIfError: true` (line 155) makes the runtime value `undefined` on failure — `usePromiseResult.ts:341-351` calls `setResult(undefined)`.

So the declared type and the runtime disagree. Since `assetsList` is typed `IBorrowAssetsList`, TypeScript narrows `!assetsList` at line 370 to `false`, which makes the entire new `<Empty>` / retry branch (lines 370-388) provably dead code by the declared types. For the same reason `assetsList?.assets ?? EMPTY_ASSETS_LIST.assets` (line 163) reads as a redundant optional chain plus a redundant `??`.

## Impact

The retry UI added for OK-62747 works today only because no rule enforces the type. A future reader — or a later `typescript/no-unnecessary-condition` enablement — has explicit license to delete the error UI and the `?.`/`??` as provably-unreachable, silently reverting this fix to the pre-PR behavior where a validation failure renders as an empty list indistinguishable from "no results". The compiler cannot flag that removal, because by the types it is correct.

## Recommendation

Make the type match the runtime by widening the type argument at line 115:

```ts
} = usePromiseResult<IBorrowAssetsList | undefined>(
  async () => { ... },
  [accountId, networkId, provider, marketAddress, action, cacheKey],
  {
    initResult: initialAssetsList,
    watchLoading: true,
    undefinedResultIfError: true,
    revalidateOnReconnect: true,
  },
);
```

`initResult: initialAssetsList` still satisfies the overload, `result` becomes `IBorrowAssetsList | undefined`, and lines 163 and 370 become type-justified rather than type-contradicting.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/borrowHistoryList.utils.ts
```diff
@@ -1,6 +1,82 @@
...
+          scope.marketAddress === normalizedMarketAddress
+        );
+      })
+        ? 'setCollateral'
+        : undefined;
+    }
+    return 'setCollateral';
+  }
+
+  return parsedTags[0]?.action;
```
sidmorizon, line 119: **Local actions leak across markets**

`getBorrowHistoryActionForLocalTx` accepts `claim` and `setEMode` tags from any market sharing the provider. The selected market can display another market's actions.

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -370,28 +353,115 @@ function BorrowHistoryList() {
...
+            type: action,
+            direction: tx.stakingInfo?.send ? 'send' : 'receive',
+            network: networkMap.get(tx.decodedTx.networkId),
+            token,
+            protocolLogoURI: tx.stakingInfo?.protocolLogoURI,
+          };
+
+          return {
+            item: historyItem,
+            isPending: tx.decodedTx.status === EDecodedTxStatus.Pending,
```
sidmorizon: # 🟠 P1 Failed and dropped local txs render as successful history rows

## Problem

The only status discrimination applied to local txs is `isPending` on line 408. Everything that is not `Pending` falls into `localHistoryItems` (lines 423-425) and is merged into the dated sections at line 426 as an ordinary row.

`EDecodedTxStatus` also has `Failed`, `Dropped`, `Removed`, and `Offline` (`packages/shared/types/tx.ts:54-62`), and `ServiceHistory` deliberately keeps failed txs rather than discarding them: a locally-pending tx whose on-chain receipt is not `Success` is pushed into `confirmedTxs` with `status: EDecodedTxStatus.Failed` (`packages/kit-bg/src/services/ServiceHistory.ts:1490-1493`), then merged into the returned `txs` at `:1727-1737`. `filterHistoryTxs` only drops scam/low-value entries, not failed ones.

Nothing downstream can recover the distinction: `IEnrichedHistoryItem` (`BorrowHistoryList.tsx:58-62`) carries no status field, and `HistoryItem` (lines 129-167) renders only avatar, title, subtitle, and amount — no failure badge, no color change, no status text.

## Impact

A user disables a reserve as collateral on Aave and the tx reverts on-chain. On the next 30s poll `fetchAccountHistory` marks it `Failed`, `isPending` becomes `false`, and the row lands in today's section titled "Collateral" — visually identical to a successful one. The user reads the history as "collateral setting changed" while the on-chain collateral flag, and therefore the health factor and liquidation exposure, is unchanged. The same applies to a failed `setEMode`, where the e-mode category drives LTV and liquidation threshold. Dropped and Removed txs behave identically.

This is new surface: the established precedent for local rows shows *pending only* — `ServiceStaking.fetchLocalStakingHistory` reads `getAccountLocalHistoryPendingTxs` (`packages/kit-bg/src/services/ServiceStaking.ts:297-324`), consumed at `packages/kit/src/views/Staking/pages/HistoryList/index.tsx:355-360`. This increment is the first place that admits non-pending local txs into a history list.

## Recommendation

Classify the status explicitly at lines 406-409 rather than collapsing to a boolean:

```ts
return {
  item: historyItem,
  isPending: tx.decodedTx.status === EDecodedTxStatus.Pending,
  isFailed:
    tx.decodedTx.status === EDecodedTxStatus.Failed ||
    tx.decodedTx.status === EDecodedTxStatus.Dropped ||
    tx.decodedTx.status === EDecodedTxStatus.Removed,
};
```

Then either drop failed entries before the merge at lines 423-425 — matching the remote endpoint, which only reports effective actions — or carry the flag onto `IEnrichedHistoryItem` (line 58) and render a failure state in `HistoryItem` (title color `$textCritical` plus a status suffix at line 147). Dropping is the smaller change and keeps the list's "these actions happened" semantics intact.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -370,28 +353,115 @@ function BorrowHistoryList() {
...
+          } => Boolean(entry),
+        );
+
+      const pendingLocalItems = localHistoryEntries
+        .filter((entry) => entry.isPending)
+        .map((entry) => entry.item);
+      const localHistoryItems = localHistoryEntries
+        .filter((entry) => !entry.isPending)
+        .map((entry) => entry.item);
+      const listMap = [...enrichedList, ...localHistoryItems].reduce(
```
sidmorizon: # 🟡 P2 Local rows are appended after remote rows within a day, breaking descending time order

## Problem

`[...enrichedList, ...localHistoryItems].reduce(...)` (lines 426-440) groups by day while preserving array order, so every local item lands **after** every remote item in the same day bucket, regardless of its own timestamp.

Only the *sections* are sorted, and only by their first element: `.toSorted((a, b) => b.data[0].timestamp - a.data[0].timestamp)` (line 447). Items inside a section are never sorted. The previous implementation had no such mixing — it grouped the server-ordered `enrichedList` alone — so intra-day ordering was implicitly correct.

## Impact

Today the user supplies at 10:00 (remote row) and then sets collateral at 14:00 (local-only row). Today's section renders "Supply · 10:00" first and "Collateral · 14:00" beneath it — the newest action of the day sits at the bottom of a list that is descending everywhere else. With several local actions in one day the intra-day order becomes arbitrary rather than merely inverted.

Section-to-section ordering stays correct, since day buckets cannot interleave, which is why this is P2 rather than P1 — but the visible ordering inside the current day is wrong whenever a local action follows a remote one, which is the common case right after any collateral or e-mode change.

## Recommendation

Sort the merged array before grouping, at line 426:

```ts
const listMap = [...enrichedList, ...localHistoryItems]
  .toSorted((a, b) => b.timestamp - a.timestamp)
  .reduce(
    (map, item) => { /* unchanged */ },
    new Map<string, IEnrichedHistoryItem[]>(),
  );
```

That also makes the existing section sort at line 447 exact rather than approximate, since `data[0]` then genuinely holds each day's newest item. Worth applying the same sort to `pendingLocalItems` (lines 420-422) so multiple outstanding pending actions list newest-first too.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Staking/hooks/useUtilsHooks.ts
```diff
@@ -64,21 +64,24 @@ export function useTrackTokenAllowance({
...
+  refreshOnMount?: boolean;
   tokenAddress: string;
   spenderAddress: string;
   approveType?: EApproveType;
 }) {
   const isLegacyApprove = approveType === EApproveType.Legacy;
   const isExistApproveTarget = !!spenderAddress;
   const shouldFetchInitialAllowance =
-    initialValue === undefined && isExistApproveTarget;
+    isExistApproveTarget && (initialValue === undefined || refreshOnMount);
```
sidmorizon, line 84: # 🟡 P2 `refreshOnMount` is not mount-scoped, leaving the reset effect able to strand `loading` at `true`

## Problem

`shouldFetchInitialAllowance` is a derived value, not a one-shot latch. With `refreshOnMount: true` it stays `true` for the hook's entire lifetime, not just on mount. That changes the meaning of the reset effect at lines 112-120:

```ts
114:    setLoading(shouldFetchInitialAllowance);   // now true, not false
118:        prev.targetKey === allowanceTargetKey ? (initialValue ?? '0') : '0',
120:  }, [allowanceTargetKey, initialValue, shouldFetchInitialAllowance]);
```

The reset effect depends on `initialValue`; the fetch effect (lines 171-184) deliberately does **not**. So when the seed changes while `allowanceTargetKey` is unchanged:

1. `allowance` is reset from the freshly fetched chain value back to the seed (line 118) — the seed that the call site's own comment at `ProtocolLendingActionDialogContent.tsx:1551-1553` describes as lagging;
2. `loading` is set back to `true` (line 114) with nothing left to clear it — the only `setLoading(false)` lives in the fetch effect's `finally` (lines 157-163) and in its short-circuit (line 141), neither of which re-runs.

Downstream that is a permanently spinning, `shouldApprove === false` confirm button: `useBorrowApproval.ts:482` returns `false` while `loadingAllowance`, and `ProtocolLendingActionDialogContent.tsx:1884-1890` keeps the button in `loading`.

Related, same root: the JSDoc on line 75 (and `useBorrowApproval.ts:156`) says "on mount", but the flag also disables the `!txDetails && !shouldFetchInitialAllowance` short-circuit at line 139 for the hook's whole life, so it refetches on every fetch-effect dep change rather than once.

## Impact

**Latent, not reachable at this HEAD — I could not build a live repro, and this is P2 for that reason.** The only opted-in call site is `ProtocolLendingActionDialogContent.tsx:1555`, whose seed is `protocolInfo?.approve?.allowance`. That `useManagePage` passes `undefinedResultIfReRun: true` (`ProtocolLendingActionDialogContent.tsx:1087`), so every refetch nulls `protocolInfo` first, which drives `approveTarget` to `undefined` (line 1331-1333), empties `spenderAddress`, flips `allowanceTargetKey`, and re-runs the fetch effect. It self-heals.

The concern is that the correctness of a public, defaulted-`false` flag on two shared hooks (`useUtilsHooks.ts:76`, `useBorrowApproval.ts:157`) now rests on an invisible invariant owned by an unrelated hook's option. The seed-changes-in-place pattern already exists here — `StakeSection.tsx:860-863` and `WithdrawSection.tsx:1419` pass `currentAllowance={initialAllowanceResult?.allowanceParsed}`, which goes `undefined` → value with account/network/token/spender all stable. The first call site to combine that pattern with `refreshAllowanceOnMount: true` gets a dead confirm button in an approval flow with no obvious cause.

## Recommendation

Add `initialValue` to the fetch effect's dep array at lines 171-184 so the reset and the refetch always move together, and drop the now-redundant `refreshOnMount` at line 182 (it is not read in the effect body — `shouldFetchInitialAllowance` on line 183 already subsumes it).

Alternatively, make the flag genuinely mount-scoped by latching it in a ref consumed once, and correct the comments at line 75 and `useBorrowApproval.ts:156` to match whichever semantics you keep.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Staking/hooks/useUtilsHooks.ts
```diff
@@ -176,6 +179,7 @@ export function useTrackTokenAllowance({
     allowanceTargetKey,
     isLegacyApprove,
     isExistApproveTarget,
+    refreshOnMount,
```
sidmorizon, line 182: 🟡 P2: Allowance refresh stays permanently loading

Severity: `non-severe`

When `initialValue` changes under `refreshOnMount`, `shouldFetchInitialAllowance` stays true without retriggering the fetch. Approval remains blocked in its loading state.

Suggested fix:

```
    refreshOnMount,
    initialValue,
```

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -370,28 +359,120 @@ function BorrowHistoryList() {
...
+        )
+        .map((tx: IAccountHistoryTx) => {
+          const action = getBorrowHistoryActionForLocalTx({
+            tx,
+            provider,
+            networkId,
+            marketAddress,
+          });
+          if (!action || BORROW_HISTORY_REMOTE_ACTIONS.has(action)) {
+            return undefined;
```
sidmorizon: 🟡 P2: Pending Borrow transactions disappear

Severity: `non-severe`

A local pending `supply`, `borrow`, `withdraw`, or `repay` is discarded before the remote endpoint indexes it. Borrow history omits the transaction meanwhile.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/useEModeSwitch.ts
```diff
@@ -61,10 +61,14 @@ export function useEModeSwitch({
       protocol: earnUtils.getEarnProviderName({ providerName: provider }),
       tags: [
         EEarnLabels.Borrow,
-        buildBorrowTag({ provider, action: 'setEMode' }),
+        buildBorrowTag({
+          provider,
+          action: 'setEMode',
+          setEModeScope: { networkId, marketAddress },
```
sidmorizon, line 67: # 🟠 P1 Scoped `setEMode` tag defeats the e-mode pending guard, allowing a second broadcast over a pending one

## Problem

This commit changed the `setEMode` emitters to emit **only** the scoped tag, but the matcher that decides whether an e-mode action is pending was not updated and is a strict equality against the **unscoped** tag.

Emitter (this line, plus the second emitter at `BorrowEModeNeedAction/useEModeNeedActionFlow.ts:333-339`):
```ts
tags: [EEarnLabels.Borrow, buildBorrowTag({ provider, action: 'setEMode', setEModeScope: { networkId, marketAddress } })]
// → 'borrow:aave:setEMode:v1:evm--1:0x87870bca…'
```

Matcher — `BorrowEModeSwitch/emodeUtils.ts:33`, unchanged since `d042cd71` and not part of this PR:
```ts
return actions.some((action) => tag === buildBorrowTag({ provider, action }));
// buildBorrowTag({ provider, action: 'setEMode' }) falls through to `return base` (Staking/utils/utils.ts:119)
// → 'borrow:aave:setEMode'
```

`'borrow:aave:setEMode:v1:evm--1:0x87870bca…' === 'borrow:aave:setEMode'` is `false`. There is no prefix or parse fallback to rescue it: `useStakingPendingTxs.ts:647-648` applies the matcher as a plain per-tag predicate, `tags.some((tag) => tagMatcher(tag))`.

The asymmetry with the sibling emitter shows the intended pattern: `CollateralSwitchCell.tsx:540-555` deliberately emits the legacy tag **and** the scoped tag, which is exactly why `setCollateral` still matches.

## Impact

Concrete scenario on Aave `evm--1`, market `0x87870Bca3F3fD6335C3F4ce8392D69350B4fA4E2`:

1. User confirms an e-mode switch → tx broadcast with the scoped tag; the page pops (`BorrowEModeSwitch/index.tsx:115-118`).
2. While that tx is still pending, the user re-enters the E-Mode page.
3. `pendingTagMatcher` (`index.tsx:143-151`) matches nothing → `pendingCount === 0` → `isEModePendingGuardActive` returns `false` → `pendingGuardActive === false` (`index.tsx:175-180`).
4. The category selector stays enabled (`index.tsx:358`), the footer is neither disabled nor loading (`index.tsx:405`, `:332`), and the guard in `onFooterConfirm` (`index.tsx:265`) passes — so the user can **broadcast a second `setEMode` transaction on top of a pending one**, decided against a stale reading of the current e-mode.

E-mode drives LTV and liquidation threshold, so two racing e-mode transactions are a real risk to position safety, not just a UI nit.

Secondarily, `useStakingPendingTxsByInfo({ onRefresh: refreshManagementState, onRefreshDelayMs: 3000 })` (`index.tsx:153-163`) never observes the tx entering or leaving the pending set, so the automatic refresh when the switch confirms is lost. The same applies to the Need Action screen's `switch` step (`BorrowEModeNeedAction/index.tsx:399-404`, `:490-495`).

This is a regression introduced here: at `c23391d3` the emitted tag was `'borrow:aave:setEMode'` and the guard worked.

## Recommendation

Emit both tags, mirroring `CollateralSwitchCell.tsx:540-555` — the new history reader already prefers the scoped form when both are present (`borrowHistoryList.utils.ts:80-81`, `:103-119`), so market scoping is preserved:

```ts
tags: [
  EEarnLabels.Borrow,
  buildBorrowTag({ provider, action: 'setEMode' }),
  buildBorrowTag({ provider, action: 'setEMode', setEModeScope: { networkId, marketAddress } }),
],
```

Apply the same at `useEModeNeedActionFlow.ts:333-339`.

Alternatively, keep the single scoped tag and rewrite `emodeUtils.ts:33` to compare via `parseBorrowTag(tag)` on `provider`/`action` instead of whole-string equality — but that must still match legacy unscoped tags already persisted in users' local history. `hasPendingSetCollateral` (`collateralControls.utils.ts:126-142`) is a working model for the parse-based form.

Worth adding a regression test asserting that the tag a `setEMode` transaction actually carries is matched by `isEModeBorrowActionTag`, since that emitter/matcher pair is what broke.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -370,28 +359,124 @@ function BorrowHistoryList() {
...
+          const action = getBorrowHistoryActionForLocalTx({
+            tx,
+            provider,
+            networkId,
+            marketAddress,
+          });
+          const isPending = tx.decodedTx.status === EDecodedTxStatus.Pending;
+          if (
+            !action ||
+            (BORROW_HISTORY_REMOTE_ACTIONS.has(action) && !isPending)
```
sidmorizon, line 392: # 🟡 P2 Pending rows are invisible under every filter except "All"

## Problem

Local history is fetched **only** when `filterType === 'all'` — `BorrowHistoryList.tsx:331-338` resolves `localHistoryResp` to `Promise.resolve(undefined)` for any other filter. So `localHistoryEntries`, and therefore `pendingLocalItems`, are empty whenever a filter is selected.

That restriction was harmless before this commit: the only local rows the gate at line 385 could previously admit were `setCollateral`, `setEMode` and `claim`, none of which appear in the filter dropdown. This line now also admits pending `supply` / `borrow` / `withdraw` / `repay` — which are exactly the four options the dropdown offers (`BorrowHistoryList.tsx:219-225`).

## Impact

A user submits a supply, then taps the filter and selects "Supply" to check on it. The pending row they are looking for is not there — it only appears under "All". The confirmed supplies show normally, so the list looks authoritative while silently omitting the in-flight one. This is the filter a user is most likely to reach for right after submitting, so the feature this commit adds disappears in precisely the view where it is most wanted.

No incorrect data is displayed; the row is simply unreachable outside the default view, which is why this is P2 rather than P1.

## Recommendation

Fetch local history for the remote-action filters too, and filter the entries by the selected action. At `BorrowHistoryList.tsx:331-338`:

```ts
filterType === 'all' || BORROW_HISTORY_REMOTE_ACTIONS.has(filterType as IBorrowAction)
  ? backgroundApiProxy.serviceHistory.fetchAccountHistory({ accountId, networkId }).catch(/* unchanged */)
  : Promise.resolve(undefined),
```

then, in the `.map` at `BorrowHistoryList.tsx:382-388`, drop entries that do not match the active filter:

```ts
if (filterType !== 'all' && action !== filterType) {
  return undefined;
}
```

The pending/dated partition at `BorrowHistoryList.tsx:434-441` needs no change.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -370,28 +359,124 @@ function BorrowHistoryList() {
...
+          ) {
```
sidmorizon, line 393: # 🟡 P2 Newly admitted pending rows render with no amount when the remote token table misses

## Problem

For a local row, `amount` comes from the **local** transaction (`BorrowHistoryList.tsx:410`, `actionToken?.amount`), but `token` is resolved **exclusively** from the remote payload — `tokenMap` is built from `historyResp.tokens` (`:345-347`) and the fallback at `:393-398` searches that same array.

`HistoryItem` renders the amount only when both are present (`BorrowHistoryList.tsx:152`):
```tsx
{item.amount && item.token?.info.symbol ? ( … ) : null}
```

Before this commit the admitted local actions were `setCollateral` / `setEMode` / `claim`, which carry no displayable send/receive amount, so a token miss was invisible. The four actions newly admitted by this gate are precisely the ones whose amount is the row's primary content — and the local transaction already holds a complete token at `tx.stakingInfo.send.token` / `receive.token` (symbol and `logoURI` included) that the code never falls back to.

## Impact

`historyResp.tokens` is a lookup table for the returned `list` — the existing `console.warn('Missing network or token for item: …')` at `BorrowHistoryList.tsx:354-358` treats a miss for a remote row as anomalous, which is only coherent if `tokens` covers exactly what `list` references.

So the miss is reachable whenever the pending action's token is not already referenced by the market's remote history. The clearest case is the first supply on a market the account has no history on: `list` is empty, `tokens` is empty, and the pending row renders as a bare "Supply · 14:32" line with a generic protocol avatar and no number at all. Same for the first action on a reserve the account has not touched before.

That is the first-use moment this commit exists to serve, and the row arrives conveying neither amount nor asset.

## Recommendation

In the `.map` at `BorrowHistoryList.tsx:390-399`, fall back to the locally-known token when the remote lookup misses, keeping the remote entry preferred so confirmed rows stay priced:

```ts
const remoteToken = actionToken
  ? tokenMap.get(actionToken.token.address) ||
    historyResp.tokens.find(/* unchanged */)
  : undefined;
const token =
  remoteToken ??
  (actionToken
    ? {
        price: '0',
        price24h: '0',
        info: { ...actionToken.token, uniqueKey: actionToken.token.address },
      }
    : undefined);
```

Shape per `IBorrowHistory['tokens'][number]` / `IEarnToken` in `packages/shared/types/staking.ts`. The price fields are unused by `HistoryItem`, which reads only `token.info.symbol` and `token.info.logoURI`.

One caveat on confidence: the mechanism and the render gate are pinned down exactly, but whether the backend ever returns tokens beyond those referenced by `list` is not verifiable from this repo. The empty-history case does not depend on that question.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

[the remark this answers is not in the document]
zhaono1: Fixed in d526e4b377 (including acb4cf5432). Pending balance-changing borrow actions now carry a v1 network/market scope at all write sites, and getBorrowHistoryActionForLocalTx rejects scoped tags for other markets while preserving legacy fallback. Added scoped supply regression tests.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/useEModeSwitch.ts
zhaono1: Fixed in acb4cf5432. isEModeBorrowActionTag now parses borrow tags and matches the action, so both scoped and legacy setEMode tags activate the pending guard. Added a scoped-tag regression test.

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
zhaono1: Fixed in d526e4b377. Local history now loads for all balance-changing filters and filters local entries by the selected action, so pending rows remain visible in Supply/Borrow/Withdraw/Repay tabs.

packages/kit/src/views/Borrow/pages/BorrowEModeNeedAction/useEModeNeedActionFlow.ts
```diff
@@ -328,9 +328,18 @@ export function useEModeNeedActionFlow({
     (action: 'repay' | 'setCollateral' | 'setEMode'): IStakingInfo => ({
       label: EEarnLabels.Borrow,
       protocol: earnUtils.getEarnProviderName({ providerName: provider }),
-      tags: [EEarnLabels.Borrow, buildBorrowTag({ provider, action })],
+      tags: [
+        EEarnLabels.Borrow,
+        buildBorrowTag({
```
sidmorizon, line 333: # 🟠 P1 Missed `borrowScope` write site: E-Mode Need-Action repay still emits an unscoped tag

## Problem

`acb4cf5432` scoped the balance-changing actions at three write sites — `StakeSection.tsx:742-747`, `WithdrawSection.tsx:1141-1146` and `:1304-1309`, and `ProtocolLendingActionDialogContent.tsx:1396-1401` — each now pushing the legacy tag **and** the scoped one. This factory was not updated, and it scopes only `setEMode`:

```ts
buildBorrowTag({
  provider,
  action,
  ...(action === 'setEMode' ? { setEModeScope: { networkId, marketAddress } } : {}),
}),
```

So `stakingInfo('repay')` still produces the bare `borrow:aave:repay`. It is consumed for real repay transactions at `useEModeNeedActionFlow.ts:725` (the blocker-repay step) and `:843` (`useBorrowApproval({ action: 'repay', … })`). An exhaustive grep of `buildBorrowTag` callers confirms this is the only remaining unscoped `supply`/`borrow`/`withdraw`/`repay` emitter.

Hand-trace, repaying USDC in Aave market A and then opening market B's history:
- `parseBorrowTag('borrow:aave:repay')` → `parts = ['borrow','aave','repay']`; `parts[3]` is `undefined ≠ 'v1'` → `borrowScope` undefined.
- `scopedRemoteActionTags` (`borrowHistoryList.utils.ts:110-112`) is empty, so the new market-scoped branch at `:113-117` is skipped.
- Falls through to `return parsedTags[0]?.action` (`:119`) → `'repay'`, with no market check.

`networkId` and `marketAddress` are already hook params and already in this `useCallback`'s dep array (`:342`), so nothing blocks passing the scope.

## Impact

The cross-market leak this pair of commits set out to close stays open for every repay started from the E-Mode Need-Action flow, and this PR makes it newly *visible*: before it, local pending repay rows were discarded outright by `BORROW_HISTORY_REMOTE_ACTIONS`. Now `2a77b507` admits pending ones and `d526e4b3`'s `shouldLoadLocalHistory` also loads local history under the **Repay** filter — so the leaked row shows in market B under both "All" and "Repay".

Same callback, same root cause: `stakingInfo('setCollateral')` (used at `:935`) is also unscoped, so `hasPendingSetCollateral` (`collateralControls.utils.ts:140-142`) falls to the legacy `tags.includes(providerTag)` branch and provider-wide-locks every collateral switch across markets. One fix covers both.

## Recommendation

Scope all three actions here, and push the legacy tag alongside the scoped one to match the three sites already fixed (so legacy consumers such as `E_MODE_PENDING_GUARD_ACTIONS` and `hasPendingSetCollateral` keep matching):

```ts
tags: [
  EEarnLabels.Borrow,
  buildBorrowTag({ provider, action }),
  buildBorrowTag({
    provider,
    action,
    ...(action === 'setEMode' ? { setEModeScope: { networkId, marketAddress } } : {}),
    ...(action === 'repay' ? { borrowScope: { networkId, marketAddress } } : {}),
    ...(action === 'setCollateral' && step?.reserveAddress
      ? { setCollateralScope: { networkId, marketAddress, reserveAddress: step.reserveAddress } }
      : {}),
  }),
],
```

A regression test asserting that a repay tag emitted from this flow is rejected by `getBorrowHistoryActionForLocalTx` for a different `marketAddress` would lock this down — the scoped-supply test added in this commit does not cover this emitter.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Borrow/pages/BorrowHistoryList.tsx
```diff
@@ -340,14 +319,30 @@ function BorrowHistoryList() {
         };
       }
 
-      const historyResp =
-        await backgroundApiProxy.serviceStaking.getBorrowHistory({
+      // The remote Borrow history endpoint currently covers balance-changing
+      // actions only. Refresh local history for the "all" view and the
+      // balance-changing filters so pending rows remain visible; metadata-only
+      // actions (for example setCollateral) remain available in "all" only.
+      const shouldLoadLocalHistory =
```
sidmorizon, line 326: # 🟡 P2 Approval transactions surface as fully-formed pending rows for the business action

## Problem

`shouldLoadLocalHistory` now pulls local history under the four action filters, not just "All". But a borrow **approval** transaction inherits the business `stakingInfo` verbatim — `useBorrowApproval.ts:639`, `:717` and `:893` pass the caller's `stakingInfo` straight through alongside `approvesInfo`, so the approve tx persists the same label, the same borrow tags (now including the new scoped one), and an intact `send`/`receive`.

Trace an Aave repay of 100 USDC that needs an ERC-20 approve. The approve transaction persists:

```
tags: ['Borrow', 'borrow:aave:repay', 'borrow:aave:repay:v1:evm--1:0xmarket', <stakeTag>]
send: { token: USDC, amount: '100' }
```

In the `.map` at `BorrowHistoryList.tsx:381-450`: `getBorrowHistoryActionForLocalTx` returns `'repay'` (scoped tag matches the market), `filterType === 'repay'` passes the equality gate at `:391`, and `isPending` is true so the `!isPending` drop at `:392` does not fire. `actionToken = stakingInfo.send`, so the row renders as **"Repay −100 USDC"** — but its `txHash` is the *approve* transaction, so tapping it opens the approve details.

The Solana LUT-setup path (`useUniversalBorrowHooks.ts:419`) uses `buildBorrowTrackingStakingInfo`, which strips `send`/`receive` but keeps `tags`, so that variant renders as an amount-less row instead.

## Impact

Under the four action filters — newly reachable as of this line — a pending row that is actually an approval is indistinguishable from the real borrow action, and shows an amount the transaction does not move.

Two things bound it: `settleApprovalTransaction` waits for the approve to confirm before firing the business tx, so the rows are sequential rather than simultaneous; and once confirmed, `BORROW_HISTORY_REMOTE_ACTIONS.has('repay') && !isPending` at `:392` drops the approve row. So it self-heals, which is why this is P2 — but it is a wrong, tappable row for the duration of the approve, and a USDT reset-to-zero produces two of them in sequence.

## Recommendation

Prefer the approach this codebase already has: give approval transactions a tracking-only `stakingInfo` at `useBorrowApproval.ts:639/717/893` — one that keeps `tags` for order tracking but drops `send`/`receive`, as `buildBorrowTrackingStakingInfo` (`useUniversalBorrowHooks.ts:63-81`) already does for the LUT case.

If you would rather keep the change inside this file, drop entries whose `decodedTx` contains only a token-approve action, near the existing gate at `BorrowHistoryList.tsx:389`.

Confidence note: the call sites and the `:389-393` gate are hand-traced and certain; how often users sit on a filtered view during the approve window is not something I can determine from the repo.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

packages/kit/src/views/Staking/utils/utils.ts
```diff
@@ -45,6 +52,15 @@ export type IBorrowClaimScope = {
   marketAddress: string;
 };
 
+export type IBorrowActionScope = IBorrowClaimScope;
+
+const BORROW_MARKET_SCOPED_ACTIONS = new Set<IBorrowAction>([
```
sidmorizon, line 57: # 🟡 P2 Two independent copies of the same four-action set gate the write and read halves of the new tag format

## Problem

The new `borrowScope` shape is gated on `BORROW_MARKET_SCOPED_ACTIONS` (this line) — used by `buildBorrowTag:135` and twice by `parseBorrowTag:231,236` — while the matcher that consumes it is gated on a separately declared set, `BORROW_HISTORY_REMOTE_ACTIONS` (`borrowHistoryList.utils.ts:23-28`):

```ts
const scopedRemoteActionTags = parsedTags.filter(
  (tag) => BORROW_HISTORY_REMOTE_ACTIONS.has(tag.action) && tag.borrowScope,
);
```

Both currently list exactly `Supply`, `Borrow`, `Withdraw`, `Repay` — I compared them member for member, so **there is no live bug at this head**. The issue is that one invariant now has two sources of truth, and divergence fails silently in both directions:

- Added to `BORROW_MARKET_SCOPED_ACTIONS` only → the tag is written and parsed with a `borrowScope`, but `scopedRemoteActionTags` never picks it up, so `getBorrowHistoryActionForLocalTx` falls through to `parsedTags[0]?.action` at `:119` — the scope is written to disk and then ignored, reopening the cross-market leak with no error anywhere.
- Added to `BORROW_HISTORY_REMOTE_ACTIONS` only → `parseBorrowTag` never populates `borrowScope`, the filter is always empty, same silent fall-through.

This is the same emitter/matcher desync shape that already bit `emodeUtils.ts` earlier in this PR, and it is now baked into a `:`-delimited format persisted in users' local transaction history, where a wrong read is not recoverable after the fact.

## Impact

Maintainability, with a silent-failure shape rather than a loud one. No user impact today.

## Recommendation

Export the set once and import it, so the emitter, the parser and the matcher agree by construction:

```ts
// utils.ts:57
export const BORROW_MARKET_SCOPED_ACTIONS = new Set<IBorrowAction>([ … ]);
```

then in `borrowHistoryList.utils.ts:23-28` replace the local declaration with a re-export of that set, keeping the `BORROW_HISTORY_REMOTE_ACTIONS` name as an alias if other consumers depend on it. `buildBorrowTag:135`, `parseBorrowTag:231/236` and `borrowHistoryList.utils.ts:111` then all read the same object.

[Claude session ↗](https://claude.ai/code/session_01LCkWH9NpnHqUpLpPBo93VN)

---

zhaono1:
Superseded by PR #13331, which rebuilds the ten issue fixes as one commit per issue.

PR #13317, opened 2026-09-10
feat: export cloud backups as password-protected zip files

sidmorizon:
The developer-only **Download Backup Data** action now exports a password-protected ZIP containing plaintext `backup.json`, so inspecting or recovering its wallet/account data does not require the original backup password or cloud account ID after extraction.

- Ask for the existing cloud backup password locally, then decrypt the selected Google Drive or iCloud backup in the background runtime.
- Generate a fresh cryptographically random 32-character password for every export and create an AES-256 ZIP entirely in memory. Only the encrypted archive is downloaded or written to the native sharing cache.
- After download/share returns, show a dialog displaying the ZIP extraction password with a **Copy password** button. Native share dismissal includes cancellation; no saved-success toast is inferred from it.
- Preserve plaintext recovery credentials and account data without exporting cloud account identifiers, the original password, or wrapped credential copies. Reject backups with additional encrypted credentials that cannot be exported as plaintext.

- Pin zip.js to 2.10.0 to meet the seven-day release-age gate. Use the physical Streams polyfill entry for Metro, register the new native module IDs, and patch the optional URI codec loader so Hermes can compile the archive dependency.

Validation: 29 targeted Jest tests cover cloud decryption, password-only ZIP extraction, fresh passwords, wrong-password rejection, nested encrypted credentials, dialog sequencing/copying, binary downloads, native temporary-file cleanup, Metro resolution on iOS/Android, and Hermes compilation. An additional independent 7-Zip test verifies AES-256 compatibility and rejects an incorrect password (30 tests total). The strict iOS union build, module-ID registry check, bundle architecture check, and both startup budgets pass locally: main 13.09/13.3 MB and background 18.29/19.5 MB. The minimum-release-age check passes for both added dependencies. All local commit/PR lint, format, and TypeScript checks pass; The previously failing minimum-release-age and native startup graph checks also pass on GitHub for commit 86e752baaa. Verification uses local code and synthetic data; no installed OneKey app or user wallet database is opened.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1ca551ea-708e-5dd8-835a-067af74c8870)

packages/kit/src/views/Onboardingv2/components/KeylessWalletBackupInfo.tsx
```diff
@@ -42,8 +44,23 @@ function KeylessWalletBackupDetails({
...
+      const { userId } =
+        await backgroundApiProxy.serviceCloudBackupV2.getCloudAccountInfo();
+      if (!userId) {
+        throw new OneKeyLocalError(
+          'Google user ID is required to export backup data',
+        );
+      }
+      const payload: IBackupDataExportPayload = {
+        ...backup.payload,
+        googleUserId: userId,
```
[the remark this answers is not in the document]
sidmorizon: 已在 ec8db3085f4a085f084ae38ff34fc412672e4401 修复。导出流程已改为在 bg 中解密备份，并使用每次新生成的 32 字符密码学安全随机密码创建 AES-256 ZIP；文件不再携带云账号 userId、原备份密码或原备份密文。main 与 RNFS/分享组件只处理加密 ZIP，解压密码在下载/分享结束后单独通过 Dialog 展示和复制。已通过随机密码、错误密码拒绝、明文 JSON 解压测试及独立 7-Zip 兼容性验证。本评论对应的旧实现已被替换，标记为已解决。

packages/kit/src/views/Onboardingv2/components/KeylessWalletBackupInfo.tsx
```diff
@@ -30,35 +33,66 @@ function KeylessWalletBackupDetails({
...
+        const { default: backgroundApiProxy } =
+          await import('@onekeyhq/kit/src/background/instance/backgroundApiProxy');
+        const { archiveBase64, password: zipPassword } =
+          await backgroundApiProxy.serviceCloudBackupV2.exportBackupArchive({
+            recordId: backupRecordId,
+            password,
+          });
+        if (!passwordDialog.isExist()) return;
+        await passwordDialog.close();
+        await downloadAsFile({
```
sidmorizon, line 56: **Save failures disappear silently**

When `downloadAsFile` rejects, `onSubmit` has already closed its dialog and returns the error to an unmounted footer. Users see neither the password nor a save-failure message.

sidmorizon:
Claude review session: https://claude.ai/code/session_01SFBZbEvZqYSinLseZHo4pt

packages/kit-bg/src/services/ServiceCloudBackupV2/createBackupExportArchive.ts
```diff
@@ -0,0 +1,34 @@
+import type { IBackupDataExportArchive } from '@onekeyhq/shared/src/cloudBackup/cloudBackupTypes';
+import stringUtils from '@onekeyhq/shared/src/utils/stringUtils';
+import type { IPrimeTransferData } from '@onekeyhq/shared/types/prime/primeTransferTypes';
+
+export async function createBackupExportArchive(
+  data: IPrimeTransferData,
+): Promise<IBackupDataExportArchive> {
+  // Hermes has no Streams API. Load it before evaluating zip.js, on demand.
+  if (typeof globalThis.TransformStream === 'undefined') {
+    await import('web-streams-polyfill/polyfill');
```
sidmorizon: # 🔴 P1 `web-streams-polyfill/polyfill` is unresolvable by Metro — the native background graph fails to build

## Problem

`createBackupExportArchive.ts:10` imports the bare subpath `web-streams-polyfill/polyfill`. `web-streams-polyfill@4.2.0` exposes that specifier **only through its `exports` map** — the published tarball's root contains just `dist/`, `types/`, `package.json`, `LICENSE`, `README.md`. There is no `polyfill.js` and no `polyfill/` directory; the map entry is `"./polyfill": { "default": "./dist/polyfill.js" }`.

`apps/mobile/metro.config.js:152` sets `config.resolver.unstable_enablePackageExports = false;` ("Fix for Metro resolver with 'subpath exports'"), so Metro ignores `exports` entirely and resolves the specifier as a literal path under the package root. Nothing exists there, so resolution fails.

This is not a missing dependency — `packages/kit-bg/package.json:13` declares `web-streams-polyfill: 4.2.0` and `yarn.lock` resolves it. It is a missing Metro alias. That same config file already carries the established workaround for exactly this class of package: `hyperliquidSigningPath` (~line 155), `HWK_SUBPATH_ALIASES` (~line 166), `MYSTEN_SUI_SUBPATH_PREFIX` (~line 176), `LEDGER_CJS_ENTRY_PACKAGES` (~line 209) — with a comment instructing contributors to append new sub-paths. No entry was added for `web-streams-polyfill`.

The sibling import on line 13, `@zip.js/zip.js/index-native.js`, is fine by contrast: that file physically exists at the package root, so Metro resolves it as a literal path.

## Impact

The iOS/Android **background** Metro graph fails to build outright — a hard bundling error, not a runtime fallback:

```
UnableToResolveError: Unable to resolve module web-streams-polyfill/polyfill from
  packages/kit-bg/src/services/ServiceCloudBackupV2/createBackupExportArchive.ts
```

The entire native app cannot be built from this commit, so the breakage extends far beyond this feature. The `typeof globalThis.TransformStream === 'undefined'` guard on line 9 does not help: Metro resolves dynamic `import()` statically at graph-build time regardless of the runtime branch.

Desktop/web/extension are unaffected — `development/webpack/webpack.base.config.js` and `development/rspack/rspack.base.config.ts` override neither `exportsFields` nor `conditionNames`, so they honor the `exports` map. Jest also resolves it (jest-resolve honors `exports`, picking the `require` condition). That is why the 26 new unit tests pass while the native build breaks — the suite gives false confidence on this specific line.

## Recommendation

Either add a Metro alias alongside the existing ones, resolving in Node-land where `exports` is honored:

```js
const webStreamsPolyfillPath = require.resolve('web-streams-polyfill/polyfill');
// ...in resolveRequest:
if (moduleName === 'web-streams-polyfill/polyfill') {
  return { type: 'sourceFile', filePath: webStreamsPolyfillPath };
}
```

or import the physical path directly at `createBackupExportArchive.ts:10`:

```ts
await import('web-streams-polyfill/dist/polyfill.js');
```

The second resolves under both Metro (literal path) and webpack/rspack (`"./dist/*": "./dist/*"` is present in the exports map), and needs no config change.

[Claude session ↗](https://claude.ai/code/session_01SFBZbEvZqYSinLseZHo4pt)

---

packages/kit-bg/package.json
```diff
@@ -4,11 +4,13 @@
   "private": true,
   "main": "src/index.tsx",
   "dependencies": {
+    "@zip.js/zip.js": "2.14.0",
```
sidmorizon: # 🔴 P1 `@zip.js/zip.js@2.14.0` is 1 day old and violates the repo's 7-day minimum-release-age gate

## Problem

`packages/kit-bg/package.json:7` pins `"@zip.js/zip.js": "2.14.0"`, published one day before this head commit.

The repo enforces a minimum release age. Root `package.json` sets:

```json
"minimumReleaseAge": {
  "days": 7,
  "allowlist": ["@onekeyhq/*", "@onekeyfe/*", "shell-quote"],
  "blockOnFailure": true,
  "registryUrl": "https://registry.npmjs.org"
}
```

and `development/scripts/minimum-release-age/checkPackageAge.ts:112` fails any package where `ageDays < config.days`. `@zip.js/*` is not on the allowlist.

The check derives its package set from `git diff x -- yarn.lock`, and this PR adds a brand-new `"@zip.js/zip.js@npm:2.14.0"` resolution block, so the failure is introduced here rather than pre-existing. CI confirms:

```
FAIL: 1 package(s) younger than 7 days:
  - @zip.js/zip.js@2.14.0 (1 day(s) old)
Summary: 1 ok, 1 too young, 0 skipped, 0 errors
Blocking: one or more packages are younger than the minimum release age threshold.
```

(`web-streams-polyfill@4.2.0` is the "1 ok".)

## Impact

The "Supply Chain - Minimum Release Age (24.x)" job is blocking (`blockOnFailure: true`), so the PR cannot merge as-is.

Beyond CI, this newly introduced third-party package sits directly on the path that handles plaintext wallet secrets: `createBackupExportArchive.ts:28` feeds `stringUtils.stableStringify(data)` — which contains `privateData.decryptedCredentials`, i.e. plaintext mnemonics and private keys for every wallet in the backup — straight into `ZipWriter`. It runs in the background runtime on iOS/Android/extension and in the single shared runtime on desktop/web. A one-day-old release of a dependency in that position is precisely the exposure the 7-day window exists to catch.

## Recommendation

Pin to a release older than 7 days. **`2.9.0`** (published 2026-08-31, 10 days old) is the safest nearby choice; `2.8.61` (13 days) also works. Both expose the `./index-native.js` subpath, ship `index-native.js` at the package root, and declare the identical API surface used here — `ZipWriter`, `Uint8ArrayWriter`, `Uint8ArrayReader`, and every option passed at `createBackupExportArchive.ts:18-24` (`password`, `encryptionStrength`, `zipCrypto`, `useWebWorkers`, `useCompressionStream`, `dataDescriptor`, `level`). No code change is needed to downgrade; the range is already an exact pin, so only the literal string and the lockfile change.

Avoid `2.10.0` — it is exactly 7 days old and sits on the `<` boundary, so it will pass or fail depending on when CI runs.

[Claude session ↗](https://claude.ai/code/session_01SFBZbEvZqYSinLseZHo4pt)

---

packages/kit-bg/src/services/ServiceCloudBackupV2/ServiceCloudBackupV2.ts
```diff
@@ -329,6 +330,47 @@ class ServiceCloudBackupV2 extends ServiceBase {
...
+  @backgroundMethod()
+  @toastIfError()
+  async exportBackupArchive(params: {
+    recordId: string;
+    password: string;
+  }): Promise<IBackupDataExportArchive> {
+    if (!params.recordId || !params.password) {
+      throw new OneKeyLocalError('Backup record ID and password are required');
+    }
+    const backup = await this.download({ recordId: params.recordId });
```
sidmorizon, line 342: # 🟡 P2 Plaintext export of every wallet seed skips the local password verification that `delete()` requires

## Problem

`exportBackupArchive` (`ServiceCloudBackupV2.ts:335-372`) decrypts the cloud backup and writes every wallet's plaintext credentials into `backup.json`, gated **only** by the cloud backup password. It never calls `servicePassword.promptPasswordVerify`.

In the same class, two strictly less sensitive operations do gate on it:

```ts
// ServiceCloudBackupV2.ts:489-495
async delete(params: { recordId: string; skipManifestUpdate?: boolean }): Promise<void> {
  await this.backgroundApi.servicePassword.promptPasswordVerify({
    reason: EReasonForNeedPassword.Security,
  });
```

`deleteAllBackups` at line 516 does the same. Deleting a remote file destroys data but reveals nothing; this new path reveals every mnemonic and private key in the backup, and is the one that skips the check.

## Impact

Anyone holding an unlocked device plus the backup password can extract plaintext mnemonics for all wallets without knowing the device/app password. That inverts the risk ordering the rest of this service follows, where secret-revealing flows are exactly the ones that gate on `promptPasswordVerify`.

The dev-settings gate on the entry point (`KeylessWalletBackupInfo.tsx:110` `devSettingsOnly`, plus the 10-tap `MultipleClickStack` trigger in release builds) reduces exposure but does not remove it — dev settings are user-toggleable in production builds. That is why this is P2 rather than P1.

## Recommendation

Mirror `delete()`: add the verification as the first statement of `exportBackupArchive`, before the `this.download(...)` call on line 342. `EReasonForNeedPassword` is already imported and used in this file, so this is a two-line change:

```ts
await this.backgroundApi.servicePassword.promptPasswordVerify({
  reason: EReasonForNeedPassword.Security,
});
```

[Claude session ↗](https://claude.ai/code/session_01SFBZbEvZqYSinLseZHo4pt)

---

packages/kit/src/views/Onboardingv2/components/KeylessWalletBackupInfo.tsx
```diff
@@ -30,35 +33,66 @@ function KeylessWalletBackupDetails({
...
+      onSubmit: async (password) => {
+        const { default: backgroundApiProxy } =
+          await import('@onekeyhq/kit/src/background/instance/backgroundApiProxy');
+        const { archiveBase64, password: zipPassword } =
+          await backgroundApiProxy.serviceCloudBackupV2.exportBackupArchive({
+            recordId: backupRecordId,
+            password,
+          });
+        if (!passwordDialog.isExist()) return;
+        await passwordDialog.close();
```
sidmorizon, line 55: **Dismissed export still opens sharing**

When export finishes during dismissal, `passwordDialog.isExist()` remains true through the closing animation. The canceled export then opens the share sheet.

originalix (approved):

PR #13318, opened 2026-09-11
fix: perps layout and inputs(OK-61375 OK-61397 OK-61481 OK-61655)

Minnzen:
[OK-61375](<https://onekeyhq.atlassian.net/browse/OK-61375>) [OK-61397](<https://onekeyhq.atlassian.net/browse/OK-61397>) [OK-61481](<https://onekeyhq.atlassian.net/browse/OK-61481>) [OK-61655](<https://onekeyhq.atlassian.net/browse/OK-61655>)

----------

## Summary
- Keep the desktop order book symmetric and vertically centered during live resizing (OK-61375).
- Preserve the total holdings count when hiding small balances and align table headers with rows at narrow widths (OK-61397, OK-61481).
- Restore a localized native Done button for iOS TP/SL numeric inputs (OK-61655).

## Intent & Context
These four independently committed fixes address verified Perps UI issues. Resizing should update the order book continuously with fixed row height and spacing. Hiding assets below $5 should filter the list without changing the holdings tab total. Historical trade headers should stay aligned at every horizontal scroll position. iOS users should be able to dismiss the TP/SL decimal keyboard without leaving the editor.

## Root Cause
- The desktop vertical order book stretched rows and could allocate an extra bid row. Its deferred display updates also needed to account for row-count changes during resizing.
- The holdings count used the small-balance visibility filter, so the label counted displayed assets instead of all holdings.
- Header and body scroll content could resolve to different widths because of intrinsic cell text widths and omitted row padding. The trade direction cell also overrode its configured minimum width.
- TP/SL inputs relied on custom accessory views that did not provide a usable dismissal button in the reported iOS flow.

## Design Decisions
- Recompute the visible depth during dragging, keep bid/ask counts equal, and distribute unused vertical space above and below the ladder. Apply fixed row sizing only to the desktop vertical order book.
- Count holdings independently of the list filter, preserving the existing zero-balance and USDC deduplication rules.
- Give header and body the same minimum width and zero flex basis in the shared desktop table. Preserve the existing scrolling mechanism.
- Use the native iOS input accessory with the existing `global_done` translation, retaining decimal input and avoiding custom accessory IDs. Dismissing the keyboard does not submit a trade.

## Changes Detail
- `OrderBook/index.tsx` and `perpLayoutUtils.ts`: symmetric fixed-height rows, centered content, and immediate display reconciliation when the visible row count changes.
- `useVisibleSpotHoldingsCount.ts`: remove the visibility preference from the count data source.
- `CommonTableListView.tsx` and `TradesHistoryRow.tsx`: unify scroll content sizing and preserve the direction column minimum width.
- `TpslInput.tsx`: replace custom keyboard accessories with localized native dismissal controls.
- Three focused test files cover order-book layout and empty-data resizing, plus holdings-count stability.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Desktop, Web, Extension, and Mobile; the keyboard behavior is iOS-only.
- **Risk Areas**: The table sizing fix is shared by Perps info-panel lists. Order-book resizing interacts with coalesced display updates. Native keyboard presentation depends on iOS behavior.

## Test plan
- [x] User verified continuous desktop order-book resizing with symmetric rows, fixed spacing, and centered remaining space.
- [x] User verified that hiding small holdings changes the list while preserving its total count.
- [x] Verified narrow historical-trade tables at the start, middle, and end of horizontal scrolling, including long direction text.
- [x] Reviewed shared column sizing across all info-panel lists and TWAP subtables.
- [x] User verified iOS TP/SL keyboard dismissal and the localized Done label.
- [x] Rebase onto the latest `origin/x` completed without conflicts; all four patches remained unchanged.
- [x] Focused regression tests: 3 suites / 14 tests passed after rebase and dependency synchronization. Jest reported an existing duplicate desktop package-name warning.
- [x] Changed-file lint and formatting passed after rebase; TypeScript passed after synchronizing dependencies and reapplying the latest base-branch patches.
- [ ] Full agent gate: existing repository skill-metadata budgets exceed their limits (23 implicit skills vs. 22; 3172 description characters vs. 3000). This is unrelated to the nine changed files.
- [ ] GitHub CI and required review.

[OK-61375]: https://onekeyhq.atlassian.net/browse/OK-61375?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61397]: https://onekeyhq.atlassian.net/browse/OK-61397?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61481]: https://onekeyhq.atlassian.net/browse/OK-61481?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-61655]: https://onekeyhq.atlassian.net/browse/OK-61655?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-671a0a72-2480-55d3-8686-9d520fa8ecac)

sidmorizon:
Claude review session: https://claude.ai/code/session_01NuWLUK6cDS7Nah3xv36ZBZ

packages/kit/src/views/Perp/components/TradingPanel/inputs/TpslInput.tsx
```diff
@@ -24,64 +21,19 @@ import {
...
-      {platformEnv.isNativeIOS ? (
-        <InputAccessoryView nativeID={accessoryViewId}>
-          <TpslDoneButton onDone={onDone} />
-        </InputAccessoryView>
-      ) : null}
-    </>
+    <Input
+      {...inputProps}
+      returnKeyType={platformEnv.isNativeIOS ? 'done' : undefined}
+      inputAccessoryViewButtonLabel={
```
originalix, line 30: P1: [Restore a usable iOS numeric-keyboard dismiss action]

When a TP/SL field uses `keyboardType="decimal-pad"` on iOS, the keypad has no return key. This change removes the `InputAccessoryView` that previously rendered a Done button calling `Keyboard.dismiss()`.

`inputAccessoryViewButtonLabel` only labels an accessory view when one exists; it does not create a dismiss control. Keep or replace the accessory view with a real Done action wired to `Keyboard.dismiss()` so users can close the keypad after entering TP/SL values.

packages/kit/src/views/Perp/components/OrderBook/OrderBook.empty.test.tsx
```diff
@@ -113,6 +126,36 @@ jest.mock('../../hooks/useTradingPrice', () => ({
 }));
 
 describe('OrderBook empty vertical state', () => {
+  it('updates both sides immediately as the available height changes', () => {
```
sidmorizon, line 129: # 🟡 [P2] Resize test never reaches the coalesced path it is meant to guard

## Problem
The behavioural change in this commit is the new flush key: `visualEpoch` now carries `resolvedMaxLevelsPerSide` alongside `depthEpoch` so `useRafCoalesced` flushes synchronously when the visible row count changes (`OrderBook/index.tsx:1147-1151`, applied at `:1152-1169`). This test renders with `asks: []` / `bids: []`, and on the empty branch the rendered rows do not come from the coalesced ladders at all:

```ts
// OrderBook/index.tsx:1176-1181
verticalAsks = isEmpty ? verticalEmptyLevels.slice(0, resolvedMaxLevelsPerSide) : reversedAskLadder.levels;
verticalBids = isEmpty ? verticalEmptyLevels : bidLadder.levels;
```

`verticalEmptyLevels` is a plain `useMemo` over `resolvedMaxLevelsPerSide` (`index.tsx:1004-1013`). The only other consumer of the coalesced output in this render is `DepthBarColumn`, which line 116 mocks to `() => null`. So the `--` counts (73/55/61/73/55) are a pure function of `resolvedMaxLevelsPerSide`, and reverting `visualEpoch` back to `depthEpoch` leaves all five assertions green.

Related, in the same file: the sibling test derives its expectation from `getVerticalOrderBookLayout(containerHeight, maxLevelsPerSide)` (`:162-165`) without the new `fixedRows` argument, while the component under test renders with `isDesktopVertical === true` and therefore takes the `fixedRows` branch (`index.tsx:950-957`). It passes only because both branches happen to return `levelsPerSide: 12` at 640/18.

## Impact
The commit titled "stabilize order book resizing" ships with no regression protection for the stale-ladder / row-count-vs-fixed-box-height problem it fixes: a future refactor can drop the `resolvedMaxLevelsPerSide` term from the flush key and this suite stays green.

## Recommendation
Add a non-empty case to this test: supply more than 13 levels per side, stop mocking `DepthBarColumn` away, resize 640 → 500 inside `act`, and assert that the number of rendered rows equals `getVerticalOrderBookLayout(500, 18, true).levelsPerSide` in the same commit (without letting a rAF tick run). That path goes through `useRafCoalesced` and fails when `visualEpoch` is reverted. Separately, pass `true` as the third argument at `:162-165` so the expectation tracks the branch the component actually executes.

[Claude session ↗](https://claude.ai/code/session_01NuWLUK6cDS7Nah3xv36ZBZ)

---

packages/kit/src/views/Perp/layouts/perpLayoutUtils.test.ts
```diff
@@ -62,33 +62,24 @@ describe('getPerpDesktopChartSplitSizes', () => {
...
-    expect(layout.extraBidLevels).toBe(0);
-    expect(layout.rowHeight).toBeCloseTo(22.92, 5);
-  });
-
-  it('stretches rows to fill the pane once the level cap is reached', () => {
-    const layout = getVerticalOrderBookLayout(1000, 12);
-    expect(layout.levelsPerSide).toBe(12);
-    expect(layout.extraBidLevels).toBe(0);
-    expect(layout.rowHeight).toBeCloseTo(36.52, 5);
+  it('preserves the existing layout for other platforms', () => {
```
sidmorizon: # 🟡 [P2] Deletes the only coverage of the `fixedRows === false` branch, which this PR leaves live

## Problem
Four existing tests are replaced by this two-line "preserves the existing layout" test plus an `it.each` that passes `fixedRows = true` on every row. The legacy branch at `perpLayoutUtils.ts:180-197` is untouched by this PR but loses both of its specific guards:

- `getVerticalOrderBookLayout(660, 12)` → `levelsPerSide: 12, extraBidLevels: 0` — the only assertion covering the `levelsPerSide < maxLevelsPerSide` condition at `perpLayoutUtils.ts:182` that stops the spare bid row from pushing past the cap.
- `getVerticalOrderBookLayout(1000, 12)` → `rowHeight ≈ 36.52` — the only assertion covering the row-stretch formula at `perpLayoutUtils.ts:191-196`.

What survives for that branch is `rowHeight ≈ 22.12` at 640/18 and `extraBidLevels === 1` at 660/18 (lines 66-67); `levelsPerSide` is no longer asserted for it at all.

That branch is still the live one everywhere `isDesktopVertical` is false, since the flag is `variant === 'web' && !platformEnv.isNative && !horizontal` (`OrderBook/index.tsx:925-926`) and it is what gets forwarded as `fixedRows` at `index.tsx:950-957`. The vertical call site at `PerpOrderBook.tsx:1238-1256` renders under `gtMd`, so a native large-screen render takes the legacy path with the cap guard and the stretch formula both unprotected.

## Impact
Unexplained coverage loss on code the PR deliberately did not change, against the repo rule not to remove existing code unless the request requires it. A regression in the extra-bid-row cap there would put one more bid row on screen than `maxLevelsPerSide` allows, with nothing failing in CI.

## Recommendation
Keep the two deleted cases verbatim next to the new `it.each` — `getVerticalOrderBookLayout(660, 12)` asserting `{ levelsPerSide: 12, extraBidLevels: 0, rowHeight ≈ 22.92 }` and `getVerticalOrderBookLayout(1000, 12)` asserting `rowHeight ≈ 36.52` — and restore the `levelsPerSide` assertions to the `preserves the existing layout for other platforms` test at lines 65-68. They cost two lines each and are the only thing pinning the branch this PR left alone.

[Claude session ↗](https://claude.ai/code/session_01NuWLUK6cDS7Nah3xv36ZBZ)

---

packages/kit/src/views/Perp/components/TradingPanel/inputs/TpslInput.tsx
Minnzen: This is handled by the native implementation in the React Native 0.86.2 version used by this project. For `decimal-pad`, `setDefaultInputAccessoryView` creates a `UIToolbar` when a supported return key (including `done`) or an accessory button label is present. The button invokes `handleInputAccessoryDoneButton`, which ends editing. Removing the custom accessory ID allows this default path to run.

Source: [toolbar creation and dismissal in RN 0.86.2](https://github.com/react/react-native/blob/v0.86.2/packages/react-native/React/Fabric/Mounting/ComponentViews/TextInput/RCTTextInputComponentView.mm#L644-L706).

The current input supplies `returnKeyType="done"` and the existing `global_done` translation. The reporter also verified keyboard dismissal on iOS. Keeping the current implementation and resolving this thread.

packages/kit/src/views/Perp/components/OrderBook/OrderBook.empty.test.tsx
Minnzen: Fixed in 849d9fbc8c. Added a populated order-book test with 18 levels per side and an active tick option. It renders the real depth-bar components, blocks animation-frame callbacks, and resizes 640 → 500 → 683 → 500. It asserts both visible row and depth-bar counts, plus the prices and indices returned when selecting the outermost ask and bid. The placeholder test now explicitly uses `fixedRows=true`.

Verified the regression guard by temporarily replacing `visualEpoch` with `depthEpoch`: the new test fails at 500px with 24 rows instead of the expected 18. Restoring the production code makes it pass. Both affected suites pass (17 tests).

packages/kit/src/views/Perp/layouts/perpLayoutUtils.test.ts
Minnzen: Fixed in 849d9fbc8c. Restored all four original `fixedRows=false` layout tests alongside the new fixed-row cases, including the maximum-level cap at 660/12 and row stretching at 1000/12. The original `levelsPerSide`, `extraBidLevels`, and row-height assertions are preserved. Both affected suites pass (17 tests); production layout code is unchanged.

originalix (dismissed):

originalix (approved):

sidmorizon (approved):

PR #13319, opened 2026-09-11
feat: improve native crash diagnostics

huhuanming:
## Summary

- initialize the native Sentry SDK during Android/iOS application startup, independently of the React Native main and background runtimes
- retain sanitized native fatal-event JSON under the existing `logs/crashes` tree on the next launch; Android NDK outbox replay passes through the configured `beforeSend` callback before upload
- include the same logs tree in both normal log ZIP export and Recovery-only export, without a React Native diagnostics module or a second export path
- require `SENTRY_DSN_REACT_NATIVE` for EAS Release builds while keeping local Release builds available for profiling and diagnosis
- enable Android native symbol/source upload and include native source context in the existing iOS dSYM upload phase

## Privacy and retention

- redact sensitive keys and values, wallet material, mnemonic sequences, credentials, identifiers, URL queries/fragments, and common PII before writing exportable crash JSON
- keep full non-sensitive native diagnostic context for QA; retain at most 5 native crash reports for 7 days
- keep screenshots, view hierarchy capture, and default PII disabled
- remove normal and Recovery ZIP artifacts on the next healthy launch's post-first-render background cleanup, leaving the current share artifact readable for deferred consumers

## Validation

- `git diff --check` and standalone Android Java 17 compilation
- latest `yarn agent:check --profile commit`: non-TypeScript checks pass; `tsc-staged` is currently blocked by existing ImageV2/NativeList type mismatches outside this change
- focused Jest coverage for React Native Sentry initialization options and event sanitization
- Android Java compilation plus Sentry 8.33 envelope/native-event filtering and sanitizer fixtures for prefixed mnemonic arrays, wallet-secret fields, 128-character extended private keys, and ordinary diagnostic text
- iOS Objective-C syntax compilation and Xcode workspace parsing
- isolated iOS config-injection checks for local Release without DSN, EAS Release without DSN, and EAS Release with DSN/wordlist reuse
- post-fix privacy, security, concurrency, Recovery-only export, and log-ZIP audit: no P1/P2 blockers

## Notes

- native debug-file upload still requires the authenticated EAS/Sentry environment
- no additional package build was run for this audit follow-up


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-0b94edd1-4bde-5e91-8624-09fd8ffb228f)

sidmorizon:
Claude review session: https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,655 @@
...
+
+    private static JSONArray buildExceptions(List<SentryException> exceptions) throws Exception {
+        JSONArray result = new JSONArray();
+        if (exceptions == null) {
+            return result;
+        }
+        for (SentryException exception : exceptions) {
+            JSONObject item = new JSONObject();
+            putString(item, "type", exception.getType());
+            putString(item, "value", exception.getValue());
```
huhuanming: **[P1] 收紧 native crash 自由文本的持久化边界**

这里会把 `exception.value`（同类还有 mechanism description、Android exit description/ANR trace，以及 iOS MetricKit payload 中的字符串）写入 `cache/logs/crashes`。当前 native 正则可以拦截部分私钥/助记词/token，但实测不会拦截 `0x` 40-byte 钱包地址、邮箱和常见 44 字符 base58 地址；这些文件随后会被现有整目录日志 ZIP、客服日志上传和系统分享读取。建议改成字段白名单，默认不落盘自由文本；确需保留的字段统一走经过测试的脱敏器，并覆盖各链钱包地址、邮箱、URL query token、支持的私钥/助记词格式。

packages/kit/src/views/Setting/pages/Tab/DevSettingsSection/SentryCrashSettings.tsx
```diff
@@ -30,9 +32,19 @@ export function SentryCrashSettings() {
   ];
   if (platformEnv.isNative) {
     sections.push(
+      <SectionPressItem
+        icon="DownloadOutline"
+        key="SentryCrashDiagnosticsExport"
+        testID="sentry-native-crash-export"
+        title="Export Logs With Native Crash Diagnostics"
+        onPress={() => {
+          void exportLogs(`OneKeyNativeCrashDiagnostics-${Date.now()}`);
```
huhuanming: **[P1] 让导出 ZIP 遵守崩溃日志的保留策略**

每次点击都会在 `Caches/logs_zip` 生成一个带时间戳的新 ZIP；当前只轮转 `logs/crashes/*.json`，没有代码删除或轮转这些 ZIP。因此崩溃副本可能在 7 天/5 份限制之后继续留存，并继续包含在后续分享/上传边界内。建议导出/上传完成后删除 ZIP，并在创建前清理历史 ZIP；Android 原子写入遗留的 `.json.tmp` 也应纳入清理。

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,655 @@
...
+        BeforeEnvelopeCallback existingBeforeEnvelope =
+            options.getBeforeEnvelopeCallback();
+        options.setBeforeEnvelopeCallback((envelope, hint) -> {
+            if (existingBeforeEnvelope != null) {
+                existingBeforeEnvelope.execute(envelope, hint);
+            }
+            persistCrashEvents(
+                getReactApplicationContext(),
+                envelope,
+                options.getSerializer()
```
sidmorizon: # 🟠 [P1] `beforeEnvelope` hook persists JS fatal errors as "native" crash reports and evicts real native reports

## Problem

`persistCrashEvents` (`OneKeySentryCrashDiagnosticsModule.java:350-370`) is installed as the SDK-wide `BeforeEnvelopeCallback` (line 159) and writes every envelope `Event` item whose `isCrashed()` is true under the name `"sentry-native-" + eventId + ".json"` (line 385). It never checks where the event came from, and every JavaScript error the RN SDK sends passes through this callback:

- The RN global error handler tags fatal JS errors with `mechanism: { handled: false, type: 'onerror' }` (`@sentry/react-native@8.2.0` `reactnativeerrorhandlers.ts`), so `SentryEvent.isCrashed()` (`getUnhandledException() != null`, sentry-java 8.33.0 `SentryEvent.java:218-233`) returns true for them.
- The JS transport sends them via `NATIVE.sendEnvelope` → `RNSentryModuleImpl.captureEnvelope` → `InternalSentrySdk.captureEnvelope` → `scopes.captureEnvelope(...)` → `SentryClient.sendEnvelope`, which invokes `options.getBeforeEnvelopeCallback()` (`SentryClient.java:882-888`). `enableNative` stays true in the manual-init branch, so this path is live after this PR.

The same file-name and `cleanupReports` rotation (`MAX_REPORT_COUNT = 5`, line 60; `cleanupReports`, line 572) is shared with the SIGSEGV/NDK, ANR and `ApplicationExitInfo` reports, and `persistPendingNativeCrashEnvelopes` (line 334) applies the same unfiltered logic to cached `*.envelope` files at startup.

## Impact

A handful of JS fatals (which are already fully captured in app logs and Sentry) rotate the genuinely native crash/ANR/exit reports out of `cache/logs/crashes` before anyone can export them, which defeats the stated purpose of the PR. The surviving files are also mislabeled `sentry-native-*` while containing only `platform: "javascript"` frames, so anyone reading the export is misled. Each JS envelope is additionally JSON-deserialized a second time on the send path.

## Recommendation

Skip non-native events in both hooks. In `persistCrashEvents` (line 364) and the `beforeSend` wrapper (line 148), bail out when `"javascript".equals(event.getPlatform())` (the RN JS SDK sets `event.platform = 'javascript'`; sentry-java events are `java`, NDK events `native`). Alternatively restrict to `event.getSdk()` names `sentry.native.android` / `sentry.java.android.react-native`. Consider a separate file prefix (or counter) for JS-originated reports if you decide to keep them at all, so they cannot compete with native reports for the five retained slots.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/build.gradle
```diff
@@ -18,7 +18,7 @@ sentry {
     // for Sentry. This executes sentry-cli automatically so
     // you don't need to do it manually.
     // Default is disabled.
-    uploadNativeSymbols = false
+    uploadNativeSymbols = true
```
sidmorizon, line 32: # 🟡 [P2] `uploadNativeSymbols = true` is not gated on `SENTRY_DISABLE_AUTO_UPLOAD`, so local release builds now fail or silently upload

## Problem

This flips `uploadNativeSymbols` to `true` unconditionally (`includeNativeSources = true` on line 28 rides along). In the pinned Sentry Android Gradle Plugin 4.13.0 (`apps/mobile/android/build.gradle:38`), `configureNativeSymbolsTask` registers `uploadSentryNativeSymbolsFor<Variant>` for every non-debuggable variant and wires it with `hookWithAssembleTasks` → `assemble.finalizedBy(uploadTask)` (plus the install and bundle tasks). The task runs `sentry-cli debug-files upload … --include-sources` through `asSentryCliExec()`, which throws `SentryCliException` (a `GradleException`) on any non-zero exit other than the "outdated endpoint" case. Nothing in the plugin reads `SENTRY_DISABLE_AUTO_UPLOAD`; that variable only gates the React Native source-map upload in `packages/core/sentry.gradle`, which this file applies on line 147.

The repo's committed `apps/mobile/android/sentry.properties` has no `auth.token` (CI appends it only in `release-android.yml:98-100`), yet several release-variant builds run with the upload explicitly disabled and no token:

- `development/scripts/android-release-build-deploy.sh:220-227` runs `SENTRY_DISABLE_AUTO_UPLOAD=true ./gradlew :app:assembleProdRelease`.
- `apps/mobile/.detoxrc.js:170-171` runs `SENTRY_DISABLE_AUTO_UPLOAD=true ./gradlew assembleProdRelease …`.

## Impact

On a machine without a Sentry auth token, `:app:assembleProdRelease` now fails in the finalizer after the APK is already built, so the release script never reaches its "Native build complete" step and the Detox Android release build breaks. On a machine that does have a personal token or `~/.sentryclirc`, a local build silently uploads native symbols and native source files (`--include-sources`) from the developer's checkout into the production Sentry project. CI is unaffected because `react-native-unittest.yml` builds the debuggable `assembleProdDebug` and `release-android.yml` writes the token.

## Recommendation

Gate the plugin flag on the same switch the RN upload already honors, e.g. on line 21:

```groovy
uploadNativeSymbols = System.getenv('SENTRY_DISABLE_AUTO_UPLOAD') != 'true'
```

and keep `includeNativeSources` tied to the same gate. If you want the task to still exercise locally without uploading, set `autoUploadNativeSymbols = false` under that condition instead (the plugin then passes `--no-upload`), and note the behavior in `development/scripts/android-release-build-deploy.sh`.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

packages/shared/src/modules3rdParty/sentry/index.native.ts
```diff
@@ -55,13 +73,38 @@ export const initSentry = () => {
...
+    attachViewHierarchy: false,
+    sendDefaultPii: false,
+  };
+  const crashDiagnosticsModule = NativeModules.OneKeySentryCrashDiagnostics as
+    | IOneKeySentryCrashDiagnosticsModule
+    | undefined;
+  let didInitializeNativeSdk = false;
+  try {
+    didInitializeNativeSdk =
+      crashDiagnosticsModule?.initialize(nativeInitializationOptions) === true;
```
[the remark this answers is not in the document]
huhuanming: Fixed: native Sentry now starts from Android Application.onCreate and iOS AppDelegate before either React runtime; bg JS also initializes its own Sentry client.

apps/mobile/ios/OneKeyWallet/OneKeySentryCrashDiagnostics.m
```diff
@@ -0,0 +1,515 @@
...
+  result = [result stringByReplacingOccurrencesOfString:@"\r" withString:@" "];
+
+  static NSArray<NSRegularExpression *> *patterns;
+  static dispatch_once_t onceToken;
+  dispatch_once(&onceToken, ^{
+    NSArray<NSString *> *rawPatterns = @[
+      @"(?:0x)?[0-9a-fA-F]{64}",
+      @"\\b[5KL][1-9A-HJ-NP-Za-km-z]{50,51}\\b",
+      @"\\b[xyzXYZ](?:prv|pub)[1-9A-HJ-NP-Za-km-z]{107,108}\\b",
+      @"(?:\\b[a-z]{3,8}\\b[\\s,]+){11,}\\b[a-z]{3,8}\\b",
```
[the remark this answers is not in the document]
huhuanming: Fixed: Android/iOS use the bundled exact 2048-word BIP39 set and redact any sequence of 3+ words, including arrays and multiline forms.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/BaseMainApplication.java
```diff
@@ -452,6 +456,9 @@ public void onCreate() {
       return;
     }
 
+    OneKeySentryCrashDiagnosticsModule.persistHistoricalProcessExitDiagnostics(this);
+    OneKeySentryCrashDiagnosticsModule.persistPendingNativeCrashEnvelopes(this);
```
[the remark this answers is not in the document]
huhuanming: Fixed: only lightweight native handler initialization remains on launch; ApplicationExitInfo/envelope and MetricKit history processing run on serial utility/background queues.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,655 @@
...
+        for (File report : reports) {
+            if (now - report.lastModified() > MAX_REPORT_AGE_MS) {
+                if (!report.delete()) {
+                    android.util.Log.w(NAME, "Unable to remove expired crash diagnostics");
+                }
+            } else {
+                retained.add(report);
+            }
+        }
+        retained.sort(Comparator.comparingLong(File::lastModified).reversed());
```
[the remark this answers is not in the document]
huhuanming: Fixed: report mtime is set to the actual event/exit/payload time before grouped rotation.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
huhuanming: Addressed per final product decision: full local diagnostics are retained, but all strings are recursively sanitized (credentials, URLs, emails, wallets/keys, and BIP39 including multiline/JSON), top-level user/request are excluded, and network transport remains strict. The final privacy/security audit found no blockers.

packages/kit/src/views/Setting/pages/Tab/DevSettingsSection/SentryCrashSettings.tsx
huhuanming: Fixed: logs_zip is cleaned before/after export (7 days/max 5), share/upload cleanup runs in finally, and Android temporary reports are cleaned; Recovery ZIPs are also deleted.

apps/mobile/android/app/build.gradle
huhuanming: Fixed: uploadNativeSymbols/includeNativeSources remain enabled, while autoUploadNativeSymbols is disabled when SENTRY_DISABLE_AUTO_UPLOAD or SENTRY_DISABLE_NATIVE_DEBUG_UPLOAD is true; SAGP emits --no-upload locally.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
huhuanming: Fixed without dropping requested JS fatals: JS-origin events use sentry-js-* and rotate independently from native/ANR/exit reports, so they cannot evict native diagnostics.

apps/mobile/ios/OneKeyWallet/OneKeySentryCrashDiagnostics.m
```diff
@@ -0,0 +1,515 @@
...
+  return value;
+}
+
+static NSString *OneKeyCrashDiagnosticsDirectory(void)
+{
+  NSString *cachesDirectory = NSSearchPathForDirectoriesInDomains(
+      NSCachesDirectory,
+      NSUserDomainMask,
+      YES).firstObject;
+  return [cachesDirectory stringByAppendingPathComponent:@"logs/crashes"];
```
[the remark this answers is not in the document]
huhuanming: Fixed: iOS Recovery recursively exports logs and crashes/*.json, protects staging/final ZIP, and deletes it after sharing.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,939 @@
...
+        });
+    private static final Set<String> SAFE_CONTEXT_KEYS = new HashSet<>(
+        Arrays.asList("app", "device", "os", "runtime", "gpu")
+    );
+    private static boolean initialized;
+    private static volatile Set<String> mnemonicWords = Collections.emptySet();
+
+    private static final Pattern SENSITIVE_KEY_PATTERN = Pattern.compile(
+        "(?i).*(?:password|passwd|passphrase|secret|token|authorization|cookie|sessionid|apikey|privatekey|mnemonic|seed|recoveryphrase|credential|bearer).*"
+    );
```
sidmorizon: 🟠 P1: Wallet secrets survive crash redaction

Crash fields such as `pinHash` or `backendShare` bypass `isSensitiveKey` and remain in persisted diagnostics. Recovery log exports can expose those wallet secrets.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,939 @@
...
+            persistHistoricalProcessExitDiagnostics(applicationContext);
+            persistPendingNativeCrashEnvelopes(applicationContext);
+        });
+    }
+
+    public static void persistPendingNativeCrashEnvelopes(Context context) {
+        cleanupReports(
+            new File(context.getCacheDir(), "logs/crashes"),
+            System.currentTimeMillis()
+        );
```
sidmorizon: **Startup cleanup erases active reports**

Concurrent `persistPendingNativeCrashEnvelopes` and `writeReport` can delete the active `.tmp` file. `cleanupReports` runs outside `FILE_LOCK`, so committing that crash report fails.

packages/kit/src/views/Setting/pages/Tab/exportLogs/index.native.ts
```diff
@@ -325,3 +335,14 @@ export const uploadLogBundle = async ({
...
+export const uploadLogBundle = async (args: {
+  uploadToken: string;
+  digest: ILogDigest;
+}): Promise<{ digest: ILogDigest; result: ILogUploadResponse }> => {
+  try {
+    return await uploadLogBundleInternal(args);
+  } finally {
+    await disposeLogDigest(args.digest);
+  }
+};
```
sidmorizon: **Uploaded digest references deleted file**

Successful `uploadLogBundle` returns a digest whose `filePath` was already deleted by `disposeLogDigest`. Callers cannot reuse, inspect, or export the returned bundle.

Suggested fix:

```
export const uploadLogBundle = uploadLogBundleInternal;
```

packages/kit/src/provider/Container/ErrorToastContainer/ErrorToasts.tsx
```diff
@@ -37,84 +36,20 @@ interface IErrorActionParams {
...
-        const token = await backgroundApiProxy.serviceLogger.requestUploadToken(
-          {
-            sizeBytes: digest.sizeBytes,
-            sha256: digest.sha256,
-          },
-        );
-        await uploadLogBundle({
-          uploadToken: token.uploadToken,
-          digest,
-        });
```
sidmorizon, line 95: **Support contact drops diagnostic upload**

Pressing `ContactSupportButton` now only opens Intercom and skips the existing diagnostic upload. No log bundle reaches the support backend from this action.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
sidmorizon: # 🟠 [P1] Unlocked startup `cleanupReports` deletes the `.tmp` file another thread is committing, losing the crash report

## Problem

`writeReport` (lines 706-744) serializes on `FILE_LOCK`, writes `fileName + ".tmp"`, fsyncs, then `renameTo(target)`. `cleanupReports` (lines 746-758) unconditionally deletes every `*.tmp` in the directory. Two callers invoke `cleanupReports` **outside** `FILE_LOCK`:

- `persistPendingNativeCrashEnvelopes(Context)` here (lines 241-244)
- `persistHistoricalProcessExitDiagnostics` (line 269)

Both run on `DIAGNOSTICS_EXECUTOR` at every launch via `collectHistoricalDiagnosticsAsync` (`BaseMainApplication.java:464`). If a crash occurs during startup (the crash-loop case this feature is meant to capture), Sentry's `beforeSend` on the crashing thread calls `persistCrashEvent(..., true)` → `writeReport`, while the executor thread runs the unlocked cleanup and unlinks the in-flight `.tmp`. `renameTo` then fails (`IllegalStateException("Unable to commit crash diagnostics file")`, swallowed at lines 483-485) or the bytes land in an unlinked inode.

## Impact

Exactly the startup crash that the diagnostics are supposed to preserve produces no `sentry-native-*.json`. The window recurs on every cold start, and the unlocked rotation in the same call can also race the locked writer's own rotation.

## Recommendation

Take `FILE_LOCK` inside `cleanupReports` (simplest: wrap its body in `synchronized (FILE_LOCK)`, which is reentrant for the `writeReport` caller at line 742), or wrap the two call sites at lines 241 and 269. Optionally only delete `.tmp` files older than a few seconds so an in-flight write is never swept.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

packages/kit/src/views/Setting/pages/Tab/exportLogs/index.native.ts
```diff
@@ -51,15 +51,19 @@ export const exportLogs = async (filename: string) => {
...
+    await Share.shareAsync(logFilePath, {
+      dialogTitle: 'OneKey Logs',
+      mimeType: 'application/zip',
+      UTI: 'public.zip-archive',
+    }).catch(() => {
+      /** ignore */
+    });
+  } finally {
+    await utils.removeLogFilePath(logFilePath);
+  }
```
sidmorizon: # 🟡 [P2] Android share target can find the log ZIP already deleted

## Problem

`exportLogs` now awaits `Share.shareAsync(...)` and unlinks the archive in `finally` (lines 64-66). On Android, expo-sharing 57.0.13 (`SharingModule.kt`) starts `Intent.createChooser(...)` with `startActivityForResult` and resolves the promise in `OnActivityResult` (lines 80-83 of the module). The chooser activity returns its result as soon as the chosen target activity is launched, not when the target has finished reading the `content://` stream. Targets that read lazily (Drive "Save to Drive", Files "Save", mail clients that attach on send) then hit `FileNotFoundException` on a file this code has already removed.

`RecoveryActivity.java` has the same new pattern: `exportArchivePendingCleanup = true` (line 207) and the deletion in `onResume` (lines 150-156) fires when the chooser returns, which is before the receiving app reads the URI.

## Impact

Log export on Android becomes nondeterministically broken depending on which app the user picks, which is the worst possible time for a support flow to fail. Before this commit the archive stayed in cache and export was reliable.

## Recommendation

On Android skip the immediate unlink and rely on the new `cleanupLogArchives` (7-day / 5-file bound in `packages/shared/src/logger/utils/index.native.ts:39-66`) plus the pre-export unlink at `index.native.ts:95`. Keep immediate deletion on iOS, where the share sheet completion callback is real. For `RecoveryActivity`, delete only on the next `onCreate` (line 142 already does this) and drop the `onResume` deletion.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,939 @@
...
+    }
+
+    private static boolean isRelevantExitReason(int reason) {
+        return reason == ApplicationExitInfo.REASON_ANR ||
+            reason == ApplicationExitInfo.REASON_CRASH ||
+            reason == ApplicationExitInfo.REASON_CRASH_NATIVE ||
+            reason == ApplicationExitInfo.REASON_EXCESSIVE_RESOURCE_USAGE ||
+            reason == ApplicationExitInfo.REASON_INITIALIZATION_FAILURE ||
+            reason == ApplicationExitInfo.REASON_LOW_MEMORY;
+    }
```
sidmorizon: # 🟡 [P2] Routine low-memory exit records still share the 5-slot "native" pool and evict real crash reports

## Problem

The latest commit isolates `sentry-js-*` into its own rotation pool, but `cleanupReports` (lines 771-780) still puts every non-JS file into one `nativeReports` list capped at `MAX_REPORT_COUNT = 5`. That list mixes real crash reports (`sentry-native-*`) with the informational `android-exit-*` records written here, and `isRelevantExitReason` (lines 311-318) accepts `REASON_LOW_MEMORY` and `REASON_EXCESSIVE_RESOURCE_USAGE`, which are routine LMK kills on Android. Up to `MAX_PROCESS_EXIT_COUNT = 10` such records can be written per launch, each stamped with the exit timestamp, so they sort newest in `rotateReports`.

iOS has the same shape: `metrickit-*` payloads (hang, CPU, disk-write diagnostics arrive daily) share the native pool with `sentry-native-*` in `OneKeyCleanupCrashReports` (`OneKeySentryCrashDiagnostics.m:506-534`).

## Impact

Five background LMK kills after a native crash rotate the crash's `sentry-native-*.json` out within the 7-day window. The pool isolation this commit claims for JS events is not provided for the reports the feature was written to preserve.

## Recommendation

Rotate per source prefix with independent caps (`sentry-native-`, `android-exit-` / `metrickit-`, `sentry-js-`) in `cleanupReports` and `OneKeyCleanupCrashReports`, or drop `REASON_LOW_MEMORY` and `REASON_EXCESSIVE_RESOURCE_USAGE` from `isRelevantExitReason` so only crash/ANR/init-failure exits compete for slots.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
```diff
@@ -0,0 +1,939 @@
...
+        }
+        return redact(String.valueOf(value));
+    }
+
+    private static void sanitizeEventForTransport(SentryEvent event) {
+        event.setUser(null);
+        event.setRequest(null);
+        event.setBreadcrumbs(null);
+        event.setExtras(null);
+        event.setTags(null);
```
sidmorizon: # 🟡 [P2] `sanitizeEventForTransport` wipes the RN SDK's `event.origin` / `event.environment` tags on every native event

## Problem

`RNSentryStart.updateWithReactFinals` (`@sentry/react-native` 8.2.0, `RNSentryStart.java:305-316`; iOS `RNSentryStart.m:128-146`) wraps the user `beforeSend`: it first calls `setEventOriginTag(event)` (tags `event.origin=android|ios`, `event.environment=native`) and then invokes our callback. `sanitizeEventForTransport` runs `event.setTags(null)` here (line 550) and `event.tags = nil` in `OneKeySentryCrashDiagnostics.m:355`, so those tags are dropped for every event captured by the native SDK (Java/NDK crashes, ANRs, app hangs, handled native exceptions).

## Impact

Sentry's React Native integration uses `event.origin` / `event.environment` to classify native-vs-JS issues; after this PR all native-captured issues lose that classification in the Sentry UI and in saved searches, which makes the native crash triage this PR is trying to improve harder rather than easier.

## Recommendation

Whitelist those two SDK-owned tags instead of clearing the whole map: read `event.getTag("event.origin")` / `event.getTag("event.environment")` before `setTags(null)` and re-apply them afterwards (same on iOS around line 355). If other tags are considered sensitive, keep the wipe for everything else.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

packages/shared/src/modules3rdParty/sentry/basicOptions.ts
```diff
@@ -252,6 +294,7 @@ export const sanitizeSentryEvent = <T extends ISentrySanitizableEvent>(
         // Sanitize error message
         if (exceptionValue.value) {
           const newErrorText = sanitizeText(exceptionValue.value);
+          sanitizeStacktrace(exceptionValue.stacktrace);
```
sidmorizon: # 🟡 [P2] Stack-frame sanitization is now skipped whenever `exception.value` is falsy (all platforms)

## Problem

This commit moved `sanitizeStacktrace(exceptionValue.stacktrace)` inside `if (exceptionValue.value) { ... }` (line 297) and removed the unconditional call that existed at the previous head (`0309cb3`, old lines 267-269, under the comment "Sanitize stacktrace (local variables, context lines)"). That comment now dangles above the `isFilterErrorAndSkipSentry` check at line 311 with no sanitization beneath it.

`sanitizeSentryEvent` is the shared `beforeSend` for every platform through `buildBasicOptions` (line 334), so desktop, web, extension and native are affected. An exception whose `value` is empty or undefined (e.g. `throw new Error()`, `throw ''`, or non-`Error` throws that the SDK renders with an empty value) now ships `frame.vars`, `context_line`, `pre_context` and `post_context` unsanitized.

## Impact

A regression of an existing privacy control on the Sentry transport path. The Electron main / node SDK enables context lines by default, so this is reachable in production desktop builds, not only in theory.

## Recommendation

Call `sanitizeStacktrace(exceptionValue.stacktrace)` unconditionally before the `if (exceptionValue.value)` block (that keeps the new "sanitize before `onError`" ordering for both branches), and move or delete the orphaned comment at line 310. A unit test in `sentry.test.ts` with `value: ''` and a frame carrying `vars` would lock this in.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

sidmorizon: 🟠 P1: Valueless exceptions bypass stack sanitization

An exception without `value` skips `sanitizeStacktrace`, even when its stacktrace exists. Sensitive frame variables and source context can reach Sentry unredacted.

packages/kit/src/provider/Container/ErrorToastContainer/ErrorToasts.tsx
huhuanming: This is intentional. Contact Support now only opens Intercom; diagnostic collection/upload remains behind the explicit Export Logs flow. Restoring the previous behavior would upload diagnostics without a dedicated user confirmation, so we are preserving this privacy boundary.

apps/mobile/android/app/build.gradle
```diff
@@ -287,6 +298,17 @@ def prepareDevShellDebugAssets = tasks.register("prepareDevShellDebugAssets") {
 }
 
 android.applicationVariants.configureEach { variant ->
+    if (variant.buildType?.name?.equalsIgnoreCase("release")) {
+        variant.preBuildProvider.configure {
+            doFirst {
+                if (sentryDsnReactNative.isEmpty()) {
```
sidmorizon: # 🟡 [P2] Release-only DSN gate breaks every existing DSN-less release build path

## Problem

This `preBuild` hook (lines 301-309) throws for every `release` build type when `SENTRY_DSN_REACT_NATIVE` is empty, and the iOS inject script does the same for any `*Release*` configuration (`apps/mobile/scripts/inject-sentry-native-config.sh:37-45`). The value is read only from `.env` / `.env.expo` / the process environment (`defEnvStr(appEnvConfig, ...)` at lines 150-151), and nothing in the repository provisions it outside CI: `.env.example:21` lists `NATIVE_SENTRY_DSN`, a different key that nothing reads, and CI writes `SENTRY_DSN_REACT_NATIVE` only through `.github/actions/shared-env/action.yml:229-251`.

Existing release-variant entry points that run without a DSN on a developer machine:

- `development/scripts/android-release-build-deploy.sh:225` (`:app:assembleProdRelease`)
- `development/scripts/ios-release-build-deploy.sh:141` and `apps/mobile/package.json:14` `ios:build` (`-configuration Release`)
- `apps/mobile/.detoxrc.js:133` (`ios.sim.release`) and `:171` (`android.emu.release` → `assembleProdRelease`)

Xcode has only `Debug` and `Release` configurations, so the script's pattern matches all of the iOS cases.

## Impact

The documented local startup-profile release builds and both Detox release configurations stop building on any machine that does not have a production DSN in `.env`, and the failure only appears at `preBuild`, after the JS bundle and native compile are already queued. CI release builds are unaffected.

## Recommendation

Scope the hard failure to real release contexts, for example only when `CI` or `EAS_BUILD` is set or `ENVIRONMENT=production`, or add an explicit opt-out such as `ONEKEY_ALLOW_MISSING_SENTRY_DSN=true` that the two dev scripts and the Detox configs pass. Apply the same condition in `inject-sentry-native-config.sh`, and replace `NATIVE_SENTRY_DSN` with `SENTRY_DSN_REACT_NATIVE=` in `.env.example` so developers know which key is required.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

packages/shared/src/modules3rdParty/sentry/index.native.ts
```diff
@@ -49,19 +53,54 @@ export const initSentry = () => {
...
+    enableNdk: true,
+    enableWatchdogTerminationTracking: false,
+    attachScreenshot: false,
+    attachViewHierarchy: false,
+    sendDefaultPii: false,
+  };
+  init({
+    ...nativeInitializationOptions,
+    ...nativeBasicOptions,
+    autoInitializeNativeSdk: false,
```
sidmorizon, line 92: # 🟡 [P2] Native options passed to the JS `init` are now inert, and the comment describing them is stale

## Problem

With `autoInitializeNativeSdk: false` hard-coded here (line 103), the RN wrapper returns early in `initNativeSdk` (`@sentry/react-native` 8.2.0 `wrapper.ts:250-257`) and never calls `RNSentry.initNativeSdk(filteredOptions)` (`wrapper.ts:310`), which is the only consumer of `nativeInitializationOptions` (lines 86-99): `maxCacheItems`, `enableAppHangTracking`, `appHangTimeoutInterval`, `enableNativeCrashHandling`, `enableNdk`, `enableWatchdogTerminationTracking`, `attachScreenshot`, `attachViewHierarchy`, `sendDefaultPii`. The effective native settings now live only in `BaseMainApplication.initializeNativeSentry` (`BaseMainApplication.java:444-466`) and `AppDelegate.initializeNativeSentry` (`AppDelegate.swift:124-140`). The comment at lines 112-115 ("enableNativeCrashHandling and enableNdk are kept enabled because…") still presents these JS values as the ones in force.

## Impact

Anyone tuning native crash handling, app-hang thresholds or PII settings here gets no effect and no warning, while the two native blocks must be kept in parity by hand. The earlier increment used these values for the real native init, so the trap is new to this refactor.

## Recommendation

Trim the JS `init` call to what the JS SDK actually consumes (`dsn`, `enabled`, `maxBreadcrumbs`, `beforeSend`, `autoInitializeNativeSdk: false`, `integrations`, `enableAutoPerformanceTracing`), delete `nativeInitializationOptions`, and move the explanatory comment next to the two native init sites with a note that they must stay in sync with each other.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
```diff
@@ -0,0 +1,511 @@
...
+    Context applicationContext = context.getApplicationContext();
+    scheduleReportCleanup(applicationContext);
+    BeforeSendCallback existingBeforeSend = options.getBeforeSend();
+    options.setBeforeSend((event, hint) -> {
+      SentryEvent preparedEvent = existingBeforeSend == null
+        ? event
+        : existingBeforeSend.execute(event, hint);
+      if (preparedEvent != null && isNativeCrash(preparedEvent)) {
+        persist(applicationContext, preparedEvent);
+      }
```
sidmorizon, line 118: 🟡 P2: Native crashes skip local diagnostics

Severity: `non-severe`

On Android, NDK crashes bypass `BeforeSendCallback`, so `persist` never writes their reports. The native SDK sends cached crash envelopes without re-entering Java. Recovery exports therefore omit these crashes.

apps/mobile/ios/OneKeyWallet/OneKeyNativeCrashDiagnostics.m
```diff
@@ -0,0 +1,580 @@
...
+                            }
+                                 error:nil];
+    NSString *path = [directoryPath stringByAppendingPathComponent:fileName];
+    if (![data writeToFile:path options:NSDataWritingAtomic error:nil]) {
+      return NO;
+    }
+    [fileManager setAttributes:@{
+      NSFileProtectionKey: NSFileProtectionComplete,
+      NSFileModificationDate: eventDate,
+    } ofItemAtPath:path error:nil];
```
sidmorizon, line 522: # 🟡 [P2] Crash report write silently fails on a locked-device background launch because it uses `NSFileProtectionComplete`

## Problem

`OneKeyWriteCrashDiagnosticsData` creates `Caches/logs/crashes` with `NSFileProtectionComplete` (lines 493-498) and then writes with `NSDataWritingAtomic` (line 500), whose temporary file is created inside that class A directory. Class A files cannot be created or written while the device is locked; only `CompleteUnlessOpen` / `CompleteUntilFirstUserAuthentication` allow creation in that state.

This hook does not run at crash time. SentryCrash processes the on-disk crash report on the **next** launch, asynchronously from its report sink, and only then reaches `beforeSend` and `OneKeyPersistNativeSentryCrashEvent` (lines 540-552). `Info.plist:118-121` declares `UIBackgroundModes = remote-notification`, so after a native crash a silent push can relaunch the app in the background while the phone is locked. In that case `writeToFile` returns `NO`, the return value is discarded by the caller (line 549), nothing is logged, Sentry still uploads the event, and SentryCrash deletes its report after sending, so there is no second chance to persist it.

## Impact

The local artifact for the most recent native crash is lost non-deterministically depending on how the app was next launched, which is exactly the case this file exists to cover. Android is unaffected because it has no equivalent protection class.

## Recommendation

Use `NSFileProtectionCompleteUntilFirstUserAuthentication` for both the directory attribute (line 496) and the file attribute (line 504); the report contains only sanitized crash metadata and the same trade-off is already accepted for the shared log export directory. At minimum log the `writeToFile` error and keep a retry path (for example re-persist from `applicationDidBecomeActive` when a pending marker exists), but since the sink deletes the report, changing the protection class is the only option that actually preserves the data.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
```diff
@@ -0,0 +1,511 @@
...
+    Pattern.compile("(?i)\\b[A-Z0-9._%+-]+@[A-Z0-9.-]+\\.[A-Z]{2,}\\b"),
+    Pattern.compile("\\b0x[0-9a-fA-F]{40,64}\\b"),
+    Pattern.compile("\\b[0-9a-fA-F]{64}\\b"),
+    Pattern.compile("\\b[5KL][1-9A-HJ-NP-Za-km-z]{50,51}\\b"),
+    Pattern.compile("\\b[xyzXYZ](?:prv|pub)[1-9A-HJ-NP-Za-km-z]{107,108}\\b"),
+    Pattern.compile("(?i)\\b(?:[a-z0-9]{1,20}1)[a-z0-9]{20,90}\\b"),
+    Pattern.compile("\\b[1-9A-HJ-NP-Za-km-z]{32,128}\\b"),
+    Pattern.compile("\\beyJ[A-Za-z0-9_-]+\\.[A-Za-z0-9_-]+\\.[A-Za-z0-9_-]+\\b"),
+    Pattern.compile("(?iu)\\b(?:mnemonic|seed(?:\\s+phrase)?)\\s*[=:]\\s*(?:\\p{L}{2,16}[\\s,]+){2,}\\p{L}{2,16}\\b"),
+    Pattern.compile("(?iu)(?:\\b\\p{L}{2,16}\\b[\\s,]+){11,}\\b\\p{L}{2,16}\\b")
```
sidmorizon: # 🟡 [P2] The generic "12+ words" regex redacts ordinary exception messages and defeats the diagnostic

## Problem

`SENSITIVE_PATTERNS` (line 80) ends with `(?iu)(?:\b\p{L}{2,16}\b[\s,]+){11,}\b\p{L}{2,16}\b` (line 90), and `redact` (lines 285-303) applies every pattern to every string in the report, including `exception.values[].value`, `message`, and breadcrumb messages. The pattern matches any run of twelve or more alphabetic words separated by spaces or commas and replaces the whole run with `[REDACTED]`. Ordinary framework messages are exactly that shape, for example `"The specified message queue synchronization barrier token has not been posted or has already been removed"` (16 words, fully redacted) or RecyclerView's `"The content of the adapter has changed but RecyclerView did not receive a notification"` (first 12 words redacted). `OneKeyNativeCrashDiagnostics.m:214` carries the identical pattern for `NSException` reasons.

This pattern was carried over from the deleted first implementation, where it was the only mnemonic defense. The new BIP39-aware sequence redaction (`redactMnemonicSequences`, lines 343-377; iOS lines 33-90) already redacts three or more dictionary words with far better precision, so the prose pattern now adds only false positives.

## Impact

Reports lose the very string a developer reads first; the stack trace survives but the message does not. It applies to Java exceptions, NDK abort messages, ObjC exception reasons and breadcrumbs alike.

## Recommendation

Remove the `{11,}` prose pattern from both `SENSITIVE_PATTERNS` arrays (Java line 90, iOS line 214) and rely on the BIP39 sequence redaction, keeping the labelled `mnemonic|seed ... :` variant on line 89 / iOS line 213. If a dictionary-less fallback is wanted for the case where the word list fails to load, gate the prose pattern on `mnemonicWords.isEmpty()` (iOS: `OneKeyMnemonicWords().count == 0`) instead of applying it unconditionally.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

packages/shared/src/modules3rdParty/sentry/index.native.ts
huhuanming: Fixed: removed the native-only cache, app-hang, NDK, and watchdog options plus the stale comment from the JS init path. attachScreenshot, attachViewHierarchy, and sendDefaultPii remain because the React Native JS client/default integrations still consume them when autoInitializeNativeSdk is false; focused tests now lock this boundary.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
huhuanming: Fixed: native Sentry configuration now queues a bounded cache/sentry envelope scan on the SDK single-thread executor before its cached-sender tasks. It parses up to 60 .envelope files, persists only crashed non-JavaScript events through the existing sanitizer/atomic writer, and leaves the SDK source envelopes untouched.

apps/mobile/ios/OneKeyWallet/RecoveryViewController.swift
```diff
@@ -506,6 +507,29 @@ final class RecoveryViewController: UIViewController {
...
+      return []
+    }
+
+    var paths: [String] = []
+    for case let fileURL as URL in enumerator {
+      let values = try fileURL.resourceValues(forKeys: [.isRegularFileKey])
+      guard values.isRegularFile == true else { continue }
+      let relativePath = String(fileURL.path.dropFirst(rootURL.path.count + 1))
+      if fileURL.pathExtension == "log" ||
+        (fileURL.pathExtension == "json" && relativePath.hasPrefix("crashes/")) {
```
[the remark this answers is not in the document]
huhuanming: Addressed by the existing cross-file cleanup path: both Android cache/onekey_logs.zip and iOS tmp/onekey-logs.zip are included in RECOVERY_LOG_ARCHIVE_PATHS and removed on the next healthy launch by the post-first-render background idle pruneStaleArtifacts call. Per the product decision, the current share artifact is intentionally not deleted immediately from the share completion callback.

apps/mobile/ios/OneKeyWallet/OneKeyNativeCrashDiagnostics.m
huhuanming: Fixed: the crash directory and report files now use NSFileProtectionCompleteUntilFirstUserAuthentication. The writer also updates the protection class of an already-existing directory for upgrades and logs a fixed, payload-free message if persistence fails.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
huhuanming: Fixed: removed the generic 12-plus-word prose regex from both Android and iOS. Mnemonic protection continues to use the bundled npm BIP39 English word list with the three-word threshold, so ordinary framework exception messages remain intact.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
huhuanming: Fixed in the current native implementation: pinHash and backendShare are treated as sensitive keys/labels on Android and iOS, alongside the existing wallet-secret filters. The sanitizer fixtures cover both fields before exportable JSON is written.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeySentryCrashDiagnosticsModule.java
huhuanming: Fixed by the simplified implementation: the removed NativeModule no longer collects ApplicationExitInfo or competes across report sources. Pending envelope processing and report cleanup/writes now serialize through the SDK executor and the same FILE_LOCK, so cleanup cannot delete an active temporary report.

packages/kit/src/views/Setting/pages/Tab/exportLogs/index.native.ts
huhuanming: Fixed: native log archives are no longer deleted by uploadLogBundle or disposeLogDigest, so the returned digest continues to reference a readable file for deferred consumers. Previous-launch archives are removed by the post-first-render background idle cleanup.

packages/shared/src/modules3rdParty/sentry/basicOptions.ts
huhuanming: Fixed: sanitizeStacktrace now runs unconditionally before checking exception.value, so frame vars and source context are sanitized even for empty or missing exception values. Focused test coverage exercises the valueless-exception case.

apps/mobile/android/app/build.gradle
huhuanming: Fixed: the missing-DSN hard failure is now scoped to EAS Release markers only on both Android and iOS, preserving local Release, profiling, and Detox builds. The example variable is also corrected to SENTRY_DSN_REACT_NATIVE, and isolated config-injection checks cover all three branches.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
```diff
@@ -0,0 +1,603 @@
...
+        Log.w(TAG, "Unable to rotate native crash diagnostics");
+      }
+    }
+  }
+
+  private static void schedulePendingNativeCrashProcessing(
+    Context context,
+    SentryAndroidOptions options
+  ) {
+    options.getExecutorService().submit(() -> {
```
sidmorizon: # 🟠 [P1] Startup task is submitted to the SDK's no-op executor and never runs

## Problem

`configure` (line 117) calls `schedulePendingNativeCrashProcessing`, which submits the cleanup plus the new envelope scan to `options.getExecutorService()` (line 593). `configure` runs inside the `RNSentrySDK.init` options callback (`BaseMainApplication.java:452-463`), and at that point the executor is still the placeholder:

- `SentryOptions.java:319` (sentry-java 8.33.0) initializes `executorService = NoOpSentryExecutorService.getInstance()`.
- The real executor is created only in `SentryOptions.activate()` (`SentryOptions.java:651-655`), which `Sentry.init` calls at `Sentry.java:332`, **after** `applyOptionsConfiguration` (`Sentry.java:241`) has run the user callback (`SentryAndroid.java:143`).
- `NoOpSentryExecutorService.submit(Runnable)` (lines 20-22) returns `new FutureTask<>(() -> null)` and never executes the runnable.

So the runnable submitted at line 593 is silently discarded on every launch.

## Impact

The "preserve cached native crash diagnostics" scan (lines 134-190) never executes, so the behaviour this commit describes does not exist at runtime. It is also a regression against the previous head: the 7-day expiry cleanup used to run on the dedicated `onekey-crash-log-cleanup` thread, and now it only runs as a side effect of `writeReport` (line 547), so expired reports linger until the next crash is written.

## Recommendation

Do not take the executor from the options object inside the configuration callback. Restore the dedicated low-priority thread for the cleanup, or register an `Integration` via `options.addIntegration(...)` and do the work in `register()`, which runs after `activate()` (`Sentry.java:374-376`). If you do keep the work on the SDK executor, note that it queues ahead of `SendCachedEnvelopeIntegration` on the same single thread, and that integration blocks the init thread on `future.get(startupCrashFlushTimeout)` when a startup-crash marker exists (`SendCachedEnvelopeIntegration.java:172-182`), so the scan's regex sanitization and fsync would eat that flush budget in exactly the startup-crash case.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
```diff
@@ -0,0 +1,603 @@
...
+      }
+      return preparedEvent;
+    });
+  }
+
+  private static boolean isNativeCrash(SentryEvent event) {
+    return event.isCrashed() && !"javascript".equalsIgnoreCase(event.getPlatform());
+  }
+
+  private static void persistPendingNativeCrashEnvelopes(Context context) {
```
sidmorizon: # 🟡 [P2] The cached-envelope scan is redundant: every event it can find is already persisted by `beforeSend`

## Problem

`persistPendingNativeCrashEnvelopes` (lines 134-190) re-reads `cache/sentry` on each launch, but neither class of crash it targets needs it:

- **Java uncaught crash.** `UncaughtExceptionHandlerIntegration` → `SentryClient.captureEvent` → `executeBeforeSend` (`SentryClient.java:163-166`) runs in-process at crash time, so the `beforeSend` hook (lines 119-127) has already written `sentry-native-<eventId>.json`. The `.envelope` later cached under `cache/sentry/<dsnHash>/` holds the same event, so the scan rewrites an identical file via `REPLACE_EXISTING` (lines 531-536).
- **NDK crash.** sentry-native-ndk 0.12.7 writes outbox files as `<outbox>/<uuid>` with no suffix (`ndk_sentry.c:238-245`), so the `.envelope` filter at line 177 never opens them; the only suffixed NDK files live transiently under `.sentry-native/<uuid>.run/` and are forwarded and deleted by `sentry__process_old_runs` during `SentryNdk.init` on the init thread. The outbox file is then processed by `OutboxSender.processEnvelope` → `scopes.captureEvent(event, hint)` (`OutboxSender.java:131-147`), and the cached-envelope hint only suppresses scope application, not `executeBeforeSend`, so the hook persists NDK crashes on the next launch already. `isNativeCrash` passes for them because sentry-native sets `mechanism.handled = false` and `platform = native`.

The 60-file cap at line 165 is also consumed in unordered `listFiles()` order, and with `setMaxCacheItems(60)` the top-level cache alone can exhaust it before `outbox/` is visited.

## Impact

Extra I/O, JSON parsing and full regex sanitization on every launch for no additional coverage, plus the executor-ordering cost described in the P1 above if the task is ever wired to a real executor.

## Recommendation

Remove `persistPendingNativeCrashEnvelopes` / `persistNativeCrashEvents` and the `EnvelopeReader`, `IEnvelopeReader`, `SentryEnvelope`, `SentryEnvelopeItem` and `SentryItemType` imports, keeping `beforeSend` as the single capture point (it covers Java, NDK-via-outbox and ANR events). If an explicit NDK regression test is wanted, assert on the outbox → `beforeSend` path rather than adding a second reader.

[Claude session ↗](https://claude.ai/code/session_014Df21yCieMCimvz97Nxbkb)

---

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
```diff
@@ -0,0 +1,603 @@
...
+      if (file.isDirectory()) {
+        persistPendingNativeCrashEnvelopes(
+          context,
+          file,
+          envelopeReader,
+          serializer,
+          depth + 1,
+          visitedEnvelopeCount
+        );
+      } else if (file.getName().endsWith(".envelope")) {
```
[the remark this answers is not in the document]
huhuanming: Fixed in 41301e59d2. I removed the custom cache scanner instead of expanding its filename matching. In the installed Sentry 8.33 SDK, OutboxSender accepts regular outbox filenames including raw NDK UUID files, replays event items through IScopes.captureEvent, and SentryClient then executes the configured beforeSend callback. Native crash JSON is therefore persisted by the SDK replay path without a competing traversal or scan cap.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
huhuanming: Fixed in 41301e59d2. The startup work no longer uses SentryOptions.getExecutorService during the options callback. The redundant envelope scan was removed, and report rotation now runs on a dedicated minimum-priority background thread that does not contend with or precede Sentry cached-event delivery.

apps/mobile/android/app/src/main/java/so/onekey/app/wallet/sentry/OneKeyNativeCrashDiagnostics.java
huhuanming: Fixed in 41301e59d2. The cached-envelope traversal and its parser imports were removed. The configured beforeSend callback is now the single persistence point for in-process Java fatal events and next-launch NDK outbox replay, matching the Sentry 8.33 OutboxSender capture path.

originalix (approved):

PR #13321, opened 2026-09-11
perf: overlap badges request with address validation in queryAddress

weatherstar:
## Summary
- Start the `/wallet/v1/account/badges` request as soon as the local validator accepts the recipient input, instead of waiting for `/wallet/v1/account/validate-address` to finish.
- Thread the in-flight badges promise into `checkAccountBadges` via a new `pendingBadges` param so the early result is reused rather than re-fetched.
- Add `ServiceAccountProfile.queryAddress.test.ts` pinning the request ordering and the cases where speculation must not fire.

## Intent & Context
Reported on Slack (#C02207B6FNC, 2026-09-11): after pasting a recipient address in the Send address picker, the input spinner stayed visible for over 1 s. The network panel showed `validate-address` (572 ms) followed by `badges` (578 ms) back to back, and the reporter asked whether the two calls could run in parallel.

## Root Cause
`ServiceAccountProfile.queryAddress` awaited `serviceValidator.validateAddress` (server round trip) before reaching `checkAccountBadges` (second server round trip). The badges call only depends on the target address, account and token, not on the server validation verdict, so the serialisation was incidental to code order rather than a data dependency. Total spinner time was the sum of both round trips plus the 300 ms UI debounce.

## Design Decisions
- **Speculate only after local validation passes.** `localValidateAddress` already runs first and is a strong predictor of the server verdict, so the only wasted badges call is local-valid + server-invalid (or prohibit-send-to-self). Garbage input never fires an extra request.
- **Explicit promise threading instead of relying on `fetchBadgesDeduped`.** The dedup map drops entries on completion, so if the badges response arrived before validation finished, a second call would re-fetch. Passing the promise directly makes reuse deterministic.
- **Name-resolution guard.** The early promise is reused only when `speculativeBadges.toAddress === resolveAddress`; if a domain resolved to a different address the original fetch path runs unchanged.
- **Error semantics preserved.** A noop `.catch` prevents an unhandled rejection when the early promise is never consumed; when it is consumed, a rejection still propagates exactly as before (UI falls back to `validStatus: 'unknown'`).
- **Out of scope (left serial on purpose):** on EVM, when badges do not report INTERACTED, `serviceHistory.fetchTransferRecipients` still runs afterwards as a third hop. Parallelising it would add one request per query for already-interacted addresses unless it gets a short per-account cache; that is a traffic trade-off to decide separately.

## Changes Detail
- `packages/kit-bg/src/services/ServiceAccountProfile.ts`: track `isLocalValid`; hoist `shouldCheckBadges` / `checkInteractionStatus`; start `fetchBadgesDeduped` speculatively; `checkAccountBadges` accepts `pendingBadges` and awaits it when provided.
- `packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts`: 5 cases — badges fires before validation settles and is reused once; speculative result dropped on `invalid`; no unhandled rejection when the unused request fails; no speculation for locally invalid input; no badges call when neither contract nor interaction checks are enabled.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Extension / Mobile / Desktop / Web (shared `kit-bg` path used by Send AddressInput, Swap incognito recipient input and BulkSend)
- **Risk Areas**: the reuse guard for name-resolved addresses; slight increase in badges traffic for addresses the server rejects but the local validator accepts.

## Test plan
- [x] `yarn jest packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts` (ordering case fails on `x`, passes with this change)
- [ ] Send flow: paste a valid EVM address, confirm in the network panel that `validate-address` and `badges` overlap and the spinner ends after the longer of the two
- [ ] Paste an ENS/domain name and a random invalid string: badges/labels still correct, no extra badges call for the invalid string
- [ ] Swap incognito recipient and BulkSend address inputs still show interaction/label badges

### DIFF
--- a/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
+++ b/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
@@ -96,8 +96,14 @@ jest.mock('../dbs/simple/simpleDb', () => ({
 
 /* eslint-disable import/first, import/order */
 import { OneKeyLocalError } from '@onekeyhq/shared/src/errors';
+import {
+  EAddressInteractionStatus,
+  EServerInteractedStatus,
+} from '@onekeyhq/shared/types/address';
 
 import ServiceAccountProfile from './ServiceAccountProfile';
+
+import type { IServerAccountBadgeResp } from '@onekeyhq/shared/types/address';
 /* eslint-enable import/first, import/order */
 
 const NETWORK_ID = 'evm--1';
@@ -126,16 +132,25 @@ async function flushMicrotasks() {
   }
 }
 
+// Node only emits `unhandledRejection` after the microtask queue drains and
+// control returns to the event loop, so assertions on it need a macrotask
+// boundary, not just microtask flushes.
+async function flushMacrotask() {
+  await new Promise<void>((resolve) => {
+    setTimeout(resolve, 0);
+  });
+}
+
 function makeService({
   localValid = true,
   badgesResp = {
-    interacted: 'true',
+    interacted: EServerInteractedStatus.TRUE,
     badges: [{ type: 'success', label: 'ok' }],
   },
   getAccount,
 }: {
   localValid?: boolean;
-  badgesResp?: Record<string, unknown>;
+  badgesResp?: Partial<IServerAccountBadgeResp>;
   getAccount?: jest.Mock;
 } = {}) {
   const validate = deferred<'valid' | 'invalid'>();
@@ -228,6 +243,9 @@ describe('ServiceAccountProfile.queryAddress request overlap', () => {
 
     expect(result.validStatus).toBe('valid');
     expect(result.addressBadges).toEqual([{ type: 'success', label: 'ok' }]);
+    expect(result.addressInteractionStatus).toBe(
+      EAddressInteractionStatus.INTERACTED,
+    );
     // The speculative response is reused; no second badges round trip.
     expect(
       clientGet.mock.calls.filter(([path]) => path === BADGES_PATH),
@@ -264,7 +282,8 @@ describe('ServiceAccountProfile.queryAddress request overlap', () => {
     await flushMicrotasks();
     validate.resolve('invalid');
     const result = await pending;
-    await flushMicrotasks();
+    // A rejection only becomes "unhandled" once Node reaches the event loop.
+    await flushMacrotask();
 
     expect(result.validStatus).toBe('invalid');
     expect(unhandled).toHaveLength(0);
--- a/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
+++ b/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
@@ -100,10 +100,10 @@ import {
   EAddressInteractionStatus,
   EServerInteractedStatus,
 } from '@onekeyhq/shared/types/address';
+import type { IServerAccountBadgeResp } from '@onekeyhq/shared/types/address';
 
 import ServiceAccountProfile from './ServiceAccountProfile';
 
-import type { IServerAccountBadgeResp } from '@onekeyhq/shared/types/address';
 /* eslint-enable import/first, import/order */
 
 const NETWORK_ID = 'evm--1';
--- a/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
+++ b/packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
@@ -1,0 +1,301 @@
+/*
+yarn jest packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
+
+Latency guard for the recipient AddressInput:
+
+`queryAddress` used to run `/wallet/v1/account/validate-address` and
+`/wallet/v1/account/badges` strictly one after the other, so the input spinner
+lasted the sum of both round trips (~1.1 s observed). Once the local validator
+accepts the input, the badges request no longer depends on the server
+validation result, so it is started speculatively and awaited later. These
+tests pin that ordering and the cases where speculation must NOT fire.
+*/
+
+// --- mocks MUST be defined before the import of ServiceAccountProfile below ---
+
+jest.mock('@onekeyhq/shared/src/background/backgroundDecorators', () => ({
+  backgroundClass: () => () => undefined,
+  backgroundMethod: () => (_t: unknown, _k: unknown, d: PropertyDescriptor) =>
+    d,
+  toastIfError: () => (_t: unknown, _k: unknown, d: PropertyDescriptor) => d,
+}));
+
+jest.mock('./ServiceBase', () => ({
+  __esModule: true,
+  default: class ServiceBase {
+    backgroundApi: any;
+
+    constructor({ backgroundApi }: { backgroundApi: any }) {
+      this.backgroundApi = backgroundApi;
+    }
+  },
+}));
+
+jest.mock('@onekeyhq/shared/src/locale', () => ({
+  ETranslations: {},
+}));
+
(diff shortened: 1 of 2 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-327cb9ca-e73f-5f41-88f1-0eafbc6713b9)

sidmorizon:
Claude review session: https://claude.ai/code/session_01M5sGsoZaNSiK9y9uUBDFZq

originalix:

Review of commit `7cd2d198a480` did not finish successfully.

packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
```diff
@@ -0,0 +1,301 @@
...
+    const { service, validate } = makeService({ getAccount });
+
+    const pending = service.queryAddress(baseArgs);
+    await flushMicrotasks();
+    validate.resolve('invalid');
+    const result = await pending;
+    await flushMicrotasks();
+
+    expect(result.validStatus).toBe('invalid');
+    expect(unhandled).toHaveLength(0);
```
sidmorizon, line 289: # 🟡 P2 Unhandled-rejection test asserts before Node can emit the event, so it can never fail

## Problem

This is the only test guarding the `promise.catch(() => undefined)` at `packages/kit-bg/src/services/ServiceAccountProfile.ts:583`, but it asserts at a point where the event it checks cannot have fired yet.

The whole flow is microtask-only: `localValidateAddress`, `getAccount`, `getClient` and `validate.promise` are all `jest.fn(async …)` / manually-resolved promises, and `flushMicrotasks()` (lines 121–127) is just `await Promise.resolve()` ×20. V8 drains the *entire* microtask queue in `runMicrotasks()` and only afterwards does Node call `processPromiseRejections()`. So line 267's flush cannot advance past the microtask checkpoint, and line 270 runs strictly before `unhandledRejection` is emitted.

Verified by executing the exact shape on Node 22 (a rejected `fetchBadgesDeduped`-style promise with no handler, 20 microtask flushes, then the assertion):

```
at assertion point (microtasks only): unhandled.length = 0   <- where line 270 asserts
after macrotask boundary:             unhandled.length = 1   <- where Node actually fires
```

The rejection itself is real — `_fetchBadgesUncached` (`ServiceAccountProfile.ts:380-385`) awaits `serviceAccount.getAccount` with no `try/catch`, and `fetchBadgesDeduped`'s `.catch` re-throws after cleaning the dedup map (lines 352-358) — so the test targets the right path, it just measures it too early. Deleting line 583 leaves this test green.

Compounding it: `afterEach` (line 210) calls `process.off` before that macrotask boundary, so the escaping rejection is not merely unobserved here — it surfaces later under jest's global handler and gets attributed to an unrelated test.

## Impact

The PR's only regression guard for the unhandled-rejection safety of the speculative fetch is inert. A future refactor that drops or relocates the `.catch` — for example storing `promise.catch(...)`'s return value instead of `promise` — ships with a green suite, while `queryAddress` runs in the `bg` runtime on iOS/Android/extension and starts emitting unhandled rejections on every locally-valid-but-server-rejected address.

## Recommendation

Add a real macrotask boundary before the assertion — replace line 267's `await flushMicrotasks();` with `await new Promise((r) => { setTimeout(r, 0); });`, or append that after it. Then confirm the test actually fails when `promise.catch(() => undefined)` at `ServiceAccountProfile.ts:583` is removed; that mutation check is what makes this case earn its name.

[Claude session ↗](https://claude.ai/code/session_01M5sGsoZaNSiK9y9uUBDFZq)

---

packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
```diff
@@ -0,0 +1,301 @@
...
+  for (let i = 0; i < 20; i += 1) {
+    // eslint-disable-next-line no-await-in-loop
+    await Promise.resolve();
+  }
+}
+
+function makeService({
+  localValid = true,
+  badgesResp = {
+    interacted: 'true',
```
sidmorizon: # 🟡 P2 Badges fixture uses an `interacted` value outside `EServerInteractedStatus`, so interaction-status reuse is never covered

## Problem

`packages/shared/types/address.ts:127-131` defines `EServerInteractedStatus { FALSE = '0', TRUE = '1', UNKNOWN = '2' }`, and `getAddressAccountBadge` maps the server field through `statusMap[interacted] ?? EAddressInteractionStatus.UNKNOWN` (`ServiceAccountProfile.ts:293-308`).

The fixture here sends `'true'`, which is not a member of that enum, so `statusMap['true']` is `undefined` and every test in this file falls through to `EAddressInteractionStatus.UNKNOWN` — never `INTERACTED`. The mismatch is invisible to `tsc` only because `badgesResp` is typed `Record<string, unknown>` (line 139) rather than `IServerAccountBadgeResp`.

Two consequences follow. First, the case named "starts the badges request before server validation settles" comments that "The speculative response is reused" but only asserts `addressBadges` (line 229); the interaction-status half of that reuse is never exercised. Second, because the status lands on `UNKNOWN`, every run also takes the EVM transfer-recipients override branch at `ServiceAccountProfile.ts:775-849` (gated on `result.addressInteractionStatus !== INTERACTED`) instead of the INTERACTED short-circuit the fixture reads as intending — so the INTERACTED path through `pendingBadges` has zero coverage.

## Impact

No false green today, since nothing asserts `addressInteractionStatus` on the success path. But the suite reads as covering interaction-status propagation when it does not, and that is the security-relevant half: `INTERACTED` is what suppresses the first-transfer warning on the Send recipient input. A regression that broke interaction status while leaving `addressBadges` intact would pass this file.

## Recommendation

Set line 132 to `interacted: EServerInteractedStatus.TRUE` (i.e. `'1'`), importing the enum from `@onekeyhq/shared/types/address`, and type `badgesResp` as `Partial<IServerAccountBadgeResp>` so a future invalid value is a compile error. Then add `expect(result.addressInteractionStatus).toBe(EAddressInteractionStatus.INTERACTED);` alongside the existing `addressBadges` assertion at line 229, which pins the speculative promise's reuse for interaction status and not only for badges.

[Claude session ↗](https://claude.ai/code/session_01M5sGsoZaNSiK9y9uUBDFZq)

---

packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
weatherstar: Fixed in e035b26eb6: replaced the trailing `flushMicrotasks()` with a `setTimeout(0)` macrotask boundary (`flushMacrotask`) before asserting on `unhandled`. Mutation-checked: with `promise.catch(() => undefined)` removed from `ServiceAccountProfile.ts`, this case now fails (`1 failed, 4 passed`); with it restored, all 5 pass.

packages/kit-bg/src/services/ServiceAccountProfile.queryAddress.test.ts
weatherstar: Fixed in e035b26eb6: fixture now sends `interacted: EServerInteractedStatus.TRUE`, `badgesResp` is typed `Partial<IServerAccountBadgeResp>` so an out-of-enum value is a compile error, and the overlap case asserts `result.addressInteractionStatus === EAddressInteractionStatus.INTERACTED` next to the `addressBadges` check, pinning interaction-status reuse of the speculative promise.

originalix:

Review of commit `e035b26eb611` did not finish successfully.

originalix:

Review of commit `79825d876c44` did not finish successfully.

originalix (approved):

PR #13322, opened 2026-09-11
feat: expand a borrow position card in place on phones

labrinyang:
## What

Makes a supplied or borrowed position on phones tap-to-expand, revealing its actions inline instead of sending the user to a detail page for every operation.

The card is the pointer and keyboard target end to end; the collateral switch inside it stops the press from reaching the card. Chevron rotation follows the disclosure-triangle convention and is gated on `useReducedMotion`.

## Why

On phones the position rows had no room for action buttons, so every supply/withdraw/borrow/repay started with a navigation step. Expanding in place removes it.

## Notes for review

**Rebased onto current `x`.** Upstream #13284 changed `hasPendingSetCollateral` to reserve-scoped locking and updated its own call site in `CollateralSwitchCell.tsx`. This branch's only change to that file is a comment; the edits land in different regions and upstream's 7 `hasPendingSetCollateral` tests and 2 `CollateralSwitchCell` tests pass unchanged.

**`getCollateralCellState`** is new and distinguishes "off, but you can turn it on" from "this market never accepts this asset as collateral" — both previously rendered as the same faded OFF switch. Its visibility gate agrees with the existing `getCollateralSwitchState.render` across all nine input combinations.

**Test harness caveat, stated rather than fixed.** `jest.config.js` maps `@onekeyhq/components` to a module-wide mock for every suite in this repo, so no test here renders real Tamagui and no DOM assertion proves the real stack forwards `role`/`tabIndex`/`aria-*`. The card's stub now forwards every `aria-*` it is given rather than a whitelist, and `ariaForwarding.test.ts` pins the assumption those stubs stand on — that react-native-web still forwards these attributes. Removing the global mock is separate infrastructure work.

## Independence

Split out of a combined branch alongside #13323 (e-mode category picker). The two share no file and no symbol; either can merge first. Verified by building both from the merge base and confirming the test counts add up exactly (baseline 347 → 397 here, 401 there, 451 merged).

## Not covered

- Android + TalkBack card expansion was never verified on a device.
- Wide-window / iPad footer collapse was never verified.

## Test plan

- [ ] Phone: tap a supplied position, confirm actions appear and a second tap collapses it
- [ ] Phone: tap the collateral switch inside an expanded card, confirm the card does not collapse
- [ ] Android + TalkBack: expand and collapse a card via the activate action
- [ ] Reduced motion on: confirm the chevron changes state without animating
- [ ] Desktop table: confirm the collateral switch still hits and does not overhang the next column

---

https://claude.ai/code/session_016ehg7sC5ui1wKPLfT6eKVp


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

Review of commit `9b2f60ebea7c` did not finish successfully.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-03ca1b87-e23e-5182-81e9-4a21e87e55f7)

sidmorizon:
Claude review session: https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -78,10 +125,17 @@ export function BorrowMobilePositions({
   const navigation = useAppNavigation();
   const { reserves, market, borrowDataStatus, earnAccount } =
     useBorrowContext();
+  // One card at a time: tapping another position replaces the open one.
+  const [expandedKey, setExpandedKey] = useState<string | null>(null);
+
+  const toggleExpanded = useCallback((key: string) => {
+    setExpandedKey((current) => (current === key ? null : key));
```
sidmorizon, line 128: 🟡 P2: Provider switches preserve expanded cards

Severity: `non-severe`

Switching providers with the same network and market address leaves `expandedKey` selected. The matching card opens automatically in the new provider.

Suggested fix:

```
  // Scoped keys hide the old card; clear the selection so returning stays collapsed.
  useEffect(() => {
    setExpandedKey(null);
  }, [accountId, networkId, normalizedMarketAddress, market?.provider]);
```

originalix:

Review of commit `a12a2d881efb` did not finish successfully.

packages/kit/src/views/Borrow/components/BorrowTableList/CollateralBadge.tsx
```diff
@@ -8,17 +8,22 @@ import { collateralBadgeVariant } from '../collateralControls.utils';
...
+  bg?: IIconProps['bg'];
 }) {
   const variant = collateralBadgeVariant(canBeCollateral);
   if (!variant) return null;
   const can = variant === 'can';
   return (
     <Icon
       br="$1"
-      bg={can ? '$bgSuccess' : '$bgSubdued'}
+      bg={can ? '$bgSuccess' : bg}
```
sidmorizon: # 🟡 P2 New `bg` prop is silently ignored for the `can` variant

## Problem

This PR adds a public `bg` prop to the shared `CollateralBadge` — declared at `CollateralBadge.tsx:18` as `bg?: IIconProps['bg']`, defaulted at line 14 to `'$bgSubdued'`, and documented by the new comment at lines 11–13 as the escape hatch for callers whose surface is itself `$bgSubdued`.

But line 26 only consults it on one branch:

```tsx
bg={can ? '$bgSuccess' : bg}
```

The `cannot` chip honours the caller. The `can` chip is hard-wired to `$bgSuccess` and discards the prop entirely. The name and the type promise control over the chip fill; the implementation delivers it for one of the two rendered states.

This is latent today — the only caller that passes `bg` is `BorrowMobilePositions.tsx:297`, which hard-codes `canBeCollateral={false}`, so the unreachable branch is never taken. I'm raising it as an API-shape defect in code this PR introduces, not as a present visual bug.

## Impact

`CollateralBadge` is a three-state component (`collateralBadgeVariant` in `collateralControls.utils.ts:24-30` returns `'can' | 'cannot' | null`), and two existing call sites render the positive state:

- `BorrowTableList/AssetField.tsx:41`
- `BorrowTableList/AssetWithAmountField.tsx:40`

The next caller that renders a badge on a non-default surface — exactly the contrast problem the new comment describes, just for the green chip rather than the neutral one — will pass `bg`, get no effect for eligible reserves, and see no type error and no warning. A silent contrast-only failure on one of two states is hard to catch in review and invisible in a screenshot that happens to show the other state.

## Recommendation

Either narrow the name to what it does, or make it do what the name says.

Option A — scope the prop to its one state (smaller change, matches the comment's actual intent):

```tsx
unavailableBg = '$bgSubdued',
…
bg={can ? '$bgSuccess' : unavailableBg}
```

Option B — let the caller override both, preserving today's defaults:

```tsx
bg,
…
bg={bg ?? (can ? '$bgSuccess' : '$bgSubdued')}
```

Option A leaves `BorrowMobilePositions.tsx:297` needing only a prop rename. Option B keeps that call site's current behaviour unchanged but means `bg="$bgStrong"` would also restyle the green chip if `canBeCollateral` ever became dynamic there.

[Claude session ↗](https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena)

---

sidmorizon (commented):
Automated review — see inline comment.

---

originalix:

Review of commit `0155f035146b` did not finish successfully.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -50,41 +53,94 @@ function buildDisabledByReserve<T extends { reserveAddress: string }>(
...
   const networkId = market?.networkId ?? '';
+  const marketAddress = market?.marketAddress ?? '';
+  const normalizedMarketAddress = marketAddress.toLowerCase();
   const eModeId = eModeStatus?.eModeId;
   const hasCollateralControls = Boolean(market && accountId);
 
+  // Scoped keys hide the old card; clear the selection so returning stays collapsed.
+  useEffect(() => {
+    setExpandedKey(null);
+  }, [accountId, networkId, normalizedMarketAddress]);
```
sidmorizon, line 164: # 🟠 P1 A transient blank `accountId` permanently collapses the user's open card

## Problem

`accountId` is `earnAccount.data?.account?.id || ''` (line 131), so it becomes `''` whenever the context publishes `data: null`. This new effect treats *every* `accountId` transition as a scope change — including `X → '' → X`, which is a **loading** transition, not a scope change.

That transition is reachable with `market` completely unchanged:

1. `BorrowDataGate.tsx:404-419` publishes `data: scopedEarnAccountData ?? null` with no keep-previous guard. Its sibling `loading` expression (`hasAccountContext && marketNetworkId && scopedEarnAccountData === undefined`) exists precisely because `scopedEarnAccountData` can be `undefined` while account context is live — the gate already anticipates this state.
2. `useEarnAccount.ts:144-148` subscribes to `GlobalDeriveTypeUpdate` / `NetworkDeriveTypeChanged` and calls `refreshNetworkDeriveType({ alwaysSetState: true })`. Its own comment states the intent: *"Clear the previous derive scope before resolving the new authoritative value."*
3. That result is configured `undefinedResultIfReRun: true` (`useEarnAccount.ts:79`), so `usePromiseResult.ts:279-281` runs `setResult(undefined)` → `networkDeriveType` undefined → `hasResolvedAccountScope` false → `swrKey` undefined (`useEarnAccount.ts:92-101`).
4. `usePromiseResult.ts:148-151` then takes the `key→undefined` branch: `setResult(options.initResult)`, and `useEarnAccount` passes no `initResult`, so `earnAccount` becomes `undefined`.
5. Gate publishes `data: null` → `accountId === ''` → **this effect fires and clears `expandedKey`**. The derive type then resolves back to the same value (these events are app-global and carry no network payload) and `accountId` returns byte-identical.

The component stays mounted throughout, so the state loss is real rather than incidental: `useBorrowPositionEntries` (`hooks/useBorrowPositionEntries.ts:8-19`) depends only on `reserves`, never on `earnAccount`, so `hasPositions` at `BorrowHome.tsx:273` stays true and the render guard at `BorrowHome.tsx:433` keeps this component alive across the blank.

A second, independent route: a cancelled market switch. `BorrowDataGate.tsx:100-104` nulls `requestedMarket` when the target disappears from a refreshed market list, and the `earnAccountResult.networkId === networkId` guard (`useEarnAccount.ts:171-174`) fails for a frame — again publishing `data: null` with `networkId` / `normalizedMarketAddress` constant.

Ordinary refetches are *not* affected: the earnAccount `usePromiseResult` sets `undefinedResultIfReRun: false` (`useEarnAccount.ts:133`), so polling, `refresh()`, focus revalidation and reconnect all retain the previous value. The blank specifically requires the `swrKey → undefined` reset or the network-guard mismatch above.

## Impact

This is a regression introduced by this commit, not a pre-existing issue. Before the effect existed, the same blank made the key merely fail to match for those frames — the card rendered collapsed and then **re-expanded** when `accountId` returned, because `expandedKey` was never cleared. The effect converts a self-healing flicker into permanent loss.

A user with a position card open on a phone loses that expansion with no input from them, triggered by a background event unrelated to the borrow scope — a derive-type change made anywhere in the app (including on another network, while Borrow is mounted but blurred, since `alwaysSetState` defeats the focus gate), or a market-list refresh that cancels an in-flight market switch. They must tap again to recover. It is the same class of bug the effect was written to fix, in the opposite direction.

## Recommendation

Distinguish "scope changed" from "scope not yet resolved". The gate already publishes `loading` for exactly these blanks, so the smallest correct fix is to make the reset ignore unresolved scopes rather than react to them:

```tsx
useEffect(() => {
  if (!accountId && earnAccount.loading) return;
  setExpandedKey(null);
}, [accountId, networkId, normalizedMarketAddress, earnAccount.loading]);
```

Alternatively, hold the last **non-blank** scope tuple in a ref and clear `expandedKey` only when a new non-blank tuple differs from it — that also handles `X → '' → Y`, where the reset genuinely is required.

Either way, please add a case beside the new `it.each` at `BorrowMobilePositions.test.tsx:402-435` that rerenders with `scope.accountId = ''` and then back to `'account-1'`, asserting the card is **still expanded**. Every one of the added +42 lines round-trips through populated scopes only, so this exact transition is the gap in the new coverage.

[Claude session ↗](https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena)

---

sidmorizon (commented):
Automated review — see inline comment.

---

originalix (dismissed):

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
labrinyang: Not reachable in practice: a pool address on a given network belongs to exactly one protocol, so no market list can contain two providers that share the same `networkId` and `marketAddress`. Any real provider switch changes the network or the address, and both of those already reset the expansion.

Leaving the key as is; resolving.

labrinyang:
Rebased onto `x` (389d2ff49a). All four conflicts come from #13353:

- `BorrowMobilePositions.tsx`, `BorrowPositionCard.tsx`: kept this branch's layout. #13353's horizontal wrapper and dropped negative margin are already achieved by the card's structure here, and its actions-row margin has nothing left to separate, because actions sit on their own line when the card is expanded.
- `CollateralSwitchCell.tsx`: took #13353's press handler (no handler on native, `stopPropagation` only on web) and kept this branch's comment on why the wrapper has no padded halo. The branch's concern still holds: on web `stopPropagation` keeps the row press from firing, and on native the switch is its own touch responder.
- `CollateralSwitchCell.test.tsx`: kept #13353's two handler tests plus this branch's halo test. Dropped this branch's "swallows the press" test: it asserted the `preventDefault` call that #13353 removed on purpose, and the desktop propagation test covers the same protection.

222 tests across 20 suites and `agent:check --profile pr` pass.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -172,104 +252,149 @@ export function BorrowMobilePositions({
           reserveAddress: entry.asset.reserveAddress,
         });
 
+        const positionKey = getPositionKey({
+          kind: entry.kind,
+          accountId,
+          networkId,
```
sidmorizon, line 258: 🟡 P2: Loading frames collapse expanded positions

Severity: `non-severe`

When `accountId` temporarily clears, `getPositionKey` assigns the visible position a new key and renders it collapsed. The actions disappear and reappear when the same account resolves.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -49,41 +53,116 @@ function buildDisabledByReserve<T extends { reserveAddress: string }>(
...
+  accountId: string;
+  networkId: string;
+  marketAddress: string;
+  reserveKey: string;
+}): string {
+  return [
+    kind,
+    accountId,
+    networkId,
+    marketAddress.toLowerCase(),
```
sidmorizon, line 110: 🟡 P2: Solana markets share expansion state

Severity: `non-severe`

On Solana, `getPositionKey` lowercases case-sensitive market addresses, so distinct markets can receive the same expansion key. The scope reset also compares the lowercased value. Switching between colliding markets can leave the next market's matching position expanded.

packages/kit/src/views/Borrow/components/CollateralSwitchCell.tsx
```diff
@@ -666,6 +666,14 @@ export function CollateralSwitchCell({
...
       ai="center"
       jc="center"
+      // No padded halo here. Neither platform needs one: on web, desktop and
+      // the extension ESwitchSize.small renders a 38x24 Tamagui track, at the
+      // 24x24 floor of WCAG 2.5.8, and on native the Switch delegates to the
+      // platform control, which is larger again. Growing it with padding and a
+      // negative margin also put the halo outside this view's parent, where
+      // Android's ViewGroup never hit-tests and hitSlop is ignored, while on
+      // web it swallowed the desktop table's row press and overhung the next
+      // column.
```
sidmorizon, line 676: # 🟡 P2 The added comment is wrong on iOS, next to the one invariant the rebase left unasserted

## Problem

This comment — the PR's only change to this file — states that *"on native the Switch delegates to the platform control, which is larger again"* (lines 671-672). That is true on Android but false on iOS.

Fifteen lines below, line 691 reads:

```tsx
native={!platformEnv.isNativeIOS}
```

and `Switch/index.tsx:143` spreads `{...restProps}` *after* the hard-coded `native` at line 130, so the caller's `native={false}` wins. On iOS this renders the same 38×24 Tamagui track as web — which the existing comment at line 686 already says outright: *"The shared press-based switch avoids native row hit-testing issues on iOS."* The new comment directly contradicts its own neighbour.

The same incorrect sentence is duplicated at `CollateralSwitchCell.test.tsx:356-360` (*"native hands off to the platform control, which is larger"*).

This matters more than a stray comment usually would, because of what the rebase changed around it. The wrapper's press guard is now `platformEnv.isNative ? undefined : …stopPropagation()` (lines 677-683), so on iOS and Android **nothing here stops the press** — the only thing keeping a switch tap from also collapsing the card is responder negotiation, i.e. that the switch claims the touch before the card's `onPress` (`BorrowPositionCard.tsx:138`) can fire. On iOS that claim depends specifically on the press-based Tamagui switch this comment denies exists.

And that invariant is no longer asserted anywhere. The rebase dropped `swallows the press so the surrounding card or row never fires` (present at `563c08d9`). What replaced it covers web only: `CollateralSwitchCell.test.tsx:330` pins `platformEnv.isNative = false`, and `:309` asserts only that native has *no* handler. Nothing can assert the native path — `jest.config.js:60` maps `@onekeyhq/components` to a module-wide mock, as this PR's own `ariaForwarding.test.ts:1-5` states. `BorrowPositionCard.test.tsx:322-323` still reads *"the real swallow lives in CollateralSwitchCell and is covered there"*, which no longer holds for native.

## Impact

iOS and Android only; no user-visible defect today. I traced the responder path and found no propagation bug: on Android RN's platform `Switch` sets `onStartShouldSetResponder` unconditionally, and on iOS the Tamagui press-based switch claims the responder as a descendant (the card's `YStack` uses no capture-phase handler, so it cannot pre-empt a child — cf. the same mechanism documented at `ListItem/index.tsx:366-370`).

The cost is to the next change. A future edit to line 691, or a Tamagui upgrade altering press attachment, would break the behaviour silently — no test covers it, and the comment a maintainer would reason from points at the wrong switch on the platform where the reasoning is most delicate.

## Recommendation

- Correct lines 671-672 to separate the two native targets: Android delegates to the platform control; iOS renders the same 38×24 Tamagui track and relies on the explicit `hitSlop` already set at line 708. Apply the same fix to the duplicated text at `CollateralSwitchCell.test.tsx:356-360`.
- Realign `BorrowPositionCard.test.tsx:322-323` with what actually remains covered after the rebase.
- Since the native path cannot be unit-asserted, add it to the test plan explicitly: on both iOS and Android, tap the collateral switch on an expanded card and confirm the card does not collapse.

One thing I could not settle from source, offered as uncertainty rather than a claim: whether `@tamagui/web` strips press handlers when `disabled` is true. If it does, a tap on a **disabled** iOS switch (`isSwitchDisabled`, lines 662-666) would fall through to the card and toggle it. Android is unaffected either way. Worth covering in the same device check.

[Claude session ↗](https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena)

---

sidmorizon (commented):
Automated review — see inline comment.

---

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -172,104 +252,149 @@ export function BorrowMobilePositions({
...
+                </SizableText>
+                {collateralState === 'unavailable' ? (
+                  // Match the unavailable mark used by the desktop table.
+                  <Stack
+                    testID={BorrowTestIDs.positionCardCollateralUnavailable(
+                      suppliedAsset.reserveAddress,
+                    )}
+                    accessible
+                    accessibilityRole="text"
+                    accessibilityLabel={[
```
labrinyang, line 319: **[P1] The unavailable-collateral mark has no accessible name on web, desktop and the extension**

This `Stack` only carries the React Native props (`accessible`, `accessibilityRole`, `accessibilityLabel`). Tamagui 2.7.7 does not map those on web: `@tamagui/web` ships an empty `accessibilityDirectMap` and renders stacks straight to a `div`, so the element comes out as `<div accessibilityRole="text" accessibilityLabel="…">` with no `role` and no `aria-label`. The mark is silent on screen by design, so on those targets a screen-reader user gets nothing at all. This list is not native-only: `BorrowHome` renders it whenever `!gtMd`, which includes the extension popup and narrow web/desktop windows.

`BorrowMobilePositions.test.tsx` passes only because `MockAccessibleStack` rewrites `accessibilityLabel` into `aria-label`, so the "spells the state out for screen readers" case pins the mock rather than the component.

Suggested fix: mirror the disclosure row's browser twins (a bare `div` ignores `aria-label` under ARIA naming rules, so it needs a role):

```tsx
const collateralUnavailableLabel = [
  suppliedAsset.token.symbol,
  labels.collateral,
  labels.collateralNotAvailable,
].join(', ');
// …
<Stack
  testID={BorrowTestIDs.positionCardCollateralUnavailable(
    suppliedAsset.reserveAddress,
  )}
  accessible
  accessibilityRole="text"
  accessibilityLabel={collateralUnavailableLabel}
  {...(platformEnv.isRuntimeBrowser
    ? { role: 'img' as const, 'aria-label': collateralUnavailableLabel }
    : {})}
>
```

and switch the suite's `Stack` mock to a prop-forwarding one (like the card suite's `asDom`) so the test asserts what the component actually emits. `CapabilityBadge` in `EModeAssetsTable.tsx` has the same gap, but that one predates this PR.

_— Auto-review by Claude, cross-validated with Codex_

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
labrinyang: Fixed in a3984a3008.

The key no longer carries the account, network or market. That duplicated what the scope reset already owns, and duplicated it without the reset's one piece of judgement: that a blank account is a loading frame rather than a different account. Keying on the position alone leaves the reset as the single owner of scope invalidation.

Your description was exact, including the part the earlier test comment glossed as a self-healing flicker. It was not only a collapse: the new key remounted the card, so the chevron animation restarted too. A test now pins node identity and `data-expanded` during the blank frame itself, not just after the round trip:

```
it('keeps the card mounted and open through the blank frame itself', ...)
```

It fails against the previous key.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
labrinyang: Fixed in a3984a3008.

Two parts to it. The expansion key no longer carries the market address at all, and the scope reset now compares through `earnUtils.normalizeBorrowAddress`, which folds case on EVM only — the same helper the reserve key already used one line below, so the raw `toLowerCase()` was the odd one out.

The suite could not have caught this: its `earnUtils` stub lowercased unconditionally, ignoring `networkId`, so it folded the two addresses together on the component's behalf. The stub now mirrors the real helper's contract, and a test switches between two base58 markets differing only in case:

```
it('treats case-differing Solana markets as separate scopes', ...)
```

It fails against the previous code, and it also failed against the fixed component while the stub was still unconditional.

packages/kit/src/views/Borrow/components/CollateralSwitchCell.tsx
labrinyang: Correct on every point. Fixed in 872459278d.

I confirmed the mechanism: `Switch` hard-codes `native` and then spreads `{...restProps}` after it (`forms/Switch/index.tsx:130,143`), so the `native={!platformEnv.isNativeIOS}` passed here wins and iOS renders the same 38x24 Tamagui track as web. The comment now separates the targets — web, extension and iOS on the track, iOS reaching the WCAG floor through the hitSlop set beside it, Android alone delegating to the platform control — and `CollateralSwitchCell.test.tsx` carries the corrected text.

`BorrowPositionCard.test.tsx` no longer claims the native swallow is covered elsewhere. It now says what you established: the path is unreachable in any suite, because `jest.config.js` maps `@onekeyhq/components` to a module-wide mock, so the switch that has to claim the touch is never the real one. That makes it a device check.

On the uncertainty you flagged — Tamagui does strip press handlers when disabled. `createComponent.cjs:773` gates every attached event on `shouldAttach = !disabled && ...`. So your hypothesis holds: on iOS a tap on a *disabled* collateral switch attaches no responder and falls through to the card, toggling it. Web is unaffected, since the wrapper's `stopPropagation` branch runs there regardless of the switch's state, and Android's platform control is unaffected too.

I have left that one alone rather than guess at a fix I cannot exercise here. The candidate is to give the wrapper a press handler on iOS when, and only when, the switch is disabled — there is no switch press handling to interfere with in that state, which is what made the native branch `undefined` in the first place. Both it and the enabled path belong in the device check you outlined:

- iOS and Android, expanded card, tap an **enabled** collateral switch: the switch toggles, the card does not collapse.
- iOS, expanded card, tap a **disabled** collateral switch: expect the card to collapse today.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
labrinyang: Fixed in 0e1087ca52.

The mark now ships `role="img"` and `aria-label` under `platformEnv.isRuntimeBrowser`, keeping the React Native props for native.

The suite was the more useful half of this. Its `Stack` mock rewrote `accessibilityLabel` into `aria-label`, so the assertion measured the mock rather than the component. It now forwards `role` and `aria-*` verbatim and translates nothing, and a second case pins that native still gets neither DOM prop. The web case fails against the previous component.

`CapabilityBadge` in `EModeAssetsTable.tsx` has the same gap and predates this branch, so I have left it out of this PR.

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -172,104 +250,157 @@ export function BorrowMobilePositions({
...
+            fiatValue={amount.description}
+            apyDetail={entry.asset.apyDetail}
+            statusLabel={labels[entry.kind]}
+            statusBadgeType={entry.kind === 'supplied' ? 'success' : 'critical'}
+            platformBonusApy={entry.asset.platformBonusApy}
+            collateral={collateral}
             actions={actions}
+            isExpanded={expandedKey === positionKey}
+            onToggleExpand={() => toggleExpanded(positionKey)}
           />
```
sidmorizon, line 446: 🟡 P2: Recreated position opens automatically

Severity: `non-severe`

When an expanded position disappears, `expandedKey` retains its key. Recreating that position in the same scope exposes its actions automatically.

packages/kit/src/views/Borrow/components/CollateralSwitchCell.tsx
```diff
@@ -666,6 +666,16 @@ export function CollateralSwitchCell({
...
+      // No padded halo here, and each target reaches the 24x24 floor of WCAG
+      // 2.5.8 its own way. Web, desktop, the extension and iOS all render the
+      // same 38x24 Tamagui track, because native={!isNativeIOS} below lands
+      // after the Switch's own hard-coded native and wins; iOS adds the
+      // hitSlop set alongside it. Android is the one target that delegates to
+      // the platform control, which is larger again. Growing the target here
+      // with padding and a negative margin also put the halo outside this
+      // view's parent, where Android's ViewGroup never hit-tests and hitSlop
+      // is ignored, while on web it swallowed the desktop table's row press
+      // and overhung the next column.
```
sidmorizon, line 685: # 🟠 P1 On iOS, tapping a *disabled* collateral switch toggles the card's expansion

(Anchored on the comment block because that is this PR's own line here; the defect is in the `onPress` at lines 679-685 just below.)

Filing the deferred item from the `676` thread as its own finding, now that it is confirmed rather than a hypothesis. I verified the Tamagui half independently against the pinned package rather than relying on the citation in that thread — and it holds, including for the **native** build.

## Problem

**1. On iOS the switch is a Tamagui frame, and it receives `disabled`.**
Line 693 is `native={!platformEnv.isNativeIOS}`; `Switch/index.tsx:143` spreads `{...restProps}` after the hard-coded `native` at `:130`, so `false` wins on iOS. In `@tamagui/switch@2.7.7` (the `yarn.lock` pin), `src/createSwitch.tsx:129-140` returns `null` from `useSwitchNative` when the platform check fails, so the Tamagui `Frame` renders, and `:159` passes `disabled` to it explicitly.

**2. `disabled` removes the press handlers *and* the responder claim.**
In `@tamagui/web@2.7.7`, `dist/cjs/createComponent.native.js:754` — the native build, not the web one:

```js
var shouldAttach = !disabled && !props.asChild && Boolean(attachFocus || attachPress || attachHover || ...);
var events = shouldAttach ? { ... } : null;   // :769
```

With `events === null`, `src/helpers/mainThreadPressEvents.native.ts:48` returns before writing any responder props, and `wrapWithGestureDetector` returns the content unwrapped — skipping the `cloneElement(content, { onStartShouldSetResponder: responderClaim })` that file documents as what stops a parent Pressable from firing. The switch's only press handler comes from `@tamagui/switch-headless`'s `useSwitch` native branch, which is exactly what `shouldAttach` gates off.

For the repair: `useComponentState.ts:312-320` derives `isDisabled` from `props.disabled || props.accessibilityState?.disabled || props['aria-disabled'] || …`. This cell sets **both** `disabled={isSwitchDisabled}` (line 694) and `accessibilityState: { disabled: isSwitchDisabled }` (lines 702-705), so dropping one does not avoid the gate.

**3. Nothing in between intercepts.** The wrapper at lines 679-685 is `onPress={platformEnv.isNative ? undefined : …}` — no handler and no `pressStyle` on native, so `shouldAttach` is false for it too. The opacity wrapper (line 687) has no handlers, the spinner overlay carries `pointerEvents="none"`, and the `hitSlop` at line 709 is irrelevant with no responder attached.

**4. The card claims the touch instead.** `BorrowPositionCard.tsx:142` puts `onPress={onToggleExpand}` on the outer `YStack` wrapping this row, and `BorrowMobilePositions.tsx:403` always passes `onToggleExpand`, so `isPressable` is always true on phones.

**5. The disabled state is routine, not exotic.** `isSwitchDisabled` (lines 658-662) is `previewLoading || isNativeActionUnsupported || disabled || (!value && requiresEModeId && eModeId === undefined)`, and `disabled` from `getCollateralSwitchState` (`collateralControls.utils.ts:46-51`) adds `submitting || pendingSetCollateral || (usageAsCollateral === false && canBeCollateral !== true)`. Because `getCollateralCellState` only diverts to the dash mark when `canBeCollateral === false`, the switch renders *and* is disabled whenever the server omits the eligibility flag — the shape this PR's own test `keeps the switch when the eligibility flag is missing` pins. `previewLoading`, `submitting` and `pendingSetCollateral` also mean **the ordinary toggle flow** sits in this state for seconds at a time.

## Impact

iOS only. During and after any collateral toggle — and permanently for reserves whose eligibility flag is absent — tapping the switch expands or collapses the card instead of doing nothing. It is worst right after the user confirms a collateral change: the switch locks while settling, the user taps it again, and the card jumps open or shut with its action buttons, reading as though the tap changed something about the position.

## Recommendation

Give the wrapper a real press handler on native precisely when the inner switch has none, so the wrapper becomes the responder claimant:

```tsx
onPress={
  platformEnv.isNative
    ? isSwitchDisabled
      ? () => {}
      : undefined
    : (e) => {
        e.stopPropagation();
      }
}
```

Conditioning on `isSwitchDisabled` leaves the enabled path byte-identical to today — the enabled switch still wins arbitration as the deeper claimant — so it cannot regress the working case. The comment above would need a line about it, since it currently explains only the halo.

One scope caveat, stated rather than glossed: I verified **iOS**. On Android `native` is `true`, so RN's platform `Switch` renders, and whether a *disabled* Android `Switch` still cancels the JS responder depends on RN internals I could not read here. I would not assert Android either way — which is an argument for gating the fix on `platformEnv.isNative` as above rather than `isNativeIOS`, since that makes the question moot at no cost.

[Claude session ↗](https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena)

---

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -172,104 +250,157 @@ export function BorrowMobilePositions({
...
+                </SizableText>
+                {collateralState === 'unavailable' ? (
+                  // Match the unavailable mark used by the desktop table.
+                  <Stack
+                    testID={BorrowTestIDs.positionCardCollateralUnavailable(
+                      suppliedAsset.reserveAddress,
+                    )}
+                    accessible
+                    accessibilityRole="text"
+                    accessibilityLabel={unavailableLabel}
```
sidmorizon: # 🟡 P2 The unavailable mark still ships React Native a11y props to the web DOM

## Problem

`0c76d8c8` made the card's RN a11y props native-only, and `0e1087ca` gave this mark its browser twins at lines 327-329 — but left the three RN props unconditional:

```tsx
accessible
accessibilityRole="text"
accessibilityLabel={unavailableLabel}
...
{...(platformEnv.isRuntimeBrowser
  ? { role: 'img' as const, 'aria-label': unavailableLabel }
  : {})}
```

The premise the sibling commit relied on cuts both ways. In `@tamagui/web@2.7.7`, `src/helpers/skipProps.ts:31-33` merges `nativeOnlyProps` into `skipProps` only when `TAMAGUI_TARGET === 'web'`, and `src/helpers/nativeOnlyProps.ts` lists `onAccessibilityAction`, `onAccessibilityTap` and `hitSlop` — **not** `accessible`, `accessibilityRole` or `accessibilityLabel`. `accessibilityDirectMap` is `{}` in the web build, and OneKey's `Stack` on web is `styled(View)` from `@tamagui/core` (`packages/components/src/shared/stacks.web.ts`), i.e. a plain `div`, not an RNW `View` — so nothing filters them either.

So on web, desktop and the extension this `div` receives all three RN props *in addition to* `role="img"` and `aria-label`. `accessible={true}` triggers React's "Received `true` for a non-boolean attribute" warning; the other two produce unknown-prop casing warnings and land in the shipped DOM as `accessibilityrole="text"` / `accessibilitylabel="…"`. The stray `accessibilityRole="text"` also disagrees with the `role="img"` that actually takes effect.

The new test `leaves the DOM-only props off the mark on native` only asserts the browser props are absent on native, and since `MockAccessibleStack` forwards only `aria-*`/`role`, no test can observe the leftovers on web. Unfixed and unobservable.

## Impact

Web, desktop and browser extension. Three React console warnings per rendered unavailable mark — so N per list — plus junk attributes in the production DOM. No screen-reader breakage: `role="img"` + `aria-label` do the real work. Cosmetic, but it is exactly the class of defect `0c76d8c8` was written to eliminate, left asymmetric within the same PR.

## Recommendation

Mirror what the card now does and make the two sides exclusive:

```tsx
{...(platformEnv.isRuntimeBrowser
  ? { role: 'img' as const, 'aria-label': unavailableLabel }
  : {
      accessible: true,
      accessibilityRole: 'text' as const,
      accessibilityLabel: unavailableLabel,
    })}
```

To make it assertable, either widen `MockAccessibleStack`'s filter to forward `data-*`-prefixed copies of the RN props, or park the raw prop set on `globalThis` the way `BorrowPositionCard.test.tsx:58-61` already does, then assert `accessibilityLabel` is undefined in the browser case.

[Claude session ↗](https://claude.ai/code/session_01DRdoZd77yDRD7N6Wruxena)

---

sidmorizon (commented):
Automated review — see inline comments.

---

packages/kit/src/views/Borrow/components/BorrowMobilePositions.tsx
```diff
@@ -49,41 +54,113 @@ function buildDisabledByReserve<T extends { reserveAddress: string }>(
...
+    const resolvedScope = [accountId, networkId, normalizedMarketAddress].join(
+      '|',
+    );
+    const previousScope = lastResolvedScopeRef.current;
+    lastResolvedScopeRef.current = resolvedScope;
+    if (previousScope !== null && previousScope !== resolvedScope) {
+      setExpandedKey(null);
+    }
+  }, [accountId, networkId, normalizedMarketAddress]);
+
```
sidmorizon: 🟡 P2: New scope briefly inherits expansion

Severity: `non-severe`

A scope change with the same `positionKey` renders the new card expanded. The passive reset runs after paint, briefly exposing its actions.

originalix (approved):

PR #13230, opened 2026-09-08
fix: improve Earn and Borrow UX (OK-61508, OK-61800, OK-61798, OK-61797, OK-58110, OK-58111)

zhaono1:
[OK-61508](https://onekeyhq.atlassian.net/browse/OK-61508) [OK-61800](https://onekeyhq.atlassian.net/browse/OK-61800) [OK-61798](https://onekeyhq.atlassian.net/browse/OK-61798) [OK-61797](https://onekeyhq.atlassian.net/browse/OK-61797) [OK-58110](https://onekeyhq.atlassian.net/browse/OK-58110) [OK-58111](https://onekeyhq.atlassian.net/browse/OK-58111)

----------

## Issues
- Fixes OK-61508
- Fixes OK-61800
- Fixes OK-61798
- Fixes OK-61797
- Fixes OK-58110
- Fixes OK-58111

## Summary
- restore the Earn banner loading state from EarnHome through EarnMobileHomeContent
- reserve the current banner block with a layout-aligned skeleton on true cache misses
- keep resolved empty states empty and retain existing banners during refresh
- pre-mount native Select portals in the closed state so sheets animate from the bottom on first open
- prevent the Borrow E-Mode category sheet from competing with list scroll gestures
- stage Borrow market changes and publish the target market, account, status, and reserves as one settled snapshot
- keep the previous complete Borrow page visible while the target request is loading, instead of returning to full-page skeletons
- scope Earn derive results to their network so a previous network cannot create a duplicate or stale account request
- cancel an in-flight selection cleanly if a refreshed market list removes its target
- prefer amount-specific server lending alerts and use the generic liquidation warning only as a fallback
- allow long Aave repay debt amounts to wrap within a bounded right-aligned value area so the remaining-debt label stays visible
- register the Borrow market transition utility for native split bundles and explicitly ignore the async Select story callback result

## Test plan
- [x] yarn jest packages/kit/src/views/Earn/components/EarnHomeBanner.test.tsx --runInBand
- [x] yarn jest packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.test.tsx --runInBand
- [x] yarn jest packages/kit/src/views/Borrow/components/Markets.utils.test.ts packages/kit/src/views/Borrow/components/borrowDataGate.utils.test.ts packages/kit/src/views/Borrow/borrowDataStatus.test.ts packages/kit/src/views/Staking/hooks/useEarnAccount.test.tsx --runInBand (46 tests)
[...]

PR #13323, opened 2026-09-11
feat: give the e-mode category picker its own screen

labrinyang:
## What

Replaces the `Select` popover in the e-mode category switch with a pushed modal screen, and reworks the need-action screen's footer and shortfall layout.

## Why

The category rows carry two token groups each, and `Select` hard-codes its desktop panel to `$56` (224px), which cannot fit them.

## Notes for review

**This removes `sheetProps={{ disableDrag: true }}` from `EModeCategorySelect.tsx`, which #13230 added for OK-61798. That is not a revert.** The prop reached a Tamagui `Sheet` that only exists because `Select` renders `Popover` on native (`Popover/index.native.tsx` hard-codes `dismissOnSnapToBottom`). This branch renders no `Select`, no `Popover`, and no `Sheet` anywhere in the category flow, so the drag gesture that ticket disabled has no surface to attach to. The replacement is a navigation screen registered in `StakingModalRouter`: `slide_from_right` on iOS, `presentation: 'modal'` on Android, and the modal option factories never set `gestureEnabled`.

**This is a net accessibility gain over what it replaces.** `forms/Select` carries no `role`, no `aria-*` and no `tabIndex`, and its `Trigger` carries no `onKeyDown`, so the category flow was pointer-only on web, desktop and the extension. The rows now announce as radios inside a radiogroup with a real keyboard path.

**Every accessibility state ships twice on purpose.** react-native-web 0.21 dropped `accessibilityState` from its forwarded props, so on the web it reaches no DOM attribute at all; native reads it and ignores `aria-*`.

**`shouldPopOnClickBackdrop: true`** is set on the new route. Without it the web navigator sends a backdrop click to the parent and drops the whole e-mode flow — the `Select` this replaced absorbed that click itself. Note none of the other 21 screens in `StakingModalRouter` set this; only the drill-down screen needs it.

## Independence

Split out of a combined branch alongside #13322 (mobile position card expand). The two share no file and no symbol; either can merge first. Verified by building both from the merge base and confirming the test counts add up exactly (baseline 347 → 401 here, 397 there, 451 merged).

## Known, not fixed here

- The picker renders a header over an empty body if the status cache is cold — no skeleton, no error, no retry. The switch page has a retry branch; this screen does not. Hard to reach because both screens share an SWR key that hydrates synchronously.
- Opening the picker costs one extra `getBorrowEModeStatus` request; `ServiceStaking.ts` has no dedupe on that call.

## Test plan

- [ ] Web: open the category picker, click the backdrop, confirm you land back on the switch page rather than out of e-mode
- [ ] Web: Tab through the categories, pick one with Enter and one with Space
- [ ] Web + VoiceOver/NVDA: confirm the picked category is announced as selected
- [ ] iOS/Android + screen reader: same, through `accessibilityState`
- [ ] Native: confirm the picker cannot be dismissed by dragging it down
- [ ] Pick a category and watch the network panel: exactly one `borrowSwitchCheckEMode`, not two
- [ ] Confirm a `setEMode` that lands while the picker is open does not make the next pick compare against a stale current category

---

https://claude.ai/code/session_016ehg7sC5ui1wKPLfT6eKVp


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

Review of commit `7a7471513369` did not finish successfully.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-43733da2-d478-55eb-813a-a2d4eed9af63)

sidmorizon:
Claude review session: https://claude.ai/code/session_01HwvkXzwstkKjBN9dm3b2kM

packages/kit/src/views/Borrow/pages/BorrowHome.tsx
```diff
@@ -444,6 +445,15 @@ const BorrowHomeContent = memo(
...
             />
+            {/* E-Mode is a market-wide setting rather than a headline number,
+                so on phones it closes the page under the positions and the
+                summary instead of interrupting the metrics at the top. */}
+            <BorrowEModeMetric
+              eModeStatus={eModeStatus}
+              isError={isEModeError}
+              isLoading={isEModeInitialLoading}
+              variant="bar"
+            />
```
sidmorizon: **Reserve errors hide mobile E-Mode**

When reserves fail on phones, `renderCards` returns before the relocated E-Mode bar. Users cannot manage E-Mode until reserves recover.

packages/kit/src/views/Borrow/pages/modal/BorrowEModeCategorySelectModal.tsx
```diff
@@ -0,0 +1,187 @@
...
+
+  // Same scope key as the switch page that pushed this screen, so this reads
+  // straight out of the cache and then keeps following it.
+  const { eModeStatus } = useBorrowEModeStatus({
+    networkId,
+    provider,
+    marketAddress,
+    accountId,
+    enabled: !!accountId,
+  });
```
sidmorizon, line 128: **Picker failures leave empty screen**

When the status request fails without cached data, `eModeStatus` stays null and the picker renders no rows. Users get no error or retry action.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.tsx
```diff
@@ -200,17 +208,40 @@ function BorrowEModeSwitchView() {
...
+      // picker is on top, so its own copy can be the stale one.
+      const currentId = observedCurrentEModeId ?? eModeStatus?.eModeId;
+      if (eModeId === currentId) {
         setUserSelection(null);
         resetTarget();
         return;
       }
       setUserSelection(eModeId);
+      pickRanOwnCheckRef.current = true;
       void runCheck(eModeId);
```
sidmorizon: # 🟠 P1 Pick is discarded when the picker's current id differs from the page's copy

## Problem

This handler decides "already current?" against `observedCurrentEModeId` (the picker's copy), but the selection it stores is still reconciled against the page's own copy on the very next render:

- `index.tsx:123-131` calls `reconcileEModeSelection({ statusCurrentId: eModeStatus?.eModeId ?? null, userSelection, ... })`
- `emodeUtils.ts:181-190` returns `resetTarget: true` when `userSelection === statusCurrentId`
- `index.tsx:190-195` then runs `setUserSelection(null); resetTarget();`

The two copies really can disagree: the picker's `useBorrowEModeStatus` instance fetches on mount, while the page's instance only refetches on focus or through the pending-tx watcher (`refreshManagementState`, which lands via `alwaysSetState` while blurred). Neither instance polls.

Walk the case the PR targets. Page copy = A (stale), picker showed current = B, user picks A:

1. Line 223: `currentId = B`; `A !== B`, so `setUserSelection(A)`, mark set, `runCheck(A)`.
2. Re-render: reconcile sees `userSelection (A) === statusCurrentId (A)` → `resetTarget: true`.
3. Effect at 190-195: `setUserSelection(null)` and `resetTarget()`, which bumps `checkSeqRef` and drops the in-flight check (`useEModeSwitch.ts:141-147`).
4. Page refocuses, fetches B, and lands on "Current: B" with no target.

The mirror case (page = B fresh, picker still showed A, user picks A) is discarded at line 224 as "already current" even though the page knows B is current. Either way the user's pick vanishes with no feedback.

The PR's own test locks this in without noticing: `index.test.tsx` "prefers the current id the picker displayed over its own copy" does `moveCurrentTo(view, 0)` then `pick(0, 1)` and asserts only `mockRunCheck(0)`. With the real `reconcileEModeSelection` the reset effect fires right after, so adding `expect(mockResetTarget).not.toHaveBeenCalled()` fails.

## Impact

In exactly the scenario this ref/observed-id mechanism was added for (status moved while the picker was open), the pick is silently thrown away and the check it started is cancelled. The mechanism only changes which request is fired before the selection is wiped.

## Recommendation

Make the pick decision and the reconcile use the same current id. In `index.tsx`, when `observedCurrentEModeId !== eModeStatus?.eModeId` in this handler, store the observed id (ref or state), call `void refresh()`, and feed `statusCurrentId: observedOverride ?? eModeStatus?.eModeId ?? null` into `reconcileEModeSelection` at line 126 until the page's `eModeStatus.eModeId` catches up to the override, then clear it. Also tighten the test at `index.test.tsx` ("prefers the current id the picker displayed…") to assert `mockResetTarget` is not called and the selector still receives `value: 0` after the pick.

[Claude session ↗](https://claude.ai/code/session_01HwvkXzwstkKjBN9dm3b2kM)

---

originalix:

Review of commit `08f0300de904` did not finish successfully.

packages/kit/src/views/Borrow/pages/modal/BorrowEModeCategorySelectModal.tsx
```diff
@@ -0,0 +1,181 @@
...
```
sidmorizon, line 128: **Status errors leave stale picker active**

When status revalidation fails, `eModeStatus` can retain cached data while `isError` becomes true. The picker ignores `isError` and accepts stale categories.

originalix:

Review of commit `ca99a1acf490` did not finish successfully.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.tsx
```diff
@@ -60,21 +67,46 @@ function BorrowEModeSwitchView() {
...
+    isLoading: statusLoading,
+    isError,
+    refresh,
+  } = useBorrowEModeStatus({
+    networkId,
+    provider,
+    marketAddress,
+    accountId,
+    enabled: !!accountId,
+    revalidateOnFocus: !returningFromPicker,
```
sidmorizon, line 108: 🟡 P2: Picker cancellation leaves stale status

Severity: `non-severe`

When E-Mode changes while the picker is open, cancellation makes `returningFromPicker` suppress the parent refresh. The switch page retains its old current category and prior check.

originalix: P2: [取消分类选择器后父页继续显示旧 E-mode 状态]

父页持有分类 A，picker 的请求拿到更新的分类 B 后，如果用户按返回或点 backdrop 取消，就不会触发 onSelect。此时 returningFromPicker 又跳过了父页的 status/pending history 刷新和已有目标的 runCheck，返回后仍展示旧 Current 和旧检查结果。

相同 SWR key 不能弥补这一点：usePromiseResult 只在初始化或 key 变化时读取缓存，没有订阅其他实例的写入。用真实 useBorrowEModeStatus/usePromiseResult 和共享缓存验证，父页 0、picker 1 的场景在取消返回后仍得到 0；开启返回刷新后，同一断言通过。

建议区分“选择后已执行检查”和“取消”的返回。取消时恢复 status、pending history 刷新并重查保留目标；已选择分支继续保留去重，避免重新引入重复请求。

packages/kit/src/views/Borrow/pages/modal/BorrowEModeCategorySelectModal.tsx
originalix: P2: [选择器无可用缓存时请求失败后没有恢复入口]

这里仅消费 eModeStatus，忽略了 hook 返回的 isError 和 refresh。缓存未命中且请求失败时，或新请求返回空 payload 被 hook 判为错误时，buildEModeRows(null) 返回空数组，页面只剩标题和空列表，没有失败提示或重试入口，用户必须退出重进。

父页此前成功通常能提供缓存，但缓存是 best-effort，不能保证 picker 一定有可用数据。基于当前页面测试注入 eModeStatus: null、isError: true、isLoading: false，错误提示断言得到 null，确认这个分支尚未处理。

建议复用父页的无数据错误提示与 Retry，按钮调用 refresh，并为首次请求提供 loading/skeleton；补一条失败后重试成功能够恢复分类列表的测试。

packages/kit/src/views/Borrow/pages/BorrowHome.tsx
labrinyang: Fixed in afaa2d6a74: on phones the e-mode bar now renders alongside the reserves error state, so the entry point survives a reserves outage. Covered by `BorrowHome.test.tsx` ("keeps the e-mode bar reachable on phones while reserves are in error").

packages/kit/src/views/Borrow/pages/modal/BorrowEModeCategorySelectModal.tsx
labrinyang: Not a correctness issue. When the picker's own refetch fails it falls back to the last-good status, which is the same cached copy the switch page is showing, so its rows are never staler than the page that opened it. A pick from those rows is then validated server-side by the switch check, and once more by the final check inside the switch flow before any transaction is built.

The only remaining gap is that the failure is not surfaced on this screen, which is cosmetic. The no-data dead end is handled separately in the empty-state thread on this file. Resolving.

packages/kit/src/views/Borrow/pages/modal/BorrowEModeCategorySelectModal.tsx
labrinyang: Fixed in e091d83fad. The picker now consumes `isInitialLoading`, `isError` and `refresh` from the hook. With nothing to list it shows the switch page's load error with a Retry that re-runs the request, and a row skeleton while the first request is still in flight.

Tests: skeleton while resolving; error, then retry, then the recovered category list.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.tsx
labrinyang: Fixed in 4673760189. The page now records whether the picker reported a pick before it popped (`onSelect` fires ahead of the pop, so the answer is known by the time the page refocuses).

A pick still skips the return as before. A cancelled return refreshes the status by itself: if the current id moved, the existing revalidation effect re-runs the retained check, and a target that became current is cleared by reconciliation. The pending-history reload, which is the part that locks the footer, stays skipped, since no transaction can start while the picker is up, so the earlier P1 flash does not come back.

The skip-on-cancel test is flipped to assert the refresh, and a moved-current recheck test is added.

labrinyang:
Rebased onto `x` (389d2ff49a). Two conflicts, plus one thing git could not see:

- `BorrowEModeNeedAction/index.tsx`: kept this branch's two-way footer and carried #13379's `isWatchingAccount` into the ordinary footer's `disabled` list. The funding footer's confirm is already disabled.
- The "match the send footer" commit had removed the `accountUtils` import once its last use went away; #13379 needs it again for the watched-account guard. The import is restored inside that commit, so every commit compiles.
- `useStakingPendingTxs.test.tsx`: both sides appended a test at the same spot; kept both. The hook itself merged cleanly.

Also added the two new files to the mobile module-id registry as its own commit (690841b56d). That is what failed the native startup graph budget job on the previous head. Verified locally with the CI-identical strict union build.

238 tests across 29 suites and `agent:check --profile pr` pass.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.tsx
```diff
@@ -160,23 +201,49 @@ function BorrowEModeSwitchView() {
...
+    const cancelledPicker = returningFromPicker && !pickerPickedRef.current;
+    pickerScopeRef.current = null;
+    if (cancelledPicker) {
+      // A cancelled picker reported nothing, yet its own request may have seen
+      // a newer status than this page holds. Refresh the status alone: a moved
+      // current id re-runs the retained check through the revalidation effect
+      // below, while the pending-history reload, which is what locks the
+      // footer, stays skipped because no transaction could have started while
+      // the picker was up.
+      void refresh();
```
sidmorizon, line 224: 🟡 P2: Cancelled picker preserves stale verdict

Severity: `non-severe`

When balances or collateral change while the picker is open, cancellation runs only `refresh` and retains `check`. If `currentEModeId` stays unchanged, no `runCheck` follows, so the footer presents an obsolete switch verdict.

packages/kit/src/views/Borrow/pages/BorrowEModeNeedAction/EModeGetFundsAction.tsx
```diff
@@ -1,44 +1,41 @@
...
-          {intl.formatMessage({ id: ETranslations.global_top_up })}
-        </Button>
+    <Button
+      testID={BorrowTestIDs.eModeNeedActionGetFundsBtn}
+      variant="primary"
+      loading={loading}
+      flexGrow={1}
+      flexShrink={1}
+      // Split the footer evenly with the blocked action.
+      flexBasis={0}
```
labrinyang, line 27: **[P1] `flexBasis={0}` at every width breaks the two-way footer above `$md`**

Both footer buttons (this one and the disabled "Insufficient funds" twin at `index.tsx:802-804`) get `flexGrow 1 / flexShrink 1 / flexBasis 0` unconditionally. `Page.Footer`'s own confirm button applies those only under `$md` (`PageFooterActions.tsx` `confirmButtonMdStyle`), because at `gtMd` the button container is `ml: 'auto'` inside a row and is therefore content-sized. Basis-0 children in a content-sized container have no definite space to share, so the pair cannot split evenly: Chromium sizes the container from the sum of the items' content widths and then hands each item half of it, which truncates the longer label.

Measured in Electron 43 with a minimal reproduction of the footer structure (desktop app and the extension's expanded views both land in `gtMd`):

| buttons | Swap USDC | Insufficient funds |
|---|---|---|
| `flexBasis: 0` (this PR) | 139 px | 139 px, label ellipsized |
| `flexBasis: auto` (Send footer) | 124 px | 154 px, intact |

The Send footer this copies (`SendAmountInputContainer.tsx`, `renderDefaultInsufficientFooterButtons`) deliberately omits `flexBasis`.

Suggested fix: move the even split into the `$md` override, mirroring `confirmButtonMdStyle`, on both buttons:

```tsx
flexGrow={1}
flexShrink={1}
textEllipsis
$md={{ size: 'large', flexBasis: 0 } as IButtonProps}
```

and update `EModeGetFundsAction.test.tsx:63-67`, which pins `flexBasis === 0` at the top level.

_— Auto-review by Claude, cross-validated with Codex_

labrinyang: Fixed in c22f53639a.

Both footer buttons now carry `flexBasis: 0` inside `$md` rather than at the top level, matching `confirmButtonMdStyle` in `PageFooterActions.tsx` and the send page's insufficient-funds row, which omits `flexBasis` for the same reason.

Measured in Chromium before and after, on a reproduction of the footer's structure at `gtMd`:

| buttons | Swap USDC | Insufficient funds |
|---|---|---|
| `flexBasis: 0` | 139 px | 139 px, truncated |
| `flexBasis: auto` | 124 px | 154 px, intact |

`EModeGetFundsAction.test.tsx` pinned the wrong value; it now asserts the top level carries no `flexBasis` and that `$md` does.

packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/index.tsx
labrinyang: The mechanism is exactly as you describe, and it is deliberate. Leaving it open so you can overrule the tradeoff, since it is a close call and you are the second reviewer to land on this line.

Cancellation refreshes the status alone because re-checking on a cancel puts the footer into `checking`, which is loading and disabled. That is the outcome of a no-op interaction: the user opened the list, changed their mind, backed out, and the button they were about to press goes dead for the duration of a request they did not ask for. The refresh-only shape came out of the earlier thread on picker return locking the footer, and `index.test.tsx` pins it as `refreshes only the status on picker cancellation while preserving a checked target`.

What that leaves uncovered is narrower than the general case. A moved current id does re-check, through the revalidation effect. What survives is a verdict whose *inputs* moved while the picker was up, with the current id unchanged.

That residue is bounded at the point it would matter. `confirmSwitch` does not act on the on-screen verdict: it calls `setEMode`, which returns an authoritative check, and on `!latestCheck.canSwitch` it feeds that through `applyAuthoritativeCheck` instead of proceeding (`useEModeSwitch.ts:148-190`). So pressing a stale-looking Switch re-validates server-side and corrects the screen rather than performing a switch the backend would reject.

If you would rather take the flicker and be fresh on every return, it is one line in the cancel branch — `void runCheck(selection.userSelection)` alongside the refresh — plus reversing that test's `runCheck` assertion. Say the word and I will make the change; I did not want to undo a previous reviewer's P1 fix on my own judgement.

originalix (approved):

PR #13325, opened 2026-09-11
fix: format K-line prices and align financial performance axes ｜OK-62760 & OK-62704

zy2000-star:
[OK-62760](<https://onekeyhq.atlassian.net/browse/OK-62760>) [OK-62704](<https://onekeyhq.atlassian.net/browse/OK-62704>)

----------

### Summary

Fixes [OK-62760](https://onekeyhq.atlassian.net/browse/OK-62760) and [OK-62704](https://onekeyhq.atlassian.net/browse/OK-62704).

### Changes

- Improve K-line price-axis labels:
  - Compress more than five leading decimal zeros using subscript notation.
  - Keep the displayed amount within the configured character limit.
  - Replace the last two amount characters with `...` when the value is too long.
  - Preserve the currency symbol and K/M/B unit suffix.
  - Remove floating-point precision artifacts such as `$0.7000...`.
  - Apply the same formatting behavior to native WebView charts.
  - Reserve enough axis width to prevent labels from being clipped.

- Correct the Financials Performance chart:
  - Keep revenue and net income bars mapped to the right amount axis.
  - Keep net margin percentage mapped to the left percentage axis.
  - Use stable chart intervals that match the expected Financials reference scale.
  - Handle negative, cross-zero, constant, and missing values safely.

### Validation

- Added price formatter tests for:
  - Small decimal values.
  - Leading-zero compression.
  - K/M/B conversion.
  - Long-value truncation.
  - Floating-point precision artifacts.
  - Native WebView serialization.

- Added Financials chart tests for:
  - Normal values.
  - Negative values.
  - Cross-zero values.
  - Constant values.
  - Missing values.

- Targeted Jest tests passed.

[OK-62760]: https://onekeyhq.atlassian.net/browse/OK-62760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62704]: https://onekeyhq.atlassian.net/browse/OK-62704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix:

Review of commit `44b490d740f8` did not finish successfully.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-b9bbb1c3-2845-5393-8cee-93087ab931d5)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy

packages/kit/src/components/LightweightChart/utils/htmlTemplate.ts
```diff
@@ -14,7 +15,13 @@ function getStyles(): string {
 
 function getChartInitScript(): string {
   return `
+      var compactPriceFormatter = ${formatChartPrice.toString()};
```
sidmorizon: # 🟠 P1 Native compact formatter is shipped via `Function.prototype.toString()`, which Hermes does not preserve in release builds

## Problem

`getChartInitScript()` embeds the formatter into the WebView by serializing the live function object:

```js
var compactPriceFormatter = ${formatChartPrice.toString()};
```

`generateChartHTML()` runs inside the React Native JS runtime, which is Hermes on both mobile targets (`apps/mobile/android/gradle.properties:47` → `hermesEnabled=true`; `apps/mobile/ios/Podfile:133`). A release build precompiles the bundle to Hermes bytecode, and Hermes does not retain function source text for bytecode-compiled functions — `Function.prototype.toString()` yields a placeholder body (`function formatChartPrice() { [bytecode] }`) unless the function carries the `'show source'` directive. `formatChartPrice` (`packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts:2`) does not, and there is no such directive anywhere in this directory.

## Impact

In an iOS/Android **release** build the generated HTML contains a function whose body is the placeholder rather than the formatter. It still parses, so the page loads and the chart is constructed with no visible error at init. But `getPriceFormatter` returns a wrapper around it (lines 20-24 below) and lightweight-charts invokes that wrapper for every price-axis tick and crosshair label, so each call throws from inside the chart's paint path instead of returning a string. For a tick of `123.45` the axis gets a thrown error instead of `$123.45`, and the price scale fails to render.

Reached from `StockPriceLineChart.tsx:246` — the only caller that sets `compactPriceMaxCharacters` — via `StockSimpleChart` and `SwapStockDesktopContainer`.

Nothing in this PR's validation reaches the failing path: Metro dev builds compile Hermes from source, so `toString()` works there, and both new tests (`formatChartPrice.test.ts:47`, `lightweightChartsRuntime.test.ts:299`) execute under `@swc/jest` in Node. So "Apply the same formatting behavior to native WebView charts" holds in dev and silently fails in release.

## Recommendation

Don't depend on `Function.prototype.toString()` to cross the WebView boundary. This directory already has the right mechanism: `lightweightChartsRuntime.ts` ships the chart library as source text through `babel-plugin-inline-import`, with the `.text-js` extension registered at `development/babelTools.js:120-124`. Move the formatter body into a `.text-js` file, import it as a string for the template, and have `formatChartPrice.ts` evaluate or re-export that same text so the TS and WebView copies cannot drift.

A `'show source'` directive on `formatChartPrice` would also work, but it is Hermes-specific and breaks again the moment anything wraps or re-exports the function.

To confirm before fixing: dump `generateChartHTML(...)` from a release build and grep the output for `[bytecode]`.

[Claude session ↗](https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy)

---

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -0,0 +1,41 @@
...
+    { exponent: 9, suffix: 'B' },
+    { exponent: 6, suffix: 'M' },
+    { exponent: 3, suffix: 'K' },
+  ].find((item) => value >= 10 ** item.exponent);
+  const unit = scale?.suffix ?? '';
+  const parts = value.toExponential().split('e');
+  const digits = parts[0].replace('.', '');
+  const exponent = Number(parts[1]) - (scale?.exponent ?? 0);
+  let body: string;
+  if (value > 0 && !unit && exponent < -6) {
```
sidmorizon: # 🟠 P1 Zero compression starts one decade too late, so small-price axis labels all collapse to the same string

## Problem

Compression is gated on `exponent < -6`, i.e. it only engages at six or more leading decimal zeros. But the character limit is applied to `body` at line 35, and in the **five**-leading-zero decade `body` is already 8-10 characters. Those values therefore fall through to the `else if (exponent < 0)` branch at line 26 and get chopped by the ellipsis rule instead of compressed — the ellipsis eats the significant digits and leaves only leading zeros.

## Impact

Executing the function as written, with the two limits the caller actually passes (`StockPriceLineChart.tsx:91,246` → `8`, or `7` when `chartWidth < 400`):

| price | limit 8 | limit 7 |
|---|---|---|
| `0.0000011` | `$0.0000...` | `$0.000...` |
| `0.0000012` | `$0.0000...` | `$0.000...` |
| `0.0000013` | `$0.0000...` | `$0.000...` |
| `0.000122` | `$0.000122` | `$0.000...` |
| `0.000124` | `$0.000124` | `$0.000...` |

Every tick on the axis of an instrument priced around `1.2e-6` renders the identical label `$0.0000...`, so the price scale conveys nothing. On phones (`chartWidth < 400`, limit 7) the collapse starts a decade earlier and produces a mixed, partly-useless axis such as `$0.00012 / $0.000... / $0.000... / $0.00013`.

This is a regression against the formatter being replaced here: `formatPrice` kept `4 + countLeadingZeroDecimals(val)` digits (`packages/shared/src/utils/numberUtils.ts:605-609`) and applied no character cap, so those ticks were distinct.

The new tests don't catch it because the only case in this decade uses an artificial limit — `formatChartPrice.test.ts:8` asserts `formatChartPrice(0.000_001_2, 20)`, i.e. 20 characters rather than the 7 or 8 production passes.

## Recommendation

Drive compression from the display budget rather than a fixed exponent: compress whenever the plain decimal form would exceed `limit` (test `-exponent - 1 >= 4`, or build the plain body first and compress when `body.length > limit`), and spend the ellipsis on the significant digits rather than on the leading zeros. `0.0000012` then renders `$0.0₅12` — 7 characters, fits both limits — and neighbouring ticks stay distinguishable. Worth adding a test at the default limit for the `1e-6`/`1e-5` decades.

[Claude session ↗](https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy)

---

packages/kit/src/components/StockPriceLineChart/StockPriceLineChart.tsx
```diff
@@ -87,11 +88,8 @@ export function StockPriceLineChart({
   const [chartWidth, setChartWidth] = useState(0);
   const priceFormatter = useCallback(
     (price: number) =>
-      numberFormat(String(price), {
-        formatter: 'price',
-        formatterOptions: { currency: '$' },
-      }),
-    [],
+      formatChartPrice(price, chartWidth > 0 && chartWidth < 400 ? 7 : 8),
+    [chartWidth],
```
sidmorizon: # 🟡 P2 `priceFormatter` keyed on raw `chartWidth` destroys and rebuilds the chart on every resize pixel

## Problem

The dependency array changed from `[]` to `[chartWidth]`, but the formatter only has two possible behaviours — 7 or 8 characters. `chartWidth` is updated for every distinct layout width (lines 215-220), so the callback takes a new identity on each intermediate pixel while its behaviour is unchanged.

That identity flows into `useChartConfig` (`hooks/useChartConfig.ts:169,211`), and `chartConfig.priceFormatter` is a dependency of the web chart-creation effect (`LightweightChart.tsx:506`), whose cleanup calls `chart?.remove()` (`LightweightChart.tsx:487`).

The sibling prop on line 246 is already bucketed correctly: `compactPriceMaxCharacters={chartWidth > 0 && chartWidth < 400 ? 7 : 8}` is a number, so it is stable within a bucket.

## Impact

Dragging a browser/desktop window edge, or toggling the Market detail side panel, fires `onLayout` for each intermediate width (900 → 899 → 898…). Each step tears down the lightweight-charts instance and rebuilds it through the async `getChartLib()` path, re-running `setData` + `fitContent` and resetting `setLastPointPosition(null)` (`LightweightChart.tsx:193`). That bypasses the dedicated `ResizeObserver` handler at `LightweightChart.tsx:452-469`, which exists precisely to apply `applyOptions({ height, width })` without recreating anything — so the chart flickers and the pulse-dot / crosshair anchor is dropped throughout the drag.

On native the same identity change makes `nativeConfig` new (`LightweightChart.native.tsx:106-113`) and re-injects a full `applyChartConfig` on every layout width, even though the serialized payload is byte-identical — `priceFormatter` is a function and `JSON.stringify` drops it.

## Recommendation

Memoize on the bucket rather than the pixel, and reuse it for both props:

```ts
const maxPriceCharacters = chartWidth > 0 && chartWidth < 400 ? 7 : 8;
const priceFormatter = useCallback(
  (price: number) => formatChartPrice(price, maxPriceCharacters),
  [maxPriceCharacters],
);
```

then pass `compactPriceMaxCharacters={maxPriceCharacters}`. Identity changes at most once per threshold crossing, and the duplicated ternary on line 246 disappears.

[Claude session ↗](https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy)

---

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -0,0 +1,41 @@
+// Keep this function self-contained: native charts embed it in the WebView.
+export function formatChartPrice(price: number, maxCharacters = 8): string {
```
sidmorizon, line 2: # 🟠 P1 New module is not in the mobile bundle-id registry, so the iOS union build aborts

## Problem

`formatChartPrice.ts` is a new module that enters the native module graph through two static, top-level imports:

- `utils/htmlTemplate.ts:1` — `import { formatChartPrice } from './formatChartPrice';`
- `StockPriceLineChart.tsx:12` — `import { formatChartPrice } from '../LightweightChart/utils/formatChartPrice';`

But it was never added to `apps/mobile/bundle-registry/module-id-registry.json`, and the PR does not touch that file. Every other module in this directory is registered — `chartColor.ts`, `constants.ts`, `htmlTemplate.ts`, `lightweightChartsRuntime.ts`, `lightweightChartsRuntimeSource.ts`, `priceFormatterType.ts` — so the new one is the only gap.

## Impact

`Native startup graph budget (24.x)` fails. `buildSegmentAllocation` throws after both graphs are built:

```
packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
    at buildSegmentAllocation (apps/mobile/scripts/unionBuild.js:562:36)
```

The iOS union build exits 1 before the budget check runs, so no allocation report is produced at all. This has now failed on two consecutive heads (`c7a8e55e` and `41ff5bad`); the three `Merge branch 'x'` commits in between did add 58 registry lines, but those came from the base branch and are for unrelated modules, so the gap is still open on the current head.

## Recommendation

Run the command the build error names and commit the result:

```
```

Worth doing before the other fixes on this PR: if `formatChartPrice` is later moved into a `.text-js` file (see the separate comment on `htmlTemplate.ts:18` about `Function.prototype.toString()` not surviving Hermes release builds), the registry will need regenerating for the new path anyway.

[Claude session ↗](https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy)

---

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -70,7 +71,9 @@ export function FinancialChart({
   const selectedIndex = rows.findIndex((row) => row.key === selected);
   const selectedRow = rows[selectedIndex];
   const lineIndex = series.findIndex((item) => item.kind === 'line');
-  const domain = getFinancialDomain(
+  const domain = (
+    lineIndex >= 0 ? getFinancialPerformanceDomain : getFinancialDomain
+  )(
```
sidmorizon, line 76: # 🟠 P1 Domain swap changes the performance axis ticks and breaks an existing test that was not updated

## Problem

Switching the amount domain to `getFinancialPerformanceDomain` when `lineIndex >= 0` changes the tick values the performance chart renders, but `StockFinancials.test.tsx` still asserts the old axis. That file is not touched by this PR.

For the existing test input at `StockFinancials.test.tsx:368-391` — a single performance row with `revenue: 100_000_000_000`, `netIncome: 20_000_000_000`, and `series = [revenue, netIncome, margin]` so `lineIndex === 2` and the amount domain receives `[100e9, 20e9]`:

| | domain | rendered ticks |
|---|---|---|
| before (`getFinancialDomain`) | `{min: 0, max: 110e9}` | `110B, 82.5B, 55B, 27.5B, 0B` |
| after (`getFinancialPerformanceDomain`) | `{min: 0, max: 100e9}` | `100B, 75B, 50B, 25B, 0B` |

The old top tick came from `max + span * 0.1` = `100e9 + 10e9`. The new one comes from `getFinancialTickStep(100e9 / 4)` → `2.5e10`, giving `lower + step * 4` = exactly `100e9`.

## Impact

`Unit Tests (3/5)` fails, and with it the `unittest (24.x)` rollup, on both heads so far (`c7a8e55e` and `41ff5bad`). The failing assertion is `StockFinancials.test.tsx:391`:

```ts
expect(
  screen.getByTestId('chart').querySelector('svg')?.textContent,
).toContain('110B');
```

`110B` no longer appears anywhere in the SVG — the axis now reads `100B / 75B / 50B / 25B / 0B`. The jest output is not in the job log (it is only inside the `ci-ai-unittest-diagnostics-shard-3-of-5` artifact), so the check reports a bare `exit code 1` and the cause is not visible from the PR page.

Note the new tick values look like the intended behaviour — `0 / 25B / 50B / 75B / 100B` is a cleaner scale than `0 / 27.5B / 55B / 82.5B / 110B`, which is what "use stable chart intervals" asks for. The defect is that the behaviour change was not carried into the existing test.

## Recommendation

Update the assertion at `StockFinancials.test.tsx:391` to the new top tick:

```ts
).toContain('100B');
```

Worth also asserting one interior tick (e.g. `25B`) so the test pins the interval spacing rather than only the maximum — that is the property this PR is actually changing. The tooltip assertions immediately below (`Revenue 100B`, `Net income 20B`, `Net margin % 20%`) read raw row values, not the domain, so they stay correct.

[Claude session ↗](https://claude.ai/code/session_01Bc4rjA6XXyinvVkPnHSEpy)

---

packages/kit/src/components/LightweightChart/utils/formatChartPrice.text-js
```diff
@@ -0,0 +1,37 @@
...
+  }
+  if (value > 0 && !unit && exponent < 0 && body.length > limit) {
+    var zeros = String(-exponent - 1).replace(/[0-9]/g, function (digit) {
+      return '₀₁₂₃₄₅₆₇₈₉'.charAt(Number(digit));
+    });
+    body = '0.0' + zeros + digits;
+  }
+  if (body.length <= limit) return prefix + body + unit;
+  var end = Math.max(1, limit - 2);
+  return prefix + body.slice(0, end) + '...' + unit;
```
sidmorizon, line 38: 🟡 P2: Native price axis can cut a subscript zero-count

Severity: `non-severe`

The native WebView embeds [formatChartPrice](packages/kit/src/components/LightweightChart/utils/formatChartPrice.text-js), which omits the subscript-indivisibility loop that [formatChartPrice](packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts:40) has. When a compressed label overflows and the cut index lands inside a multi-digit zero-count, native truncates mid-subscript, so web and native render different, misleading prices.

Suggested fix:

```
  var end = Math.max(1, limit - 2);
  // A zero-count subscript is indivisible, including counts with multiple digits.
  while (/[₀₁₂₃₄₅₆₇₈₉]/.test(body.charAt(end))) end += 1;
  return prefix + body.slice(0, end) + '...' + unit;
```

zy2000-star:
fixed+resolve

packages/kit/src/components/LightweightChart/utils/htmlTemplate.ts
zy2000-star: fixed+resolve

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
zy2000-star: fixed+resolve

packages/kit/src/components/StockPriceLineChart/StockPriceLineChart.tsx
zy2000-star: fixed+resolve

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
zy2000-star: fixed+resolve

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
zy2000-star: fixed+resolve

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -0,0 +1,42 @@
...
+  let body: string;
+  if (decimalPosition <= 0) {
+    body = `0.${'0'.repeat(-decimalPosition)}${digits}`;
+  } else {
+    body =
+      decimalPosition >= digits.length
+        ? digits + '0'.repeat(decimalPosition - digits.length)
+        : `${digits.slice(0, decimalPosition)}.${digits.slice(decimalPosition)}`;
+  }
+  if (value > 0 && !unit && exponent < 0 && body.length > limit) {
```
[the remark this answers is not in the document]
zy2000-star: fixed+resolve

packages/kit/src/components/LightweightChart/utils/formatChartPrice.text-js
zy2000-star: fixed+resolve

packages/kit/src/components/LightweightChart/utils/htmlTemplate.ts
zy2000-star: fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
zy2000-star: fixed

packages/kit/src/components/StockPriceLineChart/StockPriceLineChart.tsx
zy2000-star: fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
zy2000-star: fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.text-js
zy2000-star: fixed

PR #13327, opened 2026-09-11
fix: Ledger install dialog under the stage, and a smaller PIN stage (OK-62656, OK-62091)

franco-chan:
[OK-62656](<https://onekeyhq.atlassian.net/browse/OK-62656>) [OK-62091](<https://onekeyhq.atlassian.net/browse/OK-62091>)

----------

## Summary

Two Device Stage fixes from the mobile QA round.

### OK-62656 — Ledger core-app install dialog covered by the stage
- Onboarding holds one stage burst across the whole wallet creation, and the Ledger core-app install dialog opens inside that hold, so the stage's **processing capsule stayed docked at the bottom of the phone screen right over the sheet's Install button**.
- The dialog helper now asks the stage to yield first (`deviceStageYieldToDialog`, the same hand-off the bootloader dialog uses since OK-62105) and waits for its exit before the sheet rises. The hold's bookkeeping is untouched: the install progress and the device's next word bring the stage back, and the hold's own end still releases it.
- Not a literal `endBurst`/`ensureBurst` pair: a vendor burst's `end()` closes a successful run with the ✓ "done" beat, which would flash before the install. Placing the yield in the dialog helper also covers the accounts-phase check that opens the same dialog under the same hold.

### OK-62091 — on-device PIN card too big on phones
- The replica width in play drops from the spec's 280pt to **220pt**: the PIN card goes from 481 to 408pt (55% → 47% of an iPhone 17 Pro screen) while the device's own screen stays 181pt wide, its keypad still readable.
- The spec's numbers stay stated at a new `STAGE_DESIGN_WIDTH` (280) and `DeviceStage` scales them to the width in play, so the confirm miniature and the capsule thumbnail keep their absolute sizes. The compact/full discriminator is now the arrangement kind rather than the port value.
- New `replicaWidth` prop with a range control in the Pin, Passphrase and Console stories for tuning.
- On-device storybook: the preview mounts the stage on the overlay window with the window's safe-area insets (matching the app's container order); the stories' driver bar now clears the status bar.

## Verification
- `yarn agent:check --profile commit` and `--profile pr` green.
- Unit tests: DeviceStage + MorphOverlay + kit DeviceStageContainer + ThirdPartyHardwareUiStateContainer (78 + 33) pass.
- OK-62091 checked on iPhone 17 Pro simulator storybook (Pin / Passphrase flows at 220).
- OK-62656 needs a device check: Ledger without apps through onboarding — the capsule should leave before the sheet rises and the Install button is tappable; tapping Install brings the stage back with the batch checklist; closing the sheet keeps the stage off and errors as before.

[OK-62656]: https://onekeyhq.atlassian.net/browse/OK-62656?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62091]: https://onekeyhq.atlassian.net/browse/OK-62091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/apps/mobile/.rnstorybook/preview.tsx
+++ b/apps/mobile/.rnstorybook/preview.tsx
@@ -14,6 +14,7 @@ import { Portal } from '@onekeyhq/components/src/hocs/Portal';
 import { ConfigProvider } from '@onekeyhq/components/src/hocs/Provider';
 import { OverlayContainer } from '@onekeyhq/components/src/layouts/OverlayContainer';
 import { Stack } from '@onekeyhq/components/src/primitives/Stack';
+import { HARDWARE_STAGE_Z_INDEX } from '@onekeyhq/shared/src/consts/zIndexConsts';
 
 import { HyperlinkTextStub } from './HyperlinkTextStub';
 
@@ -84,18 +85,25 @@ const preview: Preview = {
               FullWindowOverlay window, because that is where the app
               mounts it since 8c0391dfa8 (the stage must cover native
               modal pages on iOS). Keeping the shell on the same window
-              means on-device rounds exercise the real geometry — the
-              overlay window's own touch delivery included. The wrapper
-              gives the portal a viewport on the pass-through platforms:
-              OverlayContainer is a full-window host only on iOS, and
-              MorphOverlay's layer anchors absolute to fill it. */}
+              means on-device rounds exercise the window's geometry and
+              touch delivery; the app additionally nests the stage in its
+              own OverlayContainer with a raise token (OK-62422), which
+              this shell does not replay. The wrapper mirrors the app's
+              (FullWindowOverlayContainer): a viewport on the
+              pass-through platforms — OverlayContainer is a full-window
+              host only on iOS, and MorphOverlay's layer anchors absolute
+              to fill it — kept as a native view, since RN 0.86 Fabric
+              flattens a layout-only box-none container and kills
+              hit-testing for the whole portal subtree. */}
           <Stack
             position="absolute"
             top={0}
             left={0}
             right={0}
             bottom={0}
+            zIndex={HARDWARE_STAGE_Z_INDEX}
             pointerEvents="box-none"
+            collapsable={false}
           >
             <Portal.Container name={Portal.Constant.HARDWARE_UI_STATE_DIALOG} />
           </Stack>
--- a/apps/mobile/.rnstorybook/preview.tsx
+++ b/apps/mobile/.rnstorybook/preview.tsx
@@ -1,6 +1,10 @@
 import type { PropsWithChildren } from 'react';
 
 import { useColorScheme } from 'react-native';
+import {
+  SafeAreaInsetsContext,
+  initialWindowMetrics,
+} from 'react-native-safe-area-context';
 
 import {
   ShowToastProvider,
@@ -14,6 +18,13 @@ import { Stack } from '@onekeyhq/components/src/primitives/Stack';
 import { HyperlinkTextStub } from './HyperlinkTextStub';
 
 import type { Preview } from '@storybook/react';
+
+const WINDOW_INSETS = initialWindowMetrics?.insets ?? {
+  top: 0,
+  right: 0,
+  bottom: 0,
+  left: 0,
+};
 
 function ShellProvider({ children }: PropsWithChildren) {
   // Theme follows the system appearance — the on-device UI has no toolbar
@@ -30,7 +41,16 @@ function ShellProvider({ children }: PropsWithChildren) {
       locale="en-US"
       HyperlinkText={HyperlinkTextStub}
     >
-      {children}
+      {/* Window insets for the whole decorated tree: ConfigProvider's own
+          SafeAreaProvider measures the story canvas — a view already
+          under the Storybook chrome, so its top inset settles at 0 once
+          the native view reports — while components that seat
+          themselves against the real status bar band (the hardware
+          stage, the toasts) read the app's root provider in production.
+          The main window's launch metrics stand in for it here. */}
+      <SafeAreaInsetsContext.Provider value={WINDOW_INSETS}>
+        {children}
+      </SafeAreaInsetsContext.Provider>
     </ConfigProvider>
   );
 }
@@ -60,26 +80,28 @@ const preview: Preview = {
             Storybook UI. */}
         <OverlayContainer>
           <Portal.Container name={Portal.Constant.FULL_WINDOW_OVERLAY_PORTAL} />
+          {/* The hardware stage's mount point — INSIDE the
+              FullWindowOverlay window, because that is where the app
+              mounts it since 8c0391dfa8 (the stage must cover native
+              modal pages on iOS). Keeping the shell on the same window
+              means on-device rounds exercise the real geometry — the
+              overlay window's own touch delivery included. The wrapper
+              gives the portal a viewport on the pass-through platforms:
+              OverlayContainer is a full-window host only on iOS, and
+              MorphOverlay's layer anchors absolute to fill it. */}
+          <Stack
+            position="absolute"
+            top={0}
+            left={0}
+            right={0}
+            bottom={0}
+            pointerEvents="box-none"
+          >
+            <Portal.Container name={Portal.Constant.HARDWARE_UI_STATE_DIALOG} />
+          </Stack>
           <ShowToastProvider />
           <Toaster />
         </OverlayContainer>
-        {/* The hardware stage's mount point — deliberately OFF the
-            FullWindowOverlay window, matching the app's own container
-            order (FullWindowOverlayContainer mounts it beside, not
-            inside): the stage sits at the main window's dialog level, so
-            presentations opened over it — the in-app browser, system
-            sheets — actually cover it. Canvas-wide and box-none: the
-            stage positions itself, the UI behind stays live. */}
-        <Stack
-          position="absolute"
-          top={0}
-          left={0}
-          right={0}
-          bottom={0}
-          pointerEvents="box-none"
-        >
-          <Portal.Container name={Portal.Constant.HARDWARE_UI_STATE_DIALOG} />
-        </Stack>
       </ShellProvider>
     ),
   ],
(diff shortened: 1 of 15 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5a476468-78af-5c1e-86fb-c04161385850)

sidmorizon:
Claude review session: https://claude.ai/code/session_01UTp2xQRWJKuBYxomKfM7Tq

apps/mobile/.rnstorybook/preview.tsx
```diff
@@ -60,26 +81,35 @@ const preview: Preview = {
...
           <Portal.Container name={Portal.Constant.FULL_WINDOW_OVERLAY_PORTAL} />
+          {/* The hardware stage's mount point — INSIDE the
+              FullWindowOverlay window, because that is where the app
+              mounts it since 8c0391dfa8 (the stage must cover native
+              modal pages on iOS). Keeping the shell on the same window
+              means on-device rounds exercise the window's geometry and
+              touch delivery; the app additionally nests the stage in its
+              own OverlayContainer with a raise token (OK-62422), which
+              this shell does not replay. The wrapper mirrors the app's
+              (FullWindowOverlayContainer): a viewport on the
```
sidmorizon, line 92: # 🟡 [P2] Storybook stage wrapper moves into the overlay window without `collapsable={false}`, so its touch delivery is exactly what stays unfaithful

## Problem

This wrapper (`apps/mobile/.rnstorybook/preview.tsx:92-101`) is now mounted inside `OverlayContainer` — the iOS `FullWindowOverlay` window — and the comment above it states the reason as "on-device rounds exercise the real geometry — the overlay window's own touch delivery included."

The app's structurally identical wrapper omits nothing here. `packages/kit/src/provider/Container/FullWindowOverlayContainer/index.tsx:84-98` wraps the same `Portal.Constant.HARDWARE_UI_STATE_DIALOG` portal in the same `position="absolute"` + full inset + `pointerEvents="box-none"` `Stack`, inside the same `OverlayContainer`, and carries:

```tsx
// RN 0.86 Fabric flattens this layout-only box-none container out
// of the native hierarchy, which kills hit-testing for the whole
// portal subtree (draws fine, touches dead). Keep the native view.
collapsable={false}
```

RN is `0.86.2` (`apps/mobile/package.json:162`), so that note applies to this build. `MorphOverlay`'s own layer already carries the flag (`packages/components/src/composite/MorphOverlay/index.tsx:1285`) and the app still needed it on the wrapper as well, so the inner flag does not cover this container.

To be clear on provenance: the flag was also absent before this PR, when the wrapper sat outside `OverlayContainer`. What the PR changes is that the wrapper now occupies the precise structural position where the app found the flag necessary, and it does so specifically to make overlay-window touch delivery representative.

## Impact

On iOS the storybook stage can render inside the `FullWindowOverlay` while its touches are dead — draws fine, taps do nothing — for the whole portal subtree. That is the single property the relocation is meant to validate, so an on-device round can read as "stage geometry confirmed" while silently exercising a hit-testing path the app does not have. Storybook-only, hence P2 rather than P1.

## Recommendation

Add the flag to the `Stack` at `preview.tsx:92`, mirroring `FullWindowOverlayContainer/index.tsx:95`:

```tsx
<Stack
  position="absolute"
  top={0}
  left={0}
  right={0}
  bottom={0}
  pointerEvents="box-none"
  // RN 0.86 Fabric flattens this layout-only box-none container out of
  // the native hierarchy, which kills hit-testing for the portal subtree.
  collapsable={false}
>
```

For full parity with the app's cross-platform ordering guarantee, consider also applying the same `zIndex` the app sets at `FullWindowOverlayContainer/index.tsx:90`.

[Claude session ↗](https://claude.ai/code/session_01UTp2xQRWJKuBYxomKfM7Tq)

---

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
```diff
@@ -1640,7 +1641,7 @@ export class DeviceStageBurstScope {
     // by the burst's own end, a user close, a new burst, or the device
     // asking again (see onHardwareUiEvent).
     this.yieldedToDialog = true;
-    await this.forceOff({ force: true });
+    return this.forceOff({ force: true });
```
sidmorizon: 🟡 P2: Concurrent stage claim covers dialog

Severity: `non-severe`

When a stage claim races `silence()`, `forceOff` returns false and `yieldDeviceStageToDialog` opens the dialog immediately. The surviving modal stage can cover the Ledger install dialog.

apps/mobile/.rnstorybook/preview.tsx
franco-chan: Addressed in 1d9e5ede5f: the storybook stage wrapper now carries collapsable={false} and zIndex={HARDWARE_STAGE_Z_INDEX}, mirroring the app's FullWindowOverlayContainer wrapper; the comment above it no longer claims the raise-token layering is replayed.

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
franco-chan: Fixed in c92dfa9289: silence() now re-runs the exit while a claim raced it (bounded to three passes) and stands down only when the device asked again and lifted the yield, so a straggler processing paint can no longer stay up under the install sheet. Regression test added (DeviceStageBurst.test.ts: "takes down a wait a straggler paints while the yield reads the stage").

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.test.ts
```diff
@@ -731,6 +734,34 @@ describe('DeviceStageBurstScope', () => {
...
+    const scope = new DeviceStageBurstScope();
+    const token = await scope.beginExplicit({
+      connectId: CONNECT_ID,
+      vendor: EHardwareVendor.ledger,
+    });
+    await paintOpeningBeat();
+    expect(stage?.step).toBe('connecting');
+    stageAtom.get.mockImplementationOnce(async () => {
+      const read = stage;
+      await scope.onThirdPartyState({
```
sidmorizon, line 755: # 🟡 [P2] The new regression test never bumps the claim, so it passes against the pre-fix commit and does not pin the race

## Problem

The straggler this test injects does not take the code path the fix defends against. `claimSeq` has exactly one mutation site — `DeviceStageBurst.ts:515`, inside `clearOffTimer()` — and every `clearOffTimer()` call site is at or before `:1747`, while `setStep` begins at `:1798` and never calls it.

Tracing `onThirdPartyState({ ui: undefined, install: undefined, batch: undefined })` at this head:

- `install` is undefined → the `clearOffTimer()` at `:1330` is skipped
- `ui` is undefined → the `clearOffTimer()` at `:1367` is skipped
- falls through to `:1381`, and `depth > 0` from the `beginExplicit` hold → `setStep('processing', {})` at `:1395`

So the straggler repaints `processing` — which is why the in-mock assertion at `:757` passes — but it **does not bump `claimSeq`**, contrary to the test's own comment at `:741` ("and bumps the claim").

The consequence is that `forceOff`'s guard at `:1764` (`claim !== this.claimSeq`) passes on the very first attempt, the off is written from the stale snapshot, and `true` is returned. `silence()` then takes the `claim === this.claimSeq` break at `:1653` with `attempt === 0`: the retry loop body executes exactly once and never iterates.

At `1d9e5ede`, `silence()` was a bare `return this.forceOff({ force: true })` — the same single `forceOff` call, the same write, the same `true`. Both assertions here (`:760` `resolves.toBe(true)` and `:761` `step === 'off'`) therefore hold on the pre-fix commit.

## Impact

The test named as the regression test for this fix is green against the bug it is meant to pin. The retry loop added at `:1650-1656` has no coverage: it could be deleted, or its bound changed to one pass, and this test would still pass. The race fixed in `c92dfa9289` is free to regress silently.

The product fix itself is sound — this is about the guard around it, not the fix.

## Recommendation

Use a straggler that actually reaches `clearOffTimer()`. The closest match to the scenario the comment describes (the probe's call-end clear) is the call-end close, which under a live yield paints nothing and bumps the claim:

```ts
await scope.onHardwareUiEvent({
  action: EHardwareUiStateAction.CLOSE_UI_WINDOW,
  connectId: CONNECT_ID,
});
```

With `yieldedToDialog` already true, `DeviceStageBurst.ts:1083` is false, so it takes `scheduleOff(OFF_GRACE_MS, ...)` at `:1089` → `clearOffTimer()` → `claimSeq += 1`, painting nothing. Attempt 0 then aborts at `:1764` and returns `false` with the stage still `connecting`; attempt 1 re-reads and writes the off. That version fails at `1d9e5ede` and passes here — which is what makes it a regression test.

If you want to keep a repaint in the scenario as well, `onThirdPartyState({ install: { ... }, ui: undefined, batch: undefined })` reaches the `clearOffTimer()` at `:1330` *and* repaints, giving both halves.

Either way, if the straggler is left as-is, drop the "and bumps the claim" clause from the comment at `:741`, since it does not.

[Claude session ↗](https://claude.ai/code/session_01UTp2xQRWJKuBYxomKfM7Tq)

---

franco-chan: Good catch. Fixed in 395314dbfb: the straggler is now a third-party ui action (EThirdPartyHardwareUiAction.processing), which reaches clearOffTimer and bumps the claim as well as repainting, so the first exit stands down on the stale read and only the retry writes the off. Verified it fails with the bare-forceOff silence() (Expected true, Received false) and passes with the retry; comment updated to match.

originalix:

Review of commit `395314dbfb3c` finished, but some comments could not be published automatically.

originalix (approved):

PR #13330, opened 2026-09-11
fix: restore simulation asset layout and nft quantity (OK-62907)

yikZero:
[OK-62907](<https://onekeyhq.atlassian.net/browse/OK-62907>)

----------

NFT transaction previews truncated collection names in a fixed desktop column and hid ERC-721 quantities, leaving a lone `+` or `-`.

Restore the previous left-aligned icon → signed quantity → asset name layout on desktop and mobile while keeping the current typography. Long names can wrap, and NFTs show their returned quantity. The reported empty-metadata Uniswap position now displays `+1  Uniswap v4 Positions NFT`. Fungible assets continue to use parsed amounts.

Issue: [OK-62907](https://onekeyhq.atlassian.net/browse/OK-62907)

## Validation

- Simulation utility tests: 46 passed, including the empty-metadata NFT regression, outgoing NFT quantity, and raw-unit guards.
- `yarn agent:check --profile commit`: passed on the committed changes.
- Tamagui static rendering confirmed the web wrapping CSS is generated.
- Desktop and mobile visual verification is pending.

[OK-62907]: https://onekeyhq.atlassian.net/browse/OK-62907?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/SignatureConfirm/components/SecurityCheckCard/TransactionPreview.tsx
+++ b/packages/kit/src/views/SignatureConfirm/components/SecurityCheckCard/TransactionPreview.tsx
@@ -29,19 +29,6 @@ import {
 
 import type { ISimulationAsset, ISimulationGroup } from './utils';
 
-const DESKTOP_ASSET_LIST_STYLE = {
-  alignSelf: 'flex-start',
-  width: 'auto',
-} as const;
-const DESKTOP_ASSET_ROW_STYLE = { justifyContent: 'flex-start' } as const;
-const DESKTOP_NAME_SLOT_STYLE = { flexGrow: 0, flexShrink: 0 } as const;
-const DESKTOP_NAME_TEXT_STYLE = {
-  width: 56,
-  flexGrow: 0,
-  flexShrink: 0,
-} as const;
-const DESKTOP_AMOUNT_STYLE = { textAlign: 'left' } as const;
-
 type IProps = {
   simulationComponents?: IDisplayComponentSimulation[];
 };
@@ -60,21 +47,18 @@ function SignGuardMark() {
 function SimulationAssetText({ asset }: { asset: ISimulationAsset }) {
   const amount = getSimulationAssetAmount(asset);
   const direction = getSimulationAssetDirection(asset);
-  // Assets.tsx renders the direction sign unconditionally and hides only the
-  // numeric amount for non-ERC1155 NFTs — keep the lone '-'/'+' so an outgoing
-  // unique NFT still reads as leaving the wallet.
   const sign = getSimulationAssetSign(asset);
   const color = direction === ETransferDirection.In ? '$textSuccess' : '$text';
   return (
     <SizableText
       size="$bodyMdMedium"
       color={color}
-      numberOfLines={1}
-      textAlign="right"
-      flexShrink={0}
-      $gtMd={DESKTOP_AMOUNT_STYLE}
+      $platform-web={{
+        wordBreak: 'break-all',
+        overflowWrap: 'break-word',
+      }}
     >
-      {`${sign}${amount}`}
+      {`${sign}${amount}  ${getSimulationAssetLabel(asset)}`}
     </SizableText>
   );
 }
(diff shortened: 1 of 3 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-f54c693d-39ec-50ee-827f-42527ff3e8f4)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Dd7D4XBVmkkhm1Wt4DNvqq

packages/kit/src/views/SignatureConfirm/components/SecurityCheckCard/TransactionPreview.tsx
```diff
@@ -60,21 +47,18 @@ function SignGuardMark() {
...
-      textAlign="right"
-      flexShrink={0}
-      $gtMd={DESKTOP_AMOUNT_STYLE}
+      $platform-web={{
+        wordBreak: 'break-all',
+        overflowWrap: 'break-word',
+      }}
     >
-      {`${sign}${amount}`}
+      {`${sign}${amount}  ${getSimulationAssetLabel(asset)}`}
```
sidmorizon, line 61: # 🟡 P2 Merged asset row has no line cap, so a third-party asset name can grow the signing card without bound

## Problem

Merging the amount and the asset name into one `SizableText` dropped **both** `numberOfLines={1}` props that previously bounded this row:

- the amount text was `numberOfLines={1} textAlign="right" flexShrink={0}`
- the name text was `<SizableText size="$bodyMdMedium" color="$text" numberOfLines={1}>`

What replaces them is `$platform-web={{ wordBreak: 'break-all', overflowWrap: 'break-word' }}` (lines 56–59). Those are breaking *hints*, not a cap — they choose where a long string wraps, not how many lines it may occupy. `SizableText` has no default `numberOfLines`, so on every platform the row now renders as many lines as the string needs.

The string is `` `${sign}${amount}  ${getSimulationAssetLabel(asset)}` `` (line 61), and `getSimulationAssetLabel` returns raw third-party data with no length bound — `token.info.symbol`, `nft.metadata?.name`, `nft.collectionName`, or `name`/`symbol` for internal assets (`utils.ts:125–136`). An ERC-20 whose `symbol()` returns several hundred characters, or an NFT collection named to flood, expands this row without limit. Nothing upstream caps asset count or card height either (`TransactionPreview.tsx:186–205`).

This is now the only uncapped text in the card family — `TransactionPreview.tsx:85` (network name), `TransactionPreview.tsx:105` (group label), and `SecurityCheckCard.tsx:289,356` all still use `numberOfLines={1}`.

## Impact

On a transaction confirmation surface, an attacker-controlled asset name can push `TxConfirmDetails` — recipient, network fee, advanced settings (`TxConfirm.tsx:539–543`) — arbitrarily far down the scroll, so the user approves after scrolling past adversarial content instead of reading the details next to the preview.

Two things keep this at P2 rather than P1: the sign and amount are concatenated *first*, so the financially load-bearing part of the string is never what gets pushed onto a later line; and `TxConfirmAlert` / `SecurityCheckCard` render *above* the preview (`TxConfirm.tsx:516–533`), so risk warnings are not displaced.

Note that wrapping itself is clearly intended here ("Long names can wrap") — the gap is that the wrap is unbounded, not that it wraps.

## Recommendation

Put a cap back on the combined text at line 53 — `numberOfLines={2}` preserves the intended wrap for realistic long names while bounding the adversarial case — and keep `$platform-web` `wordBreak`/`overflowWrap` as the web-side breaking hint for long-but-capped strings. (That `$platform-web` spelling is fine as written; `TokenDetailsHeader.tsx:200–205` uses the same pattern on a `SizableText`.)

Alternatively, bound `getSimulationAssetLabel`'s return length in `utils.ts:125–136` so every consumer inherits the limit.

[Claude session ↗](https://claude.ai/code/session_01Dd7D4XBVmkkhm1Wt4DNvqq)

---

originalix (approved):

PR #13332, opened 2026-09-11
fix: align Perps empty states and restore pagination keyboard toolbar (OK-56261, OK-62884)

Kahnchan:
[OK-56261](<https://onekeyhq.atlassian.net/browse/OK-56261>) [OK-62884](<https://onekeyhq.atlassian.net/browse/OK-62884>)

----------

## Changes

- Fix desktop and wide-screen Perps empty states shifting between tabs. Render empty content independently of table column widths and compensate for TWAP sub-tabs; preserve the mobile empty-state layout.
- Restore the iOS pagination keyboard accessory by assigning each input its own accessory ID. Keep the native page-number caret at the end for appending and deleting digits.

## Validation

- `yarn agent:check --profile pr` passed locally.
- iOS 26.5 Simulator: confirmed the page count and Done toolbar appears, pagination updates, and tapping the left edge then pressing backspace deletes the final digit.
- iOS 18.3.1 Simulator: app startup verified; keyboard interactions still need comparison. Keyboard dismissal after Done remains to be verified.

## Issues

- [OK-56261](https://onekeyhq.atlassian.net/browse/OK-56261)
- [OK-62884](https://onekeyhq.atlassian.net/browse/OK-62884)

[OK-56261]: https://onekeyhq.atlassian.net/browse/OK-56261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62884]: https://onekeyhq.atlassian.net/browse/OK-62884?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix (dismissed):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-7dbea981-1d8e-5177-83ec-200e216ac592)

sidmorizon:
Claude review session: https://claude.ai/code/session_01YHmAq4xbVNaFtX1v5vquoc

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/PerpDesktopEmptyState.tsx
```diff
@@ -10,7 +10,7 @@ export function PerpDesktopEmptyState({
   return (
     <YStack
       flex={1}
-      justifyContent="center"
+      justifyContent="flex-start"
```
sidmorizon: 🟡 P2: Deposit tracker empty state pulled off the top and clipped

Severity: `non-severe`

The shared empty state now top-aligns content via `justifyContent="flex-start"`. The deposit tracker still passes `contentOffsetY={-96}`, so its content is translated 96px above the top instead of above center. The illustration and title overlap the intro card and clip. See [UnifoldTrackerContent](packages/kit/src/views/Perp/components/TradingPanel/modals/UnifoldDeposit/UnifoldTrackerContent.tsx:960-965).

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/PerpPositionsEmptyState.tsx
```diff
@@ -134,7 +134,7 @@ export function PerpPositionsEmptyState({ isMobile }: { isMobile?: boolean }) {
   return (
     <YStack
       flex={1}
-      justifyContent="center"
+      justifyContent={isMobile ? 'center' : 'flex-start'}
```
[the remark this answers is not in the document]
Kahnchan: Fixed: the desktop positions early-return branch now uses the same shared 40px top inset as CommonTableListView. Mobile spacing is unchanged.

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/PerpDesktopEmptyState.tsx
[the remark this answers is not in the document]
Kahnchan: Fixed: restored centered alignment as the default for PerpDesktopEmptyState. Only table and TWAP callers opt into top alignment, so Unifold keeps its centered layout and existing offset.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -1040,7 +1046,18 @@ export function CommonTableListView<T>({
         nestedScrollEnabled
         showsVerticalScrollIndicator
       >
-        {desktopTable}
+        {showDesktopEmptyState ? (
```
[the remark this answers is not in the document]
Kahnchan: Fixed: entering the desktop empty state now resets the header horizontal offset to zero, matching the body scroller when data returns.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -1040,7 +1046,18 @@ export function CommonTableListView<T>({
...
+          // Anchor empty content independently of column widths, fixed columns,
+          // pagination and optional sub-tabs above the table.
+          <YStack
+            testID="perp-desktop-table-empty"
+            pt={Math.max(0, 40 - desktopEmptyStateTopInset)}
```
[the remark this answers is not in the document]
Kahnchan: Fixed: the outer scroll content now grows to fill the viewport and the empty-state wrapper flexes to fill that space, preserving the available height for custom Failed/Retry content.

packages/kit/src/views/Perp/components/OrderInfoPanel/Components/PerpDesktopEmptyState.tsx
Kahnchan: Fixed: restored centered alignment as the default for PerpDesktopEmptyState. Table callers explicitly opt into top alignment; the deposit tracker retains its original centered layout with contentOffsetY=-96.

originalix (dismissed):

Minnzen (dismissed):

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -1034,13 +1046,29 @@ export function CommonTableListView<T>({
     <YStack flex={1} testID="perp-desktop-table">
       {desktopHeader}
       <ScrollView
+        contentContainerStyle={{ flexGrow: 1 }}
```
sidmorizon: # 🟠 P1 `flexGrow: 1` leaks into the non-empty branch and stretches every desktop table row

## Problem

`contentContainerStyle={{ flexGrow: 1 }}` is unconditional, but this `ScrollView` hosts **both** arms of the ternary below it — the empty state (`:1056-1068`) *and* `desktopTable` (`:1070`). Only the empty arm needs it.

Adding `flexGrow: 1` gives the content container a definite height (previously it was content-sized, which is exactly why this idiom is used to fill a short `ScrollView`). That definiteness now propagates all the way down the non-empty chain:

- `:1051` `desktopTable` → `<XStack flex={1}>` (`:951`)
- `:952-967` horizontal body `ScrollView` `style={{ flex: 1 }}`, cross-axis stretch
- `:969` `<YStack flex={1} minWidth={scrollableMinWidth}>`
- `:970` / `:1021` `<YStack flex={1} pb={…}>` — the direct parent of the rows

Every desktop row root is itself `flex={1}`, i.e. `flex-basis: 0`:

- `Components/PositionsRow.tsx:700-706`
- `Components/OpenOrdersRow.tsx:414-415`
- `Components/TradesHistoryRow.tsx:349-350`
- `List/PerpTwapList.tsx:407-408`, `:703-704`, `:1026-1027`

`DebugRenderTracker` returns `props.children` unchanged in production (`packages/components/src/utils/DebugRenderTracker.tsx:36-38`) and `Fragment` emits no box, so those row `XStack`s are direct flex children. With a definite parent height and `flex-basis: 0`, the rows **split the pane height equally** instead of keeping their natural `py="$1.5"` height.

## Impact

Affects the normal, non-empty case on every Perps desktop table — Positions, Open Orders, TWAP (×3), Trades History, Funding, Account, Balances — whenever the row stack is shorter than the panel. With one open position in a ~300px order-info pane the single row grows from ~44px to the full pane: contents float vertically centred (`alignItems="center"`), the alternating `$bgSubdued`/`$bgApp` banding becomes full-height blocks, and the row hover highlight covers the entire pane. Columns stay aligned (both the scrollable and fixed sections stretch together), so the damage is purely vertical — but it lands on the shared path this PR did not intend to touch, to fix a problem that only exists on the empty path.

## Recommendation

Scope the growth to the branch that needs it — the two arms are mutually exclusive, so this fully preserves the empty-state fill that `flex={1}` on `perp-desktop-table-empty` (`:1061`) depends on:

```tsx
contentContainerStyle={showDesktopEmptyState ? { flexGrow: 1 } : undefined}
```

This is already the idiom used for the mobile list in this same file (`:810-813`):

```tsx
contentContainerStyle={{
  flexGrow: paginatedData.length === 0 ? 1 : undefined,
  paddingBottom: enablePagination && totalPages > 1 ? 0 : 16,
}}
```

Avoid `minHeight="100%"` as the alternative — a percentage min-height against an auto-height content container resolves to `0`.

[Claude session ↗](https://claude.ai/code/session_01YHmAq4xbVNaFtX1v5vquoc)

---

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -1034,13 +1046,29 @@ export function CommonTableListView<T>({
...
         style={{
           flex: 1,
         }}
         nestedScrollEnabled
         showsVerticalScrollIndicator
       >
-        {desktopTable}
+        {showDesktopEmptyState ? (
```
sidmorizon, line 1059: # 🟡 P2 Branch swap still leaves `useFixedColumnShadow` bound to a detached scroller

## Problem

The header-offset reset added at `:603-607` fixes one consequence of this ternary, but the same unmount/remount breaks the fixed-column shadow, and that half is untracked.

`scrollViewRef` is attached only inside `desktopTable` (`:952-953`), which this ternary moved into the *else* arm — so the body scroller is now destroyed and recreated as the list empties and refills. `useFixedColumnShadow` cannot survive that:

```ts
// packages/kit/src/hooks/useFixedColumnShadow.ts:144-161
const element = getScrollElement();
if (!element) return;              // bails, no observer registered
handleWebScroll();
const resizeObserver = new ResizeObserver(handleWebScroll);
resizeObserver.observe(element);
…
}, [enabled, handleWebScroll, getScrollElement, isNative, forceVisible]);
```

All five deps are stable for a mounted table (`handleWebScroll` ← `[getScrollElement, forceVisible, position]`; `getScrollElement` ← `[isNative]`), so the effect runs once per mount and binds to whatever element existed at that instant. Two consequences:

- **Mounting while empty** (no open orders, an empty TWAP sub-tab, a stale page index) → `getScrollElement()` is `null` → it bails with no observer and no measurement, and `shadowTransitionEnabled` stays `false`. When rows arrive the effect does not re-run, so `showShadow` keeps its seeded `initialVisible: true` (`:522`) and the right fixed column paints a separator shadow even in a wide window with nothing to scroll.
- **non-empty → empty → non-empty** leaves the `ResizeObserver` attached to a detached node. The remounted scroller is never observed, so column-width and window resizes stop updating the shadow, and a user who had scrolled to the right edge (shadow correctly hidden) gets none back after the remount resets `scrollLeft` to 0.

Before this PR `desktopTable` was always mounted — `c239fef9` removed `{showDesktopEmptyState ? desktopEmptyComponent : null}` from inside it and introduced this ternary — so the element resolved at mount lived for the component's lifetime.

## Impact

Decorative only (a 4px fixed-column separator shadow), but it stays wrong until an unrelated horizontal scroll or a window-width crossing of 768px flips `forceVisible` and re-runs the effect. Applies to every desktop table with fixed columns: Positions, Open Orders, TWAP, Trades History.

## Recommendation

Give the hook a remount signal. Add an optional `resetKey?: unknown` to `IUseFixedColumnShadowOptions`, append it to the deps at `useFixedColumnShadow.ts:161`, and pass `resetKey: showDesktopEmptyState` from the call site at `CommonTableListView.tsx:519-523`.

A callback ref (`setScrollNode` in state rather than `useRef`) is the more robust variant and would also cover any future conditional mount of the scroller. Alternatively, keeping `desktopTable` permanently mounted and rendering the empty content as an absolutely-positioned overlay removes this class of problem entirely — and would also retire the now-unreachable placeholder at `:1023-1025`, whose condition is verbatim `showDesktopEmptyState` (`:601-602`) and so can never be true inside `desktopTable`.

[Claude session ↗](https://claude.ai/code/session_01YHmAq4xbVNaFtX1v5vquoc)

---

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Kahnchan: Fixed: flexGrow is now applied only when showDesktopEmptyState is true. The non-empty table retains its original content-sized scroll container and natural row heights.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Kahnchan: Fixed: shadow management is now enabled only while the desktop body scroller is mounted (hasFixedColumns && !isMobile && !showDesktopEmptyState). The existing enabled dependency disconnects the old ResizeObserver on empty state and remeasures/rebinds it after the body remounts, including when the table starts empty.

originalix (approved):

PR #13333, opened 2026-09-11
fix: align market stars and paginate stock selector｜OK-62872&OK-62891

zy2000-star:
[OK-62872](<https://onekeyhq.atlassian.net/browse/OK-62872>) [OK-62891](<https://onekeyhq.atlassian.net/browse/OK-62891>)

----------

## Summary

- Fix inconsistent star icon sizing in market stock and top coin lists.
- Add cursor-based pagination to the market stock selector.
- Support pagination for both stock search results and the default stock list.
- Prevent duplicate stocks during page loading.
- Add loading, retry, and end-of-list states.

## Related Issues

- OK-62872
- OK-62891

## Validation

- Targeted Jest tests passed: 41 tests.
- Oxlint and Oxfmt checks passed.
- Desktop app started successfully and rendered the Market detail page.
- Full TypeScript check remains blocked by two existing dependency errors:
  - Missing `react-native-draggable-flatlist/src/context/globalRef`
  - Unsupported `bringToFrontToken` prop on iOS `FullWindowOverlay`

[OK-62872]: https://onekeyhq.atlassian.net/browse/OK-62872?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62891]: https://onekeyhq.atlassian.net/browse/OK-62891?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1481e64f-14ce-549b-89a6-bb068402e099)

sidmorizon:
Claude review session: https://claude.ai/code/session_01EQNqs3qQ3aMY9Ysmir6Ctr

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
```diff
@@ -103,7 +132,7 @@ const MarketStockSelectorList = memo(
       );
     }
 
-    if (result.items.length === 0) {
+    if (items.length === 0) {
```
sidmorizon: # ⚠️ [P1] Zero-item early return dead-ends pagination

## Problem

`loadMore()` has exactly one runtime trigger: `Table`'s `onEndReached` (`MarketStockSelectorList.tsx:173` → `handleEndReached` at `:57-61`). But this early return at `:135` unmounts the `Table` entirely whenever `items.length === 0`:

```tsx
if (items.length === 0) {
  return (
    <YStack height={STOCK_SELECTOR_TABLE_HEIGHT} ...>
      <Empty illustration="QuestionMark" title={... global_no_results} />
```

Meanwhile the hook deliberately derives "more pages exist" from the cursor alone, independent of how many rows came back (`useMarketStockSelectorList.ts:177`):

```ts
canLoadMore: Boolean(nextCursor) && hasCurrentData && !isLoading,
```

So a first page of `{ items: [], nextCursor: 'abc' }` produces `canLoadMore === true` with no mounted component able to act on it. That response shape is not rejected anywhere: `IMarketStockPublicListResponse` (`packages/shared/types/marketV2.ts:803-805`) makes `items` and `nextCursor` independent, and neither `fetchMarketStockList` nor `searchMarketStocks` (`ServiceMarketV2.ts:1125-1177`) normalizes it — both return `response.data.data` verbatim. Returning a short or empty window plus a continuation cursor is ordinary cursor-pagination behavior (server-side filtering applied after the window is taken).

Note this is a regression created by the migration: the pre-PR code fetched a single `limit: 50` page, so "no rows" and "no more pages" were necessarily the same condition. Splitting them into 20-row cursor pages breaks that equivalence, but `:135` still assumes it.

## Impact

The selector renders a terminal "No results" for a query that does have matches, with no spinner, no retry affordance, and no scroll surface — the footer's retry/end states at `:62-95` are unreachable too, since they live inside the unmounted `Table`. The only user recovery is retyping the query, which re-runs the identical request and reproduces the same dead end.

## Recommendation

The sibling implementation this was forked from already handles it: `MarketStockList.tsx` mounts the `Table` unconditionally and routes the empty state through `TableEmptyComponent` (`MarketStockList.tsx:197-224`, passed at `:300`), keeping `onEndReached` and the footer alive at zero rows. Either mirror that — move the `Empty` at `:142-145` into a `TableEmptyComponent` prop and delete this early return — or, minimally, stop claiming "no results" while the hook says another page is pending:

```tsx
if (items.length === 0 && !canLoadMore) {
```

`canLoadMore` is already destructured at `:53`, so the narrow fix is a one-line change. Worth adding a hook-level case to `useMarketStockSelectorList.test.tsx` covering `items: []` with a non-null `nextCursor`, since no current test exercises that boundary.

[Claude session ↗](https://claude.ai/code/session_01EQNqs3qQ3aMY9Ysmir6Ctr)

---

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/useMarketStockSelectorList.ts
```diff
@@ -0,0 +1,181 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
```
sidmorizon, line 1: # ⚠️ [P1] New hook missing from mobile module-id registry

## Problem

This file is new in this PR, but `apps/mobile/bundle-registry/module-id-registry.json` was not updated to match. The native union build enforces that registry and fails hard — `Native startup graph budget (24.x)` is red on `c6650d1` for exactly this reason:

```
packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/useMarketStockSelectorList.ts.
```

It reaches this only after both graphs build successfully (main 14688 modules, background 14643), so it is a registry-completeness gate, not a graph-size regression.

This is not optional for this directory — every other non-test module beside it is already registered, including the plain `.ts` ones:

```
packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/constants.ts
packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/navigateToMarketTokenDetail.ts
packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/useLiveTokenOverride.ts
packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
```

`useMarketStockSelectorList.ts` is the only new non-test file in the PR (`git diff --diff-filter=A` against the merge base returns it plus `useMarketStockSelectorList.test.tsx`), and the registry contains zero `useMarketStockSelectorList` entries. The `.test.tsx` needs no entry — no test file appears in the registry.

## Impact

The iOS/Android bundle cannot be produced from this branch: `buildSegmentAllocation` throws before segment allocation completes, so no allocation report or budget report is emitted (the `bundle-architecture-reports` artifact upload found no files). This blocks the native build gate regardless of the rest of the PR, and it will not clear on a re-run — the registry is a committed file, so the failure is deterministic until it is updated.

## Recommendation

```
```

(`apps/mobile/package.json:34` → `node scripts/module-id-registry.js update`; `module-id:check` at `:35` reproduces the gate locally before pushing.) Regenerate the registry with that tooling rather than hand-editing the JSON, so the assigned module id stays consistent with the rest of the file.

Worth noting for the PR description: the stated validation covered Jest, oxlint/oxfmt, and a desktop launch, none of which exercise this gate — the native union build is the only check that does.

[Claude session ↗](https://claude.ai/code/session_01EQNqs3qQ3aMY9Ysmir6Ctr)

---

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
zy2000-star: Fixed

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/useMarketStockSelectorList.ts
zy2000-star: Fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -0,0 +1,42 @@
+// Keep this function self-contained: native charts embed it in the WebView.
+export function formatChartPrice(price: number, maxCharacters = 8): string {
```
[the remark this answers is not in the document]
zy2000-star: Fixed

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -0,0 +1,42 @@
...
+  let body: string;
+  if (decimalPosition <= 0) {
+    body = `0.${'0'.repeat(-decimalPosition)}${digits}`;
+  } else {
+    body =
+      decimalPosition >= digits.length
+        ? digits + '0'.repeat(decimalPosition - digits.length)
+        : `${digits.slice(0, decimalPosition)}.${digits.slice(decimalPosition)}`;
+  }
+  if (value > 0 && !unit && exponent < 0 && body.length > limit) {
```
sidmorizon: # 📝 [P2] Subscript branch fires with zero leading zeros

## Problem

The zero-compaction guard triggers on `exponent < 0` — i.e. *any* value below 1 — rather than on there actually being leading zeros worth compressing:

```ts
if (value > 0 && !unit && exponent < 0 && body.length > limit) {
  const zeros = String(-exponent - 1).replace(/[0-9]/g, (digit) =>
    '₀₁₂₃₄₅₆₇₈₉'.charAt(Number(digit)),
  );
  body = `0.0${zeros}${digits}`;
}
```

For `exponent === -1` the count is `-(-1) - 1 === 0`, so it emits a literal subscript **zero**; for `exponent === -2` it emits `₁`. The resulting string is longer than the plain decimal it replaced, so the truncation at `:38-41` then cuts deeper.

To be clear about what is *not* wrong: the notation stays arithmetically consistent (`0.0₀123` = zero repeated 0 times = `0.123`), so no value is misrepresented. The cost is display quality and precision. Running the committed `.text-js` copy directly:

```
0.123456789   ->  $0.0₀12...     (plain truncation would give  $0.1234...)
0.99999999    ->  $0.0₀99...     (plain truncation would give  $0.9999...)
0.0123456789  ->  $0.0₁12...     (plain truncation would give  $0.0123...)
```

Each loses two significant digits versus the plain form, and `0.0₀` is not a notation a reader will recognise — the subscript exists precisely to stand in for a run of zeros, so a run of length zero is degenerate.

This also contradicts the module's own documented contract. The test at `formatChartPrice.test.ts:6` is titled *"compresses only more than five leading decimal zeros"*, and `:9` pins the no-compression case — but nothing exercises `0.01 ≤ value < 1` with an overflowing body, so the guard's actual range is untested.

## Impact

Sub-$1 prices with more than ~6 decimal digits render on the price axis and crosshair with two significant digits instead of four, in a degenerate subscript form. Both platforms are affected: web calls `formatChartPrice` directly from `StockPriceLineChart.tsx`, native executes the mirrored `.text-js` copy injected via `htmlTemplate.ts`. It is a readability/precision regression rather than a wrong number, hence P2.

## Recommendation

Compute the zero count first and require at least two before compacting:

```ts
const zeroCount = -exponent - 1;
if (value > 0 && !unit && zeroCount >= 2 && body.length > limit) { ... }
```

`0.99999999` then falls back to `$0.9999...` and `0.0123456789` to `$0.0123...`. The existing assertions at `formatChartPrice.test.ts:7-10, 20, 26-27` all use zero counts of 5, 6, 15 and 29, so none of them change.

Apply the same edit to `formatChartPrice.text-js:28` — the two copies are hand-mirrored with nothing pinning them to each other (`formatChartPrice.test.ts:50` evaluates `formatChartPrice.toString()`, the TS module, never the committed `.text-js`; `lightweightChartsRuntime.test.ts` asserts only two outputs against the mirror). They are byte-equivalent in behaviour today, so this is the moment to keep them that way. Worth adding one assertion that evaluates the committed `.text-js` text and compares it against the module across the existing case table, so a future one-sided edit fails the build instead of silently diverging native from web.

[Claude session ↗](https://claude.ai/code/session_01EQNqs3qQ3aMY9Ysmir6Ctr)

---

zy2000-star: Fixed

originalix:

Review of commit `85b95d31d72c` did not finish successfully.

originalix (approved):

PR #13306, opened 2026-09-10
fix: compact chart axis and floating price labels (OK-61390)

limichange:
[OK-61390](<https://onekeyhq.atlassian.net/browse/OK-61390>)

----------

Sub-indicators that cross zero reserve excess Y-axis width for theoretical scientific-notation labels. Small volume-style indicator values also render long decimals. Remove that unconditional reservation and compact four or more leading fractional zeros, for example `0.0000319479` becomes `0.0₄319479`, while preserving six significant digits for volume indicators.

Current-price and Y-axis crosshair label backgrounds fit their rendered text with equal left and right padding. Short labels start at the axis boundary; long labels extend toward the chart to remain within the canvas. Axis measurement follows the same number formatting.

Related: OK-61390

Validation:
- Chart layout and sub-indicator coordinates: 66 tests passed. Shared chart scene: 29 tests passed, including short and long labels with different axis widths.
- `yarn agent:check --profile commit` passed.
- `yarn agent:check --profile pr --pr 13306`: local checks passed; GitHub review is still required.

Visual follow-up: remaining overall Y-axis spacing still needs verification on iOS/Android.

[OK-61390]: https://onekeyhq.atlassian.net/browse/OK-61390?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13334, opened 2026-09-11
fix: prevent Android TradingView price worklet crash(OK-62918)

huhuanming:
[OK-62918](<https://onekeyhq.atlassian.net/browse/OK-62918>)

----------

## Summary
- keep the TradingView Native price formatter default precision self-contained in its serialized worklet
- document why the literal must remain manually synchronized with `PRICE_SIGNIFICANT_FRACTION_DIGITS`

## Root cause
Worklets evaluates default parameters before the function body initializes captured values from `this.__closure`. When callers omitted the precision argument, the serialized UI-runtime function tried to resolve `PRICE_SIGNIFICANT_FRACTION_DIGITS` as a global and threw a fatal `ReferenceError` on Android.

## Validation
- existing `chartLayout.test.ts` suite: 46 tests passed, with no test-file changes in this PR
- reinstalled and launched on physical Android device; requester verified the original crash flow no longer crashes

[OK-62918]: https://onekeyhq.atlassian.net/browse/OK-62918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/components/TradingView/TradingViewNative/utils/chartLayout.ts
+++ b/packages/kit/src/components/TradingView/TradingViewNative/utils/chartLayout.ts
@@ -231,7 +231,9 @@ function compactTradingViewNativePriceLeadingZeros(value: string) {
 
 export function formatTradingViewNativePriceTick(
   price: number,
-  significantFractionDigits: 4 | 6 = PRICE_SIGNIFICANT_FRACTION_DIGITS,
+  // Worklet default parameters cannot read captured constants before __closure is initialized.
+  // Keep this literal in sync manually with PRICE_SIGNIFICANT_FRACTION_DIGITS.
+  significantFractionDigits: 4 | 6 = 4,
 ) {
   'worklet';
 


sidmorizon:
@codex review

sidmorizon:
@codex security review

wabicai (approved):

weatherstar (approved):

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-00759a03-97a7-5f55-84ef-11952bd7fe54)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Gn44TgZmNoYAZisQuoMNpW

originalix (approved):

packages/kit/src/components/TradingView/TradingViewNative/utils/chartLayout.ts
```diff
@@ -231,7 +231,9 @@ function compactTradingViewNativePriceLeadingZeros(value: string) {
 
 export function formatTradingViewNativePriceTick(
   price: number,
-  significantFractionDigits: 4 | 6 = PRICE_SIGNIFICANT_FRACTION_DIGITS,
+  // Worklet default parameters cannot read captured constants before __closure is initialized.
+  // Keep this literal in sync manually with PRICE_SIGNIFICANT_FRACTION_DIGITS.
+  significantFractionDigits: 4 | 6 = 4,
```
sidmorizon, line 236: # 🟡 [P2] Hand-synced literal breaks a coupling the same file already solves in-worklet-body

## Problem

The worklet limitation is real and the fix is correct at this head (`PRICE_SIGNIFICANT_FRACTION_DIGITS = 4` at `chartLayout.ts:119` matches the new literal `4` at `:236`). But this remedy is the one that severs the compiler-checked link between the tick formatter's default and the constant that sizes the axis the formatter must fit into — and the *same file* already carries the correct in-repo remedy for the identical root cause:

```ts
// chartLayout.ts:957-964, inside a 'worklet' function
'worklet';

// Imported constants used in default parameters are not captured by the
// Native worklet serializer. Resolve them inside the worklet body instead.
const resolvedPriceAxisTickCount =
  priceAxisTickCount ?? TRADING_VIEW_NATIVE_PRICE_AXIS_TICK_COUNT;
const resolvedTimeAxisHeight =
  timeAxisHeight ?? TRADING_VIEW_NATIVE_TIME_AXIS_HEIGHT;
```

The coupling is tightest inside a single function. `getTradingViewNativePriceAxisCandidateLabel` (`:291`) builds the widest-possible price label — the axis width budget — and now reads the number from two independently-mutable places:

- `:296` `formatTradingViewNativePriceTick(price)` — relies on the new hardcoded `4`
- `:304` / `:309` `leadingZeroCount + PRICE_SIGNIFICANT_FRACTION_DIGITS` and `'8'.repeat(PRICE_SIGNIFICANT_FRACTION_DIGITS)` — reads the constant

`getTradingViewNativePlainDecimalPriceAxisLabel` (`:335`) does the same. These labels feed `getTradingViewNativePriceAxisWidth`, which reserves the price-axis gutter, while `:256`/`:258` render the actual tick text through the hardcoded default. After this change only the comment at `:235` keeps them in agreement — `significantFractionDigits: 4 | 6` at `:236` accepts either value completely independently of `:119`.

## Impact

Not a crash and no behavior change today. If `PRICE_SIGNIFICANT_FRACTION_DIGITS` is later changed and `:236` is missed, the reserved gutter and the rendered labels disagree: in the direction where the rendered precision exceeds the reserved width, price-axis labels clip or overlap on native. TypeScript no longer catches it.

The hazard is not hypothetical for this code — the `significantFractionDigits: 4 | 6` parameter itself was only introduced two commits ago (`c5fda454`, #13306), and a second hand-derived copy of the same value already exists on this path (`subIndicatorRender/coordinates.ts:13`, `INHERITED_PRICE_PLAIN_DECIMAL_BOUNDARY = 0.000_888_8`). A test cannot close the gap either: `PRICE_SIGNIFICANT_FRACTION_DIGITS` is not exported, so pinning `default === constant` would mean widening the module's public surface.

## Recommendation

At `chartLayout.ts:232-236`, mirror the body-resolution pattern already shipping at `:961-964`, keeping the explicit `4 | 6` annotation so the compiler — not a comment — enforces the relationship:

```ts
export function formatTradingViewNativePriceTick(
  price: number,
  // Worklet default parameters cannot read captured constants before
  // __closure is initialized. Resolve inside the worklet body instead.
  significantFractionDigits?: 4 | 6,
) {
  'worklet';

  const resolvedSignificantFractionDigits: 4 | 6 =
    significantFractionDigits ?? PRICE_SIGNIFICANT_FRACTION_DIGITS;
```

then use `resolvedSignificantFractionDigits` at `:256` and `:258`. Without the explicit `: 4 | 6` annotation TS would just widen the union, so that annotation is what restores the enforcement.

This is safe on the Android UI runtime path: this function already reads module-scope constants in its own body today (`PRICE_INTEGER_FRACTION_DIGITS` at `:249`, `MAX_TO_FIXED_FRACTION_DIGITS` at `:257`), and `getTradingViewNativeChartLayout` uses the exact `?? CONSTANT` body form and is reached via the same chain (`TradingViewNativeChart.tsx:362` `useDerivedValue` → `createTradingViewNativeSkiaPicture` → `buildTradingViewNativeChartScene` → `getTradingViewNativeChartLayout`). No call-site changes are needed — the only caller passing the argument explicitly is `subIndicatorRender/coordinates.ts:31` (`, 6`), and no caller passes `undefined`.

[Claude session ↗](https://claude.ai/code/session_01Gn44TgZmNoYAZisQuoMNpW)

---

PR #13303, opened 2026-09-10
fix: correct stock charts and estimates (OK-62700, OK-62704, OK-62706)

zy2000-star:
[OK-62700](<https://onekeyhq.atlassian.net/browse/OK-62700>) [OK-62704](<https://onekeyhq.atlassian.net/browse/OK-62704>) [OK-62706](<https://onekeyhq.atlassian.net/browse/OK-62706>)

----------

## Summary

Fix stock chart currency consistency, financial chart number formatting, and missing earnings forecast points.

Related Jira issues:
- [OK-62700](https://onekeyhq.atlassian.net/browse/OK-62700)
- [OK-62704](https://onekeyhq.atlassian.net/browse/OK-62704)
- [OK-62706](https://onekeyhq.atlassian.net/browse/OK-62706)

### OK-62700 — Financial chart number formatting

Financial chart values need consistent formatting across locales and enough space to remain readable at narrow widths.

- Add a shared formatter for financial chart axes and tooltips.
- Use K/M/B suffixes for large values and handle rounding across unit boundaries.
- Preserve significant digits for small decimals, using subscript notation for leading zeros.
- Limit axis labels to 8 characters, or 7 on narrow charts, while retaining unit suffixes when truncating.
- Adjust the chart's right margin for narrow layouts.
- Default the reporting currency to USD when the response omits it.

### OK-62704 — Stock price and K-line axis mismatch

Stock requests previously inherited the application's selected currency through the shared request interceptor. This could produce chart values in a different currency from the USD price displayed on the stock page.

- Explicitly request USD for stock list, search, detail, financials, and chart endpoints.
- Apply the currency override to the chart endpoint used by both the simple price chart and the K-line data flow.
- Add service tests to verify the request currency header.

### OK-62706 — Missing earnings estimates

Two frontend rules could hide available forecast data: annual estimates were suppressed when the analyst count was below 3, and quarterly forecasts without fiscalPeriod were removed by period validation.

- Remove the analyst-count restriction on annual estimates.
- Preserve quarterly estimate-only records when fiscalPeriod is missing.
- Display the supplied date for those forecasts instead of inferring a fiscal quarter.
- Retain date, fiscal-year, and currency validation.
- Keep Actual as filled circles and Estimate as outlined circles.
- Preserve missing values as unavailable rather than substituting zero.

## Validation

- 45 financials tests passed after the final forecast-filtering fix.
- Added regression coverage for:
  - Quarterly forecast rendering when fiscalPeriod is absent.
[...]

PR #13335, opened 2026-09-11
feat: improve stock financial charts and polish market details

erikzou:
Stock financials redesign plus a round of Market detail polish. Everything is scoped to Market, except the wallet home Top Coins tab label.

## Stock financials

- Redesign the four charts: theme tokens instead of fixed hex so dark mode follows, rounded bars (negative bars round downward), a column highlight on hover, hollow markers on the net-margin line, and x-axis labels that wrap instead of truncating.
- Replace the text row under the chart with a floating tooltip anchored to the column. Desktop is hover-only by design; on touch a tap pins/unpins it; Escape closes it on web. The currency moves from the card header into the tooltip.
- Below 560px the revenue-to-profit waterfall collapses to Revenue → COGS → Gross profit → Expenses & adjustments → Net income, and the collapsed bar's tooltip lists what it folds in. Only a reconciled waterfall collapses; missing or unreconciled data stays on the simplified subtotal chart.
- Cards sit in two columns once the content column is at least 800px wide.
- The card header wraps when the title and the Annual/Quarterly switch cannot share a line: the switch moves onto its own line, left-aligned, and the title can wrap too.
- X-axis labels render as platform text below the SVG instead of `SvgText`, so native and browser line breaking handles every script: long German compounds, CJK labels without spaces and Thai/Bengali combining marks wrap correctly with no hand-rolled splitting (web adds `word-break: break-word` for over-wide words).
- The net-margin axis is start-anchored at the card edge so it lines up with the card title, and the two-column layout uses a 40px gap.
- Labels come from `ETranslations` through `useStockFinancialLabels()` (26 `market.stock_financials_*` keys), replacing the temporary hard-coded English. Annual period labels drop the `FY` prefix.

Rebase notes: #13303 and #13333 landed in the meantime. This keeps its `formatFinancialValue` and estimate rules together with the redesign's layout and tooltip. Two of its tests asserted details of the old tooltip, so they now follow the redesign: hollow estimate dots fill with `bgApp` rather than `none`, and the forecast date is asserted on the x-axis label because the tooltip no longer carries a period title. #13333's `getFinancialPerformanceDomain` is kept for the performance chart's axis.

## Stock chart and header

- Drop the simple chart → header price sync: hovering the line no longer rewrites the header price and change. `StockPriceLineChart` keeps `onHoverChange` so a complete version can wire back in later; the in-chart hover card still shows time and price.
- The Market simple chart hover card shows its price at `$bodyMdMedium`, 2px under the time. It is opt-in (`hoverLabelLargePrice`), so the Swap stock chart is unchanged.

## Tooltips

- Hover-intent delay (the component library's interactive Tooltip: 250ms open / 300ms close) on the stock header price, the token price info button, the stock Price column header (list page and the token selector's stock list), and the list tab tooltips.
- The header change goes through `HeaderColumn`; `titleTooltip` is only used by the Market stock columns, so nothing outside Market changes.

## Stock variant selector

- Always opens rightward from the trigger: `allowFlip={{ flipAlignment: false }}` still allows a vertical flip but no longer swaps to end alignment when the right side is tight.
- Width 384 → 376, so on a gutter-limited window it keeps a 12px margin instead of landing inside the popper's 10px frame padding and getting pushed left.
- Columns weighted 1.3 : 1 : 0.9 (name / issuer / price) so a symbol plus its 24/7 tag fits; the Price header loses its decorative underline.

## Labels

- Stock list columns (shared with the token selector's stock list) and the desktop stock detail stat show the latest trading day, not a rolling 24h. "24h change / 24h volume / 24H price range" become the new stock-only keys `market.stock_change__title`, `market.stock_volume__title` and `market.stock_day_range__title`, pulled from Lokalise. The existing keys stay, since Top Coins and the crypto lists still use them.
- Wallet home Top Coins tab: the fallback name is now `market.top_coins` instead of hard-coded English. A server-provided name still wins, as #13129 intended.

## Narrow stock overview (mobile)

- Analyst ratings are hidden on narrow screens for now. This is a product decision; the desktop layout keeps the analyst gauge.
- The About description collapses once it exceeds 200 characters: four lines on mobile, two on desktop, with Show more / Show less. The threshold is now shared (`STOCK_ABOUT_DESCRIPTION_COLLAPSED_LENGTH`).

## Not in this PR

- The mobile stock detail stats still read "24h volume": `useStockSecurityStats` is shared with Perps and is left for a separate change.

## Verification

After rebasing on `x` (`44945aaa14`): `yarn agent:check --profile pr` clean and the Market, Home PopularTrading, Table, Tooltip and Swap suites pass. The follow-up axis label and layout commits pass `yarn agent:check --profile commit` and the financials and stock overview suites.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-fb8e5a1e-b023-561b-8cf4-a0f539b9f42a)

sidmorizon:
Claude review session: https://claude.ai/code/session_01KzwjLwXufgSY2bJcMTXk7F

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -298,23 +365,138 @@ export function FinancialChart({
             tabIndex={0}
             testID={`${testID}-point-${index}`}
             aria-label={`${row.label}: ${row.values.map((value, valueIndex) => `${series[valueIndex]?.label ?? ''} ${formatValue(value, valueIndex === lineIndex)}`).join(', ')}`}
-            onHoverIn={() => setSelected(row.key)}
-            onHoverOut={() => setSelected(undefined)}
-            onPress={() => setSelected(row.key)}
-            onFocus={() => setSelected(row.key)}
-            onBlur={() => setSelected(undefined)}
+            onHoverIn={() => setSelection({ key: row.key, pinned: false })}
+            onHoverOut={() => setSelection(undefined)}
```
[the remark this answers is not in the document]
erikzou: Intentional. The desktop chart is hover-only by design, so leaving the column is meant to hide the tooltip. Press-to-pin exists for touch devices: there is no hover there, a tap is the only way to open the tooltip, and native never fires `onHoverOut`, so a pinned tooltip stays until the next tap. No change.

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/StockFinancials.tsx
```diff
@@ -65,34 +75,51 @@ function FinancialCard({
...
         justifyContent="space-between"
         gap="$2"
-        flexWrap="wrap"
+        minHeight="$10"
       >
-        <YStack gap="$1" flexShrink={1}>
-          <SizableText size="$headingSm">{labels[kind]}</SizableText>
+        <XStack alignItems="baseline" gap="$3" flexShrink={1}>
+          <SizableText size="$headingMd" flexShrink={0}>
+            {labels[kind]}
```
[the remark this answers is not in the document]
erikzou: Fixed in e1179022ba. The card header row now wraps: when the title and the Annual/Quarterly switch cannot share a line, the switch moves onto its own line (left-aligned), and the title itself can wrap instead of overflowing.

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -28,48 +32,98 @@ import { formatFinancialValue } from './financialValueFormat';
...
+
+function wrapLabel(label: string, columnWidth: number) {
+  const maxCharacters = Math.max(5, Math.floor((columnWidth - 4) / 5.5));
+  const lines: string[] = [];
+  label.split(/[\s/]+/).forEach((word) => {
+    const last = lines.at(-1);
+    if (last && last.length + word.length + 1 <= maxCharacters) {
+      lines[lines.length - 1] = `${last} ${word}`;
+    } else {
+      lines.push(word);
```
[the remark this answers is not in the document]
erikzou: Fixed in e1179022ba. `wrapLabel` now breaks any word wider than its column instead of letting it run into the neighboring columns. Width is estimated in 5.5px units with CJK/full-width glyphs counted as two: CJK labels such as `费用及调整项` have no spaces and overflowed the same way as the German compounds. Covered by a new test ("breaks labels wider than their column instead of letting them overlap").

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -28,48 +32,123 @@ import { formatFinancialValue } from './financialValueFormat';
...
+    // running into the neighboring columns.
+    const pieces: string[] = [];
+    let piece = '';
+    Array.from(word).forEach((char) => {
+      if (piece && textUnits(piece) + glyphUnits(char) > maxUnits) {
+        pieces.push(piece);
+        piece = '';
+      }
+      piece += char;
+    });
```
sidmorizon: # [🟠 P1] Forced label breaks split Thai and Bengali grapheme clusters

## Problem
`wrapLabel` breaks an over-wide word per code point (`Array.from(word)` at `FinancialChart.tsx:101-107`), so a forced break can land between a base consonant and its combining vowel or tone mark, and `glyphUnits` (line 84-86) charges those zero-width marks a full unit, which makes the break come earlier than it should.

Running the exact function against the shipped translations:

- 335px card (a 375px phone, compact 5-column waterfall): `กำไรขั้นต้น` (gross profit, `th_TH`) → `["กำไรขั", "้นต้น"]`; second line starts with U+0E49.
- 320px card: `กำไรสุทธิ` (net income) → a line that is only `ิ` (U+0E34).
- 560px card (8-column chart): `รายได้/ค่าใช้จ่ายนอกการดำเนินงาน` → last line `ินงาน`; Bengali `পরিচালন-বহির্ভূত আয়/ব্যয়` (`bn`) → a line starting with `ির্ভূত`.

Sweeping card widths 280-900px in 1px steps: Thai produces a mark-initial line at 135 distinct widths (280-751), Bengali at 104 (560-751). Hindi and Vietnamese are clean.

## Impact
A line that begins with a combining mark renders as a dotted circle plus a detached mark, so the x-axis labels for gross profit, net income, operating expenses and the non-operating row are unreadable in Thai and Bengali at the default mobile width and in the two-column desktop layout. These are the labels the previous overlap fix targeted: the overlap is gone, but the glyphs are now broken instead.

## Recommendation
Iterate grapheme clusters rather than code points in the split loop at lines 101-107. The repo already has `splitGraphemes` (Intl.Segmenter with a code-point fallback) in `packages/kit/src/views/Onboardingv2/pages/GetStarted.tsx:61-76`; lift it to a shared util and use it here. In `glyphUnits`, count marks (`/\p{M}/u`) as 0 units so the width estimate does not over-break, and in the fallback path never start a piece with a mark. Extend the new "breaks labels wider than their column" test with a Thai label at a 7-unit budget asserting that no line matches `/^\p{M}/u`.

[Claude session ↗](https://claude.ai/code/session_01KzwjLwXufgSY2bJcMTXk7F)

---

packages/kit/src/views/Market/MarketDetailV2/components/TokenOverview/StockTokenOverview.tsx
```diff
@@ -101,54 +108,8 @@ export function StockTokenOverview() {
...
-      <Divider my="$1" />
-
       {stockId ? (
-        <StockFinancials
-          stockId={stockId}
-          labels={stockFinancialLabels}
-          withHorizontalPadding={false}
-        />
+        <StockFinancials stockId={stockId} withHorizontalPadding={false} />
       ) : null}
```
sidmorizon, line 113: # [🟠 P1] Mobile stock overview drops the Analyst ratings section without replacement

## Problem
Commit e1179022 deletes the "Analyst ratings" block (consensus plus Buy/Hold/Sell rows) that the mobile overview rendered between the stats and the financials at the merge base (the block around line 107 of `StockTokenOverview.tsx` at 639eb621). After this change `StockAnalystGauge` and `market_stock_analyst_ratings` are referenced only from `layouts/StockDesktopLayout.tsx:919` and `:957`, and `layouts/MobileLayout.tsx:893` mounts only `LazyStockTokenOverview` for stocks, so no mobile path renders analyst data any more. The test was updated to assert the heading is absent (`StockFinancials.test.tsx:610-614`), so the removal is deliberate, but the commit message ("tidy narrow stock overview") and the PR description do not mention it; the description scopes the mobile overview changes to the financials cards.

## Impact
iOS and Android users lose analyst consensus and the rating breakdown for every stock while desktop keeps them. Reviewers reading the PR description will not know a product surface was removed, and the repository rule is to not remove existing components unless the request requires it.

## Recommendation
Either restore the section here (lines 109-113, or mount the existing `StockAnalystGauge` from `components/StockAnalystGauge` in the mobile overview as the desktop equivalent), or state the removal explicitly in the PR description with product sign-off so it is reviewed as a scope change rather than shipped under a wrapping fix.

[Claude session ↗](https://claude.ai/code/session_01KzwjLwXufgSY2bJcMTXk7F)

---

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -298,23 +390,138 @@ export function FinancialChart({
...
+            onHoverOut={() => setSelection(undefined)}
+            onPress={() =>
+              setSelection((current) =>
+                current?.key === row.key && current.pinned
+                  ? undefined
+                  : { key: row.key, pinned: true },
+              )
+            }
+            onFocus={() => setSelection({ key: row.key, pinned: false })}
+            onBlur={() => setSelection(undefined)}
```
sidmorizon, line 358: 🟡 P2: Pinned financial tooltip clears immediately

Severity: `non-severe`

After a point is pressed, `onHoverOut` and `onBlur` clear its pinned `selection`. Moving away or losing focus closes the tooltip.

originalix (dismissed):

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
erikzou: Fixed in 7abca5ce05. `wrapLabel` now breaks between grapheme clusters, and a cluster takes the width of its base character, so combining marks add nothing and never start a line. The splitter is lifted into `stringUtils.splitGraphemes` as suggested, with one change: the `GetStarted` copy fell back to a per-code-point split, which on Hermes without an Intl polyfill would have kept this bug on native. The new fallback attaches combining marks (hand-rolled Thai/Bengali/Devanagari/diacritic ranges, since Hermes lacks reliable `\p{…}`) and joins the consonant after a virama and the character after a ZWJ; `GetStarted` now uses the shared helper. Tests: the chart test adds `กำไรขั้นต้น` at a 9-unit budget (the old split produced `้น`) and asserts no line matches `/^\p{M}/u`; `stringUtils` tests cover both the Segmenter and the fallback path, plus `ক্ষ` conjuncts and ZWJ emoji.

packages/kit/src/views/Market/MarketDetailV2/components/TokenOverview/StockTokenOverview.tsx
erikzou: Intentional: hiding analyst ratings on narrow screens is a product decision for now, and the desktop layout keeps the analyst gauge. It was not called out in the commit message or the description, so the PR description now lists it under "Narrow stock overview (mobile)" as a product decision.

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
erikzou: Intentional, same as the Codex thread on this file: the desktop chart is hover-only by design, so leaving the column or losing focus is meant to hide the tooltip. Press-to-pin is the touch interaction, where native never fires `onHoverOut`, so a pinned tooltip stays until the next tap. No change.

packages/shared/src/utils/stringUtils.ts
```diff
@@ -320,6 +321,81 @@ function stripLineBreaks(value: string) {
...
+function splitGraphemesByCodePoint(text: string): string[] {
+  const graphemes: string[] = [];
+  let joinNext = false;
+  for (const char of text) {
+    const codePoint = char.codePointAt(0) ?? 0;
+    if (graphemes.length > 0 && (joinNext || isCombiningMark(codePoint))) {
+      graphemes[graphemes.length - 1] += char;
+    } else {
+      graphemes.push(char);
+    }
```
sidmorizon: # [🟡 P2] Fallback joiner rule glues the next character unconditionally, including spaces and punctuation

## Problem
In `splitGraphemesByCodePoint` (`stringUtils.ts:361-374`), `joinNext` is set after any virama (U+094D, U+09CD) or ZWJ (U+200D) and the next code point is appended to the previous cluster no matter what it is (line 366-367). ICU only joins when a consonant or emoji actually follows. Running the exact fallback against ICU:

- `ক্ ম` → fallback `["ক্ ", "ম"]`, ICU `["ক্", " ", "ম"]`
- `ক্-ম` → fallback `["ক্-", "ম"]`, ICU `["ক্", "-", "ম"]`
- `ক্!` → fallback `["ক্!"]`, ICU `["ক্", "!"]`
- `👩 💻` → fallback `["👩 ", "💻"]`, ICU `["👩", " ", "💻"]`

None of the currently shipped chart labels or onboarding hero words end a word in a bare virama or ZWJ, so the two paths agree on today's strings.

## Impact
This helper is now a general-purpose export of `@onekeyhq/shared`, and the fallback is the production path on iOS and Android (Hermes has no `Intl.Segmenter` and the repo's intl shim does not polyfill it). A future caller that passes a phrase with a word-final halant or a dangling ZWJ gets a space or punctuation mark welded into the preceding cluster. In `GetStarted.tsx`'s hero animation that would bypass the `' '` to non-breaking-space substitution because the space is no longer its own cluster, and in `wrapLabel` a trailing `-` or `!` would be under-counted as part of a 1-unit cluster.

## Recommendation
Only honor `joinNext` when the following code point is a plausible continuation: skip whitespace and ASCII punctuation at line 366, e.g. `const joins = joinNext && !(codePoint <= 0x7f && !/[A-Za-z0-9]/.test(char));` (kept ASCII-only to stay clear of `\p{…}` on Hermes), then use `joins || isCombiningMark(codePoint)`. Add cases to `stringUtils.test.ts` asserting that a virama or ZWJ followed by a space or punctuation yields separate clusters on the fallback path.

[Claude session ↗](https://claude.ai/code/session_01KzwjLwXufgSY2bJcMTXk7F)

---

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/StockFinancials.test.tsx
```diff
@@ -401,17 +461,145 @@ it('uses the same compact amount units for axes and hover values in every locale
...
+    screen.getByTestId('labels').querySelectorAll('text'),
+    (node) => node.textContent,
+  );
+  expect(lines).toEqual(
+    expect.arrayContaining(['Betriebsf', 'remde', '费用及调', '整项']),
+  );
+  expect(lines).not.toContain('Betriebsfremde');
+  expect(lines).not.toContain('费用及调整项');
+  // A forced break must never leave a combining mark at the start of a line.
+  expect(lines.filter((line) => /^\p{M}/u.test(line ?? ''))).toEqual([]);
```
sidmorizon: # [🟡 P2] The Thai case in the wrap test never forces a break under the new algorithm

## Problem
The fixture `กำไรขั้นต้น` at `StockFinancials.test.tsx:557` is 11 code points but only 7 grapheme clusters, each 1 unit under `graphemeUnits`. With ten columns the budget is 9 units, so the current `wrapLabel` emits the label on a single line and the `/^\p{M}/u` assertion at line 581 is checked against lines that were never split. The case does fail on the previous per-code-point code (which produced `้น`), so it guards against that regression, but it does not exercise the new cluster-boundary break path at all.

## Impact
A future change to the break loop in `FinancialChart.tsx` that again splits inside a cluster, for example counting marks as 1 unit or iterating `Array.from` instead of `splitGraphemes`, would still pass this test as long as short Thai labels fit in one line. The real labels that broke in the previous head (`ค่าใช้จ่ายและรายการปรับปรุง`, `รายได้/ค่าใช้จ่ายนอกการดำเนินงาน`) are longer than any budget the compact chart gets and are exactly the cases this test should pin.

## Recommendation
Use a label that exceeds the budget, e.g. replace the fixture with `ค่าใช้จ่ายและรายการปรับปรุง` (22 clusters, 22 units against a 9-unit budget, so it must break at least twice), keep the `/^\p{M}/u` assertion, and add `expect(lines).not.toContain('ค่าใช้จ่ายและรายการปรับปรุง')` so the test proves both that a break happened and that no piece starts with a mark.

[Claude session ↗](https://claude.ai/code/session_01KzwjLwXufgSY2bJcMTXk7F)

---

packages/shared/src/utils/stringUtils.ts
```diff
@@ -320,6 +321,81 @@ function stripLineBreaks(value: string) {
...
+    joinNext = JOINING_CODE_POINTS.has(codePoint);
```
sidmorizon: 🟡 P2: Hermes splits valid grapheme clusters

Severity: `non-severe`

Without `Intl.Segmenter`, `splitGraphemes` splits regional-indicator pairs and other valid clusters. Flags can animate or wrap as separate characters on native.

packages/shared/src/utils/stringUtils.ts
erikzou: Resolved by removal in 44c26c0124. The hand-rolled splitter is gone: axis labels now render as platform text below the SVG, so native and browser line breaking handles clusters, and `stringUtils` / `GetStarted` are back to their `x` versions. There is no fallback joiner left to fix.

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/StockFinancials.test.tsx
erikzou: Superseded in 44c26c0124: `wrapLabel` and its break loop are removed, and labels render as platform text below the SVG. The test is rewritten to assert that a German compound, a CJK label and a Thai label render intact as text outside the SVG, since breaking them is now the platform's job.

packages/shared/src/utils/stringUtils.ts
erikzou: Removed in 44c26c0124: `splitGraphemes` is no longer exported from `stringUtils`, and the chart no longer splits text at all (labels are platform text). `GetStarted` is back to its own local helper exactly as on `x`; its code-point fallback is pre-existing and unchanged by this PR.

packages/components/src/composite/Table/components/HeaderColumn.tsx
```diff
@@ -138,6 +138,9 @@ function HeaderColumn<T>({
     titleContent = titleTooltip ? (
       <Tooltip
         placement="top"
+        // Interactive mode for its hover-intent delay: sweeping the pointer
+        // along the header row must not flash each column's tooltip.
+        hovering
```
sidmorizon, line 143: 🟡 P2: Hover mode blocks keyboard tooltips

Severity: `non-severe`

When keyboard users focus these triggers, `hovering` ignores the focus-driven open state. The tooltip only opens on mouse enter, hiding its explanation.

erikzou: Verified, and not changed in this PR. Of the five tooltips switched to `hovering` here, only one has a keyboard-focusable trigger: the token price info `Button` (`role: "button"`, `focusable: true`, so react-native-web gives it `tabIndex=0`). The other four — including the `HeaderColumn` trigger this comment points at, plus the header price label and the list tab items — are DashText or plain Stacks with no role and no `focusable`, and `createDOMProps` only makes native `button`/`a`/`input`/`select`/`textarea`, the button/checkbox/link/radio/textbox/switch roles, or explicitly `focusable` elements into tab stops. So there is no focus-driven open to lose at the flagged site.

For the one Button trigger the observation holds: in `hovering` mode `isOpen` reads the hover state and ignores the focus-driven open. That is a property of the shared interactive Tooltip rather than of this PR — `hovering` has 19 call sites on `x` with the same behaviour — so making focus-visible open a `hovering` tooltip belongs in `packages/components/src/actions/Tooltip`, not in this Market change. Keeping the hover-intent delay as is here.

originalix (approved):

PR #13338, opened 2026-09-11
fix: read cached device state in about dialog OK-62935

wabicai:
[OK-62935](<https://onekeyhq.atlassian.net/browse/OK-62935>)

----------

## Summary
- Read persisted `deviceStateInfo` when rendering the About Device dialog.
- Avoid triggering a hardware `getDeviceState` request when the dialog opens.
- Preserve legacy Features and third-party fallback behavior.

## Issue
- OK-62935

## Validation
- Verified disconnected cached Pro2, V1, legacy V1, third-party, stale Features, and missing-state scenarios.

[OK-62935]: https://onekeyhq.atlassian.net/browse/OK-62935?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-65e98eb8-7e5e-53e0-8011-2739c3b43d29)

sidmorizon:
Claude review session: https://claude.ai/code/session_01KRYK3p7duUcGJ6DqQzdXeR

originalix (dismissed):

packages/kit/src/views/DeviceManagement/pages/DeviceDetailsModal/dialog/DialogDeviceAbout.tsx
```diff
@@ -118,15 +117,7 @@ function DialogDeviceSpecsContent({ data }: { data: IHwQrWalletWithDevice }) {
...
-        : await backgroundApiProxy.serviceHardware
-            .getDeviceState({
-              connectId: device.connectId,
-              params: {
-                scope: 'firmware',
-              },
-              silentMode: true,
-            })
-            .catch(() => device.deviceStateInfo);
+        : device.deviceStateInfo;
```
sidmorizon: # 🟠 P1 About dialog bypasses `resolveDeviceState`, so it can print a stale firmware version while the header on the same screen shows the new one

## Problem

Line 120 now reads the raw persisted `device.deviceStateInfo`. Every other consumer of device state in this feature resolves it through `resolveDeviceState({ persistedState, snapshot })` (`packages/kit/src/states/jotai/contexts/deviceDetails/deviceStateManagement.ts:122-130`, i.e. `snapshot?.state ?? persistedState`): `buildDeviceMetaStatic` (`actions.ts:53-58`) — which is otherwise the same `isThirdParty ? undefined : <state>` branch as the new code — plus `buildDeviceMetaState` (`actions.ts:158-161`) and `applyDeviceStateEvent` (`actions.ts:253-256`). This dialog is now the only reader that ignores `deviceStateSnapshotAtom`.

The `device` it receives comes from `actions.getWalletWithDevice()` (`DeviceSectionSupport.tsx:39`) → `walletWithDeviceStateAtom` (`actions.ts:402-404`). `refresh()` sets that atom at `actions.ts:334` from the `getAllHwQrWalletWithDevice` DB read taken at `actions.ts:315` — *before* the live read at `actions.ts:349`, whose fresher result lands only in `deviceStateSnapshotAtom` (`actions.ts:358-371`). That DB read is explicitly treated as possibly-stale in this codebase: `getAllDevices()` is served from a short-lived LRU cache, and `pickNewerDeviceStateSnapshot` (`deviceStateManagement.ts:132-160`) exists specifically to stop such a read from regressing a newer snapshot ("Refresh reads can be served from short-lived DB record caches that may predate the write"). The dialog now reads the very value that guard was written to shadow.

## Impact

Post-firmware-update flow: `DeviceDetailsModal/index.tsx:148` runs `refresh(walletId, { refreshFirmwareInfo: true })`. `walletWithDeviceStateAtom` receives the pre-update DB record, while the fresh firmware versions go to `deviceStateSnapshotAtom`. The page header then renders the **new** version (`DeviceBasicInfo.tsx:174` ← `deviceMetaStatic`, snapshot-resolved), but opening "About device" on that same screen renders `getDeviceVersionsFromState({ state: device.deviceStateInfo })` (line 129) — the **old** firmware/bootloader/BLE version, along with a potentially stale `serialNo` / `bleName` / `deviceType`. Before this PR the dialog performed its own live `getDeviceState` read and therefore could not lag the header.

The divergence does not reliably self-heal: the focus refresh uses `skipDeviceStateSnapshot: true` (`index.tsx:137`), and on that path `pickNewerDeviceStateSnapshot` (`actions.ts:372-380`) deliberately keeps the newer snapshot over the older DB copy. So the wallet atom's `deviceStateInfo` can remain behind while the header stays correct, and the two surfaces keep disagreeing about the user's actual firmware version.

## Recommendation

Keep the intended "no hardware call when the dialog opens" behavior, but feed the dialog the same *resolved* state the rest of the page uses instead of the raw persisted field. The dialog is shown via `Dialog.show()` outside `ProviderJotaiContextDeviceDetails`, so resolve it at the call site: add a context action returning `resolveDeviceState({ persistedState: get(walletWithDeviceStateAtom())?.device?.deviceStateInfo, snapshot: get(deviceStateSnapshotAtom()) })` (both atoms are already exported at `deviceDetails/atoms.ts:31-34`), call it next to `getWalletWithDevice()` in `DeviceSectionSupport.tsx:38-43`, pass it into `showDialogDeviceAbout`, and use it in place of `device.deviceStateInfo` at `DialogDeviceAbout.tsx:120`. The `else` fallback to `deviceUtils.getDeviceVersion()` at line 131 stays correct as-is.

[Claude session ↗](https://claude.ai/code/session_01KRYK3p7duUcGJ6DqQzdXeR)

---

wabicai:
已处理 review 反馈。

About Device 现在在 `DeviceSectionSupport` 中读取 `deviceStateSnapshotAtom`，通过 `resolveDeviceState` 选择最新 snapshot，并显式传入弹窗；没有 snapshot 时回退到持久化 `deviceStateInfo`。弹窗打开仍不会发起硬件查询。

已重新通过 `yarn agent:check --profile commit`，并推送 commit `d6f9c5f006`。

originalix (approved):

PR #13339, opened 2026-09-11
fix: preserve Bluetooth unpaired errors (OK-62230)

originalix:
[OK-62230](<https://onekeyhq.atlassian.net/browse/OK-62230>)

----------

BLE pairing failures reported as error 705 (`BleDeviceNotBonded`) could be replaced by “Device disconnected” when DeviceStage checked the already-disconnected transport. Preserve the existing localized “Bluetooth unpaired” / “蓝牙未配对” message so the user sees the pairing failure.

This is an App-only change in the shared DeviceStage error handler. It adds no retry, SDK, dependency, or translation changes; protocol V2 errors reported as 720 retain their existing behavior. A focused regression covers error serialization across the UI/background boundary and a disconnected transport, including avoiding a duplicate toast.

Issue: [OK-62230](https://onekeyhq.atlassian.net/browse/OK-62230)

Validation:
- DeviceStageBurst and useDeviceStageBurst: 2 suites, 86 tests passed.
- `yarn agent:check --profile commit` passed before the signed commit; `yarn agent:check --profile pr` local checks passed before PR creation.
- [release-app-bundles.yml](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34574639104) completed successfully for the PR head `b52474b541e65cc110e35f7b26acdd406dd2946d`.
- [iOS QA bundle 1527](https://bundle-test.onekey-asset.com/utility/js-bundle/6.6.0/b52474b5-1527/ios.zip) is available (HTTP 200).

[OK-62230]: https://onekeyhq.atlassian.net/browse/OK-62230?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wabicai (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-980136f0-982d-580c-81ac-30f7cb15f71c)

sidmorizon:
Claude review session: https://claude.ai/code/session_01JeXnJ3t3uRWUqKcYvwkj3G

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
```diff
@@ -873,6 +873,7 @@ export class DeviceStageBurstScope {
       !isHardwareErrorByCode({
         error,
         code: [
+          HardwareErrorCode.BleDeviceNotBonded,
```
sidmorizon, line 876: # 🟠 P1 705 is not exclusively "unpaired" — the Protocol V1 write path reports ordinary BLE disconnects as 705

## Problem

`BleDeviceNotBonded` (705) has two unrelated producers in the pinned transport SDK (`@onekeyfe/hd-transport-react-native@1.2.2-alpha.111`, `yarn.lock:9046`):

1. **A genuine bonding failure** — `src/index.ts:1076` (`!bondState.bonded` → `TypedError(705, 'device is not bonded')`) and `src/BleManager.ts:61/97/113` (bond timeout / `BOND_BONDING → BOND_NONE` failure). Every one of these sits behind `if (Platform.OS === 'android')` (`src/index.ts:1069`) or the Android bond-state listener, and each carries `params: { phase: 'bond', reason }`.
2. **A plain peripheral disconnect during a Protocol V1 characteristic write** — inside `callProtocolV1` (`src/index.ts:1547`), at `src/index.ts:1733-1734`:
   ```js
   if (e.errorCode === BleErrorCode.DeviceDisconnected) {
     throw ERRORS.TypedError(HardwareErrorCode.BleDeviceNotBonded);
   }
   ```
   `BleErrorCode.DeviceDisconnected` is ble-plx 201 — the peripheral powered off, went out of range, or dropped the link. This branch is platform-independent and carries no `params`.

The Protocol V2 write path deliberately does *not* do this: it routes the identical native error through `toBleDisconnectHardwareError` (`src/index.ts:2684-2685`, `src/bleNativeDisconnect.ts:30-39`) to `BleDeviceDisconnected` (720), which *is* in `DEVICE_STAGE_DISCONNECTED_CODES` (`packages/shared/src/hardware/deviceStageErrorCodes.ts:12-18`). So the PR description's "protocol V2 errors reported as 720 retain their existing behavior" is accurate — but that asymmetry is precisely the problem: only the V1 family funnels a real disconnect into 705, and V1 is the Classic/Mini/Touch/Pro path.

Until this line, the `isDeviceStillConnected` probe at `DeviceStageBurst.ts:884-910` was the *only* discriminator the app had between the two meanings: a `false` verdict re-classified those 705s as `'disconnected'`. Line 876 removes that discriminator unconditionally, and the inline comment above it ("These failures identify the cause even when the transport tracker has already cleared the connection") does not hold for 705 on V1.

## Impact

A Protocol V1 device that goes out of range or powers off mid-call now lands on the stage with `errorReason: undefined` and `errorI18n: { key: 'feedback.bluetooth_unpaired' }` (`packages/shared/src/errors/errors/hardwareErrors.ts:318-329`) — "Bluetooth unpaired" / "蓝牙未配对". The user is told their pairing is broken and is implicitly directed to forget and re-pair the device in system settings: a destructive, unnecessary remediation for a device that is simply not nearby. On iOS the effect is total — producer 1 is Android-only, so *every* iOS 705 originates from the disconnect branch at `src/index.ts:1734`.

The two cases are also indistinguishable by message: `TypedError(705)` with no message is backfilled from `HardwareErrorCodeMessage[705]` in the `HardwareError` constructor (`@onekeyfe/hd-shared` `src/HardwareError.ts:58-72`), producing the same `'device is not bonded'` string that producer 1 sets explicitly.

## Recommendation

Disambiguate where the native error is still intact, rather than blanket-excluding the code:

- **Preferred (SDK-side).** At `hd-transport-react-native/src/index.ts:1733-1734`, keep 705 only when the native error actually indicates an authentication/pairing failure, and map the remaining `BleErrorCode.DeviceDisconnected` cases to `HardwareErrorCode.BleDeviceDisconnected`, mirroring what the V2 path already does. Once that lands, line 876 here is correct as written.
- **App-side interim.** Drop `HardwareErrorCode.BleDeviceNotBonded` from the array at line 876 and add a narrow guard for the bond-failure shape instead, e.g. `isHardwareErrorByCode({ error, code: HardwareErrorCode.BleDeviceNotBonded }) && (error as IOneKeyError)?.payload?.params?.phase === 'bond'`. `convertDeviceError` spreads the SDK payload (including `params`) onto `error.payload` (`packages/shared/src/errors/utils/deviceErrorUtils.ts:83-92`), and `payload` survives `toPlainErrorObject` + the bg→main JSON hop (`packages/shared/src/errors/utils/errorUtils.ts:45`), so this discriminator is reachable at this call site.

Note the interim option restores "Device disconnected" for the iOS path. Before landing the blanket exclusion, please confirm which producer the OK-62230 iOS repro actually hits — if it is `src/index.ts:1734`, that is the same line an ordinary out-of-range disconnect hits, and the fix cannot be made correct on the app side alone.

[Claude session ↗](https://claude.ai/code/session_01JeXnJ3t3uRWUqKcYvwkj3G)

---

PR #13341, opened 2026-09-11
fix: phone Earn detail page QA round 2 (OK-61140, OK-62885, OK-62886, OK-62888, OK-62922, OK-62923, OK-62924, OK-62925, OK-62926, OK-62942, OK-62943, OK-62948, OK-62949)

ezailWang:
[OK-61140](<https://onekeyhq.atlassian.net/browse/OK-61140>) [OK-62885](<https://onekeyhq.atlassian.net/browse/OK-62885>) [OK-62886](<https://onekeyhq.atlassian.net/browse/OK-62886>) [OK-62888](<https://onekeyhq.atlassian.net/browse/OK-62888>) [OK-62922](<https://onekeyhq.atlassian.net/browse/OK-62922>) [OK-62923](<https://onekeyhq.atlassian.net/browse/OK-62923>) [OK-62924](<https://onekeyhq.atlassian.net/browse/OK-62924>) [OK-62925](<https://onekeyhq.atlassian.net/browse/OK-62925>) [OK-62926](<https://onekeyhq.atlassian.net/browse/OK-62926>) [OK-62942](<https://onekeyhq.atlassian.net/browse/OK-62942>) [OK-62943](<https://onekeyhq.atlassian.net/browse/OK-62943>) [OK-62948](<https://onekeyhq.atlassian.net/browse/OK-62948>) [OK-62949](<https://onekeyhq.atlassian.net/browse/OK-62949>)

----------

Second QA round on the phone Earn detail page and the positions page. All changes are phone-only unless noted; wide layouts keep their current rendering.

Detail page
- OK-61140 The Team / Investors / Audits dialogs drag closed from the grabber and title only. A new `sheetDragArea: 'header'` Dialog option turns the sheet's own drag off so the list scrolls natively (no more jumping) and puts a pan on the header that moves the body and dismisses past 120pt / 800pt/s.
- OK-62885 Links opened from inside those dialogs (audit "View", member social icons) close the dialog first on native, so the sheet is not still there after the in-app browser.
- OK-62886 The header divider between the vault name and the maturity date is a filled 1pt `$border` line; iOS never painted the hairline right border of the vertical Divider.
- OK-62888 The page refreshes again on `HistoryTxStatusChanged`, i.e. once the claim / stake / withdraw transaction confirms; the refresh at broadcast time still returned pre-transaction balances.
- OK-62923 Long token names in the Token info cell wrap inside their column instead of running under the price.
- OK-62925 The Protocol tab is hidden when the provider sends no intro data (Lista), same rule the Portfolio tab already follows.
- OK-62942 A Lista reward check refetches the detail once the signature is accepted, so the reward amount and its Claim button appear without reopening the page.
- OK-62922 New `copy` action icon for server-driven cells; the contract address cell uses it (server side lands separately).
- OK-62949 The Protocol Tips "View all" dialog uses the header-only drag as well, so scrolling its body no longer moves the sheet (reported on Android).
- OK-62943 Android: the dotted rule under the APY figure is an SVG line; Android Text ignores `textDecorationStyle`. iOS / web keep the text decoration.
- OK-62948 Android: pager pages that have fully left the viewport are hidden, so pixel rounding of the translation cannot leave a sliver of the next page visible.

Positions page
- OK-62926 The provider header rule and the row rules are filled 1pt lines on native (the header's Divider had a size token as its bg).
- OK-62924 A reward's Claim button only spins while a pending claim for that same token exists; a MORPHO claim in flight no longer spins the KAT button.

Verification
- iOS simulator: every item above except OK-62943 / OK-62948 / OK-62949 was checked on device; those three are Android-only in effect (OK-62949's mechanism was checked on iOS) and still need an Android pass.
- `yarn agent:check --profile commit` passes; new unit tests cover the tab visibility rules.

[OK-61140]: https://onekeyhq.atlassian.net/browse/OK-61140?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62885]: https://onekeyhq.atlassian.net/browse/OK-62885?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62886]: https://onekeyhq.atlassian.net/browse/OK-62886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62888]: https://onekeyhq.atlassian.net/browse/OK-62888?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62922]: https://onekeyhq.atlassian.net/browse/OK-62922?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62923]: https://onekeyhq.atlassian.net/browse/OK-62923?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62924]: https://onekeyhq.atlassian.net/browse/OK-62924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62925]: https://onekeyhq.atlassian.net/browse/OK-62925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62926]: https://onekeyhq.atlassian.net/browse/OK-62926?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62942]: https://onekeyhq.atlassian.net/browse/OK-62942?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62943]: https://onekeyhq.atlassian.net/browse/OK-62943?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62948]: https://onekeyhq.atlassian.net/browse/OK-62948?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62949]: https://onekeyhq.atlassian.net/browse/OK-62949?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-301953dc-1c17-5951-832f-1ccaeccda7a0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj

packages/kit/src/views/Earn/pages/EarnProtocolDetails/components/ProtocolIntroSection.tsx
```diff
@@ -580,6 +591,24 @@ function ExpandableDescription({ text }: { text: IEarnProtocolIntroText }) {
...
+// it opens lands on top of the sheet, which is still sitting there when the
+// user comes back (OK-62885), so the sheet goes first. On desktop the link
+// opens a browser tab and the dialog stays where it was. Outside a dialog the
+// close is a no-op.
+function useOpenLinkFromDialog() {
+  const dialog = useDialogInstance();
+  return useCallback(
+    (url: string) => {
+      if (platformEnv.isNative) {
+        void dialog.close();
```
originalix: P2: [Native intro links race dialog dismissal]

When a native user opens a team social or audit link from this dialog, `dialog.close()` is started but not awaited before `openUrlExternal(url)` presents the in-app browser. The browser can therefore be pushed while the sheet is still exiting, which leaves the overlay transition that this helper is meant to prevent.

Make this callback async and await `dialog.close()` on native before calling `openUrlExternal(url)`.

huhuanming (dismissed):

packages/kit/src/views/Earn/pages/EarnProtocolDetails/components/ProtocolIntroSection.tsx
```diff
@@ -580,6 +591,24 @@ function ExpandableDescription({ text }: { text: IEarnProtocolIntroText }) {
...
+      }
+      openUrlExternal(url);
```
sidmorizon, line 608: 🟡 P2: Native links outrun dialog dismissal

Severity: `non-severe`

On native, `useOpenLinkFromDialog` opens the URL without awaiting `dialog.close()`. The destination appears while the sheet still exists underneath it.

packages/components/src/composite/Dialog/index.tsx
```diff
@@ -330,18 +352,76 @@ function DialogFrame({
...
+
+          if (
+            event.translationY > HEADER_DRAG_DISMISS_DISTANCE ||
+            event.velocityY > HEADER_DRAG_DISMISS_VELOCITY
+          ) {
+            runOnJS(dismissFromHeaderDrag)();
+            return;
+          }
+          headerDragY.value = withSpring(0, HEADER_DRAG_SPRING);
+        }),
```
sidmorizon: 🟡 P2: Cancelled drags leave sheets displaced

Severity: `non-severe`

When the pan cancels after moving, `headerDragY` retains its translation because `onEnd` never runs. The sheet stays displaced until another drag completes.

packages/kit/src/views/Staking/components/ProtocolDetails/EarnActionIcon.tsx
```diff
@@ -609,6 +618,10 @@ function BasicEarnActionIcon({
       icon = 'OpenOutline';
       onPress = () => openUrlExternal(actionIcon.data.link);
       break;
+    case 'copy':
+      icon = 'Copy3Outline';
+      onPress = () => copyText(actionIcon.data.text);
+      break;
```
sidmorizon, line 626: 🟡 P2: Disabled copy actions remain active

Severity: `non-severe`

A `copy` action with `disabled: true` still assigns `copyText` as its handler. The rendered button remains enabled and copies the value.

packages/components/src/composite/Dialog/index.tsx
```diff
@@ -330,18 +352,76 @@ function DialogFrame({
...
+        .hitSlop({ top: 8, bottom: 8 })
+        .onUpdate((event) => {
+          'worklet';
+
+          headerDragY.value =
+            event.translationY >= 0
+              ? event.translationY
+              : event.translationY * HEADER_DRAG_UPWARD_RESISTANCE;
+        })
+        .onEnd((event) => {
```
sidmorizon: # ⚠️ P1 Cancelled or upward pan can dismiss the sheet

## Problem

The new header pan's `onEnd` (`packages/components/src/composite/Dialog/index.tsx:385-396`) drops the RNGH `success` flag and never checks the drag direction:

```js
.onEnd((event) => {
  if (
    event.translationY > HEADER_DRAG_DISMISS_DISTANCE ||
    event.velocityY > HEADER_DRAG_DISMISS_VELOCITY
  ) {
    runOnJS(dismissFromHeaderDrag)();
    return;
  }
  headerDragY.value = withSpring(0, HEADER_DRAG_SPRING);
})
```

Two gaps:

1. **Missing `success` guard.** In `react-native-gesture-handler` v2 (`apps/mobile/package.json` pins `~2.32.0`), `Gesture.Pan().onEnd(cb)` is invoked with `(event, success)` on ACTIVE → END **and** on ACTIVE → FAILED/CANCELLED. The handler here declares only `event`, so an interrupted pan still evaluates the dismiss thresholds against the last known `translationY`/`velocityY` and calls `runOnJS(dismissFromHeaderDrag)` → `handleOpenChange(false)` → `onClose()` (lines 367-369).

2. **Missing direction guard on the velocity branch.** `onUpdate` (lines 380-383) damps upward drags by `HEADER_DRAG_UPWARD_RESISTANCE` but keeps `translationY` negative. Pulling the header up and then flicking down while `translationY` is still negative satisfies `event.velocityY > 800` on its own, so a sheet sitting *above* its resting position dismisses.

## Impact

On iOS/Android phones — the only place `isHeaderDragOnly` can be true, since it requires `media.md && !disableDrag` (line 359-360) and both call sites pass `disableDrag: platformEnv.isRuntimeBrowser` — the Protocol-intro and Protocol-tips sheets can close without a completed downward drag: after an interrupted/cancelled pan, or after a rubber-band pull-up released with downward momentum. These are informational sheets, so no data is lost, but it reads as the sheet dismissing itself, which is the class of behaviour OK-61140 set out to fix.

## Recommendation

At `index.tsx:385`, accept the second parameter and require both a successful end and a downward translation before honouring velocity:

```js
.onEnd((event, success) => {
  'worklet';

  const shouldDismiss =
    success &&
    (event.translationY > HEADER_DRAG_DISMISS_DISTANCE ||
      (event.translationY > 0 &&
        event.velocityY > HEADER_DRAG_DISMISS_VELOCITY));
  if (shouldDismiss) {
    runOnJS(dismissFromHeaderDrag)();
    return;
  }
  headerDragY.value = withSpring(0, HEADER_DRAG_SPRING);
})
```

Consider adding `.onFinalize` only if you also need to reset `headerDragY` when the gesture never activates; the `open` effect at lines 362-366 already covers the re-open case.

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

packages/kit/src/views/Earn/components/PortfolioTabContent.tsx
```diff
@@ -196,12 +196,24 @@ const WrappedActionButtonCmp = ({
...
+      ) {
+        return false;
+      }
+      // A vault can pay more than one reward token (Morpho on Katana pays
+      // MORPHO and KAT), and the tag is per vault: a MORPHO claim in flight
+      // put a spinner on the KAT button as well (OK-62924). Match the token
+      // the pending claim receives; a claim recorded without one keeps the
+      // vault-wide match.
+      const pendingRewardSymbol = tx.stakingInfo.receive?.token?.symbol;
+      if (pendingRewardSymbol && rewardSymbol) {
```
sidmorizon, line 217: # 🔍 P2 New reward-symbol guard is inert on the reward rows

## Problem

The new guard at `packages/kit/src/views/Earn/components/PortfolioTabContent.tsx:211-213` only narrows the pending-claim match when the `rewardSymbol` prop is set; otherwise it falls through to `return true`, i.e. the old vault-wide match.

`rewardSymbol` is passed at exactly one of the four `WrappedActionButton` call sites — line 719, the `ProtocolAirdrop` group renderer (which already passed it before this PR). The remaining three do not:

- line 1078 — the phone-only expanded row, `asset.rewardAssets?.map(...)` inside `renderExpandedContent` (gated by `!isDesktopLayout` at line 1003)
- line 1054 — `asset.assetsStatus?.map(...)` in the same expanded row
- lines 562 / 512 — the desktop `ActionField` / status field, same two loops

Every button inside one `asset` shares a `stakeTag` built from `(providerName, symbolForConfig, vaultForConfig)` (lines 186-192), which for a non-airdrop asset derives from `asset.token.info.symbol` and is therefore identical across all of that asset's `rewardAssets`. The pending tx does carry the distinguishing value — `usePortfolioAction.ts:124,168-170` sets `stakingInfo.receive.token` from `convertEarnTokenToIToken(token?.info)`, where `token` is `getRewardButtonToken(reward.button)` (line 263) — so the comparison would work; it is simply never reached on these paths.

## Impact

On the positions page this PR targets, a vault paying more than one reward token (the MORPHO + KAT case named in the comment at lines 205-209) still spins every reward button in the expanded phone row while a single claim is pending. OK-62924 is fixed for airdrop groups only; the `rewardAssets` rows that produce the reported symptom are unchanged.

## Recommendation

Pass each button's own reward symbol at the remaining call sites, mirroring line 719. Since the filter compares against `tx.stakingInfo.receive?.token?.symbol`, the matching source is the button's own token:

```jsx
<WrappedActionButton
  asset={asset}
  reward={reward}
  rewardSymbol={getRewardButtonToken(reward.button)?.info?.symbol}
/>
```

at lines 1078 and 562, and the `actionableStatus` equivalent at lines 1054 and 512. If some of those paths are known to be single-reward, a short comment saying so would keep the vault-wide fallback at line 214 from looking accidental.

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

packages/components/src/composite/Dialog/index.tsx
```diff
@@ -17,9 +17,12 @@ import { FocusScope } from '@tamagui/focus-scope';
 import { setStringAsync } from 'expo-clipboard';
 import { isNil } from 'lodash';
 import { useIntl } from 'react-intl';
+import { Gesture, GestureDetector } from 'react-native-gesture-handler';
```
sidmorizon: # ⚠️ P1 New gesture-handler import breaks Dialog.focus.test.tsx

## Problem

This import is new — `59fbc785:packages/components/src/composite/Dialog/index.tsx` has no `react-native-gesture-handler` reference — and it makes `Dialog/index.tsx` unloadable in the Web jest project. `Unit Tests (4/5)` on this head fails with:

```
FAIL packages/components/src/composite/Dialog/Dialog.focus.test.tsx
  ● Test suite failed to run

    TypeError: Cannot read properties of undefined (reading 'create')

      at Object.create (node_modules/react-native-gesture-handler/src/components/GestureHandlerRootView.web.tsx:20:27)
      at Object.require (node_modules/react-native-gesture-handler/src/components/gestureHandlerRootHOC.tsx:4:1)
      at Object.require (node_modules/react-native-gesture-handler/src/index.ts:6:1)
      at Object.<anonymous> (packages/components/src/composite/Dialog/index.tsx:46:36)
      at Object.<anonymous> (packages/components/src/composite/Dialog/Dialog.focus.test.tsx:114:11)
```

(Test Suites: 1 failed, 310 passed — [job 103226596319](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34586292717/job/103226596319).)

`Dialog.focus.test.tsx` runs under `@jest-environment jsdom` with `@jest-environment-options {"customExportConditions": ["node", "node-addons"]}` (lines 1-4), so `react-native` → `react-native-web` resolves its node condition and `StyleSheet` is undefined. The suite deliberately stubs every RN-touching dependency of this module — `react-native-reanimated` (line 64), `react-native-safe-area-context` (line 70), `expo-clipboard` (line 71), `SheetGrabber` (line 87). The gesture-handler barrel added here is now the one un-stubbed import, and its web root view calls `StyleSheet.create` at module scope. The test file itself is unchanged by this PR, so nothing on the test side could have anticipated it.

## Impact

CI is red on `e5fa9d97` and the PR cannot merge. The failure is not a flake and will not clear on a re-run — it is a module-load error caused by this line.

## Recommendation

Stub the barrel in `Dialog.focus.test.tsx` alongside the existing mocks. Note that a bare `{}` will not work: `headerDragGesture` is built in a `useMemo` that runs on **every** render (`index.tsx:370-398`), not only when `isHeaderDragOnly` is true, so `Gesture.Pan()` must return a chainable builder:

```ts
jest.mock('react-native-gesture-handler', () => {
  const gesture = new Proxy({}, { get: () => () => gesture });
  return {
    Gesture: { Pan: () => gesture },
    GestureDetector: ({ children }: { children?: ReactNode }) => children,
  };
});
```

The sibling `MorphOverlay/index.dismiss.test.tsx:23-28` takes the `requireActual` route instead, but only because it runs plain jsdom with `jest.unmock('react-native')` (lines 1-3, 21) — that is not available to this suite under its node export conditions.

While you are there: this PR also adds `runOnJS` and `withSpring` to the reanimated import (line 22, 24), which the existing reanimated mock at `Dialog.focus.test.tsx:64-68` does not provide. They are only called inside `onEnd`, so nothing fails today, but the mock is now incomplete.

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

packages/kit/src/views/Earn/pages/EarnProtocolDetails/index.tsx
```diff
@@ -1199,6 +1272,30 @@ const EarnProtocolDetailsPage = ({ route }: { route: IRouteProps }) => {
...
+  useEffect(() => {
+    if (!isMobileLayout) {
+      return undefined;
+    }
+    const handleHistoryTxStatusChanged = () => {
+      void refreshData();
+    };
+    appEventBus.on(
+      EAppEventBusNames.HistoryTxStatusChanged,
+      handleHistoryTxStatusChanged,
```
sidmorizon, line 1289: 🟡 P2: Confirmed Earn balances stay stale

Severity: `non-severe`

On phone, `HistoryTxStatusChanged` misses normal transaction confirmations, so `refreshData` does not run. [ServiceHistory](packages/kit-bg/src/services/ServiceHistory.ts:1557) emits `LocalPendingTxConfirmed` for these transitions. Claims, stakes, and withdrawals keep stale balances until another refresh.

packages/components/src/composite/Dialog/HeaderDragZone.native.tsx
```diff
@@ -0,0 +1,79 @@
+import { useMemo } from 'react';
```
sidmorizon, line 1: # ⚠️ P1 New native file is missing from the module-id registry

## Problem

This new file is not in `apps/mobile/bundle-registry/module-id-registry.json`, so the iOS union build aborts after building both graphs:

```
Main graph modules: 14688
Background graph modules: 14643
packages/components/src/composite/Dialog/HeaderDragZone.native.tsx.
```

(`Native startup graph budget (24.x)`, [job 103241061836](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34592591537/job/103241061836).)

I confirmed the gap locally: the registry contains 79 `.native.tsx` entries and no `HeaderDragZone` entry, and the PR does not touch the registry file. The convention is visible on this file's own sibling — `Content.tsx` / `Content.native.tsx` are a platform pair, and only `packages/components/src/composite/Dialog/Content.native.tsx` is registered (registry line 16112); the same holds for `renderToContainer.android.tsx` and `renderToContainer.ios.tsx`. So `HeaderDragZone.native.tsx` needs an entry and the web `HeaderDragZone.tsx` does not.

## Impact

The iOS union build fails, so `Native startup graph budget (24.x)` is red and the PR cannot merge. This is a new failure introduced by `c3a421f3` — it did not exist on `62911a49`, where the gesture still lived inline in `index.tsx`. Worth noting the extraction itself is the right call for the jest failure it targets; only the registry step is missing.

## Recommendation

```
```

Registry IDs are allocation-sensitive, so regenerate with the tooling rather than hand-adding a line — per `CLAUDE.md`, generated files are regenerated with the repo's own tooling, never edited by hand.

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

packages/components/src/composite/Dialog/HeaderDragZone.tsx
```diff
@@ -0,0 +1,20 @@
...
+  dragY: SharedValue<number>;
+  /** called once a completed downward pull passes the dismiss thresholds */
+  onDismiss: () => void;
+  /** keeps the zone tall enough to grab when the dialog renders no title */
+  minHeight?: number;
+  children: React.ReactNode;
+};
+
+// Web: callers disable the header drag in the browser, and the gesture
+// handler has no web build to load here, so the zone is a plain container.
```
sidmorizon, line 16: # 🔍 P2 Web fallback's comment gives a reason that is not true

## Problem

"the gesture handler has no web build to load here" is incorrect. `react-native-gesture-handler` ships a web build, and this repo already loads it in the web bundle from two plain (web-resolving) files:

- `packages/components/src/composite/MorphOverlay/index.tsx:12` — `import { Gesture, GestureDetector } from 'react-native-gesture-handler'`; that directory has no `.native.tsx`, so this is the file web resolves.
- `packages/components/src/layouts/SwipeableCell/index.tsx:11` — `import Swipeable from 'react-native-gesture-handler/Swipeable'`.

The irony is that RNGH's web build is precisely what threw in `Dialog.focus.test.tsx` (`GestureHandlerRootView.web.tsx` calling `StyleSheet.create`), so the commit's own motivation contradicts the stated reason.

## Impact

The first clause is the real invariant and it is the load-bearing one: every caller must pass `disableDrag` in the browser, which both do today (`ProtocolIntroSection.tsx:1650`, `ProtocolTipsSection.tsx:118`, via `platformEnv.isRuntimeBrowser`). Nothing enforces it, and the comment directs attention to the inaccurate half instead. A maintainer reading this concludes a web header drag is technically impossible and will not revisit it; someone adding a third caller gets no signal that omitting `disableDrag` silently yields a sheet whose header does not drag on web.

## Recommendation

State the actual reason on lines 15-17:

```ts
// Web/desktop/ext: every caller passes disableDrag in the browser, so
// isHeaderDragOnly is never true here and this zone is a plain container.
// Keeping the pan in the native file also keeps the gesture-handler barrel
// out of the web bundle — and out of Dialog.focus.test.tsx, which runs jsdom
// with node export conditions.
```

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

packages/kit/src/views/Earn/pages/EarnProtocolDetails/index.tsx
```diff
@@ -1199,6 +1273,48 @@ const EarnProtocolDetailsPage = ({ route }: { route: IRouteProps }) => {
...
+      return undefined;
+    }
+    const handleHistoryTxStatusChanged = () => {
+      void refreshData();
+    };
+    const handleLocalPendingTxConfirmed = (
+      payload: IAppEventBusPayload[EAppEventBusNames.LocalPendingTxConfirmed],
+    ) => {
+      if (payload.networkId === networkId) {
+        void refreshData();
```
sidmorizon, line 1295: 🟡 P2: Confirmation refresh can remain stale

Severity: `non-severe`

When `LocalPendingTxConfirmed` precedes provider indexing, `refreshData` immediately caches unchanged balances. No delayed retry follows, leaving the page stale until another event.

sidmorizon, line 1295: # 🔍 P2 The confirmation refresh is dropped while the page is blurred

## Problem

`refreshData` is `run` from `usePromiseResult` (`useProtocolDetailData.ts:186`, `refreshData: run`), and calling it with no config is focus-gated — it does not merely skip the state update, it never issues the request:

```ts
// usePromiseResult.ts:207-216
const shouldSetState = (config?: IRunnerConfig) => {
  let flag = true;
  if (checkIsMounted && !isMountedRef.current) flag = false;
  if (checkIsFocused && !isFocusedRef.current) flag = false;   // ← here
  if (alwaysSetState || config?.alwaysSetState) flag = true;
  return flag;
};
...
// usePromiseResult.ts:272 — the fetch is dispatched *inside* this guard
if (shouldSetState(config)) {
  didStartRequest = true;
  ...
  await methodWithNonce({ nonce: requestNonce });
}
```

`checkIsFocused` defaults to `true` (`usePromiseResult.ts:174`), and the options block in `useProtocolDetailData.ts:85-102` passes only `watchLoading`, `swrKey` and `swrShouldPersist` — no `checkIsFocused: false`, no `alwaysSetState`, no `revalidateOnFocus`, no `pollingInterval`.

Nothing recovers the skipped run afterwards. The refocus effect (`usePromiseResult.ts:535-541`) re-runs only when `revalidateOnFocus` is set or when `isDepsChangedOnBlur.current` is true, and that flag is set only for deps-triggered runs (`config?.triggerByDeps`, line 252) — never for a manual `run()` from an event handler.

## Impact

The OK-62888 symptom survives whenever the page is not the focused route at the moment the transaction settles: the user claims or withdraws, moves to another tab or lets the app lock while it confirms, `LocalPendingTxConfirmed` fires, this handler calls `refreshData()`, the call silently no-ops, and on return the page still shows pre-settlement numbers until a manual pull-to-refresh (`index.tsx:1574`). The focused case — user sitting on the page when it confirms — does work, which is why this is P2 and not P1.

To be clear about scope: the sibling `HistoryTxStatusChanged` handler two lines up has exactly the same property, so this increment widens an existing gap rather than creating one — but both handlers are this PR's own new code, so the feature ships with the gap either way.

## Recommendation

Preferred, since it also covers the case where the provider has not indexed the settlement yet when the event arrives — add to the `usePromiseResult` options in `useProtocolDetailData.ts:85-102`:

```ts
revalidateOnFocus: true,
```

Or, minimally, make both handlers bypass the gate:

```ts
void refreshData({ alwaysSetState: true });
```

(`IRunnerConfig.alwaysSetState`, `usePromiseResult.ts:19-23`.)

Note this overlaps with the separate stale-balance thread on this effect — `revalidateOnFocus: true` addresses both, so they want one fix, not two.

[Claude session ↗](https://claude.ai/code/session_01CvnkswZpVujcF3FMHuhpjj)

---

originalix (approved):

PR #13344, opened 2026-09-11
chore: improve worktree merge detection

sidmorizon:
## Summary

- Add layered worktree merge detection using commit ancestry, merged PR metadata, stable patch IDs for squash merges, and final blob comparison.
- Report `NEEDS_MANUAL_REVIEW` when later edits or conflict resolution prevent an automatic conclusion.
- Restrict PR lookup to PRs targeting `x` and detect stale directories under both `.worktree/` and `.worktrees/`.
- Keep cleanup candidates limited to clean worktrees with positive merge evidence or no committed branch delta.

## Validation

- `bash -n .skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh`
- `git diff --check`
- `bash .skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh check 2`
- `yarn agent:check --profile commit` (lint and repository checks passed; `tsc-staged` remains blocked by existing repository-wide TypeScript errors unrelated to this Shell/Markdown-only change)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-09d6d21b-d098-574d-87d9-cfde67ddd3d4)

sidmorizon:
Claude review session: https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -527,12 +577,24 @@ check_one_worktree() {
     fi
   done <<< "$candidate_output"
 
-  if [[ $candidate_count -eq 0 ]]; then
+  if git -C "$wt" merge-base --is-ancestor HEAD origin/x; then
+    result="MERGED_TO_ORIGIN_X_BY_ANCESTOR"
+    merge_evidence="HEAD is an ancestor of origin/x"
+  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ]]; then
+    result="MERGED_TO_ORIGIN_X_BY_PR"
+    merge_evidence="${WT_PR_LABELS[$idx]}"
```
sidmorizon, line 585: 🟡 P2: Merged PR hides later commits

Severity: `non-severe`

After a merged branch gains commits, `WT_PR_LABELS` still marks its current HEAD merged. Cleanup offers a worktree containing unmerged commits for removal.

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -428,6 +435,47 @@ resolve_pr_metadata() {
...
+  base="$(git -C "$wt" merge-base origin/x HEAD)"
+  branch_patch_id="$(patch_id_for_range "$wt" "$base..HEAD")"
+  [[ -n "$branch_patch_id" ]] || return 1
+
+  while IFS= read -r commit; do
+    [[ -n "$commit" ]] || continue
+    commit_patch_id="$(patch_id_for_commit "$wt" "$commit")"
+    if [[ "$branch_patch_id" == "$commit_patch_id" ]]; then
+      printf '%s\n' "$commit"
+      return 0
```
sidmorizon, line 472: 🟡 P2: Whitespace changes falsely match

Severity: `non-severe`

When branch and `x` changes differ only in whitespace, `git patch-id --stable` can produce equal IDs. Cleanup then offers distinct code for removal.

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -527,12 +577,24 @@ check_one_worktree() {
     fi
   done <<< "$candidate_output"
 
-  if [[ $candidate_count -eq 0 ]]; then
+  if git -C "$wt" merge-base --is-ancestor HEAD origin/x; then
+    result="MERGED_TO_ORIGIN_X_BY_ANCESTOR"
+    merge_evidence="HEAD is an ancestor of origin/x"
+  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ]]; then
```
sidmorizon, line 583: # 🔴 [P0] Merged-PR label overrides local merge proof, marking unmerged worktrees removable

## Problem

The new `elif` at line 583 accepts a branch-**name**-derived PR label as merge proof, and because it short-circuits it is evaluated *before* every local check:

```bash
580	  if git -C "$wt" merge-base --is-ancestor HEAD origin/x; then
581	    result="MERGED_TO_ORIGIN_X_BY_ANCESTOR"
583	  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ]]; then
584	    result="MERGED_TO_ORIGIN_X_BY_PR"
586	  elif [[ $candidate_count -eq 0 ]]; then
589	  elif [[ $unmatched_count -eq 0 ]]; then
```

`WT_PR_LABELS[$idx]` comes solely from `resolve_pr_metadata` (line 405):

```bash
gh pr list --head "$branch" --base x --state all --limit 1 \
  --json number,state,isDraft,mergedAt,url
```

No SHA is requested or compared anywhere — no `headRefOid`, no `mergeCommit`, no correlation against `git -C "$wt" rev-parse HEAD`. The label only means "some PR from a branch with this name was merged at some point"; line 584 promotes it to "this worktree's committed content is in `origin/x`". The blob-equality verdict computed just above at lines 566–578 is discarded, so `unmatched_count > 0` is silently ignored.

Then `run_cleanup_candidates` accepts the verdict with no additional gate:

```bash
705	      MERGED_TO_ORIGIN_X_BY_PR)
706	        reason="merged-by-pr"
707	        ;;
```

Three real repo states reach this. (1) **Post-merge commits on the same branch** — the common case after a squash merge, since `--is-ancestor` at line 580 can never be true for a squash-merged branch, so line 583 is the arm that fires. (2) **Branch-name reuse** — another author's `fix/typo` merged into `x`, and your local unmerged `fix/typo` inherits their `MERGED #NNN`. (3) **Merged then reverted in `x`** — the label stays `MERGED`.

This also contradicts the skill's own stated contract: `SKILL.md:16` promises `NEEDS_MANUAL_REVIEW` when layers "disagree or cannot prove the result" (the code short-circuits instead of comparing), `SKILL.md:18` calls PR status "best-effort only", and `SKILL.md:136` says "Never conclude purely from 'this commit was merged'".

## Impact

Reproduced end to end (branch squash-merged into `origin/x`, then one further commit adding `f3.txt`, with `gh` stubbed to return the merged PR):

```
=== [2] wt-foo ===
PR: MERGED #123
Branch-side candidate files: 3
Matched files: 2
Unmatched files: 1          <-- f3.txt exists only on the branch
Result: MERGED_TO_ORIGIN_X_BY_PR
```

and `cleanup-candidates A` then emits:

```
REMOVABLE_WORKTREE|2|wt-foo|feat/foo|MERGED_TO_ORIGIN_X_BY_PR|merged-by-pr|...
```

Before this PR the same state produced `NOT_FULLY_MERGED_TO_ORIGIN_X_BY_CODE` and was `continue`d out of the candidate list, so this change newly promotes demonstrably unmerged worktrees to "removable", after which the skill drives `git worktree remove <path>` (`SKILL.md:106`) on a confirmation the user gave on a false premise.

Calibrating the blast radius honestly: `git worktree remove` leaves `refs/heads/<branch>` intact, so committed work is recoverable from the branch ref. The unrecoverable parts are (a) gitignored local files — the `clean` gate uses `git status --short --untracked-files=all`, which does not flag ignored files, so a worktree holding e.g. `.env.local` reports clean, passes the gate, and is destroyed; and (b) the affirmatively wrong `已合并到远端 x` verdict reported to the user, which invites discarding the branch too.

For contrast, the other two new arms are sound and need no change: `MERGED_TO_ORIGIN_X_BY_ANCESTOR` is a strict subset of the old `candidate_count -eq 0` case, and `MERGED_TO_ORIGIN_X_BY_PATCH_ID` correctly declines to fire once an extra commit exists. Line 583 is the only arm that widens deletion unsoundly.

## Recommendation

Stop treating the PR label as merge proof. Either **demote it** — move the `elif` at 583–585 *below* the local checks (`candidate_count -eq 0`, `unmatched_count -eq 0`, patch-id) so it can corroborate but never override — or **correlate by SHA**: add `headRefOid` (and/or `mergeCommit`) to the `--json` list at line 406, carry it in a `WT_PR_HEAD_OIDS` array alongside `WT_PR_LABELS`, and require `headRefOid == $(git -C "$wt" rev-parse HEAD)` before accepting `MERGED_TO_ORIGIN_X_BY_PR`.

The minimum safe gate if you want to keep the arm in place: require `unmatched_count -eq 0` alongside the `MERGED #*` label, and route label-says-merged-but-files-differ to `NEEDS_MANUAL_REVIEW` — exactly what `SKILL.md:16` already promises.

Separately worth pinning down: `--limit 1 --state all` at line 405 returns whichever PR the default sort puts first, so if an older merged PR can outrank a newer open one for a reused branch name, case (2) fires even without post-merge commits. Adding the SHA check above closes that too.

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -506,12 +563,16 @@ check_one_worktree() {
...
+  local merge_evidence=""
+  local squashed_commit=""
+  local head_oid=""
 
   git -C "$wt" rev-parse --verify origin/x^{commit} >/dev/null 2>&1 || \
     die "origin/x is missing for $wt. Fetch origin x before running check."
 
   candidate_output="$(collect_candidate_paths "$wt")"
   dirty_output="$(collect_dirty_files "$wt" || true)"
+  head_oid="$(git -C "$wt" rev-parse HEAD)"
```
sidmorizon: # 🟠 [P1] Unborn-HEAD worktree aborts the whole run, truncating the audit

## Problem

Line 575 is new in this commit and is the first *unguarded* `git` call in `check_one_worktree`:

```bash
575	  head_oid="$(git -C "$wt" rev-parse HEAD)"
```

Every other git invocation in this function tolerates failure — line 570 ends in `|| die`, line 574 uses `|| true`, `merge-base --is-ancestor` at 591 is an `if` condition. This one is a bare command substitution in an assignment, and the script runs under `set -euo pipefail` (line 2).

For a worktree whose HEAD is unborn — `git worktree add --orphan` (git ≥ 2.42), or a freshly created empty branch — `git rev-parse HEAD` exits 128, so the assignment fails and the entire script dies mid-loop.

## Impact

`check <selection>` and `cleanup-candidates` — the primary flow in `SKILL.md` Steps 3–4 — terminate with `rc=128` after emitting a **partial** report. The user gets a truncated audit that silently omits every worktree after the orphan one, with only git's raw `fatal: ambiguous argument 'HEAD'` on stderr and none of the script's own `die` framing to explain it.

Reproduced on a repo with three worktrees (`main`, one normal branch, one `--orphan`):

```
=== CURRENT script: check A ===
rc=128
Result lines: 2          <-- 2 of 3 worktrees reported
stderr: fatal: ambiguous argument 'HEAD': unknown revision or path not in the working tree.
```

`cleanup-candidates A` on the same repo also exits 128, and dies before printing the `=== CLEANUP SUMMARY ===` section at all.

This is a regression introduced here, not pre-existing. The same repo against the version currently on `x` (merge base `197b9c74`... lineage) and against this PR's own previous revision both complete cleanly:

```
=== BASE (pre-PR, currently on x): check A ===   rc=0   Result lines: 3
=== PREV revision of this PR:      check A ===   rc=0   Result lines: 3
```

The truncated `cleanup-candidates` output is the dangerous part: a user who does not notice the non-zero exit can read a short candidate list as "nothing else to clean" when the script simply never reached the remaining worktrees.

## Recommendation

Make the read fail-soft and let the existing arms handle an empty value. Line 597 already guards the other side with `-n "${WT_PR_HEAD_OIDS[$idx]}"`, and an empty `$head_oid` can never equal a non-empty `headRefOid`:

```bash
head_oid="$(git -C "$wt" rev-parse HEAD 2>/dev/null || true)"
```

Also add `-n "$head_oid"` to the conjunction at line 597 so an empty local HEAD is never compared at all. An unborn HEAD then falls through to `NEEDS_MANUAL_REVIEW`, which is the correct verdict for it.

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -527,12 +588,25 @@ check_one_worktree() {
...
+  elif [[ $candidate_count -eq 0 ]]; then
     result="NO_BRANCH_CODE_DELTA_FROM_COMMON_BASE"
+    merge_evidence="no branch-side committed code delta"
+  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* && -n "${WT_PR_HEAD_OIDS[$idx]}" && "${WT_PR_HEAD_OIDS[$idx]}" == "$head_oid" && $unmatched_count -eq 0 ]]; then
+    result="MERGED_TO_ORIGIN_X_BY_PR"
+    merge_evidence="${WT_PR_LABELS[$idx]} head matches current HEAD and all branch-side blobs match origin/x"
   elif [[ $unmatched_count -eq 0 ]]; then
     result="MERGED_TO_ORIGIN_X_BY_CODE"
+    merge_evidence="all branch-side candidate blobs match origin/x"
+  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ]] && \
```
sidmorizon: # 🟡 [P2] Local squash detection now requires `gh`, contradicting SKILL.md

## Problem

The new conjunct at line 603 gates the patch-id arm on the GitHub PR label:

```bash
603	  elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ]] && \
604	    squashed_commit="$(find_squashed_patch_match "$wt" 2>/dev/null)"; then
605	    result="MERGED_TO_ORIGIN_X_BY_PATCH_ID"
```

`find_squashed_patch_match` is a purely local git computation — merge-base, `git diff`, `git patch-id`, `rev-list`. Nothing in it needs the network. This commit newly makes it reachable only when a **best-effort, network-dependent** signal succeeds. When `gh` is missing, unauthenticated, or rate-limited, the label becomes `UNAVAILABLE (…)` and the arm is unreachable. Note `mark_pr_lookup_unavailable` latches `GH_PR_LOOKUP_STATE` process-wide (line 346), so one failed query disables squash detection for every worktree in the run.

`SKILL.md` promises the opposite in two places:

> line 18 — `PR status is best-effort only. Use `gh` when available; if auth or network is unavailable, report that explicitly and **continue the local git-based check**.`

> line 122 — `` `MERGED_TO_ORIGIN_X_BY_PATCH_ID`: the branch's aggregate patch matches a non-merge commit in `origin/x`, covering squash merges. `` — no PR precondition stated.

Line 16 is also now stale: it says "a **stable** patch-id match", but lines 450/459 switched to `--verbatim`.

## Impact

Fail-safe in direction (no wrong deletion), but it silently disables the tool's squash-merge detection for anyone without working `gh` auth — and squash is the dominant merge style on `x`, so this is the common case, not an edge case. Affected worktrees report `NEEDS_MANUAL_REVIEW` with the misleading evidence string `branch-side files differ from origin/x after later changes`, which blames later edits rather than the missing `gh`. `cleanup-candidates` then yields 0 removable and the skill's purpose is defeated.

Reproduced (branch squash-merged into `origin/x`, then `x` edits the same file so blobs differ):

```
=== WITH gh stub (MERGED label present) ===
Merge evidence: matching squash commit 8b88ae985c90
Result: MERGED_TO_ORIGIN_X_BY_PATCH_ID

=== WITHOUT gh (not installed) ===
PR: UNAVAILABLE (gh-missing)
Merge evidence: branch-side files differ from origin/x after later changes
Result: NEEDS_MANUAL_REVIEW
```

This PR's own previous revision produced `MERGED_TO_ORIGIN_X_BY_PATCH_ID` on that same repo with no `gh` present, so the gating is what changed. (To be precise about scope: patch-id detection is new in this PR, so this is a regression within the PR, not against what is on `x` today.)

## Recommendation

The P0 this commit fixes was about a **label-only** verdict with no local evidence. Patch-id is local evidence, so it does not need the label to be safe. Either drop the conjunct at 603:

```bash
elif squashed_commit="$(find_squashed_patch_match "$wt" 2>/dev/null)"; then
```

or, if you want GitHub to be able to veto, relax the gate to "not contradicted by GitHub" — accept `MERGED #*` **or** a label starting with `UNAVAILABLE`/`NONE` — so a missing `gh` degrades to local-only rather than disabling the check.

Either way, update `SKILL.md` lines 16, 18 and 122 to match the code; all three currently describe behaviour the script no longer has.

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -425,7 +438,49 @@ resolve_pr_metadata() {
...
+  PR_CACHE_HEAD_OIDS+=("$head_oid")
+  printf '%s\x1f%s\x1f%s\n' "$label" "$url" "$head_oid"
+}
+
+patch_id_for_range() {
+  local wt="$1"
+  local range="$2"
+
+  git -C "$wt" diff --binary --no-ext-diff "$range" -- |
+    git -C "$wt" patch-id --verbatim |
```
sidmorizon: # 🟡 [P2] `patch-id --verbatim` needs git ≥ 2.39 and fails silently below it

## Problem

This commit switches both patch-id helpers from `--stable` to `--verbatim` (lines 450 and 459):

```bash
449	  git -C "$wt" diff --binary --no-ext-diff "$range" -- |
450	    git -C "$wt" patch-id --verbatim |
451	    awk 'NR == 1 { print $1; exit }'
```

`--verbatim` was introduced in **Git 2.39.0**; it is absent in 2.38. I verified this against the tagged documentation: `Documentation/git-patch-id.txt` at `v2.39.0` documents `--verbatim` ("Calculate the patch-id of the input as it is given, do not strip any whitespace"), while the same file at `v2.38.0` documents only `--stable` and `--unstable`. `--stable` has existed since 2.13.

On older git the option is rejected, `branch_patch_id` comes back empty, `find_squashed_patch_match` returns 1, and the `2>/dev/null` at line 604 swallows the error completely. This matters for macOS, which this repo targets — Ventura's Apple system git is 2.37.x.

## Impact

Silent, unreported loss of squash-merge detection, with a misleading diagnosis. Simulated with a shim that rejects `--verbatim` and otherwise execs real git, against a squash-merged repo with the PR label present:

```
=== simulated git < 2.39, gh stub present ===
PR: MERGED #1000
Merge evidence: branch-side files differ from origin/x after later changes
Result: NEEDS_MANUAL_REVIEW
rc=0

=== control: real git 2.43, same gh stub ===
Merge evidence: matching squash commit 8b88ae985c90
Result: MERGED_TO_ORIGIN_X_BY_PATCH_ID
```

Exit status stays 0, so nothing signals that a capability was lost — the user is told their files differ due to later edits when in fact the comparison never ran. Combined with the `gh` gating at line 603, `MERGED_TO_ORIGIN_X_BY_PATCH_ID` now has two independent silent failure modes.

To be clear, the hardening intent is correct and worth keeping: `--stable` ignores whitespace, so two *different* whitespace-only changes hash identically, which `--verbatim` correctly distinguishes. This is about detecting the unsupported option, not reverting it.

## Recommendation

Feature-detect once and fall back instead of failing silently:

```bash
PATCH_ID_MODE="--stable"
if git patch-id --verbatim </dev/null >/dev/null 2>&1; then
  PATCH_ID_MODE="--verbatim"
fi
```

Use `$PATCH_ID_MODE` at both 450 and 459. When the fallback is taken, surface it — e.g. append `patch-id mode: --stable (git < 2.39)` to `merge_evidence` — so the weaker whitespace-insensitive comparison is visible rather than silently in effect.

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -506,33 +567,59 @@ check_one_worktree() {
...
+    elif [[ $candidate_count -eq 0 ]]; then
+      result="NO_BRANCH_CODE_DELTA_FROM_COMMON_BASE"
+      merge_evidence="no branch-side committed code delta"
+    elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* && -n "${WT_PR_HEAD_OIDS[$idx]}" && "${WT_PR_HEAD_OIDS[$idx]}" == "$head_oid" && $unmatched_count -eq 0 ]]; then
+      result="MERGED_TO_ORIGIN_X_BY_PR"
+      merge_evidence="${WT_PR_LABELS[$idx]} head matches current HEAD and all branch-side blobs match origin/x"
+    elif [[ $unmatched_count -eq 0 ]]; then
+      result="MERGED_TO_ORIGIN_X_BY_CODE"
+      merge_evidence="all branch-side candidate blobs match origin/x"
+    elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ||
```
sidmorizon: # 🟠 [P1] `--stable` fallback plus the widened guard marks unmerged work MERGED and removable

## Problem

Two changes that are individually reasonable combine into an unsound one.

The guard at 613–615 now reaches the patch-id arm with no GitHub confirmation at all:

```bash
613	    elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ||
614	      "${WT_PR_LABELS[$idx]}" == NONE ||
615	      "${WT_PR_LABELS[$idx]}" == UNAVAILABLE\ \(* ]] && \
616	      squashed_commit="$(find_squashed_patch_match "$wt" 2>/dev/null)"; then
```

And line 59 substitutes a whitespace-insensitive algorithm when `--verbatim` is unavailable:

```bash
59	PATCH_ID_MODE="--stable"
60	if git patch-id --verbatim </dev/null >/dev/null 2>&1; then
61	  PATCH_ID_MODE="--verbatim"
62	fi
```

`--stable` strips whitespace before hashing — exactly the property `--verbatim` exists to remove. So two *different* whitespace-only changes to the same lines produce the same patch ID.

Before this increment, neither half could bite. The arm required a `MERGED #N` label, and on git < 2.39 `patch-id --verbatim` failed, `branch_patch_id` came back empty, the `[[ -n … ]] || return 1` guard fired, and the result fell through to `NEEDS_MANUAL_REVIEW` — **fail-safe**. Now the same conditions **fail open**.

## Impact

Reproduced end to end against this head. Base `g.txt` contains `foo(); bar();`; the branch reindents them to 4 spaces; `origin/x` independently reindents the *same* lines to tabs. Nothing was merged.

```
branch --stable : 722e8b76a7bade731f7b2a23394b099c5655092f
x      --stable : 722e8b76a7bade731f7b2a23394b099c5655092f   >> STABLE COLLIDES
>> verbatim correctly differs
```

With `gh` absent (label `UNAVAILABLE (gh-missing)`) and `unmatched_count = 1`:

```
=== real git (--verbatim) ===
Merge evidence: branch-side files differ from origin/x after later changes
Result: NEEDS_MANUAL_REVIEW            <- correct

=== git < 2.39 (--stable fallback) ===
Merge evidence: matching squash commit 6f6d0a17a1a1 (patch-id mode: --stable)
Result: MERGED_TO_ORIGIN_X_BY_PATCH_ID <- wrong
```

and `cleanup-candidates` then offers it for deletion:

```
Removable worktrees: 1
REMOVABLE_WORKTREE|2|ws|feat/ws|MERGED_TO_ORIGIN_X_BY_PATCH_ID|merged-by-patch-id|…
```

while the branch demonstrably differs from `x`:

```
-	foo();
-	bar();
+    foo();
+    bar();
```

Rating this P1 rather than P0 deliberately: `git worktree remove` keeps the `feat/ws` ref, the dirty-tree gate still blocks uncommitted edits, and deletion needs explicit user confirmation. The defect is a false *merged* verdict driving a destructive recommendation, not unconditional data loss.

To be clear about what is **not** wrong here: patch-id equality with `unmatched_count > 0` is sound under `--verbatim`. A non-zero unmatched count just means `x` moved on after the squash landed, and equality of `base..HEAD` against a single non-merge commit is real evidence the net change was applied. Only the whitespace-stripping fallback breaks the inference — so the widened guard is fine on its own.

## Recommendation

Don't let `--stable` and the no-GitHub-confirmation paths combine. Either:

- keep the widened guard but require `--verbatim` for the `NONE` / `UNAVAILABLE (*` cases, restricting the arm back to `MERGED #*` when running on `--stable`; or
- treat an unavailable `--verbatim` as "patch-id evidence unusable" so the arm yields `NEEDS_MANUAL_REVIEW` rather than a weaker comparison.

If the fallback is kept for `MERGED #*`, make the evidence string say why it is weaker — e.g. `patch-id mode: --stable (whitespace-insensitive)` — so the degraded mode is visible at the point of decision.

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

.skillshare/skills/1k-worktree-clean/scripts/1k-worktree-clean.sh
```diff
@@ -506,33 +567,59 @@ check_one_worktree() {
...
+      "${WT_PR_LABELS[$idx]}" == NONE ||
+      "${WT_PR_LABELS[$idx]}" == UNAVAILABLE\ \(* ]] && \
```
sidmorizon: # 🟡 [P2] Detached worktrees are excluded from the patch-id arm despite having no GitHub evidence

## Problem

The widened allow-list covers `MERGED #*`, `NONE` and `UNAVAILABLE (*`, but not `DETACHED`:

```bash
613	    elif [[ "${WT_PR_LABELS[$idx]}" == MERGED\ #* ||
614	      "${WT_PR_LABELS[$idx]}" == NONE ||
615	      "${WT_PR_LABELS[$idx]}" == UNAVAILABLE\ \(* ]] && \
```

`resolve_pr_metadata` short-circuits for a detached HEAD and returns `DETACHED` **without ever calling `gh`**. That makes it the strongest possible "GitHub has not contradicted this" case — stronger than `UNAVAILABLE (*)`, which at least implies a lookup was attempted. Excluding it means this commit's stated goal ("keep local merge checks available") doesn't reach detached worktrees.

Excluding `OPEN #*`, `OPEN DRAFT #*` and `CLOSED #*` is correct — there GitHub actively contradicts the patch-id evidence. `DETACHED` carries no such signal.

## Impact

Two worktrees at a byte-identical commit get opposite verdicts, decided purely by the label. Verified on a squash-merged branch with `gh` absent, both worktrees at `a1abf269b7f5db0595c9633239ec0faa4bd5c04d`:

```
=== branch worktree (feat/bar) ===
PR: UNAVAILABLE (gh-missing)
Merge evidence: matching squash commit 8b88ae985c90 (patch-id mode: --verbatim)
Result: MERGED_TO_ORIGIN_X_BY_PATCH_ID

=== detached worktree, SAME commit ===
PR: DETACHED
Merge evidence: branch-side files differ from origin/x after later changes
Result: NEEDS_MANUAL_REVIEW
```

Fail-safe in direction, so no risk of wrong deletion — but detached worktrees are a normal way to park a build or bisect, and they are exactly the kind of stale checkout this skill exists to clean up. They now always require manual review, with an evidence string that misattributes the cause to later edits rather than to the missing label.

## Recommendation

Add `|| "${WT_PR_LABELS[$idx]}" == DETACHED` to the guard. Better, invert it to a deny-list — reach the arm unless the label matches `OPEN *`, `CLOSED *`, or the `${state} #N` fallback — so the intent ("GitHub has not contradicted this") is expressed directly and any future label defaults to the safe-but-useful side rather than silently dropping out.

(`N/A`, for a worktree sitting on `x` itself, is also excluded; that one is harmless, since such a worktree resolves via the ancestor or no-delta arm first.)

[Claude session ↗](https://claude.ai/code/session_013eSGm6YUzipKfupBckxEop)

---

originalix:

Review of commit `fb31f2b8ff98` finished, but some comments could not be published automatically.

originalix (approved):

PR #13345, opened 2026-09-11
fix: hide iOS 26 modal top scroll edge blur OK-61458 OK-62750

originalix:
[OK-61458](<https://onekeyhq.atlassian.net/browse/OK-61458>) [OK-62750](<https://onekeyhq.atlassian.net/browse/OK-62750>)

----------

## Summary

- Default `scrollEdgeEffects: { top: 'hidden' }` for iOS 26+ `modal` and `fullScreen` pages in `makeModalStackNavigatorOptions`.
- Restore the edge effect natively before a configured scroll view can be recycled, so a later unrelated list cannot inherit the hidden top effect (addresses the two P2 review findings).
- Fixes [OK-61458](https://onekeyhq.atlassian.net/browse/OK-61458) (Perps Search asset / History) and [OK-62750](https://onekeyhq.atlassian.net/browse/OK-62750) (Aave supply/borrow token select) without per-page config.

## Intent & Context

iOS 26 draws its automatic top scroll-edge effect on a modal page's main list even though modal headers are opaque, so the first rows (and in-list table headers) get a gradient blur. Reproduced on OK-62750: kill the app, then open the Aave token selector for a market/action combination for the first time in the session (cold cache, so the list mounts after the loading skeleton).

## Root Cause

- The list's `UIScrollView` mounts after a loading state. react-native-screens only reconfigures a late-mounted list when the screen sets a custom edge effect.
- #12948 added that native path (largest-visible scroll view finder + re-apply on layout/mount). It has shipped in native builds since 2026-08-22, but only WalletAddress and Unifold opted in.
- Causal check from the #12948 investigation (iOS 26 simulator): flipping the list's `topEdgeEffect.hidden` from 0 to 1 at runtime removed the blur.

### JS default

- The default lives in the only modal screen-options factory. Route options and `Page.Header` still override it; they replace the whole object, so an override must restate `top` and never pass `undefined` or a partial object.
- `fullScreen` is included: `pushFullModal` mounts the same `modalRouter` under the same opaque header (Prime, AppUpdate, DApp full-screen requests, Perps search on iPad split view).
- Excluded: `onboarding` (its Liquid Glass header relies on the effect), `webView`, `fullScreenPush`, and tab stacks (they never use this factory).
- Top edge only; bottom/left/right stay automatic.
- Supersedes the per-page change on `codex/fix-perps-scroll-edge` (f8b5af479b0). The existing WalletAddress/Unifold settings are identical to the default and are kept until this is verified.

### Native cleanup (`patches/react-native-screens+4.26.0.patch`)

Two hunks in `ios/RNSScreen.mm`, both keeping the existing `isDescendantOfView:self` ownership guard:

1. `willBeUnmountedInUpcomingTransaction` now calls `resetLastContentScrollViewEdgeEffects`. `RNSScreenStackView` invokes this from `mountingTransactionWillMount`, before the first mutation of the unmounting transaction, so the scroll view is still owned by the screen and the reset applies. The pre-existing reset in `didMoveToWindow` runs too late on a JS-driven close (`navigation.popStack()`): Fabric removes and recycles the list first, the guard fails, and the scroll view reaches the recycle pool with `topEdgeEffect.hidden = YES`. `RCTScrollViewComponentView.prepareForRecycle` does not restore edge effects.
2. `resetLastContentScrollViewEdgeEffects` also resets when the remembered scroll view is off-window. This covers a list unmounted while its screen stays mounted (a list re-created with a new `key`, e.g. the Perps token selector switching FAVS/Perps/Spot). An off-window view belongs to no visible screen, and a screen that owns it re-applies its own configuration on `didMoveToWindow`, so this cannot clobber another screen's state.

Rejected alternatives: patching `RCTScrollViewComponentView.prepareForRecycle` (new fork of the most upgrade-expensive package, and its patch set is sequential), and widening the guard alone (its only trigger is the post-dismissal `didMoveToWindow`, by which time the pool may already have handed the view over).

## Changes Detail

- `packages/components/src/layouts/Navigation/GlobalScreenOptions.native.ts`: module-level `{ top: 'hidden' }` applied on iOS 26+ when `pageType` is `modal` or `fullScreen`.
- `patches/react-native-screens+4.26.0.patch`: the two `ios/RNSScreen.mm` hunks above. Regenerated from the pristine 4.26.0 package; verified to apply cleanly to pristine and to reproduce the installed sources byte-for-byte.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: iOS 26+ only. Android, web, desktop, extension and iOS < 26 are unchanged (the native applicator is gated by `RNS_IPHONE_OS_VERSION_AVAILABLE(26_0)` + `@available`, and the JS default by `isNativeIOS26Plus`).
- **Runtime Scope**: main UI runtime only.
- **Release channel**: this PR now contains a native patch, so it must ship with an App Shell release. The JS default must not be delivered bundle-only to an older shell: that would ship the hidden default without the recycle cleanup, which is worse than today.
- **Known limits**:
  - Only the largest visible scroll view per screen is configured. Lists nested inside a larger scroll view (`Page scrollEnabled`, PagerView, collapsible Tabs) are unverified.
  - A scroll view reached through an `RNSContentScrollViewProviding` provider can be a non-descendant; it is configured on apply but skipped by every reset path. Pre-existing, unchanged here.
  - Screens torn down without a mounting transaction (surface stop/restart) still get no cleanup.
  - Every iOS 26 modal becomes a custom edge-effect screen, so the native finder re-runs on layout and Fabric mount transactions. Not profiled yet.

## Test plan

- [x] `yarn agent:check --profile commit` (JS change and native patch commits)
- [x] Patch regenerated per `1k-patch-package-workflow`: applies to pristine with `git apply --check -p1`, reproduces installed sources, contains no build artifacts, survives a clean `yarn install`
- [ ] iOS 26 native build: kill the app, then first-open Aave Core/Arbitrum × supply/borrow token select — column header and first row stay sharp
- [ ] Perps Search asset (FAVS/Perps/Spot, type/clear search), History (Trades/TWAP/Funding/Account), deposit token select
- [ ] Home copy-address modal; one Prime list opened via `pushFullModal`
- [ ] Recycle check (identity-based, needs the native build): in the Perps token modal note the list's inner `UIScrollView` address in the View Debugger, pick a token (`popStack()`), open a Home/Market list, confirm the same address is reused and read `topEdgeEffect.isHidden` — expect `NO`, and the top fade visible. A run where the address differs proves nothing and must be repeated.
- [ ] Liquid Glass preserved: Home tab list, one Onboarding v2 page
- [ ] iOS < 26 and Android smoke

[OK-61458]: https://onekeyhq.atlassian.net/browse/OK-61458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62750]: https://onekeyhq.atlassian.net/browse/OK-62750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-830867ed-9e81-5cf5-8f15-11c3d62db0a0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01SraMnDmNgGSVEAL1QKYAu4

packages/components/src/layouts/Navigation/GlobalScreenOptions.native.ts
```diff
@@ -109,6 +117,15 @@ export function makeModalStackNavigatorOptions({
...
       contentStyle: { backgroundColor: bgColor ?? 'transparent' },
     };
+    // fullScreen (pushFullModal) mounts the same modal routes under the same
+    // opaque header, so it must match modal. Onboarding keeps the effect for
+    // its Liquid Glass header; webView and fullScreenPush are out of scope.
+    if (
+      platformEnv.isNativeIOS26Plus &&
+      (pageType === EPageType.modal || pageType === EPageType.fullScreen)
+    ) {
+      options.scrollEdgeEffects = MODAL_SCROLL_EDGE_EFFECTS;
```
sidmorizon, line 127: 🟡 P2: Recycled lists lose top fade

Severity: `non-severe`

When JS closes an affected modal, `scrollEdgeEffects` can remain hidden on its recycled scroll view. The native reset skips views moved outside the closing screen. A later non-modal list can lose its top glass fade.

originalix: P2: 扩大 hidden 默认配置时，未补齐回收列表的原生清理

当前 head 仍有这个清理缺口。选择 Perps token 后会执行 `navigation.popStack()`；Fabric 卸载时先移除/回收子列表，之后原 screen 的 `resetLastContentScrollViewEdgeEffects` 才执行。此时 `isDescendantOfView:self` 为 false，reset 被跳过，但引用仍被清空。`RCTScrollViewComponentView.prepareForRecycle` 也没有恢复 `topEdgeEffect.hidden`，后续 Home/Market 等使用默认 automatic 的页面复用该列表时，可能丢失顶部 Liquid Glass 渐变。

这是已有原生补丁的漏洞，但这里把 hidden 默认值扩大到全部 modal/fullScreen，会扩大受影响范围。建议配套修复原生卸载/回收清理，在视图进入复用池前恢复 edge effect，并验证“选 token 关闭弹窗 → 打开新的非 modal 列表”仍保留渐变。

不要仅删除 `isDescendantOfView` 判断后无条件 reset：旧 screen 的延迟清理可能覆盖新页面已经设置的状态，需要保证清理时机或视图所有权。该修复需要原生包，单独更新 JS bundle 无法补齐。

zhaono1 (approved):

Minnzen (approved):

huhuanming (approved):

PR #13347, opened 2026-09-11
fix: align network selector feedback and index

huhuanming:
## What
- trigger selection haptics when opening the unified network selector, matching the account selector
- request window-centered section indexes for both single-network and all-network lists on iOS and Desktop
- add focused regression coverage for the haptic trigger and Desktop capability

## Why
The network selector did not provide the same selection feedback as the account selector, and the section index could be centered relative to list content instead of the screen/window.

## Impact
The change is scoped to the unified network selector. Desktop window centering requires the upcoming `@onekeyfe/react-native-native-list` package update from app-modules; that dependency bump is intentionally left for follow-up.

## Checks
- focused Jest: 2 suites, 9 tests passed
- runtime behavior previously checked on iOS and Desktop with the local NativeList implementation

### DIFF
--- a/packages/kit/src/components/AccountSelector/hooks/useUnifiedNetworkSelectorTrigger.test.tsx
+++ b/packages/kit/src/components/AccountSelector/hooks/useUnifiedNetworkSelectorTrigger.test.tsx
@@ -1,0 +1,64 @@
+/**
+ * @jest-environment jsdom
+ */
+import { act, renderHook } from '@testing-library/react';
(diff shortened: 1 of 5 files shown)


huhuanming:
The complete patch was ported unchanged to PR #13350 as commit bb929127ae. PR #13350 already carries the required unified native module version 3.0.123, so the deferred Desktop section-index capability and app-side opt-in can land together there.

PR #13272, opened 2026-09-09
feat: migrate mobile market to native pager with resilient lists

huhuanming:
## Summary

Migrate the iOS and Android `MarketHomeV2` home screen to `CollapsiblePagerView` and native-backed Market lists while preserving the production data, filters, navigation, pagination, watchlist actions, and banner behavior.

The migration covers Favorites, Trending, Stocks, Robinhood, Top Coins, and Perps. Pager selection, sticky headers, native view attachment, refresh state, list snapshots, and recovered data remain consistent across page changes, navigation returns, lifecycle transitions, pagination, and network recovery. Shared pager/list contracts and fallback behavior stay aligned across iOS, Android, and Web; the existing Web/Desktop Market layout remains unchanged.

The UI runs in the native main runtime, while requests run in the independently initialized background runtime. Platform native views own attachment and cell recycling; main/background readiness is not assumed to be simultaneous.

## Changes

- coordinate pager selection, retained page position, sticky-header bounds, horizontal and edge gestures, and short/offscreen page layout on iOS and Android
- match the existing Market rows with native typography, badges, spacing, row heights, images, network decoration, long-press anchors, loading/retry/no-data rows, and recycled-cell resets
- preserve existing rows during refresh and failures, reject stale category/cursor results, restore successful page-owned data after unmounts, and recover loading/error state after reconnecting
- keep snapshots, quote-only patches, row identities, watchlist mutations, banner readiness, and pager tab guards synchronized across reloads and returns
- deduplicate overlapping ranking pages by the same business identity used for NativeList row keys while preserving requested page progress
- wait for the first successful banner response before locking the native pager header height, so an initial failure can recover without retaining the no-banner height
- restore checked and disabled favorite actions for Stocks and Top Coins without forwarding those presses to detail navigation
- on Android background runtime only, invalidate the matching rejected token-list request on an explicit retry so polling cannot reuse a failed promise
- keep the shared Market identity helper in an existing registered native bundle module so three-bundle Union Build remains deterministic
- consume formally published app-modules packages and remove the temporary NativeList and Pager patch-package files

## Published package verification

- upstream completion: [OneKeyHQ/app-modules PR #97](https://github.com/OneKeyHQ/app-modules/pull/97), release commit `9c64500201eb5cfd2c9fbf73122db535c76e98f0`
- CI publication: [package-publish run 34532390534](https://github.com/OneKeyHQ/app-modules/actions/runs/34532390534) completed successfully
[...]

PR #13350, opened 2026-09-11
fix: stabilize mobile Market pager and network selector interactions

huhuanming:
## Intent

The native Market migration is already present on `x` through #13272. This PR is the follow-up stabilization layer: align Android Market pager and gesture behavior with the validated iOS interaction model, fix pagination stability and recovery, and preserve the existing iOS host-retention behavior.

All platform differences remain opt-in and scoped to the mobile Market page. Default behavior for other pager and NativeList callers is unchanged.

## What changed

- enable the native primary tab bar, Stocks subheader, and smooth-header gesture bridge for Android Market
- continuously interpolate primary-tab text, underline, and horizontal title-row centering from native pager progress
- keep Stocks subcategory horizontal gestures owned by the subheader while vertical gestures collapse and expand the shared Header
- cancel banner, filter, and time-control presses after vertical or clear horizontal drags while preserving real taps
- preserve the first visible Market row key and decorated top offset across pagination-only NativeList relayouts, preventing sticky-header bounce near the list bottom
- disable pull-to-refresh only for Android Market NativeLists so Android does not expose loading UI absent from the accepted iOS behavior
- add token load-more failure state, stop repeated `onEndReached` retries, expose the native retry footer, and resume only from an explicit retry
- keep the Market and DeFi `react-freeze` bypass iOS-only; Android logs did not show the same native host teardown
- consume the unified `@onekeyfe` release `3.0.123` across app packages and Android/iOS locks, including `react-native-native-list`, pager behavior, `react-native-image`, and the other version-locked native modules

Native implementation and release: [OneKeyHQ/app-modules#99](https://github.com/OneKeyHQ/app-modules/pull/99).

## Root causes

1. Android centered the title row from `round(progress)`, so the underline moved continuously while the titles jumped at discrete page boundaries.
2. A Stocks horizontal child could consume `ACTION_UP` without completing the parent nested gesture lifecycle, leaving responder ownership stuck for later taps.
3. Pagination-only relayouts did not restore the decorated top offset of the first visible row, allowing the sticky Header to oscillate when more rows were appended.
4. Token pagination failures left `canLoadMore` enabled without a load-more error or retry path, allowing repeated end-reached attempts.

## Android acceptance

- continuous main-tab title and underline progress: PASS
- rapid main-tab, Stocks subcategory, and Market to DeFi switching without blank lists or native host rebuild: PASS
- Stocks row horizontal ownership, accurate selection, and selected-item recentering: PASS
- vertical Header collapse and expansion from non-list regions: PASS
- drag press cancellation with light-tap recovery: PASS
- pagination at the list bottom with a stable sticky Header: PASS
- Android Market pull-to-refresh loading UI disabled: PASS
- token pagination failure pauses automatic retries and provides explicit retry: implemented in review follow-up; GitHub checks are authoritative for the final patch

The final dedicated 60 Hz AVD pagination window recorded 927 rendered frames with 24 missed-deadline or janky frames (2.59%), P50 16 ms, P90 20 ms, P95 21 ms, P99 26 ms, and zero missed-vsync frames. The simulator cannot prove physical-device 120 fps; 120 Hz Android hardware validation remains the only unconfirmed performance item.

## iOS regression boundary

- the previously accepted iOS native tab, subheader, gesture bridge, and Market to DeFi retention behavior remains in place
- relevant `3.0.123` pager interaction changes are Android-specific
- iOS packages and Pod lock resolve to the same unified version
- no dedicated post-`3.0.123` iOS runtime pass was fabricated because this worktree has no registered iOS simulator

## Validation and evidence

Validation completed before the final review follow-up:

- focused Market Jest: 5 suites / 63 tests passed
- JDK 17 NativeList Kotlin compilation against installed `3.0.123`: passed
- full Android DevSession rebuild against published `3.0.123`: passed, 1982 tasks, `userNoticeRequired=false`
- installed published native sources matched the validated app-modules worktree

After CI exposed the nondeterministic pagination retry mock, the focused hook suite passed 16/16, the companion NativeList suite passed 15/15, and `yarn agent:check --profile commit` passed. No native build or full regression suite was run. Required PR checks remain enabled and auto-merge does not bypass them.

Primary evidence roots:

- `/Volumes/T7Shield/Project/app-monorepo-market-pager-ios-transition/.tmp/android/bottom-jitter-20260912/final-3.0.123`
- `/Volumes/T7Shield/Project/app-monorepo-market-pager-ios-transition/.tmp/android/market-centering-fix-20260912`
- `/Volumes/T7Shield/Project/app-monorepo-market-pager-ios-transition/.tmp/android/evidence/final-3.0.121-published`

Key artifacts include `stocks-upward-pagination.mp4`, `stocks-upward-pagination-tail-contact-sheet.png`, `native-final-session.log`, `gfxinfo-framestats.txt`, `display-refresh-rate.txt`, `interaction-smoke.mp4`, `main-tab-progress-contact-sheet.png`, `native-pager-focused.log`, and `subcategory-healthcare-selected.png`.

## Network selector follow-up

Ported unchanged from #13347 in `bb929127ae`:

- trigger selection haptics when opening the unified network selector, matching the account selector
- request window-centered section indexes for both single-network and all-network lists on iOS and Desktop
- retain focused regression coverage for the haptic trigger and Desktop capability

The original PR deferred the Desktop behavior until the next NativeList release. This PR already consumes unified native version `3.0.123`, so the required capability and app-side opt-in now land together. PR #13347 had green hosted checks before the exact patch was ported; no additional local regression run was started after the transfer.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9f6bc7e6-d2fb-57b8-8d63-9ee523da77bd)

sidmorizon:
Claude review session: https://claude.ai/code/session_01KtprGYr57mvjxBiYLX2uen

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -194,16 +195,28 @@ function transformMarketTokenListResponse({
...
 }
 
+function dedupeMarketTokens(items: IMarketToken[]) {
+  const seen = new Set<string>();
+  return items.filter((item) => {
+    const key = marketTokenKey(item);
+    if (seen.has(key)) return false;
+    seen.add(key);
+    return true;
+  });
```
sidmorizon, line 217: 🟡 P2: Stock variants disappear from rankings

Severity: `non-severe`

When ranking pages contain multiple tokenized variants of one stock, `dedupeMarketTokens` keeps only the first. `marketTokenKey` keys every variant by `stockId`, although each variant has a distinct network and contract. Users cannot open the omitted variants.

huhuanming: 已核对当前身份契约，这里不修改。marketTokenKey 会优先使用 assetId；没有 assetId 时，stockId 明确表示跨链共享的同一股票身份，现有 tokenListHelpers 测试也固定了同一股票的链上变体共享 stockId，详情导航同样按 stockId 打开。保留同一 stockId 的多行只会重复同一个详情目标，并不能提供独立变体导航，因此按 false positive 关闭。

packages/kit/src/views/Market/MarketHomeV2/components/MarketNativeList/MobileMarketNativeLists.tsx
```diff
@@ -0,0 +1,1345 @@
...
+            ? intl.formatMessage({ id: ETranslations.global_an_error_occurred })
+            : undefined
+        }
+        loadingMore={result.isLoadingMore}
+        canLoadMore={result.canLoadMore}
+        contentPaddingBottom={listContainerProps.paddingBottom}
+        emptyContentHeight={listContainerProps.emptyContentHeight}
+        onRowAction={handleRowAction}
+        onActionAnchorInvalidated={badgeInfo.onActionAnchorInvalidated}
+        onEndReached={() => {
```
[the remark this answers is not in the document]
huhuanming: 已在 17f347043b 修复：token hook 记录并仅对当前请求设置 load-more error；NativeList 在失败后阻止 onEndReached 自动重试、展示 load-more error footer，并通过 load-more-retry 显式恢复。补充了 hook 和 native list 的聚焦回归用例。按用户要求本轮未本地运行测试，交由 PR checks 验证。

packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
```diff
@@ -368,418 +444,238 @@ function MobileLayoutComponent({
...
+    },
+    [stockCategories],
+  );
 
+  const handlePageSelected = useCallback(
+    (event: CollapsiblePagerViewOnPageSelectedEvent) => {
       if (
-        tabName === lastAcceptedTabChangeNameRef.current &&
-        latestTabState.activeTabName === tabName
+        !shouldHandleMarketPagerPageSelected(isPagerUserDraggingRef.current)
```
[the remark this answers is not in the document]
huhuanming: 已核对当前 3.0.123 原生实现，这条不适用。iOS CollapsiblePager 在 direct pager 和 controller 两条完成路径里都先 emitPageSelected，再发送 idle；Android 也由 ViewPager2 的 onPageSelected 回调发选择事件。当前 guard 是为忽略 setPage 的程序化回调并避免旧页/unfreeze 竞态。增加 post-idle 接受窗口会重新接收这类程序化回调，因此保持现状并按 false positive 关闭。

packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
```diff
@@ -503,6 +602,26 @@ function MobileLayoutComponent({
     [handleTabChange, setPagerIndex, tabNames],
   );
 
+  const handleNativeTabPress = useCallback(
+    (event: CollapsiblePagerViewOnNativeTabPressEvent) => {
+      const { key, position } = event.nativeEvent;
+      const tabName = tabNames[position] === key ? key : tabNames[position];
```
sidmorizon: # 🟡 P2 Native tab press resolves by position and ignores the pressed `key`

## Problem

`handleNativeTabPress` (`packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx:605-612`) receives `{ key, position }` from the native tab bar but the selector on line 608 is equivalent to `tabNames[position]` in both branches:

```ts
const tabName = tabNames[position] === key ? key : tabNames[position];
```

When `tabNames[position] === key` it returns `key` (same string); otherwise it returns `tabNames[position]`. The `key` never influences the result, so the mismatch branch, the only case where the two disagree, silently selects a different tab than the one whose title was pressed.

`tabNames` (`:337-350`) is rebuilt at runtime: `spotTabItems` arrive asynchronously from categories, and `showPerpsTab` follows `perpDisabled` (`layouts/hooks/useMarketTabsLogic.ts:74`). The native `items` (`:376-402`) are regenerated from the new list, but a press event emitted against the previous layout still carries the old `position`. The sibling `handleNativeSubHeaderPress` (`:614-623`) handles the same race correctly by refusing a `key`/`position` pair that does not match.

## Impact

A tap on, for example, "Perps" that is delivered while the spot category list is being inserted maps to the spot category that now occupies that position: `handleTabPress` writes the wrong tab to the synced market tab atom and scrolls the pager to the wrong page. Rare, but wrong by construction whenever the fallback branch runs, and the conditional as written reads as if `key` were validated when it is not.

## Recommendation

Resolve by `key` and drop the positional fallback, mirroring the sub-header handler:

```ts
const { key } = event.nativeEvent;
if (tabNames.includes(key)) handleTabPress(key);
```

or, if positional lookup must stay, ignore mismatched pairs (`if (tabNames[position] !== key) return;`) rather than falling back to `tabNames[position]`.

[Claude session ↗](https://claude.ai/code/session_01KtprGYr57mvjxBiYLX2uen)

---

huhuanming: 已在 03511164a8 修复。native tab press 现在只按稳定 key 解析，并要求该 key 仍存在于当前 tabNames；旧 layout 事件中的 position 不再参与选择。这样 key 仍存在但位置已变化时会打开正确 tab，key 已移除时则安全忽略。按用户要求本轮未运行本地构建或回归。

zhaono1 (dismissed):

originalix (dismissed):

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.initialData.test.tsx
```diff
@@ -787,6 +787,44 @@ describe('useMarketTokenList initial data', () => {
...
+  it('holds a failed native page until an explicit load-more retry', async () => {
+    mutablePlatformEnv.isNative = true;
+    mutablePlatformEnv.isWeb = false;
+    let latestResult: ReturnType<typeof useMarketTokenList> | undefined;
+    mockFetchMarketTokenList
+      .mockResolvedValueOnce({
+        ...createResponse('first', 'First', 'FIRST'),
+        total: 2,
+      })
+      .mockRejectedValueOnce(new Error('offline'))
```
sidmorizon: # 🟠 P1 New load-more retry test fails deterministically in CI: the rejection is consumed by a second page-1 request, not by `loadMore`

## Problem

`Unit Tests (1/5)` on `0351116` fails in `holds a failed native page until an explicit load-more retry` at line 817 (`expect(isLoadMoreError).toBe(true)` received `false`). The test relies on call order: page 1 resolves, the next call rejects, the third resolves. On native the hook issues **two** page-1 requests on a cold mount, so the ordered `mockRejectedValueOnce` on line 799 is eaten by the second page-1 request and `loadMore()` receives the third mock (success).

Mechanism, all in code this test exercises with `isNative = true`:

- `usePromiseResult` dispatches the mount run from its deps effect (`packages/kit/src/hooks/usePromiseResult.ts:451-478`) **and** a second run from the reconnect-simulation branch (`:525-532`) when `platformEnv.isNative && !isLoadingRef.current && isEmptyResultRef.current && revalidateOnReconnect`. `isLoadingRef` is created at `:482` and never assigned, so `!isLoadingRef.current` is always true; the result is empty because `swrCacheUtils.clearAll()` runs in `beforeEach`; `useMarketTokenList` passes `revalidateOnReconnect: true` (`useMarketTokenList.ts:579`). That second run goes through `requestIdleCallback`, which this file maps to `setTimeout(0)` (`:207-210`).
- Run 1 lands `first` (`total: 2`) and the unlock effect sets `remoteFirstPageLoadedQueryKey`. Run 2 then calls `fetchMarketTokenListForPlatform` again with `page: 1` and receives `Error('offline')`; the native catch path (`useMarketTokenList.ts:490-507`) sets `errorQueryKey` and returns `latestAuthoritativeFirstPageResultRef` (`first`), so `canLoadMore` is true and `waitFor` on line 814 passes.
- `loadMore()` on line 816 now consumes the third mock (`second`), succeeds, and leaves `isLoadMoreError` `false`.

The ordering is fixed by the two zero-delay timers, so this is not a flake; it fails the same way on every run. The production `loadMore` catch (`useMarketTokenList.ts:906-912`) is not what is broken here, but the test that is supposed to prove the Cursor P1 fix never exercises it, and the required check is red.

## Impact

The PR cannot merge with `Unit Tests (1/5)` failing, and the only regression coverage for the `isLoadMoreError` path is currently asserting the wrong request. The sibling native tests in this file already avoid the trap by keying the mock on `page` (`:749-753`, `:833-836`).

## Recommendation

Drive the mock by request shape instead of call order, with a one-shot page-2 failure, for example:

```ts
let page2Attempts = 0;
mockFetchMarketTokenList.mockImplementation(async ({ page }) => {
  if (page === 1) {
    return { ...createResponse('first', 'First', 'FIRST'), total: 2 };
  }
  page2Attempts += 1;
  if (page2Attempts === 1) throw new Error('offline');
  return { ...createResponse('second', 'Second', 'SECOND'), total: 2 };
});
```

and, to make the intent explicit, assert `page2Attempts` is `1` after the first `loadMore()` and `2` after the retry. Separately, the always-true `!isLoadingRef.current` in `usePromiseResult.ts:527` means every native `revalidateOnReconnect` consumer double-fetches on an empty cold mount; that is pre-existing on `x` and outside this PR, but worth a follow-up.

[Claude session ↗](https://claude.ai/code/session_01KtprGYr57mvjxBiYLX2uen)

---

huhuanming: 已在 9c57553808 修复。测试 mock 现在按请求 page 分支：所有 page 1 初始化/重验证都返回首屏数据，第一次 page 2 明确失败，第二次 page 2 明确成功，因此真实覆盖 isLoadMoreError 和显式重试路径，不再依赖调用顺序。验证：useMarketTokenList.initialData.test.tsx 16/16、MobileMarketNativeLists.test.tsx 15/15，yarn agent:check --profile commit 通过。usePromiseResult 的冷启动重复请求属于 x 上既有问题，本 PR 未扩大范围处理。

originalix (approved):

PR #13353, opened 2026-09-12
fix: refresh Aave allowance and reuse Stock trade

zhaono1:
## Summary

- Replace Market desktop Stock trading UI/state with the shared Swap Stock trade container.
- Keep Market-specific behavior through Swap extensions: stock variant selector, oracle price, estimated shares, speed config, and Market sizing data.
- Remove the duplicated Market Stock trade panel and its standalone state/review/input implementations.
- Keep the embedded trade panel mounted when switching Stock variants, so the shared channel can refresh in place.
- Scope Stock pay-token state, preferences, candidates, and selector content by the active Stock network.
- Show partial skeletons while Stock metadata, pay-token candidates, speed config, or quote state is refreshing; avoid blank states and large loading spinners.
- Match the legacy Market behavior by hiding Recent Trades and pending History from the embedded Market panel while retaining them in the regular Swap Stock page.
- Reconcile a seeded Aave borrow allowance with the current chain allowance when opening the portfolio lending dialog, so an already approved WBTC repay skips the stale authorization step.
- Keep the last successfully loaded token image visible while a replacement source loads, preventing a blank frame during Borrow protocol switches and ignoring stale image callbacks.

## Validation

- Desktop Electron verification: switching a Stock variant keeps the trade panel mounted and shows skeleton content during the refresh.
- Desktop Electron verification: switching to the xStocks Solana variant resolves the Solana USDC token and network icon instead of retaining the previous BSC token.
- Desktop Electron verification: switching Borrow from Kamino to Aave and back showed the previous token image underneath the loading replacement until the new image loaded; no empty image frame was observed.
- Targeted Jest suites: 95 existing tests plus 30 borrow allowance tests passed.
- Targeted Borrow and Image suites: 18 tests passed.
- Targeted Oxfmt, Oxlint, and `git diff --check` passed.
- `yarn agent:check --profile commit` and `yarn agent:check --profile pr`: lint, format, dependency, API contract, context, and staged lint checks passed; `tsc-staged` remains blocked by five existing `react-native-pager-view` export/prop errors in `packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx` on the base branch.

Fixes OK-62893
Fixes OK-62868

[OK-62893]: https://onekeyhq.atlassian.net/browse/OK-62893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62868]: https://onekeyhq.atlassian.net/browse/OK-62868


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5e6c915f-a371-53f0-8c15-80c351dc0b51)

sidmorizon:
Claude review session: https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw

packages/kit/src/states/jotai/contexts/marketV2/actions.ts
```diff
@@ -179,6 +179,58 @@ class ContextJotaiActionsMarketV2 extends ContextJotaiActionsBase {
...
 
+  prepareStockTokenDetail = contextAtomMethod(
+    (
+      get,
+      set,
+      target: { tokenAddress: string; networkId: string; isNative?: boolean },
+    ) => {
+      // A stock route can be entered before its token variant is resolved.
+      // Clear the previous identity so it cannot be rendered for the new stock.
+      if (!target.networkId || (!target.tokenAddress && !target.isNative)) {
```
sidmorizon, line 190: 🟡 P2: Stock-list navigation still clears detail

Severity: `non-severe`

Stock-list navigation still calls `clearTokenDetail()` despite having a resolved variant. This path discards loaded state and can expose the generic Swap fallback.

packages/kit/src/views/Borrow/borrowUtils.ts
```diff
@@ -33,6 +33,18 @@ export const BorrowNavigation = {
...
+  popToBorrowHome(navigation: IAppNavigation) {
+    if (platformEnv.isNative) {
+      void safePushToEarnRoute(navigation, ETabEarnRoutes.BorrowHome);
+      return;
+    }
+
+    // Reserve details are pushed on top of BorrowHome on desktop. Pop the
+    // detail route so the existing BorrowProvider and selected market remain
+    // mounted instead of creating a second, empty BorrowHome instance.
+    navigation.pop();
```
sidmorizon: 🟡 P2: Share-route breadcrumb misses Borrow Home

Severity: `non-severe`

On a desktop share-link entry, `popToBorrowHome` blindly pops the current route. Without a Borrow Home predecessor, the breadcrumb does nothing or reveals unrelated content.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useToMarketDetailPage.ts
```diff
@@ -295,7 +295,11 @@ export function useToDetailPage(options?: IUseToDetailPageOptions) {
         closeExtensionPopupAfterExpandTabOpen();
       } else if (options?.switchToMarketTabFirst) {
         if (stockId) {
-          tokenDetailActions.current.clearTokenDetail();
+          tokenDetailActions.current.prepareStockTokenDetail({
```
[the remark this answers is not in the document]
zhaono1: 已处理，提交 8777e9e6c0：useToMarketStockDetailPage 现在统一调用 prepareStockTokenDetail，有已解析的 token variant 时传入 network/token identity，未解析时走安全清理分支。该 hook 同时覆盖 Market Home/native stock list 和 Token Selector 入口；新增回归测试已通过。

packages/kit/src/views/Borrow/pages/borrowHistoryList.utils.ts
```diff
@@ -1,6 +1,73 @@
...
+  tx: IAccountHistoryTx;
+  provider: string;
+  networkId: string;
+  marketAddress: string;
+}): IBorrowAction | undefined {
+  const providerName = provider.toLowerCase();
+  const parsedTags = (tx.stakingInfo?.tags ?? [])
+    .map((tag) => parseBorrowTag(tag))
+    .filter((tag): tag is NonNullable<typeof tag> => Boolean(tag))
+    .filter((tag) => tag.provider === providerName);
```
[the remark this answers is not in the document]
zhaono1: 已处理，提交 8777e9e6c0：replacement history 在新交易没有 stakingInfo 时继承原 pending transaction 的 metadata，Speed Up/Cancel 后仍保留 setCollateral tags，Borrow 历史不会消失。

packages/kit/src/states/jotai/contexts/marketV2/actions.ts
zhaono1: 已处理，提交 8777e9e6c0：Stock list navigation 已改用 prepareStockTokenDetail，同一 variant 保留已有 detail 和 in-flight request，不再无条件清空共享 detail。

packages/kit/src/views/Borrow/borrowUtils.ts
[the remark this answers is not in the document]
zhaono1: 已处理，提交 8777e9e6c0：Borrow breadcrumb 现在调用 EarnNavigation.popToEarnHome({ mode: 'borrow' })，不再依赖当前 route 恰好有可 pop 的 Borrow Home 前置页。

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/StockTradePanelContent.tsx
```diff
@@ -457,12 +459,16 @@ export function StockTradePanelContent(props: IStockTradePanelContentProps) {
...
-  const shouldUseSwapFallbackAction = shouldJumpToMarketTradeFallback({
-    supportSpeedSwap: supportSpeedSwap.enabled,
-    isAccountNetworkSupported: supportSpeedSwap.isAccountNetworkSupported,
-    isWrapped,
-    isRefreshQuote,
-  });
+  // ActionButton enables its fallback action, so keep it out of transient stock loading states.
+  const shouldUseSwapFallbackAction =
+    speedConfigReady &&
+    !isActionDisabled &&
```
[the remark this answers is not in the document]
zhaono1: 已处理，提交 8777e9e6c0：完整 Swap fallback 不再被通用 isActionDisabled 阻断；只有 speed config、交易 loading 和 quote loading 等瞬态会阻止 fallback 渲染。新增回归测试覆盖 quote 不可用但仍应打开完整 Swap 的场景。

packages/kit/src/views/Borrow/borrowUtils.ts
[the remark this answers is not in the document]
zhaono1: 已处理，提交 8777e9e6c0：桌面/Web 的 popToBorrowHome 改为复用 EarnNavigation.popToEarnHome({ mode: 'borrow' })，使用目标 stack 的 popTo(EarnHome)，找不到目标 stack 时再走 navigate fallback，分享路由也会回到 Borrow Home。

zhaono1: 已处理，提交 8777e9e6c0：Borrow breadcrumb 已改为按目标 stack 回到 EarnHome 的 borrow mode，share/deeplink 入口没有父级 Borrow Home 时也能正确恢复 Borrow 首页。

packages/kit-bg/src/services/ServiceHistory.ts
```diff
@@ -3222,6 +3232,9 @@ class ServiceHistory extends ServiceBase {
         xpub,
       });
       if (prevTx) {
+        if (prevTx.stakingInfo && !newHistoryTx.stakingInfo) {
+          newHistoryTx.stakingInfo = prevTx.stakingInfo;
+        }
```
sidmorizon, line 3278: 🟡 P2: Cancellations masquerade as collateral actions

Severity: `non-severe`

Cancelling a collateral transaction copies its `stakingInfo` onto the replacement. Borrow history then presents the cancellation as a collateral action.

Suggested fix:

```
        if (
          replaceTxInfo.replaceType !== EReplaceTxType.Cancel &&
          prevTx.replacedType !== EReplaceTxType.Cancel &&
          prevTx.stakingInfo &&
          !newHistoryTx.stakingInfo
        ) {
          newHistoryTx.stakingInfo = prevTx.stakingInfo;
        }
```

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/StockTradePanelContent.tsx
```diff
@@ -457,12 +459,18 @@ export function StockTradePanelContent(props: IStockTradePanelContentProps) {
...
-    isAccountNetworkSupported: supportSpeedSwap.isAccountNetworkSupported,
-    isWrapped,
-    isRefreshQuote,
-  });
+  // ActionButton enables its fallback action even when its generic disabled
+  // flag is set. Keep it out of transient stock loading states only.
+  const shouldUseSwapFallbackAction =
+    speedConfigReady &&
+    !isLoading &&
+    !quoteLoading &&
```
sidmorizon: # 🟡 P2 Swap fallback gate still hides on non-transient `quoteLoading`

## Problem

`quoteLoading` is not a purely transient signal. The container passes it as `!executionReady || quoteActionLoading` (`StockTradePanel.tsx:869`), and `executionReady` (`StockTradePanel.tsx:449-461`) requires a `paymentToken` that is one of `paymentTokenCandidates`. Those candidates come from `filterStockPayTokenCandidates(defaultTokens)`, which keeps only `USDC`/`USDT` (`swapStockChannelUtils.ts:141`).

When speed swap is unsupported for a network, `useSpeedSwapInit` serves `buildSwapPairFallbackConfig`, whose `defaultTokens` are `swapDefaultSetTokens[networkId]` (`useSpeedSwapInit.tsx:25-56`). For `evm--4663` (Robinhood) that set is ETH + USDG (`SwapProvider.constants.ts`), so `compatibleDefaultTokens` is empty, `paymentToken` never matches a candidate, `executionReady` is permanently `false`, `quoteLoading` is permanently `true`, and `shouldUseSwapFallbackAction` never becomes `true`. The same holds for any unsupported network whose default token set lacks USDC/USDT, and for any other non-loading reason that keeps `executionReady` false.

## Impact

In exactly the "speed swap unsupported" case the fallback exists for, the panel renders `SwapActionsState` with `forceQuoteActionLoading` (an endless spinner) instead of the enabled jump-to-Swap `ActionButton`. The earlier `isActionDisabled` regression is fixed for the no-amount / quote-error case, but it persists for networks without a USDC/USDT default token, leaving no route to Swap.

## Recommendation

Gate on genuinely transient signals only. In `StockTradePanel.tsx:869` pass a dedicated prop (e.g. `stockPreparing = quoteActionLoading || !isReady`) alongside `quoteLoading`, and here use `speedConfigReady && !isLoading && !stockPreparing && shouldJumpToMarketTradeFallback(...)`; keep `quoteLoading` for the skeleton and `SwapActionsState` paths. Add a `StockTradePanelContent.test.tsx` case with `supportSpeedSwap.enabled = false`, `quoteLoading = true`, `hasInitialReady = false` (missing payment token) asserting `actionButtonMock` is still called.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketStockTradeTarget.tsx
```diff
@@ -0,0 +1,44 @@
+import { XStack } from '@onekeyhq/components';
```
sidmorizon, line 1: # 🟠 P1 New module missing from mobile module-id registry breaks the iOS union build

## Problem

This new file is not registered in `apps/mobile/bundle-registry/module-id-registry.json`. The union build's segment allocation (`apps/mobile/scripts/unionBuild.js:571-573` → `apps/mobile/plugins/moduleIdRegistry.js:391`) requires every module in the main/background graphs to have a registry entry and throws otherwise. The failing `Native startup graph budget (24.x)` job on this head aborts with exactly:

```
packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketStockTradeTarget.tsx.
```

Verified locally at `926bfb5c`: `grep MarketStockTradeTarget module-id-registry.json` returns nothing, while the earlier new file in this PR (`packages/kit/src/views/Borrow/pages/modal/borrowTokenSelect.utils.ts`, registry line 20892) was registered. The same commit also leaves 13 stale entries for the modules it deleted (`StockTradePanel.tsx`, `StockTradePanelContent.tsx`, `useSpeedSwapActions.tsx`, `useSwapPanel.ts`, `MarketSwapReviewDialog.tsx`, `TokenInputSection/*`, `market*Utils.ts`), so the registry no longer matches the tree.

## Impact

The iOS union bundle build fails, so the native startup-graph budget check is red on this head and the change cannot ship to mobile until the registry is regenerated. Because the file is reached from the Market detail desktop layout through shared kit code, the module is in both the main and background graphs.

## Recommendation

Regenerate the registry with the repo tooling rather than by hand: run the union build once locally to produce the module-id map, then `yarn workspace @onekeyhq/mobile module-id:update --map <module-id-map.json>` and commit the updated `apps/mobile/bundle-registry/module-id-registry.json` alongside this file. Confirm the regenerated registry also drops the 13 stale entries for the deleted `SwapPanel` modules.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -713,11 +769,20 @@ export function useSwapStockAmountInputState({
...
       const balanceBN = new BigNumber(resolvedInputTokenBalance);
-      setInputAmount(balanceBN.multipliedBy(stage / 100));
+      let amount = balanceBN.multipliedBy(stage / 100);
+      if (inputToken?.isNative) {
+        const reserveGasBN = new BigNumber(reserveGas ?? '');
+        if (reserveGasBN.isFinite() && reserveGasBN.gt(0)) {
+          amount = BigNumber.max(0, amount.minus(reserveGasBN));
+        }
+      }
+      setInputAmount(amount);
```
sidmorizon: 🟡 P2: Partial percentages over-reserve native gas

Severity: `non-severe`

For native stock payments, `onSelectPercentageStage` subtracts the full gas reserve from every percentage. Selecting 25% of 1 ETH with 0.01 reserved enters 0.24 ETH instead of 0.25.

Suggested fix:

```
      const balanceBN = new BigNumber(resolvedInputTokenBalance);
      const amount = balanceBN.multipliedBy(stage / 100);
      setInputAmount(amount);
```

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
```diff
@@ -132,11 +244,19 @@ function MarketEmbeddedSwapDraft({
   }, []);
 
   // Keep only the draft while inactive, since other routes share the modal store.
-  return disabled ? null : (
+  return disabled ? (
+    <MarketEmbeddedSwapLoading />
+  ) : (
```
sidmorizon: 🟡 P2: Unavailable trades show endless skeleton

Severity: `non-severe`

When trading is permanently unavailable, `MarketEmbeddedSwapDraft` renders a loading skeleton instead of no ticket. Unsupported assets appear to load forever and retain a 520-pixel placeholder.

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
```diff
@@ -1,22 +1,40 @@
...
 import type {
   ISwapInitParams,
+  ISwapStockSpeedConfig,
+  ISwapStockTradeConfig,
   ISwapToken,
 } from '@onekeyhq/shared/types/swap/types';
 
 import { useSpeedSwapInit } from '../components/SwapPanel/hooks/useSpeedSwapInit';
 
+import { MarketStockTradeTarget } from './components/MarketStockTradeTarget';
```
sidmorizon, line 25: # 🟠 P1 New static import chain makes `MarketDetailEmbeddedSwap.test.tsx` fail to load

## Problem

This import is new in the increment and adds a module-scope chain: `MarketEmbeddedSwap.tsx:25` → `MarketStockTradeTarget.tsx:5` (`StockTokenInfoPopover`) → `StockTokenInfoPopover.tsx:18` (`Token` from `@onekeyhq/kit/src/components/Token`) → `Token.tsx:30` (`useAccountData`) → `useAccountData.ts:15` (`backgroundApiProxy`) → kit-bg `BackgroundApiProxy` → shared `backgroundDecorators` → `backgroundUtils` → `engineConsts` → `networkIds` → `presetNetworks.ts:30` (`ENetworkStatus.LISTED`).

`presetNetworks.ts:7` imports `ENetworkStatus` from `@onekeyhq/shared/types`. `MarketDetailEmbeddedSwap.test.tsx:78-80` replaces that module with `{ EAccountSelectorSceneName: { swap: 'swap' } }`, so `ENetworkStatus` is `undefined` and module evaluation throws. This is the exact failure in the `Unit Tests (4/5)` job on `926bfb5c`:

```
FAIL packages/kit/src/views/Market/MarketDetailV2/components/MarketDetailEmbeddedSwap.test.tsx
● Test suite failed to run
TypeError: Cannot read properties of undefined (reading 'LISTED')
    at Object.LISTED (packages/shared/src/config/presetNetworks.ts:30:28)
    ...
    at (packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketStockTradeTarget.tsx:14:32)
    at (packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx:18:33)
```

At `0f27ff25` the suite's mock set covered every module-scope import of `MarketEmbeddedSwap.tsx`; the deleted `StockTradePanel.tsx` imported `StockTokenInfoPopover` only via `SwapPanel.tsx`, which this test never loaded. The test file itself has no diff in this increment. There is no import cycle: `packages/shared/types/index.ts` defines the enum at line 20 and has only type-level imports, so the real bundle is unaffected.

## Impact

CI `Unit Tests` is red on this head, and the Market embedded-swap regression coverage (init-param seeding via `buildMarketEmbeddedSwapInitParams`, draft handling) no longer runs. `MarketEmbeddedSwap` is also rendered by `TokenDesktopLayout.tsx:31` and `TopCoinsDesktopLayout.tsx:49`, so stock-only UI is now pulled into every desktop Market detail layout.

## Recommendation

Preferred: drop this import and the `stockTradePortfolioData` / `stockTradeResolvedVariantKeys` plumbing (lines 36-37, 117-118, 178-186, 204-205, 229-230, 258-259, 282-283, 298-299), and have `StockDesktopLayout.tsx` (around lines 1401-1414) render `<MarketStockTradeTarget token={swapToken} portfolioData={stockPortfolioData} resolvedVariantKeys={resolvedVariantKeys} />` and pass it through the already-supported `stockTradeHeader` prop (`MarketEmbeddedSwap.tsx:34`, resolved at 179). Minimal alternative: add `jest.mock('../layouts/components/MarketStockTradeTarget', () => ({ MarketStockTradeTarget: () => null }))` next to `MarketDetailEmbeddedSwap.test.tsx:87`. Either way, run that suite before pushing.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -320,6 +352,29 @@ export function useSwapStockChannel() {
...
+    if (
+      !nextStockToken ||
+      getTokenIdentityKey(nextStockToken) === currentStockTokenKey
+    ) {
+      return;
+    }
+    setPayTokenState(undefined);
+    payTokenSnapshotRef.current = undefined;
+    manualStockPayTokenKeyRef.current = '';
+    selectStockSwapToken(nextStockToken, { resetReceiveAmount: true });
```
sidmorizon, line 387: # 🟡 P2 Controlled variant switch writes a cross-network execution pair before the pay-token hook corrects it

## Problem

The new controlled effect clears `payTokenSnapshotRef` and then calls `selectStockSwapToken(nextStockToken, …)`. That calls `syncStockExecutionTokens({ stockToken })` (line 345-347), whose `payToken` default is `payTokenSnapshotRef.current ?? stockOwnedPayToken` (lines 264-265). The ref was just nulled, so the default falls through to `stockOwnedPayToken` captured in the `useCallback` closure of the current render. That value is `resolveStockPayTokenState({ channelToken: payTokenState, … }).displayToken` (lines 216-224), and `channelToken` is the previous network's pay token because `setPayTokenState(undefined)` (line 367) has not re-rendered yet. Only `swapPairStockPayToken` and `stockPairPayToken` are network-filtered (lines 192-214); `channelToken` is not.

`selectStockExecutionTokens` (`states/jotai/contexts/swap/actions.ts:1087-1107`) therefore sets `swapSelectFromTokenAtom` to the old-network USDC and `swapStockExecutionTokensAtom` to `{ old USDC → new stock }`. `useSwapQuote.ts:516-530` only rejects when `fromToken.networkId !== activeAccount.networkId` or from ≡ to, so with the preserved draft amount (`marketEmbeddedSwapUtils.ts:27-35` keeps `fromTokenAmount`) a STOCK quote for a cross-chain pair is issued until `useSwapStockPayTokens` finishes its speed-config and token-detail fetches and calls `selectPayToken(nextPayToken, false)` (`useSwapStockPayTokens.ts:683`). The comment above the effect says the previous network's pay-token override is cleared; the execution pair is not.

## Impact

Each in-place variant switch across networks in the embedded Market panel (BSC ↔ Solana ↔ Robinhood variants) fires one invalid quote request and can flash the quote/alert state before the corrected pair re-quotes. The pay-token input is masked by `isStockTradeChannelInitializing`, so this is a quote-layer transient, not a fund-safety issue.

## Failure scenario

Stock detail on the BSC variant, pay token USDC, amount 12 entered → switch to the Solana variant → store holds `fromToken = BSC USDC`, `toToken = Solana stock` → `useSwapQuote` fires (account still BSC) → a few hundred ms later `selectPayToken` writes Solana USDC and re-quotes.

## Recommendation

Resolve the pay token before selecting: in this effect compute `const scopedPayToken = stockOwnedPayToken?.networkId === nextStockToken.networkId ? stockOwnedPayToken : undefined;` and forward it to `syncStockExecutionTokens` (extend `ISelectStockSwapTokenOptions` with `payToken`), and make `selectStockExecutionTokens` (`actions.ts:1087-1092`) clear the pay-side `swapSelectFromTokenAtom` / `swapSelectToTokenAtom` when `payToken` is explicitly `undefined` instead of leaving the previous atom in place. Alternatively add a `fromToken.networkId === toToken.networkId` guard to the STOCK branch at `useSwapQuote.ts:516`.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -378,20 +417,53 @@ export function ImageV2({ style: defaultStyle, ...props }: IImageV2Props) {
...
+          }
+          onLoad={(event: ImageLoadEvent) =>
+            handleLoad(
+              event,
+              resolvedSourceIdentity,
+              resolvedSource as ImageURISource,
+            )
+          }
+          onLoadEnd={() => handleLoadEnd(resolvedSourceIdentity)}
+          onLoadStart={() => handleLoadStart(resolvedSourceIdentity)}
```
sidmorizon: # 🟠 P1 Inline image handlers fail lint, restart web image loads on every re-render, and the stale-callback guard is inert

## Problem

Before this commit `ImageComponent` received the stable `useCallback` handlers directly (`onError={handleError}` etc., `907552c0:ImageV2.tsx:389-392`). To thread `resolvedSourceIdentity` into them, these lines now create four arrow functions inside the `content` memo. Three consequences:

1. **Lint is red on this head.** The `lint (24.x)` job on `31d874c5` reports exactly six oxlint errors that all map to this file: four `JSX attribute values should not contain functions created in the same scope` (these four arrows), `'source' is already declared in the upper scope` (`handleLoad`'s third parameter at line 290 shadows the `source` prop destructured at line 141), and `React Hook useMemo has a missing dependency: 'resolvedSourceIdentity'` (the `content` memo deps at 470-486 do not list it although the arrows read it).
2. **The identity guard cannot reject anything.** Each handler compares its `identity` argument with the `resolvedSourceIdentity` it closes over (263-266, 291-293, 354-356, 364-366) and lists that value as a dep, and `content` lists the handlers as deps. Both sides are therefore always captured by the same render, so `identity !== resolvedSourceIdentity` is a tautology. The "ignoring stale image callbacks" behavior described in the PR summary is not implemented.
3. **Web/desktop/extension loads restart on every parent render.** `ImageV2.tsx` is the non-native file (native has `ImageV2.native.tsx`). `content` recomputes whenever `imageProps` (a fresh rest object from `usePropsAndStyle`, 127-163) or the inline `fallback` JSX changes, which is essentially every render. react-native-web's `Image` lists `onError`, `onLoad`, `onLoadEnd` and `onLoadStart` in the deps of the effect that starts the `ImageLoader` request and aborts the pending one on cleanup (react-native-web 0.19-0.21 `exports/Image/index.js`). New callback identities each render therefore abort and restart the in-flight load and re-fire `handleLoadStart` (`setIsImageLoaded(false)`, placeholder timer restart, `onLoadStart` prop call).

## Impact

CI lint blocks the PR. Functionally, any token or network logo inside a frequently re-rendering host on web/desktop/extension (live-price rows, the quote-polling Swap and Market panels, Borrow cards) has its load aborted on each tick, so uncached images appear late and the `bgStrong` background flickers behind transparent logos. This regresses the 140+ `Image` consumers in `packages/kit`, not only the Borrow protocol switch the commit targets.

## Failure scenario

Market stock detail with the embedded Swap panel polling quotes → a slow-CDN token logo starts loading → each quote tick re-renders the row → `content` recomputes → four new arrows → react-native-web aborts the request and calls `onLoadStart` again → the logo only appears once a tick lands after the browser cache is warm.

## Recommendation

Keep passing `handleError` / `handleLoad` / `handleLoadEnd` / `handleLoadStart` as stable references. For the stale-callback guard, hold the live identity in a ref updated during render (`latestIdentityRef.current = resolvedSourceIdentity`) and build the four handlers in one `useMemo` keyed on `resolvedSourceIdentity` that captures `const identity = resolvedSourceIdentity` and compares against `latestIdentityRef.current` at call time. Rename the `source` parameter at line 290 (e.g. `loadedSource`) and add `resolvedSourceIdentity` to the `content` deps only if it is still read there. Run `yarn lint` on the file before pushing; the six errors above must clear.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/components/src/primitives/Image/ImageV2.tsx
```diff
@@ -355,11 +388,17 @@ export function ImageV2({ style: defaultStyle, ...props }: IImageV2Props) {
...
       scheduleRetry,
+      resolvedSourceIdentity,
       shouldUseRawSourceFallback,
     ],
   );
 
   const ImageComponent = ReactNativeImage as ComponentType<any>;
+  const hasVisibleLoadedImage = Boolean(
+    loadedImage && loadedImage.identity !== resolvedSourceIdentity,
+  );
```
sidmorizon: # 🟡 P2 Previous image survives an empty source and is shown under an unrelated next source

## Problem

`loadedImage` is only cleared on terminal error (line 380). When `source` becomes empty or `undefined`, `content` returns `null` (417-419) but `loadedImage` is retained. When a new, unrelated non-empty source arrives, `hasVisibleLoadedImage` is `true` because `loadedImage.identity !== resolvedSourceIdentity`, so the stale image is mounted as the "previous" layer (450-464), and it also suppresses the placeholder / skeleton (513-515) and the `bgStrong` background (495).

## Impact

The wrong token or protocol logo is displayed for the duration of the new image's load, which contradicts the commit's intent of showing the *replaced* image only during a direct A→B transition. The placeholder is hidden in that window as well.

## Failure scenario

`AmountInput` (`packages/kit/src/components/AmountInput/index.tsx:274`, constant `recyclingKey`) shows token A → the user clears the selection (source `undefined`, nothing rendered) → the user selects token B → A's logo renders as the "previous" overlay until B finishes loading.

## Recommendation

Reset `loadedImage` when the resolved source becomes empty, e.g. an effect on `resolvedSourceIdentity` that calls `setLoadedImage(null)` when `isEmptyResolvedSource(resolvedSource)`; or track `previousIdentityRef` and treat `loadedImage` as "previous" only when it was the immediately preceding non-empty identity (ImageV2.tsx:397-401). Add an `ImageV2` test for the A → empty → B sequence asserting no previous layer is rendered.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
zhaono1: 已复核当前 head：该评论针对 926bfb5c 的旧公式，现已由 e7542a2fd1 修正。当前百分比先对 native balance 扣除 reserve，再乘 stage；例如 1 - 0.2 的可用余额，50% 是 0.4，100% 是 0.8，相关测试已覆盖。这里的百分比语义是可用余额，不是扣 reserve 前的总余额，因此原问题在当前 head 不成立，关闭。

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
zhaono1: 已复核当前 head：该评论针对 926bfb5c 中 disabled 分支持续渲染 MarketEmbeddedSwapLoading。b251b7508d 已将 terminal disabled 改为不渲染 loading；TopCoins 调用方有明确 unavailable panel，transient 初始化和切换仍使用局部 skeleton。因此评论描述的 520px 永久 loading 在当前 head 已不存在，原评论已过期，关闭。

packages/kit-bg/src/services/ServiceHistory.ts
zhaono1: 复核后该评论仍成立，但需要区分 Speed Up 与 Cancel：当前 ServiceHistory 会在设置 replacement type 之前无条件继承 prevTx.stakingInfo；Cancel 交易由 zero-value + empty data 生成，Borrow history 会据此把新记录识别为 setCollateral。此前回复只覆盖 Speed Up 的 metadata 保留，没有解决 Cancel 被标记为抵押操作的问题。我已重新打开此线程，先保留待确认 Cancel history 的产品语义。

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
zhaono1: 补充核对最近聊天：b251b7508d 确实移除了 terminal disabled 的永久 skeleton，但 DesktopLayout 会把 cold-start/metadata 未就绪也并入 shouldDisableTrade，Stock/Token 外层随后得到空 trade column。此前用户明确要求未就绪阶段不能出现空容器，因此这条评论的旧现象虽已变更，terminal 与 transient readiness 的分层仍需产品确认；我先保留线程，不把它作为纯误报关闭。

zhaono1: 最终结论：按原评论所指的“terminal unsupported 状态持续显示 520px skeleton”，当前 head 已由 b251b7508d 移除，故该线程保持 resolved。最近聊天中发现的 cold-start metadata 未就绪时可能出现空 trade column，是独立的 UX follow-up，不等同于本评论的永久 skeleton 断言。

packages/kit-bg/src/services/ServiceHistory.ts
zhaono1: Fixed in 9a845495a2. Cancel replacements still retain stakingInfo because the pending collateral-state guards need that metadata, but getBorrowHistoryActionForLocalTx now ignores transactions with replacedType === Cancel, so Borrow history no longer renders a cancellation as setCollateral. SpeedUp remains visible when its metadata is present. Added regression coverage for Cancel and SpeedUp; the targeted Borrow tests pass (8 utility cases and 38 related collateral cases).

packages/kit/src/views/Borrow/pages/borrowHistoryList.utils.ts
```diff
@@ -1,6 +1,80 @@
...
+  tx: IAccountHistoryTx;
+  provider: string;
+  networkId: string;
+  marketAddress: string;
+}): IBorrowAction | undefined {
+  // Keep cancellation metadata on the local transaction for pending-state
+  // guards, but do not render the replacement itself as a borrow action.
+  if (tx.replacedType === EReplaceTxType.Cancel) {
+    return undefined;
+  }
```
sidmorizon, line 35: # 🟠 P1 Cancel guard relies on `replacedType`, which the indexer merge drops, so a confirmed cancel still renders as setCollateral

## Problem

`replacedType` exists only on the local history record (`ServiceHistory.ts:3242` and `3252-3255`). Borrow history loads its local rows through `serviceHistory.fetchAccountHistory` (`BorrowHistoryList.tsx:314-324`), whose result is `finalPendingTxs + confirmedTxsToSave + onChainHistoryTxs` (`ServiceHistory.ts:1737-1747`). Once the indexer lists the cancel tx, the local record is replaced by `mergeLocalTxDisplayFields(...)` (`ServiceHistory.ts:2283-2289`). That merge spreads the on-chain object and re-attaches the PR-added `stakingInfo` carry-over (`ServiceHistory.ts:741-743` and `812-814`) but never `replacedType`; `rg replacedType packages/kit-bg/src` shows no path that copies it onto an indexer-built tx.

So the merged, `Confirmed` cancel row carries the inherited `borrow:aave:setCollateral…` tags (from `ServiceHistory.ts:3235-3237`) and no `replacedType`. It is not in `HIDDEN_LOCAL_BORROW_HISTORY_STATUSES` (`BorrowHistoryList.tsx:184-188`), the remote borrow history does not know a zero-value self-transfer so `remoteTxHashes` (`:358-360`) cannot dedupe it, and this guard falls through to tag parsing and returns `'setCollateral'`.

The same inheritance also reaches the Staking/Earn pending lists: `fetchLocalStakingHistory` (`ServiceStaking.ts:324-330`) and `useStakingPendingTxsByInfo` filter pending txs by `stakingInfo.tags` only, and `Staking/pages/HistoryList/index.tsx:363-379` renders `stakingInfo.label` and amount. A pending cancel of a stake or withdraw therefore shows as a pending "Stake"/"Withdraw" row until it confirms. Before this PR the cancel had no `stakingInfo` and never matched either consumer.

## Impact

The symptom this commit targets is fixed only while the cancel is pending or not yet indexed. After confirmation the Borrow "All" view shows a bogus completed collateral action for a tx that changed nothing (title via `getBorrowTxTitle` from the inherited tags, `BorrowHistoryList.tsx:396-397`), and Earn pending lists briefly show cancels as the original action. The new tests only exercise a raw local record with `replacedType` present, so they cannot catch the merged case.

## Failure scenario

Toggle collateral → pending → open History Details and Cancel → cancel confirms → indexer returns the 0-value self-transfer → next Borrow history load: the original is `Dropped` (hidden), and the cancel itself appears as a confirmed "Collateral" row.

## Recommendation

Pick one root-cause fix rather than per-consumer guards:
- (a) Do not inherit `stakingInfo` for `replaceType === Cancel` at `ServiceHistory.ts:3235-3237`, and drive `hasPendingSetCollateral` from the dropped `prevTx` (its `replacedNextId` is already set two lines below) or a dedicated pending marker; or
- (b) if the metadata must stay for the pending guards, also preserve `replacedType` and `replacedPrevId` in both branches of `mergeLocalTxDisplayFields` (`ServiceHistory.ts:741-743`, `812-814`) so this guard survives the merge, and skip `replacedType === Cancel` in `fetchLocalStakingHistory` (`ServiceStaking.ts:324-330`) as well.

Either way, add a test that feeds a merged record (on-chain shape, `stakingInfo` present, `replacedType` as it would arrive from `fetchAccountHistory`) through `getBorrowHistoryActionForLocalTx` and asserts `undefined`.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit/src/views/Market/MarketDetailV2/components/SwapPanel/StockTradePanelContent.tsx
zhaono1: Reviewed against the current head: the referenced StockTradePanelContent quote-loading path was removed and reworked by the x sync, so this line-level finding is outdated. The current fallback flow is covered by the updated Market embedded-swap tests; cold-start readiness is handled separately from the terminal unsupported-trade state. Keeping resolved/outdated.

packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketStockTradeTarget.tsx
zhaono1: Fixed in 2edc76cd82 and included in the merged head 54f05a5e4e. Added packages/kit/src/views/Market/MarketDetailV2/layouts/components/MarketStockTradeTarget.tsx as module 21857 to apps/mobile/bundle-registry/module-id-registry.json. The strict module-id check, main/background startup budgets, and bundle architecture check all pass locally, and the registry now matches the generated module map without stale deleted SwapPanel entries. Keeping resolved.

packages/kit/src/views/Swap/hooks/useSwapStockTradeInputs.ts
```diff
@@ -713,11 +769,20 @@ export function useSwapStockAmountInputState({
         syncStockTokenDetail(authoritativeStockInputToken);
       }
       const balanceBN = new BigNumber(resolvedInputTokenBalance);
-      setInputAmount(balanceBN.multipliedBy(stage / 100));
+      let amount = balanceBN.multipliedBy(stage / 100);
+      if (inputToken?.isNative) {
+        const reserveGasBN = new BigNumber(reserveGas ?? '');
+        if (reserveGasBN.isFinite() && reserveGasBN.gt(0)) {
+          amount = BigNumber.max(0, amount.minus(reserveGasBN));
```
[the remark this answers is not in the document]
zhaono1: This finding duplicates the native percentage-reserve issue. Fixed in e7542a2fd1: native available balance is reduced by the gas reserve before applying the selected percentage; targeted tests cover the reserve semantics at partial and maximum stages. Keeping resolved/outdated.

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
zhaono1: Fixed in d8679607f1 by isolating the Market embedded-swap mocks from the module-scope network enum dependency. The affected MarketDetailEmbeddedSwap test suite now loads and passes. Keeping resolved.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
zhaono1: Fixed in b251b7508d: controlled stock variant changes now clear and select pay-token state within the active network scope, preventing a stale cross-network execution pair. The related selector and channel tests pass. Keeping resolved.

packages/components/src/primitives/Image/ImageV2.tsx
zhaono1: The ImageV2 change is no longer part of this PR; the current head has no diff for packages/components/src/primitives/Image/ImageV2.tsx. This finding is therefore outdated and out of scope for this branch. Keeping resolved/outdated.

packages/components/src/primitives/Image/ImageV2.tsx
zhaono1: The ImageV2 change is no longer part of this PR; the current head has no diff for packages/components/src/primitives/Image/ImageV2.tsx. This finding is therefore outdated and out of scope for this branch. Keeping resolved/outdated.

packages/kit/src/views/Borrow/pages/borrowHistoryList.utils.ts
zhaono1: Fixed in 489b7acb1f. History merging now carries local replacement linkage (replacedType, replacedPrevId, replacedNextId, and replacedMethod) when the indexer omits it, so a confirmed Cancel remains distinguishable from the collateral metadata it inherits. Pending staking/Borrow filters also exclude Cancel replacements while retaining stakingInfo for pending-state guards. Added merged-record and pending-filter regression tests; the targeted suites and commit checks pass. Keeping resolved.

packages/kit-bg/src/services/ServiceHistory.ts
zhaono1: Follow-up in 489b7acb1f also preserves replacedType/replacedPrevId through indexer merges and excludes cancelled replacements from staking pending lists, covering the confirmed-indexer path as well as the original pending-history case. The Cancel history regression tests pass. Keeping resolved.

packages/kit/src/views/Earn/hooks/useStakingPendingTxs.ts
```diff
@@ -641,6 +642,7 @@ export const useStakingPendingTxsByInfo = ({
         }
 
         return pendingTxs.filter((tx): tx is IStakePendingTx => {
+          if (tx.replacedType === EReplaceTxType.Cancel) return false;
```
sidmorizon, line 645: # 🟡 P2 Cancel filter in `useStakingPendingTxsByInfo` also unlocks the Borrow pending-state guards while a cancel is in flight

## Problem

This hook returns a single `filteredTxs` / `pendingCount` that serves two purposes: display of pending rows, and the Borrow lock guards. `BorrowHomePage.tsx:90` takes `filteredTxs` as `borrowPendingTxs`, passes it as `pendingTxs` into `BorrowHome` (`:106`) and the provider (`BorrowHome.tsx:94-98`), and `CollateralSwitchCell.tsx:264-271` feeds it to `hasPendingSetCollateral`. `BorrowEModeSwitch/index.tsx:153-176` uses the same hook's `pendingCount` for `isEModePendingGuardActive`.

When the user cancels a `setCollateral` / `setEMode` tx, `ServiceHistory.ts:3279` marks the original `Dropped`, so the cancel is the only pending record carrying the borrow tag. The new line 645 removes it before tag matching, so both guards see nothing pending. This contradicts the intent the PR itself records at `borrowHistoryList.utils.ts:31-32` ("Keep cancellation metadata on the local transaction for pending-state guards") and the comment in `getLocalReplacementFields` about pending-state guards.

## Impact

The collateral switch and EMode selector become interactive while the cancel is still unconfirmed and the original toggle may still mine. `isPending` flips false at cancel submission rather than at cancel confirmation, so the `onRefresh` callbacks (`refreshAllBorrowData`, `refreshManagementState`) fire early and do not fire again when the cancel actually confirms.

## Failure scenario

Enable collateral → pending → Cancel from History Details → return to Borrow: the switch is interactive; toggling again queues a second collateral tx behind a cancel that may lose the nonce race; when the cancel confirms later, no refresh runs and the displayed state can disagree with chain.

## Recommendation

Keep cancels in the guard path and drop them only for display. Either revert line 645 and apply the Cancel exclusion at the display consumers through the existing `filter` param (`EarnHome.tsx:436`, `PortfolioTabContent.tsx:200`), or add an option such as `includeCancellations` that defaults to `true` for `BorrowHomePage.tsx:90`, `BorrowEModeSwitch/index.tsx:156` and `BorrowEModeNeedAction`, and `false` for list rendering. Adjust `useStakingPendingTxs.test.tsx:188-212` to assert the guard path still contains the cancel.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

packages/kit-bg/src/services/ServiceHistory.mergeLocalTxDisplayFields.test.ts
```diff
@@ -0,0 +1,88 @@
...
+        },
+        actions: [],
+      },
+    });
+
+    const merged = mergeLocalTxDisplayFields({
+      localTx,
+      onChainHistoryTx,
+    });
+
```
sidmorizon, line 56: # 🟡 P2 New merge test only covers the labeled-indexer branch, not the branches the fix targets

## Problem

Both cases in this suite set `payload: { label: 'Transfer', … }` (lines 45-53 and 77-85), so they exercise only the early `payload.label` return of `mergeLocalTxDisplayFields` (`ServiceHistory.ts:776-784`). The two unlabeled paths that the fix also patches, the non-staking early return at `ServiceHistory.ts:795` and the main return at `852-855`, have no assertion on `replacedType`. Those are precisely the paths taken for metadata-only Borrow txs on networks where the indexer supplies no label, which is the scenario the earlier finding described. There is also no test feeding a merged record through `getBorrowHistoryActionForLocalTx`.

## Impact

A later edit that reorders the spreads or removes `...localReplacementFields` from the main return would leave CI green while re-introducing the confirmed-cancel-as-setCollateral row.

## Recommendation

Add a case with a `payload` lacking `label` plus local `stakingInfo` (hits line 852), and a case with neither staking nor internal-swap metadata (hits line 795), each asserting `replacedType` and `replacedPrevId` survive. Optionally add a case in `borrowHistoryList.utils.test.ts` that passes such a merged record to `getBorrowHistoryActionForLocalTx` and expects `undefined`.

[Claude session ↗](https://claude.ai/code/session_01QAeVMo7EPtceYbJ7EXwbJw)

---

originalix (approved):

zy2000-star (approved):

PR #13354, opened 2026-09-12
fix: enable mobile indicator settings and color editing OK-61423

limichange:
[OK-61423](<https://onekeyhq.atlassian.net/browse/OK-61423>)

----------

Fixes OK-61423: mobile chart users can open a dedicated indicator settings page and edit individual indicators in a bottom sheet. Reuses the existing settings fields, color palette, and reset/confirm behavior with a borderless mobile layout. Adaptive popovers make line colors editable on mobile, and a bounded scroll area keeps the footer accessible on short screens.

Opening Settings preserves pending indicator selections and waits for the native background write before closing and navigating. A failed write shows an error and preserves the selection for retry, including when the user undoes the failed edit. Successful writes advance the comparison baseline; later close/navigation failures do not repeat the save. The active button shows progress, duplicate actions are disabled, and local error logs identify the failed phase. Saving an individual indicator preserves its active state and other indicator settings; Market and Swap retain separate settings. Registers the new modules, plus missing registrations for the restored market modules introduced by the merge from x, so strict mobile builds can resolve them.

Validation:

- `CI=true yarn lint`: passed across the full repository, including TypeScript, Oxlint, formatting, module IDs, and Electron checks.
- `yarn agent:check --profile pr --pr 13354`: local checks passed; remote CI and review approval are tracked on the PR.
- GitHub CI on `9a2c5a3029`: all 25 checks passed; no unresolved review threads.
- Targeted Jest suites for selection/navigation, chart container behavior, and short-screen layout passed: 4 suites, 53 tests. Regression tests cover delayed persistence, Market/Swap Promise propagation, failure/retry for both entry points, corrective writes after undo, progress state, and close/navigation failure after a successful save. Indicator merge and confirmation suites also passed.
- Android emulator at 1080 × 1100: scrolled to MA3, verified the fixed Confirm button, changed the color, saved, and reopened to verify persistence.
- Browser at 390 × 844: opened the MA color palette, changed a line from green to pink, confirmed, and reopened to verify the saved color.
- Strict iOS union build, fresh module-map validation, and bundle architecture check: passed.
- Native startup budgets: main 2,641 modules / 13.09 MB; background 2,439 modules / 18.31 MB; both passed.

iOS device UI verification is still pending.

[OK-61423]: https://onekeyhq.atlassian.net/browse/OK-61423?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-07a3e18f-2a66-57c0-816d-7217f62d11be)

sidmorizon:
Claude review session: https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB

packages/kit/src/views/Market/router/index.tsx
```diff
@@ -18,6 +18,10 @@ const MobileTokenSelectorModal = LazyLoadPage(
 const MarketChartSettingsModal = LazyLoadPage(
   () => import('../MarketDetailV2/components/MarketChartSettingsModal'),
 );
+const MarketIndicatorSettingsModal = LazyLoadPage(
+  () =>
+    import('@onekeyhq/kit/src/components/TradingView/TradingViewNative/TradingViewMobileIndicatorSettingsPage'),
```
originalix, line 23: P1: [Missing mobile module-ID registry entry]

When the new `MarketIndicatorSettings` route is lazy-loaded in a production or union mobile bundle, Metro resolves `TradingViewMobileIndicatorSettingsPage.tsx` through this import. The file has no entry in `apps/mobile/bundle-registry/module-id-registry.json`, and strict registry mode throws before the bundle can be produced.

Regenerate and commit the module-ID registry entry for this new screen, then run the mobile module-ID or bundle check.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -290,6 +292,20 @@ export function IndicatorListDialogContent({
...
+          justifyContent="flex-start"
+          variant="tertiary"
+          onPress={async () => {
+            await dialog.close();
+            onSettingsPress();
+          }}
+        >
+          {intl.formatMessage({ id: ETranslations.global_settings })}
+        </Button>
+      ) : null}
```
sidmorizon, line 368: # 🟡 P2 New "Settings" button silently discards pending indicator toggles

## Problem

The new entry button at `NativeIndicatorSelector.tsx:295-308` does `await dialog.close(); onSettingsPress();` without committing the user's pending selection.

In `IndicatorListDialogContent`, tapping an indicator pill is purely local: `handleIndicatorPress` (lines 220-244) only writes `activeIndicatorValuesRef.current` and `setActiveIndicatorValues`. The selection reaches the chart exclusively through `handleConfirmPress` (lines 246-257), which calls `commitNativeIndicatorSelection` (`indicatorSelector/indicatorUtils.ts:129-167`) → `onSelectionConfirm` → `handleIndicatorSelectionConfirm` → `reconcileTradingViewNativeIndicatorActiveState` in `TradingViewNativeContainer.tsx`.

The dialog is created by `Dialog.show` in `TradingViewNativeChartControlsContainer.tsx:143-172` with no `onClose`/dismiss commit hook, so closing from the Settings button drops every pending toggle.

## Impact

Mobile flow (Market detail, Swap K-line, AssetDetails chart): open the Indicators dialog, tap MA and RSI to enable them, then tap **Settings** to configure one of them. The dialog closes, the new indicator-settings modal opens, and MA/RSI were never enabled. After editing and confirming in that modal, the user returns to a chart showing neither indicator, with no feedback that their taps were dropped. This is the natural flow the new entry point invites — enable an indicator, then go configure it — so the loss is easy to hit and confusing, since the settings page will also show parameters for indicators the user believes they just turned on.

## Recommendation

Commit before closing in the Settings `onPress` (`NativeIndicatorSelector.tsx:301-304`), mirroring `handleConfirmPress`:

```ts
commitNativeIndicatorSelection({
  indicators,
  nextActiveIndicatorValues: activeIndicatorValuesRef.current,
  onSelect,
  onSelectionConfirm,
  originalActiveIndicatorValues: originalActiveIndicatorValuesRef.current,
});
await dialog.close();
onSettingsPress();
```

`commitNativeIndicatorSelection` already no-ops when nothing changed (`indicatorUtils.ts:149-151`), so this is safe when the user made no toggles. Extracting the shared commit into a `useCallback` used by both `handleConfirmPress` and the Settings button keeps the two paths from drifting. If discarding is intentional here, the button should at least be positioned/labelled so it does not read as an in-flow continuation of the pending selection.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

packages/kit/src/components/TradingView/TradingViewChartControls/chartSettings/TradingViewIndicatorContent.tsx
```diff
@@ -94,32 +98,37 @@ function TradingViewIndicatorContent({
...
   );
+  const contentPaddingTop = mobileLayout ? 0 : 20;
+  const contentPaddingHorizontal = mobileLayout ? 0 : 16;
 
   return (
     <ScrollView
       testID="trading-view-indicator-settings-content"
       h={compact ? undefined : TRADING_VIEW_INDICATOR_BODY_HEIGHT}
-      flex={compact ? 1 : undefined}
+      flex={compact && !mobileLayout ? 1 : undefined}
```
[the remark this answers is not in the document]
limichange: Fixed in a5cf08a70f. The mobile sheet now has an explicit bounded height; the content Stack and compact ScrollView use flex=1/minHeight=0, while the footer keeps flexShrink=0. The layout regression covers a 390px-high viewport. I also verified the real Android app at 1080×1100: MA3 scrolls into view while Confirm stays fixed and clickable, and selecting a color, confirming, and reopening preserves it.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
limichange: Fixed in 305bec372c. Settings and Confirm now share commitSelection, which reuses commitNativeIndicatorSelection. Settings commits the pending selection before closing the dialog and waits for close to finish before navigating. Regression tests verify pending MA/RSI selections are committed, navigation waits for the close Promise, and an unchanged selection does not trigger a redundant commit.

packages/kit/src/views/Market/router/index.tsx
limichange: Fixed in a5cf08a70f and 305bec372c. Registered TradingViewMobileIndicatorSettingsPage.tsx and mobileIndicatorSettingsUtils.ts using the repository's updateRegistryFromModulePaths/writeRegistry helpers. After the remote merge from x, I also registered the three restored AssetDetails market modules that were missing IDs; existing IDs were preserved.

The same strict iOS union build used by CI now passes locally. The fresh module map check, both startup budgets (main: 2,641 modules / 13.09 MB; background: 2,439 modules / 18.31 MB), and bundle architecture check also pass.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -290,6 +296,21 @@ export function IndicatorListDialogContent({
...
+      {onSettingsPress ? (
+        <Button
+          testID="trading-view-native-indicators-settings-button"
+          icon="SettingsOutline"
+          justifyContent="flex-start"
+          variant="tertiary"
+          onPress={async () => {
+            commitSelection();
+            await dialog.close();
+            onSettingsPress();
```
sidmorizon: 🟡 P2: Pending indicator selections can disappear

Severity: `non-severe`

On iOS and Android, `commitSelection` starts the cross-runtime atom write but opens settings without awaiting it. A quick settings save can overwrite those pending selections from the stale main-runtime copy.

limichange: Fixed in 457d24a287. The native container now returns the atom setter's Promise, and `commitNativeIndicatorSelection` / `commitSelection` propagate it. Both Settings and Confirm await persistence before closing; Settings also waits for the dialog to close before navigating.

This keeps the native main/background synchronization in the existing atom implementation. Added regressions for a delayed selection commit and Promise propagation in both Market and Swap. All 4 targeted suites (47 tests) and `yarn agent:check --profile commit` passed, including lint and TypeScript.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -290,6 +296,21 @@ export function IndicatorListDialogContent({
...
   return (
     <YStack gap="$6" pb="$2">
+      {onSettingsPress ? (
+        <Button
+          testID="trading-view-native-indicators-settings-button"
+          icon="SettingsOutline"
+          justifyContent="flex-start"
+          variant="tertiary"
+          onPress={async () => {
+            await commitSelection();
```
sidmorizon: # 🟡 P2 Awaited persistence has no rejection path, so a failed write strands the dialog

## Problem

This commit turns a fire-and-forget write into a UI gate, but neither call site handles rejection:

- `handleConfirmPress` (`NativeIndicatorSelector.tsx:258-261`): `await commitSelection(); await dialog.close();`
- Settings `onPress` (`NativeIndicatorSelector.tsx:305-309`): `await commitSelection(); await dialog.close(); onSettingsPress();`

`commitSelection` now returns the persistence promise all the way from `TradingViewNativeContainer.tsx:632` (`return setIndicatorSettings(...)`). On iOS/Android/extension that resolves through `wrapAtomPro.ts` `doSet` → `proxyStateUpdateActionFromUiToBg` → `backgroundApi.setAtomValue` → `callBackground` (`jotaiBgSync.ts:137-148`), a cross-runtime RPC that can reject: transport/timeout failures, and on the background side `broadcastStateUpdateFromBgToUi` throws `OneKeyLocalError('native background thread jotai bridge is not ready')` (`jotaiBgSync.ts:277-281`).

Nothing attaches a `catch`. These buttons don't set `onPressLoadingEnabled`, so `Button` returns the raw handler (`primitives/Button/index.tsx:289-290`) and `useSharedPress` only debounces it (`useEvent.ts:19-38`) — the returned promise is dropped by the press dispatcher.

## Impact

If the atom write fails, `await` never proceeds: `dialog.close()` and `onSettingsPress()` are skipped. Confirm becomes a dead control that does nothing visible, and Settings never navigates — the user is left in an open dialog with no error, their selection lost, recoverable only by dismissing manually. The escaping rejection also surfaces as an unhandled promise rejection (RN warning / Sentry noise). Before this commit the write was `void`-ed, so a failure was contained and the dialog always closed; the regression is introduced by making the close conditional on the write.

## Recommendation

Keep the ordering guarantee but make failure non-stranding, in both `NativeIndicatorSelector.tsx:258-261` and `305-309`:

```ts
try {
  await commitSelection();
} catch (error) {
  defaultLogger.market.tradingView.indicatorSettingsPersistFailed({ error });
} finally {
  await dialog.close();
}
onSettingsPress();
```

so the dialog always closes and the error is reported rather than silently dead-ending the tap. Surfacing a toast on failure would be better still, since the user's selection genuinely did not persist.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/indicatorUtils.ts
```diff
@@ -150,7 +150,7 @@ export function commitNativeIndicatorSelection({
     return;
   }
   if (onSelectionConfirm) {
-    onSelectionConfirm({
+    return onSelectionConfirm({
```
sidmorizon, line 169: # 🟠 P1 Return-type widening makes three existing call sites floating promises — `lint (24.x)` is red

## Problem

Returning the confirm result here (`indicatorUtils.ts:153`) together with widening the callback to `void | Promise<void>` (`:140`) changes `commitNativeIndicatorSelection`'s inferred return type from `void` to `void | Promise<void>`.

`indicatorUtils.test.ts` is **not** touched by this PR, and its three call sites discard the result:

- `indicatorUtils.test.ts:72` — "commits the full selection atomically when supported"
- `indicatorUtils.test.ts:95` — "keeps sub-indicators out of a main-only selection replacement"
- `indicatorUtils.test.ts:118` — "keeps per-indicator updates for legacy callers"

Each is a bare `commitNativeIndicatorSelection({ ... });`, so they became floating promises the moment the return type widened. That is exactly the CI failure on this head: `lint (24.x)` reports `Promises must be awaited, add void operator to ignore.` three times and `Found 0 warnings and 3 errors` ([run 34678030100](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34678030100/job/103511311151)). The count and rule match these three sites precisely; the sites in `TradingViewNativeContainer.test.tsx` are not flagged because their local prop casts still declare `=> void`.

## Impact

The required lint check fails on `457d24a2`, blocking merge, and the repo's own rule in `CLAUDE.md` ("Do not commit code that fails lint or TypeScript checks") is violated. This is a real consequence of the type change, not a test-only artifact: any other caller that ignores the result now silently drops a rejection-capable promise, which is the same class of bug the awaited-persistence change set out to fix.

## Recommendation

Update the three call sites in `indicatorUtils.test.ts` to match the new contract. Prefer awaiting, since these tests assert post-commit state:

```ts
await commitNativeIndicatorSelection({ ... });
```

(making each `it` callback `async`). If a given case deliberately ignores persistence, `void commitNativeIndicatorSelection({ ... });` satisfies the rule, but awaiting is the better fit here and also guards the new promise-propagation contract. Also consider giving `commitNativeIndicatorSelection` an explicit `: void | Promise<void>` return annotation so future widenings are visible at the signature rather than inferred.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

limichange: Fixed in 3c8180508c. All three `indicatorUtils.test.ts` callers now await `commitNativeIndicatorSelection` before asserting results, and the helper explicitly declares `void | Promise<void>` as its return type. I checked all callers of the helper.

Validation: `CI=true yarn lint` passed for the full repository, including TypeScript, Oxlint, formatting, module IDs, and Electron checks. The commit gate and all 4 targeted suites (49 tests) also passed. The previous staged-only check missed these unchanged dependent tests; this full run covers them.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
limichange: Fixed in 3c8180508c. Confirm and Settings now share a submission handler that catches rejection, shows the existing localized error Toast, and restores interaction in `finally`. It keeps the dialog and pending selection available for retry after a failed write. On success, it still awaits persistence and dialog close before navigating.

The shared in-progress guard also prevents duplicate submissions, reset, or selection edits during the write. Added regressions for both entry points: reject persistence, verify feedback and no close/navigation, then retry successfully with the same MA/RSI selection. All 49 targeted tests and full CI lint passed.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -241,18 +247,42 @@ export function IndicatorListDialogContent({
...
+
+  const handleSelectionSubmit = useCallback(
+    async (onSuccess?: () => void) => {
+      if (submittingRef.current) {
+        return;
+      }
+      submittingRef.current = true;
+      setIsSubmitting(true);
+      try {
+        await commitSelection();
```
sidmorizon, line 290: # 🟠 P1 After a failed write, reverting the selection silently skips the corrective persist

## Problem

Keeping the dialog open for retry is right, but the delta is always computed against the **mount-time** set, and the persist is optimistic.

- `originalActiveIndicatorValuesRef` is initialised once (`NativeIndicatorSelector.tsx:216`) and never advanced.
- `commitSelection` always diffs `activeIndicatorValuesRef.current` against it (`:250-260`).
- The write applies in memory *before* it can reject: `jotaiStorage.ts:474-475` does `set(baseAtom, newValue); return storage.setItem(key, newValue);`.

So on a rejection the in-memory atom already holds the failed value while storage still holds the old one. If the user responds to the error Toast by restoring the original selection and pressing Confirm again, `getNativeIndicatorSelectionUpdates` yields an empty delta and `commitNativeIndicatorSelection` returns at `indicatorUtils.ts:149-151` without ever calling `onSelectionConfirm`.

Scope: this optimistic-apply path is the single-runtime one (desktop/web, and native `main` when `enableNativeBackgroundThread` is off). On extension UI and native main-with-bg-thread, `wrapAtomPro.ts:45-49` proxies to bg and returns before applying locally, so those targets don't show the divergence.

## Impact

The second Confirm reports success — dialog closes, and from the Settings button it navigates — while no write is issued at all. The chart keeps rendering the failed, unpersisted selection the user just tried to undo, and in-memory state diverges from the persisted store until the app restarts and the old value reloads. Preconditions are narrow (a rejected write plus a revert), but the failure is completely silent, and this is precisely the retry path the commit was written to support.

## Recommendation

Stop treating "no delta vs. mount time" as "nothing to do" once a submission has failed. Either:

- track a `hasFailedSubmitRef` and, when set, bypass the empty-delta short-circuit so the full `activeIndicatorValues` payload is re-sent even when it equals the original; or
- have `handleSelectionSubmit` diff against the last **successfully committed** set (advance a `lastCommittedRef` only in the success path) rather than the mount-time original.

The second is the more robust shape, since it also keeps repeat submissions correct after a successful commit without remounting the dialog.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -241,18 +247,42 @@ export function IndicatorListDialogContent({
...
+        await dialog.close();
+        onSuccess?.();
+      } catch (_error) {
+        Toast.error({
+          title: intl.formatMessage({
+            id: ETranslations.global_an_error_occurred,
+          }),
+        });
```
sidmorizon, line 308: # 🟡 P2 The catch covers close/navigation too, and discards the error entirely

## Problem

Two issues in the same `try`/`catch` (`NativeIndicatorSelector.tsx:269-278`):

1. **Scope is too wide.** The `try` spans `await dialog.close()` (`:271`) and `onSuccess?.()` (`:272`) in addition to the persist. `onSuccess` is `onSettingsPress`, which pushes a modal (`TradingViewNativeContainer.tsx:644-649`). If the close or the navigation throws *after* the write already succeeded, the user is shown "An error occurred" for a save that in fact persisted.
2. **The error is swallowed.** `catch (_error)` shows the Toast and drops the error — no `defaultLogger`, no Sentry breadcrumb, nothing.

## Impact

Misattributed error copy on a successful save, and zero observability for exactly the failure mode this commit exists to handle. A field report of "my indicator settings don't stick" would have no diagnostic trail — no error class, no message, no indication whether the RPC rejected or storage failed.

## Recommendation

Narrow the `try` to the persist and log before reporting:

```ts
try {
  await commitSelection();
} catch (error) {
  defaultLogger.market.tradingView.indicatorSettingsPersistFailed({ error });
  Toast.error({
    title: intl.formatMessage({ id: ETranslations.global_an_error_occurred }),
  });
  return; // finally still clears the guard
}
await dialog.close();
onSuccess?.();
```

That keeps the retry behaviour identical for a failed write while letting a close/navigation fault surface as itself instead of as a bogus save error, and gives the failure a trace.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -282,14 +313,29 @@ export function IndicatorListDialogContent({
       testID="trading-view-native-indicators-confirm-button"
       variant="primary"
       size="large"
-      onPress={handleConfirmPress}
+      disabled={isSubmitting}
+      onPressLoadingEnabled
```
sidmorizon: # 🟡 P2 `onPressLoadingEnabled` is inert here — no spinner can ever show

## Problem

Both new `onPressLoadingEnabled` props (`NativeIndicatorSelector.tsx:317` on Confirm, `:333` on Settings) cannot work, because `Button` debounces the handler *before* the loading wrapper sees it:

- `useSharedPress` returns `debounce(onPress, 0)` with lodash's defaults `{ leading: false, trailing: true }` (`Button/useEvent.ts:19-25`, returned at `:37`).
- `handlePressWithLoading` then does `await onPress(event)` on that debounced wrapper (`Button/index.tsx:288-303`).

A trailing-edge lodash debounce returns its cached `result` synchronously — `undefined` on the first press, and the **previous** press's return value on later ones — never the promise from `handleSelectionSubmit`. The `// eslint-disable-next-line @typescript-eslint/await-thenable` sitting on that `await` (`Button/index.tsx:292`) is the compiler already saying the awaited value isn't a thenable.

So `setInternalLoading(true)` and `(false)` batch within one tick into a no-op, and on a later press the wrapper can await a stale promise from the previous submission.

## Impact

The prop reads as the mechanism that shows progress during the write, but the only thing actually gating the UI is `disabled={isSubmitting}`. A future reader will reasonably assume the spinner works and won't add real progress feedback; the stale-promise path also makes `internalLoading` non-deterministic across repeated presses.

## Recommendation

Drop `onPressLoadingEnabled` from both buttons and drive the spinner from the state this component already owns:

```tsx
loading={isSubmitting}
```

`isLoading = loading || internalLoading` (`Button/index.tsx:243`) feeds both the spinner and the disabled state, so `loading={isSubmitting}` also makes the separate `disabled={isSubmitting}` redundant.

[Claude session ↗](https://claude.ai/code/session_01NzVhWHhnA4dX9z6xUq6wbB)

---

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
limichange: Fixed in 9a2c5a3029. An unacknowledged submission now retains its main/sub replacement scopes. `commitNativeIndicatorSelection` merges these with subsequent edits, so reverting to the original values still sends a corrective write without replacing an untouched scope. The retained scopes are cleared only after a successful write, and the comparison baseline advances before closing the dialog.

Regression tests cover both MA and RSI failing then being reverted, as well as successful persistence followed by a failed close/navigation and retry without a duplicate write. All 53 targeted tests, full `CI=true yarn lint`, and the commit gate passed.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
limichange: Addressed in 9a2c5a3029. Errors now go through the existing `defaultLogger.app.error.log` with an explicit `persist`, `close`, or `navigate` phase and name/className/message/code. Error payloads and indicator settings are excluded from the log.

The successful-write baseline is advanced immediately after persistence, so a later close/navigation failure cannot cause an unnecessary retry of the save. I kept all three phases handled by the shared catch to avoid introducing an unhandled rejection at the press boundary; the existing generic error Toast does not claim that saving failed. Added separate close/navigation failure regressions verifying the logged phase and no repeated write. All 53 targeted tests and full CI lint passed.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
limichange: Fixed in 9a2c5a3029. Removed both `onPressLoadingEnabled` props. The current submit action now directly drives each button's `loading` prop, so only the pressed button shows progress and the state does not depend on the debounced handler's return value. The other actions remain disabled during submission.

Both Confirm and Settings regression cases verify loading during a pending write and restoration after rejection, followed by a successful retry. All 53 targeted tests and full CI lint passed.

packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorSelector.tsx
```diff
@@ -241,18 +257,62 @@ export function IndicatorListDialogContent({
...
+      submittingRef.current = true;
+      setSubmittingAction(action);
+      let phase = 'persist';
+      try {
+        await commitSelection();
+        phase = 'close';
+        await dialog.close();
+        if (action === 'settings') {
+          phase = 'navigate';
+          onSettingsPress?.();
```
sidmorizon, line 295: 🟡 P2: Navigation failure closes settings entry

Severity: `non-severe`

`handleSelectionSubmit` closes the selector before `onSettingsPress` runs. A navigation failure leaves no open control to retry Settings.

limichange: The selector sheet closes before navigation, but the persistent chart entry remains available. `TradingViewChartControls.tsx:338-345` renders the Indicators button with `onShowIndicatorsDialog`, and `TradingViewNativeChartControlsContainer.tsx:143-163,225` opens a fresh selector each time. Closing this dialog does not unmount those chart controls.

If navigation throws, the error Toast and `navigate` log are emitted, and the user can retry via Indicators → Settings. The selection has already been persisted, so reopening retains it. Keeping the close-before-navigation order avoids presenting the new modal while the previous sheet is still active. I am treating automatic reopening as an optional UX enhancement, rather than a missing recovery path, and leaving the code unchanged for this observation.

originalix (approved):

PR #12498, opened 2026-07-15
refactor: migrate extension builds to rspack

huhuanming:
## Summary

- switch extension development, production, release, and performance builds from Webpack to Rspack without a fallback kill switch
- consolidate five independent compilers into three runtime domains:
  - Pages: `ui-popup`, `ui-passkey`, and `offscreen` share one module graph and emitted chunks
  - Background: stable single-file service worker
  - Content Script: stable single-file bundle
- add collision-safe staging/finalization, byte-identical asset deduplication, cross-platform zip generation, stats output, and production artifact quality gates
- remove the extension-specific Webpack configuration and helpers

## Why

The Webpack release build was slow and emitted multiple semantically duplicated entry graphs. Besides build cost, this inflated both the unpacked extension and the release zip, produced thousands of JavaScript files, and left remote-code/import-meta checks implicit.

The new topology shares code where browser page runtimes can safely reuse cached chunks while preserving the isolated Background and Content Script runtime boundaries required by Manifest V3.

## Artifact comparison

Webpack baselines:

- 6.4.0 stable release: [release-ext run 27440620691](https://github.com/OneKeyHQ/app-monorepo/actions/runs/27440620691), artifact `onekey-extension-rn-6.4.0-2026061232-14069831-1affb7b37ff168265c6353b51b85444495b9e7c9`
- 6.6.0 current `x`: [release-ext run 29363675791](https://github.com/OneKeyHQ/app-monorepo/actions/runs/29363675791), artifact `onekey-extension-rn-6.6.0-2026071428-16833314-2684bf7d0bd334912a3e72aabf08801f10e9c205`

All rows use the production `OneKey-Wallet-<version>-chrome_v3-extension.zip` and sum the uncompressed zip entry sizes.

| Metric | Webpack 6.4.0 | Webpack 6.6.0 | Rspack 6.6.0 | Rspack vs 6.4 | Rspack vs 6.6 |
| --- | ---: | ---: | ---: | ---: | ---: |
| Release zip | 70.61 MiB | 103.53 MiB | 50.78 MiB | -28.1% | -50.9% |
| Unpacked files | 985 | 3,036 | 831 | -15.6% | -72.6% |
| Unpacked bytes | 223.60 MiB | 320.86 MiB | 146.31 MiB | -34.6% | -54.4% |
| JavaScript files | 721 | 2,767 | 683 | -5.3% | -75.3% |
[...]

PR #13356, opened 2026-09-12
fix: restore EXT_CHANNEL in extension builds (OK-62971)

huhuanming:
[OK-62971](<https://onekeyhq.atlassian.net/browse/OK-62971>)

----------

## Problem

The extension never detects a pushed higher version. With the Chrome Extension
target enabled in the release dashboard, `/utility/v1/app-update` still answers
the extension with an empty payload.

## Root cause

The extension sends `X-Onekey-Request-Platform: extension` instead of
`extension-chrome`. The update service has no `EPlatform` for `extension`, so
neither the full-update lookup nor the current-version lookup matches a row.

#12498 moved the extension build from webpack to rspack and dropped the
write-back that used to live in `development/webpack/ext/devUtils.js`:

```js
process.env.EXT_CHANNEL = buildTargetBrowser; // removed in #12498
```

`development/rspack/constant.ts` resolved `targetBrowser` but never assigned it
back, and nothing else sets `EXT_CHANNEL` — not `apps/ext/package.json`, not
`release-ext.yml`, not `.env`. So every extension build since 2026-07-15 folds
`isExtChrome` / `isExtFirefox` / `isExtEdge` to `false`:

1. `buildTimeEnv.js` — `isExtChrome = process.env.EXT_CHANNEL === 'chrome'` → `false`
2. `platformEnv.ts` — `getAppChannel()` falls through → `undefined`
3. `InterceptorConsts.ts` — `['extension', undefined].filter(Boolean).join('-')` → `'extension'`

### Neither channel resolver could work

`getBuildTargetBrowser()` looked like it supported `--firefox` / `--chrome` /
`--edge`. In the rspack copy those branches were unreachable:

```
$ rspack build --firefox --config-loader jiti
[rspack] CLI failed:
CACError: Unknown option `--firefox`
```

The CLI rejects undeclared flags before the config is ever loaded. The webpack
copy was worse — it ended in an `else` that overwrote the result with `'chrome'`
unconditionally, so `EXT_CHANNEL` was never read there at all. Both only ever
inspected `process.argv[process.argv.length - 1]`, so even a tolerated flag had
to be the very last argument.

That left `EXT_CHANNEL` as the only working selector — and nothing set it. Every
build silently fell through to the `'chrome'` literal default.

## Blast radius

The same value feeds analytics (`platform` on every extension event since
2026-07-15) and the WalletConnect client name. A firefox build would have picked
the Chrome manifest, since `apps/ext/src/manifest/index.js` tests
`process.env.EXT_CHANNEL === 'firefox'` too.

## Fix

`constant.ts` owns the mirroring through `applyExtChannelToEnv()`, next to the
`targetBrowser` resolution it belongs to. The ext config calls it; DefinePlugin,
`loadBuildTimeEnv` and the manifest builder all read the env var lazily
afterwards.

It is deliberately **not** assigned unconditionally at module scope.
`EXT_CHANNEL` is not namespaced by platform, so a global assignment leaks into
the web/desktop bundles: `getAppChannel()` tests the ext channels before the
desktop ones, and the WalletConnect client url is built from the same variable
(`https://desktop-web.app.onekey.so` would become `desktop-chrome-web`).

The dead argv branches are removed from both resolvers, collapsing each to
`EXT_CHANNEL || 'chrome'`. Since a selector that looks supported but always
resolves to the default is exactly what hid this for two months,
`applyExtChannelToEnv()` now rejects a channel outside `chrome|firefox|edge`
rather than quietly shipping a Chrome manifest under a bogus platform header.

`EXT_CHANNELS` carries a checklist of everything a new channel has to touch:
`buildTimeEnv.js`, `platformEnv.ts` (`IAppChannel`, the destructure,
`getAppChannel()`, the exported object), `platformEnvDefine.js`, the manifest
variants, and the server-side `EPlatform` / `fullUpdatePlatformMapping` /
`DefaultPlatformInfo` wiring whose absence produces exactly this bug.

Writing that checklist turned up two spots `edge` is already missing from —
`platformEnv`'s exported object (so `platformEnv.isExtEdge` is `undefined`,
though the channel still resolves via `getAppChannel()`) and
`platformEnvDefine`'s fold map (so edge branches ship unfolded). Both are noted
in the comment and left alone here.

## Verification

Both builds installed unpacked in Chrome for Testing 145, with
`/utility/v1/app-update` forwarded to the test environment (the extension's own
request headers preserved):

| build | `X-Onekey-Request-Platform` | test-env response |
| --- | --- | --- |
| `x` | `extension` (32 requests) | `{"code":0,"message":"Success","data":{}}` |
| this PR | `extension-chrome` (32 requests) | `version: "9906.17.2"` + changelog |

`background.bundle.js` goes from `isExtChrome:!1` to `isExtChrome:!0`.

Channel resolution after the cleanup:

| `EXT_CHANNEL` | result |
| --- | --- |
| unset | `chrome`, written back to `process.env` |
| `firefox` / `edge` | that channel, written back |
| `Firefox` (wrong case) | throws `Invalid EXT_CHANNEL "Firefox", expected one of chrome, firefox, edge` |
| `safari` | throws |

The web kill-switch webpack config (`apps/web/webpack.config.kill-switch.js`,
the only remaining consumer of `development/webpack/constant.js`) still loads,
and still resolves `TARGET_BROWSER` to `'chrome'` in a web build — unchanged,
since the old code always returned `'chrome'` and the web build never sets
`EXT_CHANNEL`.

Cross-checked directly against the test service: `extension-chrome` returns
9906.17.2 and `extension-edge` returns 9906.17.0, so the dashboard config and
the server were already correct — only the header the client sent was wrong.

`yarn agent:check --profile pr` passes.

## Follow-up (not in this PR)

`development/envExposedToClient.js`: of the 38 variables exposed to the client,
11 have no source anywhere — not `explicitDefines`, not CI, not a package
script, not `.env.example`. `EXT_CHANNEL` was one. The other 10 are inert today
(zero source references, or the value is hardcoded in a constant next to them),
so they cause no bug, but they are the same shape: listed as configurable, never
actually set.

```
EXT_CHANNEL                ← this PR
ONEKEY_BUILD_TYPE          ← 0 source references
WALLETCONNECT_PROJECT_ID   ← project id is hardcoded in walletConnect/constant.ts
SUPABASE_PROJECT_URL       ← same-named constant lives in authConsts.ts
SUPABASE_PUBLIC_API_KEY    ← same
PERF_BG_CALL_MARK_MS, PERF_JS_BLOCK_THRESHOLD_MS, PERF_STORAGE_MARK,
PERF_STORAGE_MARK_MS, PERF_SIMPLEDB_MARK, PERF_SIMPLEDB_MARK_MS  ← 0 references
```

[OK-62971]: https://onekeyhq.atlassian.net/browse/OK-62971?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-b09a2bf9-e519-509f-8d31-466652b49a89)

sidmorizon:
Claude review session: https://claude.ai/code/session_01CDqm85z8DH5hL1REjXLDhL

development/rspack/rspack.ext.config.ts
```diff
@@ -23,6 +24,11 @@ import type {
   RspackPluginInstance,
 } from '@rspack/core';
 
+// Extension builds resolve their channel here (chrome unless EXT_CHANNEL says
+// otherwise) and publish it to process.env before anything reads it. Scoped to
+// this ext-only config on purpose — see applyExtChannelToEnv.
+applyExtChannelToEnv();
```
sidmorizon: # 🟡 [P2] `applyExtChannelToEnv()` at module scope is still reachable from the `development/rspack` barrel

## Problem

The refactor moved the assignment into a named function but kept the **invocation** at module scope. `applyExtChannelToEnv()` on line 30 sits in the module body, between the import block and `const platform = 'ext'` (line 32), so it fires on *import* of `rspack.ext.config.ts`, not on *use* of `createExtConfig()`.

`development/rspack/index.ts:9` does `export * from './rspack.ext.config';`. `export *` forces full evaluation of the re-exported module, so importing *anything* from the `development/rspack` barrel — `createWebConfig`, `isDev`, `getOutputFolder` — sets `process.env.EXT_CHANNEL = 'chrome'` process-wide, on any target.

That is exactly the leak the new JSDoc warns against at `development/rspack/constant.ts:58-61` ("Call this ONLY from the extension config"). The call site honours it, but the barrel quietly re-exports the module that makes the call, so the invariant is documented rather than structural.

To be clear about severity: this is **latent, not live**. I grepped for barrel importers at this head and found none — all four app configs (`apps/ext`, `apps/web`, `apps/desktop`, `apps/web-embed`) import their platform config module directly, and the only references to `applyExtChannelToEnv` anywhere are `constant.ts:63` and lines 9/30 here. The fix itself works: I traced `apps/ext/rspack.config.ts` → line 30 → `buildDefineMap()` (`rspack.base.config.ts:177-183`), `loadBuildTimeEnv()` (`:79-89`), and the `manifest/index.js` require inside `createCopyPlugin`'s `transform()`, and every consumer reads the env var after this line runs.

## Impact

Nothing breaks today. But the barrel exists to be imported, and the failure mode is silent: `buildTimeEnv.js:15` computes `isExtChrome` from `EXT_CHANNEL` alone with no `isExtension` guard, and `getAppChannel()` (`packages/shared/src/platformEnv.ts:271-274`) tests the ext channels before the desktop ones. A web or desktop build that ever picked up the barrel would ship `appChannel: 'chrome'`, a wrong `X-Onekey-Request-Platform` header and a wrong analytics `platform`, with no error and no test covering it — the same class of silent, header-shaped bug this PR is fixing.

## Recommendation

Move the single call from line 30 into `createExtConfig()` (`development/rspack/rspack.ext.config.ts:458`) as its first statement, before the `isManifestV3` check and `createPagesConfig(basePath)`:

```ts
export function createExtConfig({
  basePath,
}: IExtConfigOptions): RspackOptions[] {
  applyExtChannelToEnv();

  if (!isManifestV3) {
```

That is still strictly earlier than every consumer identified above (all of them run inside `createPagesConfig` / `createBackgroundConfig` / `createContentScriptConfig`, or later at asset-emit time), so the fix keeps working identically — and the side effect becomes unreachable except from an actual extension build, making the "ext-only" guarantee structural instead of conventional.

[Claude session ↗](https://claude.ai/code/session_01CDqm85z8DH5hL1REjXLDhL)

---

huhuanming: Good catch — fixed in 56cc4a33b6. Verified both premises before acting: `development/rspack/index.ts:9` does re-export this module with `export *`, and there are no barrel importers today, so this was latent as you said.

Moved the call to the first statement of `createExtConfig()` exactly as recommended. The ext-only guarantee is now structural rather than conventional: `applyExtChannelToEnv` has a single call site in the repo and it sits inside the factory, so module evaluation — which is what `export *` forces — no longer has the side effect.

Re-verified the fix still works after the move, since the whole question was ordering:

- `yarn agent:check --profile commit` — all pass
- ext production build — `background.bundle.js` still folds to `isExtChrome:!0`
- unpacked install in Chrome for Testing 145 with `/utility/v1/app-update` forwarded to the test environment — still sends `extension-chrome` and still receives `version: 9906.17.2`

One thing I could not do, for the record: I wanted a runtime before/after showing that importing the barrel leaves `EXT_CHANNEL` unset. Loading `development/rspack/index.ts` standalone through jiti fails with `Cannot use 'import.meta' outside a module` — the rspack CLI configures its own jiti differently — so that assertion rests on the static fact (one call site, inside the function) plus the build verification above, not on a runtime comparison.

originalix (approved):

PR #13357, opened 2026-09-12
fix: keep large account indices and page numbers readable in Manage Accounts(OK-62844)

weatherstar:
[OK-62844](<https://onekeyhq.atlassian.net/browse/OK-62844>)

----------

## Summary
- Give the Manage Accounts (batch-create preview) index column a digit-aware minimum width so 10-digit account indices stay on one line.
- Unpin the `ButtonGroup.Item` fixed width and pad the pager's page-number item so 9-digit page numbers no longer overflow into the chevron buttons.

## Intent & Context
QA reported (OK-62844) that in Manage Accounts on desktop, with very large account indices (e.g. 2147483641, near the int32 ceiling), the "No." column wrapped the index onto two lines. A follow-up screenshot of the same page after the first fix showed the second defect on the same screen: the bottom pager rendered `‹214748365›` with both chevrons pressed against the digits.

## Root Cause
- Index column: the column only received a 20% flex share of the row, and the checkbox label size override (`labelProps.size`) is silently ignored because `Label` sizes text only through `variant`, so the label really renders at 16px Medium and 10 digits do not fit.
- Pager: `ButtonGroup.Item` wraps Tamagui `ToggleGroup.Item`, whose default `size="$true"` variant pins `width` to the 44px size token. The item only overrode `height` and `minWidth`, so the page-number item stayed 44px wide while a 9-digit page number measures ~80px (verified over CDP on the running desktop app: item width 44px, text child 80px, overlapping the chevron items on both sides).

## Design Decisions
- The index column floor is computed from page arithmetic (`fromInt + page * pageSize`) rather than from loaded rows, so the header and skeleton rows already use the final width and nothing shifts when data arrives. Per-digit width uses the real 16px Medium metrics rounded up (11px) to stay safe on fallback fonts.
- The width unpin lives in `ButtonGroupItem` itself (this component's only consumer is this pager) since a fixed 44px width is wrong for any variable-width content; the page-number item adds `px="$3"` so small page numbers keep the previous 42–44px footprint via `minWidth`.

## Changes Detail
- `packages/kit/src/views/AccountManagerStacks/pages/BatchCreateAccount/BatchCreateAccountPreview.tsx`: `indexColumnMinWidth` floor for the checkbox column, drop the dead `labelProps.size` override, add horizontal padding to the page-number item.
- `packages/components/src/primitives/ButtonGroup/index.tsx`: `width="auto"` on `ToggleGroup.Item` to release the size-token width.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Desktop / Web / Extension / Mobile (same component on all; verified on desktop dev)
- **Risk Areas**: `ButtonGroup` width change; the pager is its only usage. Page-1 layout on desktop is unchanged.

## Test plan
- [x] Desktop dev: Manage Accounts with start index 2147483641 shows each 10-digit index on one line; header/skeleton widths match loaded rows.
- [x] Desktop dev: pager shows `‹ 214748365 ›` with normal spacing (page-number item 104px, chevrons 42px each, measured over CDP).
- [ ] Small page numbers (1–99) keep the previous pager footprint.
- [ ] Mobile/device pass on Manage Accounts preview.

[OK-62844]: https://onekeyhq.atlassian.net/browse/OK-62844?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-75a660bc-da63-5e38-8e21-d18b13e13e90)

sidmorizon:
Claude review session: https://claude.ai/code/session_018PsZdcdG5DYLfu8MVDCPJs

packages/kit/src/views/AccountManagerStacks/pages/BatchCreateAccount/BatchCreateAccountPreview.tsx
```diff
@@ -1081,6 +1100,7 @@ function BatchCreateAccountPreviewPage({
               <ButtonGroup.Item
                 testID="batch-create-account-preview-page-number"
                 opacity={1}
+                px="$3"
```
sidmorizon, line 1117: 🟡 P2: Large pages overflow mobile footer

Severity: `non-severe`

On narrow screens, `px` helps expand a nine-digit pager to about 188px. Adjacent selection and settings controls then overflow the footer row.

packages/kit/src/views/AccountManagerStacks/pages/BatchCreateAccount/BatchCreateAccountPreview.tsx
```diff
@@ -655,6 +675,7 @@ function BatchCreateAccountPreviewPage({
         columnProps: {
           flexGrow: 2,
           flexBasis: 0,
+          minWidth: indexColumnMinWidth,
```
sidmorizon, line 688: 🟡 P2: Large indices crush mobile details

Severity: `non-severe`

On narrow screens, `minWidth` reserves 138px for ten-digit indices. Native gaps then leave address and balance columns too narrow to read.

originalix:

Review of commit `8629e3adbae2` finished, but some comments could not be published automatically.

packages/kit/src/views/AccountManagerStacks/pages/BatchCreateAccount/BatchCreateAccountPreview.tsx
weatherstar: Fixed in 0d5d7884b6: the select-all block now has `flexShrink={1} minWidth={0}` (plus `labelContainerProps={{ flexShrink: 1 }}` on the Checkbox), so on a 375pt screen a 9-digit pager makes the "Select all" label wrap instead of pushing the settings and pager controls past the footer edge. The page number itself stays fully visible. (Actual pager width for 9 digits is ~115px with `px="$3"`, not 188px, but the row did overflow by ~14px at 375pt.)

packages/kit/src/views/AccountManagerStacks/pages/BatchCreateAccount/BatchCreateAccountPreview.tsx
weatherstar: Fixed in 0d5d7884b6: on native the row `gap` now drops from `$8` (32px) to the web value `$4` (16px) once the index column floor reaches 8 digits. At 375pt with a 10-digit index that leaves ~103px for the address column and ~62px for balance instead of ~83px / ~50px, while the index stays on one line (the original OK-62844 report). Header and body rows share `rowProps`, so they remain aligned.

originalix (approved):

PR #13253, opened 2026-09-08
fix: keep lock reset dialog interactive(OK-62416)

huhuanming:
[OK-62416](https://onekeyhq.atlassian.net/browse/OK-62416)

----------

## Summary

- Keep the lock-screen reset dialog inside the app-state lock portal on non-native platforms.
- Preserve the existing native over-top rendering behavior.
- Add focused regression coverage for desktop and native dialog options.

## Intent & Context

Fixes [OK-62416](https://onekeyhq.atlassian.net/browse/OK-62416). On Electron, choosing **Forgot password?** displayed the reset confirmation dialog, but its input and actions could not be clicked.

## Root Cause

`isOverTopAllViews` re-portaled the dialog to `document.body`. The lock screen's `MutationObserver` marks newly added body children outside the lock container as `inert`, so the reset dialog remained visible while pointer and focus events were blocked.

## Design Decisions

Keep the existing lock-screen inert isolation intact and only avoid the body-level portal on non-native platforms. Native keeps its previous over-top behavior.

## Changes Detail

- Set `isOverTopAllViews` from `platformEnv.isNative` for the reset dialog.
- Add focused hook tests covering desktop and native behavior.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Desktop, Web, Extension
- **Risk Areas**: Dialog stacking on locked non-native surfaces; native behavior is explicitly preserved.

## Test Plan

- [x] `yarn jest packages/kit/src/views/Setting/hooks/index.test.tsx --runInBand`
- [x] Electron: open **Forgot password?** > **Reset**, focus and enter `RESET`, verify Confirm becomes enabled, then cancel the dialog.
- [x] `yarn agent:check --profile commit`

[OK-62416]: https://onekeyhq.atlassian.net/browse/OK-62416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13359, opened 2026-09-12
fix: make lock screen Forgot passcode usable on extension and iOS (OK-62416)

huhuanming:
[OK-62416](<https://onekeyhq.atlassian.net/browse/OK-62416>)

----------

## Summary

Three independent defects kept the lock screen's **Forgot passcode?** flow unusable after #13253, one per platform family. Each is fixed and verified on a real target.

- **Extension popup / side panel** — a sheet-form lock dialog portalled itself to `document.body` and was caught by the lock screen's `inert` sweep. Kept inside `APP_STATE_LOCK_CONTAINER_OVERLAY` via `portalProps.passThrough`, with the portal root sized to the viewport so the sheet lands on screen.
- **iOS** — the button never received the tap. A `tertiary` Button's touch target is a ~26pt band offset above its own label; `hitSlop={NATIVE_HIT_SLOP}` fixes it. The iOS portal container is also now hosted inside the lock screen's own window overlay.
- **All platforms** — the confirm button could not enable, because the footer subscribed once to a `Dialog.Form` that registers late and re-registers. Replaced with an explicit registration channel.

Collected the props every lock-screen dialog needs into one `inAppStateLockDialogProps` so the reset dialog, the biometric-changed warning and the log-upload dialog cannot drift apart again — scoped, so only the lock screen uses them.

## Intent & Context

Follow-up to #13253. QA came back with: extension popup and side panel still freeze on **Forgot passcode?**, and on iOS the button does nothing at all. Desktop, the extension's expanded tab and Android were reported fine.

## Root Cause

**Extension popup / side panel (any viewport below `$gtMd`).** The dialog renders as a Tamagui `Sheet`, and `DialogFrame` passes `modal: true` off native. A modal sheet portals itself to `document.body` through Tamagui's own `Portal` — straight into the sweep `AppStateLockContainer` runs to mark every other `document.body` child `inert`, so nothing can steal focus from the passcode input. The sheet drew normally and swallowed every tap and keystroke. #13253's `isOverTopAllViews` did not help, because the sheet escapes on its own rather than through that flag.

Reproduced in the **real extension**, driven over CDP:

```json
"bodyChildrenOfInterest":[
  {"i":7,"tag":"DIV","inert":false,"hasLock":true,"hasReset":false},
  {"i":8,"tag":"SPAN","inert":true, "hasLock":false,"hasReset":true}],
"lockAndResetShareOneBodyChild":false, "resetSubtreeInert":true
```

Playwright refused every click on it: `<div … AppStateLock.tsx">…</div> from <div>…</div> subtree intercepts pointer events`.

**iOS.** Two separate things, and only one of them was the dialog.

The button's handler never ran. `variant="tertiary"` renders the Button as its label plus 4pt of padding, and its `my: -5 / mx: -9` negative margins pull the frame above the glyphs, so the real touch target is a ~26pt band sitting slightly higher than the text — well under the 44pt Apple asks for, with no `hitSlop`. Taps on the visible label missed it.

Separately, `AppStateContainer/index.ios.tsx` was just `OverlayContainer` and never registered the portal container, so a lock-screen dialog fell through to the `isOverTopAllViews` branch of `renderToContainer.ios` and was mounted into the overlay added at app start — below the lock screen's own, which is added at lock time.

**Confirm button, all platforms.** `Dialog.Form` is lazy. It registers its react-hook-form instance *after* the footer's effect has run, and registers a fresh one while the open animation settles. `useConfirmButtonDisabled` subscribed once, so `disabledOn` watched a form nobody typed into and the confirm button stayed disabled no matter what the user entered.

## Design Decisions

- **`portalProps.passThrough` rather than `modal: false` on the sheet.** `DialogFrame` deliberately forces `modal: true` on web (OK-36893 / OK-38624: non-modal sheets collapse into `position: relative` and break z-index stacking). `passThrough` opts only the portal hop out and leaves the rest of the modal sheet behaviour intact. Tamagui reads `portalProps` only in its web modal branch, so it is inert on native.
- **Sizing the portal root with `position: absolute`, not `fixed`.** `document.body` is a flex container, so the portal root collapsed to a zero-width flex item beside `#root`; the sheet resolved its static position against that and rendered one viewport to the right, off screen. `absolute` keeps `z-index: auto` and creates no stacking context, so the lock screen's z-index keeps competing with the rest of `document.body` exactly as before.
- **The lock-screen props are opt-in per call site.** `APP_STATE_LOCK_CONTAINER_OVERLAY` exists only while the lock screen is mounted, so `PasswordVerify` takes them only when `PasswordVerifyContainer` says it *is* the lock screen (`name === 'lock'`). An unlocked biometric-changed warning would otherwise have had nowhere to render.
- **No `isOverTopAllViews` on the unlocked paths.** `dialogShow` reads that flag only when a `portalContainer` is also named, which the unlocked paths never do — so it has always been inert there, and spelling it out only claims stacking these dialogs do not get. The unlocked biometric warning does not need it either: the passcode prompt that raises it renders with no container on native, so the warning lands in the same portal as a later child and stacks above it.
- **An event for the form registration, not a poll.** `Dialog.Form` announces itself through `registerForm` on the dialog instance; `BaseDialogContainer` fans that out to subscribers; the footer moves its `watch` subscription when it fires. It also clears the registration on unmount, guarded on identity so a replacement that registered first is not wiped.

## Changes Detail

| file | change |
|---|---|
| `Setting/hooks/index.tsx` | `inAppStateLockStyle` → `inAppStateLockDialogProps`, now also carrying `portalContainer`, `isOverTopAllViews: false` and the sheet's `portalProps.passThrough` |
| `Password/components/PasswordVerify.tsx` | new `inAppStateLock` prop gates the lock-screen props |
| `Password/container/PasswordVerifyContainer.tsx` | passes `inAppStateLock={isLock}` |
| `Setting/pages/Tab/exportLogs/showExportLogsDialog.tsx` | uses the shared props |
| `AppStateContainer/index.tsx` | portal root fills the viewport |
| `AppStateContainer/index.ios.tsx` | hosts `Portal.Container` inside the lock screen's `FullWindowOverlay` |
| `AppStateLockContainer/components/AppStateLock.tsx` | `hitSlop={NATIVE_HIT_SLOP}` on the reset button |
| `Dialog/type.ts`, `Dialog/index.tsx` | `registerForm` / `subscribeFormChange` on the dialog instance |
| `Dialog/DialogForm.tsx` | announces its form, and clears the registration on unmount |
| `Dialog/Footer.tsx` | follows the announcement instead of subscribing once |
| `Dialog/index.tsx` | dev-only guard: reports a caller that combines `portalContainer` with `isOverTopAllViews: true` |

## Verification

Every claim below comes from unmodified product code — CDP probes of the real extension's DOM, the device's own native log file, and screenshots.

**Real extension, popup and side panel.** Built `apps/ext`, loaded it unpacked over CDP, onboarded with a **passcode** (PassKey off, `setBiologyAuthEnable(false)` before locking so no biometric prompt can auto-unlock the screen under test), locked via `lockApp()`, then `Forgot passcode?` → type `RESET` → `Cancel`.

| surface | body child | inert | typed | confirm `aria-disabled` | Cancel |
|---|---|---|---|---|---|
| popup `ui-popup.html` 375×600 | one child holds lock **and** dialog | `false` | `"RESET"` | `true` → **`false`** | dismissed |
| side panel `ui-side-panel.html` 400×700 | one child holds lock **and** dialog | `false` | `"RESET"` | `true` → **`false`** | dismissed |

**iOS, same-coordinate A/B.** `useResetApp` calls `serviceApp.isAppLocked()`, so a handler run leaves an unforgeable record in `Library/Caches/logs/app-latest.log`. Same device, same build shape, same tap at `(201, 815)` — the centre of the label:

| build | `isAppLocked` lines before → after |
|---|---|
| without `hitSlop` | 15 → **15** (handler never ran) |
| with `hitSlop` | 15 → **20** (handler ran, sheet opened) |

Tapping `(201, 808)` — 7pt higher — reaches the handler on *both* builds, which pins the cause to the target's size and offset. Also verified on iOS: the log-upload dialog (3 logo taps) renders and dismisses above the lock screen.

**Android.** Emulator, real build: dialog opens, accepts `RESET`, Confirm enables, Cancel dismisses.

**Tests — 27 across 7 suites, every new one mutation-checked:**

| test | mutation that breaks it |
|---|---|
| `leaves the warning in the default portal everywhere else` | unconditional `...inAppStateLockDialogProps` |
| `drops the replaced form’s subscription instead of leaking it` | drop `subscription?.unsubscribe()` in `watchCurrentForm` |
| `drops its subscriptions when the footer unmounts` | drop the footer effect's cleanup |
| `clears the registration when it unmounts` | drop `DialogForm`'s cleanup |

`AppStateContainer/index.test.tsx` renders the real web container and asserts the portal genuinely shares one `document.body` child with the lock screen — the property the inert sweep depends on.

## Risk Assessment

- **Risk Level**: Medium
- **Affected Platforms**: Extension, Web, Desktop, Mobile (iOS, Android)
- **Risk Areas**:
  - The web portal root is now a viewport-sized `absolute` box. It exists only while the lock screen is mounted, and the lock screen already covered the viewport, but it is the change most likely to have a layout side effect.
  - `Footer.tsx` is shared by every dialog in the app. The work added is one subscription moved on an event, torn down on unmount, and covered by two leak tests.
  - No change to global dialog stacking: `dialogShow` and `renderToContainer` are untouched.
  - Android behaviour is unchanged — `renderToContainer.android` ignores `isOverTopAllViews` and already rendered into this container.

## Known Limitations

- **The lock-screen biometric-changed warning was not driven end-to-end on iOS.** `checkBiometricAuthChanged` compares `LAContext.evaluatedPolicyDomainState` against a keychain baseline, and the simulator offers no reachable "changed but still supported" state: un-enrolling makes `isSupportBiologyAuth()` false so the check early-returns, and toggling enrollment leaves the domain state identical (the native module logs every outcome it can distinguish and stays silent after a toggle cycle). Its scoping is covered by a mutation-checked unit test on the real component, and its rendering path is the same `Dialog.confirm` with the same props into the same container as the reset and log-upload dialogs, both verified on that device.
- **`isOverTopAllViews` silently does nothing without a `portalContainer`.** Pre-existing on `x` and left alone here — `GlobalErrorHandlerContainer` has carried it inert for a long time. Making it work also needs a lock-state gate (iOS stacks window overlays by add order, and the lock screen's is added at lock time) and a toast-container re-front, which together are their own change with their own iOS verification.

  What this PR does add is a **tripwire, not a fix**: `Dialog.show` now reports in dev when a caller combines `portalContainer` with `isOverTopAllViews: true`. That combination is the only reachable route to a dialog creating a window overlay after the lock screen has added its own, and `renderToContainer.ios` does not survive it anyway — it mounts the element twice and returns only the second manager. Nothing in the repo does it today; the guard is there so the next caller that tries finds out immediately instead of debugging a stacking bug. It deliberately stays quiet for `isOverTopAllViews: false` alongside a container, which is exactly the shape `inAppStateLockDialogProps` uses.

## Environment Notices (`ONEKEY_USER_NOTICE`)

Neither affects the result, but both must be surfaced rather than described as clean runs:

1. **iOS dev-shell, first attempt — failed.** `[devVendor] Missing android cache…`, then `action=run-failed reason="[nativeDevShell] Metro exited with code 1"`, `status=failed userNoticeRequired=true`. The retry ran clean and is the run all iOS evidence comes from.
2. **Android dev-shell fell back to a local Gradle build** after a transient APK download failure: `action=local-build reason="terminated"`, `status=running shell.source=local-build userNoticeRequired=true`. Not a cache hit.

[OK-62416]: https://onekeyhq.atlassian.net/browse/OK-62416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-13691381-0977-59c8-8357-1cccf92323f9)

sidmorizon:
Claude review session: https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt

packages/kit/src/components/Password/components/PasswordVerify.tsx
```diff
@@ -208,9 +207,7 @@ function PasswordVerify({
           Dialog.confirm({
             icon: 'ErrorOutline',
             tone: 'warning',
-            ...inAppStateLockStyle,
-            isOverTopAllViews: true,
-            portalContainer: Portal.Constant.APP_STATE_LOCK_CONTAINER_OVERLAY,
+            ...inAppStateLockDialogProps,
```
sidmorizon: 🟡 P2: Biometric warning disappears outside lock

Severity: `non-severe`

On iOS, `inAppStateLockDialogProps` routes every biometric-change warning into the lock portal. Unlocked password prompts lack that portal, so disabling biometric authentication shows no warning.

sidmorizon: # 🟠 [P1] Biometric-changed warning stops rendering on iOS when the app is not locked

## Problem

This call site is not lock-screen-only, but it now unconditionally spreads the lock-screen props, which carry `isOverTopAllViews: false` (`packages/kit/src/views/Setting/hooks/index.tsx:57`) together with `portalContainer: APP_STATE_LOCK_CONTAINER_OVERLAY` (`index.tsx:58`).

`checkAuthChanged()` runs on mount and on every `onVisibilityStateChange` whenever `platformEnv.isNativeIOS || platformEnv.isDesktopMac`, with no lock-state guard (`PasswordVerify.tsx:240-275`). `PasswordVerifyContainer` is mounted by the ordinary in-app password prompt (`packages/kit/src/components/Password/container/PasswordVerifyPromptMount.tsx:97`) and by onboarding (`packages/kit/src/views/Onboardingv2/pages/CreatePasscodePage.tsx:69`), not only by the lock screen (`packages/kit/src/provider/Container/AppStateLockContainer/index.tsx:202`).

`APP_STATE_LOCK_CONTAINER_OVERLAY` is registered only by `AppStateContainer` (`.../AppStateContainer/index{,.ios,.android}.tsx`), which is mounted only inside `AppStateLock`, i.e. only while `isLocked` is true (`AppStateLockContainer/index.tsx:188-213`).

Failure path on iOS while unlocked: `Dialog.confirm` takes the `portalContainer` branch, `renderToContainer.ios.tsx:11` sees `isOverTopAllViews === false` and skips the `FULL_WINDOW_OVERLAY_PORTAL` render, so only `Portal.Render(APP_STATE_LOCK_CONTAINER_OVERLAY, element)` runs — and `renderToPortal` finds no manager, retries 10 times logging `renderPortalFailed`, then gives up (`packages/components/src/hocs/Portal/index.tsx:62-78`).

Before this PR the site passed `isOverTopAllViews: true`, so `renderToContainer.ios.tsx:11-16` also rendered into `FULL_WINDOW_OVERLAY_PORTAL`, which always exists — the dialog was visible. Web/desktop already required the container even with the flag (`renderToContainer.tsx:12-16`) and Android always ignored it, so this is an iOS-specific regression.

## Impact

On iOS, `setBiologyAuthEnable(false)` at `PasswordVerify.tsx:205` still executes, but the user is never told that biometric unlock was silently turned off — the only surface reporting that state change disappears. The orphaned dialog instance is still registered, so a later `close()` burns another retry loop.

## Recommendation

Gate the spread on the lock state instead of applying it unconditionally. `PasswordVerifyContainer` already derives `isLock` from `name === 'lock'` (`PasswordVerifyContainer.tsx:101`); forward it into `PasswordVerify` and use it at `PasswordVerify.tsx:210`, e.g. `...(isLock ? inAppStateLockDialogProps : { isOverTopAllViews: true })` — or read `useAppIsLockedAtom()` here. The other two consumers already guard correctly (`views/Setting/hooks/index.tsx:102`, `showExportLogsDialog.tsx:495`); this is the one site that lost its guard.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/Footer.tsx
```diff
@@ -22,24 +22,50 @@ import type { IDialogFooterProps } from './type';
...
-        }
-      };
-    }
+      // The instance changed underneath the previous subscription, so its
+      // current value has never been through `disabledOn` — evaluate it now
+      // instead of waiting for a change that `watch` would report.
+      updateStatus((i) => i + 1);
+    };
+    syncSubscription();
+    const timer = setInterval(syncSubscription, FORM_LOOKUP_INTERVAL_MS);
```
sidmorizon: # 🟡 [P2] 10 Hz poll runs for the whole lifetime of every dialog using `disabledOn`, including those with no `Dialog.Form`

## Problem

The interval created at `Footer.tsx:64` is the only mechanism used to notice that `Dialog.Form` registered, and nothing ever stops it short of the effect cleanup at `Footer.tsx:65-68`. It therefore keeps firing every 100 ms for as long as the dialog is open, even after the form has been found and the subscription is stable — and also for dialogs that pass `disabledOn` without any `Dialog.Form`, where `getForm()` returns `undefined` forever and `syncSubscription` (`Footer.tsx:48-62`) can only hit its early return at line 50-52. There are ~24 `disabledOn` call sites across Swap, Market, LiteCard and Setting, so this is not a one-off path.

The event this polls for is deterministic, not racy: `Dialog.Form` is a lazy module component (`Dialog/index.tsx:104-134`), and `DialogForm` assigns `dialogInstance.ref.current = form` (`DialogForm.tsx:13-18`) — a single, observable point. A push channel for exactly this already exists in `DialogContext.footerRef.notifyUpdate` (`Footer.tsx:77-78`, `type.ts:29-32`).

## Impact

A permanent background timer on every such dialog on all platforms including React Native, and correctness that depends on a poll interval rather than on the registration itself. It also leaves the dependency invisible: nothing in `DialogForm.tsx` indicates the footer's subscription hinges on that assignment.

## Recommendation

Replace the poll with an explicit notification: add a registration callback to the dialog context (or reuse `footerRef`), invoke it in `DialogForm.tsx:13-18` right after `dialogInstance.ref.current = form`, and in `Footer.tsx:42-69` run `syncSubscription()` once on mount plus once per notification. The existing `form === watchedForm` identity guard at `Footer.tsx:50` already makes repeat notifications a no-op, so the re-register case stays covered.

If you prefer to keep the poll as a safety net, at minimum stop it once a form has been found — `clearInterval(timer)` inside `syncSubscription` after the first successful `form?.watch(...)` at `Footer.tsx:55` — so it does not run for the whole dialog lifetime.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

huhuanming:
## Review follow-up — pushed `9eef3e2`

All four points were right. One of them exposed that my own iOS root-cause claim was wrong, so I corrected it rather than building on it.

### 1. iOS main entry — fixed, with a controlled A/B

The button never ran its handler, and it was **not** the dialog. `variant="tertiary"` renders the Button as its label plus 4pt of padding, and its `my: -5 / mx: -9` negative margins pull the frame above the glyphs — the real touch target is a ~26pt band sitting higher than the text, with no `hitSlop`. Taps on the visible label miss it.

Fix: `hitSlop={NATIVE_HIT_SLOP}`, which is what `IconButton` already applies to small buttons.

Evidence is the **background runtime's own log file** on the simulator (`Library/Caches/logs/app-latest.log`, written by the native logger), not a screenshot. `useResetApp` calls `serviceApp.isAppLocked()`, so a handler run leaves an unforgeable `[BgTransport] … method=serviceApp.isAppLocked` record. Same device, same build shape, same tap coordinate `(201, 815)` — the centre of the label:

| build | tap | `isAppLocked` lines before → after |
|---|---|---|
| without `hitSlop` | (201, 815) | 15 → **15** (handler never ran) |
| with `hitSlop` | (201, 815) | 15 → **20** (handler ran, sheet opened) |

Tapping `(201, 808)` — 7pt higher — reaches the handler on *both* builds, which is what pins the cause to the target's size and offset rather than to anything being dead.

```
[BgTransport] callRemoteRequest: ready, type=service-call, method=serviceApp.isAppLocked
[BgTransport] dispatchRemoteRequest: callId=158, type=service-call, method=serviceApp.isAppLocked
[BgRPC] exec-start callId=158, request=service-call:serviceApp.isAppLocked
[BgRPC] exec-done callId=158, request=service-call:serviceApp.isAppLocked, elapsedMs=3
```

**Correction to my earlier analysis.** I had also changed `useAnimatedStyle` to stop carrying `flex: 1`, on the theory that Reanimated was desynchronising paint from hit-testing by ~35pt. The same A/B shows that change makes no difference to this button — the earlier 35pt reading came from my scratch probe rows shifting the layout between measurements, not from the animated style. That change is **reverted**; it is not in this PR.

### 2. PasswordVerify scope — fixed

`PasswordVerifyContainer` already knows whether it is the lock screen (`name === 'lock'` → `isLock`). That is now passed down as `inAppStateLock`, and the biometric-changed warning only takes the lock-screen props when it is actually on the lock screen. Outside it the dialog keeps ordinary rendering instead of targeting a container nothing hosts.

### 3. Footer poll — removed

No timer. `Dialog.Form` announces itself through a new `registerForm` on the dialog instance, `BaseDialogContainer` fans that out to subscribers, and the footer moves its `watch` subscription when it fires. Registration is now an event, which is what it should have been.

### 4. Tests — rewritten around behaviour

- `Dialog.confirmDisabled.test.tsx` drives a **real `react-hook-form`** through the real registration channel: late registration after the footer mounted, a remount that replaces the instance (asserting the stale one no longer drives the button and the live one does), and the value going back out of range.
- `AppStateContainer/index.test.tsx` renders the real web container and asserts the portal genuinely shares one `document.body` child with the lock screen — the property the inert sweep depends on — and that the root fills the viewport.

The config-shape assertion in `hooks/index.test.tsx` is still there, but it is now the smallest of the three rather than the only one.

### Still not verified — not claiming otherwise

- **Real extension popup / side panel.** Verified so far only on the web dev server at 375×812, which runs the same `AppStateLockContainer`, `renderToContainer.tsx` and Tamagui sheet path. A CDP run against a loaded unpacked extension is not done.
- **Android**, and the lock screen's **biometric-changed** and **log-upload** dialogs.

Happy to hold the PR until the extension run is done — say the word and I will not ask for review before that.

packages/kit/src/provider/Container/AppStateLockContainer/components/AppStateLock.tsx
```diff
@@ -23,6 +23,7 @@ import {
   useSafeAreaInsets,
 } from '@onekeyhq/components';
 import Logo from '@onekeyhq/kit/assets/logo_round_decorated.png';
+import { NATIVE_HIT_SLOP } from '@onekeyhq/components/src/utils/getFontSize';
```
sidmorizon: # 🟡 [P2] This import's position fails `lint (24.x)` on the current head

## Problem

The new import at `AppStateLock.tsx:26` was placed directly after `import Logo from '@onekeyhq/kit/assets/logo_round_decorated.png';` (line 25). oxlint rejects that ordering, and the `lint (24.x)` job on `9eef3e2` fails with exactly one error:

```
@onekeyhq/components/src/utils/getFontSize import should occur before import of @onekeyhq/kit/assets/logo_round_decorated.png
Found 0 warnings and 1 error.
```

([run 34703526771, job 103579350991](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34703526771/job/103579350991)) This is the job's only failure — TypeScript and `check:third-party-native-storage` both passed — and it is introduced by this commit, not inherited from `x`.

## Impact

CI is red on the current head, so the PR cannot merge. `CLAUDE.md` also requires that code failing lint is not committed.

## Recommendation

Move line 26 above line 25 so the `@onekeyhq/components` specifier precedes the `@onekeyhq/kit` one. Better still, fold it into the existing `@onekeyhq/components` block at lines 12-24 — `NATIVE_HIT_SLOP` is re-exported from the package root and every other kit consumer (`TxHistoryAddressInfo.tsx`, `NetworkSelectorTrigger.tsx`, `SwapProviderListItem.tsx`) imports it from there, which removes the deep `src/utils/...` path as well as the ordering error.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/Dialog.confirmDisabled.test.tsx
```diff
@@ -0,0 +1,201 @@
...
+        listeners.add(listener);
+        return () => {
+          listeners.delete(listener);
+        };
+      },
+    },
+    footerRef: { notifyUpdate: undefined, props: undefined },
+  } as unknown as IDialogContextType;
+
+  render(<Harness controller={controller} contextValue={contextValue} />);
```
sidmorizon: # 🟡 [P2] The rewrite drops every teardown assertion, leaving the new subscription's cleanup untested

## Problem

The previous version of this suite asserted teardown directly — `expect(firstForm.listenerCount()).toBe(0)` after the form was replaced, and a whole `it('drops its subscription when the footer unmounts')` asserting `listenerCount() === 0` after `unmount()`. Both are gone, and `renderFooter` no longer captures `unmount` from `render(...)` at line 135 or returns it (lines 137-148). Grepping the new file for `unmount`, `unsubscribe` or `listenerCount` returns nothing.

Nothing left covers cleanup. The remount test at lines 165-185 does not: line 178's `expect(confirmButton().disabled).toBe(true)` passes whether or not `Footer.tsx:46` unsubscribed, because the render-time `disabledOn?.({ getForm })` at `Footer.tsx:64` reads `getForm()`, which the harness has already re-pointed at `secondForm` (value `''`). A leaked `firstForm` subscription only causes an extra re-render, and the assertion stays green.

## Impact

The exact property this redesign needs — that replacing the poll with `subscribeFormChange` is leak-free — is now unverified. Dropping `subscription?.unsubscribe()` from `watchCurrentForm` (`Footer.tsx:46`) or the cleanup return (`Footer.tsx:59-62`) keeps all three tests passing, and the resulting react-hook-form listener leak (one per form remount, per dialog) ships silently. `Footer.tsx` is on every dialog in the app.

## Recommendation

Restore teardown coverage. Make the subscription observable in the harness — spy on `form.watch` and capture the `{ unsubscribe }` it returns — then assert (a) after `remountForm()` that the first form's `unsubscribe` was called, and (b) in a restored unmount test, that it is called on teardown. That needs `render(...)` at line 135 to keep its `unmount` and `renderFooter` to return it alongside the existing helpers.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/kit/src/provider/Container/AppStateLockContainer/components/AppStateLock.tsx
```diff
@@ -62,11 +63,19 @@ function diagLog(msg: string) {
   webAuth.log(msg);
 }
 
+// `flex: 1` is static and belongs in a plain style, NOT in the animated one.
```
sidmorizon: # 🟡 [P2] This comment documents a root cause that was retracted, and the code below it contradicts the mechanism it describes

## Problem

Two independent signals say this comment is wrong about why the button was dead:

1. The follow-up comment on this PR states the `useAnimatedStyle` change "makes no difference to this button", that the 35pt reading came from scratch probe rows shifting the layout between measurements, and that the change "is **reverted**; it is not in this PR". It is not reverted — lines 66-98 are present at `9eef3e2`, comment included.
2. The shipped code refutes the stated mechanism on its own terms. The comment says declaring a layout prop inside `useAnimatedStyle` hands the view to Reanimated's UI-thread path and desyncs paint from hit-testing — but the worklet at lines 78-80 still declares `bottom`, which is just as much a layout prop. If that were the mechanism, removing only `flex: 1` would not close it. `flex: 1` was also a constant equal to the committed Yoga value, so it cannot by itself produce a 35pt offset.

The code change is fine and worth keeping: hoisting a static value into `lockScreenFillStyle` (line 98) and composing it at line 209 is correct regardless of the iOS story, and the web branch returning `undefined` (line 96) stays equivalent because the array restores `flex: 1`. The defect is that the only durable record of why this file is shaped this way asserts a disproven cause as fact.

## Impact

This comment plus the `hitSlop` comment at lines 258-264 give two mutually exclusive causes for the same symptom, and the file's own worklet contradicts this one. The next person debugging a lock-screen touch issue is sent into a Reanimated layout-prop investigation that is not the mechanism, and may "restore consistency" by moving `bottom` out of the worklet — which would break keyboard avoidance — or drop `hitSlop` as redundant.

## Recommendation

Keep both code edits and cut this comment to what was actually observed: static style values belong outside the worklet, so `flex: 1` moves to `lockScreenFillStyle`. Drop the 35pt figure, the paint-vs-hit-test narrative and the claim that this is what made the button dead. Leave the verified cause in the `hitSlop` comment at lines 258-264, which the A/B in the PR thread does support.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/DialogForm.tsx
```diff
@@ -10,12 +10,19 @@ import type { IDialogFormProps } from './type';
...
       dialogInstance.ref.current = form;
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [form]);
+  // …and announce it from an effect, where subscribers may re-render. This
+  // module is lazy, so the announcement is how the footer learns the form it
+  // watches has appeared or been replaced. (OK-62416)
+  useEffect(() => {
+    dialogInstance?.registerForm?.(form);
+  }, [dialogInstance, form]);
```
sidmorizon, line 38: 🟡 P2: Removed forms remain active

Severity: `non-severe`

When `DialogForm` unmounts, `registerForm` retains it indefinitely. Confirmation keeps validating and reading the removed form's stale values.

originalix:

Review of commit `9eef3e2ed3b4` finished, but some comments could not be published automatically.

huhuanming:
## Acceptance complete — real extension, real Android, real iOS (`59f3245`)

The remaining items from the review are done. Every claim below is backed by output from unmodified product code — CDP probes of the real extension's DOM, the native log file on device, and screenshots — not by bypassing an entry point.

### 1. Real extension: popup **and** side panel, pre-fix vs after-fix

Built the real extension (`apps/ext`), loaded it unpacked into Chrome over CDP (`Extensions.loadUnpacked` — Chrome 153 ignores `--load-extension`), onboarded a wallet with a **passcode** (PassKey toggle explicitly turned off, and `servicePassword.setBiologyAuthEnable(false)` before locking, so no biometric prompt can auto-unlock the screen under test), locked via `servicePassword.lockApp()`, then drove `Forgot passcode?` → type `RESET` → `Cancel`.

The probe reports who owns which `document.body` child, because that is exactly what the inert sweep keys on.

**Before the fix — `ui-popup.html` @ 375×600:**

```json
"bodyChildrenOfInterest":[
  {"i":7,"tag":"DIV","inert":false,"hasLock":true,"hasReset":false},
  {"i":8,"tag":"SPAN","inert":true, "hasLock":false,"hasReset":true}],
"lockAndResetShareOneBodyChild":false,
"resetSubtreeInert":true
```

The sheet escaped to its own body child and the lock screen's sweep marked it `inert`. Playwright then refused every click on it:

```
<div … AppStateLock.tsx">…</div> from <div>…</div> subtree intercepts pointer events
```

That is the QA report reproduced in the shipping artifact — the freeze is an inert subtree, not a hang.

**After the fix — same build steps, same script:**

| surface | body child | inert | typed value | confirm `aria-disabled` | Cancel |
|---|---|---|---|---|---|
| popup `ui-popup.html` 375×600 | one child holds lock **and** dialog | `false` | `"RESET"` | `true` → **`false`** | dismissed = `true` |
| side panel `ui-side-panel.html` 400×700 | one child holds lock **and** dialog | `false` | `"RESET"` | `true` → **`false`** | dismissed = `true` |

```json
"lockAndResetShareOneBodyChild":true, "resetSubtreeInert":false,
"resetInputValue":"RESET", "confirmAriaDisabled":"false",
"sheetRect":{"x":0,"y":600,"w":375,"h":136}      // popup
"sheetRect":{"x":0,"y":348,"w":400,"h":352}      // side panel
```

`confirmAriaDisabled` going `true → false` is the Footer fix observed in the real extension: the confirm button now follows the lazily-registered `Dialog.Form`.

### 2. Android — emulator, real build

Lock screen → `Forgot passcode?` → dialog opens, accepts `RESET`, Confirm becomes enabled, Cancel dismisses. (Two environment quirks cost time and are worth knowing: the LogBox dev toast covers the button, and the soft keyboard shifts the button ~180px so taps land on IME keys — hide the keyboard first.)

### 3. iOS — real button, real device log

Covered in the previous comment: same-coordinate A/B at `(201, 815)` on the simulator, `serviceApp.isAppLocked` records in `Library/Caches/logs/app-latest.log` going 15 → **15** without `hitSlop` and 15 → **20** with it.

Also verified on the same build: the **log-upload dialog** (3 taps on the logo) renders above the lock screen and dismisses correctly — it uses the same `inAppStateLockDialogProps`.

### 4. Biometric-changed warning — scoping proven, on-device trigger not reproducible

This is the one item I could not drive end-to-end, and I am not going to claim otherwise.

`checkBiometricAuthChanged` compares `LAContext.evaluatedPolicyDomainState` against a keychain baseline. On the simulator that gives no reachable "changed but still supported" state:

- un-enrolling Face ID makes `isSupportBiologyAuth()` false, and `checkAuthChanged` early-returns before the dialog;
- toggling enrollment off and back on leaves `evaluatedPolicyDomainState` **identical**. The native module logs on every outcome it can distinguish, and after a toggle cycle plus relaunch the log stays silent — no `[Biometric] Biometric auth change detected`, only the earlier one-time
  `01:16:56 | INFO : [Biometric] No previous biometric state stored, saving initial state`,
  which is `changed == false` by that module's own reckoning.

So instead the regression itself is pinned down in code. New suite `PasswordVerify.test.tsx` renders the **real** `PasswordVerify` with a real `react-hook-form`, drives the real `checkAuthChanged` path, and asserts both sides:

- on the lock screen the warning carries `inAppStateLockDialogProps`;
- everywhere else it carries **no** `portalContainer`, no `sheetProps`, no `isOverTopAllViews`.

Mutation-checked: restoring the unconditional spread (the exact regression raised in review) fails the second test and passes the first.

```
✕ leaves the warning in the default portal everywhere else
```

What remains unproven is only that *this particular dialog's* pixels land above the lock screen on iOS — and that path is the same `Dialog.confirm` with the same props object into the same container as the reset dialog and the log-upload dialog, both of which are verified on that device.

### Test suite

11 tests, all passing: `PasswordVerify.test.tsx` (2), `AppStateContainer/index.test.tsx` (2), `Dialog.confirmDisabled.test.tsx` (3), `Setting/hooks/index.test.tsx` (4). `yarn agent:check --profile commit` passes.

### Environment notices (`ONEKEY_USER_NOTICE`) — reporting as required

Neither affects the result, but both must be surfaced rather than described as clean runs:

1. **iOS dev-shell, first attempt — failed.** `Error: [devVendor] Missing android cache…`, then
   `[ONEKEY_USER_NOTICE] … action=run-failed reason="[nativeDevShell] Metro exited with code 1 signal null."`, `status=failed userNoticeRequired=true`. The retry ran clean (`status=running userNoticeRequired=false`) and is the run all iOS evidence comes from.
2. **Android dev-shell — fell back to a local Gradle build** after a transient APK download failure:
   `[ONEKEY_USER_NOTICE] … resource=shell action=local-build reason="terminated"`, `status=running shell.source=local-build userNoticeRequired=true`. Not a cache hit.

packages/kit/src/components/Password/components/PasswordVerify.test.tsx
```diff
@@ -0,0 +1,220 @@
...
+      onBiologyAuth={() => {}}
+      onInputPasswordAuth={() => {}}
+      {...props}
+    />,
+  );
+}
+
+describe('PasswordVerify biometric-changed warning', () => {
+  beforeEach(() => {
+    jest.useFakeTimers();
```
sidmorizon: # 🟡 [P2] Faking `queueMicrotask` makes this suite burn React's one-shot "act without await" warning

## Problem

`jest.useFakeTimers()` at line 183 fakes `queueMicrotask` as well as the clock — that is the Jest 29 modern-timers default. React's `act(async cb)` schedules its "did the caller await this act?" check through `queueMicrotask`. `flush()` calls `jest.advanceTimersByTime(100)` at line 161 from *inside* the async `act` callback, which drains the faked microtask queue while `act()` has not yet returned, so React's `didAwaitActCall` is still `false` and it logs `You called act(async () => ...) without await` — even though both tests do `await flush()` correctly.

Reproduced by running this file against a jest 29.7 / React 19 / `@testing-library/react` 16.3 harness pointed at these sources: the warning fires on every run, with the stack `react.development.js:835 ← runJobs (@sinonjs/fake-timers) ← advanceTimersByTime ← PasswordVerify.test.tsx:161`. Both tests still pass.

## Impact

Two consequences, neither of which fails the suite today:

1. A permanent `console.error` in the unit-test CI output for this file.
2. React's `didWarnNoAwaitAct` is a one-shot module-level flag. Burning it here means a genuinely missing `await act(...)` added to this file later is swallowed silently. With `queueMicrotask` faked, React's act queue in this file is drained only by `advanceTimersByTime`, so a future assertion that depends on a state update becomes ordering-sensitive instead of failing loudly.

## Recommendation

One line, at line 183:

```ts
jest.useFakeTimers({ doNotFake: ['queueMicrotask'] });
```

Verified against the same harness: the warning disappears, both tests still pass, and the mutation check still holds — restoring the unconditional `...inAppStateLockDialogProps` spread at `PasswordVerify.tsx:216` still fails the "leaves the warning in the default portal everywhere else" test and passes the lock-screen one.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

huhuanming (commented):
结论：核心修复方向现在是合理的——锁屏 dialog 被约束在同一个 portal/overlay 内，`PasswordVerify` 也已经按锁屏场景收口；Footer 改为由 `Dialog.Form` 注册事件驱动，100ms 定时轮询已经移除。

但当前实现还不建议直接合并，至少需要处理下面这些问题：

1. **表单生命周期没有闭环。** `DialogForm` 只在挂载/替换时调用 `registerForm(form)`，卸载时没有清空注册。这样 `formRef` 以及 Footer 可能继续持有、监听已卸载的旧表单，Confirm 也可能读取或校验陈旧值。`registerForm` 已经接受 `undefined`，应在 effect cleanup 中带身份判断地注销，避免旧表单 cleanup 覆盖刚注册的新表单。
   - https://github.com/OneKeyHQ/app-monorepo/pull/13359#discussion_r3996830745

2. **清理逻辑没有被测试。** 当前三条 Footer 用例覆盖了晚注册、实例替换和字段值变化，但没有真正卸载 Footer/DialogForm，也没有断言旧 `watch` 的 `unsubscribe` 被调用。删除现有 unsubscribe 逻辑后，这些行为断言仍可能继续通过。
   - https://github.com/OneKeyHQ/app-monorepo/pull/13359#discussion_r3996773959

3. **仍保留了已被验证为错误的 Reanimated 根因说明。** 后续 A/B 已确认按钮问题来自实际触摸区域，应删除“Reanimated 导致约 35pt 绘制/命中错位”的错误归因，并撤掉与本修复无关的改动，避免给后续排障留下错误结论。
   - https://github.com/OneKeyHQ/app-monorepo/pull/13359#discussion_r3996774748

4. **PR 描述与当前实现严重不一致。** 目前仍写着 identity poll/100ms interval、iOS 没有修复、真实扩展和 Android 未验收，但后续代码与验收评论已经推翻了这些描述；“unmount cleanup”也与实际测试不符。请在合并前按最终实现和真实验收结果更新正文。

建议：补齐 `DialogForm` 卸载注销和对应回归测试，验证 watch 在替换与卸载时都被释放，清理错误归因并更新 PR 描述后再合并。

Reviewed head: `e7e61303a1f4c71224f82f93f1ec877693d76566`

packages/kit/src/views/Setting/hooks/index.tsx
```diff
@@ -70,14 +99,10 @@ export function useResetApp(
       }
     }
     Dialog.show({
-      ...(inAppStateLock ? inAppStateLockStyle : undefined),
+      ...(inAppStateLock ? inAppStateLockDialogProps : undefined),
```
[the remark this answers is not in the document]
huhuanming: Correct, and this was a live regression — fixed in `0cf6027`.

`isOverTopAllViews: platformEnv.isNative` is now passed unconditionally at the top of `Dialog.show`, and the lock-screen spread overrides it with `false` + the lock container only in the `inAppStateLock` branch:

```tsx
isOverTopAllViews: platformEnv.isNative,
...(inAppStateLock ? inAppStateLockDialogProps : undefined),
```

Two tests pin both sides: `keeps the native top-level overlay for an unlocked reset dialog` and `still hands the lock screen its own container on native`.

packages/kit/src/components/Password/components/PasswordVerify.tsx
huhuanming: Fixed in `0cf6027`. `PasswordVerifyContainer` now forwards its existing `isLock` as `inAppStateLock`, so the dialog only takes the lock-screen props on the lock screen.

Also fixed the half my first pass got wrong: the non-lock path keeps `isOverTopAllViews: true`, which it had before this PR — see the reply on the P1 thread.

huhuanming: Right on both counts, including the part my first fix missed. Fixed in `0cf6027`.

`inAppStateLock` is threaded from `PasswordVerifyContainer.isLock`, and — as this comment recommended — the non-lock path keeps `isOverTopAllViews: true` rather than dropping to `undefined`:

```tsx
isOverTopAllViews: true,
...(inAppStateLock ? inAppStateLockDialogProps : undefined),
```

`PasswordVerify.test.tsx` renders the real component and drives the real `checkAuthChanged` path for both cases; the non-lock test asserts `isOverTopAllViews === true` and no `portalContainer`. Mutation-checked: restoring the unconditional spread fails it.

packages/components/src/composite/Dialog/Footer.tsx
huhuanming: Agreed and done — the poll is gone (`9eef3e2`), replaced by exactly the push channel this comment describes.

`BaseDialogContainer` exposes `registerForm` / `subscribeFormChange` on the dialog instance; `DialogForm` calls `registerForm(form)` from an effect right after assigning `ref.current`; the footer moves its `watch` subscription when that fires. No timers.

Teardown is now covered too, after [the follow-up finding on the test file](https://github.com/OneKeyHQ/app-monorepo/pull/13359#discussion_r3996773959): `unsubscribe` is asserted on both form replacement and footer unmount, and `DialogForm` clears its registration on unmount unless a replacement already claimed it.

packages/kit/src/provider/Container/AppStateLockContainer/components/AppStateLock.tsx
huhuanming: Fixed in `0cf6027`, and taking the better half of the suggestion: `NATIVE_HIT_SLOP` now comes from the `@onekeyhq/components` package root, folded into the existing import block, so the deep `src/utils/...` path is gone along with the ordering error.

The ordering itself was already corrected in `59f3245`; this removes the cause rather than the symptom.

packages/components/src/composite/Dialog/Dialog.confirmDisabled.test.tsx
huhuanming: You're right — and the analysis of *why* the remount test doesn't cover it is exactly correct: `disabledOn` reads `getForm()`, which the harness has already re-pointed, so a leaked subscription only costs an extra re-render and the assertion stays green.

Fixed in `0cf6027`, using the spy approach suggested here. `LateForm` patches `form.watch` during render (before the footer subscribes from its effect) and keeps every subscription it hands out, so teardown is asserted rather than inferred:

```tsx
it('drops the replaced form’s subscription instead of leaking it', …)
it('drops its subscriptions when the footer unmounts', …)
```

The first asserts the outgoing form's `unsubscribe` was called and the incoming one's was not; the second asserts it on `unmount()`, plus `subscribeFormChange`'s own listener set going back to empty. `renderFooter` keeps `unmount` from `render(...)` as recommended.

Both mutation-checked against the two lines they protect:

- dropping `subscription?.unsubscribe()` from `watchCurrentForm` → `✕ drops the replaced form’s subscription instead of leaking it`
- dropping the cleanup `return () => {…}` → `✕ drops its subscriptions when the footer unmounts`

Each fails only its own test, so they are independent rather than one assertion covering both.

packages/kit/src/provider/Container/AppStateLockContainer/components/AppStateLock.tsx
huhuanming: You are right, and thank you for catching it — my PR comment said that change was reverted, and it was not. That was my error, not a subtlety.

Both of your technical points also hold: `bottom` in the same worklet is just as much a layout prop, and a constant `flex: 1` equal to the committed Yoga value cannot produce a 35pt offset. The mechanism in that comment is not real.

Rather than keep the edit and rewrite the comment, I reverted the whole `useAnimatedStyle` / `lockScreenFillStyle` change in `0cf6027`. It was a no-op refactor that this bug fix does not need, and shipping it under a disproven rationale in a fix PR is not worth the ambiguity. `AppStateLock.tsx` is now base plus exactly two things: `hitSlop={NATIVE_HIT_SLOP}` and the comment explaining the touch-target measurement the A/B actually supports.

Corrected on the PR thread as well, so the record does not stand as written.

packages/components/src/composite/Dialog/DialogForm.tsx
huhuanming: Correct — fixed in `0cf6027`. The effect now returns a cleanup:

```tsx
const instance = dialogInstance;
instance?.registerForm?.(form);
return () => {
  if (instance?.ref?.current === form) {
    instance?.registerForm?.(undefined);
  }
};
```

The identity guard matters: when the dialog swaps one `Dialog.Form` for another, the replacement can register before the outgoing instance's cleanup runs, and an unconditional clear would wipe the live form.

New suite `DialogForm.test.tsx` covers all three cases with the real `DialogForm` and a real `react-hook-form`: it announces its form, it clears the registration on unmount, and it leaves a newer registration alone. Mutation-checked — removing the cleanup fails `clears the registration when it unmounts`.

packages/kit/src/components/Password/components/PasswordVerify.test.tsx
huhuanming: Confirmed and fixed in `0cf6027` — the warning reproduces here exactly as described:

```
console.error
  You called act(async () => ...) without await. …
✓ renders the warning inside the lock overlay on the lock screen
✓ leaves the warning in the default portal everywhere else
```

Applied your one-line fix:

```ts
jest.useFakeTimers({ doNotFake: ['queueMicrotask'] });
```

The `console.error` is gone, both tests still pass, and the mutation check still holds — restoring the unconditional `...inAppStateLockDialogProps` spread still fails `leaves the warning in the default portal everywhere else`.

The second consequence you flagged is the one that mattered to me: burning the one-shot `didWarnNoAwaitAct` flag would have silently swallowed a genuinely missing `await act(...)` added to this file later.

huhuanming:
## Review round 2 addressed — `27a9413` (all 9 threads replied and resolved)

Four findings were already fixed by later commits; four were live and are fixed in `0cf6027`. One of them is a correction to something I stated wrongly in my previous comment.

### Correction first

I wrote that the `useAnimatedStyle` / `flex: 1` change "is **reverted**; it is not in this PR". **It was not reverted** — it was still there at `9eef3e2`, disproven comment included. That was my mistake.

The review is also right that the comment's mechanism was self-refuting: `bottom` in the same worklet is equally a layout prop, and a constant `flex: 1` matching the committed Yoga value cannot produce a 35pt offset. So rather than keep the edit and rewrite the comment, I reverted the whole thing. `AppStateLock.tsx` is now base plus exactly two things — `hitSlop={NATIVE_HIT_SLOP}` and the comment describing the touch-target measurement the A/B actually supports.

### Live findings fixed

**1. Non-lock paths lost their top-level overlay (P1).** The real bug, and I had it in two places. Before this PR both call sites passed `isOverTopAllViews` unconditionally — `platformEnv.isNative` for Settings → Reset, `true` for the biometric prompt. My conditional spread dropped it for the non-lock case, so on iOS either dialog could land under a native modal page. Both now pass it up front and let the lock-screen spread override it:

```tsx
isOverTopAllViews: platformEnv.isNative,      // useResetApp
...(inAppStateLock ? inAppStateLockDialogProps : undefined),
```

**2. `Dialog.Form` stayed registered after unmount (P2).** The effect now clears the registration on teardown, guarded on identity so a replacement that registered first is not wiped.

**3. Footer test dropped all teardown coverage (P2).** Restored with the suggested spy approach — `LateForm` patches `form.watch` during render and keeps every subscription, so `unsubscribe` is asserted on replacement and on unmount, plus `subscribeFormChange`'s listener set returning to empty.

**4. `useFakeTimers` faked `queueMicrotask` (P2).** Reproduced the `act(async () => ...) without await` error and applied the one-line `doNotFake` fix. Gone.

**5. Import path (P2).** `NATIVE_HIT_SLOP` now comes from the package root, removing the deep `src/utils/...` path along with the ordering error.

### Every new test is mutation-checked

Each protects one specific line, and fails only when that line is removed:

| test | mutation that breaks it |
|---|---|
| `leaves the warning in the default portal everywhere else` | unconditional `...inAppStateLockDialogProps` |
| `drops the replaced form’s subscription instead of leaking it` | drop `subscription?.unsubscribe()` in `watchCurrentForm` |
| `drops its subscriptions when the footer unmounts` | drop the effect's cleanup `return` |
| `clears the registration when it unmounts` | drop `DialogForm`'s cleanup `return` |

18 tests across 5 suites, all passing. `yarn agent:check --profile commit` passes. Branch merged up to `e7e6130`.

packages/kit/src/views/Setting/hooks/index.tsx
```diff
@@ -70,14 +99,15 @@ export function useResetApp(
       }
     }
     Dialog.show({
-      ...(inAppStateLock ? inAppStateLockStyle : undefined),
+      // Settings → Reset is reached from a native modal page on iOS, which is
+      // added to the window stack after the app's own overlay container: the
+      // dialog has to opt into a fresh top-level overlay or it renders beneath
+      // that page. Only the lock-screen branch below replaces this.
+      isOverTopAllViews: platformEnv.isNative,
```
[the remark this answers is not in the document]
huhuanming: Confirmed and fixed in `f6ad7f6b`. The analysis is exactly right — `isOverTopAllViews` was only ever read inside `renderToContainer`, which only runs when `portalContainer` is truthy, so both of my container-less call sites were passing a no-op prop and my comments described behaviour the code did not have.

Worth separating the two sites, because they were not equally broken:

- **`useResetApp` unlocked** — inert before this PR too (`x` passed the same flag with `portalContainer: undefined`), so no regression, but the comment I added claimed otherwise. Now true rather than deleted.
- **`PasswordVerify` unlocked** — a real regression. On `x` that site passed `isOverTopAllViews: true` **and** `portalContainer`, so `renderToContainer.ios` did mount the extra `OverlayContainer`; it was visible only as a side effect of naming a container it had no business using. Dropping the container took the overlay with it.

Fix follows the recommendation, minus the double-mount trap you flagged — the flag is honoured on the container-less branch instead of routing back through `renderToContainer`:

```tsx
if (portalContainer) {
  return renderToContainer(portalContainer, element, isOverTopAllViews);
}
return Portal.Render(
  Portal.Constant.FULL_WINDOW_OVERLAY_PORTAL,
  isOverTopAllViews ? <OverlayContainer>{element}</OverlayContainer> : element,
);
```

One `Portal.Render`, one manager returned, so `destroy()` tears down what was mounted. Off iOS `OverlayContainer` is `({ children }) => children`, so web and Android are byte-for-byte unchanged.

**Deliberate collateral, flagging it rather than slipping it in:** `GlobalErrorHandlerContainer.tsx:37` has been passing `isOverTopAllViews: true` with no container since long before this PR, silently inert. It now works, so on iOS the hardware passphrase dialog gets its own window overlay. That is what the call site asked for, and no other site in the repo is affected — those three are the only `isOverTopAllViews` callers outside the Dialog folder.

**On the test criticism — fair, and addressed.** Mocking `Dialog.show` did mean those assertions could not see the runtime path. The decision now lives in `renderDialogPortal` with its own suite that exercises it for real: container-less + flag wraps in `OverlayContainer`, container-less without the flag stays bare, a named container defers to `renderToContainer` with the flag passed through, and the returned manager is the one that was rendered. Mutation-checked — removing the wrap fails `gives a container-less dialog its own overlay when it asks to be on top` and nothing else.

27 tests across 7 suites pass, including the pre-existing Dialog focus suites that cover this render path.

packages/components/src/composite/Dialog/renderDialogPortal.tsx
```diff
@@ -0,0 +1,42 @@
+import type { ReactElement } from 'react';
```
sidmorizon: # 🟠 [P1] New module is missing from the mobile module-id registry, failing the native union build

## Problem

`renderDialogPortal.tsx` is a new module in the native graph, but `apps/mobile/bundle-registry/module-id-registry.json` was not updated for it — this increment touches only `Dialog/index.tsx`, `renderDialogPortal.tsx` and `renderDialogPortal.test.tsx`.

`Native startup graph budget (24.x)` fails on `f6ad7f6b` after both graphs build:

```
Main graph modules: 14690
Background graph modules: 14644
packages/components/src/composite/Dialog/renderDialogPortal.tsx.
```

([run 34712093080, job 103602568928](https://github.com/OneKeyHQ/app-monorepo/actions/runs/34712093080/job/103602568928)) Confirmed locally at this head: the registry contains every sibling in this folder — `index.tsx`, `Footer.tsx`, `DialogForm.tsx`, `context.ts`, `hooks.ts`, `dialogInstances.ts`, `renderToContainer.ios.tsx`, `renderToContainer.android.tsx` — and zero entries matching `renderDialogPortal`.

## Impact

The iOS union build errors out during segment allocation, so the native bundle cannot be produced at this head and the check stays red. **Auto-merge is currently armed on this PR**, so this is the gate standing between the branch and a merge. It is introduced by this commit, not inherited from `x` — the same check passed on `27a9413a`.

## Recommendation

Run the command the error names and commit the result:

```
```

Note this is a generated file — regenerate it with that tooling rather than hand-adding an entry, so the assigned id stays consistent with how the other Dialog modules were allocated. Worth re-checking whenever a future revision of this PR adds another module under `packages/components/src`.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/renderDialogPortal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  portalContainer?: EPortalContainerConstantName;
+  isOverTopAllViews?: boolean;
+}): IPortalManager {
+  if (portalContainer) {
+    return renderToContainer(portalContainer, element, isOverTopAllViews);
+  }
+  return Portal.Render(
+    Portal.Constant.FULL_WINDOW_OVERLAY_PORTAL,
+    isOverTopAllViews ? (
+      <OverlayContainer>{element}</OverlayContainer>
```
sidmorizon: # 🟠 [P1] Activating the flag puts the hardware-passphrase dialog above the iOS lock screen, with a device-mutating button

## Problem

Wrapping in `OverlayContainer` at line 37 gives the dialog its own `FullWindowOverlay` on iOS (`layouts/OverlayContainer/OverlayContainer.ios.tsx:10`). iOS stacks these by the order they are added — the repo states this itself at `provider/Container/FullWindowOverlayContainer/index.tsx:21-29` ("iOS stacks window overlays in the order they were added … anything presented after it still lands on top") and patches `react-native-screens` with `bringContainerToFront` precisely to work around it.

The lock screen's overlay is **not** added at app start. `AppStateLockContainer/index.tsx:188-213` renders `AppStateLock` only while `isLocked`, and `AppStateContainer/index.ios.tsx:7` wraps it in its own `OverlayContainer`. So the lock window is added at lock time, and any dialog overlay created afterwards lands on top of it.

`GlobalErrorHandlerContainer` is one of the three call sites this commit newly activates (`GlobalErrorHandlerContainer.tsx:37`, `isOverTopAllViews: true`, no `portalContainer`). It is mounted at `provider/Container/index.tsx:84`, inside the subtree `AppStateLockContainer` renders at `:141-145`, and that subtree stays mounted while locked — so its listeners stay live. It fires on `EAppEventBusNames.ShowHardwareErrorDialog` and on any unhandled `DeviceNotOpenedPassphrase` routed through `globalErrorHandler`. Auto-lock tripping while a hardware call is in flight is an ordinary sequence.

Before this commit the same dialog went to `FULL_WINDOW_OVERLAY_PORTAL`, which lives in the overlay added at app start — i.e. *below* the lock window, hidden. That is what changes here.

I also checked the two escape hatches: nothing closes open dialogs on lock (no `dialogInstances` or close-all usage anywhere in `AppStateLockContainer/` or the password containers), and the `inert` sweep that protects web is explicitly skipped on native (`AppStateLockContainer/index.tsx:146-182` returns early for `platformEnv.isNative`).

## Impact

iOS only. The passcode screen is covered by a dialog, and its confirm button calls `serviceHardware.setPassphraseEnabled` (`GlobalErrorHandlerContainer.tsx:50-57`) — a hardware-wallet state mutation reachable without unlocking. This inverts the invariant the rest of OK-62416 establishes, that a dialog over the lock screen must go through `APP_STATE_LOCK_CONTAINER_OVERLAY`.

## Recommendation

The gate belongs kit-side, since `components` cannot import `kit-bg`. Either:

- narrowest — have `GlobalErrorHandlerContainer` suppress or defer the dialog while `useAppIsLockedAtom()` is true; or
- systemic — give the lock screen's `OverlayContainer` (`AppStateContainer/index.ios.tsx:7`) a `bringToFrontToken` that bumps while locked, so the lock window re-fronts above later overlays. The plumbing already exists (`layouts/OverlayContainer/type.ts:10`, used at `FullWindowOverlayContainer/index.tsx:83,104`).

`renderDialogPortal` itself is fine as written — the issue is which callers the change switches on.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/renderDialogPortal.tsx
```diff
@@ -0,0 +1,42 @@
...
+  isOverTopAllViews,
+}: {
+  element: ReactElement;
+  portalContainer?: EPortalContainerConstantName;
+  isOverTopAllViews?: boolean;
+}): IPortalManager {
+  if (portalContainer) {
+    return renderToContainer(portalContainer, element, isOverTopAllViews);
+  }
+  return Portal.Render(
```
sidmorizon: # 🟡 [P2] Newly-wrapped dialogs also stack above the toast overlay on iOS

## Problem

Separate consequence of the same activation, so noting it apart from the lock-screen issue. `<Toaster />` lives in its own `OverlayContainer` mounted at app start (`FullWindowOverlayContainer/index.tsx:104-108`). A dialog overlay created later is added to the window after it and therefore draws over the toast layer.

That file already anticipates this ordering problem and re-fronts the toast container via `bringToFrontToken` — but the token only bumps on `stageRaiseToken` changes (`FullWindowOverlayContainer/index.tsx:44-47`), never when a dialog's overlay mounts.

## Impact

On iOS, a toast arriving while one of the newly-wrapped dialogs is open is drawn behind the dialog's scrim and cannot be tapped. Reachable from `InAppNotification` (`provider/Container/InAppNotification/index.tsx:11`) and `ErrorToastContainer` — e.g. a swap or push banner landing while Settings → Reset is open (`Setting/hooks/index.tsx:101-145`, now wrapped on iOS). Previously these dialogs shared the app overlay and the toast layer sat above them.

Lower severity than the lock-screen case and purely a visual/interaction regression. Note the hardware-passphrase error itself sets `defaultAutoToast: false` (`shared/src/errors/errors/hardwareErrors.ts:95`), so that particular flow does not self-trigger it.

## Recommendation

Bump the toast overlay's `bringToFrontToken` when a dialog overlay mounts and unmounts, reusing the mechanism already there for `stageRaiseToken` at `FullWindowOverlayContainer/index.tsx:44-47`.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

packages/components/src/composite/Dialog/renderDialogPortal.tsx
huhuanming: This is the decisive one, and it retires the module rather than registering it — `f6ad7f6b` is reverted in `eea00aa9`, so `renderDialogPortal.tsx` no longer exists and there is nothing for the registry to carry.

Thank you for the exact failing job; the finding was correct as written. Noting the general rule for later: a new module under `packages/components/src` needs `module-id:update` committed alongside it.

packages/components/src/composite/Dialog/renderDialogPortal.tsx
huhuanming: You are right, and this is the finding that settles it. Reverted in `eea00aa9` — `f6ad7f6b` is gone in full, `dialogShow` is byte-for-byte back to base, and `renderDialogPortal` is deleted.

I had this backwards. I called the `GlobalErrorHandlerContainer` activation "deliberate collateral" on the basis that the call site asked for `isOverTopAllViews: true`, so honoring it matched intent. The chain you traced shows why that reasoning was wrong: the lock overlay is added at lock time, not app start, so a dialog overlay created afterwards lands above it; that container stays mounted while locked and its listeners stay live; nothing closes open dialogs on lock; and the `inert` sweep that saves web returns early on native. So the end state is a dialog over the passcode screen whose confirm button calls `serviceHardware.setPassphraseEnabled`. A flag that has been inert for a long time is not a statement of intent I get to act on inside an unrelated fix — a device-mutating dialog reachable without unlocking is the exact invariant this PR exists to defend.

**Nothing in this PR needed the change.** I checked the caller I was actually trying to serve: `PasswordVerifyPromptMount.tsx:90-92` shows the unlocked passcode prompt itself passes `portalContainer: undefined` on native, so it renders into `FULL_WINDOW_OVERLAY_PORTAL`. The biometric-changed warning raised from inside it goes to the same portal as a later child and stacks above it with no window of its own. The unlocked reset dialog never named a container either, so the flag it used to pass was inert before this PR and inert after.

Both call sites now say that plainly instead of carrying a prop that asserts stacking they do not get, and the tests assert the absence rather than a value. Your two suggested gates — a lock-state guard in `GlobalErrorHandlerContainer`, or a `bringToFrontToken` that re-fronts the lock window — are the right shape for the underlying problem, but they belong to that problem and not to OK-62416: the `isOverTopAllViews`-without-a-container hole is pre-existing on `x` and deserves its own change with its own iOS verification. I have not touched it here.

packages/components/src/composite/Dialog/renderDialogPortal.tsx
huhuanming: Also correct, and gone with the same revert (`eea00aa9`). No dialog is newly wrapped in an `OverlayContainer` any more, so nothing in this PR moves above the toast overlay — Settings → Reset and the unlocked biometric warning both stay in the app-start overlay exactly as on `x`.

Worth keeping the observation on record though: the toast container's `bringToFrontToken` only bumps on `stageRaiseToken`, so it never re-fronts for a dialog overlay. That is latent for the dialogs that already nest one today (the ones passing a container *and* the flag), independent of this PR. It belongs with the same follow-up as the lock-screen thread rather than here.

huhuanming:
## Review round 3 — `f6ad7f6b` reverted in `eea00aa9`

All three findings were correct, and together they say my previous commit was the wrong call. It is reverted in full: `Dialog/index.tsx` is byte-for-byte back to base and `renderDialogPortal` is deleted.

### What I got wrong

Last round I made `isOverTopAllViews` work without a `portalContainer`, and flagged the effect on `GlobalErrorHandlerContainer` as "deliberate collateral" — the call site asked for it, so honoring it matched intent.

That reasoning does not survive the chain in [the lock-screen finding](https://github.com/OneKeyHQ/app-monorepo/pull/13359#discussion_r3997220715): the lock overlay is added at **lock time**, not app start, so any dialog overlay created afterwards lands above it; that container stays mounted while locked with live listeners; nothing closes open dialogs on lock; and the `inert` sweep that protects web returns early on native. Net effect: a dialog over the passcode screen whose confirm button calls `serviceHardware.setPassphraseEnabled`.

A long-inert flag is not a statement of intent I get to act on from inside an unrelated fix — and a device-mutating dialog reachable without unlocking is precisely the invariant this PR exists to defend. The toast-layer finding is the same activation showing up a second time.

### Nothing in this PR needed it

I checked the caller I was actually trying to serve. `PasswordVerifyPromptMount.tsx:90-92` — the unlocked passcode prompt passes `portalContainer: undefined` on native, so it renders into `FULL_WINDOW_OVERLAY_PORTAL`. The biometric-changed warning raised from inside it goes to **the same portal as a later child** and stacks above it with no window of its own. The unlocked reset dialog never named a container either, so the flag it carried was inert before this PR and inert after — "restoring" it in round 2 restored nothing.

So both call sites now simply say that, rather than carrying a prop that asserts stacking they do not get, and the tests assert the absence instead of a value.

### Left for a separate change, deliberately

`isOverTopAllViews` silently doing nothing without a container is pre-existing on `x` and still there. The two gates suggested in review — a lock-state guard in `GlobalErrorHandlerContainer`, or a `bringToFrontToken` that re-fronts the lock window — are the right shape for it, along with the toast container's token never bumping for dialog overlays. All of it needs its own iOS verification and does not belong in OK-62416. Untouched here.

### State

23 tests across 6 suites pass. `yarn agent:check --profile commit` passes. The PR is back to the shape it had at `27a9413` plus accurate comments: three defects fixed (web/extension sheet escaping into the inert sweep, the iOS touch target, the footer's lazily-registered form), with no change to global dialog stacking.

packages/components/src/composite/Dialog/index.tsx
```diff
@@ -675,6 +698,21 @@ function dialogShow({
   isOverTopAllViews,
   ...props
 }: IDialogShowFunctionProps): IDialogInstance {
+  if (platformEnv.isDev && portalContainer && isOverTopAllViews === true) {
```
sidmorizon: # 🟡 [P2] The guard is platform-blind, but the hazard is iOS-only — on web it fires on the only working use of the prop

## Problem

The condition at line 701 has no platform term, yet the three `renderToContainer` implementations differ completely:

- **iOS** (`renderToContainer.ios.tsx:11-17`) — flag truthy ⇒ two `Portal.Render` calls, only the second manager returned. The double-mount and leak the message describes are real here.
- **Android** (`renderToContainer.android.tsx:5`) — the signature is `(container, element)`; the third argument isn't even declared, so the combination is identical to passing the container alone. No hazard.
- **Web/desktop/extension** (`renderToContainer.tsx:12-16`) — flag truthy ⇒ a single `Portal.Render(container, <Component/>)` where `Component` is `createPortal(element, document.body)`. One render, one manager, nothing leaked.

So on web the message at line 712 — "mounts the dialog twice, leaks the first window overlay, and can stack it above the app-state lock screen" — states three things that do not happen in that build.

It also forbids the documented behaviour. `type.ts:139-145` still declares `@platform iOS, Web` with "On web, it will be rendered to `document.body`", and `index.tsx:778-780` drops the flag entirely when no `portalContainer` is named — so container + flag is the *only* shape under which the web feature exists. `useInPageDialog` (`index.tsx:957-965`) injects a `portalContainer` into every call it wraps, so an in-page dialog on web asking to escape to `document.body` is supported code that now takes a red `console.error`.

The lock-screen rationale doesn't carry over either: the `inert` sweep is gated at `AppStateLockContainer/index.tsx:146-147` (`if (platformEnv.isNative || !isLocked) return;`), so it only exists while the lock screen is up.

## Impact

Dev-only, no production effect — `platformEnv.isDev` short-circuits first and the object literal is inside the block. But the guard's whole audience is future callers, and for the web/desktop/extension half it's a false positive that either gets a correct call deleted or teaches people to ignore `console.error` from the Dialog/Portal area, right next to the genuine ones at `hocs/Portal/index.tsx:111,147`. `Dialog.overlayTargeting.test.tsx:36-39` mocks `platformEnv` as `{ isDev: true, isNative: false }`, so narrowing the guard later fails two of its tests.

## Recommendation

Scope the condition to the platform that has the defect:

```ts
if (platformEnv.isDev && platformEnv.isNativeIOS && portalContainer && isOverTopAllViews === true) {
```

and set `isNativeIOS: true` in the test mock at `Dialog.overlayTargeting.test.tsx:36-39`.

If the combination really is meant to be illegal everywhere, then the counterpart is to drop the `isOverTopAllViews` branch at `renderToContainer.tsx:11-16` and the `@platform iOS, Web` doc at `type.ts:139-145` — otherwise the repo implements, documents, and dev-errors on the same call shape.

[Claude session ↗](https://claude.ai/code/session_01F9L52PEqyA62aoxipuwkPt)

---

huhuanming: Correct on every point — fixed in `365671f5`. The guard now carries `platformEnv.isNativeIOS`, and the message says "on iOS" rather than asserting a double-mount that only one of the three implementations performs.

I checked the two claims that decide this, and both hold: `type.ts:139-145` documents `@platform iOS, Web` with "On web, it will be rendered to `document.body`", and `useInPageDialog` injects `portalContainer: portalId` into **every** call it wraps (`index.tsx:957-965`). So container + flag is not merely tolerated on web, it is the only shape in which the web feature exists — and my guard was firing a red `console.error` on it. That is exactly how a guard trains people to ignore the real ones two files over in `hocs/Portal`.

Test updated as recommended (`isNativeIOS: true` in the mock) plus a new case that pins the narrowing itself:

```
✓ reports a dialog that asks for a container and the top of all views on iOS
✓ stays quiet off iOS, where the same pair is a documented feature
```

The second drives both non-iOS shapes — web and Android — through the same call. Mutation-checked: removing `platformEnv.isNativeIOS` from the condition fails it and nothing else, so the platform term is load-bearing rather than decorative.

On your alternative — dropping the `isOverTopAllViews` branch from `renderToContainer.tsx` and the `@platform iOS, Web` doc — I did not take it. Removing a documented, working web behaviour that `useInPageDialog` sits on top of is a real behavioural change needing its own verification, and it is not what OK-62416 is about. Scoping the guard leaves the repo consistent: web implements and documents the pair, iOS rejects it, Android ignores the flag either way.

originalix (approved):

PR #13360, opened 2026-09-12
fix: correct mobile Market banner height

huhuanming:
## Summary

- align the native Market pager header with the measured 188pt banner card plus 16pt vertical padding
- keep the loaded banner height consistent with the existing 204pt loading fallback
- prevent the third token row from being clipped or covered by the category tabs

## Validation

- focused Market banner Jest tests: 5 suites, 39 tests passed
- iOS 18.6 Simulator at 402x874 @3x: measured content height 204pt in portrait/mobile breakpoint
- verified light/dark rendering, Browser-to-Market return, and collapsed sticky-tab scrolling

### DIFF
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
@@ -121,7 +121,7 @@ const EMPTY_MARKET_STOCK_CATEGORIES: IMarketCategoryItem[] = [];
 const MARKET_TAB_ITEM_PRESS_DRAG_GUARD_MS = platformEnv.isNativeIOS ? 700 : 350;
 const MARKET_TAB_ITEM_PRESS_IDLE_GUARD_MS = platformEnv.isNativeIOS ? 180 : 120;
 const MARKET_TAB_BAR_HEIGHT = 44;
-const MARKET_BANNER_HEADER_HEIGHT = 134;
+const MARKET_BANNER_HEADER_HEIGHT = 204;
 
 const STYLES = StyleSheet.create({
   pager: { flex: 1 },
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
@@ -70,6 +70,7 @@ import {
   getMarketMobileBannerHeaderHeight,
   getMarketMobileSecondaryHeaderHeight,
   resolveMarketBannerHeaderDecision,
+  resolveMarketBannerHeaderHeight,
 } from './mobileLayoutUtils';
 
 import type { IMarketPerpsDataCache } from '../components/MarketPerpsList/hooks/useMarketPerpsTokenList';
@@ -422,21 +423,12 @@ function MobileLayoutComponent({
     scope: bannerScope,
     height: getMarketMobileBannerHeaderHeight(bannerList),
   });
-  if (bannerHeaderHeightRef.current.scope !== bannerScope) {
-    bannerHeaderHeightRef.current = {
-      scope: bannerScope,
-      height: getMarketMobileBannerHeaderHeight([]),
-    };
-  }
-  if (
-    isBannerFetched &&
-    bannerList.length > 0 &&
-    (bannerDecisionRef.current.scope !== bannerScope ||
-      !bannerDecisionRef.current.isDecided)
-  ) {
-    bannerHeaderHeightRef.current.height =
-      getMarketMobileBannerHeaderHeight(bannerList);
-  }
+  bannerHeaderHeightRef.current = resolveMarketBannerHeaderHeight({
+    current: bannerHeaderHeightRef.current,
+    scope: bannerScope,
+    isFetched: isBannerFetched,
+    bannerList,
+  });
   bannerDecisionRef.current = resolveMarketBannerHeaderDecision({
     current: bannerDecisionRef.current,
     scope: bannerScope,
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
@@ -67,6 +67,7 @@ import { getDefaultMarketStockCategoryId } from './marketStockCategoryUtils';
 import { shouldHandleMarketPagerPageSelected } from './marketTabSelectionGuards';
 import {
   MARKET_MOBILE_COLUMN_HEADER_HEIGHT,
+  getMarketMobileBannerHeaderHeight,
   getMarketMobileSecondaryHeaderHeight,
   resolveMarketBannerHeaderDecision,
 } from './mobileLayoutUtils';
@@ -121,7 +122,6 @@ const EMPTY_MARKET_STOCK_CATEGORIES: IMarketCategoryItem[] = [];
 const MARKET_TAB_ITEM_PRESS_DRAG_GUARD_MS = platformEnv.isNativeIOS ? 700 : 350;
 const MARKET_TAB_ITEM_PRESS_IDLE_GUARD_MS = platformEnv.isNativeIOS ? 180 : 120;
 const MARKET_TAB_BAR_HEIGHT = 44;
-const MARKET_BANNER_HEADER_HEIGHT = 204;
 
 const STYLES = StyleSheet.create({
   pager: { flex: 1 },
@@ -418,6 +418,25 @@ function MobileLayoutComponent({
     isDecided: false,
     hasBanners: false,
   });
+  const bannerHeaderHeightRef = useRef({
+    scope: bannerScope,
+    height: getMarketMobileBannerHeaderHeight(bannerList),
+  });
+  if (bannerHeaderHeightRef.current.scope !== bannerScope) {
+    bannerHeaderHeightRef.current = {
+      scope: bannerScope,
+      height: getMarketMobileBannerHeaderHeight([]),
+    };
+  }
+  if (
+    isBannerFetched &&
+    bannerList.length > 0 &&
+    (bannerDecisionRef.current.scope !== bannerScope ||
+      !bannerDecisionRef.current.isDecided)
+  ) {
+    bannerHeaderHeightRef.current.height =
+      getMarketMobileBannerHeaderHeight(bannerList);
+  }
   bannerDecisionRef.current = resolveMarketBannerHeaderDecision({
     current: bannerDecisionRef.current,
     scope: bannerScope,
@@ -425,7 +444,7 @@ function MobileLayoutComponent({
     bannerCount: bannerList.length,
   });
   const headerHeight = bannerDecisionRef.current.hasBanners
-    ? MARKET_BANNER_HEADER_HEIGHT
+    ? bannerHeaderHeightRef.current.height
     : 1;
   const [stickyHeaderHeight, setStickyHeaderHeight] = useState(
     MARKET_TAB_BAR_HEIGHT + getMarketMobileSecondaryHeaderHeight(),
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
@@ -15,6 +15,34 @@ export function getMarketMobileBannerHeaderHeight(
     : MARKET_MOBILE_BANNER_LEGACY_HEIGHT;
 }
 
+export type IMarketBannerHeaderHeightState = {
+  scope: string;
+  height: number;
+};
+
+export function resolveMarketBannerHeaderHeight({
+  current,
+  scope,
+  isFetched,
+  bannerList,
+}: {
+  current: IMarketBannerHeaderHeightState;
+  scope: string;
+  isFetched: boolean;
+  bannerList: readonly { tokens?: unknown }[];
+}): IMarketBannerHeaderHeightState {
+  if (isFetched && bannerList.length > 0) {
+    const height = getMarketMobileBannerHeaderHeight(bannerList);
+    return current.scope === scope && current.height === height
+      ? current
+      : { scope, height };
+  }
+  if (current.scope !== scope) {
+    return { scope, height: MARKET_MOBILE_BANNER_LEGACY_HEIGHT };
+  }
+  return current;
+}
+
 export type IMarketBannerHeaderDecision = {
   scope: string;
   isDecided: boolean;
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
@@ -2,6 +2,19 @@ export const MARKET_MOBILE_SECONDARY_HEADER_HEIGHT = 74;
 export const MARKET_MOBILE_COLUMN_HEADER_HEIGHT = 32;
 export const MARKET_MOBILE_CONTENT_TOP_GAP = 16;
 
+const MARKET_MOBILE_BANNER_MODERN_HEIGHT = 204;
+const MARKET_MOBILE_BANNER_LEGACY_HEIGHT = 134;
+
+export function getMarketMobileBannerHeaderHeight(
+  bannerList: readonly { tokens?: unknown }[],
+) {
+  const isModernBannerList =
+    bannerList.length > 0 && bannerList.every((item) => Boolean(item.tokens));
+  return isModernBannerList
+    ? MARKET_MOBILE_BANNER_MODERN_HEIGHT
+    : MARKET_MOBILE_BANNER_LEGACY_HEIGHT;
+}
+
 export type IMarketBannerHeaderDecision = {
   scope: string;
   isDecided: boolean;
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.test.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.test.tsx
@@ -18,6 +18,7 @@ jest.mock('@onekeyhq/components', () => {
     role?: string;
     'aria-label'?: string;
     color?: string;
+    $gtMd?: { h?: number | string };
   }>;
   const Component = ({
     children,
@@ -26,6 +27,7 @@ jest.mock('@onekeyhq/components', () => {
     role,
     'aria-label': accessibilityLabel,
     color,
+    $gtMd,
   }: IProps) =>
     React.createElement(
       'div',
@@ -35,6 +37,7 @@ jest.mock('@onekeyhq/components', () => {
         onClick: onPress,
         role,
         'aria-label': accessibilityLabel,
+        'data-gt-md-height': $gtMd?.h,
       },
       children,
     );
@@ -51,6 +54,11 @@ jest.mock('@onekeyhq/components', () => {
 
 jest.mock('@onekeyhq/kit/src/views/Market/components/PerpsBadges', () => ({
   LeverageBadge: () => null,
+}));
+
+jest.mock('@onekeyhq/shared/src/platformEnv', () => ({
+  __esModule: true,
+  default: { isNative: true },
 }));
 
 const makeToken = (
@@ -185,5 +193,10 @@ describe('Market theme banner', () => {
     render(<MarketBannerItem item={item} />);
     expect(screen.getByText('+1%')).toBeTruthy();
     expect(screen.queryAllByTestId('market-banner-token-row')).toHaveLength(0);
+    expect(
+      screen
+        .getByTestId('market-banner-item')
+        .parentElement?.getAttribute('data-gt-md-height'),
+    ).toBe('118');
   });
 });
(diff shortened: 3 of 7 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9b27ba4c-c63f-5385-8004-6504070acd78)

sidmorizon:
Claude review session: https://claude.ai/code/session_01S8Avvm5j5A5nGTjTRoanjU

packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
```diff
@@ -121,7 +121,7 @@ const EMPTY_MARKET_STOCK_CATEGORIES: IMarketCategoryItem[] = [];
 const MARKET_TAB_ITEM_PRESS_DRAG_GUARD_MS = platformEnv.isNativeIOS ? 700 : 350;
 const MARKET_TAB_ITEM_PRESS_IDLE_GUARD_MS = platformEnv.isNativeIOS ? 180 : 120;
 const MARKET_TAB_BAR_HEIGHT = 44;
-const MARKET_BANNER_HEADER_HEIGHT = 134;
+const MARKET_BANNER_HEADER_HEIGHT = 204;
```
sidmorizon: # 🟡 P2 Hard-coded 204 over-reserves 70pt for legacy (no-`tokens`) banner payloads

## Problem

`MARKET_BANNER_HEADER_HEIGHT` is a single hard-coded value, but the banner card has **two** live render variants with different heights:

- **Modern card** (`components/MarketBanner/MarketBannerItem.tsx:275-326`): `py="$5"` (40) + title row `h="$6"` (24) + `gap="$5"` (20) + token block `minHeight={104}` = **188**.
- **Legacy card** (`MarketBannerItem.tsx:126`): explicit **`h={118}`**.

The legacy branch is not dead code — `MarketBannerItem.tsx:273` routes to it whenever the API item omits `tokens`:

```ts
// Older API deployments do not provide token previews yet.
if (!item.tokens) return <LegacyMarketBannerItem {...props} />;
```

`tokens` is declared optional (`packages/shared/types/marketV2.ts:747`), and the branch is actively covered by tests (`MarketBannerItem.test.tsx:180`, whose `makeBanner()` helper yields `tokens: undefined`).

Adding the mobile container's 16pt padding (`MarketBannerList.tsx:56-60`, `py: '$2'`) gives the two real heights: `188 + 16 = 204` (modern) and `118 + 16 = 134` (legacy) — the latter being *exactly* the pre-PR constant. So this change does not remove the mismatch, it moves it from the modern card to the legacy card.

## Impact

Native only (this file is `.native.tsx`; iOS/Android `main` runtime). When the banner endpoint returns legacy items, the collapsible header reserves 70pt more than the content renders:

- A 70pt empty `$bgApp` band appears between the banner cards and the category tab bar (`MobileLayout.native.tsx:745`, `:759-763`).
- The header needs 70pt of extra scroll travel before the tab bar sticks.
- On Android the same 70pt also inflates empty-list content via `emptyContentPaddingTop` (`MobileLayout.native.tsx:685`).

Severity is P2 rather than P1 because the visible damage is confined to layout and only manifests against legacy payloads; the code path itself is unambiguously reachable.

## Recommendation

Derive the reservation from what will actually render instead of hard-coding one of two valid constants.

Minimal fix — the committed `bannerList` is already in scope at `MobileLayout.native.tsx:411-429`, so branch on it:

```ts
const MARKET_BANNER_HEADER_HEIGHT_MODERN = 204;
const MARKET_BANNER_HEADER_HEIGHT_LEGACY = 134;
// ...
const headerHeight = bannerDecisionRef.current.hasBanners
  ? (bannerList.every((item) => item.tokens)
      ? MARKET_BANNER_HEADER_HEIGHT_MODERN
      : MARKET_BANNER_HEADER_HEIGHT_LEGACY)
  : 1;
```

Preferred fix — measure it, mirroring what the sticky header already does at `:768-772`: attach `onLayout` to the header `YStack` at `:759-763` and feed the measured height into `headerHeight`, keeping 204 only as the initial estimate. That also closes the related 8pt under-reserve on native tablets, where `MarketHomeV2.tsx:338` selects this layout but `MarketBannerList.tsx:114-118` picks the desktop container (`pt="$4" pb="$2"` = 24pt → 212, not 204).

[Claude session ↗](https://claude.ai/code/session_01S8Avvm5j5A5nGTjTRoanjU)

---

sidmorizon: 🟡 P2: Banner height mismatches supported layouts

Severity: `non-severe`

Above 767px, `MARKET_BANNER_HEADER_HEIGHT` stays 204 while preview banners render 212 points. Legacy mobile banners render 134 points. Tabs overlap tablet banners by 8 points or leave a 70-point legacy gap.

originalix (dismissed):

packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
huhuanming: 已采纳建议中的最小固定值方案。d9505210a1 对首批整批数据做一次版本判定：全部 legacy 使用 134，全部 modern 使用 204，不做动态布局测量。你指出的 native tablet 8pt 差异也已在 a47e411ec9 修复：native 大断点容器垂直 padding 统一为 16pt，legacy 卡片继续固定 118pt，因此 native 各断点仍分别为 134/204；Web 和 Desktop 样式不变。

huhuanming: 已解决。d9505210a1 修复 legacy 列表的 70pt 预留空白；a47e411ec9 将 native 大断点容器的垂直 padding 从 24pt 对齐为 16pt，并保持 legacy 卡片 118pt，modern/legacy 在 native tablet 上也与固定 header 204/134 一致。未引入动态测量。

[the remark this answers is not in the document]
huhuanming: 已修复。d9505210a1 根据首次成功返回的整批 banner payload 选择固定 header 高度：全部 legacy 为 134，全部 modern 为 204；同一 scope 只确定一次，不使用 onLayout 或内容高度测量。a47e411ec9 进一步统一了 native 大断点的容器和 legacy 卡片高度，保证两个固定值在 native 支持的断点上都成立。相关聚焦测试已通过。

sidmorizon (dismissed):

originalix:

Review of commit `a47e411ec998` finished, but some comments could not be published automatically.

packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
```diff
@@ -2,6 +2,19 @@ export const MARKET_MOBILE_SECONDARY_HEADER_HEIGHT = 74;
...
 export const MARKET_MOBILE_CONTENT_TOP_GAP = 16;
 
+const MARKET_MOBILE_BANNER_MODERN_HEIGHT = 204;
+const MARKET_MOBILE_BANNER_LEGACY_HEIGHT = 134;
+
+export function getMarketMobileBannerHeaderHeight(
+  bannerList: readonly { tokens?: unknown }[],
+) {
+  const isModernBannerList =
+    bannerList.length > 0 && bannerList.every((item) => Boolean(item.tokens));
```
sidmorizon, line 12: # 🟠 P1 Mixed banner payloads reserve 134 while the row renders 204, re-introducing the 70pt clipping

## Problem

`every()` makes the height all-or-nothing, but the banner row's height is **max-of-children**, not a function of uniformity:

```ts
bannerList.length > 0 && bannerList.every((item) => Boolean(item.tokens));
```

Both containers are plain flex rows with no cross-size constraint — `BannerContainerMobile` (`MarketBannerList.tsx:47-65`, horizontal `ScrollView`) and `BannerContainerDesktop` (`:76-90`, `XStack`) — so the row is as tall as its tallest card. Each variant has a definite height: legacy `h={118}` (`MarketBannerItem.tsx:127`, now also pinned at `$gtMd` on native by `a47e411`), modern intrinsic 188.

So a list containing **one** modern card among legacy ones renders `188 + 16 = 204`, while this function returns **134**.

**Mixed payloads are representable and already present in this repo.** `tokens` is optional *per item* (`packages/shared/types/marketV2.ts:747`), and `fetchMarketBannerList` returns the server's `data.data` unmapped (`ServiceMarketV2.ts:1058-1073, 1076-1082`) — nothing normalises the shape. The repo's own captured sample is itself mixed: `MOCK_MARKET_BANNER_LIST` (`ServiceMarketV2.const.ts:6`) holds **6 banners, all 6 with legacy `tokenLogos`, exactly 1 with `tokens`** (`:11`; the other five at `:51, :70, :89, :108, :127, :146` have none). That is the signature of a per-banner server rollout, and it is reachable today through the shipped "Mock Market Banner Data" dev toggle (`ServiceMarketV2.ts:1077-1080`).

## Impact

On iOS/Android (`main` runtime) with a mixed batch, the collapsible header reserves 70pt too little, so the modern card's third token row is clipped/covered by the category tab bar — **the exact defect this PR exists to fix**. `a47e411` made phone and tablet agree, so this now misses identically at every native breakpoint. On Android it also mis-offsets empty-list content via `emptyContentPaddingTop` (`MobileLayout.native.tsx:685`).

Confirmed reachable in dev/internal builds via the mock. Production reachability depends on what the server currently returns, which I can't read from the repo — flagging that one step as inference rather than asserting it.

## Recommendation

Take the per-item max, which is what the layout actually computes:

```ts
const MARKET_MOBILE_BANNER_MODERN_CARD_HEIGHT = 188;
const MARKET_MOBILE_BANNER_LEGACY_CARD_HEIGHT = 118;
const MARKET_MOBILE_BANNER_CONTAINER_PADDING = 16;

export function getMarketMobileBannerHeaderHeight(
  bannerList: readonly IMarketBannerItem[],
) {
  if (bannerList.length === 0) {
    return MARKET_MOBILE_BANNER_LEGACY_CARD_HEIGHT + MARKET_MOBILE_BANNER_CONTAINER_PADDING;
  }
  return (
    Math.max(
      ...bannerList.map((item) =>
        item.tokens
          ? MARKET_MOBILE_BANNER_MODERN_CARD_HEIGHT
          : MARKET_MOBILE_BANNER_LEGACY_CARD_HEIGHT,
      ),
    ) + MARKET_MOBILE_BANNER_CONTAINER_PADDING
  );
}
```

Both uniform cases still yield 134 / 204, so the two new tests (`mobileLayoutUtils.test.ts:11-21`) keep passing unchanged. Please also add the mixed case — those tests currently pin only `[{}, {}]` and `[{tokens: []}, {tokens: []}]`, so the one broken input is exactly the one not covered, and the suite reads as green. Typing the parameter as `IMarketBannerItem` rather than `readonly { tokens?: unknown }[]` (`:9`) would check it against the real model.

## Verified, for the record

The rest of rounds 2–3 holds up: the modern card has no `$gtMd` override so 188 is breakpoint-invariant; `platformEnv.isNative` is build-time exclusive with web/desktop/ext (`buildTimeEnv.js:15`), so both `a47e411` edits are genuine no-ops off-native; and forcing `h: 118` at `$gtMd` does not clip — it raises the legacy content box from 64pt to 86pt against a 62pt worst case, which is strictly safer than the previous `h: 'auto'` + `minHeight: 96`.

[Claude session ↗](https://claude.ai/code/session_01S8Avvm5j5A5nGTjTRoanjU)

---

huhuanming: 感谢详细核对。你指出的混合输入下的布局计算，以及当前 dev mock 的可达性，都是成立的：现有 mock 为 1 条 modern + 5 条 legacy，every() 会返回 134，而该行实际需要 204，因此开启 mock 时确实可能裁切。

不过我们目前没有确认线上存在混合批次。2026-09-13 使用 App 请求头分别检查了生产和测试接口的 en-US、zh-CN 响应：生产环境均为 9/9 不含 tokens，测试环境均为 9/9 含 tokens，没有出现混合。Git 历史也显示，该 mock 原本全部是 legacy，f27521df25 仅为测试 modern UI 给第一条人工补了 tokens；它不是服务端混合响应的捕获样本。

这个 PR 按已确认的产品约束，将 banner 响应视为整批同一版本，并保留两个固定值：全 legacy 为 134，全 modern 为 204，不使用动态布局测量。因此本次不修改实现。我们保留对 dev mock 与该约束不一致、开启后可复现裁切这一事实；若需要消除该开发场景，应单独把 mock 数据规范成一致批次。

sidmorizon: You're right, and I withdraw the part of my finding that drove the P1 rating.

I checked `f27521df25` locally: it hand-added `tokens` to the first mock entry only, with obviously synthetic values (`logo: ''`, `Mock low gainer` / `Mock top gainer`, `price: '--'`). That is a hand-authored fixture for exercising the modern UI, not a captured server response — so my "signature of a per-banner server rollout" inference was unfounded. I'd flagged production reachability as inference rather than assertion, but I still leaned on the mock's shape as evidence, and that evidence doesn't support the conclusion. Combined with your 9/9 uniform results on both prod and test endpoints, there's no basis for treating mixed production batches as a live risk.

What stands, and you've already confirmed both: the layout arithmetic (a mixed list resolves to 134 while the row measures 204) and the dev-mock reproduction. Given a confirmed product constraint that batches are uniformly versioned, the remaining exposure is a dev-only inconsistency in the fixture, not a defect in this PR's implementation. Normalising the mock separately is the right place to fix it.

No change requested here. Downgrading this to informational; the implementation on `a47e411` is sound under the stated constraint.

[Claude session ↗](https://claude.ai/code/session_01S8Avvm5j5A5nGTjTRoanjU)

---

packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
```diff
@@ -2,6 +2,19 @@ export const MARKET_MOBILE_SECONDARY_HEADER_HEIGHT = 74;
...
+  return isModernBannerList
+    ? MARKET_MOBILE_BANNER_MODERN_HEIGHT
+    : MARKET_MOBILE_BANNER_LEGACY_HEIGHT;
```
sidmorizon, line 15: 🟡 P2: Mixed banners reserve legacy height

Severity: `non-severe`

A mixed response makes `getMarketMobileBannerHeaderHeight` reserve 134 points while [MarketBannerItemComponent](packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx:251-327) renders a 188-point modern card. The pager clips that card and can cover its third token row.

Suggested fix:

```
  const isModernBannerList =
    bannerList.length > 0 && bannerList.some((item) => Boolean(item.tokens));
  return isModernBannerList
    ? MARKET_MOBILE_BANNER_MODERN_HEIGHT
    : MARKET_MOBILE_BANNER_LEGACY_HEIGHT;
```

packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx
```diff
@@ -418,14 +418,33 @@ function MobileLayoutComponent({
...
+    };
+  }
+  if (
+    isBannerFetched &&
+    bannerList.length > 0 &&
+    (bannerDecisionRef.current.scope !== bannerScope ||
+      !bannerDecisionRef.current.isDecided)
+  ) {
+    bannerHeaderHeightRef.current.height =
+      getMarketMobileBannerHeaderHeight(bannerList);
```
sidmorizon: 🟡 P2: Modern refresh retains legacy height

Severity: `non-severe`

After a legacy response, a modern refresh leaves `bannerHeaderHeightRef` at 134 while [MarketBannerListComponent](packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerList.tsx:136-147) renders the new cards. The pager clips each 188-point card and can cover its third token row.

huhuanming: 确认有效，已在 ba98a86349 修复。成功的非空刷新现在会重新计算整批对应的固定档位，并且只在 134/204 实际变化时替换高度状态；因此同一 scope 从全 legacy 切换到全 modern 会更新为 204。空刷新仍保留已占用高度，未引入 onLayout、onContentSizeChange 或其他布局测量。已增加 legacy 到 modern 刷新及空刷新保留高度的回归测试。

packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
huhuanming: 感谢指出。这与 discussion_r3996813402 是同一项混合输入问题：该布局计算在当前 dev mock 下确实可复现，但 2026-09-13 实测生产环境 en-US/zh-CN 均为 9/9 legacy，测试环境均为 9/9 modern；混合 mock 是 f27521df25 在原 legacy mock 的第一条上人工补 tokens UI 数据形成，并非服务端混合响应样本。本 PR 按已确认的整批一致契约保留固定 134/204，不为混合 mock 修改实现；若要覆盖该开发场景，应单独规范 mock 数据。这里作为重复评论关闭。

packages/kit/src/views/Market/MarketHomeV2/layouts/mobileLayoutUtils.ts
```diff
@@ -2,6 +2,47 @@ export const MARKET_MOBILE_SECONDARY_HEADER_HEIGHT = 74;
...
+  if (isFetched && bannerList.length > 0) {
+    const height = getMarketMobileBannerHeaderHeight(bannerList);
+    return current.scope === scope && current.height === height
+      ? current
+      : { scope, height };
+  }
+  if (current.scope !== scope) {
+    return { scope, height: MARKET_MOBILE_BANNER_LEGACY_HEIGHT };
+  }
+  return current;
```
sidmorizon, line 43: # 🟡 P2 Height now tracks the latest list, but the empty-refresh fallback still renders the *first* one — 70pt gap after a tier change

## Problem

This retention path is correct on its own, but it introduces a divergence from `MarketBannerList`, because the two now commit on **different policies**:

- **Height** tracks the *latest* successful non-empty response (`:31-36`, new in `ba98a863`).
- **`initialBannerList`** commits *once per scope* and is never refreshed — `MarketBannerList.tsx:124-126` assigns only when it is still `undefined`.

On an empty response those two meet: `hidden = bannerList.length === 0` (`MarketBannerList.tsx:136`) and `visibleBannerList = hidden ? (initialBannerList ?? []) : bannerList` (`:140`), so the row re-renders the **first** committed list while the height keeps the **last** one's tier.

Concrete sequence, entirely inside the uniform-batch contract (no mixed payload needed):

1. Fetch 1 → all-legacy. Height `134`; `initialBannerList` = legacy. Renders 134. ✅
2. Fetch 2 → all-modern. Height recomputes to `204`; `initialBannerList` **stays legacy**. Renders 204. ✅
3. Fetch 3 → empty. `hidden = true`, so the row renders `initialBannerList` = **legacy cards, 134**. Height hits this `return current` and **stays 204**. ❌

The header stays active throughout: `resolveMarketBannerHeaderDecision` (`:69-73`) returns `current` once `isDecided` is true, so `hasBanners` remains `true` and `headerHeight` is `204` (`MobileLayout.native.tsx:439`).

Note this is **new in this commit**. Before `ba98a863` the height latched once, to the same first list `initialBannerList` committed, so step 3 agreed at 134/134 — the old frozen latch was wrong at step 2 and accidentally right at step 3. Fixing step 2 exposed step 3.

## Impact

Native only. After a tier change followed by an empty or reconnect-cleared response, the collapsible header reserves 70pt more than the row renders: a blank `$bgApp` band above the category tabs and 70pt of extra scroll before the tab bar sticks. On Android it also inflates `emptyContentPaddingTop` (`MobileLayout.native.tsx:696`). Over-reserve, not clipping, so it is cosmetic rather than content-hiding — hence P2.

The empty branch is a real path, not hypothetical: `MarketBannerList.tsx:119-120` and `:138-139` exist precisely to handle "a successful empty response" and "a reconnect removes the banners".

## Recommendation

Align the two commit policies. The smaller change is in `MarketBannerList`: refresh the retained copy on every successful non-empty response instead of only the first, so the fallback renders the same tier the height is holding.

```ts
// MarketBannerList.tsx — replace the commit-once guard at :124-126
if (isFetched && bannerList.length > 0 && initialBannerList !== bannerList) {
  setInitialBannerList(bannerList);
}
```

That keeps the stated intent (commit only successful responses; retain card dimensions when a reconnect clears the list) while making the retained list the most recent one, which is what the height now reflects.

Alternatively, make the height fall back symmetrically — retain the tier of the list that will actually be rendered when `hidden` is true — but that would mean threading `initialBannerList` into this helper, which is the more invasive of the two.

## Verified in this increment

The rest of `ba98a863` checks out, and the round-2 frozen-latch problem is genuinely fixed. `scope` and `swrKey` both derive from `(locale, enableMockMarketBanner)` (`useMarketBannerList.ts:24, 46-48`) so they flip together, and `usePromiseResult.ts:139-151` syncs state during render, so `MobileLayout` never observes a new scope paired with the previous scope's list — the stale-list-on-scope-change risk I looked for is not reachable. A failed fetch (`undefinedResultIfError: true`, `:51`) correctly falls through to `return current`, so the height cannot drop to 134 while modern cards are on screen. The two new tests do pin the claimed behaviour, including `.toBe(current)` for identity.

[Claude session ↗](https://claude.ai/code/session_01S8Avvm5j5A5nGTjTRoanjU)

---

huhuanming: Confirmed. The retained fallback was still pinned to the first successful list while the header height now follows the latest successful non-empty list. Fixed in b364269ba5 by updating the retained fallback on each subsequent successful non-empty response, so legacy -> modern -> empty now keeps the modern cards and the 204pt header aligned. The existing successful-first-empty lock is intentionally preserved, so a native page that started without a banner still does not insert a header later. Added a focused regression test for the tier-change sequence; the existing initial-empty and recovery cases still pass. Heights remain static at 134/204 with no layout measurement.

zhaono1 (approved):

originalix (approved):

PR #13362, opened 2026-09-12
fix: correct market banner data and responsive layouts OK-62951 OK-62954 OK-62957 OK-62967 OK-62968

zy2000-star:
[OK-62951](<https://onekeyhq.atlassian.net/browse/OK-62951>) [OK-62954](<https://onekeyhq.atlassian.net/browse/OK-62954>) [OK-62957](<https://onekeyhq.atlassian.net/browse/OK-62957>) [OK-62967](<https://onekeyhq.atlassian.net/browse/OK-62967>) [OK-62968](<https://onekeyhq.atlassian.net/browse/OK-62968>)

----------

## What changed

Fix Market Banner data synchronization, index presentation, detail navigation, and mobile layout issues.

- Synchronize updated banner quotes with the corresponding Market list list data.
- Correct banner type detection and data-source selection for stock indices.
- Display index banners according to the F Figma design, including fixed ordering and quote formatting.
- Correct Market Banner detail data loading and navigation.
- Prevent banner cards and content from overflowing on mobile.
- Fix Desktop app-state handling when `onAppState` is unavailable.

## Related issues

- OK-62951
- OK-62954
- OK-62957
- OK-62967
- OK-62968

## Validation

- Passed Market Banner targeted tests: 2 suites, 23 tests.
- Passed TypeScript checking.
- Verified Desktop and iOS Market Banner layouts and data rendering.

[OK-62951]: https://onekeyhq.atlassian.net/browse/OK-62951?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62954]: https://onekeyhq.atlassian.net/browse/OK-62954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62957]: https://onekeyhq.atlassian.net/browse/OK-62957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62967]: https://onekeyhq.atlassian.net/browse/OK-62967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62968]: https://onekeyhq.atlassian.net/browse/OK-62968?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit-bg/src/services/ServiceMarketV2.ts
+++ b/packages/kit-bg/src/services/ServiceMarketV2.ts
@@ -6,6 +6,10 @@ import {
   backgroundClass,
   backgroundMethod,
 } from '@onekeyhq/shared/src/background/backgroundDecorators';
+import {
+  DEFAULT_MARKET_STOCK_SORT_BY,
+  DEFAULT_MARKET_STOCK_SORT_TYPE,
+} from '@onekeyhq/shared/src/consts/marketConsts';
 import { OneKeyError } from '@onekeyhq/shared/src/errors';
 import {
   EAppEventBusNames,
@@ -1138,8 +1142,8 @@ class ServiceMarketV2 extends ServiceBase {
         cursor: params.cursor,
         limit: params.limit ?? 20,
         category: params.category,
-        sortBy: params.sortBy ?? 'default',
-        sortType: params.sortType ?? 'asc',
+        sortBy: params.sortBy ?? DEFAULT_MARKET_STOCK_SORT_BY,
+        sortType: params.sortType ?? DEFAULT_MARKET_STOCK_SORT_TYPE,
       },
       headers: { 'x-onekey-request-currency': 'usd' },
       autoHandleError: false,
--- a/packages/kit-bg/src/services/ServiceMarketV2.ts
+++ b/packages/kit-bg/src/services/ServiceMarketV2.ts
@@ -47,6 +47,7 @@ import type {
   IMarketStockPublicChartRequest,
   IMarketStockPublicChartResponse,
   IMarketStockPublicDetail,
+  IMarketStockPublicItem,
   IMarketStockPublicListRequest,
   IMarketStockPublicListResponse,
   IMarketStockPublicSearchRequest,
@@ -1055,36 +1056,25 @@ class ServiceMarketV2 extends ServiceBase {
     }
   }
 
-  private memoizedFetchMarketBannerList = memoizee(
-    async () => {
-      const client = await this.getClient(EServiceEndpointEnum.Utility);
-      const response = await client.get<{
-        code: number;
-        message: string;
-        data: IMarketBannerListResponse;
-      }>('/utility/v2/market/banner/list');
-      const { data } = response.data;
-      return data.data;
-    },
-    {
-      maxAge: timerUtils.getTimeDurationMs({ seconds: 30 }),
-      promise: true,
-    },
-  );
-
   @backgroundMethod()
   async fetchMarketBannerList(): Promise<IMarketBannerItem[]> {
     const devSettings = await devSettingsPersistAtom.get();
     if (devSettings.enabled && devSettings.settings?.enableMockMarketBanner) {
       return MOCK_MARKET_BANNER_LIST;
     }
-    return this.memoizedFetchMarketBannerList();
+    const client = await this.getClient(EServiceEndpointEnum.Utility);
+    const response = await client.get<{
+      code: number;
+      message: string;
+      data: IMarketBannerListResponse;
+    }>('/utility/v2/market/banner/list');
+    const { data } = response.data;
+    return data.data;
   }
 
   @backgroundMethod()
   async clearMarketBannerCache(): Promise<void> {
-    // memoizee's clear() is synchronous, returns void
-    void this.memoizedFetchMarketBannerList.clear();
+    // Kept for callers that clear market caches after account or network changes.
   }
 
   @backgroundMethod()
@@ -1100,6 +1090,20 @@ class ServiceMarketV2 extends ServiceBase {
     );
     const { data } = response.data;
     return data.list;
+  }
+
+  @backgroundMethod()
+  async fetchMarketBannerStockTokenList({ id }: { id: string }) {
+    const client = await this.getClient(EServiceEndpointEnum.Utility);
+    const response = await client.get<{
+      code: number;
+      message: string;
+      data:
+        | IMarketStockPublicItem[]
+        | { list?: IMarketStockPublicItem[]; items?: IMarketStockPublicItem[] };
+    }>(`/utility/v2/market/banner/stock-token-list/${encodeURIComponent(id)}`);
+    const data = response.data.data;
+    return Array.isArray(data) ? data : (data.list ?? data.items ?? []);
   }
 
   @backgroundMethod()
--- a/packages/kit-bg/src/services/ServiceMarketV2.ts
+++ b/packages/kit-bg/src/services/ServiceMarketV2.ts
@@ -1056,53 +1056,62 @@ class ServiceMarketV2 extends ServiceBase {
     }
   }
 
+  private memoizedFetchMarketBannerData = memoizee(
+    async (path: string, locale: string) => {
+      const client = await this.getClient(EServiceEndpointEnum.Utility);
+      const response = await client.get<{
+        data:
+          | IMarketBannerListResponse
+          | IMarketBannerTokenListResponse
+          | IMarketStockPublicItem[]
+          | {
+              list?: IMarketStockPublicItem[];
+              items?: IMarketStockPublicItem[];
+            };
+      }>(path, {
+        params: { currency: 'usd' },
+        headers: { 'x-onekey-request-locale': locale },
+      });
+      return response.data.data;
+    },
+    { maxAge: timerUtils.getTimeDurationMs({ seconds: 5 }), promise: true },
+  );
+
   @backgroundMethod()
   async fetchMarketBannerList(): Promise<IMarketBannerItem[]> {
     const devSettings = await devSettingsPersistAtom.get();
     if (devSettings.enabled && devSettings.settings?.enableMockMarketBanner) {
       return MOCK_MARKET_BANNER_LIST;
     }
-    const client = await this.getClient(EServiceEndpointEnum.Utility);
-    const response = await client.get<{
-      code: number;
-      message: string;
-      data: IMarketBannerListResponse;
-    }>('/utility/v2/market/banner/list');
-    const { data } = response.data;
+    const data = (await this.memoizedFetchMarketBannerData(
+      '/utility/v2/market/banner/list',
+      await this._getMarketTokenBatchCacheLocale(),
+    )) as IMarketBannerListResponse;
     return data.data;
   }
 
   @backgroundMethod()
   async clearMarketBannerCache(): Promise<void> {
-    // Kept for callers that clear market caches after account or network changes.
+    this.memoizedFetchMarketBannerData.clear();
   }
 
   @backgroundMethod()
   async fetchMarketBannerTokenList({ tokenListId }: { tokenListId: string }) {
-    const client = await this.getClient(EServiceEndpointEnum.Utility);
-    const response = await client.get<{
-      code: number;
-      message: string;
-      data: IMarketBannerTokenListResponse;
-    }>(
+    const data = (await this.memoizedFetchMarketBannerData(
       `/utility/v2/market/banner/token-list/${encodeURIComponent(tokenListId)}`,
-      { params: { currency: 'usd' } },
-    );
-    const { data } = response.data;
+      await this._getMarketTokenBatchCacheLocale(),
+    )) as IMarketBannerTokenListResponse;
     return data.list;
   }
 
   @backgroundMethod()
   async fetchMarketBannerStockTokenList({ id }: { id: string }) {
-    const client = await this.getClient(EServiceEndpointEnum.Utility);
-    const response = await client.get<{
-      code: number;
-      message: string;
-      data:
-        | IMarketStockPublicItem[]
-        | { list?: IMarketStockPublicItem[]; items?: IMarketStockPublicItem[] };
-    }>(`/utility/v2/market/banner/stock-token-list/${encodeURIComponent(id)}`);
-    const data = response.data.data;
+    const data = (await this.memoizedFetchMarketBannerData(
+      `/utility/v2/market/banner/stock-token-list/${encodeURIComponent(id)}`,
+      await this._getMarketTokenBatchCacheLocale(),
+    )) as
+      | IMarketStockPublicItem[]
+      | { list?: IMarketStockPublicItem[]; items?: IMarketStockPublicItem[] };
     return Array.isArray(data) ? data : (data.list ?? data.items ?? []);
   }
 
--- a/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
+++ b/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
@@ -73,7 +73,7 @@ function MarketBannerDetailContent({ title }: { title: string }) {
   const { tokenListId, type, assetType } = route.params;
   const isPerps = type === EMarketBannerType.Perps;
   const isMixed = type === EMarketBannerType.StockPerps;
-  const isIndex = isMarketIndexQuoteBanner({ type, assetType, title });
+  const isIndex = isMarketIndexQuoteBanner({ type, assetType });
   const isStock =
     type === EMarketBannerType.Stock ||
     type === EMarketBannerType.StockPerps ||
--- a/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
+++ b/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
@@ -39,6 +39,7 @@ import { useToDetailPage } from '../MarketHomeV2/components/MarketTokenList/hook
 import { MarketTokenListBase } from '../MarketHomeV2/components/MarketTokenList/MarketTokenListBase';
 import { MarketWatchListProviderMirrorV2 } from '../MarketWatchListProviderMirrorV2';
 import { MarketTestIDs } from '../testIDs';
+import { isMarketIndexQuoteBanner } from '../utils/marketBannerUtils';
 
 import { BannerDetailTokenFlatList } from './BannerDetailTokenFlatList';
 import { PerpsTokenListSection } from './PerpsTokenListSection';
@@ -72,20 +73,13 @@ function MarketBannerDetailContent({ title }: { title: string }) {
   const { tokenListId, type, assetType } = route.params;
   const isPerps = type === EMarketBannerType.Perps;
   const isMixed = type === EMarketBannerType.StockPerps;
-  const isIndex =
-    type === EMarketBannerType.Index ||
-    type === EMarketBannerType.StockIndex ||
-    assetType === 'index' ||
-    title.includes('指数');
+  const isIndex = isMarketIndexQuoteBanner({ type, assetType, title });
   const isStock =
     type === EMarketBannerType.Stock ||
-    type === EMarketBannerType.Index ||
     type === EMarketBannerType.StockPerps ||
-    type === EMarketBannerType.StockIndex ||
     assetType === 'stock' ||
     assetType === 'etf' ||
-    assetType === 'index' ||
-    title.includes('指数');
+    assetType === 'index';
 
   const intl = useIntl();
   const toDetailPage = useToDetailPage({ from: EEnterWay.BannerList });
--- a/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
+++ b/packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
@@ -10,6 +10,7 @@ import {
   SizableText,
   Stack,
   XStack,
+  YStack,
   useMedia,
   useSafeAreaInsets,
 } from '@onekeyhq/components';
@@ -68,8 +69,23 @@ const BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS = ['liquidity'] as const;
 
 function MarketBannerDetailContent({ title }: { title: string }) {
   const route = useRoute<IMarketBannerDetailRouteParams>();
-  const { tokenListId, type } = route.params;
+  const { tokenListId, type, assetType } = route.params;
   const isPerps = type === EMarketBannerType.Perps;
+  const isMixed = type === EMarketBannerType.StockPerps;
+  const isIndex =
+    type === EMarketBannerType.Index ||
+    type === EMarketBannerType.StockIndex ||
+    assetType === 'index' ||
+    title.includes('指数');
+  const isStock =
+    type === EMarketBannerType.Stock ||
+    type === EMarketBannerType.Index ||
+    type === EMarketBannerType.StockPerps ||
+    type === EMarketBannerType.StockIndex ||
+    assetType === 'stock' ||
+    assetType === 'etf' ||
+    assetType === 'index' ||
+    title.includes('指数');
 
   const intl = useIntl();
   const toDetailPage = useToDetailPage({ from: EEnterWay.BannerList });
@@ -85,7 +101,7 @@ function MarketBannerDetailContent({ title }: { title: string }) {
     listResult,
     mobileData,
     tickerIsLoading,
-  } = useMarketBannerDetail({ tokenListId, isPerps });
+  } = useMarketBannerDetail({ tokenListId, isPerps, isStock, isIndex });
 
   const renderHeaderLeft = useCallback(
     () => <NavBackButton onPress={handleBackPress} />,
@@ -124,6 +140,8 @@ function MarketBannerDetailContent({ title }: { title: string }) {
     },
     [toDetailPage],
   );
+  const handleIndexItemPress = useCallback(() => undefined, []);
+  const onItemPress = isIndex ? handleIndexItemPress : handleItemPress;
 
   const renderPageHeader = useMemo(() => {
     if (isWebDesktop) {
@@ -201,22 +219,34 @@ function MarketBannerDetailContent({ title }: { title: string }) {
     // Narrow layouts use the compact list to avoid the desktop table's
     // intrinsic width overflowing the viewport.
     if (!gtMd) {
-      return (
+      const stockList = (
         <BannerDetailTokenFlatList
           data={mobileData}
           isLoading={tickerIsLoading}
           changeSortType={changeSortType}
           change24hColumnTitle={change24hColumnTitle}
           onChangeSortPress={handleChangeSortPress}
-          onItemPress={handleItemPress}
+          onItemPress={onItemPress}
         />
+      );
+      if (!isMixed) return stockList;
+      return (
+        <YStack flex={1} gap="$6">
+          {stockList}
+          <PerpsTokenListSection
+            tokenListId={tokenListId}
+            changeSortType={changeSortType}
+            change24hColumnTitle={change24hColumnTitle}
+            onChangeSortPress={handleChangeSortPress}
+          />
+        </YStack>
       );
     }
 
     const tokenList = (
       <MarketTokenListBase
         result={listResult}
-        onItemPress={handleItemPress}
+        onItemPress={onItemPress}
         hideTokenAge
         clientSort
         watchlistFrom={EWatchlistFrom.BannerList}
@@ -226,10 +256,9 @@ function MarketBannerDetailContent({ title }: { title: string }) {
         hiddenDesktopColumns={BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS}
       />
     );
-    if (platformEnv.isNative) {
-      return tokenList;
-    }
-    return (
+    const stockList = platformEnv.isNative ? (
+      tokenList
+    ) : (
       <Stack
         flex={1}
         className="normal-scrollbar"
@@ -240,11 +269,24 @@ function MarketBannerDetailContent({ title }: { title: string }) {
         </Stack>
       </Stack>
     );
+    if (!isMixed) return stockList;
+    return (
+      <YStack flex={1} gap="$6">
+        {stockList}
+        <PerpsTokenListSection
+          tokenListId={tokenListId}
+          changeSortType={changeSortType}
+          change24hColumnTitle={change24hColumnTitle}
+          onChangeSortPress={handleChangeSortPress}
+        />
+      </YStack>
+    );
   }, [
     isPerps,
+    isMixed,
     tokenListId,
     listResult,
-    handleItemPress,
+    onItemPress,
     gtMd,
     tickerIsLoading,
     mobileData,
--- a/packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
+++ b/packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
@@ -5,7 +5,6 @@ import { usePromiseResult } from '@onekeyhq/kit/src/hooks/usePromiseResult';
 import { useMarketBasicConfig } from '@onekeyhq/kit/src/views/Market/hooks';
 import { useMarketBannerListSortAtom } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
 import type {
-  IMarketBannerIndexPreview,
   IMarketStockPublicItem,
   IMarketTokenListItem,
 } from '@onekeyhq/shared/types/marketV2';
@@ -58,25 +57,6 @@ function mapStockBannerItemToToken(
   };
 }
 
-function mapIndexBannerItemToToken(
-  item: IMarketBannerIndexPreview,
-): IMarketTokenListItem {
-  return {
-    address: item.symbol,
-    name: item.name,
-    symbol: item.symbol,
-    decimals: 0,
-    logoUrl: item.logo,
-    price: item.price ?? undefined,
-    priceChange24hPercent: item.priceChange24hPercent ?? undefined,
-    stock: {
-      subtitle: item.name,
-      source: 'index',
-      sourceLogoUri: item.logo,
-    },
-  };
-}
-
 export function useMarketBannerDetail({
   tokenListId,
   isPerps,
@@ -95,12 +75,8 @@ export function useMarketBannerDetail({
   const { result: tickerResult, isLoading: tickerIsLoading } = usePromiseResult(
     async () => {
       if (isPerps) return null;
-      if (isIndex) {
-        const banners =
-          await backgroundApiProxy.serviceMarketV2.fetchMarketBannerList();
-        const banner = banners.find((item) => item.tokenListId === tokenListId);
-        return (banner?.indices ?? []).map(mapIndexBannerItemToToken);
-      }
+      // Index quotes are display-only; restored legacy routes have no tradable rows.
+      if (isIndex) return [];
       if (isStock) {
         const data =
           await backgroundApiProxy.serviceMarketV2.fetchMarketBannerStockTokenList(
--- a/packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
+++ b/packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
@@ -4,6 +4,11 @@ import backgroundApiProxy from '@onekeyhq/kit/src/background/instance/background
 import { usePromiseResult } from '@onekeyhq/kit/src/hooks/usePromiseResult';
 import { useMarketBasicConfig } from '@onekeyhq/kit/src/views/Market/hooks';
 import { useMarketBannerListSortAtom } from '@onekeyhq/kit-bg/src/states/jotai/atoms';
+import type {
+  IMarketBannerIndexPreview,
+  IMarketStockPublicItem,
+  IMarketTokenListItem,
+} from '@onekeyhq/shared/types/marketV2';
 
 import {
   buildMarketNetworkLogoUriMap,
@@ -25,11 +30,58 @@ function isBannerDetailSortBy(
 type IUseMarketBannerDetailParams = {
   tokenListId: string;
   isPerps: boolean;
+  isStock?: boolean;
+  isIndex?: boolean;
 };
+
+function mapStockBannerItemToToken(
+  item: IMarketStockPublicItem,
+): IMarketTokenListItem {
+  return {
+    address: item.stockId,
+    name: item.name,
+    symbol: item.symbol,
+    decimals: 0,
+    logoUrl: item.logoUrl,
+    price: item.price,
+    priceChange24hPercent: item.priceChange24hPercent,
+    marketCap: item.marketCap,
+    stockId: item.stockId,
+    stock: {
+      stockId: item.stockId,
+      subtitle: item.name,
+      source: item.assetType,
+      sourceLogoUri: item.logoUrl,
+      marketCap: item.marketCap,
+      assetAnalysis: { volume24h: item.volume24h },
+    },
+  };
+}
+
+function mapIndexBannerItemToToken(
+  item: IMarketBannerIndexPreview,
+): IMarketTokenListItem {
+  return {
+    address: item.symbol,
+    name: item.name,
+    symbol: item.symbol,
+    decimals: 0,
+    logoUrl: item.logo,
+    price: item.price ?? undefined,
+    priceChange24hPercent: item.priceChange24hPercent ?? undefined,
+    stock: {
+      subtitle: item.name,
+      source: 'index',
+      sourceLogoUri: item.logo,
+    },
+  };
+}
 
 export function useMarketBannerDetail({
   tokenListId,
   isPerps,
+  isStock = false,
+  isIndex = false,
 }: IUseMarketBannerDetailParams) {
   const { networkList } = useMarketBasicConfig();
   const networkLogoUriMap = useMemo(
@@ -43,13 +95,24 @@ export function useMarketBannerDetail({
   const { result: tickerResult, isLoading: tickerIsLoading } = usePromiseResult(
     async () => {
       if (isPerps) return null;
-      const data =
-        await backgroundApiProxy.serviceMarketV2.fetchMarketBannerTokenList({
-          tokenListId,
-        });
-      return data;
+      if (isIndex) {
+        const banners =
+          await backgroundApiProxy.serviceMarketV2.fetchMarketBannerList();
+        const banner = banners.find((item) => item.tokenListId === tokenListId);
+        return (banner?.indices ?? []).map(mapIndexBannerItemToToken);
+      }
+      if (isStock) {
+        const data =
+          await backgroundApiProxy.serviceMarketV2.fetchMarketBannerStockTokenList(
+            { id: tokenListId },
+          );
+        return data.map(mapStockBannerItemToToken);
+      }
+      return backgroundApiProxy.serviceMarketV2.fetchMarketBannerTokenList({
+        tokenListId,
+      });
     },
-    [tokenListId, isPerps],
+    [tokenListId, isPerps, isStock, isIndex],
     {
       watchLoading: true,
     },
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
@@ -21,6 +21,7 @@ import {
   type IMarketBannerTokenPreview,
 } from '@onekeyhq/shared/types/marketV2';
 
+import { isMarketIndexQuoteBanner } from '../../../utils/marketBannerUtils';
 import { MarketTestIDs } from '../../testIDs';
 
 type IMarketBannerItemProps = {
@@ -323,9 +324,7 @@ function BannerTokenRow({ token }: { token: IMarketBannerTokenPreview }) {
 function MarketBannerItemComponent(props: IMarketBannerItemProps) {
   const { item, onPress } = props;
   const { isSmallScreen } = props;
-  const isIndexBanner =
-    item.type === EMarketBannerType.Index ||
-    item.type === EMarketBannerType.StockIndex;
+  const isIndexBanner = isMarketIndexQuoteBanner(item);
   const tokens = useMemo(() => {
     const bannerTokens = item.tokens ?? [];
     if (isIndexBanner) {
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
@@ -99,6 +99,7 @@ function LegacyMarketBannerItem({
 }: IMarketBannerItemProps) {
   const { title, description, backgroundColor, tokenLogos } = item;
   const isPerps = item.type === EMarketBannerType.Perps;
+  const isIndexBanner = isMarketIndexQuoteBanner(item);
   const bgColor = convertThemeToken(backgroundColor, '$bgSubdued');
   const descriptionColor = convertThemeToken(
     description?.fontColor ?? '',
@@ -119,13 +120,13 @@ function LegacyMarketBannerItem({
       width="$32"
       alignItems="flex-start"
       justifyContent="space-between"
-      onPress={handlePress}
+      onPress={isIndexBanner ? undefined : handlePress}
       transition="quick"
       animateOnly={ANIMATE_ONLY_BORDER_COLOR}
       borderWidth={StyleSheet?.hairlineWidth ?? 1}
       borderColor="$neutral3"
-      hoverStyle={{ borderColor: '$neutral4' }}
-      pressStyle={{ borderColor: '$neutral5' }}
+      hoverStyle={isIndexBanner ? undefined : { borderColor: '$neutral4' }}
+      pressStyle={isIndexBanner ? undefined : { borderColor: '$neutral5' }}
       h={118}
       userSelect="none"
       $gtMd={{
@@ -362,8 +363,8 @@ function MarketBannerItemComponent(props: IMarketBannerItemProps) {
   return (
     <YStack
       testID={MarketTestIDs.bannerItem}
-      onPress={handlePress}
-      role="button"
+      onPress={isIndexBanner ? undefined : handlePress}
+      role={isIndexBanner ? undefined : 'button'}
       aria-label={item.title}
       bg={convertThemeToken(item.backgroundColor, '$bgSubdued')}
       borderRadius="$3"
@@ -373,8 +374,8 @@ function MarketBannerItemComponent(props: IMarketBannerItemProps) {
       flexShrink={isSmallScreen ? 1 : 0}
       gap="$5"
       userSelect="none"
-      hoverStyle={{ opacity: 0.8 }}
-      pressStyle={{ opacity: 0.6 }}
+      hoverStyle={isIndexBanner ? undefined : { opacity: 0.8 }}
+      pressStyle={isIndexBanner ? undefined : { opacity: 0.6 }}
     >
       <XStack
         alignItems="center"
--- a/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
@@ -1,5 +1,7 @@
 import { memo, useCallback, useMemo } from 'react';
 
+// cspell:ignore GSPC IXIC DJIA
+
 import { StyleSheet } from 'react-native';
 
 import {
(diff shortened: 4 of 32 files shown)


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1da6827c-d771-5439-89dc-65a17f36bdc5)

sidmorizon:
Claude review session: https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -1102,6 +1092,20 @@ class ServiceMarketV2 extends ServiceBase {
...
+    const client = await this.getClient(EServiceEndpointEnum.Utility);
+    const response = await client.get<{
+      code: number;
+      message: string;
+      data:
+        | IMarketStockPublicItem[]
+        | { list?: IMarketStockPublicItem[]; items?: IMarketStockPublicItem[] };
+    }>(`/utility/v2/market/banner/stock-token-list/${encodeURIComponent(id)}`);
+    const data = response.data.data;
+    return Array.isArray(data) ? data : (data.list ?? data.items ?? []);
```
sidmorizon, line 1119: 🟡 P2: Stock banners fail in travel mode

Severity: `non-severe`

`fetchMarketBannerStockTokenList` is rejected for every stock banner in travel mode. [`TRAVEL_MODE_ALLOWED_METHODS`](packages/kit-bg/src/apis/travelModeCommandPolicy.ts:82-92) omits it while allowing sibling banner fetches. Cards lose live quotes, and stock banner details fail to load.

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -1102,6 +1092,20 @@ class ServiceMarketV2 extends ServiceBase {
     return data.list;
   }
 
+  @backgroundMethod()
+  async fetchMarketBannerStockTokenList({ id }: { id: string }) {
```
sidmorizon, line 1112: # 🟠 [P1] New bg method missing from Travel Mode allowlist

## Problem

`fetchMarketBannerStockTokenList` is added here as a new `@backgroundMethod()`, but it was not added to `TRAVEL_MODE_ALLOWED_METHODS.serviceMarketV2` in `packages/kit-bg/src/apis/travelModeCommandPolicy.ts:82-92` — which already lists its siblings `fetchMarketBannerList`, `fetchMarketBannerTokenList` and `fetchMarketBannerPerpsTokenList`.

`shouldRejectTravelModeServiceCall` (`travelModeCommandPolicy.ts:170-183`) returns `true` for any method not in that set, so `TravelModeCommandDispatcher.runServiceCall` (`TravelModeCommandDispatcher.ts:54-57`) routes the call through `environment.commands.run(operation)`. When Travel Mode is active the command capability is `control-plane-only`, so `RuntimeCommandCapability.run` short-circuits to `rejectTravelModeUnknownError()` (`packages/shared/src/travelMode/runtimeEnvironment.ts:147-165`) and the call never reaches the network.

The allowlist is a hand-maintained table mirrored by a hand-maintained test case list (`travelModeCommandPolicy.test.ts`), so an omission produces no type error and no test failure.

## Impact

With Travel Mode enabled — a state where market browsing is explicitly meant to keep working, hence the sibling methods already being allowlisted — this affects every platform that routes banner data through the background API (the `.web-only.ts` light-API path is unaffected):

- Market home: the new hydration call for `stock` / `stock_perps` banners rejects. It is swallowed by the `catch` in `useMarketBannerList.ts:80`, so those cards silently lose live quotes and fall back to the un-hydrated banner payload.
- Banner detail: the stock branch at `useMarketBannerDetail/index.ts:104-110` has no `try`/`catch`, so the rejection leaves `tickerResult` undefined and the detail list renders empty with no error surfaced.

## Recommendation

Add `'fetchMarketBannerStockTokenList'` to `TRAVEL_MODE_ALLOWED_METHODS.serviceMarketV2` in `packages/kit-bg/src/apis/travelModeCommandPolicy.ts:82`, and add the matching entry to the case table in `travelModeCommandPolicy.test.ts` so the pairing is covered.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -6,9 +6,85 @@ import { useDevSettingsPersistAtom } from '@onekeyhq/kit-bg/src/states/jotai/ato
...
+async function fetchMarketBannerListWithLiveTokens({
+  enableMockMarketBanner,
+}: {
+  enableMockMarketBanner?: boolean;
+}): Promise<IMarketBannerItem[]> {
+  const banners = await fetchMarketBannerListForPlatform({
+    enableMockMarketBanner,
+  });
+
+  const hydratedBanners = await Promise.all(
```
sidmorizon, line 29: # 🟡 [P2] Banner polling becomes N+1 uncached requests per tick

## Problem

This `Promise.all` fans out one extra HTTP request per banner (`fetchMarketBannerTokenListForPlatform` / `fetchMarketBannerStockTokenListForPlatform`). At the same time this PR removed the only cache that covered the banner path:

- `memoizedFetchMarketBannerList` (30s `maxAge`, `promise: true`) was deleted from `ServiceMarketV2.ts`; `fetchMarketBannerList` now hits the endpoint directly on every call.
- The two new helpers are plain async functions with no memoization — `marketLightApi.ts:523` and `marketLightApi.ts:535` — and the new `ServiceMarketV2.fetchMarketBannerStockTokenList` has no `memoizee` either. Compare `fetchMarketBannerListLight` (`marketLightApi.ts:518`), which is memoized.

The consuming hook runs with `pollingInterval: 30s` plus `revalidateOnReconnect: true` and `revalidateOnFocus: true` (`useMarketBannerList.ts:126-131`).

Separately, `clearMarketBannerCache` is now an empty method body, but it still has a live caller — the Dev Settings "clear banner cache" action at `packages/kit/src/views/Setting/pages/Tab/DevSettingsSection/index.tsx:1956` — so that button now silently does nothing. Its replacement comment ("Kept for callers that clear market caches after account or network changes") describes callers that do not exist in the tree.

## Impact

With N banners, the Market tab issues `1 + N` uncached requests every 30s while focused, plus a full fan-out on every focus and reconnect, where it previously issued one memoized request. The token-list endpoint returns the full list while the card renders only `.slice(0, 3)`, so payload volume grows alongside request count. On iOS/Android/extension each of those is also a `main`→`bg` bridge round-trip rather than a plain fetch.

## Recommendation

Wrap `fetchMarketBannerTokenListLight` and `fetchMarketBannerStockTokenListLight` (`marketLightApi.ts:523,535`) and the corresponding `ServiceMarketV2` methods in `memoizee({ maxAge: timerUtils.getTimeDurationMs({ seconds: 30 }), promise: true })`, mirroring `fetchMarketBannerListLight`; or restore the list-level memoizee so one cached call covers the fan-out. Either re-wire `clearMarketBannerCache` to clear the new caches, or delete it together with its Dev Settings caller rather than leaving a no-op behind a button.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/utils/marketBannerUtils.ts
```diff
@@ -0,0 +1,17 @@
...
+  title,
+}: {
+  type?: EMarketBannerType;
+  assetType?: string;
+  title?: string;
+}) {
+  return (
+    type === EMarketBannerType.Index ||
+    type === EMarketBannerType.StockIndex ||
+    (type === undefined && assetType === undefined && title?.includes('指数'))
```
sidmorizon: # 🟡 [P2] Index detection keyed on a hardcoded Chinese title

## Problem

The fallback branch classifies a banner as an index banner via `title?.includes('指数')`. Banner titles are server-supplied and localized: `useMarketBannerList` derives its request scope from `useLocaleVariant()` (`useMarketBannerList.ts:95,105`) and re-fetches on locale change, so the same banner returns a different `title` per language.

This helper is not cosmetic — it selects both layout and data source:

- `MarketBannerItem.tsx:328` uses it to pick the index column layout, to suppress the chevron, and (as of `3c0ddc78`) to disable the card's `onPress` entirely.
- `MarketBannerDetail.tsx:76` re-derives it from the route `title` and passes it as `isIndex` into `useMarketBannerDetail`, which switches the data source to `banner.indices` (`useMarketBannerDetail/index.ts:98-103`).

The function also returns `boolean | undefined` rather than `boolean`, because `title?.includes(...)` propagates `undefined`.

## Impact

For an untyped index banner (`type === undefined && assetType === undefined`) the app behaves differently purely by UI language. A `zh-CN` user gets the index layout and a non-clickable card, and the detail screen sources rows from `banner.indices`; a `en-US` user gets the crypto row layout, a clickable card with chevron, and a detail screen sourced from `fetchMarketBannerTokenList`. Since untyped banners are exactly the ones that generally carry no `indices` array, the `zh-CN` path also renders an empty detail list. The fallback the change adds only ever fires for Chinese locales.

## Recommendation

Classify from the payload rather than the localized string — e.g. `Boolean(banner.indices?.length)` as the fallback on the card — and stop re-deriving the flag from `title` on the detail screen: pass the resolved `isIndex` (or the already-resolved banner `type`) through the `MarketBannerDetail` route params alongside the `assetType` this PR already adds in `useToMarketBannerDetail.ts:24`. Also tighten the return type to `boolean` (`Boolean(...)` / `?? false`).

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
```diff
@@ -275,26 +363,26 @@ function MarketBannerItemComponent(props: IMarketBannerItemProps) {
...
+      onPress={isIndexBanner ? undefined : handlePress}
+      role={isIndexBanner ? undefined : 'button'}
+      aria-label={item.title}
       bg={convertThemeToken(item.backgroundColor, '$bgSubdued')}
       borderRadius="$3"
       px="$4"
       py="$5"
-      width={336}
-      flexShrink={0}
+      width={isSmallScreen ? '100%' : 336}
```
sidmorizon: # 🟡 [P2] Card width switches to 100% but skeleton stays 336

## Problem

Banner cards now size as `width={isSmallScreen ? '100%' : 336}` / `flexShrink={isSmallScreen ? 1 : 0}`, but the loading placeholder was not updated: `MarketBannerItemSkeleton` still hardcodes `width={336}` / `flexShrink={0}` (`MarketBannerItemSkeleton.tsx:15-16`).

`MarketBannerListSkeletonComponent` already receives `isSmallScreen` and uses it to choose the container and the skeleton count (`MarketBannerList.tsx:96-111`), but it does not forward it to `MarketBannerItemSkeleton`. Both the skeleton and the real cards render into the same `BannerContainerMobile` horizontal `ScrollView`, which this PR also pins to a fixed `h={MARKET_BANNER_LIST_MOBILE_HEIGHT}` (`MarketBannerList.tsx:51`).

## Impact

On phones and narrow web windows the initial load renders 336px-wide skeleton cards, then re-lays out at a different card width once `isFetched` flips and the real cards commit — a visible width and horizontal-scroll-extent jump on every cold load of the Market tab, which is the layout jitter this PR otherwise sets out to remove. Additionally, a percentage width has no definite containing width inside a horizontal `ScrollView` content container on native Yoga, so `'100%'` is not a reliable drop-in for the fixed `336` there.

## Recommendation

Thread `isSmallScreen` from `MarketBannerListSkeletonComponent` (`MarketBannerList.tsx:103`) into `MarketBannerItemSkeleton` and apply the identical `width` / `flexShrink` rule so placeholder and card agree. On native, prefer an explicit measured width (viewport width minus the container's `$4` horizontal padding) over `'100%'` inside the horizontal scroller.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit-bg/src/services/ServiceMarketV2.ts
sidmorizon: Correction to the Recommendation above: the method allowlist is only the first of two gates, so adding `'fetchMarketBannerStockTokenList'` alone is **not** sufficient.

`packages/shared/src/travelMode/runtimeNetworkAllowlist.ts:38-40` registers `/utility/v2/market/banner/list`, `/utility/v2/market/banner/token-list` and `/utility/v2/market/banner/perps-token-list` as allowed request prefixes (the dynamic-suffix match is at `runtimeNetworkAllowlist.ts:119-123`), but has no entry for `/utility/v2/market/banner/stock-token-list`. Even with the background method allowlisted, the request would still be rejected at the network gate.

Both need registering, with matching cases in `travelModeCommandPolicy.test.ts` and `runtimeNetworkAllowlist.test.ts:84`.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit-bg/src/services/ServiceMarketV2.ts
zy2000-star: fixed in c3c9c75c23. Registered the stock-banner read method and the exact stock-token-list URL prefix in both Travel Mode policy layers. Added allowed-path and rejected-extra-suffix tests.

packages/kit-bg/src/services/ServiceMarketV2.ts
```diff
@@ -1102,6 +1092,20 @@ class ServiceMarketV2 extends ServiceBase {
...
+    const client = await this.getClient(EServiceEndpointEnum.Utility);
+    const response = await client.get<{
+      code: number;
+      message: string;
+      data:
+        | IMarketStockPublicItem[]
+        | { list?: IMarketStockPublicItem[]; items?: IMarketStockPublicItem[] };
+    }>(`/utility/v2/market/banner/stock-token-list/${encodeURIComponent(id)}`);
```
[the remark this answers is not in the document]
zy2000-star: fixed in c3c9c75c23. Registered both the background read method and the dynamic stock-token-list network prefix; policy tests pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -6,9 +6,85 @@ import { useDevSettingsPersistAtom } from '@onekeyhq/kit-bg/src/states/jotai/ato
...
+    banners.map(async (banner) => {
```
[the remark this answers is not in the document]
zy2000-star: fixed in c3c9c75c23. Base banners now commit independently of optional quote hydration. Native page readiness depends only on the base result. Late hydration is guarded by source response and locale scope; a pending quote request no longer blocks the page.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -6,9 +6,85 @@ import { useDevSettingsPersistAtom } from '@onekeyhq/kit-bg/src/states/jotai/ato
...
+} from './marketBannerListPlatformApi';
+
+async function fetchMarketBannerListWithLiveTokens({
+  enableMockMarketBanner,
+}: {
+  enableMockMarketBanner?: boolean;
+}): Promise<IMarketBannerItem[]> {
+  const banners = await fetchMarketBannerListForPlatform({
+    enableMockMarketBanner,
+  });
```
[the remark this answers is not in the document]
zy2000-star: fixed in c3c9c75c23. Mock mode skips optional remote hydration. Added a regression test asserting neither token nor stock quote requests are made.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
```diff
@@ -25,11 +30,58 @@ function isBannerDetailSortBy(
...
+    name: item.name,
+    symbol: item.symbol,
+    decimals: 0,
+    logoUrl: item.logo,
+    price: item.price ?? undefined,
+    priceChange24hPercent: item.priceChange24hPercent ?? undefined,
+    stock: {
+      subtitle: item.name,
+      source: 'index',
+      sourceLogoUri: item.logo,
```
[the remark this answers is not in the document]
zy2000-star: fixed in c3c9c75c23. Index cards remain non-interactive, navigation also guards index banners, and legacy index detail routes no longer produce token/watchlist rows with invalid identities.

packages/kit-bg/src/services/ServiceMarketV2.ts
zy2000-star: fixed in c3c9c75c23. Both Travel Mode gates now allow the public stock banner read: background method and dynamic URL prefix, with positive and negative path tests.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed in c3c9c75c23. Added locale-keyed promise caches for base/token/stock requests on background and web paths, using a 5-second TTL to deduplicate bursts without extending the 30-second quote refresh interval. clearMarketBannerCache now clears the background caches; tested concurrent deduplication and immediate fresh data after clearing. Periodic per-list refresh remains intentional for live quotes.

packages/kit/src/views/Market/utils/marketBannerUtils.ts
zy2000-star: fixed in c3c9c75c23. Removed localized-title classification. The helper returns boolean and uses explicit types or an untyped indices payload. Untyped index payloads are normalized before rendering and navigation rejects indices. English and Chinese cases are covered.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerItem.tsx
zy2000-star: fixed in c3c9c75c23. Restored the explicit 336px non-shrinking carousel width shared by cards and skeletons, avoiding percentage widths in native horizontal content. Kept the overflow height fix. Narrow-web runtime verification measured both skeleton and card at 336px; index clicks remain on /market.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -55,9 +125,39 @@ export function useMarketBannerList(): {
...
+        source: bannerList,
+        scope: requestScope,
+        banners: await hydrateMarketBannerQuotes(bannerList),
+      };
+    },
+    [bannerList, enableMockMarketBanner, requestScope],
+    { checkIsFocused: !platformEnv.isWeb },
+  );
+  const displayedBanners =
+    liveQuotes?.source === bannerList && liveQuotes?.scope === requestScope
```
sidmorizon: # 🟠 [P1] Hydrated quotes discarded on every 30s refetch

## Problem

`displayedBanners` gates the hydrated result on **reference identity** of the base array:

```ts
liveQuotes?.source === bannerList && liveQuotes?.scope === requestScope
```

The base `usePromiseResult` still runs with `pollingInterval: 30s`, `revalidateOnFocus: true` and `revalidateOnReconnect: true` (lines 122-124). Every one of those runs commits a **freshly allocated** array:

- iOS/Android/extension: `fetchMarketBannerListForPlatform` goes through the `@backgroundMethod()` proxy, so the payload is deserialized into the `main` runtime's heap as a new object on each call.
- Web: `fetchMarketBannerListCached` has a 5s `maxAge`, long expired at the 30s tick, so a new fetch allocates a new array.

So `liveQuotes.source !== bannerList` after every refetch, and `displayedBanners` falls back to the raw payload for the duration of the hydration round-trip, every time. The `scope` guard on the same line already covers the locale / mock-mode invalidation this is meant to protect against; the identity check adds only over-invalidation.

Before this commit hydration happened inside the same promise, so `bannerList` was never un-hydrated after first load — this is a regression introduced by the base/hydration split.

## Impact

The raw `/banner/list` payload does not carry `tokens` for most banners — in `MOCK_MARKET_BANNER_LIST`, which mirrors the production shape, only 1 of 6 entries has `tokens`; the rest carry only `tokenLogos`. `MarketBannerItem.tsx:360` then takes `if (!item.tokens) return <LegacyMarketBannerItem {...props} />`.

Concrete sequence on desktop (`gtMd`): at t=0 hydration lands and six full 336px cards render. At t=30s the poll resolves, `displayedBanners` reverts to raw, and those cards collapse to `LegacyMarketBannerItem` (`width="$32"`, `$gtMd maxWidth: 256`) until hydration re-lands. This repeats every 30s and additionally on every tab focus and reconnect. Where the server does populate `tokens`, the symptom is milder but still visible — prices snap back to the list payload's values and forward again on each tick.

## Recommendation

Keep the scope guard, drop the array-identity guard, and carry the previously hydrated quotes across a same-scope refetch by merging on `_id`. Merge only the quote field so base fields stay authoritative (this keeps the existing "ignores quotes completed for a previous base response" test passing, which asserts the newer base `title` wins):

```ts
const displayedBanners = useMemo(() => {
  if (!bannerList || liveQuotes?.scope !== requestScope) return bannerList;
  const hydratedTokens = new Map(
    liveQuotes.banners.map((b) => [b._id, b.tokens]),
  );
  return bannerList.map((b) =>
    hydratedTokens.has(b._id)
      ? { ...b, tokens: hydratedTokens.get(b._id) ?? b.tokens }
      : b,
  );
}, [bannerList, liveQuotes, requestScope]);
```

This keeps the time-to-first-paint win the split was for (`MarketHomeV2` gates the native page on the base result) without the recurring un-hydrated frame.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -55,9 +125,39 @@ export function useMarketBannerList(): {
     },
   );
 
+  const { result: liveQuotes } = usePromiseResult(
```
sidmorizon, line 180: # 🟡 [P2] Native SWR cache now persists un-hydrated banners

## Problem

Splitting hydration into this second `usePromiseResult` moved it out from under the SWR cache. `swrKey` stayed on the **base** hook (line 116-118), and this hydration hook has none, so nothing persists the hydrated output.

`usePromiseResult` persists whatever the keyed runner returns — `swrCacheUtils.set(capturedSwrKey, r)` — and reads it back synchronously on the next mount as `effectiveInitResult`, which explicitly takes priority over `initResult`. Before this commit `r` was `fetchMarketBannerListWithLiveTokens(...)`, i.e. the list **with** `tokens`. After the split `r` is the raw `fetchMarketBannerListForPlatform(...)` payload.

## Impact

On iOS/Android cold start (`swrKey` is `platformEnv.isNative`-only), the MMKV entry is read synchronously, so `bannerList` is defined before any network call, `isFetched` is true and `isLoading` is false. `MarketHomeV2` therefore skips its loading fallback and `MarketBannerList` commits `initialBannerList` immediately — from a payload with no `tokens`. Since `MarketBannerItem.tsx:360` renders `LegacyMarketBannerItem` whenever `tokens` is absent, the first frame shows narrow legacy cards inside the 204pt strip, and only swaps to the real cards after the base fetch plus N token-list fetches complete.

The cache that exists to make the first frame correct now guarantees it is stale in shape, not just in price. This is permanent from the first launch after the update, not a one-time migration effect. Note this is not fixed by repairing the `displayedBanners` guard, because the value handed to `swrCacheUtils.set` is chosen before that merge.

## Recommendation

Persist what is actually rendered. Either move `swrKey` (and `swrShouldPersist`, if wanted) off the base hook onto this hydration hook so the cached value is the hydrated list, or keep it where it is and write the merged result explicitly once hydration matches the current scope:

```ts
useEffect(() => {
  if (!platformEnv.isNative || liveQuotes?.scope !== requestScope) return;
  swrCacheUtils.set(
    swrKeys.marketHomeBanners(locale, Boolean(enableMockMarketBanner)),
    displayedBanners,
  );
}, [displayedBanners, liveQuotes, requestScope, locale, enableMockMarketBanner]);
```

Confidence note: the mechanism is certain from the code; how visible the first frame is depends on whether the deployed `/banner/list` returns `tokens` — worst case the legacy-card swap above, best case a stale→live price snap.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.test.tsx
```diff
@@ -97,3 +142,124 @@ it('waits for successful banner data to commit before releasing the native layou
...
+it('skips remote quotes in mock mode', async () => {
+  mockEnabled = true;
+  mockResult = [makeBanner({ type: EMarketBannerType.Stock })];
+  renderHook(() => useMarketBannerList());
+  await mockHydrate();
+  expect(fetchMarketBannerStockTokenListForPlatform).not.toHaveBeenCalled();
+  expect(fetchMarketBannerTokenListForPlatform).not.toHaveBeenCalled();
+});
+
+it('ignores quotes completed for a previous base response', async () => {
```
sidmorizon, line 260: # 🟡 [P2] No test proves hydrated quotes ever reach bannerList

## Problem

Of the hook-level tests that exercise `mockHydrate`, all three assert a **negative**:

- line 236 — `expect(result.current.bannerList).toEqual(banners)`, i.e. the *raw* banners while quotes are pending
- line 247 (this test) — hydration for a previous base response is discarded
- line 257 — hydration for a previous locale is discarded

The two places that did run hydration and assert its output (lines 165 and 198) now call `hydrateMarketBannerQuotes(...)` directly; the `renderHook(() => useMarketBannerList())` above each is dead code that no longer feeds the assertion.

So no test asserts that `liveQuotes.banners` ever reaches `bannerList`. The positive path of the feature this commit restructured is unasserted — which is precisely why the `displayedBanners` identity guard regression is invisible to CI.

## Impact

Any future edit to the `displayedBanners` predicate can disable live banner quotes outright and the suite stays green. Concretely, replacing that expression with a bare `bannerList` — i.e. deleting the feature — passes all 23 tests today.

## Recommendation

Add a positive case, plus the regression case for the recurring-refetch bug, using the existing harness:

```ts
it('surfaces hydrated quotes for the current base response', async () => {
  mockResult = [makeBanner({ type: EMarketBannerType.Stock })];
  const { result, rerender } = renderHook(() => useMarketBannerList());
  mockLiveResult = await mockHydrate();
  rerender();
  expect(result.current.bannerList[0].tokens?.[0].price).toBe('100');
});

it('keeps hydrated quotes across a same-scope refetch', async () => {
  mockResult = [makeBanner({ _id: 'a', type: EMarketBannerType.Stock })];
  const { result, rerender } = renderHook(() => useMarketBannerList());
  mockLiveResult = await mockHydrate();
  rerender();
  // same content, newly allocated array — what every 30s poll produces
  mockResult = [makeBanner({ _id: 'a', type: EMarketBannerType.Stock })];
  rerender();
  expect(result.current.bannerList[0].tokens).toBeDefined();
});
```

The second one fails at `c3c9c75c` and is the guard for the identity-check issue.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed: Merge quote fields by banner ID and matching tokenListId/type/assetType while keeping current base metadata authoritative. Same-scope refreshes retain quotes, failed hydration retains the last displayed prices, and Perps/mock/index payloads keep their fresh base data. Added positive hydration, refresh/failure/recovery, identity and locale isolation tests.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed: Native SWR now persists the rendered banner snapshot including hydrated quotes. Disabled raw runner persistence so base refreshes cannot overwrite the hydrated snapshot. Added a real usePromiseResult hook test with serialized cache replay: after unmount/remount, quotes render synchronously and remain present while the fresh base response arrives and hydration stays pending.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.test.tsx
zy2000-star: fixed: Added hook-level assertions that fetched price 100 reaches bannerList, survives a newly allocated base response and failed hydration, then updates to 105 after success. Added native serialized cache replay and changed identity/locale regression coverage. All 50 banner tests and the commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -1,14 +1,108 @@
...
+      if (banner.type === EMarketBannerType.Perps) {
+        return banner;
+      }
+
+      try {
+        if (isStockBanner || banner.type === EMarketBannerType.Stock) {
+          const assets = await fetchMarketBannerStockTokenListForPlatform(
+            banner.tokenListId,
+          );
+          if (!assets.length) return banner;
```
sidmorizon: 🟡 P2: Cleared quote lists retain rows

Severity: `non-severe`

When a quote endpoint returns an empty list, `hydrateMarketBannerQuotes` retains cached tokens. Removed banner assets remain visible through later refreshes.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -1,14 +1,108 @@
...
+  banners: IMarketBannerItem[],
+  previous: IMarketBannerItem[] | undefined,
+): IMarketBannerItem[] {
+  const quotes = new Map(previous?.map((banner) => [banner._id, banner]));
+  return banners.map((banner) => {
+    const cached = quotes.get(banner._id);
+    if (
+      banner.type !== EMarketBannerType.Perps &&
+      !isMarketIndexQuoteBanner(banner) &&
+      cached?.tokens &&
```
sidmorizon, line 98: # 🟡 [P2] Cached quotes override fresher payload quotes

## Problem

The merge condition checks `cached?.tokens` but never checks whether the **incoming** banner already carries its own `tokens`, and the branch then does an unconditional `return { ...banner, tokens: cached.tokens }`. So when the fresh `/banner/list` response ships inline quotes, they are discarded in favour of the older cached array.

`tokens` is a real payload field on `IMarketBannerItem` (`packages/shared/types/marketV2.ts`), and the list endpoint does populate it for some banners — the `EMarketBannerType.Ticker` entry in `packages/kit-bg/src/services/ServiceMarketV2.const.ts` ships inline `tokens`, while the other mock entries carry only `tokenLogos`. So this is not a hypothetical shape.

This is distinct from the retention issues already reported on this file (empty-response retention at line 48, cross-scope tagging at line 202): those are about *keeping* a cached value, this one is about the cache **beating a newer value the client already holds**. It also has a different one-line repair.

## Impact

Take a Ticker banner with cached tokens `[BTC $60000]` from the last successful hydration. The next list response returns the same `_id` / `tokenListId` / `type` / `assetType` with fresh inline `tokens: [BTC $65000]`. `mergeBannerQuotes` replaces `$65000` with `$60000`.

In the normal case hydration completes shortly after and overwrites it, so the window is brief. The bite is when hydration then fails or returns empty: the card renders and persists `$60000` even though a newer `$65000` arrived in the very same response and was thrown away. Combined with the unbounded retention already reported, that stale value can survive cold starts.

No current test covers it — `uses fresh base quotes when optional hydration is skipped` only exercises the Perps and mock-mode bypasses, and `does not reuse quotes for changed identity` only exercises identity *mismatches*, never a match where the fresh banner also has `tokens`.

## Recommendation

Let the cache fill a gap rather than override delivered data — add a freshness check to the condition:

```ts
if (
  banner.type !== EMarketBannerType.Perps &&
  !isMarketIndexQuoteBanner(banner) &&
  !banner.tokens?.length &&   // only fill when the payload has no quotes
  cached?.tokens &&
  cached.tokenListId === banner.tokenListId &&
  cached.type === banner.type &&
  cached.assetType === banner.assetType
) {
```

Worth a test that asserts a fresh payload `tokens` value wins over a cached one for the same identity.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -58,18 +152,70 @@ export function useMarketBannerList(): {
...
+      isMarketIndexQuoteBanner(banner) && banner.indices?.length
+        ? {
+            ...banner,
+            type: EMarketBannerType.StockIndex,
+            tokens: banner.indices,
+          }
+        : banner,
+    );
+  }, [bannerList, liveQuotes, requestScope, enableMockMarketBanner]);
+  committedResultRef.current = {
```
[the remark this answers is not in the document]
zy2000-star: fixed: Network banner results now carry the locale/mock scope that produced them. Retained results from another scope are excluded from rendering, hydration and the committed fallback, so a failed request after switching settings cannot be relabeled as a successful current-scope result. Native SWR array snapshots retain their existing format and are accepted only from the scoped native cache key. Added real usePromiseResult tests for locale and mock changes with a pending request, failure, and reconnect recovery. All 52 banner tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed: Successful empty stock/token quote responses now produce tokens: [] instead of retaining cached rows. Error fallback still preserves the last displayed data. Regression tests cover both endpoints and prove removed rows remain cleared through subsequent base refreshes and failed quote requests. All 56 banner tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed: Cached quotes now fill only an absent tokens field; explicit fresh inline rows and explicit empty arrays win over prior responses. Hydration may override inline quotes only when it belongs to the current base response. Same-scope refreshes still retain the committed displayed quotes when the field is absent. Added tests for fresh price 105 and an empty payload beating cached price 100, surviving hydration failure and a later missing-field refresh, then accepting newly hydrated price 110. All 56 banner tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -33,43 +127,115 @@ export function useMarketBannerList(): {
...
       pollingInterval: timerUtils.getTimeDurationMs({ seconds: 30 }),
     },
   );
 
-  committedResultRef.current = { requestScope, bannerList };
+  // Native SWR replays arrays from a locale/mode-specific key. Network
+  // results carry their producing scope because hooks without a cache key retain old data.
+  let bannerList: IMarketBannerItem[] | undefined;
+  if (Array.isArray(scopedResult)) {
+    if (platformEnv.isNative) bannerList = scopedResult;
```
sidmorizon, line 175: # 🟠 [P1] Native SWR replay path skips the scope check

## Problem

The scope-tagging fix closes the cross-scope hole on web/desktop/extension, but the native branch here accepts the replayed array with **no scope comparison at all** — unlike the object branch immediately below it, which does check `scopedResult?.scope === scope`.

That matters because of how `usePromiseResult` handles a `swrKey` change. On native `swrKey` is `swrKeys.marketHomeBanners(locale, mock)`, so it changes with scope, and the hook adjusts state *during render*: `if (swrKey !== prevSwrKey) { setPrevSwrKey(...); setResult(swrCacheEntry?.data ?? initResult) }`. React discards that pass's output but **not** its ref mutations. In that discarded pass `scope` is already the new value while `scopedResult` is still the previous scope's replayed plain array — so `bannerList` is assigned from it, and the write at line 220 stamps `committedResultRef.current = { requestScope: <new>, bannerList: <previous scope's banners> }`.

Previously that self-healed, because the commit was unconditional (`bannerList: bannerList === undefined ? undefined : normalizedBanners`). The `if (bannerList !== undefined)` gate added at line 220 means pass 2 now **skips** the write, so the mis-stamp from the discarded pass survives.

## Impact

iOS/Android, `main` runtime, shared MMKV. Requires a conjunction, but each part is ordinary:

1. SWR holds `marketHomeBanners:…:en-US:live`; nothing yet under `…:zh-CN:live` (first time in that locale).
2. Provider mounts, render-time replay sets `result` to the en-US **array**; the banner fetch is still in flight.
3. `locale` flips to `zh-CN` — a user switch, or `useLocaleVariant`'s persisted atom hydrating after mount.
4. Discarded pass: `bannerList = [bannerEN]` → `committedResultRef = { zh-CN, [bannerEN] }`. Pass 2 has `bannerList === undefined`, so the gate skips and the poison stays.
5. The zh-CN request fails (offline/timeout). The catch at line 146 finds `previous.requestScope === requestScope` and returns `{ scope: 'zh-CN:…', banners: [bannerEN] }`.
6. Line 174 accepts it → English banners render as zh-CN with `isFetched: true`, **and** the effect at line 227 writes them to `swrCacheUtils.set('…:zh-CN:live', …)`.

Step 6 is worse than the pre-commit behaviour: the wrong-locale banners are now persisted under the new scope's key and replay on every cold start until a zh-CN fetch succeeds. Same applies to the dev mock-mode toggle, which is also part of the key.

Note the new tests can't catch it — `useMarketBannerList.recovery.test.tsx:23` sets `platformEnv.isNative = false`, so both new scope-change cases exercise only the web path. (Line 147 flips it to `true`, but that's the pre-existing cache-replay test, not a scope change.)

## Recommendation

Two parts:

1. Scope-check the native array too. The replayed array carries no tag, so gate it on the scope being settled — e.g. mirror the scope in state (`const [prevScope, setPrevScope] = useState(scope)`) and treat `Array.isArray(scopedResult)` as foreign while `scope !== prevScope`, which is exactly the discarded pass.
2. Restore self-healing at line 220 so a stale same-scope entry is cleared rather than left:

```ts
if (bannerList !== undefined) {
  committedResultRef.current = { requestScope, bannerList: normalizedBanners };
} else if (committedResultRef.current?.requestScope === requestScope) {
  committedResultRef.current = undefined;
}
```

Worth a native variant of the new `it.each`: `platformEnv.isNative = true`, seed the cache for `en-US` only, switch locale with the first request pending, then fail it.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.test.tsx
```diff
@@ -1,39 +1,93 @@
...
+      return { result: mockLiveResult };
+    }
     mockRequest = request;
-    return { result: mockResult, isLoading: mockLoading };
+    return {
+      result:
+        mockResult === undefined
+          ? undefined
+          : {
+              scope: `${mockLocale}:${Boolean(mockEnabled)}`,
```
sidmorizon: # 🟡 [P2] Mocked scope tag can never be stale, so this suite can't regress-detect it

## Problem

The mock builds the scope tag from the **current** render's mocks:

```ts
scope: `${mockLocale}:${Boolean(mockEnabled)}`
```

The hook derives its own `scope` from those same two values, so the guard it feeds — `scopedResult?.scope === scope` — is true by construction on every render in this file. The mock reproduces the *shape* of the scope tag but structurally cannot reproduce the *condition* it exists to guard: a retained result whose tag is stale relative to the current render.

## Impact

The two tests named for this problem — `does not inherit the previous locale completion before the new request starts` and `ignores a request callback from the previous locale` — flip `mockLocale` to `'zh-CN'` but run with `mockResult === undefined` and only assert `isLoading`, so they exercise `settledScope` alone, not the tag.

Concretely: deleting the entire tag-and-unwrap mechanism from the hook (the `{ scope, banners }` wrapping and the `scopedResult?.scope === scope` check) leaves this whole suite green. Only `recovery.test.tsx` catches it, and only on the web path — which is also why the native gap in the sibling comment on `useMarketBannerList.ts:173` is invisible to CI.

## Recommendation

Freeze the tag at the moment `mockRequest` is captured, so flipping `mockLocale` afterwards yields a genuinely stale tag:

```ts
mockRequest = request;
mockRequestScope = `${mockLocale}:${Boolean(mockEnabled)}`; // captured, not re-derived
return {
  result:
    mockResult === undefined
      ? undefined
      : { scope: mockRequestScope, banners: mockResult },
  isLoading: mockLoading,
};
```

Then give those two locale tests a non-empty `mockResult` and assert `bannerList === []` / `isFetched === false` after the switch, so they fail if the tag check is removed.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
zy2000-star: fixed: Reproduced both native locale and mock switches with a seeded old-scope SWR array, a pending initial request, and a failing new-scope request; both tests failed before the fix. Moved fallback snapshot mutation out of render into the committed effect alongside persistence, so a discarded native cache-switch render cannot relabel the old array. Also clear a missing same-scope snapshot. Both cases now pass, including late old-request completion and unmount/remount without a foreign cache entry. All 60 banner tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.test.tsx
zy2000-star: fixed: The harness now stores mockResultScope independently from mockLocale/mockEnabled instead of rebuilding the tag during render. New locale/mock tests retain a non-empty old-tagged result across the switch and assert empty banners, isFetched false, and correct pending/failure behavior, including a late old callback. Existing tests explicitly set the result tag when simulating newly delivered data. Real native hook tests additionally cover cache replay, failed switches, late completions and remount. All 60 banner tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/useMarketBannerList.ts
```diff
@@ -33,43 +127,124 @@ export function useMarketBannerList(): {
...
+            tokens: banner.indices,
+          }
+        : banner,
+    );
+  }, [bannerList, liveQuotes, requestScope, enableMockMarketBanner]);
+  useEffect(() => {
+    // A native cache-key switch can discard a render that still sees the old
+    // array. Only committed renders may update the fallback or persisted data.
+    if (bannerList !== undefined) {
+      committedResultRef.current = {
```
sidmorizon, line 226: # 🟡 [P2] catch fallback can replay a pre-hydration snapshot

## Problem

Moving this write into a **passive** effect is the right fix for the discarded-render problem, but it also desynchronises the ref from the commit that produced it. React 18 flushes passive effects through the Scheduler (a MessageChannel macrotask), so between a commit and its effect flush any already-queued promise continuation runs first — microtasks before macrotasks.

Of the three readers of `committedResultRef.current`, two are safe: the `normalizedBanners` memo already read the previous cycle's value (the old write sat *after* it in the render body), and the `liveQuotes` runner's one-commit staleness is benign because `merge(bannerList_N, normalized_{N-1})` and `merge(bannerList_N, normalized_N)` are equal under this predicate.

The base runner's `catch` at line 146 is the exception — it is the only reader that executes outside an effect flush, so it is the only one that can observe the pre-commit value.

## Impact

Requires a base payload that ships an explicit `tokens` array, which `8fbe8019` deliberately made meaningful (and which the test at `useMarketBannerList.test.tsx:416` establishes as a real shape):

1. Base succeeds for scope A with `[B(tokens: [])]`. Commit C1 → effect stamps `{A, [B(tokens: [])]}`.
2. The 30s poll (or a focus/reconnect revalidate) dispatches base run #2.
3. Hydration resolves `[B(tokens: [AAPL @100])]` → commit C2, `hasCurrentQuotes` true, `normalizedBanners = [B(tokens: [100])]`. C2's passive effects are queued.
4. Run #2 rejects **before** that flush. The `catch` reads the still-C1 ref, scope matches, and returns `{ scope, banners: [B(tokens: [])] }`.
5. Commit C3: `bannerList` has a new identity, so `liveQuotes.source === bannerList` fails and `preferQuotes` is false. The merge guard at line 95 is `(preferQuotes || banner.tokens === undefined)` — `tokens` is `[]`, not `undefined` — so the just-hydrated `[100]` is **not** grafted back.

The banner rolls back from live prices to the empty rows of the base payload. C3's effect then writes that regressed snapshot into the ref *and*, on native, into `swrCacheUtils.set(swrKeys.marketHomeBanners(...))`, so it survives the next cold start. It recovers only on a later successful hydration; while offline the re-dispatched hydration returns the banner unchanged and `preferQuotes: true` re-grafts the empty `tokens`, so it sticks.

Before this commit the render-body write happened synchronously inside C2's render task, so no continuation could interleave and the `catch` always replayed the hydrated snapshot.

## Recommendation

Use `useLayoutEffect` for the `committedResultRef` write. That keeps the property this commit is buying — it still runs only for committed renders, so the discarded render-phase-adjustment pass cannot stamp it — while running synchronously within the commit, which closes the window against the async `catch`. Leave the MMKV `swrCacheUtils.set` in the existing `useEffect` so persistence doesn't block paint.

Alternatively, keep one effect and have the `catch` mark its result as a replay so `normalizedBanners` re-merges it with `preferQuotes: true`.

Confidence note: the mechanism and the non-idempotent `preferQuotes: false` merge are both certain from the code; what's uncertain is how often the rejection lands inside that window. The `useLayoutEffect` change is a one-word fix with no cost, which is why it seems worth doing rather than reasoning about the odds.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerList.tsx
```diff
@@ -46,6 +48,7 @@ function BannerContainerMobile({
   return (
     <ScrollGuard>
       <ScrollView
+        h={MARKET_BANNER_LIST_MOBILE_HEIGHT}
```
sidmorizon: # 🟠 [P1] Fixed 204 box vs merged variable header height

## Problem

The merge with `x` reconciled two independent fixes to the same behaviour, and kept one half of each.

Base PR #13360 made the mobile banner height **conditional**: `MARKET_MOBILE_BANNER_MODERN_HEIGHT = 204` vs `MARKET_MOBILE_BANNER_LEGACY_HEIGHT = 134`, selected by `getMarketMobileBannerHeaderHeight()` (`layouts/mobileLayoutUtils.ts:8-16`). The resolution adopted that on the consumer side — `MobileLayout.native.tsx:439` now reads `bannerHeaderHeightRef.current.height` — and correctly dropped this PR's `MARKET_BANNER_HEADER_HEIGHT` edit there.

But it kept this PR's **unconditional** producer: `MARKET_BANNER_LIST_MOBILE_HEIGHT = 204` (line 20) pinned onto the `ScrollView` here. Before the merge each side was internally consistent — this PR pinned everything to 204, base left the `ScrollView` intrinsically sized so it always matched the computed value. Now the box is fixed and the header gate is variable.

`MARKET_BANNER_LIST_MOBILE_HEIGHT` is also a dead export after the merge: `grep` finds no consumer outside this file, since `MobileLayout.native.tsx` no longer imports it.

## Impact

Native phone (`useMedia().md`, so `BannerContainerMobile`), with a banner response from an older API deployment where items have no `tokens` — the path `MarketBannerItem.tsx:361` (`if (!item.tokens) return <LegacyMarketBannerItem …/>`) explicitly supports, and the one #13360 added `$gtMd.h: platformEnv.isNative ? 118 : 'auto'` for:

- `getMarketMobileBannerHeaderHeight([{…no tokens}])` → `every(Boolean(undefined))` is false → **134**, so `headerHeight = 134`.
- This `ScrollView` lays out a hard **204**, while its content is 118 (legacy card) + 16 (`py: '$2'`) = 134.

The collapsible header declares 134 while the banner box occupies 204 — a 70pt discrepancy. On Android that also feeds `pagerHeight` (`headerHeight + stickyHeaderHeight`), so the list offset is wrong by 70pt and the sticky header/tab bar overlaps the banner region. On web at a mobile viewport it is 70pt of dead space with no header gate to absorb it. That is the header-height jump this PR set out to remove, reintroduced for the legacy payload.

For reference the arithmetic works out exactly without any fixed height: legacy 8+118+8 = 134, modern 8+188+8 = 204 (`py="$5"` + title `h="$6"` + `gap="$5"` + tokens `minHeight={104}`).

## Recommendation

Remove the second source of truth. Simplest is to drop `h` from this `ScrollView` and let it size intrinsically, which the arithmetic above shows already lands on 134/204 correctly for both card kinds. If a fixed box is wanted for the web skeleton path, derive it instead:

```ts
import { getMarketMobileBannerHeaderHeight } from '../../layouts/mobileLayoutUtils';
// …
h={getMarketMobileBannerHeaderHeight(visibleBannerList)}
```

Either way delete `MARKET_BANNER_LIST_MOBILE_HEIGHT` (line 20) so 204 lives only as `MARKET_MOBILE_BANNER_MODERN_HEIGHT`.

Note this line currently has no test coverage in either direction: `MarketBannerList.test.tsx` came through the merge byte-identical to base, and its one height test asserts `pt` on the tablet `XStack` only — the `ScrollView` mock destructures just `children`/`opacity` and discards `h`. Extending that mock to expose `data-height` and asserting it equals `getMarketMobileBannerHeaderHeight(bannerList)` for both a `tokens`-bearing and a `tokens`-less fixture would pin the invariant that ties this line to `MobileLayout.native.tsx:439`.

[Claude session ↗](https://claude.ai/code/session_01NckxJb3Pm4pKEwN2N6u5xw)

---

zy2000-star: fixed: Removed MARKET_BANNER_LIST_MOBILE_HEIGHT and derive the phone ScrollView height from getMarketMobileBannerHeaderHeight(visibleBannerList), the same helper used by the native header. Retained hidden banners preserve the occupied height after empty refreshes. The helper now uses the tallest card for mixed legacy/modern lists. Added native and mobile-Web tests exposing ScrollView height and checking legacy 134, modern/mixed 204, transitions, and native empty-refresh retention. All 83 relevant tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
```diff
@@ -25,11 +29,39 @@ function isBannerDetailSortBy(
...
   isPerps: boolean;
+  isStock?: boolean;
+  isIndex?: boolean;
 };
 
+function mapStockBannerItemToToken(
+  item: IMarketStockPublicItem,
+): IMarketTokenListItem {
+  return {
+    address: item.stockId,
```
sidmorizon: 🟡 P2: Stock IDs become contract addresses

Severity: `non-severe`

Stock banner rows set `address` to `stockId`, such as `AAPL`. Desktop rows then display and copy that ticker as a contract address.

Suggested fix:

```
    address: '',
```

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
```diff
@@ -68,8 +70,16 @@ const BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS = ['liquidity'] as const;
 
 function MarketBannerDetailContent({ title }: { title: string }) {
   const route = useRoute<IMarketBannerDetailRouteParams>();
-  const { tokenListId, type } = route.params;
+  const { tokenListId, type, assetType } = route.params;
   const isPerps = type === EMarketBannerType.Perps;
+  const isMixed = type === EMarketBannerType.StockPerps;
```
originalix: P2: [Handle the declared mixed banner type]

When the API returns the newly declared `EMarketBannerType.Mixed`, this check leaves `isMixed` false. The banner is hydrated through the generic token-list route and the detail page never renders `PerpsTokenListSection`, so the composite banner is incomplete.

Classify `Mixed` through the same intended composite path as `StockPerps` (or remove the enum value if it cannot be emitted), and add coverage for that response type.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
zy2000-star: fixed: Stock banner rows now use an empty contract address while preserving stockId and stock metadata for stock detail navigation. Added a hook test covering both the mapped response and transformed UI row. All 72 related tests and commit-profile lint/TypeScript checks pass.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
zy2000-star: fixed: Mixed and StockPerps now share isMarketMixedBanner in both banner hydration and detail classification. Both use the stock quote endpoint and display the stock list plus PerpsTokenListSection. Added hydration coverage for both enum values and rendered detail-page tests for narrow and wide layouts. All 72 related tests and commit-profile lint/TypeScript checks pass.

originalix (approved):

PR #13364, opened 2026-09-12
fix: restore market filters and reset search scroll | OK-62716 & OK-62947

zy2000-star:
[OK-62716](<https://onekeyhq.atlassian.net/browse/OK-62716>) [OK-62947](<https://onekeyhq.atlassian.net/browse/OK-62947>)

----------

## Summary

Fix two Market UI issues:

- **OK-62716 — Restore the time filter layout**
  Move the time-range dropdown back to the left of the network selector, on the same row as the category tabs. Remove the separate trending toolbar from the desktop layout while preserving the existing time-range state and data-fetching behavior.

- **OK-62947 — Reset stock search scroll position**
  After scrolling the stock recommendation list, searching could retain the previous scroll position and hide the first results. Key the stock table by the normalized search query so entering a new query, changing it, or clearing it starts the list at the top. Loading additional pages for the same query preserves the scroll position.

## Validation

- Passed 7 targeted tests covering stock selection, search scroll reset, and pagination.
- Added regression coverage for query changes, clearing search, and preserving scroll when appending results.
- Passed `yarn agent:check --profile commit`, including lint, formatting, and TypeScript checks.
- User confirmed local testing passed.

[OK-62716]: https://onekeyhq.atlassian.net/browse/OK-62716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62947]: https://onekeyhq.atlassian.net/browse/OK-62947?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
+++ b/packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
@@ -150,6 +150,8 @@ const MarketStockSelectorList = memo(
     return (
       <YStack height={STOCK_SELECTOR_TABLE_HEIGHT}>
         <Table<IMarketStockPublicItem>
+          // Reset scroll for a new search, but preserve it when appending pages.
+          key={normalizedQuery}
           columns={columns}
           dataSource={items}
           keyExtractor={(item) => item.stockId}
--- a/packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
+++ b/packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
@@ -35,7 +35,6 @@ import { MarketListLoadingFallback } from '../components/MarketTokenList/MarketL
 import { MarketNormalTokenList } from '../components/MarketTokenList/MarketNormalTokenList';
 import { MarketTopCoinsList } from '../components/MarketTopCoinsList/MarketTopCoinsList';
 import { TimeRangeDropdown } from '../components/TimeRangeDropdown';
-import { TrendingDesktopToolbar } from '../components/TrendingDesktopToolbar';
 import {
   COMPACT_SPOT_HIDDEN_DESKTOP_COLUMNS,
   getMarketCategoryTooltipId,
@@ -280,9 +279,6 @@ export function DesktopLayout({
       const showNetworkSelector = shouldShowSpotNetworkSelector(
         currentSpotCategoryId,
       );
-      const usesTrendingStyle = isTrendingStyleSpotCategory(
-        currentSpotCategoryId,
-      );
       // Wrap TabBar + portal target in a single sticky container.
       // Override TabBar's own sticky with position: relative so
       // the outer wrapper controls stickiness for both.
@@ -314,12 +310,10 @@ export function DesktopLayout({
               alignItems="center"
               pr="$5"
             >
-              {usesTrendingStyle ? null : (
-                <TimeRangeDropdown
-                  value={currentFilterBarProps.timeRange}
-                  onChange={currentFilterBarProps.onTimeRangeChange}
-                />
-              )}
+              <TimeRangeDropdown
+                value={currentFilterBarProps.timeRange}
+                onChange={currentFilterBarProps.onTimeRangeChange}
+              />
               <XStack display={showNetworkSelector ? 'flex' : 'none'}>
                 <CompactNetworkSelector
                   selectedNetworkId={currentFilterBarProps.selectedNetworkId}
@@ -450,14 +444,6 @@ export function DesktopLayout({
               centerDesktopPortalContent
               desktopColumnVariant={usesTrendingStyle ? 'trending' : 'default'}
               useApiDefaultSort={usesTrendingStyle}
-              toolbar={
-                usesTrendingStyle ? (
-                  <TrendingDesktopToolbar
-                    timeRange={filterBarProps.timeRange}
-                    onTimeRangeChange={filterBarProps.onTimeRangeChange}
-                  />
-                ) : undefined
-              }
             />
           );
         }
--- a/packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.test.tsx
+++ b/packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.test.tsx
@@ -183,6 +183,36 @@ describe('MarketStockSelectorList', () => {
     expect(mockOnItemPress).toHaveBeenCalledWith(mockStock);
   });
 
+  it('resets scroll for new queries and preserves it for pagination', () => {
+    const { rerender } = render(
+      <MarketStockSelectorList onItemPress={mockOnItemPress} query="" />,
+    );
+    let table = screen.getByTestId('stock-table');
+    table.scrollTop = 180;
+
+    for (const query of ['mu', 'aapl', '']) {
+      rerender(
+        <MarketStockSelectorList onItemPress={mockOnItemPress} query={query} />,
+      );
+      const nextTable = screen.getByTestId('stock-table');
+      expect(nextTable).not.toBe(table);
+      expect(nextTable.scrollTop).toBe(0);
+      table = nextTable;
+      table.scrollTop = 180;
+    }
+
+    mockSelectorListResult.items = [
+      mockStock,
+      { ...mockStock, stockId: 'MU', symbol: 'MU' },
+    ];
+    rerender(
+      <MarketStockSelectorList onItemPress={mockOnItemPress} query=" " />,
+    );
+    expect(screen.getByTestId('stock-table')).toBe(table);
+    expect(table.scrollTop).toBe(180);
+    expect(screen.getByTestId('stock-row-MU')).toBeTruthy();
+  });
+
   it('triggers selector pagination when the table reaches the end', () => {
     mockSelectorListResult.canLoadMore = true;
     render(<MarketStockSelectorList onItemPress={mockOnItemPress} query="" />);


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-00a409e9-df2c-56fb-808e-fc1c96e4fed4)

sidmorizon:
Claude review session: https://claude.ai/code/session_01MqTpc8ScVTGzCzsmey6ZKu

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
```diff
@@ -150,6 +150,8 @@ const MarketStockSelectorList = memo(
     return (
       <YStack height={STOCK_SELECTOR_TABLE_HEIGHT}>
         <Table<IMarketStockPublicItem>
+          // Reset scroll for a new search, but preserve it when appending pages.
+          key={normalizedQuery}
```
sidmorizon: # 🟡 [P2] `key={normalizedQuery}` can never fire — the table already unmounts on every query change, and the new test cannot fail

## Problem

The `Table` never stays mounted across a query change, so keying it by the query is a no-op.

When `query` changes, `useMarketStockSelectorList` returns `items: []` **and** `isLoading: true` in the *same* render:

- `useMarketStockSelectorList.ts:92-93` — `hasCurrentData = listState.queryKey === queryKey`. `listState.queryKey` still holds the previous query (it is only advanced by the effect at lines 70-90, which requires `firstPageResult.queryKey === queryKey`), so `items` is `[]` immediately.
- `useMarketStockSelectorList.ts:97-99` — `firstPageResult` still carries the old key (or is `undefined`, given `undefinedResultIfReRun: true` at line 65), so `isFirstPagePending` is true, and the returned `isLoading` at line 173 is true.

That hits the early return at `MarketStockSelectorList.tsx:97`, which renders the Spinner subtree instead of the `Table`. React reconciles the two returned `YStack`s as the same element, but their children differ in type (`Spinner` vs `Table`), so the `Table` unmounts and its scroller is destroyed. It remounts at offset 0 when the first page for the new query arrives — with or without the `key`. On the pagination path `queryKey` is deliberately held constant (lines 112-147), so `normalizedQuery` does not change there either.

The new test at `MarketStockSelectorList.test.tsx:186-214` cannot catch this: `jest.mock('./useMarketStockSelectorList')` (lines 135-140) always returns `mockSelectorListResult` with `isLoading: false` and non-empty `items` (lines 41-50) — a state the real hook cannot be in right after a query change — and `Table` is mocked as a plain `<div>` (lines 69-117). `expect(nextTable.scrollTop).toBe(0)` (line 199) is trivially true for any freshly created jsdom node.

## Impact

Two costs. The `key` and its comment document behavior the file does not actually rely on, so the next reader will treat the remount as load-bearing. More importantly, OK-62947 is reported as fixed and covered by a regression test, but the production code path is unchanged and the test would keep passing if a real reset regressed. If the scroll bug is genuinely reproducible, its cause lies elsewhere (e.g. an ancestor scroller in the containing Popover) and is not addressed here.

## Recommendation

Either drop `key={normalizedQuery}` and its comment at `MarketStockSelectorList.tsx:153-154` — behavior is identical — or, if the intent is to keep previous results visible while the next query loads, make the key load-bearing by relaxing the gate at `MarketStockSelectorList.tsx:97` so stale items stay mounted during a query switch. Either way, the test should drive the real transition (mock the hook to return `isLoading: true, items: []` for the intermediate render) rather than a permanently-loaded stub.

[Claude session ↗](https://claude.ai/code/session_01MqTpc8ScVTGzCzsmey6ZKu)

---

packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
```diff
@@ -450,14 +444,6 @@ export function DesktopLayout({
               centerDesktopPortalContent
               desktopColumnVariant={usesTrendingStyle ? 'trending' : 'default'}
               useApiDefaultSort={usesTrendingStyle}
```
sidmorizon, line 479: # 🟡 [P2] Removing the trending toolbar orphans `TrendingDesktopToolbar` and `MarketNormalTokenList`'s `toolbar` prop

## Problem

This hunk deleted the only `toolbar={...}` argument ever passed to `MarketNormalTokenList`, and the import removal at line 38 dropped the only reference to `TrendingDesktopToolbar`.

After this commit:

- `TrendingDesktopToolbar` is referenced only by its own definition and barrel — `components/TrendingDesktopToolbar/TrendingDesktopToolbar.tsx:13,100` and `index.ts:1`. `git grep -n TrendingDesktopToolbar -- packages apps` returns no other hit: no component, no test, no reference to its `market-trending-desktop-toolbar` / `market-trending-time-*` testIDs.
- `IMarketNormalTokenListProps.toolbar` (`MarketNormalTokenList.tsx:29,55,122`) now has zero callers passing it. The three remaining call sites — `DesktopLayout.tsx:432`, `SwapProSelectTokenModal.tsx:176` and `MobileTokenSelector.tsx:220` — all omit it.

## Impact

101 lines of unreachable UI, plus a prop that keeps live-looking `MarketNormalTokenList` → `MarketTokenListBase` plumbing (`MarketNormalTokenList.tsx:122`) alive for no consumer. Both read as active behavior to the next person touching the trending layout, and the orphaned filter chips — sole users of `market_filter_chip_top_turnover` / `market_filter_chip_mid_cap` / `market_filter_chip_large_cap` / `market_filters` — will silently drift from whatever the real filters UI becomes.

## Recommendation

Delete `packages/kit/src/views/Market/MarketHomeV2/components/TrendingDesktopToolbar/` and drop `toolbar` from `IMarketNormalTokenListProps` (`MarketNormalTokenList.tsx:29,55,122`) in this PR. This is the change that made them unreachable, so removal is in scope under CLAUDE.md's "do not remove existing code unless the request requires it". Note the `toolbar` prop on `MarketTokenListBase` itself must stay — `MarketStockList.tsx:176`, `MarketPerpsTokenList.tsx:186` and `MarketWatchlistTokenList.tsx:373` still use it. If the toolbar is instead meant to re-land under the new filter design, keep it but add a one-line comment at `TrendingDesktopToolbar.tsx:13` saying so, so it is not mistaken for live code.

[Claude session ↗](https://claude.ai/code/session_01MqTpc8ScVTGzCzsmey6ZKu)

---

packages/kit/src/views/Market/MarketDetailV2/components/TokenSelector/MarketStockSelectorList.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/layouts/DesktopLayout.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/TrendingDesktopToolbar/TrendingDesktopToolbar.tsx
```diff
@@ -1,100 +0,0 @@
-import { memo } from 'react';
-
-import { useIntl } from 'react-intl';
-
-import { Icon, SizableText, Stack, XStack } from '@onekeyhq/components';
-import { ETranslations } from '@onekeyhq/shared/src/locale';
-
-import { MARKET_LIST_TOOLBAR_ITEM_HEIGHT } from '../../../marketDesktopLayoutConstants';
```
sidmorizon, line 8: # 🟡 [P2] Toolbar deletion orphans `MARKET_LIST_TOOLBAR_ITEM_HEIGHT` and leaves a stale comment describing the removed filter row

## Problem

This deleted line was the only import of `MARKET_LIST_TOOLBAR_ITEM_HEIGHT` (used at `TrendingDesktopToolbar.tsx:45` before deletion). With the component gone, the constant has no consumers left: `grep -rn "MARKET_LIST_TOOLBAR_ITEM_HEIGHT"` across the worktree now matches only its own declaration at `packages/kit/src/views/Market/marketDesktopLayoutConstants.ts:83`.

The neighbouring comment in that same file is now inaccurate too. `marketDesktopLayoutConstants.ts:74-78` justifies the fixed band height with:

> keeps the band identical on pages whose toolbar content is taller (**Trending's 40px filter row**)

That filter row was this component, and it no longer renders anywhere. The comment sits on `MARKET_DESKTOP_TOOLBAR_BAND_HEIGHT`, which is still live — it feeds `MARKET_DESKTOP_TOOLBAR_BAND_STYLE`, consumed at `MarketDesktopStickyHeader.tsx:14,27` — so the stale rationale is attached to a constant people will actually revisit.

## Impact

Neither will be caught automatically: the repo has no `knip`/`ts-prune`, and oxlint does not flag unused *exports*, so this survives indefinitely. The dead constant reads as a live layout contract, and the comment actively misleads the next person calibrating the toolbar band — they will go looking for a Trending filter row that no longer exists, and may preserve a 68px band for a case that is gone.

## Recommendation

In `packages/kit/src/views/Market/marketDesktopLayoutConstants.ts`, delete the now-unused `MARKET_LIST_TOOLBAR_ITEM_HEIGHT` export at line 83 (with its comment at lines 80-82), and reword lines 74-78 to drop the "Trending's 40px filter row" clause — cite a surviving taller toolbar if one still motivates the fixed height, otherwise state that the band is fixed purely to keep the table from shifting between tabs.

[Claude session ↗](https://claude.ai/code/session_01MqTpc8ScVTGzCzsmey6ZKu)

---

zy2000-star: fixed

originalix (approved):

PR #13366, opened 2026-09-12
fix: improve App Clip market experience

huhuanming:
## Summary
- build the App Clip Market with stocks, perpetuals, and trending tabs backed by the test-environment market APIs
- route every row through the unified chart detail, add native-style pressed feedback, and keep images cached across list reuse
- retry failed market loads after network recovery and match the main app network/time filter sheets

## Validation
- OneKeyAppClip Debug simulator build
- OneKeyAppClip Release simulator build
- iOS simulator: stock, perpetual, and trending row navigation and charts
- iOS simulator: row pressed state, network search/selection, time-range selection, keyboard avoidance, and light/dark filter sheets
- git diff --check
- Localizable.xcstrings JSON validation

## Notes
- Market APIs are currently exercised through the test App Clip invocation environment.
- Real-device App Clip invocation and AASA validation remain separate from this simulator acceptance.


apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -119,20 +262,134 @@ private struct MarketScalar: Decodable {
...
+    baseURL: URL,
+    category: String
+  ) async throws -> [AppClipMarketPerp] {
+    var components = URLComponents(
+      url: baseURL.appendingPathComponent("utility/v2/market/perps/token-list"),
+      resolvingAgainstBaseURL: false
+    )
+    components?.queryItems = [
+      URLQueryItem(name: "category", value: category),
+      URLQueryItem(name: "assetTypeVersion", value: "1"),
```
originalix: P2: [Perps request uses an obsolete asset capability version]

When a category contains a sub-DEX instrument, this request advertises `assetTypeVersion=1`. The current market client sends version 3 because the server withholds newer sub-DEX rows for older versions, so the App Clip Perps list silently omits supported instruments.

Send the current perps asset capability version (`3`) here and keep it synchronized with the supported sub-DEX set.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -185,54 +442,126 @@ actor AppClipMarketService {
...
+  ) async throws -> AppClipKlineResult {
+    let normalizedPeriod = period.lowercased()
+    guard ["1h", "1d", "1w", "1m", "1y", "all"].contains(normalizedPeriod) else {
+      throw URLError(.badURL)
+    }
     var components = URLComponents(
-      url: baseURL.appendingPathComponent("utility/v2/market/token/list"),
+      url: baseURL
+        .appendingPathComponent("utility/v1/stocks")
+        .appendingPathComponent(stockID)
```
originalix: P2: [Stock chart requests do not escape slash-containing IDs]

When `stockID` is `BRK/B`, appending it as a URL path component turns the slash into a path separator. The endpoint requires a single encoded stock-ID segment (`BRK%2FB`), so the chart request hits the wrong route and the detail screen falls into its chart-unavailable state.

Build this part of the endpoint as a percent-encoded path segment that escapes `/` before appending `/chart`.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9b00a3c2-58ae-5fc3-8b52-e6e06be5b1d2)

sidmorizon:
Claude review session: https://claude.ai/code/session_01EFzCU3MeRe8ejq2ySHvxcP

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
sidmorizon: 🟡 P2: Perpetual list omits sub-DEX assets

Severity: `non-severe`

`fetchPerps` requests asset version 1, so the backend withholds newer sub-DEX markets. The current [version contract](packages/shared/types/hyperliquid/perp.constants.ts:426-431) requires version 3.

Suggested fix:

```
      URLQueryItem(name: "assetTypeVersion", value: "3"),
```

apps/mobile/ios/OneKeyAppClip/AppClipRootView.swift
```diff
@@ -133,8 +743,8 @@ struct AppClipRootView: View {
           .padding(.top, 5)
           .accessibilityIdentifier("app-clip-detail-price")
         HStack(spacing: 6) {
-          Text(formattedPercentage(asset.priceChangePercent))
-            .foregroundColor(changeColor(asset.priceChangePercent))
+          Text(formattedPercentage(marketDetailChange(detail)))
+            .foregroundColor(changeColor(marketDetailChange(detail)))
           Text("24h")
```
sidmorizon: 🟡 P2: Token change carries wrong period label

Severity: `non-severe`

Opening a filtered token displays its selected-period `priceChangePercent` beside `24h`. The detail mislabels every 5m, 1h, and 4h change.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -220,99 +278,590 @@ final class AppClipModel: ObservableObject {
...
+        environmentID == requestEnvironmentID,
+        apiBaseURL == requestBaseURL,
+        stockRequestIDs[key] == requestID
+      else {
         return
       }
-      guard !result.isEmpty else {
-        throw URLError(.zeroByteResource)
+      var current = stockStates[key] ?? AppClipMarketListState()
+      current.items = page.items
```
sidmorizon, line 604: # [🟠 P1] Background refresh truncates a paginated Stocks list back to page 1

## Problem
Stocks is the only paginated tab (`fetchStocks` sends `limit=20` and a cursor, `AppClipMarketService.swift:302-308`; `loadMoreStocks` appends pages at `AppClipModel.swift:378-380`). The 60 s timer in `start()` (`AppClipModel.swift:211-222`) and `appDidBecomeActive` (`:237-239`) both call `refreshMarkets()` → `refreshStocks(force: false)`, and on success this line replaces the whole accumulated array with the first page and resets `nextCursor`:

```swift
current.items = page.items
current.nextCursor = page.nextCursor
```

Nothing distinguishes a user-initiated pull-to-refresh (`.refreshable` → `refreshMarkets(force: true)`, `AppClipRootView.swift:137-139`) from the background refresh, so the automatic path performs the same full replacement.

## Impact
A user who has scrolled through several pages of stocks has the list cut back to 20 rows every 60 s and on every return to foreground (after 15 s), with the scroll position jumping accordingly. Perps and Trending are unaffected because they are not paginated.

## Recommendation
In `refreshStocks` (`AppClipModel.swift:537-589`), only do the full replacement when `force == true`. For the background path (`!force`) either request `limit: max(20, current.items.count)` (the `/utility/v1/stocks` contract honours `limit`, see `packages/kit-bg/src/services/ServiceMarketV2.ts:1128-1134`) and replace, or merge the fresh first page into the existing array by `id` without touching `nextCursor`. Alternatively skip the automatic refresh while `current.nextCursor != nil && current.items.count > 20` and let the timer only refresh non-paginated tabs.

[Claude session ↗](https://claude.ai/code/session_01EFzCU3MeRe8ejq2ySHvxcP)

---

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -150,53 +185,74 @@ final class AppClipModel: ObservableObject {
...
 
   func handleInvocation(_ url: URL) {
     guard let invocation = AppClipInvocation(url: url) else {
       return
     }
-    hasHandledInvocation = true
-    marketRequestID = UUID()
+    let wasStarted = hasStarted
+    let environmentChanged = apiBaseURL != invocation.apiBaseURL
+    invalidateMarketRequests()
```
sidmorizon: # [🟡 P2] Invocation discards in-flight requests and refetches every tab even when the environment is unchanged

## Problem
`start()` is no longer gated on the invocation (the base gated it on `hasHandledInvocation`); it now runs from `onAppear` (`AppClipRootView.swift:38-40`) and from `scenePhase == .active` (`appDidBecomeActive`, `AppClipModel.swift:226-229`), and because `screen` defaults to `.market` (`:151`) it immediately fires `refreshAllMarketPages()` against the default `https://utility.onekeycn.com` base (`:182`, `:208-210`). When `handleInvocation` then runs, this line calls `invalidateMarketRequests()` unconditionally, even though `environmentChanged` was just computed one line above and is only used to gate `resetMarketData()` (`:253-255`). Rotating `environmentID` makes every in-flight response get dropped by the `environmentID == requestEnvironmentID` guards (`:488-490`, `:513-519`, `:559-565`), and the `wasStarted` branch at `:268-271` then issues `refreshAllMarketPages(force: true)` again for config + stocks + perps + trending.

## Impact
On the App Clip's primary path (cold launch from a link) the initial batch is issued twice: four requests whose responses are thrown away, then four more. The same discard-and-refetch happens on every re-invocation while the clip is already open, regardless of whether the environment or attribution changed. That delays the first paint and doubles load on the Utility backend for no data benefit when the environment is the same.

## Recommendation
Gate the invalidation and the forced refetch on `environmentChanged`: call `invalidateMarketRequests()` and `resetMarketData()` only inside the `if environmentChanged` block at `AppClipModel.swift:253-255`, and make the `Task` at `:268-271` refetch only when `environmentChanged` (otherwise leave the in-flight requests to complete, or at most call `refreshVisibleDetail()`). Alternatively restore the previous behaviour of deferring the first market fetch in `start()` until the invocation has been handled so the clip never fetches against a base URL it has not confirmed yet.

[Claude session ↗](https://claude.ai/code/session_01EFzCU3MeRe8ejq2ySHvxcP)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
huhuanming: Fixed in ba7c892d08: the App Clip now sends assetTypeVersion 3, with a source comment tying the value to the shared PERPS_ASSET_TYPE_VERSION contract.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
huhuanming: Fixed in ba7c892d08: stock IDs are encoded as RFC 3986 path segments through URLComponents.percentEncodedPath. BRK/B now produces BRK%2FB as one endpoint segment.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
huhuanming: Fixed in ba7c892d08: fetchPerps now advertises assetTypeVersion 3 so supported sub-DEX instruments are included.

apps/mobile/ios/OneKeyAppClip/AppClipRootView.swift
huhuanming: Fixed in ba7c892d08: token detail now labels the snapshot change with the selected Trending time range; stock and perpetual snapshots remain labeled 24h.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in ba7c892d08: non-forced background stock refreshes request max(20, currently loaded row count), then replace with the coherently sorted response and its cursor. Forced pull-to-refresh keeps the first-page reset behavior, and background refresh no longer races load-more.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in ba7c892d08: start and active refresh are gated until an App Clip invocation is handled. Same-environment reinvocations keep in-flight market requests; environment changes still invalidate, reset, and force-refresh. A first same-environment Web-to-Market transition still loads market data.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -220,99 +302,592 @@ final class AppClipModel: ObservableObject {
...
+    let categoryId = selectedStockCategoryId
+    let key = categoryId
+    var state = stockStates[key] ?? AppClipMarketListState()
+    guard force || (!state.isLoading && !state.isLoadingMore) else {
+      return
+    }
+    let requestID = UUID()
+    let requestEnvironmentID = environmentID
+    let requestBaseURL = apiBaseURL
+    let requestLimit = force ? 20 : max(20, state.items.count)
```
sidmorizon: # [🟡 P2] Network-recovery retry still resets a paginated Stocks list to 20 rows

## Problem
The depth-preserving limit introduced here is keyed on `force == false`, but `force: true` is also used for the automatic network-recovery retry, which is not a user-initiated reset. Both recovery paths call `refreshMarkets(force: true)` (`handleNetworkStatus`, `AppClipModel.swift:787`; `retryMarketRefreshAfterNetworkRecoveryIfNeeded`, `:801`), and their guards accept `marketRefreshFailed` (`:779`, `:797`). A background refresh that fails while offline keeps the loaded rows and only sets `failed = true` (`:609-612`), so the sequence "paginate to N > 20 rows → connectivity drops → 60 s tick fails → connectivity returns" reaches this line with `force == true`, requests `limit=20`, and the replacement at `:593-594` collapses the list to the first page.

## Impact
The same list-shrink and scroll-jump as the earlier thread, now confined to the offline/online transition (tunnel, airplane-mode toggle), which is a realistic mobile path for an App Clip. Timer and foreground refreshes are correctly fixed by this commit.

## Recommendation
Decouple "bypass the loading guard and invalidate in-flight requests" from "reset pagination depth". Either compute `let requestLimit = max(20, state.items.count)` unconditionally at `AppClipModel.swift:571` (pull-to-refresh then also keeps the current depth, which is acceptable UX), or add a separate `resetsPagination: Bool` parameter to `refreshMarkets`/`refreshStocks` that only the `.refreshable` callers (`AppClipRootView.swift:137-139`) and the environment-change path in `handleInvocation` set to `true`, leaving the recovery calls at `:787` and `:801` at `false`.

[Claude session ↗](https://claude.ai/code/session_01EFzCU3MeRe8ejq2ySHvxcP)

---

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -220,99 +302,592 @@ final class AppClipModel: ObservableObject {
...
+        return
+      }
+      var current = stockStates[key] ?? AppClipMarketListState()
+      current.items = page.items
+      current.nextCursor = page.nextCursor
+      current.isLoading = false
+      current.didLoad = true
+      current.failed = false
+      current.lastUpdated = Date()
+      stockStates[key] = current
```
sidmorizon, line 610: 🟡 P2: Stock and perp details stop updating

Severity: `non-severe`

While a stock or perp detail remains open, `refreshStocks` and `refreshPerps` update only their cached list states. `screen` retains the original value snapshot, and neither path reloads its candles. The visible price, change, and chart remain stale indefinitely.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -220,99 +302,592 @@ final class AppClipModel: ObservableObject {
...
+
+  func selectMarketTab(_ tab: AppClipMarketTab) {
+    guard selectedMarketTab != tab else {
+      return
+    }
+    selectedMarketTab = tab
+    guard !activeDidLoad, !isRefreshing else {
+      return
+    }
+    Task { await refreshMarkets() }
```
sidmorizon: 🟡 P2: Previously loaded tabs show stale quotes

Severity: `non-severe`

Selecting a previously loaded tab makes `selectMarketTab` skip its refresh regardless of `lastUpdated`. The timer refreshes only the selected tab, so dormant quotes can be arbitrarily old for another minute.

Suggested fix:

```
    guard !isRefreshing else {
      return
    }
    Task { await refreshMarkets() }
```

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in e5dde1f934: stock refresh now always requests max(20, currently loaded row count), including network-recovery and pull-to-refresh paths. Environment reset still naturally starts at 20 because the state is empty.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in e5dde1f934: successful stock, perpetual, and trending refreshes now synchronize the matching visible detail snapshot and reload candles for the current interval. The helper validates type and asset identity before updating the screen.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in e5dde1f934: switching tabs now refreshes missing, failed, or older-than-15-second data for the newly selected tab. It skips an in-flight refresh and rechecks the selected tab before the async refresh starts.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
```diff
@@ -176,7 +227,13 @@ final class AppClipModel: ObservableObject {
...
     }
-    start()
+    if !hasStarted {
+      start()
+      return
+    }
+    guard shouldRefreshMarketContent, !isLoadingConfiguration else {
+      return
+    }
     guard !isRefreshing else {
```
sidmorizon, line 240: # [🟡 P2] Foreground chart refresh now depends on the list refresh succeeding

## Problem
After this guard, `appDidBecomeActive` only calls `refreshVisibleDetail()` in the `else` branch (`AppClipModel.swift:240-244`), i.e. when the list is less than 15 s old. In the normal case after backgrounding the list is stale, so the open detail is refreshed only through the new success-path hooks in `refreshStocks` / `refreshPerps` / `refreshTrending` (`:610-613`, `:664-667`, `:724-727`). The three catch blocks (`:615-628`, `:669-682`, `:729-742`) never touch the detail, and `refreshStocks` returns early without fetching while a load-more is in flight (`:575`), a case `isRefreshing` does not cover. The pre-PR model handled the failure case by calling `refreshVisibleDetail()` from its catch block; that fallback is dropped here, and the previous revision of this PR (`ba7c892d`) avoided the problem by calling `refreshVisibleDetail()` unconditionally before the staleness check.

## Impact
Returning to a stock, perp, or token detail after time away can leave the chart and header price silently stale: if the list request fails (timeout, 5xx, brief offline blip) nothing reloads the candles and no error or spinner is shown until a later timer tick succeeds; if a stocks load-more happens to be in flight, no list refresh and no candle reload happen at all. Even on success, the chart update is serialised behind the config and list round trips instead of running in parallel, adding their full latency to the screen the user is actually looking at.

## Recommendation
Keep the unconditional `refreshVisibleDetail()` before the staleness check (as in `ba7c892d`) and avoid the duplicate candle request by having `refreshVisibleDetail(with:)` (`AppClipModel.swift:1019-1041`) skip `prepareCandleRequest` / `loadDetailCandles` when `isLoadingCandles` is already true for the same detail id and `selectedInterval`, while still applying the `screen = .detail(refreshed)` price update. Alternatively, restore the pre-PR behaviour by calling `refreshVisibleDetail()` from the three catch blocks when `screen` is a detail of that kind, so a failed list refresh does not strand the open chart.

[Claude session ↗](https://claude.ai/code/session_01EFzCU3MeRe8ejq2ySHvxcP)

---

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
sidmorizon, line 1: 🟡 P2: Failed refresh leaves chart stale

Reported at `apps/mobile/ios/OneKeyAppClip/AppClipModel.swift:240-243` (outside this PR's diff; file-level comment).

Severity: `non-severe`

When `lastUpdated` is stale and the market request fails, `appDidBecomeActive` never refreshes the visible chart. Only successful list callbacks refresh candles, so foregrounding leaves the chart stale.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in 7cc849031f. Foreground activation now refreshes the visible detail before the list loading and staleness guards, so candle refresh is independent of list success and stock load-more. Candle requests now track detail identity and interval; a later list success still updates the visible header snapshot, while an in-flight request for the same detail and interval is reused instead of duplicated. Debug and Release OneKeyAppClip simulator builds passed.

apps/mobile/ios/OneKeyAppClip/AppClipModel.swift
huhuanming: Fixed in 7cc849031f. The visible chart refresh now runs independently whenever the App Clip becomes active, before any list refresh guard. Therefore a failed or already-running market refresh can no longer strand the chart, and same-detail same-interval candle requests are deduplicated without blocking the list result from updating the detail header.

originalix (approved):

originalix (approved):

PR #13367, opened 2026-09-13
fix: stop settings modal flashing back to loading OK-53013

huhuanming:
[OK-53013](<https://onekeyhq.atlassian.net/browse/OK-53013>)

----------

## Problem

[OK-53013](https://onekeyhq.atlassian.net/browse/OK-53013) — on the extension, opening Settings paints the page and then, ~20ms later, blanks it back to the loading spinner for 150–300ms before the content returns. The reported "闪烁" is that `content → loading → content`.

Reproduces only when Settings is opened **while the wallet home is still loading** (the state in the bug video, where the balance was still updating). A settled home does not reproduce it.

## Root cause

`passwordWebAuthInfoAtom` is an async computed atom that read `webAuthCredentialId` from `passwordPersistAtom` purely to derive an `isEnable` flag **no consumer ever read**. That dependency meant every write to `passwordPersistAtom` recomputed the atom, so its value became a pending promise again and every reader re-suspended while `isSupportWebAuth()` re-ran its two WebAuthn probes — which is exactly the 150–300ms blank.

`useSettingsConfig()` reads that atom from inside `SettingsTabNavigator`, which has no nearby Suspense boundary, so the suspension reached the lazy page's boundary. On a re-suspend React does **not** unmount: it hides the already-committed subtree with `display: none !important` (`hideInstance`) and shows the fallback — so the entire painted settings page vanished behind the page-level spinner.

This is our bug, not React's: React behaves as designed (and offers `startTransition` as the escape hatch). What we got wrong was making a static device-capability probe into an async atom that depends on a frequently-written one, and reading it where the nearest boundary blanks the whole page.

The old code already hinted at it; its comment read *"handling webAuthCredentialId in suspense causes the parent container to re-render and flicker"*, but only Native got an early return.

## Fix (commit 1)

- Drop the runtime dependency from `passwordWebAuthInfoAtom`, along with the dead `isEnable` field. All five readers destructure only `isSupport`; the target atom is referenced nowhere outside its own definition.
- Memoize `isSupportWebAuth()` — a device capability cannot change within a runtime, and it was being re-probed on every recompute.

No storage impact: this is a computed atom (`makeCrossAtom('')`), so it is unnamed, unregistered and unpersisted — no schema or `LOCAL_DB_VERSION` concern.

### Verification

Driven against the **built extension** loaded in Chrome via Playwright, with a wallet imported, clicking Settings while the home was still loading, sampling the modal every animation frame:

| | `content → loading → content` |
|---|---|
| baseline (fix stashed, rebuilt) | **6 / 6** |
| this PR | **0 / 20** |

After the fix the modal opens straight to content — the first loading state is gone too, because that initial suspension came from the same atom.

## Audit + hardening (commit 2)

Swept every computed jotai atom (32 definitions) and all context atoms. Almost all use **synchronous** reads, which never suspend, so their dependencies are harmless. Context atoms: no async reads at all. Exactly three async ones existed, all in the password area:

| atom | runtime dep | awaits per recompute |
|---|---|---|
| `passwordWebAuthInfoAtom` | ~~passwordPersistAtom~~ | 2 WebAuthn probes — **commit 1** |
| `passwordBiologyAuthInfoAtom` | settingsPersistAtom | dynamic `import()` + 2 biometric probes |
| `systemIdleLockSupport` | passwordPersistAtom | **nothing** (gratuitous `async`) |

> [!IMPORTANT]
> Commit 2 is **hardening, not a fix for an observed symptom**. I predicted that changing theme/currency inside the settings modal would blank it the same way, then tried to reproduce it — theme change, currency change, and a raced variant — and **could not**. Their awaited work resolves too fast on this platform to be visible. They are the same trap, and the webauth atom showed what happens once that work gets slow, so they are worth closing; but nothing user-visible is being fixed here.

- `passwordBiologyAuthInfoAtom`: same shape — the `settingsPersistAtom` dependency existed only to derive `isEnable`. Its single consumer (`PasswordVerifyContainer`) already reads both halves, so it derives `isEnable = isSupport && isBiologyAuthSwitchOn` locally. Semantics are identical to the old expression.
- `systemIdleLockSupport`: awaits nothing, so the `async` only bought a one-tick suspension on every `passwordPersistAtom` write. Now synchronous, keeping the `boolean | undefined` shape callers already saw.

> [!NOTE]
> For both atoms, `isEnable` meant "supported **and** switched on". Anyone needing it should derive it at the call site rather than reintroducing the dependency inside the async atom.

## Checks

`yarn agent:check --profile pr` passes. Tests: 62 suites / 385 tests across Password, webAuth, biology and Settings. Unlock, settings open and theme switching re-verified against the rebuilt extension; the original repro re-run after commit 2 (0/6).

[OK-53013]: https://onekeyhq.atlassian.net/browse/OK-53013?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

packages/shared/src/webAuth/index.ts
```diff
@@ -49,18 +49,28 @@ const isCMA = async () => {
...
+      if (!platformEnv.isE2E && isContextSupportWebAuth) {
+        const isUvPaaAvailable =
+          await isUserVerifyingPlatformAuthenticatorAvailable();
+        const isConditionalMediationAvailable = await isCMA();
+        isSupport = isUvPaaAvailable && isConditionalMediationAvailable;
+      }
 
-  const finalSupport = isSupport && !!navigator?.credentials;
+      return isSupport && !!navigator?.credentials;
+    })();
```
sidmorizon: 🟡 P2: Probe failures remain cached

Severity: `non-severe`

When a WebAuthn probe rejects, `isSupportWebAuthPromise` retains the rejection. Later registration and verification fail until the runtime restarts.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1b27f1b9-96e5-5286-8555-40136dfdb6d0)

sidmorizon:
Claude review session: https://claude.ai/code/session_012aHmscuDWsKJJYhKR9T7TF

packages/shared/src/webAuth/index.ts
huhuanming: Good catch, and it was a regression I introduced with the memoization in this PR — before it, a rejection propagated per call and the next call retried; caching made it permanent and would have taken `verifiedWebAuth` down with it until a runtime restart.

Fixed in 33cffd3: only a *resolved* answer is cached now. On rejection the slot is cleared (guarded so a later retry's promise isn't clobbered) and the next caller probes again — the same pattern `LazyLoad` and `LazyPopover` already use for their cached load promises.

Added `packages/shared/src/webAuth/index.test.ts` to cover it; I verified the new test fails against the previous behaviour and passes against the fix.

originalix (approved):

PR #13368, opened 2026-09-13
fix: restore mobile market banner and header interactions

limichange:
Fix Discover → Market headers on iOS and Android when empty Favorites replaces the loading list. The banner stays visible on first entry, dragging from the banner or category bar scrolls the active page, and the Android tab underline snaps to the selected category without bouncing back.

- iOS: observe completed Fabric mounts to reattach headers after list replacement; include the shared native header in RN ScrollView hit testing. If no scroll view exists, release the observer and return headers to the pager. Retry an empty-page search only for Insert/Remove mutations inside the current page; repair an orphaned header immediately. Deduplicate changed parent tags, then traverse the page once, so unrelated mutation bursts cost O(mutations + page nodes) rather than one recursive page search per mutation.
- Android: forward header gestures to either RecyclerView or RN ScrollView. Keep empty Favorites content slightly taller than the extended pager viewport so Android also intercepts drags when recommendations fit on screen. Preserve the original RecyclerView-only search for existing inset/offset consumers; header gesture targets use a separate ScrollView/RecyclerView search.

Android underline root cause: ViewPager2 reports a programmatic target selection before the page animation sends its intermediate scroll positions. Both callbacks were writing the native tab progress, so the indicator and tab-row centering moved to the target, back to an intermediate position, and then to the target again. Update the indicator only from page selection. Disabling the underline animation is intentional for both taps and swipes; page swipe gestures remain enabled. The earlier instant-tab behavior in #12842 belonged to the old tab container; #13272 replaced that container and restored animated page commands, and #13350 enabled the native tab bar.

Validation:
- 11 iOS native-method harness cases passed, including no-list fallback, nested list insertion, orphaned headers, drag deferral, and 100 property-only commits without a scroll-owner search. The initial implementation fails 4 original cases; the previous mutation gate performs 300,300 tag reads for 300 unrelated insertions into a 1,001-node page, versus 1,001 with this change. This measures traversal work, not device frame rate. The harness extracts the actual Objective-C++ methods and uses Foundation view doubles.
- 6 Android native-method harness cases passed: RecyclerView after/inside a ScrollView, outer gesture ownership, horizontal-list exclusion, and empty Favorites. The pre-fix lookup fails the mixed-sibling case. These execute the actual Kotlin search methods with view doubles.
- `yarn agent:check --profile commit` passed, including lint, formatting, and TypeScript.
- 48 tests passed across native Market lists, mobile layout utilities, and tab selection guards.
- Applied the generated dependency patch with Git and patch-package against pristine sources; confirmed the compiled Android class includes the new scroll target handling.
- Android API 36 emulator: frame-by-frame comparison of Robinhood → Trending → Favorites confirms the underline no longer returns toward an intermediate page after selection; also verified Stocks/Robinhood taps, horizontal paging, and collapsed-header tab selection.
- Android header regression: verified cold entry with 9 real banners and empty Favorites, banner/category drags, body scrolling and sticky categories, category taps, horizontal page changes, banner carousel gestures, and RecyclerView header scrolling.
- iPhone 17 Pro simulator, rebuilt with the final mutation-filter fix (`e5e7e51e0b`): verified cold first entry with empty Favorites, 9 fetched banners, category selection before/after collapse, and dragging the category bar to expand the header. The native shell and vendor used local-build fallback after remote HTTP 404 responses.

- Final Android review regression on Pixel Fold API 36: rebuilt native app shows 9 fetched banners; verified empty Favorites banner dragging, switching to Trending while partially collapsed, full collapse from the category bar, and expanding back to the banner. Native shell and vendor used local-build fallback after remote HTTP 404 responses.
- `yarn agent:check --profile pr`: all local checks passed; remote CI and review approval remain tracked in the PR.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-d87ae2c4-00c3-5fc2-8872-9d651cb15d7e)

sidmorizon:
Claude review session: https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -0,0 +1,54 @@
...
++
++// OneKey patch: a loading list can be replaced without changing the pager's
++// bounds or page slots. Reattach after the new scroll view has finished mounting.
++- (void)mountingTransactionDidMount:(const MountingTransaction &)transaction
++               withSurfaceTelemetry:(const SurfaceTelemetry &)surfaceTelemetry
++{
++  if (_isBeingRecycled || self.window == nil || _transitioning || _isPagerDragging ||
++      _currentIndex < 0 || _currentIndex >= _pageControllers.count) return;
++  UIView *page = _pageControllers[_currentIndex].view;
++  if (_observedScrollView != nil && [_observedScrollView isDescendantOfView:page]) return;
```
sidmorizon: # 🟡 P2 Orphaned shared header is detected but never re-homed

## Problem

The new observer detects that `_observedScrollView` has left the active page, then delegates the whole repair to `attachScrollObserverForCurrentPage`. That method bails at `if (candidate == nil) return;` (patched source `RNCCollapsiblePagerViewComponentView.mm:1798`) whenever `findVerticalScrollViewInView:` (line 1767) finds nothing — it requires a `UICollectionView`, or a non-horizontal `UIScrollView`.

In that branch nothing re-parents `_sharedHeaderHostView`. It is a **strong** ivar (line 920) that `attachSharedHeadersToScrollView:` added as a subview of the *previous* scroll view (`[scrollView addSubview:_sharedHeaderHostView]`, line 1263). `_observedScrollView` is `__weak` (line 965), so once Fabric recycles or drops the old `RCTScrollViewComponentView` the header stays parented inside a view that is no longer in the window. `detachScrollObserver` (line 1833) only clears KVO, `restoreDetachedScrollInsets` (line 2057) only restores insets, and `layoutSharedHeaders` (line 1289) only sets frames — `restoreSharedHeadersToContainer` (line 1238) is the only thing that re-homes the host, and this path never calls it.

That the state is reachable is confirmed by this PR itself: the new `betterHitTest:` guards on exactly it (`[_sharedHeaderHostView isDescendantOfView:self]`, line 33 of this patch).

## Impact

When the replacement page is not itself a scroll view (a plain empty-state / error view rather than an RN `ScrollView`), the banner and category bar stay invisible and non-interactive — the same symptom this PR is fixing — and the observer re-runs on every subsequent commit without ever converging.

## Recommendation

Make the new observer converge on its own instead of relying on a helper that can no-op:

- Add `|| [_sharedHeaderHostView isDescendantOfView:self]` to the guard on line 48 of this patch, so an orphaned host forces a reattach even while `_observedScrollView` is still a live descendant of `page`.
- After `[self attachScrollObserverForCurrentPage]`, fall back to `[self restoreSharedHeadersToContainer]` when `[_sharedHeaderHostView isDescendantOfView:self]`, so the header is always re-homed in `_containerView` when no list can own it.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -0,0 +1,54 @@
...
++      ![_sharedHeaderHostView isDescendantOfView:self]) {
++    return hitView;
++  }
++  CGPoint headerPoint = [_sharedHeaderHostView convertPoint:point fromView:self];
++  return [_sharedHeaderHostView hitTest:headerPoint withEvent:event] ?: hitView;
++}
++
++// OneKey patch: a loading list can be replaced without changing the pager's
++// bounds or page slots. Reattach after the new scroll view has finished mounting.
++- (void)mountingTransactionDidMount:(const MountingTransaction &)transaction
```
sidmorizon, line 209: # 🟡 P2 Observer ignores the transaction and does unbounded work on every surface commit

## Problem

`transaction` is never read. `RCTMountingTransactionObserverCoordinator::notifyObserversMountingTransactionDidMount` (RN 0.86.2, `React/Fabric/Mounting/RCTMountingTransactionObserverCoordinator.mm`) looks up observers by `transaction.getSurfaceId()` and calls **every** registered observer in that surface, for **every** transaction — not only for mutations under this pager. RN's own `RCTMountingTransactionObserving.h` opens with: *"Observers of this protocol are being called several times on every single mount transaction. Any thoughtless or suboptimal implementation of this protocol will slow down the whole app."*

In the exact state this PR targets — the active page has no attachable list yet, so `_observedScrollView` is `nil` — the guard on line 48 of this patch falls through on every commit, and `attachScrollObserverForCurrentPage` then runs, unthrottled and with no backoff:

- `restoreDetachedScrollInsets` (`RNCCollapsiblePagerViewComponentView.mm:2057`) — `O(tracked scroll views × pages)` with an `isDescendantOfView:` ancestor walk per pair;
- `findVerticalScrollViewInView:` (line 1767) — a full recursive traversal of the page's native view tree, returning `nil` and leaving nothing cached.

This is on top of the existing `layoutSubviews` path (line 1511-1516), which already schedules the same call — but only when layout is actually dirtied. The new path adds every layout-free commit in the surface, which on the Market screen includes streaming price updates.

## Impact

Sustained main-thread cost on Discover → Market whenever the visible page has no attachable scroll view, scaling with commit rate across the whole surface rather than with anything this pager did.

## Recommendation

Gate on the transaction instead of ignoring it: walk `transaction.getMutations()` and return early unless some `Insert`/`Remove`/`Create`/`Delete` mutation's shadow view lies in this pager's tag subtree. If that is too invasive for a dependency patch, cache the "no vertical scroll view found" result and invalidate it only from `mountChildComponentView:`/`unmountChildComponentView:`/`layoutSubviews`, so the traversal is not repeated for unrelated commits.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
sidmorizon: Correction — the leading `!` (logical NOT) was stripped from three code spans in the comment above, which inverts the meaning. Read them as **negated**:

- The `betterHitTest:` guard this PR adds on line 33 of the patch is `if (... || NOT [_sharedHeaderHostView isDescendantOfView:self]) { return hitView; }` — i.e. it bails when the host is **not** a descendant.
- Recommendation bullet 1: extend the guard on line 48 with `|| NOT [_sharedHeaderHostView isDescendantOfView:self]`.
- Recommendation bullet 2: fall back to `[self restoreSharedHeadersToContainer]` when `NOT [_sharedHeaderHostView isDescendantOfView:self]`.

Substituting `NOT` for `!` above; the analysis and impact are unchanged.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -0,0 +1,212 @@
...
+index 6b20a94..1387eda 100644
+--- a/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/CollapsiblePagerViewManager.kt
++++ b/node_modules/react-native-pager-view/android/src/main/java/com/reactnativepagerview/CollapsiblePagerViewManager.kt
+@@ -43,7 +43,8 @@ class CollapsiblePagerViewManager : ViewGroupManager<CollapsiblePagerHost>(),
+     )
+     host.pager.registerOnPageChangeCallback(object : ViewPager2.OnPageChangeCallback() {
+       override fun onPageScrolled(position: Int, offset: Float, offsetPixels: Int) {
+-        host.updateNativeTabProgress(position, offset)
++        // OneKey patch: snap the indicator on selection; scroll progress can pull it back after a tab press.
++        // host.updateNativeTabProgress(position, offset)
```
originalix, line 154: P2: [Preserve drag-time native tab indicator updates]

When an Android user swipes between Market tabs, `ViewPager2` emits `onPageScrolled` updates throughout the gesture. Commenting out `updateNativeTabProgress` means the native indicator retains its previous position until a later selection callback, so it no longer follows the swipe and can jump at completion.

Keep progress updates for user-driven pager drags, and suppress or reconcile only the programmatic scroll callbacks produced by a native tab press.

sidmorizon: Independently traced this against the pristine 3.0.123 Android sources — **confirming this finding, and I'd rate it P1 rather than P2.** Not posting a separate comment since this is the same root cause.

Evidence:

1. **The indicator has exactly one input left.** `grep` over all 9 android sources: `updateNativeTabProgress` has only two call sites, `CollapsiblePagerViewManager.kt:46` (`onPageScrolled` — the line this patch comments out) and `:52` (`onPageSelected`). `CollapsiblePagerHost.kt:220` is the only caller of `CollapsiblePagerNativeTabBarView.setProgress`, which is the only writer of the `progress` field (`CollapsiblePagerNativeHeaders.kt:171-177`).

2. **Everything visible derives from that field.** The tab row is built as `CollapsiblePagerHorizontalItemsView(context, true)` (`CollapsiblePagerNativeHeaders.kt:296`), so `showsProgressIndicator` is `true` and `updatePresentation()` takes the progress branch (`:246-268`): indicator X/width interpolation, `ColorUtils.blendARGB` label colour, `selectedIndex`, and the `centerSelectionWhenIdle` auto-scroll. The `selectedKey` branch never applies to this row, so `selectPage()` / `updateSelectedKey()` do not rescue it.

3. **No surviving continuous feed.** `observedScrollListener` (`CollapsiblePagerHost.kt:552-566`) is a vertical-list listener that only writes `pageOffsets`/`setHeaderOffset`; the nested-scroll callbacks only move `headerOffsetPx`; and although `PageScrollEvent` is still dispatched, `CollapsiblePagerViewNativeComponent.ts` exposes no progress prop and `MobileLayout.native.tsx` never wires `onPageScroll`.

4. **Timing — one refinement to your wording.** Per androidx `ScrollEventAdapter`, a manual drag does not dispatch `onPageSelected` mid-gesture: DRAGGING→SETTLING only sets `mDispatchSelected = true`, and `dispatchSelected(mTarget)` runs on the *next* `onScrolled`, guarded by `mDragStartPosition != mTarget`. So the snap lands at **finger-release**, while the page is still settling — not at completion. The underline and label colours therefore sit frozen on the old tab for the whole swipe, then jump ~200-400 ms before the page finishes.

5. **The author's premise is also correct**, which is why the fix needs narrowing rather than reverting: programmatic `setCurrentItem(pos, true)` goes through `notifyProgrammaticScroll`, which calls `dispatchSelected(target)` immediately, *before* any scroll event — that is the real source of the target → intermediate → target bounce.

Concrete repair matching your suggestion: keep the `onPageScrolled` feed but gate it on a drag flag set in `onPageScrollStateChanged` — `true` on `SCROLL_STATE_DRAGGING`, cleared on `SCROLL_STATE_IDLE` — and call `host.updateNativeTabProgress(position, offset)` only while it is set. A programmatic selection goes IDLE→SETTLING→IDLE and never enters DRAGGING, so the bounce stays fixed while drags keep tracking.

(Side note: for a drag released below the snap threshold, `mDragStartPosition == mTarget` means no `onPageSelected` fires at all — under this patch the indicator simply never moves, which is actually cleaner than pristine. It does not help the forward-swipe case.)

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -0,0 +1,212 @@
...
+   private var nativeGestureStarted = false
+   private var pressCancelled = false
+   private val pageContentLayoutListener = ViewTreeObserver.OnPreDrawListener {
+@@ -503,7 +506,13 @@ class CollapsiblePagerHost(context: Context) : NestedScrollableHost(context), Ne
+     return false
+   }
+ 
+-  private fun findVerticalRecyclerView(view: View): RecyclerView? {
++  private fun findVerticalRecyclerView(view: View): RecyclerView? =
++    findVerticalScrollableView(view) as? RecyclerView
```
sidmorizon: # 🟡 P2 `findVerticalRecyclerView` silently loses its contract — aborts at the first `ScrollView`

## Problem

Redefining `findVerticalRecyclerView(view) = findVerticalScrollableView(view) as? RecyclerView` changes the semantics of a helper that four pre-existing call sites depend on.

`findVerticalScrollableView` returns at the **first** DFS match, and it tests `if (view is ScrollView) return view` **before** the `RecyclerView` branch (patch lines 31-33). So if any page subtree contains a vertical `ScrollView` that is visited before the list, the outer `as? RecyclerView` yields `null` and the search stops — where the pristine recursion (`CollapsiblePagerHost.kt:506-519`) would have kept descending and found the `RecyclerView`. Note `ReactScrollView extends android.widget.ScrollView`, so any RN `<ScrollView>` in a page matches.

The affected pre-existing consumers all reach it through `recyclerViewForPage` (`CollapsiblePagerHost.kt:522-524`) or the direct call at `:309`: `prepareAdjacentPages()` → `applyRecyclerInsets()` (the header/sticky top inset), `attachRecyclerObserver()` (header-collapse sync and saved scroll-offset restore), `selectPage()`'s saved-offset lookup, and `removeReactChild()` → `restoreRecyclerInsets`. A miss is silent — `null` is a legitimate return, so nothing logs or throws.

## Impact

Latent, and I want to be precise rather than overstate it: I could **not** find a page that triggers this today. `CollapsiblePagerView` has a single consumer (`MobileLayout.native.tsx`), and the only `ScrollView`s inside pages are `MobileMarketNativeLists.tsx:1029` / `:1033`, neither of which contains a `RecyclerView` (`MarketRecommendList` renders plain views). So there is no user-visible regression in this diff.

The defect is that a shared helper quietly narrowed its contract inside a patch-package patch — the kind of change that has to be re-reasoned on every dependency bump. The first Market page that wraps a list in a `ScrollView` loses its header inset (first rows rendering under the collapsible header) and its saved scroll offset, with no error to trace it back here.

## Recommendation

Keep the two searches independent rather than layering a cast on top of a broadened traversal. Either restore the original body of `findVerticalRecyclerView` (recursing for `RecyclerView` only) and let `findVerticalScrollableView` be its own traversal, or parameterise one traversal with a `predicate: (View) -> Boolean` so each caller states what it is looking for. Either keeps the new header-gesture behaviour without changing what the four existing inset/offset call sites resolve to.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
sidmorizon: Fixed in `086ddf2e` — resolving.

Verified against the patched source at `a3d7d3f2`, not just the commit message. Both halves of the repair landed:

- Patch lines 205-211 add the orphan fast path: `BOOL headerIsAttached = [_sharedHeaderHostView isDescendantOfView:self];` is now part of the early-return condition, and a stranded host (`headerIsAttached` false) forces `attachScrollObserverForCurrentPage` immediately.
- Patch lines 232-238 replace `if (candidate == nil) return;` with `{ [self detachScrollObserver]; [self restoreSharedHeadersToContainer]; return; }`.

Re-traced the scenario that previously stranded the header — list unmounted, host still parented under the removed scroll view, replacement page with no vertical scroll view — and it now converges: the orphan branch fires, `findVerticalScrollViewInView:` returns nil, and `restoreSharedHeadersToContainer` re-parents into `_containerView` (which is `self.contentView` and survives `prepareForRecycle`). On the next commit `headerIsAttached` is true, so the branch is not re-entered. No re-entrant loop: `attachSharedHeadersToScrollView` only refuses when `_nativeSmoothHeaderScrollEnabled` is off, a mode where the header never leaves `_containerView` anyway.

Also checked the `_pageOffsets` bookkeeping the new `detachScrollObserver` call introduces on the nil path — it is safe, because `attachScrollObserverForCurrentPage` opens with `restoreDetachedScrollInsets`, which has already detached and written `_pageOffsets[key] = _headerOffset` for any scroll view no longer under a page controller, leaving the new call a no-op.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
sidmorizon: Fixed in `a3d7d3f2` — resolving.

Verified at `a3d7d3f2`: the patch no longer modifies `findVerticalRecyclerView` at all. The hunk is now `@@ -518,6 +521,23 @@`, a pure addition of a separate `findVerticalScrollableView`, and I diffed the resulting `findVerticalRecyclerView` against the pristine 3.0.123 tarball — byte-identical to upstream. `grep` for `as? RecyclerView` over the patch returns nothing, so the unchecked cast is gone too.

The two searches are now independent, which is exactly the split this was asking for:
- `CollapsiblePagerHost.kt:509-520` — `findVerticalRecyclerView`, RecyclerView-only DFS (upstream, untouched), still what `recyclerViewForPage` and the direct `restoreRecyclerInsets` call site use, so all four pre-existing inset/offset consumers are back on their original contract.
- `CollapsiblePagerHost.kt:524-539` — `findVerticalScrollableView`, new, ScrollView-or-RecyclerView, reached only from `beginForwardingToScrollable`.

The previously failing case — a page whose subtree hits a `ScrollView` before the `RecyclerView` — now returns the RecyclerView to the inset/offset callers and the ScrollView only to the header-gesture forward.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
sidmorizon: # 🟡 P2 Partially fixed in `086ddf2e` — leaving open: the gate is now quadratic in the state it guards

## What is fixed

`transaction` is now read (patch lines 212-221), and the steady state is genuinely bounded: with a valid observed scroll view under the current page and the header attached, the observer returns after two `isDescendantOfView:` walks instead of `restoreDetachedScrollInsets` plus a full recursive `findVerticalScrollViewInView:`. That removes the dominant cost from every unrelated surface commit, which was the main point. I also verified the C++ against RN 0.86.2 rather than assuming: `ShadowViewMutation::Type` is an unscoped nested enum, `parentTag` is a public `Tag` defaulting to `-1` so the `> 0` guard is right, `getMutations()` binds the const-ref overload, and `[page viewWithTag:]` resolves because Fabric sets `view.tag = tag` on dequeue (`RCTComponentViewRegistry.mm:53`).

I confirmed the gate is not so tight that the original bug returns — loading-ScrollView to real list, non-scrollable loading state to list, and list to empty Favorites all still converge, the first and third via the new orphan fast path before the mutation loop is even consulted.

## What still stands

The gate replaced one bounded walk with an unbounded one in exactly the state it exists to protect. `-[UIView viewWithTag:]` is a full depth-first search of the receiver's subtree, and on a miss it visits every descendant. It runs **inside** the per-mutation loop, so the gate costs `O(#Insert/Remove mutations × size of the current page's subtree)` — and that loop only runs when the current page has no vertical scroll view, i.e. the loading window this patch targets, not an exceptional case.

Concretely: during the Market tab's loading window a hydration commit carrying a few hundred `Insert` mutations, against a skeleton subtree of comparable size, is on the order of 10^4-10^5 `UIView` visits, synchronously on the main thread inside a Fabric mount transaction, repeated per commit — on the tab-open path. The result is correct; the cost profile is not.

So this is better than before for quiet commits and worse for mutation bursts, which is why I am not resolving it.

## Recommendation

Walk the subtree at most once per transaction instead of once per mutation:

- Collect `page`'s descendant tags into a `std::unordered_set<Tag>` before the loop and test membership in O(1), building it only when the mutation list actually contains an `Insert`/`Remove`; or
- Walk **up** from the mutation's resolved component view to `page` (O(depth)) rather than down from `page`.

Confidence is high on the algorithmic claim and medium on the magnitude — it depends on how large the Market loading skeleton actually is, and I have not measured on-device.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

patches/react-native-pager-view+3.0.123.patch
limichange: The callback analysis is correct, including the selection update during settling rather than after the page animation completes.

The intended behavior has been explicitly reconfirmed: disable Android underline animation for both tab taps and swipes. Holding the underline and label state during a drag, then snapping on `onPageSelected`, is expected. `onPageSelected` will remain the only native progress writer; we will not restore the drag-time `onPageScrolled` feed.

Horizontal paging and `PageScrollEvent` dispatch remain enabled. Android API 36 emulator verification covered category taps, horizontal swipes, and final selected-state correctness, and confirmed that the target → intermediate → target bounce is gone. The PR description also documents this interaction decision.

patches/react-native-pager-view+3.0.123.patch
limichange: Fixed in `e5e7e51e0b`. I also checked the latest head (`9de0777d`): the pager patch is byte-identical after the merge from `x`.

The mutation loop no longer calls `viewWithTag:`. It collects valid Insert/Remove parent tags into a local `std::unordered_set`, skips traversal when that set is empty, and walks the current page once using an explicit stack. A matching parent triggers header recovery. Filtering now has expected O(mutations + page nodes) cost. The healthy-owner and orphan-header fast paths remain intact.

Validation on `e5e7e51e0b`:
- All 11 native-method harness cases passed, including nested list insertion after an unrelated mutation burst, property-only updates, and orphan recovery.
- For 300 unrelated insertions and a 1,001-node page, tag reads dropped from 300,300 with the previous gate to 1,001. This uses the actual Objective-C++ methods with view doubles; it measures traversal work, not on-device frame rate.
- The iOS native build and iPhone 17 Pro simulator checks passed: first entry with empty Favorites and real banners, header collapse, switching to Trending while collapsed, and expansion from the category bar.

Please re-check the updated gate and resolve this thread if this addresses the remaining concern.

sidmorizon: Fixed in `e5e7e51e` — resolving.

Re-verified at the current head `9de0777d` (I confirmed independently that this is a pure base merge: the patch blob is byte-identical to `e5e7e51e`, the PR's contribution against the new merge base `eae89221` is the same 3 files / +269, and `origin/x` is now an ancestor). Applied the patch to a fresh pristine 3.0.123 and read the result rather than relying on the commit message.

**Complexity is genuinely fixed.** The mutation loop now only does a type test and a hash insert; the page is traversed exactly once, with `changedParentTags.empty()` short-circuiting before any traversal so property-only commits keep their zero-traversal cost. `O(mutations + page nodes)`, and no path reintroduces per-mutation traversal.

**Semantically equivalent to the old gate.** The old and new predicates are the same existential with the quantifiers swapped — same views visited, same tags tested. Two details I checked rather than assumed:
- The traversal is seeded with `arrayWithObject:page`, so it includes `page` itself, matching `-[UIView viewWithTag:]` testing the receiver first.
- `page` is a bare `UIViewController.view` with `tag == 0`, and the `parentTag > 0` filter keeps 0 out of the set, so the tag-0 root is a non-match under both versions. Only sibling visit order differs (stack-pop reverses it), which is irrelevant to an existence test.

**No new defects.** Termination is guaranteed — the UIKit view hierarchy is a strict tree, so the loop pops exactly N nodes; `lastObject`/`removeLastObject` are guarded by the count check and `view` is a strong local under ARC. `Tag` is `int32_t` and both `UIView.tag` and the set element are `NSInteger`, so the conversion is widening and sign-preserving with no truncation collisions. `#include <unordered_set>` sits at file scope before `using namespace facebook::react;` in a `.mm`, which is valid.

**Nothing already agreed fixed regressed:** the `headerIsAttached` orphan fast path and the `candidate == nil` fallback (detach + restore) are both intact, and `findVerticalRecyclerView` is still byte-identical to upstream with `findVerticalScrollableView` as a separate function — I diffed both against the pristine tarball again.

On the "300,300 vs 1,001" figure: it reproduces from the code as written. Worth noting for anyone reading it later that it counts `page` itself as one of the 1,001 nodes, and counts only tag comparisons — it excludes the 300 `unordered_set` insertions the new path adds. That does not change the conclusion.

[Claude session ↗](https://claude.ai/code/session_01U8EPC3Bdkuqk731Vhw66Xe)

---

originalix (approved):

PR #13369, opened 2026-09-13
fix: preserve native spinner colors OK-62489

huhuanming:
[OK-62489](<https://onekeyhq.atlassian.net/browse/OK-62489>)

----------

## Summary

- make the custom Spinner a `YStack.styleable` component so Tamagui color tokens reach the native activity indicator
- restore variant-specific loading colors for Button, IconButton, and direct Spinner consumers

## Validation

- reproduced and verified Primary/Secondary button loading states on an isolated iOS simulator
- confirmed native `UIActivityIndicatorView` colors: Primary `#FCFCFC`, Secondary black at 60.8% alpha

[Jira OK-62489](https://onekeyhq.atlassian.net/browse/OK-62489)

[OK-62489]: https://onekeyhq.atlassian.net/browse/OK-62489?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_017NbEXZ2E1PUXBp8aTK7qop

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-e9b9f705-e573-56e0-8679-409a8ab36358)

packages/components/src/primitives/Spinner/Spinner.tsx
```diff
@@ -21,23 +19,17 @@ export type ISpinnerProps = Omit<YStackProps, 'children'> & {
...
-      <YStack ref={ref} {...stackProps}>
-        <ActivityIndicator size={size} color={color} />
-      </YStack>
-    );
-  }),
-  {
-    componentName: 'Spinner',
-  },
-) as any;
+> = YStack.styleable((props: ISpinnerProps, ref) => {
```
sidmorizon, line 22: # 🟡 P2 Load-bearing `styleable` contract is undocumented and erased by `as any`

## Problem

The callback body here is byte-identical to the pre-change version — `git diff e38216d8..0e43de57` shows the only real change is `themeable(forwardRef(fn), { componentName: 'Spinner' })` → `YStack.styleable(fn)`. So the entire fix for OK-62489 lives in that HOC swap, and its mechanism is in a *different* file: `Spinner/index.tsx:6` wraps this component in `styled(OriginSpinner, { color: '$icon' })`, and `styled()` only forwards `color` as a prop (instead of collapsing it into `style`, since `color` is a valid View style key) when the wrapped component carries `staticConfig.isHOC`. `YStack.styleable()` attaches that config; a bare `forwardRef` does not, which is why `colorProp` was always `undefined` and every spinner pinned to `$icon` on native.

Nothing in either file records this. Worse, the `as any` on line 35 plus the hand-written `React.ForwardRefExoticComponent<…>` annotation on lines 20-22 strip `staticConfig` from the declared type, so `styled(OriginSpinner, …)` at `index.tsx:6` type-checks *identically* before and after this fix. (The `as any` predates this PR, but it is now hiding a contract the fix depends on.) There is also no test for `Spinner` anywhere in the repo — `Spinner.stories.tsx` is the only sibling.

## Impact

A future "cleanup" back to `forwardRef`/`themeable` (both still exported from `shared/tamagui.ts:35`), or removal of the `styled()` wrapper in `index.tsx`, silently reintroduces OK-62489 across every colored spinner — `Button/index.tsx`, `IconButton/index.tsx`, `TxHistoryListView`, `SwapOrderProgress`, `OpenOrdersRow`, `NFTDetails`, and ~200 other call sites — with zero compile-time or test signal. The regression is native-only (web masks it via CSS vars/`currentColor`), so it is unlikely to be caught in local dev, exactly as happened the first time.

## Recommendation

1. Add a short comment above line 22 stating *why* `styleable` is required: it attaches `staticConfig.isHOC`, which makes the `styled()` wrapper in `Spinner/index.tsx:6` forward `color` as a prop rather than merging it into `style`; a plain `forwardRef` swallows the color and pins every spinner to `$icon` on native.
2. Drop the `as any` on line 35 and the hand-written annotation on lines 20-22, letting `styleable`'s return type flow through so the `staticConfig` this fix relies on is actually present in the type.
3. Add a regression test rendering `<Spinner color="$iconInverse" size="small" />` and asserting `ActivityIndicator` receives the resolved `iconInverse` color, not the `icon` one.

[Claude session ↗](https://claude.ai/code/session_017NbEXZ2E1PUXBp8aTK7qop)

---

originalix (approved):

PR #13370, opened 2026-09-13
fix: support ScrollView indicators on web (OK-58089)

huhuanming:
[OK-58089](<https://onekeyhq.atlassian.net/browse/OK-58089>)

----------

## Summary
- honor the existing ScrollView indicator props on web, desktop, and extension UI
- keep web scrollbars hidden unless the active axis is explicitly enabled
- preserve native prop behavior and caller-provided data attributes
- add focused regression coverage for horizontal, vertical, dynamic, disabled, and native behavior

## Verification
- `yarn jest packages/components/src/layouts/ScrollView/ScrollView.test.tsx --runInBand`
- Chromium rendering smoke test: visible scrollbar resolves to 8px/auto; hidden resolves to 0px/none while overflow remains scrollable

Jira: https://onekeyhq.atlassian.net/browse/OK-58089

[OK-58089]: https://onekeyhq.atlassian.net/browse/OK-58089?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5e363da3-15d3-51c6-8a4f-a0ae74000ee0)

sidmorizon:
Claude review session: https://claude.ai/code/session_013196KdxB6U8xFWov5KVojV

packages/components/src/layouts/ScrollView/ScrollView.test.tsx
```diff
@@ -0,0 +1,151 @@
...
+      children,
+    ),
+  );
+
+  MockScrollView.displayName = 'MockScrollView';
+
+  return {
+    ScrollView: MockScrollView,
+  };
+});
```
sidmorizon: # 🟡 [P2] The new tests mock away the one contract this fix depends on

## Problem

The whole fix rests on a two-sided contract: `index.tsx:85` emits the `dataSet` key `onekeyScrollViewScrollbar`, react-native-web hyphenates it to the DOM attribute `data-onekey-scroll-view-scrollbar`, and `packages/shared/src/web/index.css:94-110` selects on that exact attribute string.

The `react-native` mock at lines 10-37 hardcodes both sides of that mapping itself:

```tsx
'data-onekey-scroll-view-scrollbar': dataSet?.onekeyScrollViewScrollbar,
```

So RNW's hyphenation is never exercised, and the CSS is never loaded (`jest.config.js:59` maps `\.(css|less)$` to `__mocks__/styleMock.js`). Change only the selector in `index.css` to `[data-onekey-scrollview-scrollbar=…]`, and the feature is dead on web while all six tests stay green. The mock also drops every prop except `children`/`dataSet`/`testID`, so the `showsHorizontalScrollIndicator: undefined` / `showsVerticalScrollIndicator: undefined` blanking at `index.tsx:81-82` — the part that stops RNW applying its own `hideScrollbar` — is untested too.

`preserves caller data attributes` (lines 126-138) has the same issue: with `usePropsAndStyle` stubbed to a pass-through at lines 44-50, it only asserts that an object spread works. `index.tsx:120` (`// Preserve testID because RNW drops Tamagui's data-testid rewrite.`) and `packages/components/src/forms/Input/index.tsx:305-308` are this repo's own recorded evidence that Tamagui/RNW do rewrite or drop `data-*`/`dataSet`, so that assertion may not hold in production.

## Impact

The PR describes this as "focused regression coverage", but the suite cannot fail on the only realistic regression: drift between the emitted attribute name and the CSS selector, or `dataSet` being swallowed before it reaches the DOM node. A future rename on either side ships a silently broken scrollbar with a passing test suite — and because the visible/hidden state is purely cosmetic, nothing else would catch it.

## Recommendation

Drop the `react-native` mock and render the real component. `jest.config.js:55` already maps `'^react-native$': 'react-native-web'`, and the file already declares `/** @jest-environment jsdom */`, so RNW's `ScrollView` renders a real DOM node under jsdom with no extra setup. Then assert the shipped attribute rather than a hand-written one:

```tsx
expect(
  container.querySelector('[data-onekey-scroll-view-scrollbar="visible"]'),
).not.toBeNull();
```

Keep the `platformEnv` mock (needed for the native case) and the tamagui mock (`moduleNameMapper` redirects `@onekeyhq/components/*` to `__mocks__/componentsMock.ts`, which has no `usePropsAndStyle`). Ideally export the attribute name once as a shared constant referenced by `index.tsx` and by a comment above the `index.css` rules, so the two sides cannot drift unnoticed.

[Claude session ↗](https://claude.ai/code/session_013196KdxB6U8xFWov5KVojV)

---

huhuanming: Fixed in 9359a12e38. The test now unmocks react-native and renders the configured real react-native-web ScrollView, so RNW's dataSet-to-DOM attribute conversion is exercised. Added coverage proving the opposite false indicator no longer contributes RNW's hide-scrollbar class, and a contract assertion that the emitted DOM attribute matches both shipped CSS selectors. Verified the focused suite (8/8), targeted ESLint, and yarn agent:check --profile commit.

originalix (approved):

PR #13372, opened 2026-09-13
fix: preserve Market trade loading and Earn currency display (OK-62893, OK-62879)

zhaono1:
[OK-62893](<https://onekeyhq.atlassian.net/browse/OK-62893>) [OK-62879](<https://onekeyhq.atlassian.net/browse/OK-62879>)

----------

## Summary

- keep the desktop Market stock trade column mounted while stock identity and decimal metadata are loading
- show a trade skeleton on the first render and stock-id transition, then keep the mounted Swap instance for later variant transitions so the existing stock channel skeletons can render
- keep terminal `disableTrade` and route-ownership behavior unchanged
- render Earn/Pendle token-selector fiat values with the app-selected currency symbol ([OK-62879](https://onekeyhq.atlassian.net/browse/OK-62879))
- refresh and simplify the `1k-trade-swap-market` and `1k-defi-module-integration` skills, remove stale code maps, and retain the native bundle validation gate

## Issues

- Fixes OK-62893
- Fixes OK-62879

### Market stock trade loading

`DesktopLayout` combined transient `!isSwapTokenReady` with `disableTrade`. `MarketEmbeddedSwapDraft` interpreted that transient state as terminal and returned `null`, leaving the fixed-width right column visually blank. The shared Swap and stock-channel loading states were therefore never mounted.

### Earn token selector currency (OK-62879)

`/earn/v1/asset-list` already returns fiat values in the requested currency. The live `EarnTokenSelect` modal formatted every value through a local helper that hard-coded `$`, so CNY-converted values were displayed with a USD symbol. The modal now reads `settings.currencyInfo.symbol` and delegates value formatting to the shared `NumberSizeableText` formatter.

## Validation

- `yarn jest packages/kit/src/views/Market/MarketDetailV2/components/MarketDetailEmbeddedSwap.test.tsx packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.test.ts packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.test.tsx packages/kit/src/views/Market/MarketDetailV2/layouts/StockDesktopLayout.test.tsx --runInBand` — 4 suites, 21 tests passed
- `yarn jest packages/kit/src/views/Staking/pages/EarnTokenSelect/index.test.tsx --runInBand --no-cache` — 1 suite, 2 tests passed
- `yarn tsc:only` — passed
- `yarn agent:check --profile commit` — all local checks passed
- Oxfmt, ESLint, and `git diff --check` — passed
- Electron CDP smoke test: switched AAPL → MSFT → TSLA; the trade loading test ID remained present during transitions and the ready state returned afterward

## Scope

- backend/API conversion contract unchanged; this change is limited to the live Earn token selector presentation
- the normal currency-setting flow refreshes Earn data and restarts the app; live no-restart cache invalidation is a separate follow-up
- no Jira or Slack writes were made

## Skill documentation scope

The existing skill updates consolidate durable workflow contracts from recent conversations and merged PRs, remove path/test/module snapshots, and keep the native bundle validation gate.

## Status

PR kept open for review.

[OK-62879]: https://onekeyhq.atlassian.net/browse/OK-62879?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-62893]: https://onekeyhq.atlassian.net/browse/OK-62893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-3ec8873b-bbb4-5ad2-8a07-87eb679130f2)

sidmorizon:
Claude review session: https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +290,13 @@ export function DesktopLayout({
...
-    hasRenderedStockTradeRef.current = false;
-  }
-  if (shouldUseStockDesktopLayout && isSwapTokenReady) {
-    hasRenderedStockTradeRef.current = true;
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTradeLoading = shouldUseStockDesktopLayout && !isSwapTokenReady;
```
originalix: P2: [Terminal stock variant states render an endless trade skeleton]

When the variant request settles with no tradable token (for example, every variant is paused) or with `isTokenVariantsError`, `isSwapTokenReady` remains false. This predicate then marks the state as loading, so the new initial-load branch mounts `MarketEmbeddedSwapLoading` indefinitely instead of exposing the terminal unavailable or retryable outcome.

Derive this flag from the transient pending state and keep terminal no-variant/error cases disabled or render an explicit unavailable/retry surface.

sidmorizon: # 🟠 P1 Stock trade panel can render a loading skeleton that never resolves

## Problem

`isTradeLoading` is derived purely from `!isSwapTokenReady`, which conflates *transient* token loading with the *terminal* "no tradable token exists" state. The predicate it replaced mapped those same terminal states to `disabled → return null`, and the comment removed from `MarketEmbeddedSwapDraft` said so explicitly: *"Do not present a perpetual loading skeleton for a terminal state."* Nothing replaces that guard.

Verified trace on the current head:

- `DesktopLayout.tsx:177-186` — `stockTokenAddress = selectedTokenVariant?.contractAddress || routeTokenAddress`; `stockDisplayMatchesVariant` short-circuits falsy when that is `''`, so `isSwapTokenReady === false` (`DesktopLayout.tsx:282-292`).
- `MarketDetailV2.tsx:127-128` — `routeTokenAddress` defaults to `''`, and `useToMarketStockDetailPage.ts:72-82` only supplies it when the stock preview carries **both** `tokenAddress` and `networkId`, so the stock-list entry path commonly lands with `''`.
- `utils/stockTokenVariant.ts` — `getDefaultStockTokenVariant` returns `undefined` when no variant passes `isStockTokenVariantTradable` (`tradingEnabled && !isPaused && !tradingHours?.isPaused && status === 'active'`).
- `hooks/StockDetailContext.tsx:217-262` — when the variants request fails the effect returns early at `if (tokenVariantResult?.failed) return;`, and when no variant is tradable `setSelectedTokenId(undefined)`. Either way `selectedTokenVariant` stays `undefined`. In the non-tradable case `isTokenVariantPending`, `isTokenVariantsLoading` **and** `isTokenVariantsError` are all `false` (`:288-301`) — fully terminal, with no polling or retry path out of it.
- `StockDesktopLayout.tsx:1398-1417` — the fixed-width trade column is rendered unconditionally, so the skeleton is always on screen.

Net effect: `isTradeLoading` stays `true` for the life of the page and `MarketEmbeddedSwap.tsx:261-263` keeps returning `<MarketEmbeddedSwapLoading />` (a 520px shimmer) because `hasRenderedContentRef.current` never flips.

## Impact

On a stock whose variants are all paused / outside trading hours / non-`active`, and on a variants-endpoint failure, the desktop stock page shows a permanently animating 380×520 trade skeleton. Pre-PR these states rendered a blank column via `disabled → null`, so this is a blank-area → endless-skeleton regression. The page also contradicts itself: `StockTokenVariantSelector.tsx:363-372` renders a **Retry** button for `isTokenVariantsError` right beside a column claiming it is still loading. This also conflicts with the validation gate added by this same PR in `.skillshare/skills/1k-trade-swap-market/references/validation.md` ("prove the terminal path is explicit rather than an endless skeleton or disabled action").

Note the PR description's "keep terminal `disableTrade` and route-ownership behavior unchanged" does hold — but terminal *token-readiness* behavior did change, and that is the regressing case.

## Recommendation

Gate the skeleton on a genuinely transient signal instead of on `!isSwapTokenReady`. `useStockDetail()` is already destructured at `DesktopLayout.tsx:169` and exposes exactly what is needed:

```ts
const isTradeLoading =
  shouldUseStockDesktopLayout &&
  !isSwapTokenReady &&
  (isTokenVariantPending || isTokenVariantsLoading);
const shouldDisableTrade =
  disableTrade ||
  (!shouldUseStockDesktopLayout && !isSwapTokenReady) ||
  (shouldUseStockDesktopLayout && !isSwapTokenReady && !isTradeLoading);
```

Once variant resolution is terminal (`hasCurrentTokenVariants` with no tradable variant, or `isTokenVariantsError`), fall back to the previous `disabled` behavior — or better, render an explicit unavailable/retry state in the trade column; `TopCoinsDesktopLayout.tsx:681` already sets that precedent instead of returning `null`. `MarketEmbeddedSwap.tsx:258-264` needs no change if the flag is corrected here at line 293.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

zhaono1: Verified on the live head and fixed in 7e53919889.

The skeleton is now gated by the actual transient readiness signals (isTokenVariantPending, isTokenVariantsLoading, and token-detail loading for an already selected variant), and is suppressed for variant request errors. When the variant list settles with no tradable token, the existing disabled path is used instead of mounting the 520px loading shell. The retryable variant-error UI remains owned by the selector.

Added DesktopLayout regressions for terminal no-tradable and variant-error states, plus metadata loading; the focused Market tests pass.

zhaono1: This finding is valid and is fixed in 7e53919889.

isTradeLoading no longer treats every isSwapTokenReady=false state as loading. It is now true only while variant/token-detail readiness is genuinely pending, and terminal no-tradable or variant-error states set disableTrade so the embedded panel returns through the existing terminal path. This removes the blank-area to endless-skeleton regression while preserving the initial and metadata-loading skeleton behavior.

Added regression coverage for both terminal cases and the transient metadata case; focused Market tests and the commit profile checks pass.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,23 @@ export function DesktopLayout({
...
-    hasRenderedStockTradeRef.current = true;
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTradeReadinessPending =
+    !isTokenVariantsError &&
+    (isTokenVariantPending ||
+      isTokenVariantsLoading ||
```
sidmorizon: # 🟠 P1 Terminal "no tradable variant" state flashes the 520px skeleton on every 6s poll tick

## Problem

The terminal fix is only partial. `isTokenVariantsLoading` is counted as "pending" unconditionally, but the variants request **polls**, so that flag goes true again on every refresh tick — not just on cold start.

- `hooks/StockDetailContext.tsx:83` — `STOCK_TOKEN_VARIANTS_POLLING_INTERVAL = 6000`
- `hooks/StockDetailContext.tsx:196-205` — the variants `usePromiseResult` is configured `watchLoading: true` with that `pollingInterval`
- `hooks/usePromiseResult.ts:282` / `:371` / `:374-378` — each poll iteration calls `setLoadingTrue()`, settles with `setLoadingFalse()`, waits `pollingInterval`, then re-enters `run(...)`. With `watchLoading: true` this toggles `isLoading` true→false every tick.

Now take the terminal state this commit is meant to fix — variants resolved, none passes `isStockTokenVariantTradable`, so `selectedTokenVariant` is `undefined` and `isSwapTokenReady` is false permanently:

- **Between ticks:** `isTokenVariantPending` false, `isTokenVariantsLoading` false, `isTokenVariantsError` false → `isTradeReadinessPending` false → `shouldDisableTrade` true (`:315-317`) → `return null`. Blank column. ✅ intended.
- **During each in-flight tick:** `isTokenVariantsLoading` true → `isTradeReadinessPending` true → `shouldDisableTrade` **false** and `isTradeLoading` **true**. `hasRenderedContentRef.current` is still `false` (content has never rendered in this state), so `MarketEmbeddedSwap.tsx:261-262` renders `MarketEmbeddedSwapLoading` — the 520px shimmer.

## Impact

On any stock with no tradable variant, the trade column now pulses between blank and a 520px shimmer once every ~6 seconds for as long as the page is focused. The original permanent shimmer is replaced by a permanent 6-second flicker, which is arguably worse to look at than either steady state. The variants-**error** path is not affected: the previous `failed` result is retained across the retry, so `isTokenVariantsError` stays true and the `!isTokenVariantsError` gate holds. Only the no-tradable-variant branch regresses.

## Recommendation

A background refresh tick should not re-open the pending window — only the absence of a settled result should. `hasCurrentTokenVariants` is already computed at `hooks/StockDetailContext.tsx:212-213`; expose it from the context and gate on it here at `DesktopLayout.tsx:304`:

```ts
const isTradeReadinessPending =
  !isTokenVariantsError &&
  (isTokenVariantPending ||
    (isTokenVariantsLoading && !hasCurrentTokenVariants) ||
    (Boolean(selectedTokenVariant) && isTokenDetailLoading));
```

`isTokenVariantPending` already encodes the cold-start case (`StockDetailContext.tsx:288-294` is true while `!hasCurrentTokenVariants`), so dropping the bare `isTokenVariantsLoading` term costs no cold-start coverage. A regression test should assert that with no tradable variant, toggling `isTokenVariantsLoading` true does **not** flip `disableTrade` to false — the two new cases at `DesktopLayout.test.tsx:328-356` and `:358-387` both pin `isTokenVariantsLoading: false` and so cannot catch this.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,23 @@ export function DesktopLayout({
...
-  }
-  if (shouldUseStockDesktopLayout && isSwapTokenReady) {
-    hasRenderedStockTradeRef.current = true;
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTradeReadinessPending =
+    !isTokenVariantsError &&
```
sidmorizon: # 🟡 P2 Variants-poll failure disables a usable panel whose token detail is still loading

## Problem

`!isTokenVariantsError` gates the **entire** `isTradeReadinessPending` expression, including the `Boolean(selectedTokenVariant) && isTokenDetailLoading` term at `:305`, which has nothing to do with the variants request.

A failed variants poll does not necessarily mean there is no variant. `hooks/StockDetailContext.tsx:184-193` — the `catch` returns the **cached** items with `failed: true`:

```ts
const cachedResult = successfulTokenVariantsRef.current.get(normalizedStockId);
return { stockId: normalizedStockId, items: cachedResult?.items ?? [], ..., failed: true };
```

and the selection effect returns early on `failed` (`:224`), so `selectedTokenVariant` stays **truthy and tradable**. But `isTokenVariantsError` is now true (`:298-302`), which forces `isTradeReadinessPending` false. If that poll tick fails while the selected variant's token detail is still in flight — e.g. shortly after a variant switch — `isSwapTokenReady` is false and `shouldDisableTrade` (`:315-317`) becomes true, so `MarketEmbeddedSwap.tsx:258-260` returns `null` and the panel unmounts even though the variant is cached, tradable, and its detail request is progressing normally.

## Impact

An unrelated background request failure blanks the trade column and tears down the Swap instance on a page that is otherwise fully usable. It recovers only when the next 6s tick succeeds. This is a narrower window than the variant-switch case, but it is caused by a different root cause — the scope of the error gate, not the loading-atom timing.

## Recommendation

Scope the error gate to the variant-derived terms only, leaving the token-detail term independent, at `DesktopLayout.tsx:301-305`:

```ts
const isTradeReadinessPending =
  (!isTokenVariantsError && (isTokenVariantPending || isTokenVariantsLoading)) ||
  (Boolean(selectedTokenVariant) && isTokenDetailLoading);
```

This preserves the terminal fix — on a **first**-fetch failure there is no cached variant, so `selectedTokenVariant` is `undefined`, the second clause is false, and the panel still correctly falls through to `disabled` — while not punishing the cached-variant case. Combine with the `hasCurrentTokenVariants` gate suggested on line 304.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,23 @@ export function DesktopLayout({
...
+  // mounting Swap.
   const shouldDisableTrade =
     disableTrade ||
-    (shouldUseStockDesktopLayout
-      ? !hasRenderedStockTradeRef.current && !isSwapTokenReady
-      : !isSwapTokenReady);
+    (!shouldUseStockDesktopLayout && !isSwapTokenReady) ||
+    (shouldUseStockDesktopLayout &&
+      !isSwapTokenReady &&
+      !isTradeReadinessPending);
```
sidmorizon: 🟡 P2: Variant switches still remount Swap

Severity: `non-severe`

Selecting a stock variant makes `shouldDisableTrade` true before `isTokenDetailLoading` starts. [MarketEmbeddedSwapDraft](packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx:258-264) then unmounts Swap, discarding its input and quote state.

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
```diff
@@ -243,9 +250,20 @@ function MarketEmbeddedSwapDraft({
...
+
   // Disabled desktop routes render their own unavailable state (or no trade
-  // panel). Do not present a perpetual loading skeleton for a terminal state.
-  return disabled ? null : (
+  // panel). Transient stock identity loading keeps the panel shell visible;
+  // once the shell has mounted, the stock channel owns partial skeletons.
+  if (disabled) {
+    return null;
+  }
+  if (isTradeLoading && !hasRenderedContentRef.current) {
```
originalix, line 292: P2: [Unresolved stock metadata is seeded into the live Swap ticket]

When a user switches stock variants after this ticket has mounted, `DesktopLayout` keeps the panel enabled while `isTradeLoading` is true and supplies a token whose metadata is still unresolved (the new desktop test models that state with `decimals: 0` and `decimalsResolved: false`). This guard only blocks the first mount; on a later transition it forwards that provisional token into `MarketEmbeddedSwapContent`, which immediately updates `swapTokenSeed` and `swapInitParams`.

That causes shared Swap to consume the new variant with zero decimals. Its initialization identity does not include decimals, so the later resolved metadata may not replace the consumed execution token; fractional stock amounts can then be treated as integer-only during the transition.

Keep the last resolved seed/init pair while `isTradeLoading` is true, and update the seed only after the new variant's identity and decimals are resolved. Add a test that the shared Swap component never receives the provisional zero-decimal token.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,23 @@ export function DesktopLayout({
...
+    (isTokenVariantPending ||
+      isTokenVariantsLoading ||
+      (Boolean(selectedTokenVariant) && isTokenDetailLoading));
+  const isTradeLoading =
+    shouldUseStockDesktopLayout && !isSwapTokenReady && isTradeReadinessPending;
+  // Stock's embedded Swap owns transient token/config loading and renders its
+  // own skeletons. Only terminal route/ownership state should remove the
+  // trade panel; generic token layouts still wait for a resolved token before
+  // mounting Swap.
   const shouldDisableTrade =
```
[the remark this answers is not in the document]
zhaono1: Verified on the live head and fixed in 62219376d6. The comment is valid: prepareStockTokenDetail can clear the prior token identity after the render that sees the new selected variant, so isTokenDetailLoading is still false for one frame while isSwapTokenReady is already false. DesktopLayout now treats only settled no-variant or variant-error states as terminal; a selected variant with an unmatched token identity remains enabled and uses the existing in-place loading path. Added a regression covering the variant-switch gap; focused test and commit checks pass.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,28 @@ export function DesktopLayout({
...
     disableTrade ||
-    (shouldUseStockDesktopLayout
-      ? !hasRenderedStockTradeRef.current && !isSwapTokenReady
-      : !isSwapTokenReady);
+    (!shouldUseStockDesktopLayout && !isSwapTokenReady) ||
+    isTerminalStockTradeUnavailable;
```
sidmorizon, line 332: # 🟠 P1 A failed variants poll now tears down a fully-ready, in-use trade panel

## Problem

This commit dropped the `!isSwapTokenReady` conjunct that previously guarded the stock disable clause. Compare:

```ts
// 7e539198 — disable required the panel to be NOT ready
(shouldUseStockDesktopLayout && !isSwapTokenReady && !isTradeReadinessPending)

// 62219376 (this commit, line 322) — no readiness conjunct at all
isTerminalStockTradeUnavailable
```

and `isTerminalStockTradeUnavailable` (`:301-306`) is true whenever `isTokenVariantsError` is true, regardless of readiness:

```ts
const isTerminalStockTradeUnavailable =
  shouldUseStockDesktopLayout &&
  (isTokenVariantsError || (!isTokenVariantPending && !isTokenVariantsLoading && !selectedTokenVariant));
```

`isTokenVariantsError` is **not** a terminal signal. The variants request polls every 6s (`hooks/StockDetailContext.tsx:83`, `:196-205`), and its `catch` returns the **cached** items with `failed: true` (`:184-193`), which sets `isTokenVariantsError` at `:298-302`. So a single transient failure of a background poll now flips `shouldDisableTrade` true even when the variant is cached and tradable, the token detail is fully resolved, and `isSwapTokenReady` is `true`.

## Impact

A user with the stock trade panel fully loaded — mid-order, amount typed, quote live — loses the entire panel to a blank fixed-width column because an unrelated background refresh failed. `MarketEmbeddedSwap.tsx:258-260` returns `null`, unmounting `MarketEmbeddedSwapContent` and destroying the Swap instance, its quote and provider/approval state. (`inputDraftRef` lives on `MarketEmbeddedSwapDraft`, which stays mounted, so the typed amount may be restored — but the panel still blanks and Swap remounts.) It recovers only when the next 6s tick succeeds, so flaky connectivity produces repeated teardowns.

This is a regression introduced by this commit: on `7e539198` the same failure was harmless once the panel was ready, because disabling also required `!isSwapTokenReady`. It also escalates the narrower issue I raised on the previous head ([r4000004054](https://github.com/OneKeyHQ/app-monorepo/pull/13372#discussion_r4000004054), now outdated), which was limited to the token-detail-loading window.

## Recommendation

Treat "error" as terminal only when there is genuinely nothing to trade with, and never disable a panel that is already ready. At `DesktopLayout.tsx:301-306` and `:319-322`:

```ts
const isTerminalStockTradeUnavailable =
  shouldUseStockDesktopLayout &&
  !isSwapTokenReady &&
  !selectedTokenVariant &&
  (isTokenVariantsError ||
    (!isTokenVariantPending && !isTokenVariantsLoading));
```

Requiring `!selectedTokenVariant` keeps the cached-variant case alive, and re-adding `!isSwapTokenReady` guarantees a ready panel is never torn down by a background refresh. The genuine terminal cases still resolve correctly: a first-fetch failure leaves `selectedTokenVariant` undefined (the selection effect returns early on `failed`, `:224`), and the no-tradable-variant case is unchanged.

Worth adding a regression that starts from a ready state, sets `isTokenVariantsError: true` with a `selectedTokenVariant` present, and asserts `disableTrade` stays `false`.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,28 @@ export function DesktopLayout({
...
-  if (shouldUseStockDesktopLayout && isSwapTokenReady) {
-    hasRenderedStockTradeRef.current = true;
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTerminalStockTradeUnavailable =
+    shouldUseStockDesktopLayout &&
+    (isTokenVariantsError ||
```
sidmorizon: 🟡 P2: Cached stock trade disappears on errors

Severity: `non-severe`

When variant polling fails after loading, `isTerminalStockTradeUnavailable` disables trade despite retaining `selectedTokenVariant`. The usable cached trade panel disappears until polling succeeds.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,28 @@ export function DesktopLayout({
...
+      (!isTokenVariantPending &&
+        !isTokenVariantsLoading &&
+        !selectedTokenVariant));
+  const isTradeReadinessPending =
+    !isTerminalStockTradeUnavailable &&
+    (isTokenVariantPending ||
+      isTokenVariantsLoading ||
+      (Boolean(selectedTokenVariant) &&
+        (!isSwapTokenReady || isTokenDetailLoading)));
```
sidmorizon: 🟡 P2: Failed token details load forever

Severity: `non-severe`

When token-detail loading fails, `isTradeReadinessPending` stays true because `isSwapTokenReady` remains false. Initial stock trades show an endless skeleton.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: Verified on the live head and fixed in 6f90f99cfb. This polling flicker finding is valid: isTokenVariantsLoading is true during background refresh ticks even after the variant list has settled, so it must not reopen the initial trade skeleton when no tradable variant exists. The readiness predicate now relies on the cold-start/selection signals and keeps settled no-variant terminal state disabled across refresh ticks. Added coverage for the transient path and terminal behavior; focused tests and commit checks pass.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: Verified on the live head and fixed in 6f90f99cfb. This cached-variant error finding is valid: a failed refresh can retain a selected tradable variant, so the variant error must not disable the panel or suppress an unrelated token-detail loading state. The terminal error path now applies only when no selected variant exists; cached selections remain mounted while their token detail resolves. Added the mounted Swap metadata regression; focused tests and commit checks pass.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: This duplicate finding was valid on the earlier head and is fixed in 62219376d6/6f90f99cfb. The variant-switch path now keeps the existing Swap mounted through the pre-loading render; the newer terminal/loading fixes are included in 6f90f99cfb.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,27 @@ export function DesktopLayout({
...
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTerminalStockTradeUnavailable =
+    shouldUseStockDesktopLayout &&
+    ((!selectedTokenVariant && isTokenVariantsError) ||
+      (!isTokenVariantPending &&
+        !isTokenVariantsLoading &&
```
sidmorizon: # 🟠 P1 Non-tradable stock now mounts a live Swap with an empty token on every 6s poll tick

## Problem

Escalation of [r4000003560](https://github.com/OneKeyHQ/app-monorepo/pull/13372#discussion_r4000003560), which is still open — this commit did not fix it, it changed its failure mode for the worse.

`isTokenVariantsLoading` remains inside the terminal test on line 305, and the variants request polls every 6s (`hooks/StockDetailContext.tsx:83`, `:196-205`; `usePromiseResult.ts:282/371/374-378` re-enters `setLoadingTrue()` each tick). Take a stock where the variant list has settled with nothing tradable — `selectedTokenVariant` undefined, `isTokenVariantPending` false, `isTokenVariantsError` false:

- **Between ticks** (`isTokenVariantsLoading` false) → second clause is `true && true && true` → `isTerminalStockTradeUnavailable` **true** → `shouldDisableTrade` true → `return null`. Blank. ✅
- **During each tick** (`isTokenVariantsLoading` true) → second clause false, first clause false (no error) → `isTerminalStockTradeUnavailable` **false**. And because this commit removed the bare `isTokenVariantsLoading ||` from `isTradeReadinessPending` (`:307-310`), pending is now **false** too: `isTokenVariantPending` is false and `Boolean(selectedTokenVariant)` is false. So `isTradeLoading` is **false** and `shouldDisableTrade` is **false**.

In `MarketEmbeddedSwap.tsx:258-264` that combination falls through both guards: not `disabled`, not `isTradeLoading` → `hasRenderedContentRef.current = true` and `MarketEmbeddedSwapContent` **mounts**.

The token it mounts with is empty. With `selectedTokenVariant` undefined, `stockTokenAddress` falls back to `routeTokenAddress` (`:186-187`), which is `''` on the stock-list entry path (`MarketDetailV2.tsx:127-128`); `stockDisplayMatchesVariant` is then falsy, so `swapToken` is `{ contractAddress: '', symbol: '', decimals: 0, isStock: true }` (`:223-225`, `:227-250`).

## Impact

A stock with no tradable variant now renders a **fully interactive Swap panel** — not a skeleton — for the duration of every poll request, initialized with an empty contract address and zero decimals, then unmounts it again between ticks. Two problems, either of which is worse than the flicker I originally reported: trade UI is offered for a stock that is explicitly not tradable, and the Swap instance is seeded with an invalid token. Previously this state showed a shimmer; on base `x` it was a stable blank column. Note the seed-freezing added in this commit (`MarketEmbeddedSwap.tsx:129-168`) does not help here, because it is gated on `isTradeLoading`, which is false in exactly this state.

## Recommendation

The root cause is unchanged from my original comment: an in-flight **refresh** must not make a settled state non-terminal. Only the absence of a settled result should. `hasCurrentTokenVariants` is already computed at `hooks/StockDetailContext.tsx:212-213` — expose it and drop `isTokenVariantsLoading` from the terminal test at `:301-306`:

```ts
const isTerminalStockTradeUnavailable =
  shouldUseStockDesktopLayout &&
  !selectedTokenVariant &&
  (isTokenVariantsError ||
    (hasCurrentTokenVariants && !isTokenVariantPending));
```

`isTokenVariantPending` already covers cold start (it is true while `!hasCurrentTokenVariants`), so nothing regresses there, and the state stays terminal across refresh ticks instead of oscillating.

Please add a regression that pins this directly: no tradable variant, then toggle `isTokenVariantsLoading` true, and assert `disableTrade` stays `true` and `isTradeLoading` stays `false`. Every terminal-state test added so far (`DesktopLayout.test.tsx:328-356`, `:358-387`) pins `isTokenVariantsLoading: false`, which is why three rounds have not caught this.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +298,26 @@ export function DesktopLayout({
...
-  }
-  // Keep a mounted trade panel alive while a sibling chain variant is
-  // resolving its metadata. The shared Swap channel then replaces only the
-  // token-dependent controls with skeletons; the initial cold start still
-  // waits for a complete execution token.
+  const isTerminalStockTradeUnavailable =
+    shouldUseStockDesktopLayout &&
+    !selectedTokenVariant &&
+    (isTokenVariantsError ||
+      (hasCurrentTokenVariants && !isTokenVariantPending));
```
sidmorizon: # 🟡 P2 The `hasCurrentTokenVariants` conjunct is redundant, and exposing it adds a misleading context field

First: the behavioural fix in this commit is correct and my findings on it are resolved. This is a cleanup point about *how* it was expressed — and it is partly my fault, since my suggested snippet last round included this conjunct. Having traced it properly, it is dead logic.

## Problem

**The conjunct cannot change the result.** `isTokenVariantPending` is computed in the same `useMemo`, from the same render, as (`hooks/StockDetailContext.tsx:291-297`):

```ts
isTokenVariantPending: Boolean(
  normalizedStockId &&
  (!hasCurrentTokenVariants ||
    (!tokenVariantResult?.failed && tokenVariants.some(isStockTokenVariantTradable) && !selectedTokenVariant)),
)
```

Line 305 is only reachable behind `shouldUseStockDesktopLayout = isStockRoute && Boolean(stockId)` (`:178`), and `isStockRoute` is `Boolean(normalizedStockId)` with `stockId` being `normalizedStockId` (`StockDetailContext.tsx:278-279`) — so `normalizedStockId` is necessarily truthy here. Then `!isTokenVariantPending` implies `hasCurrentTokenVariants && !(…)`, i.e. **`!isTokenVariantPending` already entails `hasCurrentTokenVariants`**. So `hasCurrentTokenVariants && !isTokenVariantPending ≡ !isTokenVariantPending` at this call site.

The actual fix in this commit is the *deletion* of `!isTokenVariantsLoading`. The newly exported field contributes nothing.

**The exported field is a footgun.** `hasCurrentTokenVariants = tokenVariantResult?.stockId === normalizedStockId` (`StockDetailContext.tsx:214-215`), and `normalizedStockId = stockId?.trim().toUpperCase() || undefined` (`:99`) — so off a stock route both sides are `undefined` and the expression is `undefined === undefined` → **`true`**. The name reads as "this stock has variants", but it means "the last variant response belongs to the current stock", and it is vacuously `true` whenever there is no stock and before the first fetch resolves. It is safe *today* only because every read sits behind `shouldUseStockDesktopLayout`.

## Impact

No user-visible defect — I verified all four reachable states resolve correctly and the poll-tick oscillation is genuinely gone. The cost is maintainability: a dead term inside the predicate that decides whether trading is offered, plus a new public member on `IStockDetailContextValue` that no other consumer reads (grep: only `StockDetailContext.tsx`, `DesktopLayout.tsx`, `DesktopLayout.test.tsx`) and that returns `true` in the one situation its name suggests it should return `false`.

## Recommendation

Either drop the exposure and simplify at `DesktopLayout.tsx:301-305`:

```ts
const isTerminalStockTradeUnavailable =
  shouldUseStockDesktopLayout &&
  !selectedTokenVariant &&
  (isTokenVariantsError || !isTokenVariantPending);
```

reverting the additions at `StockDetailContext.tsx:37`, `:57`, `:290` and keeping `hasCurrentTokenVariants` provider-local — or, if the explicitness is preferred, keep it but make it honest:

```ts
const hasCurrentTokenVariants = Boolean(
  normalizedStockId && tokenVariantResult?.stockId === normalizedStockId,
);
```

and rename to something like `isTokenVariantResultForCurrentStock`.

Worth noting for coverage: no test sets `hasCurrentTokenVariants: false`, so the conjunct is never exercised either way. The new `'keeps a settled no-tradable stock disabled during variant polling'` case does earn its place — it is inert against the current code but would fail if `!isTokenVariantsLoading` were ever reintroduced, which is exactly the regression guard this needed.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
zhaono1: Verified on the live head and fixed in 6f90f99cfb, with the request-pending refinement in d57cd789be. This finding is valid: a mounted stock Swap must retain its last resolved init pair and stock execution token while the next variant metadata is unresolved. MarketEmbeddedSwap now freezes both seeds during isTradeLoading and only adopts the new token after the request settles. Added a regression asserting neither swapInitParams nor stockTradeToken receives the provisional zero-decimal token.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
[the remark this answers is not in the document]
zhaono1: Verified on the live head and fixed in d57cd789be/5d85a5d1e5. This finding is valid for settled token-detail failures: selectedTokenVariant alone must not keep isTradeLoading true forever. The panel now stays enabled after a settled invalid detail; loading is reserved for an in-flight detail request or an actual variant-identity transition. Added a regression for selected metadata settling invalid.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: This is a duplicate of the cached-variant refresh-error finding and is fixed in d57cd789be. Terminal disable now applies only when no selected variant remains; a cached selected variant stays mounted during a failed background variants refresh.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: This duplicate finding is valid and fixed in d57cd789be/5d85a5d1e5. Settled token-detail failure no longer counts as readiness pending, so the panel does not hold an endless skeleton; it remains enabled for the stock channel to surface its own recovery state. The request-pending and invalid-settled regressions pass.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: This is the polling-flash escalation already fixed in d57cd789be. A settled no-tradable result remains terminal even while the background variants request is in flight; the bare polling loading flag is not used to reopen the panel. The direct loading-tick regression passes.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
zhaono1: Agreed cleanup, fixed in 5d85a5d1e5. The redundant hasCurrentTokenVariants context field and conjunct were removed; the behavior remains covered by the polling regression, while the provider keeps the result-local state private.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +299,27 @@ export function DesktopLayout({
...
+    shouldUseStockDesktopLayout &&
+    !selectedTokenVariant &&
+    (isTokenVariantsError || !isTokenVariantPending);
+  const isTradeReadinessPending =
+    !isTerminalStockTradeUnavailable &&
+    (isTokenVariantPending ||
+      (Boolean(selectedTokenVariant) &&
+        (isTokenDetailRequestPending ||
+          isTokenDetailLoading ||
+          !stockDisplayMatchesVariant)));
```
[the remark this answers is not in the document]
zhaono1: Verified on the live head and fixed in 84721233d9. This finding is valid: display-identity mismatch is a terminal/preview condition, not proof that a token-detail request is still running. Trade loading now uses only variant pending, token-detail request generation pending, and the actual token-detail loading flag; a settled missing or mismatched display detail no longer holds the skeleton forever. Added the settled-mismatch regression.

originalix:

Review of commit `84721233d9c8` finished, but some comments could not be published automatically.

packages/kit/src/views/Market/MarketDetailV2/layouts/DesktopLayout.tsx
```diff
@@ -290,24 +299,25 @@ export function DesktopLayout({
...
-  // waits for a complete execution token.
+  const isTerminalStockTradeUnavailable =
+    shouldUseStockDesktopLayout &&
+    !selectedTokenVariant &&
+    (isTokenVariantsError || !isTokenVariantPending);
+  const isTradeReadinessPending =
+    !isTerminalStockTradeUnavailable &&
+    (isTokenVariantPending ||
+      (Boolean(selectedTokenVariant) &&
+        (isTokenDetailRequestPending || isTokenDetailLoading)));
```
sidmorizon, line 322: # 🟡 P2 Releasing the seed on request-settlement locks `decimals: 0` into the Swap stock channel

The predicate itself is now correct — every term here is a genuine in-flight signal, and I have verified the previously reported defects are all closed. This is a consequence of the new release condition interacting with a downstream identity check.

**First, the reassuring part**, since "zero decimals on a trade token" reads alarming: I traced it to the wire and there is **no financial defect**. Amounts stay human-decimal strings end to end (`ServiceSwap.ts:1203`, `fetchBuildTx` at `:1453-1462` posts `fromTokenAmount` verbatim), and every `transferInfo`/`approveInfo` spreads the server's `data.fromTokenInfo` rather than the UI token (`useSwapBuiltTx.ts:2124-2128`, `:2447`, `:2492-2493`, `:2929-2941`). `truncateDecimalPlaces` early-returns on `!decimal` (`Swap/utils/utils.ts:17-20`), so it is a no-op at `0`. No wrong on-chain amount can be built from this state.

## Problem

`isTradeLoading` now clears when the token-detail **request settles**, which is not the same as when **decimals resolve**. In the settled-invalid state the panel is enabled with `swapToken.decimals === 0` (`DesktopLayout.tsx:232-234`), and the seed freeze in `MarketEmbeddedSwap.tsx:138-169` is gated on `isTradeLoading` — now `false` — so the unresolved token is adopted as the seed.

It then reaches `useSwapStockChannel` as `controlledStockToken`, and that is where it sticks:

```ts
// hooks/useSwapStockChannel.ts:378-383
if (!nextStockToken || getTokenIdentityKey(nextStockToken) === currentStockTokenKey) {
  return;
}
```

`getTokenIdentityKey` is `networkId:contractAddress:native|token` — **`decimals` is deliberately excluded** (`hooks/swapStockChannelUtils.ts:153-160`). `MarketEmbeddedSwap.tsx:151-156` *does* include `decimals` in its seed signature, so when Market later resolves `0 → 18` for the same variant a new prop arrives — but the identity key is unchanged, the effect early-returns, and `selectStockSwapToken` is never called. The execution token keeps `decimals: 0`.

Before this increment, `isTradeReadinessPending` still contained `!isSwapTokenReady`, which held `isTradeLoading` true until decimals resolved and so froze the seed past this window. That is why the gap is newly reachable here rather than pre-existing.

## Impact

Not financial. The user-visible effect is on the sell side: `useSwapStockTradeInputs.ts:698` calls `validateAmountInput(value, 0)`, and at `decimal = 0` the regex (`utils/validateAmountInput.ts:1-11`) admits `"0."` but rejects `"0.5"` — so the field accepts the decimal point and then silently drops every subsequent keystroke, with no error shown. The input is not behind a skeleton here: `shouldRenderStockTradeInputSkeleton` requires `inputTokenStatus === Initializing`, but `stockTokenStatus` is `Ready` as soon as `currentStockToken` exists (`useSwapStockChannel.ts:635-643`). A `decimals: 0` token is also persisted into `swapStockSelectedToken` and `swapRecentTokenPairs` (`ServiceSwap.ts:3089-3099`, `:3169-3179`), where it can seed a later session.

Two recovery paths exist but neither is unconditional: `useSwapStockChannel.ts:420-437` repairs `decimals` only when the channel's own detail returns a non-empty `stock.subtitle`, and `useSwapStockTradeInputs.ts:646-657, 733-738` is sell-side only and requires `authoritativeBalance !== undefined`. **Marking permanence UNPROVEN** — I could not establish that `stock.subtitle` is always populated, so I cannot say how often this persists versus self-heals on the first `fetchSwapTokenDetails` response.

## Recommendation

Fix it at the identity check rather than by reintroducing the skeleton. At `useSwapStockChannel.ts:378-383`, when the identity key matches but `decimals` differs, fall through to a metadata-only sync (e.g. `syncStockTokenDetail(nextStockToken)`) instead of returning — so the execution token picks up resolved decimals without `resetReceiveAmount: true` wiping the user's input.

Widening `getTokenIdentityKey` to include `decimals` would also work but is riskier: it is used for identity comparisons elsewhere, and a decimals change would then read as a token change.

Reverting to `!isSwapTokenReady` here would reintroduce the endless skeleton this commit correctly fixed, so please don't.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

zhaono1: Valid finding. I reproduced the same-identity controlled token path: a settled Market detail could release the embedded Stock channel with decimals=0, and the channel then ignored the later resolved decimals because the identity key intentionally excludes metadata. Fixed in 0c2224cd7c: when the controlled token identity matches but its decimals are newly resolved, the channel now performs a metadata-only sync without resetting the user's receive amount; an unresolved decimals=0 value cannot downgrade an already resolved channel token. Added regression coverage for both directions. Focused tests (96 tests across 4 suites) and agent:check --profile commit pass.

packages/kit/src/views/Swap/hooks/useSwapStockChannel.ts
```diff
@@ -413,6 +391,42 @@ export function useSwapStockChannel(
...
+    }
+    if (getTokenIdentityKey(nextStockToken) === currentStockTokenKey) {
+      if (
+        controlledStockToken &&
+        shouldSyncControlledStockTokenMetadata({
+          controlledStockToken,
+          currentStockToken,
+        })
+      ) {
+        syncStockTokenDetail(controlledStockToken);
```
sidmorizon, line 418: # 🟡 P2 Metadata-only sync falsely stamps the "current locale" ref with a token that carries no localized metadata

The decimals fix itself is correct — I verified the metadata-only sync updates `decimals` without resetting the receive amount, the `decimals > 0` guard blocks the reverse downgrade, convergence terminates, and effect ordering is unchanged. This is a narrow side effect of reusing `syncStockTokenDetail` for a caller it wasn't written for.

## Problem

`syncStockTokenDetail` stamps the locale ref unconditionally on every successful sync (`useSwapStockChannel.ts:383`):

```ts
stockTokenMetadataLocaleRef.current = localeRef.current;
```

That ref's contract is stated at `:111-113` — *"Token-selector metadata is already localized. Keep its locale separate from the persisted token so a cold-start snapshot cannot flash old copy."* Every pre-existing caller honoured it: the subtitle effect (`:444-450`) passes `stock: detail.stock` from a freshly fetched detail, and `selectStockSwapToken` (`:358`) is fed from the token selector.

The new call site is the first to pass a token with **no `stock` field at all**. `controlledStockToken` originates in the Market `swapToken` memo (`DesktopLayout.tsx:236-275`), which builds only `networkId / contractAddress / symbol / decimals / logoURI / price / isNative / isStock` — no `stock`. It reaches the channel via `StockDesktopLayout.tsx:1410` → `MarketEmbeddedSwap.tsx:239`.

So inside `resolveStockExecutionTokenMetadata`, `mergeStockExecutionMetadata` short-circuits (`swapStockChannelUtils.ts:294-296`):

```ts
if (!currentStock || !tokenDetailStock) {
  return tokenDetailStock ?? currentStock;
}
```

`tokenDetailStock` is `undefined`, so `stock` is returned **unchanged** — potentially a persisted snapshot written under a previous app locale — while `:383` now asserts it is current-locale.

## Impact

Cosmetic/i18n only; no financial or data impact, and desktop/web only (`controlledStockToken` is undefined on native, so the branch never fires there).

After an in-app language change, or on a cold start where `swapStockSelectedTokenAtom` / the cold-start snapshot was written under a different locale, this decimals sync can fire before `stockTokenDetail` lands. In that window `hasCurrentLocaleStockMetadata` (`:724-725`) flips true, so `selectedStock` is taken from the stale token (`SwapStockDesktopContainer.tsx:1594`), `getStockMarketTokenSubtitle` returns the stale subtitle, and `hasTokenData` suppresses `showTokenLabelsSkeleton`. The header renders previous-language copy instead of the skeleton — exactly the flash the ref exists to prevent. It self-heals once the subtitle effect at `:430` runs.

**The user-visible flash is UNPROVEN**: it requires the decimals resolution to land before the swap-side `stockTokenDetail` fetch *and* a locale-mismatched persisted snapshot. The false stamp itself is proven from the lines above.

## Recommendation

Only stamp the ref when the incoming detail actually contributes localized metadata. At `useSwapStockChannel.ts:383`:

```ts
if (tokenDetail.stock) {
  stockTokenMetadataLocaleRef.current = localeRef.current;
}
```

That is a no-op for the existing subtitle caller, which only runs when `detail.stock.subtitle` is non-empty, and it leaves the decimals sync doing purely what it says on the tin.

[Claude session ↗](https://claude.ai/code/session_014UxunDDHH6FM2pxVjHbMoT)

---

zhaono1: Valid. The controlled Market token intentionally carries execution fields without localized stock metadata, so the new decimals-only sync must not mark persisted labels as current-locale. Fixed in 0053999d40: syncStockTokenDetail updates the locale ref only when tokenDetail.stock is present; the decimals-only path still updates execution metadata without resetting amounts. This preserves the existing cold-start/language-change guard. Focused regression suites remain green (81 tests across 3 suites). Resolving.

originalix (dismissed):

packages/kit/src/views/Market/MarketDetailV2/layouts/MarketEmbeddedSwap.tsx
```diff
@@ -200,7 +236,7 @@ function MarketEmbeddedSwapContent({
         stockSpeedConfig={stockSpeedConfig}
         stockTradeConfig={stockTradeConfig}
         stockTradeHeader={resolvedStockTradeHeader}
-        stockTradeToken={stockTradeToken}
+        stockTradeToken={stockTradeTokenSeed}
```
originalix, line 239: P2: [Keep stock conversion config aligned with the frozen token]

When a Stock variant changes while its token detail is still pending, this component retains `swapTokenSeed` and `stockTradeTokenSeed` for the previous variant. `stockTradeConfig`, however, is still forwarded from `StockDesktopLayout`'s current `selectedTokenVariant`.

`StockEstimatedReceive` gives that config's `tokenToAssetRatio` precedence while the quote and channel token are still the old seed, so the displayed estimated shares can be calculated with the new variant's ratio against the old variant's quote. Snapshot or derive the conversion config from the same seeded token and replace both only after the new identity is ready.

originalix (approved):

PR #13374, opened 2026-09-13
fix: release ios market header before scroll view recycling

limichange:
Entering Discover can crash with `CALayerInvalid` when Fabric reuses the loading ScrollView for a filter inside the shared sticky header within one mounting transaction. Release the old list's header, KVO observer, and insets before recycling can create that native view cycle.

Surviving lists retain their saved scroll positions, including adjacent tabs temporarily aligned to an expanded header. Native component identity and Fabric tags distinguish survivors from replacements, including same-address/tag reuse. Selection-completion callbacks restore list insets and observation synchronously. Missing lists stop retrying on unrelated commits while retaining the state needed for a later insertion or window return.

This changes iOS main UI ownership only. Android behavior, including its disabled tab indicator animation, is unchanged.

Validation:
- Captured the original layer-cycle crash and Fabric mutation order under LLDB.
- 51 UIKit regression assertions pass with extracted production native methods. These include actual tap/swipe completion, cancellation and queued navigation methods; the fixture controls animation/completion timing. They also cover surviving/replaced lists, delayed deletion, missing current/adjacent lists, window return and cycle prevention.
- All 11 existing recovery/filtering assertions pass. Missing-list recovery does no recursive searches or forced offset writes for 300 unrelated property updates; 300 unrelated insertions inspect a 1,001-node page once.
- The corresponding 49-case suite fails 11 assertions on the previous commit. A further window-return regression was reproduced and fixed before the final build.
- Clean patch-package application to the pristine package reproduces all edited native sources; Android patch sections are byte-identical.
- Rebuilt iOS and Android native shells and vendor bundles through Mobile DevSession. Both display the banner and populated lists, collapse/expand the header, and switch tabs successfully.
- iOS Trending retains the same visible rows/position after a Stocks round trip and after leaving/re-entering Discover. A separate cold launch directly into Favorites renders the banner and real list without crashing.
- Android's underline remains on Trending during a held horizontal swipe, then selects Stocks when the swipe settles; re-entering Discover also works.
- Commit validation and all local PR-profile checks pass. All 25 remote CI checks pass. No new iOS layer-cycle/uncaught-native-exception or Android native crash was recorded in the final UI regression windows.

Follow-up to merged #13368.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-297c1260-1ceb-53c9-89be-ab19b96b9f77)

sidmorizon:
Claude review session: https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -260,3 +288,78 @@ index 84eab17..ca70732 100644
...
+   }
+ }
+ 
++// OneKey patch: release observers, headers and saved state while the scroll view
++// still represents the old list, before Fabric applies its next owner's props.
++- (void)restoreScrollViewState:(UIScrollView *)scrollView
++{
++  if (scrollView == _observedScrollView) {
++    [self detachScrollObserver];
++    _pageOffsets[self.currentPageKey] = @(_headerOffset);
```
sidmorizon: # 🟠 P1 Pre-mount release discards the saved scroll offset

## Problem

`restoreScrollViewState:` unconditionally replaces the current page's saved offset with the clamped header offset:

```objc
if (scrollView == _observedScrollView) {
  [self detachScrollObserver];                          // saves the real logical offset
  _pageOffsets[self.currentPageKey] = @(_headerOffset); // then throws it away
}
```

Inside `restoreDetachedScrollInsets` that was safe: the loop only reaches a scroll view after `belongsToPage == NO`, so the list is already gone and its offset is worthless. The new `mountingTransactionWillMount:` (line 226) applies the same release to any managed scroll view that merely has a removed tag **anywhere on its ancestor chain** — and Fabric emits `ShadowViewMutation::Remove` for subtrees that *survive*. In RN 0.86.2 `Differentiator.cpp`: line 365 ("there's some reordering requiring a REMOVE mutation"), line 1060 ("only items outside the LIS are moved") and line 1182 ("the node was reparented elsewhere"). So a sibling reorder or a reparent of any wrapper around the observed list now releases a live, on-screen, deeply-scrolled list.

`mountingTransactionDidMount:` does recover for that case and calls `attachScrollObserverForCurrentPage`. But `detachScrollObserver` has already set `_observedScrollView = nil` and `_observingContentOffset = NO`, so the pre-existing fast path `if (candidate == _observedScrollView) { …; return; }` no longer applies and `applyInsetsToScrollView:…restore:YES` runs instead. With `restore == YES` that path always force-sets `contentOffset` to `restoredLogicalOffset - next.top`, reading exactly the value this line clobbered. `_headerOffset` is untouched by the release (KVO was removed first), so for a fully collapsed header the `_headerOffset < _headerHeight` override does not rescue it either.

## Impact

iOS only, `main` runtime, single native component instance. A user scrolled deep into a Market tab list sees the list snap back to the header-collapse position whenever an unrelated commit reorders or reparents any ancestor of that list — a silent, user-visible loss of scroll position that does not happen before this change, where the surviving-scroll-view fast path returned without touching `contentOffset` or `_pageOffsets`. It also degrades the round-trip offset restore for that page on the next tab switch.

## Recommendation

Only discard the saved offset when the scroll view is genuinely being given up. Parameterise the helper so the pre-mount path preserves the state the post-mount reattach needs:

```objc
- (void)restoreScrollViewState:(UIScrollView *)scrollView resetSavedOffset:(BOOL)resetSavedOffset
{
  if (scrollView == _observedScrollView) {
    [self detachScrollObserver];          // already records the live logical offset
    if (resetSavedOffset) _pageOffsets[self.currentPageKey] = @(_headerOffset);
  }
  ...
}
```

Pass `YES` from `restoreDetachedScrollInsets` (line 362), preserving today's behaviour exactly, and `NO` from `mountingTransactionWillMount:`. When the list really was replaced, the existing `replacingCurrentScrollView` branch in `attachScrollObserverForCurrentPage` still collapses the offset, so the `restore:YES` reattach keeps the user's position for survivors while still resetting it for genuine replacements.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.129.patch
```diff
@@ -204,6 +204,34 @@ index 84eab17..ca70732 100644
...
++         mutation.type == ShadowViewMutation::Delete) && mutation.oldChildShadowView.tag > 0) {
++      removedTags.insert(mutation.oldChildShadowView.tag);
++    }
++  }
++  if (removedTags.empty()) return;
++
++  // Inspect only the managed scroll views' ancestor chains, including wrappers
++  // above the pager. Row removals and unrelated commits keep ownership intact.
++  for (UIScrollView *scrollView in _originalInsets.keyEnumerator.allObjects) {
++    for (UIView *ancestor = scrollView; ancestor != nil; ancestor = ancestor.superview) {
```
sidmorizon, line 114: # 🟠 P1 Release scope exceeds did-mount recovery scope

## Problem

The two mounting callbacks are asymmetric in both their guards and their detection scope, so `willMount` can surrender ownership in situations `didMount` structurally cannot recover from in the same transaction.

- **Guards.** `willMount` (line 213) guards only `_isBeingRecycled || _originalInsets.count == 0`. `didMount` additionally bails on `self.window == nil || _transitioning || _isPagerDragging`. A commit landing mid-swipe therefore detaches the KVO observer and returns the shared header to the container with no paired reattach.
- **Detection scope.** This ancestor walk is unbounded — deliberately "including wrappers above the pager". `didMount` only looks for a changed `parentTag` within the subtree rooted at `_pageControllers[_currentIndex].view`. A `Remove` for a wrapper *above* the pager matches here and is invisible to `didMount`.
- **The escape hatch is dead.** `if (!headerIsAttached) { [self attachScrollObserverForCurrentPage]; return; }` looks like it covers this, but it can never fire after a `willMount` release: `restoreSharedHeadersToContainer` re-parents `_sharedHeaderHostView` into `_containerView`, and `_containerView` is assigned as `self.contentView`, i.e. a subview of `self`. So `[_sharedHeaderHostView isDescendantOfView:self]` stays `YES`.

The blast radius is wider than it appears: every managed scroll view shares the ancestor chain above the pager, so a single `Remove` up there releases the current page's list *and* both adjacent pages' lists in one pass.

## Impact

iOS only, `main` runtime. After such a transaction the pager has no `contentOffset` observer and the shared header is parked in the container: the sticky header stops collapsing and stops tracking the list. It is not permanent — the async tails of `layoutSubviews` / `updateProps` and the transition-completion blocks re-run `attachScrollObserverForCurrentPage` — but the header visibly desyncs for at least a frame, and every one of those recovery paths goes through the lossy `restore:YES` reattach described in the finding on line 301.

## Recommendation

Make the release hand off explicitly to the recovery, and stop walking past the component:

1. Bound the walk at `self`. An ancestor above the pager being recycled implies the pager is recycled too, and `prepareForRecycle` already performs a full release for every entry in `_originalInsets`:
   ```objc
   for (UIView *ancestor = scrollView; ancestor != nil && ancestor != self; ancestor = ancestor.superview) {
   ```
2. Record that ownership was surrendered and honour it ahead of `didMount`'s heuristics:
   ```objc
   // in willMount, where restoreScrollViewState: is called
   if (scrollView == _observedScrollView) _needsObserverReattachAfterMount = YES;

   // at the top of mountingTransactionDidMount:, after the _isBeingRecycled check
   if (_needsObserverReattachAfterMount) {
     _needsObserverReattachAfterMount = NO;
     if (self.window != nil && !_transitioning && !_isPagerDragging) {
       [self attachScrollObserverForCurrentPage];
       return;
     }
   }
   ```
   Keeping the flag set while `_transitioning` / `_isPagerDragging` hold lets the existing transition-completion `attachScrollObserverForCurrentPage` finish the handoff rather than leaving it to a layout pass.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -204,13 +263,72 @@ index 84eab17..ca70732 100644
...
++        UIView *componentView = RNCScrollComponentView(scrollView);
++        RNCCollapsiblePagerReleasedScrollState *state = [RNCCollapsiblePagerReleasedScrollState new];
++        state.scrollView = scrollView;
++        state.componentTag = componentView.tag;
++        // A deleted component can reuse both the native object and its tag.
++        if (deletedTags.find(componentView.tag) == deletedTags.end()) {
++          state.componentView = componentView;
++        }
++        _releasedPageScrollStates[pageKey] = state;
++        _pageOffsets[pageKey] = @(scrollView.contentOffset.y + scrollView.contentInset.top);
```
sidmorizon: # 🟠 P1 willMount overwrites a retained page's saved offset

## Problem

`mountingTransactionWillMount:` records the live offset for **every** page-belonging release, not just the observed one:

```objc
_releasedPageScrollStates[pageKey] = state;
_pageOffsets[pageKey] = @(scrollView.contentOffset.y + scrollView.contentInset.top);
```

For the page at `_currentIndex` this is correct — KVO keeps `_pageOffsets[currentPageKey]` in sync with the live offset, so reading it back is a no-op.

For a **retained neighbour page** it is not. `applyInsetsToScrollView:pageIndex:restore:YES` deliberately forces a non-current page's scroll view to the shared-header collapse position while the header is still visible:

```objc
if (_nativeSmoothHeaderScrollEnabled && _headerOffset < _headerHeight) {
  restoredLogicalOffset = _headerOffset;
}
```

and it does **not** write that clamped value back into `_pageOffsets`. So a retained page routinely sits at `contentOffset.y + contentInset.top == _headerOffset` while `_pageOffsets[pageKey]` legitimately still holds the user's deep offset. `_nativeSmoothHeaderScrollEnabled` is live on exactly this screen — `packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx:767` passes `nativeSmoothHeaderScrollEnabled={platformEnv.isNative}`.

Because the ancestor walk is deliberately unbounded above the pager, a single `Remove` of a wrapper above the pager matches for *all* managed scroll views, so every retained neighbour has its saved deep offset replaced by the clamped one in a single pass. Line 313 then passes `resetSavedOffset:NO`, so nothing corrects it afterwards, and the `sameList` check only *preserves* the value this line already destroyed. The previous head never wrote `_pageOffsets` for a scroll view that was not `_observedScrollView`, so this is a regression introduced by this push.

## Impact

A user deep-scrolls tab B, switches to tab A, and scrolls tab A until the shared header is fully collapsed. Any Fabric `Remove`/`Delete` touching tab B's ancestor chain while the header was still expanded silently replaces `_pageOffsets[B]` with the collapse position. When the user swipes back to B with the header collapsed — the one path where the deep offset *is* honoured — B restores to the top of its content instead of where they left it. Same user-visible symptom as the offset loss fixed in this push, relocated from the current page to retained pages, and silent and unrecoverable once the value is gone.

## Recommendation

Only refresh `_pageOffsets[pageKey]` from the live scroll view when that scroll view is the observed one, or when there is no saved value yet:

```objc
_releasedPageScrollStates[pageKey] = state;
if (scrollView == _observedScrollView || _pageOffsets[pageKey] == nil) {
  _pageOffsets[pageKey] = @(scrollView.contentOffset.y + scrollView.contentInset.top);
}
```

For a retained page the existing `_pageOffsets` entry is already authoritative — the pager's own `restore:YES` clamp is what made the live offset unrepresentative, so it must not be read back as truth.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -245,7 +363,25 @@ index 84eab17..ca70732 100644
...
+   [self detachScrollObserver];
+   [self restoreSharedHeadersToContainer];
+@@ -1766,16 +1916,29 @@ static void RNCLogNativeTabScrollBoundary(NSString *owner,
+ 
+ - (void)attachScrollObserverForCurrentPage
+ {
++  // OneKey patch: transition/window callbacks also complete deferred recovery.
++  if (_isBeingRecycled) return;
++  if (_needsScrollObserverReattachAfterMount) {
++    if (self.window == nil || _transitioning || _isPagerDragging) return;
```
sidmorizon: # 🟡 P2 Pending flag suppresses attach for existing callers

## Problem

This guard makes a *pending* flag suppress the whole method for every pre-existing caller:

```objc
if (_needsScrollObserverReattachAfterMount) {
  if (self.window == nil || _transitioning || _isPagerDragging) return;
  [self prepareAdjacentPageInsets];
}
[self restoreDetachedScrollInsets];
```

Several established callers invoke `attachScrollObserverForCurrentPage` *deliberately* while `_transitioning` or `_isPagerDragging` is still `YES`, because that is the moment the new page becomes current:

- `goTo:`'s transition completion sets `_currentIndex = index` and attaches while `_transitioning` is still `YES` — it is cleared only later, in `completeTransitionOnNextRunLoopForGeneration:`.
- `finishPagerScrollEmittingSelection:` sets `_currentIndex` and attaches, and only afterwards sets `_isPagerDragging = NO`, in both the direct and `UIPageViewController` branches.

With the flag set, all of those attaches become no-ops: no `restoreDetachedScrollInsets`, no `applyInsetsToScrollView:`, no `attachSharedHeadersToScrollView:`. The flag has therefore changed the behaviour of callers it was not meant to affect, on exactly the path a release-during-transition produces.

## Impact

For one run loop after a tab tap or a completed swipe that coincided with a release, the shared header stays parked in `_containerView` instead of riding the new page's list, and that list still carries its pre-release original insets, so its first row renders under the header. Recovery only arrives via `completeTransitionOnNextRunLoopForGeneration:` on the next run loop, producing a one-frame header/content jump on precisely the screen this PR is meant to stabilise.

## Recommendation

Let the flag gate only the extra work it was added for, never the pre-existing attach path:

```objc
if (_needsScrollObserverReattachAfterMount && self.window != nil &&
    !_transitioning && !_isPagerDragging) {
  [self prepareAdjacentPageInsets];
}
[self restoreDetachedScrollInsets];
```

Dropping the `return` keeps the deferred-recovery intent — the flag still stays set until a real attach clears it — while restoring the synchronous attach that `goTo:` and `finishPagerScrollEmittingSelection:` depend on.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.129.patch
```diff
@@ -260,3 +396,111 @@ index 84eab17..ca70732 100644
    // Legacy mode keeps the historical pager-first dependency. Smooth mode
    // arbitrates header touches by axis so the vertical list can begin directly.
    if (_pagerScrollView != nil && !_nativeSmoothHeaderScrollEnabled && !_needsPropsReapply) {
+     [candidate.panGestureRecognizer requireGestureRecognizerToFail:_pagerScrollView.panGestureRecognizer];
+   }
+   if (candidate == _observedScrollView) {
++    _needsScrollObserverReattachAfterMount = NO;
```
sidmorizon, line 274: # 🟡 P2 Reattach flag never clears without a scroll view

*(Anchored on the first clear-site; the defect is the `candidate == nil` early return just above, at line 391, which GitHub will not accept a comment on because it falls between diff hunks.)*

## Problem

When the current page genuinely has no vertical scroll view, `attachScrollObserverForCurrentPage` returns through that branch without clearing the flag:

```objc
if (candidate == nil) {
  [self detachScrollObserver];
  [self restoreSharedHeadersToContainer];
  return;
}
```

Both clear-sites for `_needsScrollObserverReattachAfterMount` — this line and line 410 — are *below* that return, and `_releasedPageScrollStates[currentPageKey]` is never consumed either, because `applyInsetsToScrollView:` is the only consumer and it is never reached. The flag therefore stays set for the lifetime of the host, and `mountingTransactionDidMount:` short-circuits to `attachScrollObserverForCurrentPage` on **every** subsequent mounting transaction on the surface, bypassing the cheap `changedParentTags` scan added precisely to avoid that. Each such call runs `prepareAdjacentPageInsets` — `restoreDetachedScrollInsets` plus up to three full recursive `findVerticalScrollViewInView:` subtree walks plus three `applyInsetsToScrollView:restore:YES` calls, each issuing an unconditional `setContentOffset:` on a neighbouring page's list. It also keeps `_releasedPageScrollStates.count != 0`, defeating the cheap early-out at line 273 for every future transaction.

## Impact

On a page that renders an empty or error state without a scroll container, a price-ticking market surface turns every Fabric commit into three page-subtree walks and three forced `setContentOffset:` writes on neighbouring lists, which can also interrupt an in-flight bounce or refresh settlement on those pages. The work is unbounded in time and grows with commit rate.

## Recommendation

Treat "the mount finished and the current page has no scroll container" as a completed recovery, matching the rule the pager already applies at line 405 (*a list replaced by an empty RN view keeps collapse, not the old row offset*):

```objc
if (candidate == nil) {
  [self detachScrollObserver];
  [self restoreSharedHeadersToContainer];
  [_releasedPageScrollStates removeObjectForKey:self.currentPageKey];
  _pageOffsets[self.currentPageKey] = @(_headerOffset);
  _needsScrollObserverReattachAfterMount = NO;
  return;
}
```

This loses no recovery: when a real list later mounts, its `Insert` carries a `parentTag` inside the current page subtree, so the existing changed-parent scan in `mountingTransactionDidMount:` re-attaches without needing the flag.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.123.patch
```diff
@@ -204,17 +266,81 @@ index 84eab17..ca70732 100644
...
++    [self attachScrollObserverForCurrentPage];
++    return;
++  }
 +  UIView *page = _pageControllers[_currentIndex].view;
 +  BOOL headerIsAttached = [_sharedHeaderHostView isDescendantOfView:self];
-+  if (_observedScrollView != nil && [_observedScrollView isDescendantOfView:page] &&
-+      headerIsAttached) return;
++  BOOL currentListIsAttached = _observedScrollView != nil &&
++    [_observedScrollView isDescendantOfView:page] && headerIsAttached;
++  if (currentListIsAttached && _releasedPageScrollStates.count == 0) return;
```
sidmorizon: # 🟡 P2 Released page states are never reaped

## Problem

`_releasedPageScrollStates` is now load-bearing for two gates — this early-out and the `didMoveToWindow` recovery trigger — but nothing ever reconciles the dictionary with the pager's live page set. Only two places remove entries: `applyInsetsToScrollView:pageIndex:restore:`, which removes exactly one key and only for an index it is called with, and `prepareForRecycle`, which clears everything.

Every caller that can reach `applyInsetsToScrollView:` with an arbitrary index is confined to the `_currentIndex ± 1` window (`prepareAdjacentPageInsets` clamps to `MAX(0, _currentIndex - 1) … MIN(count - 1, _currentIndex + 1)`). So an entry recorded in `mountingTransactionWillMount:` for a page that later sits two or more slots from `_currentIndex` has no reaper at all. `rebuildSlots` and `unmountChildComponentView:index:` do not clear it either, and `updateProps:` replaces `_pageKeys` wholesale without pruning — since `pageKeyForIndex:` reads `_pageKeys[index]`, a tab-set change strands a key that can never be produced again, so `hasReleasedList` is `NO` for every index while `count` stays non-zero and the entry becomes permanently unrecoverable. Because `scrollView` and `componentView` are both `weak`, a state whose views have been deallocated is likewise an indistinguishable tombstone keeping `count > 0` forever.

While any such entry lingers, this early-out cannot fire: `mountingTransactionDidMount:` walks every mutation in every mounting transaction to build `changedParentTags`, then loops all pages and runs a full BFS over the subtree of each page holding a released entry. On a Market or Discover screen with several tabs and websocket-driven commits, that is per-commit main-thread work that no longer terminates, and an insertion landing inside a far-away page's subtree also flips `_needsScrollObserverReattachAfterMount` and drives a full current-page recovery that cannot repair the released page, since recovery only ever targets `_currentIndex`.

This is the same "recovery re-runs on every subsequent transaction" symptom the flag-clearing change was meant to remove; clearing the flag at `candidate == nil` relocated the persistent state rather than eliminating it.

## Impact

iOS only, `main` runtime. Not a crash, hang, or loop — the forced `setContentOffset:` writes converge, because `applyInsetsToScrollView:restore:` recomputes a deterministic target from `_pageOffsets`, so a repeat write is a no-op and emits no scroll callback. The cost is bounded per transaction but unbounded in duration: sustained per-commit mutation scanning plus subtree BFS on the main thread for the lifetime of the screen, and repeated no-op inset/offset rewrites on off-screen neighbour lists. The stale-`_pageKeys` variant is permanent, with no chance of self-recovery short of `prepareForRecycle`.

## Recommendation

Give the dictionary a reaper so `count` reflects only states a recovery can still act on:

- In `rebuildSlots`, and wherever `_pageKeys` is replaced in `updateProps:`, drop every key absent from the new `_pageKeys`.
- In the existing `_releasedPageScrollStates.allValues` sweep in `mountingTransactionWillMount:`, remove any entry whose weak `scrollView` is `nil` *and* whose `componentView` is `nil` — `sameList` can only ever be `NO` for it, so it carries no remaining information.
- Optionally, gate the two `_releasedPageScrollStates.count > 0` conditions on a helper counting only entries for indices inside `_currentIndex ± 1`, so a state the recovery window cannot reach does not defeat this early-out. Keep the entry itself for the lazy consumption that happens when the user later navigates to that page.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.129.patch
```diff
@@ -0,0 +1,400 @@
...
++}
++
++// OneKey patch: selection completion restores list state synchronously, before
++// the transition flags are cleared on the next run loop. Other callers wait.
++- (void)attachScrollObserverForCurrentPageAfterTransition:(BOOL)selectionSettled
++{
++  if (_isBeingRecycled) return;
++  if (_needsScrollObserverReattachAfterMount) {
++    if (self.window == nil || ((_transitioning || _isPagerDragging) && !selectionSettled)) return;
++    [self prepareAdjacentPageInsets];
```
sidmorizon, line 235: # 🟠 P1 Released state adjudicated against transient fallback

## Problem

`attachScrollObserverForCurrentPageAfterTransition:` calls `prepareAdjacentPageInsets` here, at the top of the recovery path — *before* the candidate is resolved and before upstream's fallback debounce runs.

`prepareAdjacentPageInsets` iterates `MAX(0, _currentIndex - 1) … MIN(count - 1, _currentIndex + 1)`, so it includes the **current** page, and it resolves each index with `verticalScrollViewForPageAtIndex:`, whose last resort is `findVerticalScrollViewInView:` — which returns a `UICollectionView`. Its `scrollView != _observedScrollView` skip cannot protect the current page, because `detachScrollObserver` has already set `_observedScrollView = nil` during the release.

It then calls `applyInsetsToScrollView:pageIndex:restore:YES`, which consumes the released state: it computes `sameList`, overwrites `_pageOffsets[pageKey]` with `_headerOffset` when the identity does not match, and removes the key.

Immediately afterwards, the *same* invocation resolves the candidate and hits upstream 3.0.129's new fallback debounce: when the candidate is a `UICollectionView` differing from `_observedScrollView`, it records `_pendingFallbackScrollView`, calls `scheduleScrollObserverRetryForCurrentPage` and returns **without attaching**, because upstream does not trust that candidate yet.

That debounce exists for exactly the situation this code creates. Upstream's own comment on the retry reads: *"Fabric can mount a recycled child's native view before attaching its replacement ScrollView component. Let that transaction settle before accepting an automatically discovered UICollectionView fallback."* So during release recovery a transient `UICollectionView` is the expected intermediate state — and the released state is adjudicated and destroyed against precisely that untrusted view, one debounce window before the real list resolves.

This ordering was harmless on 3.0.123, where the attach path had no debounce and no `_pendingFallbackScrollView`. Both are new in upstream 3.0.129 (they are present in the pristine package, not added by this patch), so the migration introduced the dependency.

## Impact

A list that survives the recycle is misjudged as a replacement. `sameList` is `NO` because `released.scrollView` is the real list while `scrollView` is the transient collection view, so `_pageOffsets[pageKey]` is reset to `_headerOffset` and the key is removed. When the real list resolves — after the 0.5 s debounce window, reached via 100 ms retries — there is no released state left to consult, so the user's scroll position in that tab is silently discarded and the page snaps to the header-collapse offset. That is the same class of visible jump this PR exists to prevent. Adjacent pages are affected too, since `prepareAdjacentPageInsets` performs the same consumption for `_currentIndex ± 1`. In the opposite direction, if the transient candidate happens to be the previously released object, `sameList` can be `YES` and a stale deep offset is restored onto what is in fact a fresh list.

## Recommendation

Do not let a not-yet-trusted candidate consume the released state. Either:

- move the `prepareAdjacentPageInsets` call below the `_pendingFallbackScrollView` debounce block, so recovery re-applies insets only once upstream has accepted the candidate; or
- gate the consumption inside `applyInsetsToScrollView:pageIndex:restore:` on the candidate not being a pending fallback — skip the `sameList` adjudication and the `removeObjectForKey:` when `scrollView == _pendingFallbackScrollView`, leaving the entry for the retry that resolves the real list.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.129.patch
```diff
@@ -0,0 +1,400 @@
...
++        NSString *pageKey = [self pageKeyForIndex:index];
++        UIView *componentView = RNCScrollComponentView(scrollView);
++        RNCCollapsiblePagerReleasedScrollState *state = [RNCCollapsiblePagerReleasedScrollState new];
++        state.scrollView = scrollView;
++        state.componentTag = componentView.tag;
++        // A deleted component can reuse both the native object and its tag.
++        if (deletedTags.find(componentView.tag) == deletedTags.end()) {
++          state.componentView = componentView;
++        }
++        _releasedPageScrollStates[pageKey] = state;
```
sidmorizon, line 129: # 🟡 P2 Released page states are never reaped

*(Re-anchored onto the migrated `3.0.129` patch. This supersedes the earlier thread on the now-deleted `...3.0.123.patch`, which GitHub marks outdated; the defect survives the migration unchanged.)*

## Problem

`_releasedPageScrollStates` has no reaping path. Only two sites ever remove entries: `applyInsetsToScrollView:pageIndex:restore:` removes a single key, and only for an index it is actually called with; and `prepareForRecycle` does `removeAllObjects`. Consumption is clamped to `_currentIndex ± 1` by `prepareAdjacentPageInsets`, so a page two or more slots away never gets its entry consumed. `rebuildSlots` and `unmountChildComponentView:index:` do not clear it, and `updateProps:` replaces `_pageKeys` wholesale. Since `pageKeyForIndex:` reads `_pageKeys[index]`, a tab-set change strands the old key permanently.

## Impact

The dictionary is load-bearing twice over. `mountingTransactionDidMount:` early-outs on `if (currentListIsAttached && !currentListIsFallback && _releasedPageScrollStates.count == 0) return;`, and `didMoveToWindow` re-arms recovery on `_releasedPageScrollStates.count > 0`. Once a key is stranded, `count` stays above zero while `hasReleasedList` is `NO` for every live index, so the early-out can never fire again: every mounting transaction on the surface pays a full mutation scan plus a breadth-first walk of the current page's subview tree, and every `didMoveToWindow` re-arms `_needsScrollObserverReattachAfterMount` and forces a `prepareAdjacentPageInsets` pass. Because `scrollView` and `componentView` are `weak`, a stranded entry is also indistinguishable from a live one — both present as `nil` members — so nothing downstream can tell it is dead.

This is not a crash or a loop: the forced `setContentOffset:` writes recompute a deterministic target from `_pageOffsets`, so a repeat write is a no-op. The cost is bounded per transaction but unbounded in duration, for the lifetime of the screen.

## Recommendation

Reap the dictionary whenever the key space or the slot set changes:

- In `updateProps:`, after assigning the new `_pageKeys`, drop every entry whose key is absent from the new array.
- In `rebuildSlots` and `unmountChildComponentView:index:`, drop entries for indices that no longer have a page controller.
- As a backstop, drop any entry whose weak `scrollView` has become `nil` — a released scroll view that has been deallocated can never be re-adjudicated, so the entry carries no remaining information.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

patches/react-native-pager-view+3.0.129.patch
limichange: Confirmed on the current 3.0.129 patch with an extracted UIKit regression. A released list at logical offset 900 is detached, a transient UICollectionView fallback appears, and prepareAdjacentPageInsets consumes the released identity and resets the saved offset to the header position 200 before the debounce accepts any candidate. When the surviving real list returns, it restores to 200. The same failure reproduces for a retained adjacent page. This should be fixed before merge. Moving prepareAdjacentPageInsets below only the current-page debounce, or checking only _pendingFallbackScrollView, would not protect adjacent released pages; the fix needs to prevent any untrusted generic UICollectionView fallback from adjudicating released identity, while still allowing an accepted current fallback to consume it after the debounce.

sidmorizon: # 🟠 P1 Released state adjudicated against transient fallback — corrected recommendation

## Problem

Your objection is right, and my original recommendation was wrong on both counts. I re-checked: every reference to `_pendingFallbackScrollView` — the ivar plus the accept/reset sites — lives inside `attachScrollObserverForCurrentPageAfterTransition:` and operates only on the **current** page's resolved candidate. `prepareAdjacentPageInsets` resolves each of `_currentIndex ± 1` with `verticalScrollViewForPageAtIndex:` and calls `applyInsetsToScrollView:pageIndex:restore:YES` directly, so an adjacent page's candidate never reaches the debounce at all. Reordering the `prepareAdjacentPageInsets` call therefore only gates the current page, and a `scrollView == _pendingFallbackScrollView` test can only ever match the current page's candidate — neither protects a retained adjacent page, exactly as you describe.

The property that actually separates the safe case from the unsafe one is the **provenance** of the resolution, not which page it belongs to. `verticalScrollViewForPageAtIndex:` tries `findExplicitNativeScrollerInView:`, then `findFabricScrollViewInView:` (which returns an `RCTScrollViewComponentView`'s own `scrollView`), and only then falls through to `findVerticalScrollViewInView:`, whose last resort matches any `UICollectionView`. The first two yield trustworthy identities; the third is the untrusted guess your regression reproduces against.

## Impact

Unchanged from the finding above, and your reproduction pins it down more precisely than my analysis did: released identity is consumed and `_pageOffsets` reset to the header position before any candidate has been accepted, for the current page *and* for retained adjacent pages — which my recommendation would have left broken even after a fix.

## Recommendation

Gate adjudication on provenance, so no untrusted generic fallback can consume released identity on any page, while an accepted current fallback still can:

1. Have the resolver report provenance, e.g. `- (UIScrollView *)verticalScrollViewForPageAtIndex:(NSInteger)pageIndex isGenericFallback:(BOOL *)isGenericFallback`, setting `NO` for the explicit-native-scroller and `RCTScrollViewComponentView` paths and `YES` only for the `findVerticalScrollViewInView:` fall-through.
2. Add a parameter to the consumer — `applyInsetsToScrollView:pageIndex:restore:adjudicateReleasedState:` — and wrap the `sameList` / `_pageOffsets[pageKey] = @(_headerOffset)` / `removeObjectForKey:` block in it. Insets and offsets still apply when it is `NO`; only the identity verdict is deferred.
3. `prepareAdjacentPageInsets` passes `!isGenericFallback`, so an untrusted adjacent candidate defers its verdict and the entry survives for a later pass.
4. The current-page call site passes `YES` once the debounce has accepted the candidate — at the point where `_pendingFallbackScrollView` is cleared after the 0.5 s window — and `NO` while it is still pending.

One caveat worth designing for: if an adjacent page's list is only ever reachable through the generic fallback, step 3 defers its verdict indefinitely and the entry is never consumed, which feeds straight into the unreaped-dictionary finding on line 125. Pairing this with a per-page pending timestamp — accepting a stable adjacent fallback after the same window rather than never — keeps both properties, and the reaper in that other finding is worth adding regardless.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

limichange: Fixed in 860f91f5071a3baaee179d19552e984bc98374d3. Released state is no longer adjudicated by an unaccepted UICollectionView candidate; it remains available until the candidate has passed the upstream debounce and become the observed scroll view. This also protects retained adjacent pages, where _pendingFallbackScrollView alone would not be sufficient. Validation: 56 UIKit recovery/transition assertions plus 11 mutation-filtering assertions pass, including transient current-page recovery, transient adjacent-page recovery, and eventual acceptance of a genuine replacement fallback. The regenerated patch applies cleanly to pristine 3.0.129 and reproduces the working source byte-for-byte.

patches/react-native-pager-view+3.0.129.patch
limichange: Confirmed as a state-lifecycle cleanup gap, but it is non-blocking for this fix. Market uses pageRetentionDistance=1, so a released page two or more slots away normally has its content unmounted and only an empty slot wrapper remains. Price ticks are property Update mutations; mountingTransactionDidMount finds no Insert/Remove parent tags and returns before any subtree BFS or offset rewrite. The residual work is bounded mutation/key scanning, with no crash, loop, or scroll-position correctness impact. We are keeping this as an optional profiling-driven follow-up rather than expanding the native recovery state machine in this crash fix.

patches/react-native-pager-view+3.0.129.patch
```diff
@@ -0,0 +1,404 @@
...
+                         restore:(BOOL)restore
+ {
++  // OneKey patch: a surviving list keeps its full offset. A replacement keeps
++  // only header collapse, even if Fabric reused the same native scroll object.
++  NSString *pageKey = [self pageKeyForIndex:pageIndex];
++  RNCCollapsiblePagerReleasedScrollState *released = _releasedPageScrollStates[pageKey];
++  // OneKey patch: upstream debounces generic UICollectionView fallbacks before
++  // accepting them. Keep the released identity until that candidate is observed.
++  BOOL isUnacceptedCollectionView =
++    [scrollView isKindOfClass:UICollectionView.class] && scrollView != _observedScrollView;
```
sidmorizon: # 🟠 P1 Adjacent page lists are never adjudicated

First, confirming the fix works for what it targeted: the transient-fallback destruction is gone. `_observedScrollView = candidate` is assigned immediately **before** `applyInsetsToScrollView:` in the attach path, so an accepted current-page candidate still adjudicates, and the adjudication block sits above the `restore:` block that reads `_pageOffsets`, so a reset takes effect in the same call. The 900 → 200 loss you reproduced cannot recur on the current page.

The second conjunct, however, introduces the inverse defect on neighbour pages.

## Problem

`_observedScrollView` is a current-page-only concept: it has exactly two assignments in the whole file — `= candidate` in the attach path and `= nil` in `detachScrollObserver` — so it can never hold an *adjacent* page's list.

But `applyInsetsToScrollView:pageIndex:restore:` is also called for adjacent pages, from `prepareAdjacentPageInsets` (which calls it only when `scrollView != _observedScrollView`) and from `adjacentController:delta:` (where the target index differs from the current one by construction). On both paths `scrollView != _observedScrollView` holds unconditionally, so whenever an adjacent page's list is a `UICollectionView` the guard is permanently `YES` and the adjudication block never runs for that page.

That is not a hypothetical class here — it is OneKey's own list. `@onekeyfe/react-native-native-list@3.0.129` (pinned at `apps/mobile/package.json:94`) defines `final class NativeListView: UIView` hosting a private `UICollectionView` with a vertical flow layout (`ios/RNCNativeListView.swift:6,63`). Because it is a Nitro `UIView` rather than an `RCTScrollViewComponentView`, neither `findExplicitNativeScrollerInView:` nor `findFabricScrollViewInView:` can see it — it is reachable *only* through the generic `findVerticalScrollViewInView:` `UICollectionView` branch. And it is mounted inside this very pager: `packages/kit/src/views/Market/MarketHomeV2/layouts/MobileLayout.native.tsx` imports `CollapsiblePagerView` (line 16) and `MobileMarketNativeLists` (line 47), with `nativeSmoothHeaderScrollEnabled={platformEnv.isNative}` (line 767).

So for every neighbour tab on Market, the released-state verdict is skipped unconditionally. When that page's list is genuinely **replaced**, `_releasedPageScrollStates[pageKey]` is never consumed, `_pageOffsets[pageKey]` still holds the *old* list's deep offset, and the `restore:` path forces the **new** list to it. The `if (_nativeSmoothHeaderScrollEnabled && _headerOffset < _headerHeight)` clamp hides this while the header is partly expanded, so it surfaces once the header is fully collapsed — the normal steady state after any scroll, and the same state your own repro used.

## Impact

On iOS with the header fully collapsed, swiping toward a neighbour tab whose list was replaced shows that list at the previous list's row position instead of the top, then snapping to the top once the page becomes current and adjudication finally runs — up to the 0.5 s debounce later. It is the mirror image of the bug this commit fixes, moved from the current page to its neighbours.

It also removes the only reaper that covered adjacent pages: `prepareAdjacentPageInsets` no longer consumes anything for a `UICollectionView`, so a page released while adjacent and never made current again keeps its entry permanently. That matters for the deferral reasoning on the line-125 thread — the stranding now has a scroll-position correctness consequence, not just bounded scanning.

## Recommendation

Stop using `_observedScrollView` as the acceptance signal, since it cannot describe a non-current page. Give the deferral the same *time-bounded* semantics as upstream's own debounce, per page, so every released entry is eventually adjudicated:

- Track first-sight time per page key, e.g. `NSMutableDictionary<NSString *, NSNumber *> *_pendingFallbackDetectedAtByPage`. When `applyInsetsToScrollView:` sees a `UICollectionView` that is not `_observedScrollView` for `pageKey`, record `CACurrentMediaTime()` on first sight and skip adjudication only while `CACurrentMediaTime() - recorded < 0.5`; past that window, adjudicate normally.
- Clear that page's timestamp when its candidate is accepted or when the page is released again, and clear the dictionary in `prepareForRecycle` alongside the existing `removeAllObjects`.

This preserves the current-page behaviour this commit intends, restores adjudication for replacements on neighbour pages, and bounds entry lifetime to one debounce window instead of forever.

[Claude session ↗](https://claude.ai/code/session_01DuuDjwrtfcf2RkYRduBYoK)

---

limichange: Fixed in 705c44f0a1. Released state now tracks the pending fallback candidate and detection time per page. A surviving native list is accepted immediately; an unmatched adjacent UICollectionView stays at the safe header offset and is adjudicated after a 0.5s per-candidate debounce. Candidate changes restart the window, and stale callbacks are guarded by page state, component generation, window attachment, and recycle state. The pre-fix UIKit harness reproduced both the stale 900 offset and permanently retained state; after the fix, all 34 recovery cases, 63 transition/recovery cases, and 11 mutation-filtering cases pass. The regenerated 3.0.129 patch also applies cleanly with patch-package and matches the tested source byte-for-byte.

zhaono1 (approved):

originalix (approved):

PR #13376, opened 2026-09-13
fix: preserve extension route hash when switching tokens(OK-58148)

huhuanming:
[OK-58148](<https://onekeyhq.atlassian.net/browse/OK-58148>)

----------

## Summary
- Preserve the newly generated extension hash route when React Navigation also appends the previous hash.
- Prevent Market token switches in the extension expand tab from reverting to the originally opened token after refresh.

## Intent & Context
OK-58148 reports that switching tokens on the Market detail page updates the content but not the extension URL. Refreshing the page therefore reopens the token that was selected when entering the detail page. The same flow was checked in a Playwright-controlled Chromium instance using a production extension build.

## Root Cause
React Navigation preserves `location.hash` when navigation updates parameters on the same route key. In the extension, the hash represents the complete app route, so this produces a value containing the new route followed by the old route hash. The existing extension history adapter selected the last hash segment and discarded the newly generated token route.

## Design Decisions
The fix stays in the existing extension-only `createMemoryHistory` patch because this is the final state-to-browser-history normalization layer. The dedicated Market hash navigation hook handles the opposite direction (hash-to-navigation) and adding outbound synchronization there would duplicate linking behavior and risk hashchange/navigation loops. Non-extension platforms remain outside this branch.

## Changes Detail
- Update `@react-navigation/native`'s extension hash normalization to select the generated route segment instead of the stale trailing hash.
- Apply the same logic to both source and compiled module patch entries.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Extension
- **Risk Areas**: Extension history push/replace behavior for same-route parameter updates

## Test plan
- [x] Build the production extension bundle.
- [x] Load the unpacked extension in Playwright Chromium.
- [x] Open a Market token detail route, switch to another token, and verify the URL hash updates.
- [x] Reload and verify the selected token and updated URL remain stable.
- [x] Run 57 focused Market navigation tests.
- [x] Verify the patch applies cleanly to the pristine Yarn cache package and contains no build artifacts.

[OK-58148]: https://onekeyhq.atlassian.net/browse/OK-58148?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-3e9a9886-6e47-542e-87c5-d7489dad60b0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Yc4mGHq8tx6XtM6izEKRWN

patches/@react-navigation+native+7.1.28.patch
```diff
@@ -21,17 +21,17 @@ index 474db5e..4b47a93 100644
 +      const currentIndexHtml = globalThis.location.href.split('.html')?.[0];
 +      const currentIndexHtmlWithoutSlash = currentIndexHtml.split('/')?.pop();
 +      if (currentIndexHtmlWithoutSlash) {
-+        const currentPath = path.includes('#') ? path.split('#').pop() : path;
++        const currentPath = path.includes('#') ? path.split('#')[1] : path;
```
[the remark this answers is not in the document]
huhuanming: Fixed: regenerated the patch with patch-package 8.0.0 from the modified node_modules package. The postimage blob indexes are now 73d7254 and 88ea4de, and the regenerated patch applies cleanly to the pristine Yarn cache package.

patches/@react-navigation+native+7.1.28.patch
```diff
@@ -1,5 +1,5 @@
 diff --git a/node_modules/@react-navigation/native/lib/module/createMemoryHistory.js b/node_modules/@react-navigation/native/lib/module/createMemoryHistory.js
-index 474db5e..4b47a93 100644
+index 474db5e..73d7254 100644
```
sidmorizon: # 🟡 P2 lib/module postimage index is still wrong

## Problem
`528b48b4` regenerates the patch metadata, but only the `src` entry came out right. Reconstructing the regeneration locally from the pristine `@react-navigation/native@7.1.28` npm tarball:

| entry | preimage (patch) | preimage (computed) | postimage (patch) | postimage (computed) |
|---|---|---|---|---|
| `lib/module/createMemoryHistory.js` (L2) | `474db5e` | `474db5e` ✅ | `73d7254` | **`4f94f0a`** ❌ |
| `src/createMemoryHistory.tsx` (L69) | `1568263` | `1568263` ✅ | `88ea4de` | `88ea4de` ✅ |

Both preimages match the pristine package byte-for-byte, and the patch applies cleanly, so the postimage is fully determined by the hunks. Running the equivalent of what `patch-package` emits — `git diff --no-index` between the pristine and patched trees — produces:

```
index 474db5e..4f94f0a 100644   lib/module/createMemoryHistory.js
index 1568263..88ea4de 100644   src/createMemoryHistory.tsx
```

`73d7254` is not the blob hash of the patched `lib/module/createMemoryHistory.js`. The `src` line on L69 is now correct, which is what makes this look fixed at a glance.

## Impact
No functional breakage: plain `git apply` and `patch-package`'s own applier both ignore `index` lines, which is why CI still passes. The cost is the fallback path — `git apply --3way` resolves the postimage blob by hash, so it cannot fall back for the `lib/module` entry if a future `@react-navigation/native` bump shifts context. It also leaves the patch metadata internally inconsistent with the artifact it describes, so the next regeneration-verification pass will keep re-reporting this, and it makes the commit's own claim of regenerating both entries incorrect.

## Recommendation
Set L2 to `index 474db5e..4f94f0a 100644`. Worth checking why the regeneration produced a hash for `lib/module` that the patched file cannot have — if `patch-package` ran against a `node_modules` whose `lib/module/createMemoryHistory.js` carried edits beyond these hunks, the patch body and the index would have been generated from different content, and the same drift can recur on the next regeneration. Verify with a clean `node_modules` before re-running `/1k-patch-package-workflow`.

[Claude session ↗](https://claude.ai/code/session_01Yc4mGHq8tx6XtM6izEKRWN)

---

huhuanming: Fixed in ea4d9ac75b. The local compiled module had gained a trailing newline that patch-package used for the blob index but did not encode in the hunks. I restored the npm artifact EOF form, regenerated the patch, and verified it in a separately initialized pristine repository: lib/module resolves to 4f94f0a and src resolves to 88ea4de.

originalix (approved):

PR #13378, opened 2026-09-13
fix: avoid notification settings loading flash

huhuanming:
## Summary

- remove the notification settings data-loading spinner
- render cached settings immediately and revalidate them on each page entry
- keep the in-memory cache synchronized with optimistic updates and rollbacks
- prevent stale revalidation responses from overwriting user changes

## Test plan


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-17f6ca0f-44e2-53c6-80fb-d1582b6ef397)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -80,17 +81,23 @@ export default function NotificationsSettings() {
   const intl = useIntl();
   const [settings, setSettings] = useState<
     INotificationPushSettings | undefined
-  >();
+  >(cachedNotificationSettings ?? {});
+  const [isSettingsLoaded, setIsSettingsLoaded] = useState(
+    cachedNotificationSettings !== undefined,
```
sidmorizon: 🟡 P2: Cached toggles overwrite fresh settings

Severity: `non-severe`

On re-entry, `isSettingsLoaded` makes stale cached settings editable before revalidation completes. A quick toggle submits the cached config and discards intervening server changes.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,10 +106,19 @@ export default function NotificationsSettings() {
...
         updated ||
         (await backgroundApiProxy.serviceNotification.fetchServerNotificationSettings());
+      if (
+        !updated &&
+        requestMutationVersion !== settingsMutationVersionRef.current
+      ) {
+        return;
+      }
+      cachedNotificationSettings = result;
       setSettings(result);
```
sidmorizon: 🟡 P2: Reopened pages miss completed updates

Severity: `non-severe`

After a quick reopen, the new `reloadSettings` can finish before the flushed update. The old instance updates only its state and `cachedNotificationSettings`, leaving the visible page stale.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -43,6 +42,8 @@ import {
 } from '../../components/NotificationsTestButton';
 import { SETTINGS_PAGE_BODY_INSET_X } from '../Tab/settingsSurface';
 
+let cachedNotificationSettings: INotificationPushSettings | undefined;
```
sidmorizon: # 🟠 [P1] Module cache is not identity-scoped, unlike the background cache it shadows

## Problem

Line 45 introduces a bare module-level binding with no expiry and no invalidation hook:

```ts
let cachedNotificationSettings: INotificationPushSettings | undefined;
```

The server notification config is **identity-scoped**, and this repo already treats it that way for the background cache. `PrimeGlobalEffect.tsx:309-319` clears it on every Prime / OneKey-ID state change:

```ts
useUpdateEffect(() => { void (async () => {
  noop(isLoggedInOnServer); noop(isActive);
  await backgroundApiProxy.serviceNotification.clearServerSettingsCache();
  await backgroundApiProxy.serviceNotification.registerClientWithOverrideAllAccounts();
})(); }, [isActive, isLoggedInOnServer]);
```

`ServiceNotification.ts:92` clears it again on `AddDBAccountsToWallet`, and the memoized cache carries an explicit `// TODO clear cache if prime expired, onekeyID logout` (`ServiceNotification.ts:1575`).

The new UI cache gets none of this. `clearServerSettingsCache()` is a `@backgroundMethod()` that clears `getServerSettingsWithCache` on the service instance — it cannot reach a `kit` module binding, and on iOS/Android/extension it does not even run in the same JS runtime. Prime logout/login is an atom state change, not a bundle reload, so this binding survives it on every target.

Because the cache also seeds `isSettingsLoaded` (lines 85-87) and `prevSettings.current` (lines 93-95), a post-logout mount renders the previous identity's config as authoritative *and* fully interactive.

This is a distinct root cause from the remount/stale-revalidation findings already on this PR: those are intra-session, same-identity. Neither repair helps here — even with controls locked until revalidation, user B's screen still renders user A's configuration, and the poisoned binding persists for the whole app session.

## Impact

After a OneKey ID logout/login or Prime expiry, Notification Settings shows the previous identity's settings as live. `/notification/v1/config/update` **replaces the entire config** (documented at `notificationPermissionUtils.ts:171-176`) and `updateSettings` submits `{ ...v, ...partSettings }` (lines 183-186), so a toggle inside the revalidation window overwrites the current identity's whole server config — `pushEnabled`, `accountActivityPushEnabled`, `priceAlertsEnabled`, `perpsEnabled`, `announcementEnabled`, `dailyUpdateEnabled`, plus `maxAccount`, which feeds `saveMaxAccountCount` → `backupPrimeAccountActivityNotificationSettings`.

## Recommendation

Give the module cache the lifecycle the background cache already has:

1. Key it by identity — store `{ onekeyUserId, settings }` and treat a mismatch as a cache miss rather than a hit.
2. Reset it from the same place that already clears the background cache: export a `clearCachedNotificationSettings()` from `NotificationsSettings.tsx:45` and call it in the effect at `PrimeGlobalEffect.tsx:309`, or subscribe to the Prime/logout app event.
3. Seed `prevSettings.current` (line 93) from the cache only after that identity check passes, so a rollback cannot restore another identity's snapshot.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,10 +106,19 @@ export default function NotificationsSettings() {
...
         (await backgroundApiProxy.serviceNotification.fetchServerNotificationSettings());
+      if (
+        !updated &&
+        requestMutationVersion !== settingsMutationVersionRef.current
+      ) {
+        return;
+      }
+      cachedNotificationSettings = result;
       setSettings(result);
+      setIsSettingsLoaded(result !== undefined);
```
sidmorizon: # 🟡 [P2] A settings fetch resolving to `undefined` now blanks the page instead of showing a loading state

## Problem

Line 216 flipped the branch from `!settings ? <Spinner/> : <UI/>` to `settings ? <UI/> : null`. That repurposed `undefined` from "still loading" into "render nothing" — but `reloadSettings` still writes `undefined` straight into both state and the module cache:

```ts
const result = updated || (await backgroundApiProxy.serviceNotification.fetchServerNotificationSettings());
...
cachedNotificationSettings = result;          // 119
setSettings(result);                          // 120
setIsSettingsLoaded(result !== undefined);    // 121
```

`undefined` is a real runtime value here, not a theoretical one. `fetchServerNotificationSettings` → `fetchServerNotificationSettingsWithoutToast` returns `result?.data?.data` (`ServiceNotification.ts:1677-1689`), which is `undefined` for an empty or absent server config. The codebase handles that case explicitly in two places — `NotificationEnableAlert.tsx:64` (`// If serverSettings is undefined (API failed), don't show the alert`) and `notificationPermissionUtils.ts:178` (`const hasServerSettings = !!serverSettings && Object.keys(serverSettings).length > 0`) — and line 121 shows it was anticipated here too.

The result is a content→blank transition, which inverts the PR's stated goal: with a warm cache the page renders the full settings list, the revalidation resolves to `undefined`, and the entire body disappears, leaving only the invisible `MultipleClickStack`. The now-`undefined` module cache makes the next mount blank from the first frame as well. Pre-PR the same response produced a spinner — also not correct, but it never removed already-rendered UI and never persisted.

This also breaks the flow `notificationPermissionUtils.ts:182-193` deliberately routes empty-config users into, on the stated grounds that this page "owns the master-switch semantics for an empty config".

## Impact

Users whose config query returns an empty payload get a blank Notification Settings page with no master switch and therefore no way to enable notifications. Warm-cache users see the settings list vanish mid-session.

## Recommendation

Keep `undefined` out of the render path now that it can no longer mean "loading". Normalize at the single assignment point, lines 119-121:

```ts
const next = result ?? {};
cachedNotificationSettings = next;
setSettings(next);
setIsSettingsLoaded(true);
```

If a distinct "load failed" state is wanted, add an explicit error flag rather than reusing `undefined` — the branch at line 216 can no longer express it.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,11 +111,21 @@ export default function NotificationsSettings() {
...
+      if (
+        !updated &&
+        requestMutationVersion !== settingsMutationVersionRef.current
+      ) {
+        return;
+      }
+      const nextSettings = result ?? {};
+      cachedNotificationSettings = nextSettings;
+      setSettings(nextSettings);
+      setIsSettingsLoaded(true);
```
sidmorizon: 🟡 P2: Missing response enables destructive updates

Severity: `non-severe`

When `fetchServerNotificationSettings` returns no data, the page enables controls with an empty config. The [update contract](packages/kit/src/utils/notificationPermissionUtils.ts:171-177) replaces all settings, so one toggle can erase notification defaults.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -187,220 +219,237 @@ export default function NotificationsSettings() {
...
+        </ListItem>
+
+        {settings?.pushEnabled ? (
+          <NotificationPermissionRecoveryAlert
+            scene="settings"
+            pushEnabled={settings.pushEnabled}
+          />
+        ) : null}
+
+        {settings === undefined || settings.pushEnabled ? (
```
sidmorizon, line 445: # 🟡 [P2] The loading state renders two live controls that were previously unreachable during load

## Problem

```tsx
{settings === undefined || settings.pushEnabled ? (
```

This treats "unknown" as "enabled", so the whole push sub-section mounts during load. Only the five switch *positions* get skeletons — the last two children are fully live controls:

```tsx
<ListItem
  testID="notifications-manage-accounts"
  ...
  drillIn
  onPress={() => {
    navigation.push(EModalSettingRoutes.SettingManageAccountActivity);
  }}
/>

{/* Push notifications helper */}
<NotificationsSettingsHelper />
```

`NotificationsSettingsHelper` renders `NotificationHelperCtaButton`, whose `onPress` calls `handlePress` → `recoverOsNotificationPermission()` and `backgroundApiProxy.serviceNotification.showNotification(...)` (`NotificationsTestButton.tsx:73, 83-97, 136-137`).

In the previous revision the initial `settings` was `cachedNotificationSettings ?? {}` — truthy with `pushEnabled` `undefined` — so this block never rendered while loading and neither control existed in that state. The increment newly exposes both.

The skeletons themselves are well done: `w={38} h="$6"` matches `SWITCH_SIZE_CONFIG.small` (`trackWidth: 38, trackHeight: '$6'`) with `flexShrink={0}` matching `TMSwitch`, so the switch rows don't shift. The issue is only the two non-skeleton children.

## Impact

During a cold load a user can tap "Manage accounts" or the Test/Enable CTA, triggering a navigation push, an OS permission prompt, or a test notification in a state the page is about to reveal as "notifications off" — and the modal push races the collapse. Separately, for every user whose real config has `pushEnabled === false`, the page collapses from master row + 5 sub rows + 2 dividers + manage row + helper card down to a single row once loading resolves. That is a larger content jump than the loading flash this PR set out to remove, and it inverts the previous behaviour of expanding only on positive knowledge.

## Recommendation

Keep the block gated on confirmed state (`settings?.pushEnabled`) and render a separate, purely presentational placeholder when `settings === undefined` containing only the skeleton switch rows — no navigable or action-triggering children.

If the expanded-during-load look is deliberate, then at minimum move `notifications-manage-accounts` and `<NotificationsSettingsHelper />` behind `settings !== undefined`, so no live control is reachable while `settings` is unknown.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -187,220 +219,237 @@ export default function NotificationsSettings() {
...
+              maxWidth: '$96',
+            }}
+          />
+          {settings === undefined ? (
+            <NotificationSettingsSwitchSkeleton />
+          ) : (
+            <Switch
+              testID="setting-switch"
+              size="small"
+              disabled={!isSettingsLoaded}
```
sidmorizon: # 🟡 [P2] `isSettingsLoaded` is now inert — this `disabled` and two other guards can never fire

## Problem

After this increment, `isSettingsLoaded === (settings !== undefined)` is an exact invariant. Every write site:

**`settings`** — L87-89 `useState(cachedNotificationSettings)`; L126 `setSettings(nextSettings)` where L124 is `const nextSettings = result ?? {}` (never `undefined`); L185-197 the updater returns `v` or an object; L161 `setSettings(prevSettings.current)` guarded by L159 `if (prevSettings.current)`.

**`isSettingsLoaded`** — L90-92 `useState(cachedNotificationSettings !== undefined)`; L127 `setIsSettingsLoaded(true)`; L162 `setIsSettingsLoaded(true)`.

Both initializers read the same module variable in the same render, so mount state is correlated by construction. After mount `settings` can never return to `undefined`, and **both remaining setters pass a literal `true`**, each batched with a `setSettings(<defined>)`. No reachable state has `settings !== undefined && isSettingsLoaded === false`, or the converse.

So `disabled={!isSettingsLoaded}` sits inside the `settings === undefined ? <Skeleton/> : <Switch/>` ternary at L236 — it is only evaluated on the branch where `settings !== undefined`, hence permanently `disabled={false}`. The same applies to `if (!isSettingsLoaded) { return; }` (L181) and `if (!v) { return v; }` (L186-188).

The previous revision's `setIsSettingsLoaded(result !== undefined)` was the one thing that could leave the flag `false` after a degenerate fetch; normalizing to `true` removed it.

## Impact

Three guards remain in the source that can never fire while reading as active safety. This matters beyond tidiness: "the master switch is disabled until settings are genuinely loaded" is no longer true in any state, so it cannot be relied on as mitigation for the stale-cache and empty-config concerns raised elsewhere on this PR. `isSettingsLoaded` also keeps a redundant state slot and a redundant `updateSettings` dependency (L199).

## Recommendation

Pick one, but don't leave the current form, which is neither:

- **(a)** Delete `isSettingsLoaded` entirely along with L242, L181-183 and L186-188, and derive rendering from `settings === undefined`.
- **(b)** If a real "not yet safe to edit" gate is wanted, replace it with a flag set `true` only after a successful network revalidation completes *on this mount* (e.g. a ref set in `reloadSettings` on the `!updated` path), and apply it to the master switch **and** the five sub-switches, which currently have no `disabled` prop at all.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,10 +106,19 @@ export default function NotificationsSettings() {
 
   const reloadSettings = useCallback(
     async (updated?: INotificationPushSettings) => {
+      const requestMutationVersion = settingsMutationVersionRef.current;
       const result =
         updated ||
         (await backgroundApiProxy.serviceNotification.fetchServerNotificationSettings());
+      if (
+        !updated &&
+        requestMutationVersion !== settingsMutationVersionRef.current
```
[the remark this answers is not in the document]
huhuanming: Acknowledged. For this lightweight SWR, the product decision is to keep the optimistic cache authoritative until a valid server response arrives and to accept the remount race in the current scope. A shared mutation coordinator is intentionally not added here.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Acknowledged. The product decision is to keep cached switches interactive during revalidation. A valid server response still replaces the cache and visible state; the stale-submit window is accepted for this lightweight optimistic flow.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Acknowledged. The product decision is to retain the simple optimistic remount behavior rather than add a module-level update coordinator in this change.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 986e351e2d. The memory cache is now keyed by the authenticated OneKey ID and Prime state. A mismatch is treated as a cache miss, and responses from an identity that is no longer current cannot update the mounted view.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,11 +111,21 @@ export default function NotificationsSettings() {
...
         (await backgroundApiProxy.serviceNotification.fetchServerNotificationSettings());
-      setSettings(result);
-      prevSettings.current = result;
+      if (
+        !updated &&
+        requestMutationVersion !== settingsMutationVersionRef.current
+      ) {
+        return;
+      }
+      const nextSettings = result ?? {};
```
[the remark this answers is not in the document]
huhuanming: Fixed in d78542a284. Undefined and empty responses no longer become an empty loaded config. They leave the last valid cache untouched; on a cold load the switch skeletons remain non-interactive.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Intentional per product decision. Manage Account Activity loads its own settings, while the helper CTA only reads/recovers OS notification permission and sends a local test notification. Neither depends on the server NotificationSettings snapshot, so they may remain live during this loading state.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 3815d458f6. The redundant isSettingsLoaded state, disabled prop, and guard were removed. The presence of a valid settings snapshot now directly determines whether switches or skeletons render.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in d78542a284. Missing or empty responses are rejected before cache/state assignment, so they cannot unlock controls or lead to a partial whole-config update.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,13 +149,38 @@ export default function NotificationsSettings() {
...
+      }
+      if (
+        notificationSettingsIdentityKeyRef.current !==
+        notificationSettingsIdentityKey
+      ) {
+        return;
+      }
+      // Updates replace the complete config, so an empty snapshot must never
+      // unlock controls or replace the last known-good cache.
+      if (!result || Object.keys(result).length === 0) {
```
sidmorizon: # 🟠 [P1] An empty server config now deadlocks this page permanently, and it is the only place that can create one

## Problem

The new guard conflates "not yet loaded" with "loaded and legitimately empty":

```ts
// Updates replace the complete config, so an empty snapshot must never
// unlock controls or replace the last known-good cache.
if (!result || Object.keys(result).length === 0) {
  return;
}
```

`settings` is now the sole loading gate — `settings === undefined` renders a skeleton at lines 311, 366, 397, 428, 458, 488 — so an empty payload leaves the page indistinguishable from "still loading", forever.

The loop is closed; I enumerated every path out:

- `updateSettings` is called **only** from switch `onChange` handlers (L319, 374, 405, 436, 466, 496), each inside the `else` branch of a `settings === undefined ? <Skeleton/> : <Switch/>` ternary. With `settings === undefined` there are **zero reachable callers**.
- `reloadSettings` is called only from the identity effect (L278) and from `doUpdateSettingsToServer` (L207) — the latter reachable only via `updateSettings`. No retry, no timeout, no error surface, no pull-to-refresh. `usePromiseResult` covers `pushClient` only.
- Remount, navigate away and back, and app restart all re-run the same effect against the same empty config and hit the same early return.
- The two controls that *do* render here are not escapes, exactly as you noted when declining the earlier comment: `ManageAccountActivity` writes SimpleDB only, and the helper CTA only reads/recovers OS permission and sends a local test notification. Neither writes the server config.

Decisively: `/notification/v1/config/update` has one writer (`ServiceNotification.ts:1708`), reachable from exactly two call sites — this page's `updateSettings` (L198), and `notificationPermissionUtils.ts:196`, which **explicitly refuses to write when the config is empty** (`hasServerSettings`, L182) and instead routes the user *to this page*, on the stated grounds that it "owns the master-switch semantics for an empty config" (L176-181).

So `enableNotificationsBestEffort` funnels the new-user/empty-config population into a page that, as of d78542a2, can no longer create a config. Before this increment `result ?? {}` at least rendered live switches, so a toggle could bootstrap one. That path is gone and nothing replaced it.

## Impact

Any user whose server config is empty or absent sees six switch skeletons permanently and cannot enable notifications at all. The KYT dialog and export-history enable-notifications flows terminate on a dead page. The state is server-side, so this survives remount, restart and reinstall.

## Recommendation

Distinguish *loaded-and-empty* from *not-loaded* rather than suppressing the former, and keep the anti-wipe intent on the **write** side instead of the read side:

1. Carry an explicit loaded marker, e.g. `settingsState = { identityKey, settings, loaded: true }` with `settings = {}`, and render real switches once `loaded` is true.
2. Have the first mutation on an empty config POST a **complete** config — explicit booleans for `pushEnabled` and all five sub-keys — rather than the bare partial that motivated this guard.
3. Keep refusing to write `{}` into `cachedNotificationSettings`, so a warm cache still only ever holds a real config.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -78,19 +101,46 @@ function NotificationsSettingsHelper() {
...
+  const isLoggedIn = Boolean(
+    primePersistAtom.isLoggedIn &&
+    primePersistAtom.isLoggedInOnServer &&
+    primePersistAtom.onekeyUserId,
+  );
+  const notificationSettingsIdentityKey = isLoggedIn
+    ? `onekey-id:${primePersistAtom.onekeyUserId}:prime:${
+        primePersistAtom.primeSubscription?.isActive === true
+      }`
+    : 'anonymous';
```
sidmorizon: # 🟠 [P1] Every not-fully-logged-in state collapses into one shared `'anonymous'` bucket, so an authenticated user's config can be written to and served from it

## Problem

```ts
const isLoggedIn = Boolean(
  primePersistAtom.isLoggedIn &&
  primePersistAtom.isLoggedInOnServer &&
  primePersistAtom.onekeyUserId,
);
const notificationSettingsIdentityKey = isLoggedIn
  ? `onekey-id:${primePersistAtom.onekeyUserId}:prime:${
      primePersistAtom.primeSubscription?.isActive === true
    }`
  : 'anonymous';
```

The key requires all three fields at once, and anything short of that folds into a single literal bucket. Two windows put a genuinely authenticated session into that bucket, because the API auth token does not come from this atom — the request still succeeds as user A while the key reads `'anonymous'`:

- **Split projection.** `isLoggedIn` (local) and `isLoggedInOnServer` (server) are separate persisted flags that transiently disagree. The repo treats this as a real, recoverable state: `ServicePrime.tsx:2223` `commitIncompleteLocalOneKeyIdLogoutRepair` and the `'incompleteLogoutProjection'` repair type (`ServicePrime.tsx:2286`, `shared/src/logger/scopes/prime/scenes/subscription.ts:160`) exist specifically to repair it.
- **Degraded boot.** `GlobalJotaiReady.tsx:33` force-releases its gate after `GATE_SAFETY_TIMEOUT_MS = 5000` and mounts React with default atom values — its own comment says "first render may see default values". `primePersistAtomInitialValue` has `isLoggedIn: false, isLoggedInOnServer: false, onekeyUserId: undefined` (`atoms/prime.ts:28-31`), so the key is `'anonymous'` for a logged-in user until the atom hydrates.

Credit where due: OneKey ID logout, login-as-a-different-user, and Prime `isActive` flips all **do** change the key, so 986e351e fixes those three transitions correctly. This is the residual gap, not a rejection of the approach.

## Impact

Cross-identity bleed of the same class 986e351e set out to eliminate: user A's notification config is written into `'anonymous'` and then rendered to any later logged-out session, or to user B during B's own partial-projection window. The module cache lives for the whole process lifetime, so it outlasts the transient condition that created it.

## Recommendation

1. Derive the key from `onekeyUserId` whenever it is present, independent of the two booleans.
2. Add a third, explicitly **non-cacheable** state for "identity indeterminate" — atom not hydrated, or `isLoggedIn !== isLoggedInOnServer` — and skip both the cache read and the cache write in that state instead of aliasing it to `'anonymous'`. Gating on `usePrimeInitAtom().isReady` closes the hydration half directly.

Separately worth tracking, though outside this increment: the background cache `getServerSettingsWithCache` (`ServiceNotification.ts:1576`) still carries `// TODO clear cache if prime expired, onekeyID logout` and is not identity-scoped at all, so `fetchServerNotificationSettingsWithCache` consumers keep the unscoped behavior.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -78,19 +101,46 @@ function NotificationsSettingsHelper() {
...
+  );
   const [shouldShowDevPanel, setShouldShowDevPanel] = useState(false);
   const pendingSettings = useRef<INotificationPushSettings | undefined>(
     undefined,
   );
+  const settingsMutationVersionRef = useRef(0);
+  const notificationSettingsIdentityKeyRef = useRef(
+    notificationSettingsIdentityKey,
+  );
+  notificationSettingsIdentityKeyRef.current = notificationSettingsIdentityKey;
```
sidmorizon: # 🟡 [P2] Ref is mutated during render, so an abandoned concurrent render can make the stale-identity guard reject a valid response

## Problem

```ts
const notificationSettingsIdentityKeyRef = useRef(
  notificationSettingsIdentityKey,
);
notificationSettingsIdentityKeyRef.current = notificationSettingsIdentityKey;
```

consumed at L162-167 and L212-215. This is a render-phase mutation of non-React state. On React 19 concurrent rendering, a render may be started, interrupted and discarded — `GlobalJotaiReady` itself uses `startTransition`. If a render computes a new key, writes the ref, and is then abandoned, the ref holds a key that was never committed and the guard rejects a response belonging to the *committed* identity.

To be clear, the guard is not dead code and is worth keeping: the ref holds the latest rendered key while the closure holds the key from the render that created that `reloadSettings` instance, so after an identity change an already-suspended call from the old instance correctly compares latest-B against closure-A and bails. It is genuinely load-bearing, because the `reloadSettings(updated)` path from `doUpdateSettingsToServer` deliberately skips the mutation-version check (`if (!updated && ...)`), leaving this guard as its only protection. The defect is in *how* the ref is maintained, not in having it.

Reading the mutable module variable `cachedNotificationSettings` during render (L115-117, L133) is the same class of concurrent-unsafe access.

## Impact

A wrongly-rejected response sets no state, so `settings` stays `undefined` and the page lands in the permanent-skeleton state described in my comment on L170 — with no retry to escape it. Rare, but silent and unrecoverable when it happens.

## Recommendation

Drop the render-phase ref and make the check functional, which is concurrent-safe and already the pattern `updateSettings` uses — guard inside the updater:

```ts
setSettingsState((prev) =>
  prev.identityKey === notificationSettingsIdentityKey ? { ...next } : prev,
);
```

and gate the `cachedNotificationSettings` write on the same comparison against the closure key. If you prefer to keep the ref, assign it in a `useEffect`/`useLayoutEffect` rather than during render.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -134,12 +209,23 @@ export default function NotificationsSettings() {
...
+        };
+        setSettingsState({
+          identityKey: notificationSettingsIdentityKey,
+          settings: prevSettings.current,
+        });
       }
       throw e;
     }
-  }, [reloadSettings]);
+  }, [notificationSettingsIdentityKey, reloadSettings]);
```
sidmorizon, line 276: 🟠 P1: Identity changes flush stale settings

Severity: `severe`

When `notificationSettingsIdentityKey` changes during debouncing, `updateSettingsToServer` is recreated with pending settings from the previous identity. Its [cleanup](packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx:281-289) flushes them after the server session changes, potentially overwriting another user's configuration.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 532790b1a6. A resolved empty/absent config is represented as a loaded empty state, so the master switch becomes available instead of leaving permanent skeletons. The first mutation expands that state to a complete six-boolean config before submitting. Empty snapshots are not retained in the memory cache.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 532790b1a6. Cache access now waits for primeInitAtom.isReady. When a OneKey user ID is present it determines the key independently of the two login flags; anonymous is used only after initialization when both login flags are false. Split or incomplete identity states without a user ID are non-cacheable.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -154,22 +240,43 @@ export default function NotificationsSettings() {
...
+    settingsMutationVersionRef.current += 1;
+    const cachedSettings = getCachedNotificationSettings(
+      notificationSettingsIdentityKey,
+    );
+    setSettingsState({
+      identityKey: notificationSettingsIdentityKey,
+      settings: cachedSettings,
+    });
+    prevSettings.current = cachedSettings;
+    pendingSettings.current = undefined;
```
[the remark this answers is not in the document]
huhuanming: Fixed in 532790b1a6. Identity changes explicitly cancel the debounce and clear the old pending snapshot before reloading. The flush cleanup is now registered only for a true component unmount and calls the latest debounced function through a ref.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 532790b1a6. The identity ref is updated in an effect instead of during render. Render no longer reads the mutable module cache for identity transitions; a mismatched state renders as unloaded until the committed identity effect restores a matching cache snapshot.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 532790b1a6. Identity changes cancel the pending debounce and discard the previous identity's pending settings. Flush is isolated to the empty-dependency unmount cleanup via a ref to the latest debounced callback.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -78,19 +125,50 @@ function NotificationsSettingsHelper() {
...
-  const [settings, setSettings] = useState<
-    INotificationPushSettings | undefined
-  >();
+  const [primeInitAtom] = usePrimeInitAtom();
+  const [primePersistAtom] = usePrimePersistAtom();
+  let notificationSettingsIdentityKey: string | undefined;
+  if (primeInitAtom.isReady && primePersistAtom.onekeyUserId) {
+    notificationSettingsIdentityKey = `onekey-id:${
+      primePersistAtom.onekeyUserId
+    }:prime:${primePersistAtom.primeSubscription?.isActive === true}`;
```
sidmorizon: 🟡 P2: Login skips authenticated settings reload

Severity: `non-severe`

During server login, `notificationSettingsIdentityKey` becomes user-scoped as soon as `onekeyUserId` exists. The unchanged key prevents `reloadSettings` retrying after authentication, leaving anonymous settings or permanent skeletons.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -151,34 +268,70 @@ export default function NotificationsSettings() {
...
-    [updateSettingsToServer],
+    [notificationSettingsIdentityKey, updateSettingsToServer],
   );
 
   useEffect(() => {
-    void reloadSettings();
-  }, [reloadSettings]);
+    settingsMutationVersionRef.current += 1;
+    updateSettingsToServer.cancel();
+    pendingSettings.current = undefined;
```
sidmorizon: 🟡 P2: Identity changes discard queued settings

Severity: `non-severe`

An identity change during the debounce cancels `updateSettingsToServer` and clears `pendingSettings`. The latest notification changes never reach the server.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -78,19 +125,50 @@ function NotificationsSettingsHelper() {
 
 export default function NotificationsSettings() {
   const intl = useIntl();
-  const [settings, setSettings] = useState<
-    INotificationPushSettings | undefined
-  >();
+  const [primeInitAtom] = usePrimeInitAtom();
+  const [primePersistAtom] = usePrimePersistAtom();
+  let notificationSettingsIdentityKey: string | undefined;
+  if (primeInitAtom.isReady && primePersistAtom.onekeyUserId) {
```
sidmorizon: # 🟠 [P1] Gating the identity key on `primeInitAtom.isReady` makes this page wait on the deferred Prime bootstrap, and strands it permanently if that chain never completes

## Problem

The key is now `undefined` until `primeInitAtom.isReady`:

```ts
if (primeInitAtom.isReady && primePersistAtom.onekeyUserId) {
```

and while it is `undefined` the page issues **no request at all**:

```ts
    if (notificationSettingsIdentityKey) {
      void reloadSettings();
    }
  }, [notificationSettingsIdentityKey, reloadSettings, updateSettingsToServer]);
```

`primeInitAtom` is a **non-persist** `globalAtom` (`atoms/prime.ts:111-117`, `initialValue: { isReady: false }`) whose **only** writer is `PrimeGlobalEffect.tsx:285-288`, inside the `finally` of an effect that awaits `apiFetchPrimeUserInfo()`. Reaching that writer requires a chain of deliberate deferrals:

- `PrimeGlobalEffectLazy.tsx:15` — `const PRIME_GLOBAL_EFFECT_DELAY_MS = 1500;`, an unconditional timer before mount is even requested, on every platform;
- a dynamic `import('../../views/Prime/hooks/PrimeGlobalEffect')` (chunk fetch);
- `PrimeGlobalEffect.tsx:387` renders the child only when `useOneKeyAuth().isReady`, which on web sits behind `SupabaseAuthProvider.tsx:26` `WEB_SUPABASE_AUTH_START_DELAY_MS = 6000` (settled early only by user interaction);
- then a network round trip.

Before this increment the key came synchronously from the already-hydrated `primePersistAtom` and resolved to `'anonymous'` or a real key on first render, so `reloadSettings()` fired on mount. Note `isReady` does not mean "persisted atom hydrated" — persist hydration is already gated app-wide by `jotaiInit.preloadAtomStorageValues` before any UI renders. It means "the Prime bootstrap completed one user-info fetch attempt", which is strictly later.

The chunk-load failure path never retries — the load effect deps are `[ContainerImpl, loadRequestSeq, shouldMount]` and nothing bumps `loadRequestSeq` on failure:

```ts
      .catch((error: Error) => {
        console.error('Failed to load PrimeGlobalEffect:', error);
      });
```

so `ContainerImpl` stays `null`, `PrimeGlobalEffect` never mounts, and `isReady` never becomes `true`.

## Impact

Any user opening Settings → Notifications in the first seconds after launch gets an all-skeleton page — master switch *and* all five sub-toggles — with **zero network activity**, for at least ~1.5s plus chunk load plus a Prime user-info round trip, and on web potentially several seconds more. This is unrelated to their notification state: a user who has never touched Prime is blocked by a *Prime* bootstrap.

If the lazy chunk 404s (stale asset after a release) or `apiFetchPrimeUserInfo()` never settles, the page is permanently stuck on skeletons with no usable controls — the same "stranded on permanent skeletons" failure mode this PR set out to eliminate, reintroduced through a different gate.

## Recommendation

Gate the *bucket disambiguation*, not the load. Treat an unresolved Prime state as "fetch under the best-known key" rather than "render nothing":

1. Keep the `undefined` key only for the genuinely ambiguous mixed projection (`isLoggedIn !== isLoggedInOnServer` with no `onekeyUserId`).
2. Let `!onekeyUserId && !isLoggedIn && !isLoggedInOnServer` resolve to `'anonymous'` **without** requiring `primeInitAtom.isReady` — `primePersistAtom` is fully hydrated before first render, so it is a sound synchronous source.
3. If a readiness gate is still wanted, bound it with a timeout after which the page falls back to the persisted-atom projection.

The identity-scoping fix itself is correct and worth keeping — this is about not coupling the fetch to Prime initialization.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,19 +177,47 @@ export default function NotificationsSettings() {
...
+      ) {
+        return;
+      }
+      if (
+        notificationSettingsIdentityKeyRef.current !==
+        notificationSettingsIdentityKey
+      ) {
+        return;
+      }
+      const nextSettings = result ?? {};
```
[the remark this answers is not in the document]
huhuanming: Fixed in 911cd26caf. Empty or missing revalidation results are now ignored when the current identity already has a valid cached snapshot. A true cold empty document still enters the explicit empty state so the master switch remains usable.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 911cd26caf. Cache identity is now the OneKey user ID only, while local login, server login, and Prime status independently trigger revalidation. The transition to authenticated server state therefore retries the fetch without discarding same-user state.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Handled in 911cd26caf. Login projection and Prime changes no longer count as identity changes, so they revalidate without canceling the same user pending update. A real user/anonymous identity change still cancels the old pending write for cross-identity safety; before canceling, its optimistic cache is restored to the last server snapshot so it cannot reappear as unsaved state.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 911cd26caf. The primeInitAtom gate was removed. An explicit logged-out projection resolves to anonymous immediately from the persisted atom, a present OneKey user ID is always user-scoped, and only the mixed no-user-ID projection remains non-cacheable.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,19 +181,55 @@ export default function NotificationsSettings() {
...
+        return;
+      }
+      if (
+        !hasNotificationSettings(result) &&
+        hasNotificationSettings(
+          getCachedNotificationSettings(notificationSettingsIdentityKey),
+        )
+      ) {
+        return;
+      }
```
sidmorizon, line 207: 🟡 P2: Empty settings preserve stale controls

Severity: `non-severe`

When revalidation returns `{}`, `reloadSettings` retains non-empty cached settings. Controls stay stale, and the next edit can resend obsolete values.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -151,34 +280,94 @@ export default function NotificationsSettings() {
...
+    if (notificationSettingsIdentityKey) {
+      void reloadSettings();
+    }
+  }, [
+    notificationSettingsIdentityKey,
+    primePersistAtom.isLoggedIn,
+    primePersistAtom.isLoggedInOnServer,
+    primePersistAtom.primeSubscription?.isActive,
+    reloadSettings,
+    updateSettingsToServer,
```
sidmorizon, line 375: 🟡 P2: Revalidation can revert pending edits

Severity: `non-severe`

When Prime status changes during `pendingSettings`, revalidation races the queued update. An older query can arrive last and replace the confirmed settings.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -151,34 +280,94 @@ export default function NotificationsSettings() {
...
+      previousNotificationSettingsIdentityKeyRef.current;
+    const identityChanged =
+      previousIdentityKey !== notificationSettingsIdentityKey;
+    previousNotificationSettingsIdentityKeyRef.current =
+      notificationSettingsIdentityKey;
+    settingsMutationVersionRef.current += 1;
+    if (identityChanged) {
+      if (
+        previousIdentityKey &&
+        pendingSettings.current &&
```
sidmorizon: # 🟡 [P2] Identity-change restore misses an in-flight update, leaving the previous identity's cache holding a never-persisted value

## Problem

The new restore branch only fires while the optimistic value is still *debounced*:

```ts
if (
  previousIdentityKey &&
  pendingSettings.current &&
  prevSettings.current !== undefined
) {
  updateCachedNotificationSettings({
    identityKey: previousIdentityKey,
    settings: prevSettings.current,
  });
}
```

But `doUpdateSettingsToServer` clears `pendingSettings.current` the moment it starts the POST, before the `await`:

```ts
const settingsToUpdate = pendingSettings.current;
pendingSettings.current = undefined;
```

So if the identity changes while that POST is in flight, `pendingSettings.current` is already `undefined`, the restore is skipped, and the slot keeps the optimistic value written by `updateSettings`.

The rollback cannot repair it either, because its own guard rejects the changed identity — after the change, `notificationSettingsIdentityKeyRef.current` holds the **new** key while the closure holds the **old** one:

```ts
if (
  notificationSettingsIdentityKeyRef.current === notificationSettingsIdentityKey &&
  prevSettings.current
) { ... }
```

So on a failed POST the previous identity's cache is left holding a value the server never accepted.

## Impact

On the next visit under that identity, the module cache seeds the switches from a rejected value, and it stays interactive for one revalidation round trip; a toggle in that window submits it against the full-replace `/config/update`. It self-heals after one successful revalidation, and triggering it needs an identity change mid-POST *plus* a failure — so exposure is narrow. Flagging as P2, not a merge blocker.

## Recommendation

Widen the condition to cover the in-flight case: keep the value handed to the server (`settingsToUpdate`) in a ref alongside `pendingSettings`, and restore when **either** is outstanding. Alternatively, have the `catch` in `doUpdateSettingsToServer` repair the *captured* identity key's cache rather than only the current one — it already has `notificationSettingsIdentityKey` in scope from its own closure.

For what it's worth, the `prevSettings.current !== undefined` condition (rather than a truthiness check) is deliberately correct here — with `{}` the restore falls through and deletes the slot, which is right, since at that point it holds an optimistic never-sent value and the identity's true server state was empty.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -151,34 +280,94 @@ export default function NotificationsSettings() {
...
   useEffect(() => {
-    void reloadSettings();
-  }, [reloadSettings]);
+    const previousIdentityKey =
+      previousNotificationSettingsIdentityKeyRef.current;
+    const identityChanged =
+      previousIdentityKey !== notificationSettingsIdentityKey;
+    previousNotificationSettingsIdentityKeyRef.current =
+      notificationSettingsIdentityKey;
+    settingsMutationVersionRef.current += 1;
```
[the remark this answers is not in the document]
huhuanming: Fixed in 06cf73fa44. A flag-only revalidation now returns before bumping the mutation version or starting a query whenever the same identity has a debounced or in-flight update. The update completion path remains responsible for applying the confirmed result.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 06cf73fa44. Missing data and an explicit empty object are now handled separately: undefined preserves a valid last-known snapshot, while an explicit server {} clears the cache and updates the UI to the loaded empty state.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 06cf73fa44. Login/Prime-triggered revalidation is skipped while pendingSettings or an update request is active, so an older query cannot replace the optimistic edit. The normal update completion still reconciles the UI.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in 06cf73fa44. Identity-change restoration now treats isUpdating as outstanding in addition to pendingSettings. It therefore restores the previous identity last confirmed snapshot even after the pending ref has been cleared at POST start.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -134,12 +252,23 @@ export default function NotificationsSettings() {
...
+        updateCachedNotificationSettings({
+          identityKey: notificationSettingsIdentityKey,
+          settings: prevSettings.current,
+        });
+        setSettingsState({
+          identityKey: notificationSettingsIdentityKey,
+          settings: prevSettings.current,
+        });
       }
       throw e;
```
sidmorizon, line 274: 🟡 P2: New identity update remains unsaved

Severity: `non-severe`

After an identity switch, `isUpdating` makes the new identity's debounce skip its pending change while the old request runs. If the old request then fails, its catch never retries `pendingSettings`. The new notification choice remains optimistic locally but is never persisted.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -99,19 +181,55 @@ export default function NotificationsSettings() {
...
+        return;
+      }
+      if (
+        notificationSettingsIdentityKeyRef.current !==
+        notificationSettingsIdentityKey
+      ) {
+        return;
+      }
+      if (
+        result === undefined &&
```
sidmorizon: # 🟡 [P2] `result === undefined` misses `data: null`, splitting it from `undefined` against this repo's own convention

## Problem

The narrowed guard is strict:

```ts
      if (
        result === undefined &&
        hasNotificationSettings(
          getCachedNotificationSettings(notificationSettingsIdentityKey),
        )
      ) {
        return;
      }
```

`fetchServerNotificationSettingsWithoutToast` returns `result?.data?.data`, and optional chaining guards only the **left** operand — a body of `{"code":0,"data":null}` yields `null`, not `undefined`. Since `null === undefined` is false, the guard is skipped and `const nextSettings = result ?? {}` (line 211) produces `{}`, clearing the cache and flipping a warm user to loaded-empty. The outgoing `!hasNotificationSettings(result)` covered `null`.

The type system gives no cover: `IApiClientResponse<T>` declares `data: T` as non-optional (`packages/shared/types/endpoint.ts:28-32`), so `result` is typed `INotificationPushSettings` and `result === undefined` is an off-contract runtime-only check that TS will not flag.

The decisive point is consistency: the repo's two sibling consumers both class `null` **with** `undefined`, not with `{}`:

- `NotificationEnableAlert.tsx:64-65` — `// If serverSettings is undefined (API failed), don't show the alert` / `if (!serverSettings) {` — catches `null` and `undefined` alike.
- `notificationPermissionUtils.ts:182` — `const hasServerSettings = !!serverSettings && Object.keys(serverSettings).length > 0;` — likewise.

This is the one shape where the new guard diverges from the convention it is otherwise correctly implementing.

## Impact

If the backend ever represents "no config record" as `"data":null` — a common REST convention — rather than an omitted key, the preserve-guard becomes dead code on every query, and the "an `undefined` response preserves the last known-good snapshot" property this commit is built on is void: every such response would null the warm snapshot.

## Recommendation

Widen to the nullish form, which keeps `{}` distinct while restoring the intended "no settings object was returned" semantics:

```ts
      if (
        result == null &&
        hasNotificationSettings(
```

(or `(result === undefined || result === null) &&` if the loose `==` trips lint).

---

Separately, and for the record since it concerns this same line: I looked hard at the argument that this change should be reverted to `!hasNotificationSettings(result)`, and I disagree with it. The two cases are cleanly distinguishable at runtime and a transient failure cannot surface as `{}`: a non-2xx rejects at axios, and `code !== 0` **throws** before the guard is reached (`axiosInterceptor.ts:380`, with `autoHandleError` defaulting to `true` in `appApiClient.ts:60` — only `getRawDataClient` opts out, and `ServiceBase.getClient` does not use it). So `{}` is reachable only from 2xx + `code === 0` + a literal `"data":{}`, i.e. an explicit empty config document. Treating that as loaded-empty is also what `NotificationEnableAlert` already does (`{}` is truthy there, so it shows the enable CTA), and it is the contract `enableNotificationsBestEffort` depends on when it routes empty-config users to this page as the owner of "master-switch semantics for an empty config". Reverting would leave a warm-cache user in that population unable to ever reach the loaded-empty state. The `null` gap above is the real residual, and it is one token to close.

[Claude session ↗](https://claude.ai/code/session_01Gov5sFej8dLvVBVmoDBFuu)

---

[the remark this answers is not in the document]
huhuanming: Reviewed in f3320e831f. I am keeping explicit {} as a loaded empty server config: this is a successful code=0 document and must replace stale controls, while the first mutation still expands it to the complete defaults object before the whole-config POST. The preserve guard now covers only null/undefined, which are the missing-document shapes. Treating {} as failure would prevent a warm-cache user from ever observing a legitimate server-side empty config.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in f3320e831f. When an older identity request finishes, both success and failure paths now hand any remaining pending update to a ref containing the latest identity update function. A stale callback also checks the current identity before consuming pending settings, so it cannot take the new identity payload.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
huhuanming: Fixed in f3320e831f. The missing-document guard now handles both null and undefined. An explicit {} remains distinct and is applied as a valid loaded empty config.

packages/kit/src/views/Setting/pages/Notifications/NotificationsSettings.tsx
```diff
@@ -43,6 +44,57 @@ import {
...
+  identityKey,
+  settings,
+}: {
+  identityKey: string;
+  settings: INotificationPushSettings;
+}) {
+  if (hasNotificationSettings(settings)) {
+    cachedNotificationSettings = { identityKey, settings };
+  } else if (cachedNotificationSettings?.identityKey === identityKey) {
+    cachedNotificationSettings = undefined;
```
sidmorizon: 🟡 P2: Empty settings never enter cache

Severity: `non-severe`

A valid empty configuration makes `updateCachedNotificationSettings` remove its cache entry. Every later page entry shows loading skeletons until revalidation finishes.

Suggested fix:

```
  cachedNotificationSettings = { identityKey, settings };
```

huhuanming: Fixed in eea01e7eaf. Explicit empty settings are a valid loaded configuration and are now stored in the module cache, so subsequent page entries render the master switch immediately instead of returning to skeletons.

sidmorizon (approved):

ezailWang (approved):

limichange (approved):

PR #13379, opened 2026-09-14
[OK-62730][OK-62869] fix: stabilize wallet action flows

zhaono1:
[OK-62730](<https://onekeyhq.atlassian.net/browse/OK-62730>) [OK-62869](<https://onekeyhq.atlassian.net/browse/OK-62869>)

----------

## Summary

- Fix Android Swap cold-start loading by preventing position prefetch from being serialized behind the slow or failed Pro network configuration.
- Keep Swap, Stock, and Pro prefetch scopes stable while allowing each ready source to start independently.
- Fix Aave E-Mode repayment balance display so non-zero dust balances are not truncated to `0`.
- Block watch-only accounts before password/device signing flows, preventing them from entering hardware processing and leaving a PIN/connection state behind.

## Validation

- `yarn test packages/kit/src/views/Borrow/pages/BorrowEModeNeedAction/needActionBalances.test.ts --runInBand`
- `yarn test packages/kit/src/views/Borrow/pages/BorrowEModeNeedAction/useEModeNeedActionFlow.test.tsx --runInBand`
- `yarn test packages/kit-bg/src/services/ServiceSend.precheckReplaceTxNonce.test.ts --runInBand`
- `yarn tsc:only`
- Targeted ESLint
- `yarn agent:check --profile pr` local checks passed. Existing GitHub CI has one failure in `Native startup graph budget (24.x)`; the other 24 checks passed.
- Android emulator verification on `emulator-5554` under slow4g profile.
  - Before: Trade tap settled at 772 ms but still showed the loading spinner.
  - After: Trade tap settled at 599 ms with the Swap page rendered.

## Jira

- OK-62730
- OK-62869

[OK-62730]: https://onekeyhq.atlassian.net/browse/OK-62730?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62869]: https://onekeyhq.atlassian.net/browse/OK-62869?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-a34ae28f-352d-50b8-8a05-c0b0cd330d67)

sidmorizon:
Claude review session: https://claude.ai/code/session_019v2jjn2cPDxYoWFASo3kjm

packages/kit/src/views/Swap/utils/swapPositionPrefetchUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+      networkList: stockNetworkList,
+      stockOnly: true,
+    },
+    {
+      key: 'pro',
+      networkList: proNetworkList,
+      stockOnly: false,
+    },
+  ];
+  return scopes.filter(
```
sidmorizon, line 40: # 🟡 [P2] Prefetch scope order no longer prioritizes the visible tab, so Stock/Pro positions load last on cold start

## Problem

`buildSwapPositionPrefetchScopes` always returns a fixed `swap → stock → pro` order. The code it replaces sorted the *active* tab's scope to index 0 (the `toSorted` on `activeKey` removed from `SwapMainLand.tsx`). That order is not cosmetic — it is the fetch priority for the whole load generation:

- `packages/kit/src/views/Swap/pages/components/SwapMainLand.tsx:1357` destructures `const [primaryScope, ...additionalScopes] = positionPrefetchScopes;` and passes `primaryScope` as the primary argument.
- `packages/kit/src/views/Swap/hooks/useSwapPro.ts:1658` forwards the rest as `additionalSupportNetworkScopes`.
- `packages/kit/src/views/Swap/utils/swapProPositionsLoader.ts:76-82` builds `requestedScopes = [primary, ...additional]`, preserving that order into `scopesToLoad`.
- `packages/kit/src/views/Swap/utils/swapProPositionsLoader.ts:240-246` sorts the per-network fetch tasks by `scopesToLoad[0]?.networkIds`, and line 339 runs them through `executeBatched(tasks, 3)` — only three in flight at a time.

The rationale for dropping the sort does not hold: the effect still re-runs on `focusSwapPro`/`swapTypeSwitch` (deps at `SwapMainLand.tsx:1370-1375`), and the loader already dedups in-flight scopes via `activeRequestIds` and the runtime TTL (`swapProPositionsLoader.ts:59-70`). Reordering the scopes was already a no-op for request churn — only the priority benefit was lost.

## Impact

On a cold start straight into the Stock or Pro tab, `scopesToLoad[0]` is now the `swap` scope. Because the Stock/Pro network sets are normally subsets of the swap-capable set, the comparator returns `0` for essentially every pair and the queue degenerates to the unsorted union order. The networks the user is actually looking at are no longer pulled to the front of a 3-at-a-time queue over ~20 networks, so the visible tab's positions surface later than they did before this change — a regression against this PR's own cold-start goal.

## Recommendation

Keep the extracted builder, but restore the priority: add an `activeKey: 'swap' | 'stock' | 'pro'` parameter (derived in `SwapMainLand` from `focusSwapPro`/`swapTypeSwitch`) and move that scope to index 0 after the `filter` here at `swapPositionPrefetchUtils.ts:40-44`. The returned *set* stays stable; only index 0 changes, which is exactly what the loader consumes for priority. The new `swapPositionPrefetchUtils.test.ts` case `uses one stable tab-independent scope order` should then assert "active scope first, remaining scopes in stable order".

[Claude session ↗](https://claude.ai/code/session_019v2jjn2cPDxYoWFASo3kjm)

---

packages/kit/src/views/Borrow/pages/BorrowEModeNeedAction/index.tsx
```diff
@@ -814,6 +825,7 @@ function BorrowEModeNeedActionView() {
           loading: isBusy || pendingGuardBlocksAction || isChecking,
           disabled:
             isBusy ||
+            isWatchingAccount ||
```
sidmorizon, line 828: # 🟡 [P2] Watch-only guard also disables the read-only "Retry check" action, leaving a dead-end page

## Problem

The footer button is dual-purpose. When `canRetryCheck` is true (defined at L690-694), `onConfirm` is `refresh` (L836) — a read-only blocker recheck that never broadcasts, as the comment at L695-697 states — and the label becomes "Retry" (L818-822). The new `isWatchingAccount ||` term on this line is unconditional, so it disables that non-signing retry along with the signing action.

## Impact

`canRetryCheck` is true when `!check`. `runCheck` sets `check = null` and clears `isChecking` on a failed request (`packages/kit/src/views/Borrow/pages/BorrowEModeSwitch/useEModeSwitch.ts:93-107`). With no prior successful check, `displayCheck` is null (L393), so `isEModeBlockerDataUnavailable(null)` returns `false` (`needActionContract.ts:10-12`) and the inline retry `Alert` at L728-743 is **not** rendered. The footer "Retry" is then the only recovery affordance on the screen, and for a watch-only account it is now permanently greyed out — the user cannot re-run a read-only check and has to back out of the modal. The page is meant to be viewable by watch-only accounts; this PR's own `Alert` at L718-726 presupposes exactly that.

## Recommendation

Scope the guard to the signing branch only, at L826-834:

```tsx
disabled:
  isBusy ||
  (!canRetryCheck && isWatchingAccount) ||
  pendingGuardBlocksAction ||
  (!canRetryCheck && ( ... )),
```

The new `assertAccountCanSign` backstop in `packages/kit-bg/src/services/ServiceSend.ts:102-110` still blocks any signing path that would slip through, so narrowing the UI guard does not reopen the original bug.

[Claude session ↗](https://claude.ai/code/session_019v2jjn2cPDxYoWFASo3kjm)

---

originalix (approved):

PR #7864, opened 2025-07-23
fix: dismiss keyboard on scroll && update tab width on window resize && align networkId in invite page &&  tap-outside to hide keyboard on lock screen && add Keyboard Component

huhuanming:
## Summary by CodeRabbit

* **New Features**
  * Introduced a consolidated Keyboard utility, providing unified access to keyboard-related components and functions.
  * Enhanced Tabs components with improved width handling, scroll snapping, and customizable key extraction for list items.
  * Added support for percentage-based bottom sheet snap points in Select components.

* **Improvements**
  * Optimized keyboard behavior in list and sortable list views for smoother user interactions.
  * Refined tab container width calculation for better responsiveness on web platforms.
  * Improved initial network selection logic in referred wallet addresses view.
  * Enhanced styling flexibility and cache handling in various list and history components.
  * Enabled keyboard dismissal by tapping the lock screen background.
  * Adjusted keyboard dismissal and tap handling behavior in list views.
  * Added horizontal padding support in NFT list view for improved layout.

* **Bug Fixes**
  * Removed unused imports and redundant code for cleaner components.
  * Removed unnecessary divider component from popover for cleaner UI.

* **Chores**
  * Updated exports to include new Keyboard utilities.

PR #13380, opened 2026-09-14
fix: correct Android WebView keyboard layout

Hanssen0:
This PR fixes a bug on Android where the height of WebView is mistakenly calculated when the keyboard is shown.

---

In the original code, the height of the WebView is set to increase together with the height of the keyboard:

```tsx
height: keyboardHeight + (StatusBar.currentHeight || 0) + 60,
```

The WebView height depended on the keyboard size rather than the actual available viewport, producing incorrect layouts across different screen and keyboard sizes.

This PR replaced the workaround introduced in #7607 with the existing `Keyboard.AvoidingView` introduced by #7864. It keeps the height-changing behaviour but uses `automaticOffset` to calculate it. Tested on a physical Android device for both the Discover and Intercom pages.

---

| Before | After |
| - | - |
| Screenshot_20260914_095107_OneKey (2) | Screenshot_20260914_095242_OneKey WebView Test~2 |
| Screenshot_20260914_095142_OneKey (2) | Screenshot_20260914_100240_OneKey WebView Test~2 |


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-7945cc67-2ea1-549d-8b7e-2925eada8fbf)

sidmorizon:
Claude review session: https://claude.ai/code/session_017WuX9qCeqM7SYy9YprMvws

packages/kit/src/components/WebView/InpageProviderWebView.native.tsx
```diff
@@ -268,7 +249,12 @@ const InpageProviderWebView: FC<INativeInpageProviderWebViewProps> = forwardRef(
     }
 
     return (
-      <Stack {...containerStyle}>
+      <Keyboard.AvoidingView
+        automaticOffset
```
sidmorizon: # ⚠️ P1 `automaticOffset` mixes two coordinate origins on Android, shrinking the WebView by an extra status-bar height

## Problem

`Keyboard.AvoidingView` resolves to `react-native-keyboard-controller@1.21.9`'s `KeyboardAvoidingView` (`packages/components/src/content/Keyboard/index.native.tsx:18`; pinned in `apps/mobile/package.json:180`). Its overlap math is:

```ts
const keyboardY = screenHeight - keyboard.heightWhenOpened.value - keyboardVerticalOffset;
return Math.max(frame.value.y + frame.value.height - keyboardY, 0);
```

With `automaticOffset` set (line 253), the two sides of that subtraction are no longer in the same coordinate space on Android:

| quantity | origin | source |
| --- | --- | --- |
| `frame.y` (only when `automaticOffset`) | top of the **physical display** — `viewPositionInWindow` → `View.getLocationOnScreen()` | `modules/KeyboardControllerModuleImpl.kt:105`, `extensions/View.kt:56` |
| `screenHeight` | height of **`androidx.appcompat.R.id.action_bar_root`** | `listeners/WindowDimensionListener.kt:54-58` + `extensions/ReactContext.kt:27-31` |
| `heightWhenOpened` | `imeInset.bottom − navigationBarInset.bottom`, i.e. measured from the **bottom of that same content view** | `listeners/KeyboardAnimationCallback.kt:434-445` |

`KeyboardProvider` is mounted with no props (`packages/kit/src/provider/index.tsx:122`), so `isStatusBarTranslucent` keeps its default `false` and the library gives `action_bar_root` a top margin of `systemBarInsets.top` (`views/EdgeToEdgeReactViewGroup.kt:120-128`, applied from `enable() → setupWindowInsets()`). Nothing in this repo calls `forceStatusBarTranslucent`, so that margin is always present. Substituting:

```
keyboardY = (displayH - statusTop - navBottom) - (imeBottom - navBottom)
          = displayH - statusTop - imeBottom          // keyboard top, in CONTENT coords
frame.y + frame.height                                 // view bottom, in DISPLAY coords
⇒ overlap_computed = overlap_correct + systemBarInsets.top
```

## Impact

Android only. When the keyboard opens, `behavior="height"` applies `{ height: frame.height - overlap, flex: 0 }`, so the WebView is shrunk roughly 24–48 dp (status-bar height, more on large-cutout devices) beyond what the keyboard actually covers. The user sees a background strip between the bottom of the web content and the top of the keyboard, and the page's `window.innerHeight` is correspondingly short, so DApp pages lay out against a viewport that does not match the visible area. That is a residual instance of the defect this PR is titled to fix — notably, the deleted code carried an explicit `StatusBar.currentHeight` term, consistent with this delta being the real discrepancy the original workaround was compensating for.

## Recommendation

On a device, give the parent a contrasting background and measure the gap between the web content and the keyboard; if it equals the status-bar height, compensate on this element (`overlap` grows with `keyboardVerticalOffset`, so the offset must be negative):

```tsx
<Keyboard.AvoidingView
  automaticOffset={platformEnv.isNativeAndroid}
  behavior="height"
  enabled={platformEnv.isNativeAndroid}
  keyboardVerticalOffset={-(StatusBar.currentHeight ?? 0)}
  style={styles.container}
>
```

Gating `automaticOffset` on `platformEnv.isNativeAndroid` is worth folding into the same edit regardless: today it is unconditional while `enabled` is Android-only, so on iOS every layout pass still pays an async `viewPositionInWindow` native round-trip plus a `runOnUI` hop whose result `animatedStyle` discards at its `!enabled` short-circuit.

The alternative — passing `statusBarTranslucent` to `KeyboardProvider` — removes the margin at the source but changes global app chrome, so it is out of scope here.

[Claude session ↗](https://claude.ai/code/session_017WuX9qCeqM7SYy9YprMvws)

---

Hanssen0: > Android only. When the keyboard opens, `behavior="height"` applies `{ height: frame.height - overlap, flex: 0 }`, so the WebView is shrunk roughly 24–48 dp (status-bar height, more on large-cutout devices) beyond what the keyboard actually covers. The user sees a background strip between the bottom of the web content and the top of the keyboard, and the page's `window.innerHeight` is correspondingly short, so DApp pages lay out against a viewport that does not match the visible area.

Confirmed to be a valid finding, but only affects devices before Android 15. [On Android 15, edge-to-edge is enabled by default and is forced enabled on Android 16+](https://developer.android.com/about/versions/16/behavior-changes-16#edge-to-edge). Applying the proposed fix will cause problems on mainstream devices instead.

A possible workaround is to import `isEdgeToEdge` detection from `react-native-is-edge-to-edge` to reduce the offset for edge-to-edge disabled devices only. However, I would rather say it's an upstream bug and should be fixed by them.

Note: They actually fixed a similar issue once https://github.com/kirillzyusko/react-native-keyboard-controller/pull/468, but it's not for `automaticOffset`.

> Gating `automaticOffset` on `platformEnv.isNativeAndroid` is worth folding into the same edit regardless

Fixed in https://github.com/OneKeyHQ/app-monorepo/pull/13380/commits/7a8a4055e76c91742fcba0003f252bbb30038c9a.

sidmorizon: Thanks — the `automaticOffset` gating in 7a8a405 looks right.

On the Android 15 scoping, I don't think the version split holds, and I'd ask you to re-check before deciding this is upstream-only. The `systemBarInsets.top` margin is not a consequence of the *system's* edge-to-edge default; the library imposes edge-to-edge itself on every Android version and then re-insets its own content view:

- `KeyboardControllerViewManager.onAfterUpdateTransaction` calls `manager.setEdgeToEdge(view)` unconditionally, on both the Fabric and Paper managers (`android/src/fabric/.../KeyboardControllerViewManager.kt:30`, `android/src/paper/.../KeyboardControllerViewManager.kt:23`).
- `EdgeToEdgeReactViewGroup.setEdgeToEdge()` then calls `WindowCompat.setDecorFitsSystemWindows(window, false)` and clears `FLAG_FULLSCREEN` (`views/EdgeToEdgeReactViewGroup.kt:142-152`).

So the window is already edge-to-edge on Android 14 and below. What puts `action_bar_root` back below the status bar is `setupWindowInsets()`, whose top margin is a function of one prop only:

```kotlin
params.setMargins(
  navBarInsets.left,
  if (this.isStatusBarTranslucent) 0 else systemBarInsets.top,
  ...
)
```

`isStatusBarTranslucent` defaults to `false` (`EdgeToEdgeReactViewGroup.kt:46`), `KeyboardProvider` is mounted with no props (`packages/kit/src/provider/index.tsx:122`), and the only path that flips it is `StatusBarManagerCompatModuleImpl.forceStatusBarTranslucent`, which nothing in this repo calls. I also confirmed `action_bar_root` really exists here — `Theme.App.SplashScreen` → `AppTheme` → `Theme.AppCompat.Light.NoActionBar` (`apps/mobile/android/app/src/main/res/values/styles.xml`) — so the margin is applied rather than silently no-oping on a null content view.

Since `setDecorFitsSystemWindows(false)` is forced by the library on all versions and the margin depends only on `isStatusBarTranslucent`, `systemBarInsets.top` should be non-zero and the margin applied on Android 15/16 exactly as on Android 14. That would make the offset version-independent, and gating it on `react-native-is-edge-to-edge` would be conditioning on a variable that isn't the one driving the margin.

I can't run a device from here, so this is a source-level derivation rather than a measurement — if you can put a contrasting background behind the WebView on an Android 15 device and measure the gap, that settles it either way in a couple of minutes. If it does reproduce there, one refinement to my original suggestion: prefer the safe-area top inset from `react-native-safe-area-context` (`SafeAreaProvider` is already mounted above this tree) over `StatusBar.currentHeight`, since that is the value that actually tracks `systemBarInsets.top` under edge-to-edge.

Agreed it is an upstream bug in substance. Given this is the file the PR is specifically fixing, though, shipping it with a known residual status-bar-height error seems worth a local compensation until [kirillzyusko/react-native-keyboard-controller](https://github.com/kirillzyusko/react-native-keyboard-controller) carries a fix for `automaticOffset` — your call, and I'll leave the thread open rather than resolve it.

[Claude session ↗](https://claude.ai/code/session_017WuX9qCeqM7SYy9YprMvws)

---

Hanssen0: See https://github.com/kirillzyusko/react-native-keyboard-controller/blob/35a06b2acb8ffe96a0ccd7799e0e4d9c03338141/src/animated.tsx#L221

```tsx
statusBarTranslucent={IS_EDGE_TO_EDGE || statusBarTranslucent}
```

Also confirmed on my Android 16 device that the gap doesn't exist. See the screenshots from the PR description. I am inclined not to handle this issue additionally here, as it could silently introduce new problems when the upstream dependency is next updated. Even if the issue persists upstream, the current situation is already an improvement over the previous state.

Though, I might not be the right person to report the issue upstream, considering I don't have an Android 14 device to test the scenario where edge-to-edge is disabled.

sidmorizon: You're right and my previous reply was wrong — thanks for the pointer, it found a real hole in my trace.

I verified the line you cited in the pinned 1.21.9: `src/animated.tsx:221` passes `statusBarTranslucent={IS_EDGE_TO_EDGE || statusBarTranslucent}`, so `KeyboardProvider` forwards `true` regardless of what the app passes. My claim that mounting it bare leaves `isStatusBarTranslucent === false` was simply incorrect — I traced from the native side and from this repo's prop usage and never opened the JS provider that sits between them.

Following `IS_EDGE_TO_EDGE` through `react-native-is-edge-to-edge` confirms your version split is the right axis, not mine:

```js
// dist/index.android.js
const isEdgeToEdgeFromProperty = () =>
  TurboModuleRegistry.get("DeviceInfo")?.getConstants().isEdgeToEdge === true;
const isEdgeToEdge = () => isEdgeToEdgeFromLibrary() || isEdgeToEdgeFromProperty();
```

`react-native-edge-to-edge` isn't a dependency here, so `isEdgeToEdgeFromLibrary()` is false and it falls through to RN's `DeviceInfo.isEdgeToEdge` constant (RN 0.86.2), which reports the window's actual state. On Android 15/16 the system forces edge-to-edge, so that is `true` → `statusBarTranslucent` true → `setMargins(..., 0, ...)` → no status-bar term, and no gap. That matches your device observation exactly, and it means my suggested `keyboardVerticalOffset={-(StatusBar.currentHeight ?? 0)}` would have *introduced* a status-bar-height overcorrection on precisely the mainstream devices — a worse bug than the one I was reporting. Good catch.

So the residual is real only on pre-Android-15 devices that aren't edge-to-edge, it is upstream in `automaticOffset`, and any local compensation would have to be conditioned on the same `isEdgeToEdge()` signal to avoid regressing everything else. I agree with your call not to carry that here: it adds a version-conditional workaround that would silently rot on the next upstream bump, and the PR is a clear improvement over the previous `keyboardHeight + StatusBar.currentHeight + 60` behaviour either way.

Withdrawing the finding and resolving the thread.

[Claude session ↗](https://claude.ai/code/session_017WuX9qCeqM7SYy9YprMvws)

---

originalix:
<p>Thanks for contributing this fix and for taking the time to explain the edge-to-edge behavior. We tested the PR against the current <code dir="ltr">x</code> branch, and our results support your observations.</p><p>Using the same Discover WebView input scenario on Android 14 and Android 16, we compared <code dir="ltr">x</code> (<code dir="ltr">ec9007f441</code>) with this PR (<code dir="ltr">7a8a4055</code>):</p><div><div><div>

Device | Gap above the keyboard on x | With this PR
-- | -- | --
Android 14 | 259 px | 63 px (24 dp)
Android 16 | 101 px | 0 px

</div></div></div><p>Each result was reproduced across two keyboard open/close cycles. The screenshots below show the before/after comparison.</p><p>This is a clear improvement on both tested configurations. On Android 14, a small gap remains that matches the status-bar height; on Android 16, the WebView aligns with the keyboard correctly. These results do not support applying an unconditional negative offset across Android versions.</p><p>We're sharing the Android 14 screenshots as evidence for a potential upstream follow-up, rather than asking you to add a workaround to this PR. We haven't tested Android 15 or other WebView entry points, so the results are limited to the configurations above.</p><p>Thanks again for the contribution and the careful follow-up!</p>

Android 14 👇🏻
image

Android 16 👇🏻
image

PR #13383, opened 2026-09-14
fix: restore Market perps entry and stock token switching ｜OK-62737 OK-63127

zy2000-star:
[OK-62737](<https://onekeyhq.atlassian.net/browse/OK-62737>) [OK-63127](<https://onekeyhq.atlassian.net/browse/OK-63127>)

----------

## Summary

Fix two Market detail regressions: the missing Perps entry in standard desktop/Web token details, and stale stock token selection after navigating from the mobile detail search.

## OK-62737 — Show the Perps entry in standard token details

The Top Coins desktop layout already displayed a Perps banner, but the standard token layout only rendered the swap panel. Tokens with a valid `perpsInfo.hlTicker` therefore had no visible entry to their corresponding perpetual market.

Add the existing `PerpetualTradingBanner` above the swap panel in `TokenDesktopLayout`.

- Show the entry when the detail response provides a Perps mapping.
- Reuse the existing target-contract navigation, click tracking, and dismissal behavior.
- Keep the current API and mapping logic unchanged.

## OK-63127 — Apply the selected stock token when the route changes

When a user opens AAPL details and then searches for AAPLX, navigation can reuse the existing stock detail page with updated network and token-address parameters.

Previously, `StockDetailProvider` retained its selected variant whenever that variant still existed in the stock’s token list. This check ran before matching the new route target, so the page could continue selecting the previous variant.

Meanwhile, the search navigation prepared the shared detail state for the newly selected token. The page could then request the previous token while the state expected the new token. Successful responses were discarded by the identity checks, leaving token details empty and loading on subsequent polling cycles.

The provider now tracks the last applied route identity using the stock ID, network ID, and token address.

- Re-evaluate the selected variant when the route identity changes.
- Wait for a successful variant-list response belonging to the current stock before applying the route.
- Apply each route identity once, so later polling preserves the user’s manual variant selection.
- Preserve existing stale-response validation.

For example, navigating from an AAPL Ondo variant to AAPLX now selects AAPLX. If the user subsequently switches back to Ondo manually, refreshing the variant list retains that selection.

## Changed Files

- `TokenDesktopLayout.tsx`: render the Perps banner above the standard token trade panel.
- `StockDetailContext.tsx`: synchronize variant selection with route changes while preserving manual selection during polling.
- `StockDetailContext.test.tsx`: cover route-driven variant switching and subsequent manual selection.

## Validation

- All 21 targeted tests passed across `StockDetailContext` and `PerpetualTradingBanner`.
- The new regression test verifies Ondo → AAPLX through a route update, followed by manual selection of Ondo and a variant-list refresh.
- `yarn agent:check --profile commit` passed, including lint, formatting, TypeScript, background API contract, and repository checks.
- The iOS Simulator build launched successfully. End-to-end verification of the two issue flows has not been completed.

## Manual Verification

- Desktop/Web: open a non-Top-Coins token with a Perps mapping, confirm the entry is visible, and verify it opens the correct contract.
- Desktop/Web: confirm tokens without a mapping do not display the entry.
- Mobile: open AAPL details, search for AAPLX, and confirm the selected variant, loaded detail data, and trade panel correspond to the target token.
- Mobile: manually select another variant and confirm polling does not reset the selection.

[OK-62737]: https://onekeyhq.atlassian.net/browse/OK-62737?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63127]: https://onekeyhq.atlassian.net/browse/OK-63127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-90c00a59-2bc0-531e-8e0e-72a91f2f6174)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Kg9yJ1HFKoio2QBmNvMdSH

packages/kit/src/views/Market/MarketDetailV2/components/PerpetualTradingBanner/PerpetualTradingBanner.tsx
```diff
@@ -59,13 +61,17 @@ export function PerpetualTradingBanner({
...
+    if (!hlTicker || perpDisabled) return;
     defaultLogger.market.token.perpsBannerClick({
       tokenSymbol: tokenDetail?.symbol ?? '',
       hlTicker,
     });
     setTimeout(async () => {
       setPerpPageEnterSource(EPerpPageEnterSource.MarketBanner);
+      if (perpTabShowWeb) {
+        navigation.switchTab(ETabRoutes.WebviewPerpTrade);
+        return;
```
sidmorizon: 🟡 P2: Web Perps entry loses contract

Severity: `non-severe`

`perpTabShowWeb` sends every mapped token to the generic Webview tab without selecting `hlTicker`. The webview loads [a fixed trade URL](packages/kit/src/views/PerpTrade/pages/PageWebviewPerpTrade.tsx:53-54), so users see its previous or default contract.

packages/kit/src/views/Market/MarketDetailV2/components/PerpetualTradingBanner/PerpetualTradingBanner.tsx
```diff
@@ -59,13 +61,17 @@ export function PerpetualTradingBanner({
...
+      }
```
sidmorizon: # 🟡 P2: Perps tab routing fixed in a duplicated copy, leaving the shared hook and adjacent entries on a dead tab

## Problem

The new `perpTabShowWeb` branch is added inline in `handlePress`:

```ts
setTimeout(async () => {
  setPerpPageEnterSource(EPerpPageEnterSource.MarketBanner);
  if (perpTabShowWeb) {
    navigation.switchTab(ETabRoutes.WebviewPerpTrade);
    return;
  }
```

But `PerpetualTradingBanner.tsx:69-101` is a near-verbatim copy of the shared hook `packages/kit/src/views/Market/hooks/usePerpsNavigation.ts:17-53` — same 80 ms `setTimeout`, same `setPerpPageEnterSource` → `setPendingInitialTradeInstrument` → `switchTab` → `changeActiveAsset` → `appEventBus.emit` sequence, down to the same explanatory comment. The branch was added only to the copy; the shared hook still hardcodes the target at `usePerpsNavigation.ts:38`:

```ts
navigation.switchTab(ETabRoutes.Perp);
```

`usePerpsNavigation` has ~11 call sites, and two of them are siblings of this banner inside `MarketDetailV2`, reachable from the very page the banner sits on: `components/TokenSelector/MobileTokenSelector.tsx:50` (invoked at `:121` for `token.perpsCoin`) and `components/TokenSelector/MarketTokenSelector.tsx:151` (invoked at `:303`). Neither consults `usePerpTabConfig` — `grep` for `perpTabShowWeb`/`perpDisabled` across `MarketDetailV2/components/TokenSelector/` returns nothing, so those perps rows are not gated either.

## Impact

With server config `usePerpWeb=true`, the two perps entry points on the same market-detail screen now diverge. Tapping the banner correctly opens `WebviewPerpTrade`; tapping a perps row in the token selector still calls `switchTab(ETabRoutes.Perp)`, whose page renders nothing because `Perp.tsx:347` uses `useNativePerpFeatureGuard()`, which returns `canRender && !perpTabShowWeb` (`packages/kit/src/hooks/usePerpFeatureGuard.ts:36`) — the user lands on a blank tab. The dead-tab behaviour of `ETabRoutes.Perp` under `perpTabShowWeb` predates this PR, but the divergence between two adjacent entries does not: before this change both paths behaved identically. The duplicated 30-line navigation body also means the next change to this flow has to be made in two places.

## Recommendation

Move the branch into the shared hook rather than the copy. In `packages/kit/src/views/Market/hooks/usePerpsNavigation.ts`, call `usePerpTabConfig()` and insert before line 38:

```ts
if (perpTabShowWeb) {
  navigation.switchTab(ETabRoutes.WebviewPerpTrade);
  return;
}
```

adding `perpTabShowWeb` to the `useCallback` deps at `:55`. Then reduce `PerpetualTradingBanner.tsx:69-101` to `navigateToPerps(hlTicker)` from `usePerpsNavigation(EPerpPageEnterSource.MarketBanner)`, keeping the `perpsBannerClick` log and the new `perpDisabled` guard in `handlePress`. That fixes every call site at once and gives the separate "webview opens the default contract" issue a single place to carry the ticker through once a symbol-aware URL exists.

[Claude session ↗](https://claude.ai/code/session_01Kg9yJ1HFKoio2QBmNvMdSH)

---

packages/kit/src/views/Market/MarketDetailV2/layouts/TokenDesktopLayout.tsx
```diff
@@ -250,6 +251,7 @@ export function TokenDesktopLayout({
           width={TOKEN_DETAIL_TRADE_COLUMN_WIDTH}
           flexShrink={0}
         >
+          <PerpetualTradingBanner px="$5" py="$5" />
```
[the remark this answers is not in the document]
zy2000-star: fixed in 626970a625. The banner now respects the configured Perps tab and hides when Perps is disabled. Added native/Web routing tests.

packages/kit/src/views/Market/MarketDetailV2/hooks/StockDetailContext.tsx
```diff
@@ -100,6 +100,12 @@ export function StockDetailProvider({
       ? initialStockPreview
       : undefined;
   const [selectedTokenId, setSelectedTokenId] = useState<string>();
+  const appliedTokenRouteRef = useRef<string | undefined>(undefined);
+  const tokenRouteKey = JSON.stringify([
+    normalizedStockId,
+    initialNetworkId,
+    initialTokenAddress,
+  ]);
```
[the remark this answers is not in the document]
zy2000-star: fixed in 626970a625. Route identity now uses normalizeTokenContractAddress, preserving EVM-equivalent addresses and case-sensitive Solana identities. Added regression tests for both.

packages/kit/src/views/Market/MarketDetailV2/components/PerpetualTradingBanner/PerpetualTradingBanner.tsx
zy2000-star: fixed in 4d23d0e6f1. The shared navigation hook records the requested coin in a background-synchronized, session-only atom before opening Web Perps. The webview loads the encoded /trade/<coin> URL and remounts for each explicit request, including repeated selections of the same coin. Added tests for cold mount, existing webviews, repeated targets, encoding, and preparation failures. Verified that the xyz:AAPL deep link displays the AAPL contract and chart.

packages/kit/src/views/Market/MarketDetailV2/components/PerpetualTradingBanner/PerpetualTradingBanner.tsx
zy2000-star: fixed in 4d23d0e6f1. Moved the Perps configuration and target preparation into usePerpsNavigation so all existing Market callers select the configured native or Web tab. The banner now delegates to usePerpsNavigation(MarketBanner) while preserving its click analytics and disabled guard. Native initial selection and fallback behavior remain covered by regression tests; Web targets are prepared before navigation.

packages/kit/src/views/PerpTrade/components/WebviewPerpTradeWebView.tsx
```diff
@@ -0,0 +1,34 @@
...
+  ComponentProps<typeof WebViewWithFeatures>,
+  | 'onWebViewRef'
+  | 'onDidStartLoading'
+  | 'onDidStartNavigation'
+  | 'onDidFinishLoad'
+  | 'onDidStopLoading'
+  | 'onDidFailLoad'
+>;
+
+export function WebviewPerpTradeWebView(props: IWebviewPerpTradeWebViewProps) {
```
sidmorizon, line 17: # 🔴 P1: New component is missing from the mobile module-id registry, breaking the native union build

## Problem

This increment adds a new module, `packages/kit/src/views/PerpTrade/components/WebviewPerpTradeWebView.tsx` (`git log --diff-filter=A` confirms it first appears in `4d23d0e6`), but does not add it to `apps/mobile/bundle-registry/module-id-registry.json`. The PR never touches that registry at all — `git diff <merge-base>..HEAD -- apps/mobile/bundle-registry/module-id-registry.json` is empty — while `grep -c WebviewPerpTradeWebView` against the registry returns `0`.

The registry is not optional for this directory: the module's own siblings are already registered, including the file this component was extracted from:

```
packages/kit/src/views/PerpTrade/pages/PageWebviewPerpTrade.tsx
packages/kit/src/views/PerpTrade/router/index.ts
packages/kit/src/views/PerpTrade/testIDs.ts
```

`Native startup graph budget (24.x)` on `4d23d0e6` fails on exactly this, after both graphs build successfully:

```
Main graph modules: 14675
Background graph modules: 14645
packages/kit/src/views/PerpTrade/components/WebviewPerpTradeWebView.tsx.
```

## Impact

The iOS/Android union build aborts during segment allocation, so no bundle, allocation report, or budget report is produced (the artifact upload step then warns that no files matched). This is a hard build break on the native target rather than a threshold regression, and it blocks the PR's CI. It is specific to splitting the webview into its own module — the extraction is otherwise sound, but stable module IDs are what the union build uses to assign modules to the common/main/background segments, and an unregistered module has no ID to allocate.

## Recommendation

Run the command the build itself prints and commit the result:

```
```

Commit only the `apps/mobile/bundle-registry/module-id-registry.json` delta for `WebviewPerpTradeWebView.tsx`; do not hand-edit the file or hand-assign an ID. Worth re-running the native startup graph job afterwards to confirm the budget check actually passes, since the failure above aborted before any budget was evaluated — a green allocation step is not yet evidence that the new module sits within budget.

[Claude session ↗](https://claude.ai/code/session_01Kg9yJ1HFKoio2QBmNvMdSH)

---

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
```diff
@@ -11,29 +13,41 @@ import { ETabRoutes } from '@onekeyhq/shared/src/routes';
...
+            });
+          }
+        } catch (error) {
+          if (perpTabShowWeb) {
+            defaultLogger.app.error.log(
+              `Failed to prepare web Perps target: ${String(error)}`,
+            );
+            return;
+          }
+        }
```
sidmorizon, line 58: # 🟡 P2: A chunk-load failure now makes every web-Perps entry a dead tap, reversing the hook's stated invariant

## Problem

The new catch aborts navigation on the web path:

```ts
} catch (error) {
  if (perpTabShowWeb) {
    defaultLogger.app.error.log(
      `Failed to prepare web Perps target: ${String(error)}`,
    );
    return;
  }
}
```

The `try` it belongs to opens at line 28 and wraps the dynamic import as well as the target write:

```ts
try {
  proxy = (
    await import('@onekeyhq/kit/src/background/instance/backgroundApiProxy')
  ).default;
  if (perpTabShowWeb) {
    await proxy.serviceWebviewPerp.setTradeTarget({ coin });
```

So the `return` fires not only when `setTradeTarget` rejects, but also when the `backgroundApiProxy` chunk fails to load or the bg bridge hiccups. That inverts the invariant the previous version of this hook stated explicitly at the same spot — the comment removed by this diff read: *"Both this and the import it needs stay inside the catch: losing the record only costs the first-mount restore, while a chunk that fails to load must still leave the tap opening the tab, as it did before."* The native branch still honours it, falling through the catch to `switchTab(ETabRoutes.Perp)` at line 52; only the web branch now bails.

## Impact

Under `perpTabShowWeb`, a transient chunk-load or bridge failure makes the tap do nothing at all — no navigation, no toast, no spinner, just a local log line. Because this hook is now the single entry point (which is the right call), that dead tap applies to all 11 call sites at once: the Perps banner, `MobileTokenSelector`/`MarketTokenSelector`, `MarketTokenListBase`, `MarketPerpsTokenList`, `MobileMarketNativeLists`, `MobileMarketWatchlistFlatList`, `PerpsTokenListSection`, `PopularTrading`, and the two Swap alerts. Users retry and nothing happens. That is a worse failure mode than the one this PR set out to fix: opening the tab on a stale contract at least leaves a usable screen with a visible way forward.

## Recommendation

Separate the two failure kinds in `usePerpsNavigation.ts:28-47`. Move the `await import(...)` out of the `try` that guards the abort so a chunk failure keeps the old fall-through, and keep the early `return` only for a genuine `setTradeTarget` rejection — e.g. gate it on `error instanceof OneKeyError`. Either way, do not leave the tap with no visible outcome: either still `navigation.switchTab(ETabRoutes.WebviewPerpTrade)` (matching the native fallback at line 52, landing the user on the default contract) or surface a `Toast.error` so the failure is legible. The test at `PerpetualTradingBanner.test.tsx:194-207` currently pins the dead-tap behaviour with `expect(mockSwitchTab).not.toHaveBeenCalled()` and should be updated alongside.

[Claude session ↗](https://claude.ai/code/session_01Kg9yJ1HFKoio2QBmNvMdSH)

---

packages/kit/src/views/PerpTrade/components/WebviewPerpTradeWebView.tsx
zy2000-star: fixed in 37cd84a20d. Registered WebviewPerpTradeWebView.tsx using the repository module-ID allocator (ID 11970). The registry check and commit checks pass. Verified strict lookup reproduces the CI missing-module error without this entry and resolves successfully with it. Full native CI will rerun on the pushed commit; local unionBuild is blocked by dependency symlink resolution in the isolated worktree.

packages/kit/src/views/Market/hooks/usePerpsNavigation.ts
zy2000-star: fixed in cadc2a8594. Web Perps preparation failures now show a localized retry toast for both proxy-loading and background target-write errors. Navigation remains gated on successfully preparing the requested coin, avoiding silently opening a stale contract. Regression tests cover both failure paths and a successful second click after recovery; all 11 tests and commit checks pass.

originalix (approved):

PR #13385, opened 2026-09-14
fix: align App Clip market locale and detail header

huhuanming:
## Problems

- The App Clip sent the raw device locale identifier, so Simplified Chinese devices could send a value the market API did not recognize and receive English category names inside a Chinese UI.
- The asset logo sat beside the circular back button in market detail, producing two adjacent circular elements.
- The temporary production-to-test stock fallback added backend resolution and concurrency state that is no longer needed.

### Production stock backend

- Production App Clip stock list, pagination, and chart requests now use the configured production base URL directly.
- Removed the retry to `utility.onekeytest.com`, the 404-specific error path, and all fallback host resolution state.
- Existing explicit `app.onekeytest.com` invocations remain isolated as the pre-existing test environment; production requests never switch to it.

### Market localization

- Derive the request language from `Bundle.main.preferredLocalizations`, which follows the App Clip language selected by iOS.
- Send `zh-cn` for Simplified Chinese and `en-us` otherwise, matching the locale values supported by the market API.
- This keeps server-provided category names consistent with the localized App Clip UI.

### Detail header layout

- Keep the back button by itself in the navigation row.
- Move the asset logo into the title row before the asset name.
- Render the logo at 32 x 32 points with a 10-point gap before the title.

## Validation

- Debug and Release iOS Simulator builds succeeded for the `OneKeyAppClip` scheme.
- Launched `https://app.onekey.so/clip/market` on the dedicated iPhone 16 Pro / iOS 18.6 Simulator.
- Verified the unavailable state appears for the production stock endpoint without loading test data.
- Verified the perpetuals list and detail chart still load, with the asset icon rendered before the title.
- Verified Simplified Chinese category localization in the running App Clip.
- `git diff --check` passed.

## Known limits

- The production stock endpoint is currently unavailable, so the stock tab intentionally shows its unavailable state until the backend is restored.
- The direct Simulator run validates the App Clip UI and API behavior, not the system App Clip card or the production AASA association.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-57fa1a3a-8eb2-5d02-859e-207587a65084)

sidmorizon:
Claude review session: https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg

originalix:

Review of commit `aa4f4997b68b` finished, but some comments could not be published automatically.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -295,6 +295,32 @@ actor AppClipMarketService {
...
+    } catch {
+      guard let fallbackBaseURL = Self.stockFallbackBaseURL(for: baseURL) else {
+        throw error
+      }
+      return try await fetchStockPage(
+        baseURL: fallbackBaseURL,
+        category: category,
+        cursor: cursor,
+        limit: limit
+      )
```
sidmorizon: 🟡 P2: Mixed-host stock pagination

Severity: `non-severe`

When production availability changes between pages, `fetchStocks` mixes production and test cursors. Their datasets differ, so pagination can skip stocks or end early.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -295,6 +295,32 @@ actor AppClipMarketService {
...
+    }
```
sidmorizon: # 🟠 [P1] Fallback host is re-decided on every call, so a cursor chain and the 60s refresh can straddle two datasets

## Problem

The fallback is resolved independently inside each call and is never recorded anywhere:

```swift
// AppClipMarketService.swift:306-315
} catch {
  guard let fallbackBaseURL = Self.stockFallbackBaseURL(for: baseURL) else {
    throw error
  }
  return try await fetchStockPage(
    baseURL: fallbackBaseURL,   // :311
    category: category,
    cursor: cursor,             // :313 — cursor was minted by whichever host answered the previous page
    limit: limit
  )
}
```

Callers always pass the production origin and never learn which host actually answered. `AppClipModel.loadMoreStocks` captures `let requestBaseURL = apiBaseURL` (`AppClipModel.swift:393`, always `https://utility.onekeycn.com` per `:182`) and re-sends it at `:401` together with `cursor` from `state.nextCursor` (`:385`, set at `:415`). Nothing pins the host to the cursor chain, the category, or the session. `refreshStocks` has the same shape: `:581` re-reads `apiBaseURL` and `:591-595` re-fetches the whole list with `limit = max(20, state.items.count)`, so the 60s `refreshTask` (`:205-226`) independently re-rolls the host on every tick.

This does not require prod to recover. A single 10s timeout (`AppClipMarketService.swift:606`) on page 1 is enough: page 1 is then served by `utility.onekeytest.com`, and page 2 goes to prod first — carrying a test-issued cursor. Either the opaque cursor is rejected (load-more fails while a valid `nextCursor` is still displayed) or both environments share a cursor codec and it parses against the *other* dataset. The `knownIds.insert($0.id).inserted` dedup at `:414` suppresses duplicate IDs but cannot recover skipped or overlapping rows.

Related, same root cause: `fetchConfiguration` (`AppClipMarketService.swift:265`) has no fallback, so `stockCategories` always come from prod while rows may come from the test host; a prod-only category ID then sends a filter the test host may not recognise.

## Impact

Users of a shipped production App Clip can see stock lists assembled from two different environments: silently missing or overlapping rows on pagination, and the whole visible list flipping between datasets every 60 seconds, so prices and constituents jump. Category filters can dead-end on an empty list.

## Recommendation

Make the fallback environment-sticky instead of per-request. Store the resolved stock origin on the `AppClipMarketService` actor (e.g. `private var resolvedStockBaseURL: [URL: URL]`, keyed by the requested origin and cleared where `AppClipModel.invalidateMarketRequests()` runs at `AppClipModel.swift:262`) and consult it in `fetchStocks` (`AppClipMarketService.swift:299`) and `fetchStockCandles` (`:487`) *before* issuing the first request, so a cursor chain and its follow-up pages always hit the same host.

A smaller alternative: return the answering origin on `AppClipMarketStockPage` and have `loadMoreStocks` (`AppClipModel.swift:400-404`) pass it back for page N+1 instead of `apiBaseURL`. Either way, extend the same fallback to `fetchConfiguration` (`AppClipMarketService.swift:265`) so categories and rows resolve to one environment.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -295,6 +295,32 @@ actor AppClipMarketService {
...
     limit: Int = 20
+  ) async throws -> AppClipMarketStockPage {
+    do {
+      return try await fetchStockPage(
+        baseURL: baseURL,
+        category: category,
+        cursor: cursor,
+        limit: limit
+      )
+    } catch {
```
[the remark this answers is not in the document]
huhuanming: Fixed in 456412a116. The fallback is now limited to the production stock endpoint returning business code 404, HTTP 404/5xx, or selected transient URLError cases (timeout, DNS/host/connect, connection lost, resource unavailable, and bad server response). Cancellation, decoding failures, missing payloads, non-404 business errors, and other client/validation errors are rethrown. The first successful stock backend is also pinned for the App Clip process so pagination, refreshes, and charts stay on the same host.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
sidmorizon: # 🟠 [P1] Partially fixed in `456412a` — the pin is still established with a check-then-act across `await`

Re-checked against the full file at `456412a`. The narrowing of the trigger set is correct and the per-call host re-rolling is genuinely gone, so the 60s-refresh flipping I described is fixed. Two parts of this root cause survive, so leaving the thread open.

## Problem

**1. The write is not single-flight or write-once.** The pin is read at `:303`, but written only after a suspension point, unconditionally:

```swift
if let resolvedBaseURL = resolvedStockBaseURLs[baseURL] {   // :303 read
  ...
}
do {
  let page = try await fetchStockPage(baseURL: baseURL, ...)  // :312 suspends — actor is re-enterable here
  resolvedStockBaseURLs[baseURL] = baseURL                    // :318 unconditional write
} catch {
  ...
  resolvedStockBaseURLs[baseURL] = fallbackBaseURL            // :333 unconditional write
```

Actor isolation holds only *between* suspension points, so two callers arriving before any pin exists both read `nil`, both resolve independently, and the later one overwrites — last-writer-wins. Concurrent stock fetches are reachable: `refreshAllMarketPages` launches `refreshStocks` via `async let` (`AppClipModel.swift:526`), and `selectStockCategory` starts another through `Task { await refreshStocks() }` (`:349`) while that one is still awaiting; `refreshStocks(force:)` bypasses the `!state.isLoading` guard at `:576` entirely.

Interleaving: request A succeeds on prod and pins prod; overlapping request B times out on prod, falls back, succeeds on test, and overwrites the pin with test at `:333`. A's rows and its `nextCursor` (`AppClipModel.swift:605`) came from prod, but `loadMoreStocks` (`:400-404`) now takes the pinned branch at `:303` and sends that prod-minted cursor to `utility.onekeytest.com` — the original cross-host cursor, one level down.

**2. `fetchConfiguration` still resolves independently.** `:269-295` neither consults `resolvedStockBaseURLs` nor falls back, and `refreshConfiguration` passes raw `apiBaseURL` (`AppClipModel.swift:538`, `:547`). In this PR's own target state — prod stocks 404, rows served from test — `stockCategories` still come from prod, and those ids are sent to test at `AppClipModel.swift:591-595`. An id test does not recognise now fails *permanently*, because the pinned branch at `:303` has no fallback left to recover with.

## Impact

An opaque cursor interpreted by a backend that did not mint it: duplicate rows, silently truncated pagination, or a dead "load more". The pin can also name a host that served none of the currently visible rows, so the next refresh swaps the whole list's dataset. The category asymmetry can dead-end a tab with no recovery path.

## Recommendation

Make first-time resolution single-flight: hold `private var stockBaseURLResolutions: [URL: Task<URL, Error>]` beside `:267`, and have `fetchStocks`/`fetchStockCandles` await the existing task for that key or create it, so concurrent callers share one resolution and one write. Minimum viable fix: gate `:318` and `:333` on `resolvedStockBaseURLs[baseURL] == nil`, which removes last-writer-wins. Separately, route `fetchConfiguration` (`:269`) through the same resolved origin so categories and rows always come from one environment.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -592,6 +667,53 @@ actor AppClipMarketService {
...
+    }
+    switch urlError.code {
+    case .timedOut,
+         .cannotFindHost,
+         .cannotConnectToHost,
+         .dnsLookupFailed,
+         .networkConnectionLost,
+         .resourceUnavailable,
+         .badServerResponse:
+      return true
```
sidmorizon: # 🔴 [P0] One transient device-network error permanently pins a production session to the test backend

## Problem

The trigger set mixes "the prod stock endpoint is absent" with ordinary client-side connectivity failures:

```swift
// AppClipMarketService.swift:703-711
switch urlError.code {
case .timedOut,
     .cannotFindHost,
     .cannotConnectToHost,
     .dnsLookupFailed,
     .networkConnectionLost,
     .resourceUnavailable,
     .badServerResponse:
  return true
```

`.timedOut` (against `request.timeoutInterval = 10`, `:635`), `.networkConnectionLost`, `.dnsLookupFailed`, `.cannotFindHost` and `.cannotConnectToHost` describe the *device's* network, not `utility.onekeycn.com`. A cell handover, a lift, or a Wi-Fi→LTE switch during the first stock request produces exactly these. The retry then reaches `utility.onekeytest.com` a second later, when connectivity is back, and succeeds — and this commit makes that accident permanent:

```swift
resolvedStockBaseURLs[baseURL] = fallbackBaseURL   // :333
```

Nothing ever clears or re-evaluates that entry — the only references are `:267`, `:303`, `:318`, `:333`, `:505`, and `AppClipModel.invalidateMarketRequests()` (`AppClipModel.swift:819-844`) rotates `environmentID` and clears the model's request-ID maps but cannot reach state owned by the `AppClipMarketService` actor (`AppClipModel.swift:185`). Its lifetime is the whole App Clip process. Every subsequent list refresh, `loadMoreStocks` page, and chart then takes the pinned branch at `:303`/`:505`, which has no `catch` and therefore no way back.

## Impact

A production user whose network hiccups once during launch spends the entire App Clip session reading stock prices, 24h changes, volumes and charts from the **test** environment, presented as production data, with no visual difference and no recovery short of killing the app. This is a regression in blast radius introduced by this commit: before `456412a` the fallback was decided per call, so the next 60s tick (`AppClipModel.swift:214-225`) returned to prod on its own. Wrong market figures rendered as authoritative in a financial surface is why this is rated above the pagination issues.

## Recommendation

Separate "fall back" from "pin". Pin only on evidence the prod stock endpoint is genuinely absent — business `code == 404` (`:692-693`) and HTTP 404 (`:694-695`) — and for the transient `URLError` cases either drop them from `shouldFallbackStockRequest` (a plain retry against prod is the right answer to a client-side timeout) or let them fall back for that one call without writing the pin. Concretely: have `shouldFallbackStockRequest` at `:689` return a two-state result (e.g. `.fallbackAndPin` / `.fallbackOnce`) and guard the writes at `:318`/`:333` on the former. If the pin is kept for transient causes, it needs an expiry or a reset hook reachable from `invalidateMarketRequests`.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -467,7 +500,7 @@ actor AppClipMarketService {
         withAllowedCharacters: Self.pathSegmentAllowed
       ),
       var components = URLComponents(
-        url: baseURL.appendingPathComponent("utility/v1/stocks"),
+        url: requestBaseURL.appendingPathComponent("utility/v1/stocks"),
         resolvingAgainstBaseURL: false
       )
```
sidmorizon, line 472: # 🟡 [P2] The chart path reads the pin but never writes it, so a chart fallback is re-paid on every request

## Problem

The pin is read here at `:505`, but unlike `fetchStocks` (which writes at `:318` and `:333`) neither branch of the public `fetchStockCandles` records what it resolved:

```swift
// AppClipMarketService.swift:512-530
do {
  return try await fetchStockCandles(stockID: stockID, period: period, requestBaseURL: baseURL)
} catch {
  guard
    Self.shouldFallbackStockRequest(after: error),
    let fallbackBaseURL = Self.stockFallbackBaseURL(for: baseURL)
  else {
    throw error
  }
  return try await fetchStockCandles(stockID: stockID, period: period, requestBaseURL: fallbackBaseURL)
}
```

So this lookup is only ever populated by the list. In the partial-recovery window this temporary code exists to cover — prod serving `utility/v1/stocks` again while `utility/v1/stocks/{id}/chart` is still 404 — the list pins prod, and every chart request then takes the prod-first path and fails before retrying test: once on detail open, and again on each of the six interval taps (`:539`, `["1h","1d","1w","1m","1y","all"]`), each driven through `AppClipModel.loadDetailCandles` (`AppClipModel.swift:1106-1110`).

## Impact

Every chart interaction pays a wasted round trip for the whole session, with `isLoadingCandles` held across both attempts. With a 404 that is a fast extra hop; if the prod chart endpoint is failing by timeout instead, it is up to 10 extra seconds of spinner (`:635`) per interval tap, repeated indefinitely. It also means the commit's "charts stay on the same host" property holds only because the list happens to resolve first, not by construction.

## Recommendation

Give the chart endpoint its own memo rather than overloading the list pin — promoting a chart fallback into `resolvedStockBaseURLs` would wrongly move the list too. Add `private var resolvedStockChartBaseURLs: [URL: URL] = [:]` beside `:267`, seed its lookup from `resolvedStockBaseURLs` when absent, and record the outcome at both `:513` and `:525`, applying the same pin/no-pin split as the `shouldFallbackStockRequest` comment above so a transient timeout does not stick.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

huhuanming: Fixed in 5ff0baf8bb. Stock charts now have an independent memo keyed by the requested production base URL and stock ID. The first real chart request with the caller-provided stock and period runs through a shared single-flight task, falls back only on an explicit business or HTTP 404, and writes the host that served the result to resolvedStockChartBaseURLs. Later interval requests for that stock use the recorded chart host directly; list pagination keeps its own category-scoped host memo.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
sidmorizon: # 🟠 [P1] Concurrency half is fixed at `5ff0baf8`; the `fetchConfiguration` half is still open

Re-verified against the complete file at `5ff0baf8`. The check-then-act race is genuinely gone — resolution is single-flight per key with owner/waiter `id` guards and cleanup on both paths, and a continuation cursor can now only ride the host that already served its first page. Scoping this thread down to the remaining half.

## Problem

`fetchConfiguration` (`:274-300`) is unchanged across all four follow-up commits — no key, no resolution, no fallback:

```swift
func fetchConfiguration(baseURL: URL) async throws -> AppClipMarketConfiguration {
  var components = URLComponents(
    url: baseURL.appendingPathComponent("utility/v2/market/basic-config"),
    resolvingAgainstBaseURL: false
  )
```

So `stockCategories` are always minted by **production** (`AppClipModel.swift:547` → `:556`), while the rows for those categories may be served by the test host. The production-minted id is then sent straight back as the `category` query param to whichever host the rows landed on — `AppClipModel.swift:593` (`category: categoryId == "all" ? nil : categoryId`) into `fetchStockPage` at `:372`.

The narrowed classifier makes the consequence worse than it was at `456412a`. `shouldFallbackStockRequest` (`:809-821`) is now 404-only, which is right in general, but if the test host does not recognise a production-minted category it 404s, the fallback is attempted, prod 404s too, and `fetchStockPageResolvingBaseURL` (`:410-424`) rethrows. Nothing is pinned and nothing is cached, so the tab is simply dead — and since there is no reset path for either resolved-host map, retrying does not help.

## Impact

A category tab that exists in production's `basic-config` but not in the backend actually serving rows renders empty on every attempt for the life of the process, with no recovery and no distinguishing error surfaced to the user.

## Recommendation

Route `fetchConfiguration` (`:274`) through the same resolution the rows use, so categories and rows always come from one backend — with the per-backend pin suggested in my other comment on this commit, that is a single lookup rather than a third memo. If configuration must stay on production, then gate on it: drop or disable any category id the serving host rejects, rather than leaving a tab that can never load.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -456,6 +553,59 @@ actor AppClipMarketService {
     stockID: String,
     period: String,
     baseURL: URL
+  ) async throws -> AppClipKlineResult {
+    let key = StockChartBaseURLKey(baseURL: baseURL, stockID: stockID)
```
sidmorizon: # 🟠 [P1] Resolution is scoped per key, not per backend — list, chart and each category tab can pin different hosts

## Problem

The single process-wide pin was replaced by two mutually independent memos with mutually independent keys:

```swift
// :267, :270
private var resolvedStockListBaseURLs: [StockListBaseURLKey: URL] = [:]
private var resolvedStockChartBaseURLs: [StockChartBaseURLKey: URL] = [:]

// :789-797
private struct StockListBaseURLKey: Hashable { let baseURL: URL; let category: String? }
private struct StockChartBaseURLKey: Hashable { let baseURL: URL; let stockID: String }
```

Each key resolves from scratch against the caller-supplied **production** `baseURL`, and nothing propagates a resolution across keys — the only write sites in the repo are `:339` (list) and `:586` (chart). Two reachable divergences follow:

**1. List vs. chart.** `AppClipModel.swift:1106-1110` passes the raw `apiBaseURL` to `fetchStockCandles` regardless of which host served the row, and `:557` keys the chart solely on that production URL. So a row rendered from the test host still sends its chart request to production first. Fallback fires only on 404 (`:809-821`), so if production answers `/utility/v1/stocks/{id}/chart` with HTTP 200 + envelope `code: 0` + empty `points` for an id it does not know — a common REST shape, and one the classifier cannot see because nothing is thrown — `:586` pins **production** for that stock permanently. `AppClipModel.swift:1132` then sets `candleLoadFailed = true`, and every later interval switch or retry reuses the pinned host. No reset exists for either map (no `removeAll`, no `removeValue`, no TTL; `invalidateMarketRequests()` at `AppClipModel.swift:819-844` cannot reach actor state), so that chart is dead for the rest of the process.

**2. Tab vs. tab.** `selectStockCategory` (`AppClipModel.swift:340-350`) fires an independent `refreshStocks` per category and the key includes `category`, so `"tech"` answering 200 on production while `"energy"` 404s leaves the two tabs on different backends in the same session, with no UI signal.

## Impact

Test-environment prices and volumes rendered alongside production ones in a single market view — a header price from one backend with a chart from the other, or two tabs on two datasets. This is the mixed-dataset failure mode from earlier rounds, reintroduced through key granularity rather than through error classification. Secondary cost: in this PR's own target scenario the "production has no stock endpoints" fact is re-learned per key, so every new category tab and every stock detail opened pays a fresh wasted production 404 round trip.

## Recommendation

Resolve once per backend, not per key: keep the real-request resolution this commit introduced (`fetchStockPageResolvingBaseURL` `:396`, `fetchStockCandlesResolvingBaseURL` `:640`) but have both read and write one `resolvedStockBaseURLs[URL: URL]` keyed only on the production origin. Minimum fix: at `:557` seed the chart key from any existing list pin for the same `baseURL` before falling back to the caller's, and at `:309` seed a new category's key from any already-resolved category for the same `baseURL`.

[Claude session ↗](https://claude.ai/code/session_01T2npWs1SKCja1Cv33k6Kyg)

---

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -456,6 +553,59 @@ actor AppClipMarketService {
...
+        stockID: stockID,
+        period: period,
+        requestBaseURL: resolvedBaseURL
+      )
+    }
+    let resolution: StockChartBaseURLResolution
+    let ownsResolution: Bool
+    if let pendingResolution = pendingStockChartBaseURLResolutions[key] {
+      resolution = pendingResolution
+      ownsResolution = false
```
sidmorizon: 🟡 P2: Concurrent stock periods share failures

Severity: `non-severe`

When two periods overlap, `pendingStockChartBaseURLResolutions` makes the later request await the earlier period. An earlier failure prevents the later period's request and falsely marks it unavailable.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
```diff
@@ -295,6 +295,32 @@ actor AppClipMarketService {
...
+      )
+    } catch {
+      guard let fallbackBaseURL = Self.stockFallbackBaseURL(for: baseURL) else {
+        throw error
+      }
+      return try await fetchStockPage(
+        baseURL: fallbackBaseURL,
+        category: category,
+        cursor: cursor,
+        limit: limit
```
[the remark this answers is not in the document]
huhuanming: Resolved by removing the production-to-test stock fallback entirely in 94180a9155. fetchStocks now sends both the first page and every continuation request directly to the caller-supplied base URL, so a production invocation cannot create or reuse a cursor from the test backend.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
huhuanming: Resolved by removing all fallback host selection in 94180a9155. Stock pagination now uses one configured base URL for the whole request chain; production pages and cursors can no longer be mixed with test-host data.

apps/mobile/ios/OneKeyAppClip/AppClipMarketService.swift
huhuanming: Resolved by deleting the shared stock chart resolution tasks and maps in 94180a9155. Each period request now runs independently against the configured base URL, so a failure from one period is no longer propagated to another concurrent period.

originalix (approved):

PR #13388, opened 2026-09-14
fix: market banner and tab UI (OK-63137, OK-63138, OK-63139, OK-63142, OK-63146)

zy2000-star:
[OK-63137](<https://onekeyhq.atlassian.net/browse/OK-63137>) [OK-63138](<https://onekeyhq.atlassian.net/browse/OK-63138>) [OK-63139](<https://onekeyhq.atlassian.net/browse/OK-63139>) [OK-63142](<https://onekeyhq.atlassian.net/browse/OK-63142>) [OK-63146](<https://onekeyhq.atlassian.net/browse/OK-63146>)

----------

## Summary

Fix market banner refresh behavior, Android spacing, banner detail columns, mixed-list navigation, and native secondary tab styling.

### OK-63137 — Banner icons flicker during refresh

Banner previews and the stock quote endpoint can return different artwork for the same symbol. Refreshing quotes previously replaced the preview artwork, then the next banner refresh switched it back.

- Preserve the banner preview logo when hydrating quotes for the same symbol.
- Continue updating prices and percentage changes from the stock endpoint.
- Fall back to the stock endpoint logo when no matching preview logo exists.

### OK-63138 — Uneven vertical spacing in Android banners

Banner content used scaled spacing while preview minimum heights and the reserved header height used fixed values. Android could also stretch cards vertically inside the horizontal scroll container.

- Apply the existing scale utility to preview, skeleton, and modern banner header heights.
- Center cards vertically in the Android banner container.

### OK-63139 — Stock banner detail columns differ from the stock homepage

Stock banner details previously displayed token statistics such as transactions, unique traders, and holders.

- Use stock metadata columns and hide token-specific statistics.
- Display price, 24h change, market cap, 24h volume, and P/E when available.
- Recognize stock-heavy token lists even when the banner API type is `ticker`.
- Reuse the existing stock-list detection rule so individual rows without stock metadata do not change the whole table’s column layout.
- Hide the P/E column when the entire list has no valid P/E values; retain `--` for missing values when other rows have data.

### OK-63142 — Mixed banner details should use Spot / Perps tabs

Mixed banners previously rendered the spot and perpetual lists together.

- Add Spot / Perps tabs and display only the selected list.
- Default to Spot and reset the selection when opening a different banner.
- Support both `mixed` and `stock_perps` banner types.
- Preserve the existing single-list layout for standalone spot and perpetual banners.
- Expose the selected tab state through accessibility attributes.

### OK-63146 — Native secondary tab corner radius differs from Watchlist

The native category buttons used a fixed corner radius of 10, while the React watchlist filters used a pill shape.

- Set the iOS and Android corner radius to half the actual button height.
- Regenerate the `react-native-pager-view@3.0.129` patch while preserving existing fixes.

### Additional UI cleanup

- Remove the end-of-list separator from banner detail lists on mobile and wide layouts.

## Validation

- 88 targeted tests passed across banner hydration, banner rendering, detail tabs, column selection, and mobile layout utilities.
- Verified the native dependency patch applies to pristine `3.0.129` sources and contains no Android build artifacts.
- Rebuilt and launched the iOS app with the native patch.
- Confirmed stable banner artwork across 496 sampled frames from a 49.7-second iOS recording.
- Verified iOS Spot / Perps switching, list content, separator removal, and stock category pill styling.
- Verified Web stock detail columns with real tokenized-stock data and Spot / Perps selection behavior.

## Remaining verification

- Android spacing and native pill styling still require device verification.
- The Web SpaceX spot list returned no data during verification; its tab and column layout were checked, while populated stock rows were verified through the technology banner and iOS SpaceX detail.

[OK-63137]: https://onekeyhq.atlassian.net/browse/OK-63137?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63138]: https://onekeyhq.atlassian.net/browse/OK-63138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63139]: https://onekeyhq.atlassian.net/browse/OK-63139?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63142]: https://onekeyhq.atlassian.net/browse/OK-63142?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63146]: https://onekeyhq.atlassian.net/browse/OK-63146?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1368786e-bfa1-581a-837d-6b2083d12baf)

sidmorizon:
Claude review session: https://claude.ai/code/session_01YVUovH2KByzzRJ2e2ewtnw

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
```diff
@@ -99,6 +101,18 @@ function MarketBannerDetailContent({ title }: { title: string }) {
...
+    isStock || shouldShowStockSubtitleForTokens(listResult.data);
+  const hiddenDesktopColumns = useMemo(() => {
+    if (!useStockColumns) return BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS;
+    const hasPeRatio = listResult.data.some(
+      (item) => getStockPeRatioValue(item) !== undefined,
+    );
+    // The turnover column renders P/E in stock metadata mode.
+    return hasPeRatio
+      ? COMPACT_SPOT_HIDDEN_DESKTOP_COLUMNS
+      : ([...COMPACT_SPOT_HIDDEN_DESKTOP_COLUMNS, 'turnover'] as const);
```
sidmorizon, line 116: 🟡 P2: Valid stock P/E column stays hidden

Severity: `non-severe`

Stock responses containing P/E still make `hasPeRatio` false. [mapStockBannerItemToToken](packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts:36-58) drops `peRatio`, so the table removes the P/E column.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
```diff
@@ -99,6 +101,18 @@ function MarketBannerDetailContent({ title }: { title: string }) {
     mobileData,
     tickerIsLoading,
   } = useMarketBannerDetail({ tokenListId, isPerps, isStock, isIndex });
+  const useStockColumns =
+    isStock || shouldShowStockSubtitleForTokens(listResult.data);
+  const hiddenDesktopColumns = useMemo(() => {
+    if (!useStockColumns) return BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS;
+    const hasPeRatio = listResult.data.some(
+      (item) => getStockPeRatioValue(item) !== undefined,
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 0a638ceb2b. The stock API mapper now preserves peRatio in stock.tradingActivity. Regression tests cover positive, zero, negative, and missing values; all 14 related tests and commit checks passed.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
zy2000-star: Fixed in 0a638ceb2b. The stock API mapper now preserves peRatio in stock.tradingActivity. Regression tests cover positive, zero, negative, and missing values; all 14 related tests and commit checks passed.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
```diff
@@ -248,12 +250,13 @@ function MarketBannerDetailContent({ title }: { title: string }) {
         clientSort
         watchlistFrom={EWatchlistFrom.BannerList}
         copyFrom={ECopyFrom.BannerList}
-        showEndReachedIndicator
         change24hColumnTitle={change24hColumnTitle}
-        hiddenDesktopColumns={BANNER_DETAIL_HIDDEN_DESKTOP_COLUMNS}
+        showStockSubtitle={useStockColumns ? 'auto' : true}
+        forceStockMetadataColumns={useStockColumns}
```
sidmorizon, line 257: # 🟠 P1 Majority-stock lists blank out real metrics on non-stock rows

## Problem

`MarketBannerDetail.tsx:104-105` derives the column mode from a **majority** predicate, then feeds it to a **whole-table** switch at `MarketBannerDetail.tsx:255`:

```ts
const useStockColumns =
  isStock || shouldShowStockSubtitleForTokens(listResult.data);
...
forceStockMetadataColumns={useStockColumns}
```

`shouldShowStockSubtitleForTokens` returns true at **≥ 50%** stock rows (`MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers.ts:116-125`: `stockCount >= items.length / 2`). That is the right rule for `showStockSubtitle`, which is applied per row, but `forceStockMetadataColumns` is not per row — it short-circuits `shouldUseStockMetadataColumnsForTokens` (`tokenListHelpers.ts:127-142`), whose own auto-detection deliberately requires `items.every((item) => !!item.stock)`.

Once that switch is on, the three affected desktop renderers read **only** the stock accessors and fall back to `--`, never to the generic value already in `text`:

- `useMarketTokenColumns/useColumnsDesktop.tsx:391-393` — `getStockMarketCapValue(record) ?? EMPTY_MARKET_VALUE`
- `useMarketTokenColumns/useColumnsDesktop.tsx:423-425` — `getStockVolume24hValue(record) ?? EMPTY_MARKET_VALUE` (column also retitled from "Liquidity" to "24h volume")
- `useMarketTokenColumns/useColumnsDesktop.tsx:460-462` — `getStockPeRatioValue(record) ?? EMPTY_MARKET_VALUE`

Those accessors return `undefined` whenever `record.stock` is absent, while `transformApiItemToToken` has already populated the generic fields for the very same rows (`tokenListHelpers.ts:366-371`: `marketCap`, `liquidity`, `turnover`, with `stock: item.stock` merely passed through at line 385).

This PR is the first production caller to pass `forceStockMetadataColumns={true}`, so the partially-stock path was previously unreachable. It is reachable now, and the PR's own test encodes exactly that shape — `MarketBannerDetail.test.tsx:122-128` builds `[{stock:{…}}, {stock:{}}, {}]` (2 of 3) and asserts `data-stock-columns === 'true'`.

## Impact

On a `ticker` banner where, say, 6 of 10 rows are tokenized stocks, the 4 crypto rows render `--` for Market Cap, 24h Volume and P/E on web/desktop even though real values exist on the record. Those rows also lose their liquidity figure, since the `liquidity` column is no longer hidden in this mode but now renders stock volume. The result is silent data loss in a table that previously displayed the numbers correctly, and it worsens as the stock share of a mixed list crosses 50%.

## Recommendation

Pick one of:

1. Make the three renderers fall back per row instead of per table, e.g. `getStockMarketCapValue(record) ?? getDefaultMarketValue(text)` at `useColumnsDesktop.tsx:392`, and the same at `:424` and `:461`. Stock rows keep stock metadata; non-stock rows keep their generic metrics. The column titles would still need a per-table decision, so this only fully works if "24h volume" is acceptable for both.
2. Tighten the banner-detail predicate to match the unanimity rule already used for auto-detection — gate `forceStockMetadataColumns` on `listResult.data.every((item) => !!item.stock)` (or call `shouldUseStockMetadataColumnsForTokens` directly) while leaving `showStockSubtitle={useStockColumns ? 'auto' : true}` on the existing majority rule. A partially-stock list then keeps token columns, which matches the stated intent that "individual rows without stock metadata do not change the whole table's column layout".

Option 2 is the smaller change and is consistent with the PR description; option 1 preserves more data if mixed lists are expected to be common.

[Claude session ↗](https://claude.ai/code/session_01YVUovH2KByzzRJ2e2ewtnw)

---

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.networkLogo.test.tsx
```diff
@@ -8,6 +8,7 @@ import type {
   IMarketStockPublicItem,
   IMarketTokenListItem,
 } from '@onekeyhq/shared/types/marketV2';
+import { getStockPeRatioValue } from '../../MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers';
```
sidmorizon, line 12: # 🟠 P1 Missing blank line between import groups fails `lint (24.x)`

## Problem

This import is added directly against the preceding `@onekeyhq/shared` group with no separating blank line:

```ts
import type {
  IMarketStockPublicItem,
  IMarketTokenListItem,
} from '@onekeyhq/shared/types/marketV2';
import { getStockPeRatioValue } from '../../MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers';
```

A `../../` specifier belongs to the parent group, so oxlint's import-group rule requires an empty line before it. Every other group boundary in this same file is already separated that way (lines 3, 5, 7), and the sibling module `useMarketBannerDetail/index.ts:11` imports from that exact same path with the required blank line in place.

`lint (24.x)` on `0a638ceb` reports precisely one error and no warnings:

```
##[error]There should be at least one empty line between import groups
Found 0 warnings and 1 error.
```

I checked the import block of every `.ts`/`.tsx` file changed in `389d2ff4..0a638ceb`; this is the only group-boundary violation among them, which matches the single-error count.

## Impact

The required `lint (24.x)` check is red on the current head, so the PR cannot merge. The failure was introduced by `0a638ceb` ("fix: preserve stock banner price earnings ratios") — the preceding head `18fc558c` did not contain this import. Worth flagging because the follow-up comments on the two P/E threads report that commit checks passed; the CI lint job disagrees, so whatever was run locally did not cover this rule.

## Recommendation

Insert an empty line before the import at `useMarketBannerDetail/index.networkLogo.test.tsx:11`:

```ts
} from '@onekeyhq/shared/types/marketV2';

import { getStockPeRatioValue } from '../../MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers';
```

`yarn agent:check --profile commit` should reproduce this before pushing; the CI job also uploads the `ci-ai-lint-diagnostics` artifact (`lint-output.log`) if the exact location needs confirming.

[Claude session ↗](https://claude.ai/code/session_01YVUovH2KByzzRJ2e2ewtnw)

---

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
zy2000-star: Fixed in 93d2b42d0f. Ticker banners now use the existing all-stock metadata predicate instead of the subtitle majority heuristic. Mixed stock/crypto rows keep the generic metric columns. Added a regression test for this mixed list; 15 related tests passed.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.networkLogo.test.tsx
zy2000-star: Fixed in 93d2b42d0f. Added the missing import-group separator. Commit checks and an explicit type-aware oxlint run without --fix passed; the new commit triggers CI again.

originalix (approved):

PR #13393, opened 2026-09-14
fix: align earn apy chart second line window with base line OK-62946

ezailWang:
[OK-62946](<https://onekeyhq.atlassian.net/browse/OK-62946>)

----------

[OK-62946]: https://onekeyhq.atlassian.net/browse/OK-62946?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-1b793fa4-5e51-5c54-8716-c09db3f0897c)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Vu7XviQZTJ3UJMj9pRnTkf

packages/kit/src/views/Earn/pages/EarnProtocolDetails/components/ApyChart.utils.ts
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  // The window ends at the anchor when one is given. The second line is cut
+  // from the primary line's newest point, so a campaign or reward line whose
+  // data stops earlier shows its tail inside the same days as the base line
+  // instead of being shifted to its own last week (OK-62946).
+  const latestTimestamp =
+    anchorTimestamp ?? history[history.length - 1].timestamp;
+  const minTimestamp = latestTimestamp - windowMs;
+  return history.filter((item) => item.timestamp >= minTimestamp);
```
sidmorizon: 🟡 P2: Secondary window exceeds primary endpoint

Severity: `non-severe`

When secondary points exceed `anchorTimestamp`, `filterByTimeWindow` retains them because it enforces only the lower bound. The secondary line then extends beyond the primary window.

Suggested fix:

```
  return history.filter(
    (item) =>
      item.timestamp >= minTimestamp && item.timestamp <= latestTimestamp,
  );
```

sidmorizon: # 🟡 P2 Comment claims an upper bound the filter never applies

## Problem

The comment added at `ApyChart.utils.ts:87-90` states "The window **ends** at the anchor when one is given." The code at `ApyChart.utils.ts:91-94` does not do that — `anchorTimestamp` is used solely to derive `minTimestamp`, and the returned filter applies a lower bound only:

```ts
const latestTimestamp =
  anchorTimestamp ?? history[history.length - 1].timestamp;
const minTimestamp = latestTimestamp - windowMs;
return history.filter((item) => item.timestamp >= minTimestamp);
```

Any secondary point newer than the anchor is still kept. The anchor defines where the window *starts*, never where it ends.

## Impact

Not user-visible at this head — I traced both series to a single payload in `EarnProtocolDetails/index.tsx:472-511`: Pendle maps `underlyingApyHistoryData.results` twice into `impliedApyHistory` / `underlyingApyHistory` (identical timestamps), and the non-Pendle path derives `extraApyHistory` as a `.filter(item => item.extraApy !== undefined)` subset of `impliedApyHistory`. The secondary's timestamps are therefore always a subset of the primary's, so the missing upper bound is currently unreachable.

The defect is in maintainability: this comment is the only documentation of the OK-62946 invariant. If the campaign/reward line is later sourced from its own endpoint, or the base line goes stale while the campaign line keeps updating, the exact misalignment this PR fixes returns silently — while the comment asserts it cannot.

## Recommendation

Either make the code match the comment by clamping both ends when an anchor is supplied:

```ts
return history.filter(
  (item) => item.timestamp >= minTimestamp && item.timestamp <= latestTimestamp,
);
```

or, if lower-bound-only is deliberate, reword lines 87-90 to say the window *starts* relative to the anchor and drop the "ends at the anchor" claim. Either way the test at `ApyChart.utils.test.ts:41-56` pins only the lower edge, so it needs a case with a secondary point newer than the anchor to lock in whichever semantic is chosen.

[Claude session ↗](https://claude.ai/code/session_01Vu7XviQZTJ3UJMj9pRnTkf)

---

packages/kit/src/views/Earn/pages/EarnProtocolDetails/components/ApyChart.utils.ts
```diff
@@ -0,0 +1,135 @@
...
+  // week, and a point newer than the base line is left out rather than drawn
+  // past it (OK-62946). Without an anchor the upper bound is the series' own
+  // newest point, so it never removes anything.
+  const latestTimestamp =
+    anchorTimestamp ?? history[history.length - 1].timestamp;
+  const minTimestamp = latestTimestamp - windowMs;
+  return history.filter(
+    (item) =>
+      item.timestamp >= minTimestamp && item.timestamp <= latestTimestamp,
+  );
```
sidmorizon, line 100: 🟡 P2: Current day and week disappear

Severity: `non-severe`

When secondary data extends past the primary anchor, `buildChartHistory` drops the current daily or weekly bucket. `aggregateByPeriod` represents each bucket with its newest point, which can fall after that anchor. Valid secondary samples before the anchor vanish from the chart.

packages/kit/src/views/Earn/pages/EarnProtocolDetails/components/ApyChart.utils.ts
```diff
@@ -0,0 +1,135 @@
...
+    );
+  }
+
+  if (period === '1w') {
+    return filterByTimeWindow(
+      aggregateByPeriod(normalized, '1w'),
+      ONE_YEAR_MS,
+      anchorTimestamp,
+    );
+  }
```
sidmorizon, line 132: # 🟡 P2 Clamp runs after aggregation, dropping whole 1d/1w buckets

## Problem

The new upper bound at `ApyChart.utils.ts:97-100` is applied to the **aggregated** array, not the raw one. For `1d` and `1w`, `buildChartHistory` calls `aggregateByPeriod` first (`ApyChart.utils.ts:118-132`), and `aggregateByPeriod` represents each bucket by its **newest** point (`bucket.set(key, item)` over an ascending list, line 70). If that representative falls after the anchor, `item.timestamp <= latestTimestamp` discards the entire bucket — including every valid sample in it that predates the anchor.

I confirmed this by executing the module's logic at this head. Anchor `2026-09-09T12:00Z`, secondary series running hourly to `14:00Z` (2h past the anchor):

| period | points | current bucket kept? |
|---|---|---|
| `1h` | 121, last `12:00Z` | ✅ yes |
| `1d` | 5 (vs 6) | ❌ current **day** dropped |
| `1w` | 1 (vs 2) | ❌ current **week** dropped |

The `1w` case is the costly one: a single overshooting point removes a whole week of the second line.

## Impact

The comment added at `ApyChart.utils.ts:91-92` states that "a point newer than the base line is left out rather than drawn past it." That holds at `1h` only. At `1d`/`1w` the function does not leave that point out — it removes the point's entire bucket, so the second line now ends a full day or week short of the base line. That is the same class of misalignment OK-62946 set out to fix, reintroduced from the opposite direction.

The new test at `ApyChart.utils.test.ts:58-71` pins this scenario at `'1h'` only, which is precisely the one period with no aggregation step — so the suite goes green while `1d`/`1w` are wrong.

Reachability caveat, stated plainly: under the declared contract in `packages/shared/types/staking.ts:2429-2449` (`apy`, `impliedApy` and `underlyingApy` are all non-optional `string`), the secondary series is always a timestamp subset of the primary, so the anchor is never exceeded and this stays latent. It becomes live if `normalizeHistory` drops the newest primary point while keeping the secondary one — e.g. a server row carrying a valid `extraApy`/`underlyingApy` but a non-numeric `apy`/`impliedApy` — or if the second line is ever sourced from its own endpoint. Since this PR added a test and a comment asserting the overshoot case is handled, the handling should be correct at all three periods rather than one.

## Recommendation

Clamp before bucketing so the representative is chosen from points already inside the window. Apply the anchor's upper bound to the normalized series in `buildChartHistory`, then aggregate:

```ts
const windowed =
  anchorTimestamp === undefined
    ? normalized
    : normalized.filter((item) => item.timestamp <= anchorTimestamp);
```

and pass `windowed` into `aggregateByPeriod` for the `1d`/`1w` branches. The current UTC day/week then keeps its newest sample at or before the anchor instead of vanishing. Extend the `ApyChart.utils.test.ts:58-71` case to cover `'1d'` and `'1w'`, asserting the current bucket survives with its last pre-anchor point.

[Claude session ↗](https://claude.ai/code/session_01Vu7XviQZTJ3UJMj9pRnTkf)

---

originalix (approved):

huhuanming (approved):

PR #13394, opened 2026-09-14
chore: sync release/v6.5.2 to x

originalix:
Sync the four business PRs in release/v6.5.2 bundle release #4 into `x`: #13219, #13247, #13282 and #13186. This brings Robinhood/default-network and Receive search fixes, Earn display labels and source-token routing, iPad layout fixes, and Market loading recovery.

The four source PRs remain separate signed `cherry-pick -x` commits, preserving each original author, commit message, and source SHA. A separate follow-up commit completes the module registry and the AssetDetails Market modal handoff.

Conflict resolutions preserve the current `x` implementation:
- Forward the wallet asset identity through the AssetDetails Market modal into `MarketTradeButton`, and through the existing Legacy-to-V2 route for legacy tab entries. Keep metadata ownership checks, the active footer token, and existing fallback behavior.
- Keep travel-mode admission and native loading/query-scope guards. Recovery handles a fast first-page failure even when the loading render has not committed, without clearing another query's rows.
- Migrate the release patches to React Native 0.86.2, Tamagui native media driver 2.7.7, and tab-view 3.0.129. Register the three new runtime modules with the existing ID allocator while preserving every existing ID.

Release tracking, build numbers, dependency versions, and previously synced releases are excluded.

Validation: static diff and conflict review, module registry validation, scoped lint/format checks, background API contract checks, native-storage inventory checks, and TypeScript via `yarn agent:check --profile commit`. Local application builds and runtime/UI self-tests were not run for the follow-up fix.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-2cbc515f-8a71-59f9-8eb3-e8b2c7f7fad4)

sidmorizon:
Claude review session: https://claude.ai/code/session_01TZX5Yr4VGEbPSA418DXujw

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
```diff
@@ -78,6 +80,12 @@ function TokenDetailsFooter(props: {
     if (marketNavigationTarget?.type === 'detail') {
       navigation.push(EModalAssetDetailRoutes.MarketDetail, {
         token: marketNavigationTarget.token,
+        preferredToken:
+          !isAggregateToken &&
+          !networkUtils.isAllNetwork({ networkId }) &&
+          (isNative || Boolean(tokenAddress))
+            ? { networkId, tokenAddress: tokenAddress ?? '', isNative }
+            : undefined,
```
sidmorizon, line 88: 🟠 P1: Source asset trades on wrong chain

Severity: `severe`

When `preferredToken` comes from Token Details on desktop or web, the compact Market modal discards it. `MarketTradeButton` then uses the legacy heuristic, so asset actions can target another chain.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenList.ts
```diff
@@ -574,6 +574,7 @@ export function useMarketTokenList({
       // Native pages can mount before the default network is initialized.
       // usePromiseResult captures watchLoading when its runner is created.
       watchLoading: platformEnv.isNative || hasNetworkId,
+      undefinedResultIfError: true,
```
sidmorizon, line 577: # 🟡 P2 Web HTML-seed rows are wiped to "No data" when the forced remote page 1 fails

## Problem

On web the default query mounts with `initResult: marketTokenListSeedInitResult` (`useMarketTokenList.ts:581`), the table rows are seeded from it (`:596-608`), and the provisional-result effect immediately runs a forced remote `fetchMarketTokenList()` (`:679-702`), which on web goes straight to `fetchMarketTokenListRemoteLight` (`marketLightApi.ts:163-167`).

If that request rejects, the runner catch (`:490-521`) has no `latestAuthoritativeFirstPageResultRef` entry and no trusted cache entry (the seed is only used when there is none, `:399-417`), so it reaches `if (!platformEnv.isNative) throw error;`. With the new `undefinedResultIfError: true` (`:577`), `usePromiseResult` now runs `setResult(undefined)` instead of re-throwing and keeping the seed (`usePromiseResult.ts:341-353`). The new effect at `:636-645` then sees `previous.isLoading !== false && isLoading === false && apiResult === undefined` for the same query key and commits `setTransformedDataState({ data: [] })`.

Before this change the rethrow left `result` as the seed and the rows stayed rendered.

## Impact

Web only, first load with a bootstrap seed and a failed remote request: the list goes from populated seed rows to the "No data" empty component (`MarketTokenListBase.tsx:825-841`). `errorQueryKey` is native-only (`:491-498`), so no Retry affordance appears; the rows only come back on the next poll or a focus revalidation. This regresses the "preserve Market rows across loading transitions" intent of the cherry-picked commit on the one path that has no fallback.

## Recommendation

In the runner catch (`useMarketTokenList.ts:509-521`), treat the seed like the cold-cache fallback: when `currentQueryKeyRef.current === requestQueryKey` and the current result is `__fromSeed`, return that seed result (held in a ref alongside `trustedMarketTokenListFallbackRef`) instead of throwing, so `undefinedResultIfError` does not clear it. Alternatively set `errorQueryKey` on all platforms so the cleared state at least renders the Retry UI. Add a web test in `useMarketTokenList.initialData.test.tsx` that mounts with a seed `initResult`, rejects the forced remote request, and asserts the seed rows survive.

[Claude session ↗](https://claude.ai/code/session_01TZX5Yr4VGEbPSA418DXujw)

---

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
```diff
@@ -78,6 +80,12 @@ function TokenDetailsFooter(props: {
     if (marketNavigationTarget?.type === 'detail') {
       navigation.push(EModalAssetDetailRoutes.MarketDetail, {
         token: marketNavigationTarget.token,
+        preferredToken:
```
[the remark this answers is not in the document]
originalix: Fixed in 0a76006294. The AssetDetails Market modal now passes route.params.preferredToken through MarketDetailContent into MarketTradeButton, so both useMarketTradeActions and useMarketTradeNetwork receive the source asset identity on the actual desktop/web path. CoinGecko-only callers still omit the optional value. Static lint and TypeScript checks passed.

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
originalix: Fixed in 0a76006294. The AssetDetails Market modal now passes route.params.preferredToken through MarketDetailContent into MarketTradeButton, so both useMarketTradeActions and useMarketTradeNetwork receive the source asset identity on the actual desktop/web path. CoinGecko-only callers still omit the optional value. Static lint and TypeScript checks passed.

packages/kit/src/views/AssetDetails/pages/TokenDetails/TokenDetailsFooter.tsx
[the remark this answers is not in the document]
originalix: Fixed in 0a76006294. The AssetDetails Market modal now passes route.params.preferredToken through MarketDetailContent into MarketTradeButton, so both useMarketTradeActions and useMarketTradeNetwork receive the source asset identity on the actual desktop/web path. CoinGecko-only callers still omit the optional value. Static lint and TypeScript checks passed.

huhuanming (approved):

PR #13396, opened 2026-09-14
[OK-61674] Restore immediate Market row rendering

huhuanming:
[OK-61674](<https://onekeyhq.atlassian.net/browse/OK-61674>)

----------

## Summary

- restore immediate rich-row rendering once Market data is available
- reuse the existing Web direct API, SWR/seed, and watchlist batch cache paths without adding another cache layer
- keep the 6-second WebSocket startup deferral, first-render row limit, and content-visibility optimizations

## Root Cause

- `#12748` removed the fixed lightweight-row gate so cached/SWR rows rendered fully on the first frame
- the `#12839` release sync reintroduced the gate for rows after the first four
- `#13149` later accelerated Web data loading with direct APIs and request caches, but the independent 6-second render gate remained, so faster data did not improve visible row completion

## QA

- Jira: https://onekeyhq.atlassian.net/browse/OK-61674
- local production build with eight watchlist tokens
- hard refresh: direct batch API completed in 1,531 ms; all eight rows and eight rich star controls committed at 2,800 ms, about 372 ms after the response, with no extra 6-second wait
- in-page Hot -> Watchlist return: all eight full rows visible in 286 ms with zero additional batch requests

## Verification

- focused Market Jest suites: 3 passed, 33 tests passed
- `yarn agent:check --profile commit` passed
- production Rspack compilation passed; the existing Windows shell wrapper failed only at the final unavailable `cp` command after assets were emitted
- `git diff --check` passed

[OK-61674]: https://onekeyhq.atlassian.net/browse/OK-61674?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-91bb1662-1558-5c0f-8a0e-b65d4095a63d)

sidmorizon:
Claude review session: https://claude.ai/code/session_01SzBHxeLfmXcdqDEc8Mosa7

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketTokenColumns/useColumnsDesktop.tsx
```diff
@@ -135,7 +135,14 @@ function renderLightweightTokenIdentity(record: IMarketToken) {
       minWidth={0}
       overflow="hidden"
     >
-      <Stack width={40} height={40} borderRadius="$full" bg="$bgStrong" />
+      <Token
```
originalix: P2: [Deferred rows reintroduce image work during the startup deferral window]

When an integrated production web tab is still in its pre-ready period, this renderer is used for every row from index 4 onward specifically to defer extra media. Constructing `Token` here mounts its image and fallback-source machinery, so near-viewport deferred rows can begin token and network image work before the six-second defer window ends.

Please keep this branch non-loading (for example, retain a lightweight placeholder) and only render `Token` after deferred features are ready, or add an explicit no-load mode that preserves the startup boundary.

huhuanming: Addressed in adf8e46a12. The lightweight renderer is image-free by default again: non-watchlist rows retain the 40x40 placeholder and do not mount Token during the startup deferral window. Only watchlist mode explicitly opts into early token/network images for OK-61674. Added regression coverage for both the watchlist image path and the non-watchlist no-load path.

originalix (approved):

PR #13397, opened 2026-09-14
fix: add native market indicator toolbar OK-63134

limichange:
[OK-63134](<https://onekeyhq.atlassian.net/browse/OK-63134>)

----------

Native Market Detail was missing the indicator shortcut bar. Add one horizontally scrolling row with a divider between main and sub indicators and a fixed indicator-settings button on the right. Restore general chart settings in the native toolbar and remove its indicator and chart-type buttons.

Enable the new layout only for iOS/Android Market Detail. Web and other chart consumers retain their existing controls. Fullscreen settings navigation first dismisses the chart overlay so the destination page remains visible.

Reuse the shared indicator helpers, keep the toolbar settings element stable across candle updates, and register the new modules for production mobile bundles.

Validation:
- Seven targeted Jest suites: 83 tests passed.
- Strict iOS UNION build, module-ID registry verification, both startup budgets, and bundle architecture checks.
- iOS simulator: indicator selection updates the chart; the row scrolls while the settings button stays fixed; both settings entries open the correct page or dialog; entering indicator settings and chart settings from fullscreen exits the overlay and displays the destination page.
- Android device verification has not been performed.

Issue: OK-63134

[OK-63134]: https://onekeyhq.atlassian.net/browse/OK-63134?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorQuickBar.tsx
```diff
@@ -0,0 +1,315 @@
+import { type ReactNode, memo, useCallback, useRef } from 'react';
```
originalix, line 1: P1: [Register the new native quick-bar module]

When a production iOS/Android or UNION build reaches this newly added module through `TradingViewNativeContainer`, Metro uses the strict module-ID map. The registry has no entry for `packages/kit/src/components/TradingView/TradingViewChartControls/indicatorSelector/NativeIndicatorQuickBar.tsx`, so bundle generation throws before output is emitted.

Regenerate the mobile module-ID registry from the build map and commit the corresponding `apps/mobile/bundle-registry/module-id-registry.json` entry.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-21cca754-238c-501e-851c-47d6b180fab0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Skx2p1Q9Kznd4AtCLbvkG5

packages/kit/src/components/TradingView/TradingViewNative/TradingViewNativeContainer.tsx
```diff
@@ -759,6 +806,65 @@ const TradingViewNativeContent = memo(
...
+                size="small"
+                variant="tertiary"
+                icon="ChartTrending2Outline"
+                iconSize="$5"
+                width="$10"
+                height={TRADING_VIEW_NATIVE_INDICATOR_QUICK_BAR_HEIGHT}
+                accessibilityLabel={intl.formatMessage({
+                  id: ETranslations.market_indicators,
+                })}
+                onPress={() => handleIndicatorSettingsPress()}
```
sidmorizon, line 837: # 🟠 P1 Fullscreen indicator-settings button pushes a nav modal behind the fullscreen overlay

## Problem

Platform scope: iOS + Android, `main` UI runtime only (desktop/web are unaffected — `TradingViewNativePresentation.tsx` is a pass-through and both new props are `platformEnv.isNative`-gated at `MobileLayout.tsx:767-768`).

The quick bar is rendered **in fullscreen**: `TradingViewNativeContainer.tsx:1049-1051` is `{onNativeIndicatorQuickBarChange && !isNativeChartFullscreen ? null : nativeIndicatorQuickBar}`, so when the consumer hoists the bar (Market Detail does, `MobileLayout.tsx:769`), the fullscreen branch renders it inline.

Its settings button (this line) calls `handleIndicatorSettingsPress()` with **no** `indicator` argument. With `nativeControlsLayoutMode === 'mobile'` (`MobileLayout.tsx:777`), that takes the navigation branch at `TradingViewNativeContainer.tsx:696-702`:

```ts
if (nativeControlsLayoutMode !== 'desktop' && !indicator) {
  navigation.pushModal(EModalRoutes.MarketModal, {
    screen: EModalMarketRoutes.MarketIndicatorSettings, ...
  });
  return;
}
```

But in fullscreen the chart subtree is hoisted out of the navigator: `TradingViewNativePresentation.native.tsx:89` returns `null` and `updateLayer` moves the children into `TradingViewNativeFullscreenHost`, mounted at `FullWindowOverlayContainer/index.tsx:51` inside `<OverlayContainer>`. On iOS that is `react-native-screens`' `FullWindowOverlay` (`OverlayContainer.ios.tsx`); elsewhere it is a pass-through sibling rendered **after** `<RootNavigator/>` (`NavigationContainer.tsx:26-30`, `Container/index.tsx:57-62`). The fullscreen layer is `position:absolute` inset-0 with `bg="$bgApp"`, so it paints over the whole navigator either way. A pushed nav modal lands underneath it.

Note the ordering in `FullWindowOverlayContainer` is deliberate: the fullscreen host (line 51) sits *before* `FULL_WINDOW_OVERLAY_PORTAL` (line 53) where `Dialog` renders — dialogs work over fullscreen, nav modals do not.

This PR also removes the previously working fullscreen path. `TradingViewNativeChartControlsContainer.tsx:208-210` now reads `hasVisibleIndicators={!compactMobileLayout && !onChartClose && !mobileSettingsButton}`, and `mobileSettingsButton` (line 103, `layoutMode === 'mobile' && !onChartClose`) is non-null in fullscreen too — `mobileSettingsControl` at `TradingViewNativeContainer.tsx:911-923` is **not** gated on `!isNativeChartFullscreen`, unlike the in-chart placement at line 1029, and Market Detail never passes `onNativeChartClose`. Pre-PR, the fullscreen toolbar's indicator control opened `showIndicatorsDialog` (`TradingViewNativeChartControlsContainer.tsx:152-172`), which renders above the fullscreen layer.

Same root cause, second instance: the toolbar settings button is now rendered in fullscreen, and its "Chart settings" row calls `navigation.pushModal(... MarketChartSettings)` at `TradingViewNativeChartSettingsButton.tsx:43-48` after `dialog.close()`.

## Impact

In fullscreen on iOS/Android, tapping the indicator-settings icon — now the only indicator-configuration entry point in fullscreen — appears to do nothing, while a `MarketIndicatorSettings` screen is silently pushed onto the modal stack behind the overlay. Back gestures then act on that hidden screen, and exiting fullscreen reveals a page the user never opened. The same applies to "Chart settings" from the new toolbar settings dialog.

## Recommendation

Either exit fullscreen before navigating, or keep the dialog-based path in fullscreen:

- Preferred: in `handleIndicatorSettingsPress` (`TradingViewNativeContainer.tsx:694-702`), call `onNativeChartFullscreenChange?.(false)` when `isNativeChartFullscreen` is true before `navigation.pushModal`; apply the same to `openChartSettingsModal` in `TradingViewNativeChartSettingsButton.tsx:43-48` (needs an `isFullscreen`/`onExitFullscreen` prop).
- Alternative: add `&& !isNativeChartFullscreen` to `settingsControl` (`TradingViewNativeContainer.tsx:818-832`) and to `mobileSettingsControl` (lines 911-923), mirroring the guard already present at line 1029, so `hasVisibleIndicators` stays true in fullscreen and the Dialog path remains reachable.

[Claude session ↗](https://claude.ai/code/session_01Skx2p1Q9Kznd4AtCLbvkG5)

---

packages/kit/src/components/TradingView/TradingViewNative/TradingViewNativeContainer.tsx
```diff
@@ -802,6 +908,19 @@ const TradingViewNativeContent = memo(
             calendarAvailableTimeRange={calendarAvailableTimeRange}
             compactMobileLayout={isCompactDisplayMode}
             enableNativeChartSettings={enableNativeChartSettings}
+            mobileSettingsControl={
```
sidmorizon: # 🟡 P2 Inline `mobileSettingsControl` element defeats `memo` on the toolbar, re-rendering it on every k-line tick

## Problem

`TradingViewNativeChartControlsContainer` is wrapped in `memo(...)` (`TradingViewNativeChartControlsContainer.tsx:66`). Before this PR, every prop it received from `TradingViewNativeContent` on the Market Detail path was reference-stable across data-driven re-renders — `intervalConfig` and `calendarAvailableTimeRange` are `useMemo`'d inside `useTradingViewNativeKLine`, `activeIndicatorValues` is `useMemo`'d, all handlers are `useCallback`'d, and `fullscreenHeader` is never passed by `MobileLayout.tsx`.

This block constructs a **fresh React element on every render**:

```tsx
mobileSettingsControl={
  isMobileControlsLayout && enableNativeChartSettings && nativeChartSettingsInToolbar ? (
    <TradingViewNativeChartSettingsButton ... />
  ) : undefined
}
```

so `memo` can never bail out. `TradingViewNativeContent` re-renders on every chart data update — `chartPictureVersion` is bumped in hook state on each tick (`data/useTradingViewNativeKLine.ts:319, 2874, 3109, 3349, 3714, 4204`). The entire toolbar subtree (interval selector, chart-type select, calendar control, fullscreen control) therefore re-renders per tick on the mobile market-detail hot path.

The sibling `nativeIndicatorQuickBar` added by this same PR (line 809) already uses `useMemo` for exactly this reason — this element just missed the same treatment.

## Impact

Continuous avoidable re-render work on the mobile chart screen for as long as the websocket streams candles, on the startup/interaction path that `1k-perf-optimizer` guidance targets.

## Recommendation

Hoist the element into a `useMemo` alongside `nativeIndicatorQuickBar` (near `TradingViewNativeContainer.tsx:809`), keyed on `[isMobileControlsLayout, enableNativeChartSettings, nativeChartSettingsInToolbar, priceAxisWidth, enablePreviousClose, isChartSwitchDisabled, onChartSwitch]`, and pass the memoized value at line 911.

[Claude session ↗](https://claude.ai/code/session_01Skx2p1Q9Kznd4AtCLbvkG5)

---

packages/kit/src/components/TradingView/TradingViewNative/TradingViewNativeContainer.tsx
```diff
@@ -25,7 +31,17 @@ import { stableStringify } from '@onekeyhq/shared/src/utils/stringUtils';
...
+import {
+  TRADING_VIEW_NATIVE_INDICATOR_QUICK_BAR_HEIGHT,
+  TradingViewIndicatorQuickBar,
+} from '../TradingViewChartControls/indicatorSelector/NativeIndicatorQuickBar';
 import { TradingViewChartLoadingMask } from '../TradingViewChartLoadingMask';
+import {
+  canToggleTradingViewNativeIndicatorOn,
+  getAppNativeIndicators,
+  getIndicatorSections,
+} from '../TradingViewV2/components/indicatorControls/hooks/useNativeIndicatorActiveValues';
```
sidmorizon: # 🟡 P2 New `TradingViewNative → TradingViewV2` import edge picks the duplicate copy of helpers that already sit next to the new file

## Problem

Before this PR the folder dependency ran one way only — `TradingViewV2` imported from `TradingViewNative`, never the reverse (`git grep -n "TradingViewV2" <base> -- 'packages/kit/src/components/TradingView/TradingViewNative/*'` returns nothing; at this head it returns three hits).

This PR adds the reverse edge here (plus `TradingViewNative/types.ts:9`), so the native-chart module now depends on the webview-chart module for four pure helpers. Two of them already exist, functionally equivalent, in `TradingViewChartControls/indicatorSelector/indicatorUtils.ts`:

- `getIndicatorSections` — `indicatorUtils.ts:27` vs `useNativeIndicatorActiveValues.ts:73`
- `canToggleTradingViewNativeIndicatorOn` — `indicatorUtils.ts:57` vs `useNativeIndicatorActiveValues.ts:144`

(`getTradingViewNativeSubIndicatorCount` and `getNativeIndicatorSelectionUpdates` are duplicated across the two files as well; neither file re-exports the other.)

`indicatorUtils.ts` lives in the *same directory* as the `NativeIndicatorQuickBar.tsx` this PR just created and imports four lines above (line 35-38). The PR reached past the adjacent copy for the distant one, which both cements the duplication and inverts the existing layering.

## Impact

The native chart — also used by Perps — now pulls in the webview chart module tree for helpers it could get from the directory it already imports from. The two implementations can drift apart silently, and the next reader has no signal which copy is canonical.

## Recommendation

In `TradingViewNativeContainer.tsx:39-43`, import `getIndicatorSections` and `canToggleTradingViewNativeIndicatorOn` from `../TradingViewChartControls/indicatorSelector/indicatorUtils` instead. Move the 5-line `getAppNativeIndicators` (`TradingViewV2/components/indicatorControls/hooks/useNativeIndicatorActiveValues.ts:64-70`) into that same `indicatorUtils.ts`, and move `nativeIndicatorQuickBarState.ts` from `TradingViewV2/components/tradingViewV2/` into `TradingViewChartControls/indicatorSelector/`, re-exporting from `TradingViewV2/index.ts` so `MobileLayout.tsx:40-41` keeps resolving. Then have `useNativeIndicatorActiveValues.ts` re-export the shared copies rather than defining its own, which removes the duplication instead of adding a consumer to it.

[Claude session ↗](https://claude.ai/code/session_01Skx2p1Q9Kznd4AtCLbvkG5)

---

originalix (approved):

PR #13398, opened 2026-09-14
fix: align market display issues ｜OK-63153 OK-63151 OK-62760 OK-63158

zy2000-star:
[OK-63153](<https://onekeyhq.atlassian.net/browse/OK-63153>) [OK-63151](<https://onekeyhq.atlassian.net/browse/OK-63151>) [OK-62760](<https://onekeyhq.atlassian.net/browse/OK-62760>) [OK-63158](<https://onekeyhq.atlassian.net/browse/OK-63158>)

----------

## Summary

Fix multiple issues in the Market module involving mobile list actions, banner detail columns, price formatting, and financial chart rendering.

### OK-63153

- Removed favorite star controls and favorite toggle interactions from mobile stock and top coin lists.
- Updated native list rows and React Native flat lists consistently.
- Updated mobile market list tests to reflect the new behavior.

### OK-63151

- Aligned banner detail list columns with the selected market type.
- Added support for custom primary column titles.
- Stock lists now display company information and hide non-applicable market cap and volume fields.
- Non-stock lists display market cap and volume columns correctly.
- Updated token list item data mapping for the displayed columns.

### OK-62760

- Adjusted small decimal price formatting thresholds.
- Values with more than five leading zeros now use compact subscript formatting.
- Updated both TypeScript and native text formatter implementations.
- Updated formatter unit tests.

### OK-63158

- Fixed financial chart series ordering.
- Estimate data is rendered before Actual data so Actual values remain visible when the series overlap.

## Validation

- Passed `yarn agent:check --profile commit`
- Passed targeted Market and LightweightChart Jest tests.
- Passed TypeScript staged checks.
- Passed formatting and lint checks.

[OK-63153]: https://onekeyhq.atlassian.net/browse/OK-63153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63151]: https://onekeyhq.atlassian.net/browse/OK-63151?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62760]: https://onekeyhq.atlassian.net/browse/OK-62760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63158]: https://onekeyhq.atlassian.net/browse/OK-63158?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-3fa14ee2-a92a-5ab5-8090-8f9c06a56917)

sidmorizon:
Claude review session: https://claude.ai/code/session_011ZssaFHzcgMtUJbYJuwbbR

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
```diff
@@ -28,7 +28,7 @@ export function formatChartPrice(price: number, maxCharacters = 8): string {
         : `${digits.slice(0, decimalPosition)}.${digits.slice(decimalPosition)}`;
   }
   const zeroCount = -exponent - 1;
-  if (value > 0 && !unit && zeroCount >= 2 && body.length > limit) {
+  if (value > 0 && !unit && zeroCount > 5 && body.length > limit) {
```
sidmorizon: # 🟠 P1 Raising the subscript threshold to `> 5` truncates 3–5-zero prices to zero significant digits

## Problem

The old guard was `zeroCount >= 2 && body.length > limit` — the subscript form kicked in *exactly* when the plain decimal would overflow `limit`. Changing the first clause to `zeroCount > 5` (`formatChartPrice.ts:31`, mirrored at `formatChartPrice.text-js:29`) removes that protection for `zeroCount` 2–5, but the ellipsis fallback at `formatChartPrice.ts:39-42` is unchanged:

```ts
let end = Math.max(1, limit - 2);
while (/[₀₁₂₃₄₅₆₇₈₉]/.test(body.charAt(end))) end += 1;
return `${prefix}${body.slice(0, end)}...${unit}`;
```

For those values `slice(0, limit - 2)` now cuts *inside the leading-zero run*, so the label keeps no significant digits at all. I executed the real `.text-js` source with the old and new thresholds:

| value | before (`>= 2`) | after (`> 5`), limit 8 | after, limit 7 |
|---|---|---|---|
| 0.0000123456 | `$0.0₄12...` | `$0.0000...` | `$0.000...` |
| 0.0000015 | `$0.0₅15` | `$0.0000...` | `$0.000...` |
| 0.0000025 | `$0.0₅25` | `$0.0000...` | `$0.000...` |
| 0.0000035 | `$0.0₅35` | `$0.0000...` | `$0.000...` |
| 0.00000123 | `$0.0₅123` | `$0.0000...` | `$0.000...` |

Note that five *distinct* values all collapse to the identical string.

## Impact

`formatChartPrice` is the Y-axis tick formatter for the stock/index price line chart (`StockPriceLineChart.tsx:130`, and the WebView twin via `htmlTemplate.ts:18`), with `maxPriceCharacters = 7` when `chartWidth < 400` and `8` otherwise (`StockPriceLineChart.tsx:128`). For any sub-`$0.001` quote, consecutive gridlines render the same meaningless label — at `limit = 7` the entire `1e-6 … 1e-3` band prints `$0.000...` on every tick, making the axis unreadable and the scale impossible to infer. This is a strict regression: the pre-PR output kept 1–2 significant digits at every tick. The PR's own test now encodes the regression at `formatChartPrice.test.ts:9` (`expect(formatChartPrice(0.000_001_2)).toBe('$0.0000...')`).

Note `.ts` and `.text-js` are in exact parity after the change, so the twin is not the issue — the threshold is.

## Recommendation

Keep the "no subscripts for shallow decimals" intent, but restore the overflow fallback so truncation can never consume the whole mantissa. At `formatChartPrice.ts:31` and the identical `formatChartPrice.text-js:29`:

```ts
// Subscripts are the preferred compact form past 5 zeros, but any zero run the
// character limit would truncate into meaninglessness must also compress.
const plainWouldLoseAllDigits = zeroCount >= Math.max(1, limit - 2) - 2;
if (
  value > 0 &&
  !unit &&
  body.length > limit &&
  (zeroCount > 5 || plainWouldLoseAllDigits)
) {
```

Then update `formatChartPrice.test.ts:9` so `0.000_001_2` asserts a form that still shows digits (e.g. `'$0.0₅12'`), and add boundary cases for `zeroCount` 3 and 4 at both `maxCharacters` 7 and 8.

[Claude session ↗](https://claude.ai/code/session_011ZssaFHzcgMtUJbYJuwbbR)

---

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
```diff
@@ -263,45 +263,51 @@ export function FinancialChart({
...
-                    const value = row.values[index];
-                    if (!isFinancialNumber(value)) return null;
-                    if (item.kind === 'line') return null;
-                    if (item.kind === 'bar') {
-                      const barIndex = barSeries.indexOf(item);
+                  series
+                    .map((item, index) => ({ item, index }))
+                    .toSorted((a, b) => {
+                      if (a.item.kind === b.item.kind) return 0;
+                      return a.item.kind === 'estimate' ? -1 : 1;
```
sidmorizon: # 🟡 P2 `toSorted` comparator is asymmetric, so the new z-order is engine-defined rather than guaranteed

## Problem

`FinancialChart.tsx:268-271`:

```ts
.toSorted((a, b) => {
  if (a.item.kind === b.item.kind) return 0;
  return a.item.kind === 'estimate' ? -1 : 1;
})
```

`IFinancialChartSeries['kind']` is `'bar' | 'line' | 'actual' | 'estimate'` (`FinancialChart.tsx:37`). When neither side is `estimate` — e.g. `a.kind === 'bar'`, `b.kind === 'line'` — the comparator returns `1`, and it returns `1` for the swapped pair too. That breaks the antisymmetry ECMA-262 requires of a `comparefn`, so ordering of any array containing two different non-`estimate` kinds is implementation-defined.

Of the shapes `buildFinancialChart` produces (`financialChartData.ts`), `performance` is `[bar, bar, line]` (lines 161-168) and hits the inconsistent pair; `debt` is `[bar, bar, bar]`, `earnings` is `[actual, estimate]`. I confirmed the current rendering is still correct on V8, and a reorder of `performance` would be invisible anyway because `line` returns `null` at line 275 and bar x-positions come from `barSeries.indexOf(item)` (line 277) rather than the sorted position — so this is latent today, not a live visual bug.

## Impact

The intent of OK-63158 — a deterministic paint order so Actual stays visible over Estimate — is encoded as an asymmetric rule instead of a sort key, so the guarantee comes from the current data shapes rather than from the code. Any future series mixing two non-`estimate` kinds (a `line` overlay added to the earnings chart, a second marker series) puts that z-order at the mercy of the engine's sort implementation, and Hermes (native) and V8/JSC (web/desktop) may legitimately order the unspecified pairs differently — reintroducing the overlap bug on one platform only, which is expensive to diagnose.

## Recommendation

Use a rank subtraction, which is transitive by construction and lets `toSorted`'s stability preserve the original order within each rank:

```ts
.toSorted(
  (a, b) =>
    (a.item.kind === 'estimate' ? 0 : 1) - (b.item.kind === 'estimate' ? 0 : 1),
)
```

[Claude session ↗](https://claude.ai/code/session_011ZssaFHzcgMtUJbYJuwbbR)

---

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
```diff
@@ -236,6 +236,13 @@ function MarketBannerDetailContent({ title }: { title: string }) {
         <BannerDetailTokenFlatList
           data={mobileData}
           isLoading={tickerIsLoading}
+          primaryColumnTitle={
+            isStock
+              ? intl.formatMessage({ id: ETranslations.market_stock_company })
+              : `${intl.formatMessage({ id: ETranslations.global_name })} / ${intl.formatMessage({ id: ETranslations.market_mcap })}`
+          }
+          showMarketCap={!isStock}
+          showVolume={!isStock}
```
sidmorizon: 🟡 P2: Tokenized-stock banners use crypto columns

Severity: `non-severe`

Narrow ticker banners with stock-backed rows use crypto columns because `isStock` remains false. `useStockColumns` recognizes the same rows, so wide layouts display stock columns.

packages/kit/src/components/LightweightChart/utils/formatChartPrice.ts
zy2000-star: Fixed in 380c6e03f0.

packages/kit/src/views/Market/MarketDetailV2/components/StockFinancials/FinancialChart.tsx
zy2000-star: Fixed in 380c6e03f0.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx
zy2000-star: Fixed in 380c6e03f0.

packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.test.tsx
```diff
@@ -147,8 +147,10 @@ it('keeps token columns for a ticker banner containing stock and crypto rows', (
   mockIncludeNonStockToken = true;
   render(<MarketBannerDetail />);
   const list = screen.getByTestId('stocks');
-  expect(list.getAttribute('data-stock-columns')).toBe('false');
-  expect(list.getAttribute('data-hidden-columns')).toBe('liquidity');
+  expect(list.getAttribute('data-stock-columns')).toBe('true');
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 380c6e03f0.

originalix (approved):

PR #13399, opened 2026-09-14
fix: unify perps scrollbar and table styling OK-63144

Kahnchan:
[OK-63144](<https://onekeyhq.atlassian.net/browse/OK-63144>)

----------

Perps tables had inconsistent scrollbars across Web and Desktop, visible scrollbar gutters, and misaligned Trades History columns. Use scoped 4px overlay scrollbars with opaque theme colors and 2px corners; keep horizontal scrolling at the panel bottom while preserving natural row heights and fixed-column alignment.

Also align the Trades History Fee header and values, remove its fixed Close PnL column, hide token-selector scroll indicators on all platforms, fit the desktop token table without horizontal scrolling, and remove the Recent 2,000 funding label and duplicate page-bottom border. The funding chart uses the same scrollbar style. Shared scrollbar defaults and trading/data-loading logic remain unchanged.

Validation: `yarn agent:check --profile commit` and `yarn agent:check --profile pr` passed locally. Automated visual interaction tests were not run. GitHub CI will run on this PR.

Issue: [OK-63144](https://onekeyhq.atlassian.net/browse/OK-63144)

[OK-63144]: https://onekeyhq.atlassian.net/browse/OK-63144?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Related follow-up: [OK-63226](https://onekeyhq.atlassian.net/browse/OK-63226) — make the account overview scroll with the outer page on Web and Desktop. This follow-up is tracked separately from this merged PR.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-8844fafe-7d49-57b0-8d68-5ebb96bf4673)

sidmorizon:
Claude review session: https://claude.ai/code/session_017fKqwjs6moSL5oXVbXf63L

packages/shared/src/web/index.css
```diff
@@ -109,6 +109,50 @@ body {
...
+[data-perp-table-scrollbar='true']::-webkit-scrollbar-thumb {
+  background: linear-gradient(var(--borderStrong), var(--borderStrong)),
+    var(--bgApp);
+  border-radius: 2px;
+}
+
+[data-perp-table-scrollbar='true']::-webkit-scrollbar-thumb:hover {
+  background: linear-gradient(var(--textDisabled), var(--textDisabled)),
+    var(--bgApp);
+}
```
sidmorizon: # 📝 P2 — `[data-perp-table-scrollbar]` rules never style a visible scrollbar, and on older WebKit they un-hide the native one under the new JS overlay

## Problem

This block styles a 4px native scrollbar for elements tagged `data-perp-table-scrollbar="true"`. That attribute has exactly two producers repo-wide — `CommonTableListView.tsx:966` and `:1099` — and both tag a `PerpTableScrollView`:

```jsx
dataSet={!platformEnv.isNative ? { perpTableScrollbar: 'true' } : undefined}
```

On web/desktop/ext that resolves to `PerpTableScrollView.web.tsx`, which forces both indicators off *after* the spread (`:110-113`):

```jsx
<ScrollView {...props} ref={scrollRef}
  showsVerticalScrollIndicator={false}
  showsHorizontalScrollIndicator={false} />
```

`BaseScrollView` (`packages/components/src/layouts/ScrollView/index.tsx:72-89`) then computes `showsScrollIndicator === false` and merges `onekeyScrollViewScrollbar: 'hidden'` into the *same* `dataSet`, preserving the caller's keys. So every node carrying `data-perp-table-scrollbar="true"` also carries `data-onekey-scroll-view-scrollbar="hidden"`, matching the pre-existing rules at `index.css:94-101`:

```css
[data-onekey-scroll-view-scrollbar='hidden'] { scrollbar-width: none; }
[data-onekey-scroll-view-scrollbar='hidden']::-webkit-scrollbar { width: 0; height: 0; }
```

Both `::-webkit-scrollbar` rules have identical specificity, so the new one wins on source order and restores a 4px track. It is re-suppressed only where `scrollbar-width: none` is honored and takes precedence over `::-webkit-scrollbar` sizing. There is no code path in which these rules style a scrollbar the user is meant to see.

## Impact

Web, desktop (Electron) and extension only — native is unaffected, since `dataSet` is `undefined` there.

On Chromium-based targets (Electron desktop, Chrome/Edge, extension) these 15 lines plus the two `dataSet` props are inert dead code that reads as the active scrollbar styling for perps tables, which will mislead the next person tuning it.

On WebKit builds without `scrollbar-width` support (Safari before 18.2) the declaration is dropped and the 4px `::-webkit-scrollbar` applies. Because an explicit `::-webkit-scrollbar` width turns Safari's overlay scrollbar into a classic space-consuming one, the OrderInfoPanel tables then render a native 4px track *in addition to* the `.perp-table-overlay-scrollbar` thumb drawn by `PerpTableScrollView.web.tsx:115-125`, and that track steals 4px from the table body — the exact header/body alignment this PR works to keep consistent.

## Recommendation

Pick one mechanism rather than layering two.

Since `PerpTableScrollView.web.tsx` hard-disables both indicators and draws its own thumb, the overlay is the intended mechanism: delete `index.css:112-126` and the two `dataSet={{ perpTableScrollbar: 'true' }}` props at `CommonTableListView.tsx:965-967` and `:1098-1100`.

If instead the native-scrollbar styling is meant as a fallback for engines where the JS overlay is undesirable, scope it so it cannot co-occur with the hidden marker — e.g. `[data-perp-table-scrollbar='true'][data-onekey-scroll-view-scrollbar='visible']` — and stop forcing the indicators off for those scrollers in `PerpTableScrollView.web.tsx:110-113`.

[Claude session ↗](https://claude.ai/code/session_017fKqwjs6moSL5oXVbXf63L)

---

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
```diff
@@ -953,9 +959,12 @@ export function CommonTableListView<T>({
   );
   const desktopTable = (
     <XStack flex={1}>
-      <ScrollView
+      <PerpTableScrollView
         ref={scrollViewRef}
         testID="perp-desktop-table-body"
+        dataSet={
+          !platformEnv.isNative ? { perpTableScrollbar: 'true' } : undefined
+        }
```
sidmorizon: 🟡 P2: Desktop tables render duplicate scrollbars

Severity: `non-severe`

On Chromium, `dataSet` re-enables a native 4px scrollbar beneath the custom overlay thumb. The later [scoped scrollbar rule](packages/shared/src/web/index.css:112-126) overrides the hidden selector, leaving a gutter and overlapping thumbs.

packages/shared/src/web/index.css
Kahnchan: Removed the unused native scrollbar CSS and both perpTableScrollbar dataSet props. Perps tables now rely solely on the 4px JS overlay, so the existing hidden-indicator rules cannot be overridden by these legacy selectors.

packages/kit/src/views/Perp/components/OrderInfoPanel/List/CommonTableListView.tsx
Kahnchan: Fixed in 769c368b92, now included in the PR head: removed both perpTableScrollbar dataSet props and the matching native scrollbar CSS. The existing hidden-indicator rules remain intact, and only the custom overlay thumb is rendered.

originalix (approved):

PR #13401, opened 2026-09-14
fix: market list header, banner cards and detail trade column polish

erikzou:
Market list header, banner cards and detail trade column polish. Everything is scoped to Market.

## List page banner (desktop)

- Redesign the theme and index cards into one layout: 320px wide, title row with the quote list below, and three quote rows (20px logo, 80px name, then price and change as equal-width right-aligned columns with a 6px right inset).
- Index cards use the same row layout, localized names (`market.index_sp500__title`, `market.index_nasdaq__title`, `market.index_dow30__title`, pulled from Lokalise) and dedicated icons (S&P 500, Nasdaq Composite, Dow 30). The index card title still comes from the server.
- Hover changes the background (`$bgHover`, `$bgActive` on press) instead of fading the card. The More label shows on hover on desktop web and always on touch layouts (mobile and native).
- Horizontal scrolling follows the wallet home banner: edge arrows over a background gradient that fade in while cards are hidden.
- Spacing follows the design's `Banner` frame: 12px above it, 20px above and below the cards.

## List page banner (mobile)

- Cards are 288px wide and the scroller pads 16px top and bottom.
- The fixed native header heights follow: 212 for modern cards and 150 for legacy ones. The loading fallback uses the same card size and header height, so the page does not jump when banners arrive.

## List page tabs (desktop)

- The time range and network controls follow the last tab with a 32px gap instead of pinning to the right edge.

## Stocks list

- The day range sparkline downsamples the 390 one-minute points of a trading session to 40 evenly spaced ones (about every 10 minutes, always keeping the first and latest price).
- It now uses the Top Coins sparkline component: same line, smoothing, gradient fill and colors (`MARKET_SPARKLINE_COLORS` is shared). `SparklineChart` now draws a flat series at mid height instead of dividing by zero.

## Detail pages (desktop: stock, token, top coins)

- The trade column keeps 384px as its minimum and scales with the content row above a 1280px window (about 31% of the row, up to about 446px at the 1440px frame cap). The three layouts share `MARKET_DETAIL_TRADE_COLUMN_PROPS`.
- The trade column is sticky: it pins 16px below the top of the detail scroll container once the page header scrolls away.

## Other

- Chart settings (desktop and mobile quick settings) and the native chart's reference line label use `market.prev_close` instead of hard-coded "Prev close".
- The Top Coins token selector shows the full coin name as the row subtitle instead of repeating the symbol.

## Merge notes

#13388 landed in the meantime and scaled the banner's numeric heights with `s()` for Android high-density devices. The resolution keeps the redesigned card and applies the same scaling to its numeric heights (title row, quote list minimum height, card height and the modern native header height). #13383's `PerpetualTradingBanner` stays at the top of the token detail trade column.

## Test plan

- [x] Jest: Market home banner, layouts and detail layouts (37 suites), TradingView chart settings, token selector
- [ ] Desktop web: banner cards, hover, More, scroll arrows, tab controls position
- [ ] Desktop web: detail trade column scaling across 1280 / 1440 and sticky behavior
- [ ] iOS / Android: banner card size and header height, More always visible


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-68d45402-2932-532b-843c-15f3bb50c211)

sidmorizon:
Claude review session: https://claude.ai/code/session_01FjAB7MiygneRDVxiN3tfZH

packages/kit/src/views/Market/MarketHomeV2/components/MarketBanner/MarketBannerList.tsx
```diff
@@ -58,7 +59,7 @@ function BannerContainerMobile({
         bounces={false}
         showsHorizontalScrollIndicator={false}
         contentContainerStyle={{
-          py: '$2',
+          py: '$4',
```
originalix, line 62: P2: [Keep scaled banner padding within the Android header height]

On narrow Android devices, `getMarketMobileBannerHeaderHeight()` reserves `s(212)`, while a modern card already contains scaled `s(30)` and `s(92)` regions. Adding unscaled `$4` padding above and below makes the horizontal ScrollView's content taller than its fixed header, so the bottom of modern banner cards can be clipped.

Derive the reserved height from the same scaled card and padding values, or scale these vertical paddings consistently.

erikzou: Verified this one as a false positive: the vertical padding is already scaled.

- `$4` is not a raw 16: the space scale in `packages/components/tamagui.config.ts` defines `4: s(16)` (every space token goes through `s()`).
- The `ScrollView` wrapper resolves `contentContainerStyle` through `useStyle(contentContainerStyle, { resolveValues: 'auto' })` (`packages/components/src/layouts/ScrollView/index.tsx`), so `py: '$4'` becomes `s(16)` on narrow Android devices.
- The card is built from the same scale: `pt="$3.5"` (s(14)) + title row `s(30)` + `gap="$6"` (s(24)) + list `minHeight` `s(92)` + `pb="$5"` (s(20)) = s(180), plus 2 × s(16) padding = s(212), which is what `getMarketMobileBannerHeaderHeight()` reserves.

I also checked per-segment `PixelRatio.roundToNearestPixel` rounding at `uiScale` 0.9 for pixel ratios 2, 2.625, 2.75, 3, 3.5 and 4: content and reserved height match exactly except at 2.75, where the sum exceeds `s(212)` by 0.36dp (one physical pixel). That pixel falls inside the card's empty 20dp bottom padding and, with the Android `alignItems: 'center'` from #13388, is split between top and bottom, so no card content is clipped. No change needed.

packages/kit/src/views/Market/MarketDetailV2/layouts/StockDesktopLayout.tsx
```diff
@@ -1308,9 +1310,8 @@ export function StockDesktopLayout({
 
         <Stack
           testID="stock-token-detail-trade"
-          width={STOCK_DETAIL_TRADE_PANEL_WIDTH}
-          pt="$6"
-          flexShrink={0}
+          {...MARKET_DETAIL_TRADE_COLUMN_PROPS}
+          mt="$6"
```
sidmorizon, line 1314: 🟡 P2: Stock trade panel pins too low

Severity: `non-severe`

When the stock trade column sticks, `mt="$6"` combines with the shared 16px inset. Its panel pins 40px below the container.

originalix (approved):

PR #13381, opened 2026-09-14
feat: integrate native sheets and upgrade native modules to 3.0.131(OK-62487 OK-62488 OK-62237 OK-62495 OK-62953)

huhuanming:
[OK-62487](<https://onekeyhq.atlassian.net/browse/OK-62487>) [OK-62488](<https://onekeyhq.atlassian.net/browse/OK-62488>) [OK-62237](<https://onekeyhq.atlassian.net/browse/OK-62237>) [OK-62495](<https://onekeyhq.atlassian.net/browse/OK-62495>) [OK-62953](<https://onekeyhq.atlassian.net/browse/OK-62953>)

----------

## Summary

- add a reusable native sheet presentation bridge for arbitrary React Native content
- migrate account and wallet action sheets plus rename dialogs to native presentation while preserving JS fallbacks
- align backdrop handling, nested-sheet lifecycle, dynamic height animation, keyboard behavior, and async skeleton sizing
- upgrade the integrated `@onekeyfe/*` native modules to `3.0.131` for the latest NativeSheet, OneKeyImage, NativeList, and Pager fixes

## Package integration

- consume the complete `3.0.131` native-module release across app manifests, Yarn resolutions, iOS pods, and patch-package filenames
- the release contains the Android image lifecycle/decode fixes, NativeList interaction fixes, Pager scroller fixes, and NativeSheet lifecycle/review fixes
- all 40 publishable app-modules packages were published and independently verified with `dist-tags.latest=3.0.131`

## Validation

- focused app Jest coverage passed: 5 suites, 18 tests
- `yarn agent:check --profile commit` passed
- Android JDK 17 production-debug Java/Kotlin compilation passed with the linked native modules
- iOS pod installation and New Architecture NativeSheet code generation passed
- NativeSheet tests passed: 17/17
- OneKeyImage JavaScript tests passed: 20/20
- focused OneKeyImage Android decode-dimension tests passed: 8/8
- package release CI passed: https://github.com/OneKeyHQ/app-modules/actions/runs/34817029590

## Current branch note

- the app dependency update is committed locally as `de9a6f3cd4`
- the remote PR branch advanced independently and still needs that commit synchronized before this dependency portion appears in the PR diff

[...]

PR #13403, opened 2026-09-14
fix: keep Android send amount visible after the fiat toggle (OK-63199)

weatherstar:
[OK-63199](<https://onekeyhq.atlassian.net/browse/OK-63199>)

----------

## Summary

OK-63199: on Android the send amount was clipped after toggling to fiat (`$ .45496324`, leading digit hidden). Root cause is in the native `@onekeyfe/react-native-auto-size-input` Android implementation. The fix is upstream in app-modules ([OneKeyHQ/app-modules#102](https://github.com/OneKeyHQ/app-modules/pull/102), published as `3.0.132`); this PR bumps every app-modules package from `3.0.131` (after #13381 landed on `x`) to `3.0.132`. The interim patch-package patch from the first revision is gone; the branch is a single commit on top of `x`.

- `recalculateFontSize()` (contentAutoWidth branch) measured the prefix/suffix labels once at the **current** font size, but `applyFontSize` scales them to the new size, so `performLayout` gave the EditText a slot narrower than its text whenever the font grew next to a visible side label. The EditText scrolled to the caret and hid the leading digits. The fit now probes text + prefix + suffix (incl. safety padding) at every candidate size (`AutoWidthFontFit`), mirroring `performLayout`.
- `prefixMarginRight` / `suffixMarginLeft` setters only called `requestLayout()`, which the parent `ReactViewGroup` swallows, so a gap change arriving after the prefix/suffix props never re-laid out the row (lost `amount ETH` gap after toggling back). They now re-run the fit.
- Adds `AutoWidthFontFitTest` (Kotlin unit test) for the size-aware fit (lives in app-modules, runs in its Gradle test task).

`3.0.131` → `3.0.132` delta in app-modules: this fix plus the last native-sheet presentation-race commit of OneKeyHQ/app-modules#101. `patches/react-native-pager-view+3.0.131.patch` and `patches/@onekeyfe+react-native-tab-view+3.0.129.patch` are renamed to `3.0.132` with unchanged contents (both still apply; pager-view / tab-view sources did not change between 3.0.131 and 3.0.132).

iOS measures prefix + text + suffix together per candidate size and was never affected.

## Test plan

- [x] Reproduced on Medium_Phone_API_36 emulator before the fix (`$ ).1121062018`, half a digit hidden)
- [x] After the fix: fiat → token → fiat toggles all render the full amount; `0.00004447 ETH` gap restored after toggling back; 13-digit fiat value shrinks and stays fully visible
- [x] `./gradlew :onekeyfe_react-native-auto-size-input:testDebugUnitTest` green in app-modules (3 new + 11 existing)
- [x] npm: all 40 `@onekeyfe/*` app-modules packages resolve at `3.0.132`; installed `react-native-auto-size-input` carries `AutoWidthFontFit.kt`; patch-package applies the renamed pager-view / tab-view patches cleanly
- [x] `apps/mobile/ios/Podfile.lock` refreshed (`AutoSizeInput (3.0.132)`), `module-id:check` OK
- [x] Re-verified with the published `3.0.132` (no patch) in the app on the API 36 emulator: Send → `0.45496324 BTC` → fiat toggle shows `$ 35392.50036608` with the leading digit visible and the whole value fitted; toggling back restores `0.45496324 BTC` with the gap intact
- [ ] QA on a physical Android device with the reporter's cbBTC case

Jira: https://onekeyhq.atlassian.net/browse/OK-63199

[OK-63199]: https://onekeyhq.atlassian.net/browse/OK-63199?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-0e563fb2-d103-5611-8cde-9ee89f2b05da)

sidmorizon:
Claude review session: https://claude.ai/code/session_01224F8JPRVA9oWzqsuD7vcg

patches/@onekeyfe+react-native-auto-size-input+3.0.129.patch
```diff
@@ -0,0 +1,237 @@
...
++}
+diff --git a/node_modules/@onekeyfe/react-native-auto-size-input/android/src/test/java/com/margelo/nitro/autosizeinput/AutoWidthFontFitTest.kt b/node_modules/@onekeyfe/react-native-auto-size-input/android/src/test/java/com/margelo/nitro/autosizeinput/AutoWidthFontFitTest.kt
+new file mode 100644
+index 0000000..b9d25ac
+--- /dev/null
++++ b/node_modules/@onekeyfe/react-native-auto-size-input/android/src/test/java/com/margelo/nitro/autosizeinput/AutoWidthFontFitTest.kt
+@@ -0,0 +1,45 @@
++package com.margelo.nitro.autosizeinput
++
++import org.junit.Assert.assertEquals
```
originalix: P2: [Added Android test lacks its JUnit dependency]

When Gradle compiles `AutoWidthFontFitTest`, the new `org.junit.*` imports are resolved from the library's test compile classpath. This patch adds the Kotlin test source but no `testImplementation` dependency, so `testDebugUnitTest` fails before the regression checks run.

Add an `android/build.gradle` hunk to this patch declaring `testImplementation "junit:junit:4.13.2"` so the test task compiles and executes.

weatherstar: The library's `android/build.gradle` already declares `testImplementation "junit:junit:4.13.2"` (it is what the existing `TextUpdateEventCounterTest` compiles against), so the test compiled and ran as-is: `testDebugUnitTest` reports `AutoWidthFontFitTest` 3/3 and `TextUpdateEventCounterTest` 11/11.

That said, the patch is gone now: the fix is upstream in OneKeyHQ/app-modules#102 and published as 3.0.132, and this PR bumps the app-modules packages to 3.0.132 instead of patching 3.0.129.

originalix (approved):

PR #13405, opened 2026-09-14
fix: handle cloud backup bottom safe area(OK-62769)

huhuanming:
[OK-62769](<https://onekeyhq.atlassian.net/browse/OK-62769>)

----------

## Summary
- Add dynamic bottom safe-area padding to the Google Drive backup list and detail pages.
- Preserve the existing 20px design spacing while keeping backup, import, and delete actions above Android system navigation.

## Intent & Context
OK-62769 reports that the Google Drive backup and management/delete pages are still obstructed on Android devices using three-button navigation. The earlier dialog inset fix did not cover these standalone onboarding pages.

Related issue: https://onekeyhq.atlassian.net/browse/OK-62769

## Root Cause
Both pages disable the Page-level safe area and only provide a fixed 20px vertical content padding. Under Android edge-to-edge rendering, a three-button navigation inset can be larger than the fixed spacing, allowing the bottom content and actions to render behind the system navigation bar.

## Design Decisions
- Let each scroll content container own its bottom inset because these pages intentionally keep `safeAreaEnabled={false}`.
- Use `Math.max(bottom, 20)` so the safe area replaces, rather than stacks with, the existing bottom design padding.
- Keep the action-row `$3` padding unchanged and independent from the page inset.

## Changes Detail
- `ICloudBackup.tsx`: apply the safe-area-aware bottom padding to the backup list page.
- `ICloudBackupDetails.tsx`: apply the same padding to backup, restore/import, and management/delete detail states.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (Android primarily; iOS retains at least the existing spacing)
- **Risk Areas**: Bottom spacing on devices with unusually large or zero bottom insets.

## Test plan
- [ ] Verify the Google Drive backup list can scroll fully above Android three-button navigation.
- [ ] Verify backup, import, and delete actions remain visible and tappable above Android three-button navigation.
- [ ] Verify gesture-navigation Android and iOS retain the expected bottom spacing.

[OK-62769]: https://onekeyhq.atlassian.net/browse/OK-62769?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-c8d77cb0-c376-5cb8-8bd7-d92300107eb2)

sidmorizon:
Claude review session: https://claude.ai/code/session_0195riPfL5FCKUhPYGw2kTr3

huhuanming:
### Why this change

Both cloud-backup pages explicitly set `safeAreaEnabled={false}`. That setting was retained during the onboarding layout migration, but the current scroll layout no longer has the old `OnboardingLayout` wrapper that added the native bottom inset. As a result, `paddingVertical: 20` became the only bottom protection.

This happened to look acceptable on many iPhones because the existing detail-page spacing (`20` from the content container plus `$3`/`12` from the action row) is close to the common iOS home-indicator inset. Android three-button navigation can expose a larger bottom inset (commonly around 48dp), so the fixed spacing is insufficient and the backup/import/delete actions can render behind the system navigation bar.

The patch makes the scroll content the single safe-area owner and changes the bottom spacing to `Math.max(bottom, 20)`. This preserves the existing 20px design minimum when no native inset is present, uses the real inset when it is larger, and avoids blindly stacking `bottom + 20`. The action row keeps its independent `$3` visual spacing.

### Why this is not the optimal long-term solution

This is intentionally a scoped, tactical fix for OK-62769. It duplicates the same inset ownership logic in two page files and requires each page author to remember that `safeAreaEnabled={false}` transfers responsibility away from `Page`.

A better long-term design would let `OnboardingPage`/`Page` declare a content-owned bottom-safe-area mode and merge the inset with the scroll container's design padding centrally, or move persistent actions into a correctly configured `Page.Footer`. That would make safe-area ownership explicit and enforce it in one shared layer. Changing that shared contract in this PR would affect the entire onboarding flow and carries a much wider regression surface, so this PR keeps the change local and minimal.

originalix (approved):

PR #13409, opened 2026-09-14
fix: center device stage on Android foldables(OK-63187)

originalix:
[OK-63187](<https://onekeyhq.atlassian.net/browse/OK-63187>)

----------

## Summary
- Center Android DeviceStage overlays across the full split-view viewport without moving their portal host.
- Cap expanded foldable cards at the existing desktop maximum of 400 dp while retaining native bottom anchoring and downward dismissal.
- Preserve ordinary phone sizing and add viewport and split-view regression tests.

## Intent & Context
Fix [OK-63187](https://onekeyhq.atlassian.net/browse/OK-63187). The expanded foldable DeviceStage was shifted right and clipped, and its card could stretch across a wide screen. The requested result is desktop-like maximum width with native bottom-center positioning, without changing ordinary phone behavior.

## Root Cause
The hardware overlay host lives inside the split-view detail pane, while MorphOverlay calculates horizontal geometry from the full window width. Their coordinate origins therefore differ when the main pane is visible. Width capping and vertical anchoring also shared a media-breakpoint decision that does not represent expanded Android foldables correctly.

## Design Decisions
- Measure the actual Android detail-pane offset and extend only the hardware host back to the split-view origin. Provide zero offset when folded, onboarding, or detail-fullscreen hides the main pane.
- Keep the existing portal host, manager, context, and sibling order. Moving it to the root was considered but could change ordering against dialogs, toasts, and the app lock or remount active portal content.
- Separate foldable width capping from bottom anchoring, including safe-area clearance, animation direction, and drag dismissal. Ordinary phones retain their existing formulas; iOS, desktop, web, and extension do not opt into Android foldable behavior.
- Scope is Android UI/main layout. Background hardware communication and shared native resources are unchanged; no cross-runtime readiness or shared JS heap assumptions are introduced.
- The branch was synchronized with `x`; conflict resolution preserves the upstream keyboard-controller feed, hidden-shell opacity gate, scrim animation, and iOS stage/toast raise tokens. Emulator evidence below was collected before this base synchronization; the targeted suites were rerun on the combined code.

## Changes Detail
- `MorphOverlay/index.tsx`: distinguish card width constraints from foldable bottom-sheet anchoring.
- `TableSplitViewContainer/index.tsx` and `SplitViewDetailOffsetContext.tsx`: measure and expose the active Android split-view offset.
- `FullWindowOverlayContainer/index.tsx`: apply that offset to the existing hardware overlay host only.
- `MorphOverlay/index.dismiss.test.tsx` and `TableSplitViewContainer.test.tsx`: cover foldable postures, ordinary Android, iOS, desktop, measured offsets, onboarding, and detail fullscreen.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Android mobile; shared components have unchanged non-Android behavior covered by unit tests.
- **Risk Areas**: Touch interception outside the detail pane, posture transitions, safe-area anchoring, and overlay stacking. Existing host structure is retained, and emulator input/mask checks passed. Physical hardware, vendor-specific foldables, and password/app-lock stacking combinations have not been runtime-verified.

## Test plan
- [x] Reproduce the original offset/clipping with the foldable emulator using Gallery DeviceStageDriver simulated events through the real pipeline.
- [x] Verify expanded foldable card width, whole-window bottom centering, device illustration, three-column PIN input, and mask interception; verify folded layout.
- [x] Verify Pixel 9 API 36 ordinary-phone width, bottom centering, three-column PIN input, close button, downward dismissal, mask interception, restored navigation, and Pro 2 device illustration.
- [x] Run both targeted Jest suites (25 tests).
- [ ] Verify physical hardware and password/app-lock overlay combinations on Android.

[OK-63187]: https://onekeyhq.atlassian.net/browse/OK-63187?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wabicai (dismissed):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5f39a342-8642-50cd-829c-ca87369e2381)

sidmorizon:
Claude review session: https://claude.ai/code/session_01MU9AEJKkqMWuT4ywG96eRt

packages/kit/src/provider/Container/TableSplitViewContainer/index.tsx
```diff
@@ -25,9 +33,20 @@ export function TableSplitViewContainer({
...
+        flex={1}
+        onLayout={platformEnv.isNativeAndroid ? onDetailLayout : undefined}
+      >
         <SplitViewDetailFullscreenProvider value={setIsDetailFullscreen}>
-          {detailRouter}
+          <SplitViewDetailOffsetProvider
+            value={
+              platformEnv.isNativeAndroid && display === 'flex'
+                ? detailOffset
+                : 0
```
sidmorizon, line 68: 🟡 P2: Split transitions briefly misplace DeviceStage

Severity: `non-severe`

After folded or fullscreen layout returns to split view, `detailOffset` remains zero for the first split render. `onLayout` updates it only after that layout commits. A visible DeviceStage jumps right and clips until the follow-up render.

packages/kit/src/provider/Container/TableSplitViewContainer/SplitViewDetailOffsetContext.tsx
```diff
@@ -0,0 +1,10 @@
+import { createContext, useContext } from 'react';
```
originalix, line 1: P1: [Register the new module for strict mobile bundles]

When a production or `UNION_BUILD` mobile bundle traverses `TableSplitViewContainer`, it synchronously imports this new file. Metro's strict `fileToIdMap` has no registry entry for this path, so module-ID lookup throws before the bundle is produced.

Regenerate and commit `apps/mobile/bundle-registry/module-id-registry.json` through the repository's module-ID registry update flow so this exact file receives a stable ID.

packages/kit/src/provider/Container/TableSplitViewContainer/index.tsx
originalix: Fixed in d5c0e01a29. Preserve the last valid split geometry across hidden/fullscreen/onboarding layout events and only expose zero while the split is inactive. Added regression tests for zero/full-width hidden events followed by split restoration, plus both layout-event orders. The targeted suites pass (40 tests total). Foldable emulator confirms the card returns to whole-window bottom center.

packages/kit/src/provider/Container/TableSplitViewContainer/index.tsx
```diff
@@ -17,6 +21,10 @@ export function TableSplitViewContainer({
   const isLandscape = useIsSplitView();
   const [isOnBoardingOpen] = useIsOnBoardingOpenAtom();
   const [isDetailFullscreen, setIsDetailFullscreen] = useState(false);
+  const [detailOffset, setDetailOffset] = useState(0);
+  const onDetailLayout = useCallback((event: LayoutChangeEvent) => {
+    setDetailOffset(event.nativeEvent.layout.x);
```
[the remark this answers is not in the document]
originalix: Fixed in d5c0e01a29. Hidden or fullscreen layout events no longer overwrite the cached split geometry. The provider returns zero only while the split is inactive and restores the cached offset immediately. Tests explicitly emit x=0 and full-width hidden layouts before restoring split mode.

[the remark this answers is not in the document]
originalix: Fixed in d5c0e01a29. Measure container width minus detail width instead of physical x, and use logical start/end for the Android overlay. Verified on the foldable emulator with native forceRTL enabled: main pane on the right, PIN card centered across the whole window at the bottom, all three PIN columns interactive, right-pane Trade blocked while the overlay is open and reachable after closing. Restored the original LTR preference after testing. Added an RTL x=0 regression test.

packages/kit/src/provider/Container/FullWindowOverlayContainer/index.tsx
```diff
@@ -84,7 +86,9 @@ export function FullWindowOverlayContainer() {
         <Stack
           position="absolute"
           top={0}
-          left={0}
+          // Extend to the actual split-view origin without relocating the portal
+          // or changing its order relative to dialogs, toasts, and the app lock.
+          left={-detailOffset}
```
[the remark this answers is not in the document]
originalix: Investigated against React Native 0.86.2 and tested on the Android foldable emulator. TouchTargetHelper supports hits outside a parent through ReactOverflowViewWithInset when the host is not clipped; a negative offset is not drawing-only in this hierarchy. Runtime negative/positive controls pass: with the PIN stage open, the main-pane Trade tab cannot activate; after closing, the same coordinate activates Trade. All PIN columns work, including outside the detail pane. Repeated with LTR and RTL on d5c0e01a29. Keeping the existing Portal hierarchy avoids changing dialog/toast layering; no host relocation is needed for this reported case.

packages/kit/src/provider/Container/TableSplitViewContainer/SplitViewDetailOffsetContext.tsx
originalix: Fixed in d5c0e01a29 by registering SplitViewDetailOffsetContext.tsx as module ID 3552 using the registry tooling. module-id:check and the 13 registry tests pass. Strict lookup now succeeds for this path; removing the entry reproduces the prior failure. This was also the root cause of the failed Native startup graph CI Union Build, which is rerunning on the fix.

huhuanming (approved):

PR #13424, opened 2026-09-15
fix: keep Android send amount keyboard in sync with page transitions

weatherstar:
## Summary
- Re-run the send amount input auto-focus on every route focus so returning from the confirm page brings the keyboard back on Android.
- Blur the Nitro amount input in `beforeRemove` so the keyboard hides before the exit transition when the user taps header back with the keyboard open.

## Intent & Context
QA (lixiao, Slack #app-bug 2026-09-15, Android) reported two keyboard problems in the send flow on Android:
1. Enter the transaction confirm page, tap back to the amount page: the keyboard does not pop up again.
2. With the keyboard open on the amount page, tap the header back arrow: the keyboard stays open on the previous (recipient) page.

Both were reproduced on a Pixel 4a (Android 13, gesture navigation) with `adb` driving the flow and `dumpsys input_method` / `dumpsys activity top` used to read IME state and native focus.

## Root Cause
- The amount input is the `@onekeyfe/react-native-auto-size-input` Nitro HybridView (a plain `android.widget.EditText` inside a `HybridAutoSizeInput` view), not RN's `ReactEditText`.
- Issue 1: auto-focus was a mount-only 300 ms effect. While the confirm page is on top, react-native-screens detaches the amount screen on Android, so the native EditText loses focus. On return nothing refocuses it, so no keyboard.
- Issue 2: when the amount screen is popped, Android clears the EditText focus during the exit transition. RN's `ReactEditText` hides the IME in `clearFocusAndMaybeRefocus`; the Nitro input does not. Android focus recovery then lands on the next focusable input in the window (the home tab search `TextInputView` under the modal), GBoard re-binds to it (QWERTY layout) and the keyboard stays open. Logcat showed `onFinishInput → onStartInput → onWindowShown` and never `onWindowHidden`; the `ScreenContainer.removeView` keyboard-hide workaround in react-native-screens does not fire because focus has already left the screen.
- Hardware/gesture back is unaffected: the IME consumes `KEYCODE_BACK` itself.

## Design Decisions
- Refocus uses `useFocusEffect` (route focus) instead of the mount effect, so it also covers returning from any pushed route (confirm page, coin control). Same 300 ms delay as before.
- The blur is registered on `navigation.addListener('beforeRemove')` and gated to native. A `useFocusEffect`/unmount cleanup was tried first and does not work: by the time passive cleanups run on unmount, `amountInputRef.current` is already null (refs detach in the mutation phase). `beforeRemove` fires while the native view is alive; on device the blur produced `onWindowHidden` 9 ms later.
- `Keyboard.dismiss()` was not an option: it is a no-op for non-RN inputs on Android. The existing `dismissAmountInputKeyboardBeforeOverlayOpen` helper (native `blur()` + delayed dismiss) is left untouched.
- No native change to the Nitro input package; the fix stays in the page so it does not need an app-modules release.

## Changes Detail
- `packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx`
  - Replace the mount-only auto-focus `useEffect` with `useFocusEffect`.
  - Add a native-only `beforeRemove` listener that calls `amountInputRef.current?.blur()`.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (Android primary; iOS gets the same refocus-on-return and blur-before-pop behavior)
- **Risk Areas**: iOS was not device-tested. react-native-screens on iOS already restores the first responder in some transitions, so the added refocus should be a no-op or consistent there; worth a quick QA pass. Desktop/web only gain the refocus on route regain (blur is native-only).

## Test plan
- [x] Android (Pixel 4a, Android 13): amount page → Preview → confirm page → header back: keyboard pops up, EditText focused (`mInputShown=true`).
- [x] Android: amount page with keyboard open → header back: keyboard hides (`onWindowHidden`), recipient page shows no keyboard.
- [x] Android: fresh entry to the amount page still auto-focuses; confirm page shows no keyboard; hardware back with keyboard open still only dismisses the keyboard.
- [x] `yarn agent:check --profile commit` passes.
- [ ] iOS: same round trips (confirm → back, header back with keyboard open).


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-6c85bfda-9599-5d64-84a1-e84c49638396)

sidmorizon:
Claude review session: https://claude.ai/code/session_018a1A5MsyWbyY2FUyYCS5Kj

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -2510,13 +2514,36 @@ function SendAmountInputContainer() {
   // Ref to track submit disabled state for keyboard shortcuts
   const isSubmitDisabledRef = useRef(true);
 
-  // Auto-focus the amount input after page transition animation completes
+  // Auto-focus the amount input after the page transition animation completes.
+  // Runs on every route focus rather than only on mount: Android
+  // (react-native-screens) detaches this screen while the confirm page is on
+  // top, which drops the native focus, so returning from it needs a fresh
+  // focus() to bring the keyboard back.
+  useFocusEffect(
```
originalix, line 2525: P2: [Repeated autofocus regresses non-Android send flows]

When this route regains focus after any child route is closed, `useFocusEffect` schedules `amountInputRef.current?.focus()` after 300 ms on every platform. The old effect ran only at mount, so iOS, desktop, and web now have their active control forcibly replaced by the amount input; on iOS this can also reopen the software keyboard.

Keep the existing initial autofocus for non-Android targets, and restrict focus-on-return recovery to `platformEnv.isNativeAndroid`, where the detached-screen behavior occurs.

weatherstar: Fixed: the focus-on-return recovery is now Android-only. `useFocusEffect` still drives the initial autofocus, but a `hasAutoFocusedAmountInputRef` guard makes non-Android targets (iOS, desktop, web) bail out after the first focus, matching the previous mount-only behavior. Only `platformEnv.isNativeAndroid` re-focuses on every route focus, where react-native-screens detaching the screen drops the native focus.

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -2510,13 +2514,46 @@ function SendAmountInputContainer() {
...
-    const timer = setTimeout(() => {
-      amountInputRef.current?.focus();
-    }, 300);
-    return () => clearTimeout(timer);
-  }, []);
+    if (!platformEnv.isNative) {
+      return undefined;
+    }
+    return reactNavigation.addListener('beforeRemove', () => {
+      amountInputRef.current?.blur();
```
originalix, line 2557: P2: [Android exit handler does not globally dismiss the IME]

When header back removes this screen while the amount keyboard is open, this only blurs the Nitro HybridView. The shared keyboard helper documents that this input can leave the Android IME visible when focus is cleared, so the keyboard can still survive the transition and move to a focusable control on the previous page.

After blurring the input, call the shared `Keyboard.dismiss()` helper here so `KeyboardController` performs a global IME dismissal before the route exits.

weatherstar: Fixed: the `beforeRemove` handler now calls the shared `Keyboard.dismiss()` (KeyboardController global dismiss) right after `amountInputRef.current?.blur()`, matching what `dismissAmountInputKeyboardBeforeOverlayOpen` already does, so the Android IME is hidden before the route exits even if clearing the Nitro input's focus alone leaves it visible.

originalix (approved):

PR #13432, opened 2026-09-15
fix: stabilize iOS send amount keyboard animation

huhuanming:
## Summary

- wait for the iOS page transition to finish before auto-focusing the amount input
- keep the regular-send balance card on one iOS native compositing layer during the keyboard animation
- preserve the existing Android, desktop, and web focus behavior

## Validation

- iOS simulator: verified the regular send amount screen while the keyboard and balance content animate concurrently


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-ee7426e9-8a33-5f9f-862e-668b9cd8ba7b)

sidmorizon:
Claude review session: https://claude.ai/code/session_01Fqkd6Cd9FPRNAoZJCWsgWs

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -2514,16 +2515,28 @@ function SendAmountInputContainer() {
...
+  // bottom instead of entering sideways with the screen. Keep this initial
+  // focus one-shot so returning from a child route does not reopen the iOS
+  // keyboard.
   const hasAutoFocusedAmountInputRef = useRef(false);
+  usePageMounted(() => {
+    if (!platformEnv.isNativeIOS || hasAutoFocusedAmountInputRef.current) {
+      return;
+    }
+    hasAutoFocusedAmountInputRef.current = true;
+    amountInputRef.current?.focus();
```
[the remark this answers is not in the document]
huhuanming: Fixed in bf70ee06a6. iOS auto-focus now runs inside useFocusEffect, removes the transition listener and cancels the fallback on blur or unmount, checks navigation.isFocused(), and marks the one-shot only after a live input ref is available. The lazy-load fallback is cancellable and delayed 500 ms so it cannot immediately focus during the native push animation.

packages/kit/src/views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx
```diff
@@ -2514,16 +2516,70 @@ function SendAmountInputContainer() {
...
+          },
+        );
+
+        // A cold lazy load can attach after transitionEnd. Keep the fallback
+        // beyond the native push window and cancel it as soon as focus is lost.
+        fallbackTimer = setTimeout(focusIfNeeded, IOS_AUTO_FOCUS_FALLBACK_MS);
+
+        return () => {
+          isActive = false;
+          clearFocusSchedule();
```
sidmorizon, line 2585: 🟡 P2: Return navigation reopens keyboard

Severity: `non-severe`

When iOS leaves before initial autofocus, `hasAutoFocusedAmountInputRef` stays false and reschedules on return. Returning from a child screen then reopens the keyboard.

huhuanming: Fixed in cc23aa9afa. Added a separate hasStartedIOSAutoFocusRef that is consumed on the first iOS focus lifecycle entry, while hasAutoFocusedAmountInputRef is still set only after the active-route and live-input checks pass. Leaving before focus now cancels the schedule, and returning from a child screen does not schedule it again.

originalix (approved):

PR #13434, opened 2026-09-15
fix: trigger iOS list pull-to-refresh from overscroll distance OK-63207

weatherstar:
[OK-63207](<https://onekeyhq.atlassian.net/browse/OK-63207>)

----------

## Summary

Fix iOS wallet home pull-to-refresh not triggering from the token list when the collapsible header is tall (push-permission alert + wallet banner, OK-63207).

Root cause: the list pull-to-refresh relies on UIKit's `UIRefreshControl`, which only fires at roughly 100–115pt of overscroll. iOS rubber-banding turns finger travel into much less content offset, so when the first token row sits at ~660pt the finger reaches the screen edge at ~99pt of overscroll and the control never fires (the spinner stays partially revealed and snaps back). This is not a regression in the refresh wiring; the header simply grew (banner in 2025-07, notification alert in 6.5.2) until the travel budget ran out.

Fix: add an overscroll-distance fallback to `PullToRefresh` on iOS.

- `react-native-collapsible-tab-view` patch: the container context exposes an `isScrollDragging` shared value, set from the focused tab's `onBeginDrag` / `onEndDrag`.
- `@onekeyhq/components`: export `CollapsibleTabNameContext` next to `CollapsibleTabContext` (native reads the lib's `TabNameContext`, web its own); web `TabsContext` default gains the new field.
- `PullToRefresh`: a reanimated reaction tracks the deepest `scrollYCurrent` while the focused tab's list is dragged and, when the drag ends past 65pt of overscroll, runs the same `handleRefresh` path (haptic + `refreshing` state, which makes the native control show and hold its spinner). A `refreshingRef` dedupes against the native trigger, so whichever fires first owns the cycle. Outside a `Tabs.Container` the hook is inert; Android and web paths are unchanged.

65pt of overscroll is about 130pt of finger travel, so the gesture stays deliberate while remaining reachable from a low list start.

## Validation

- iOS simulator (iPhone 17, same 402×874pt as iPhone 16 Pro), agent-device pan + screenshot timeline:
  - Before: start y=660 (first token row) with 210pt travel never refreshed; 255pt+ travel did.
  - After: the same 210pt pull enters and holds the refreshing state; an 80pt pull still does not trigger.
- Header-area pull-to-refresh (HeaderScrollGestureWrapper) is untouched.
- Device QA on a physical iPhone still pending.

[OK-63207]: https://onekeyhq.atlassian.net/browse/OK-63207?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-ab7d3912-38ed-5291-8dac-ddf77554bae0)

sidmorizon:
Claude review session: https://claude.ai/code/session_01TVQcx8E96DLnmApqsAVBXC

patches/react-native-collapsible-tab-view+8.0.1.patch
```diff
@@ -1272,7 +1296,13 @@ index ba7dffa..fc486ed 100644
        },
        progressViewOffset:
          // on iOS we need the refresh control to be at the top if overscrolling
-@@ -265,10 +326,13 @@ export const useScrollHandlerY = (name: TabName) => {
+@@ -260,15 +321,19 @@ export const useScrollHandlerY = (name: TabName) => {
+     contentInset,
+     containerHeight,
+     scrollYCurrent,
++    isScrollDragging,
```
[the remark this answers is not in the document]
weatherstar: Fixed in 9de8b3ced2: `FALLBACK_TABS_CONTEXT` now includes `isScrollDragging: makeMutable(false)`, so orphan `Tabs.*` list renders outside `Tabs.Container` no longer dereference `undefined` in the onBeginDrag/onEndDrag worklets. Patch regenerated via patch-package and verified against the pristine 8.0.1 tarball.

packages/kit/src/views/Home/components/PullToRefresh/index.tsx
```diff
@@ -16,26 +29,116 @@ export const onHomePageRefresh = () => {
...
+      return;
+    }
+    refreshingRef.current = true;
     onRefresh?.();
     setRefreshing(true);
     setTimeout(() => {
+      refreshingRef.current = false;
       setRefreshing(false);
-    }, 1200);
+    }, REFRESHING_INDICATOR_DURATION_MS);
```
sidmorizon, line 125: 🟡 P2: Refresh errors disable future pulls

Severity: `non-severe`

When `onRefresh` throws synchronously, `refreshingRef` remains true because its reset timer never starts. Every later pull exits without refreshing.

Suggested fix:

```
    refreshingRef.current = true;
    try {
      onRefresh?.();
      setRefreshing(true);
    } catch (error) {
      refreshingRef.current = false;
      throw error;
    }
    setTimeout(() => {
      refreshingRef.current = false;
      setRefreshing(false);
    }, REFRESHING_INDICATOR_DURATION_MS);
```

weatherstar: Fixed in c4e69b859a: the `refreshingRef`/spinner reset timer is now armed before `onRefresh` is invoked, so a synchronous throw can no longer leave the guard stuck and disable later pulls.

originalix (approved):

PR #13440, opened 2026-09-15
fix: keep recipient badges and recent list visible instead of reloading on every entry OK-63407

weatherstar:
[OK-63407](<https://onekeyhq.atlassian.net/browse/OK-63407>)

----------

## Summary
- `AddressInput` keeps the existing wallet / address-book badges on screen while a focus-regain re-validation runs; the spinner now only appears for the first query of an input or an `unknown`-status retry.
- Recent recipients (`useRecentRecipientsData`) get a per-session cache keyed by account + network: a re-opened Send page paints the previous list immediately and refreshes in the background, and networks the server reported as unsupported skip the `/transfer-recipient` round trip.
- `ServiceHistory.fetchTransferRecipients` returns `errored: true` when `supported: false` came from a failed request, so a transient network error is never memoized as "unsupported".

## Intent & Context
Reported in Slack #client-end (2026-09-15, thread p1789461561035689) with two screen recordings:

1. Android: with a recipient already filled, tapping "Check for risks" opens the Prime page; on returning, the "Wallet / Account #1" label inside the recipient input flips to a spinner for ~0.5 s and then comes back. This happens on every return to the screen.
2. Desktop, Cosmos: the "Recent" tab under the recipient input shows the skeleton on every entry to the Send page, even though `/transfer-recipient` does not support the chain and the local recent store is empty. The reporter asked that unsupported chains not show loading at all, and that supported chains reuse what was already loaded instead of showing the skeleton every time.

## Root Cause
1. `AddressInput` deliberately re-runs `queryAddress` whenever `useRouteIsFocused` flips (so address-book edits made on a pushed page refresh the label). `AddressInputBadgeGroup` rendered `<Spinner />` whenever `loading` was true, even though `queryResult` for the same input was still held in state, so every focus regain blanked the badges for the duration of validate-address + badges.
2. `useRecentRecipientsData.load` reset state to `isLoadingRecent=true` + `[]` on every mount and always awaited `fetchTransferRecipients` before falling back to the local store. For unsupported chains that is a full server round trip just to learn `supported: false`, and for supported chains the previously loaded list was discarded on every entry.

## Design Decisions
- The focus-regain re-query in `AddressInput` is kept; only the presentation changed. `shouldShowAddressQuerySpinner` (pure helper in `AddressInput/utils.ts`) decides the spinner: no result yet, or a retry after `validStatus === 'unknown'`. Re-validating the same input keeps the old badges and swaps them in place when the new result arrives.
- The recent-recipients cache is an in-memory module-level `Map` (stale-while-revalidate), not simpleDb. It satisfies "already loaded once in this session = no skeleton" without a schema change. Consequence: the first open of an unsupported chain after a cold start still shows the skeleton for one round trip.
- Only a server-side `supported: false` is memoized. `fetchTransferRecipients`' `callOnce` swallowed request errors as `supported: false`, which would have poisoned the cache on a flaky network; the new `errored` flag distinguishes the two. The multi-xpub merge reports `errored` only when every derive-path request failed.
- The cache key is `${accountId}__${networkId}`, so switching account or network still shows the skeleton on first load and never reuses another account's list.
- Traffic is unchanged: the background refresh runs the same API + per-address `queryAddress` pipeline as before; the EVM focus-regain chain in `queryAddress` (validate-address → badges → transfer-recipient) is untouched.

## Changes Detail
- `packages/kit/src/components/AddressInput/utils.ts`: new `shouldShowAddressQuerySpinner`.
- `packages/kit/src/components/AddressInput/index.tsx`: `AddressInputBadgeGroup` uses the helper instead of `if (loading)`.
- `packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts`: session cache (`recipients`, `lastUsedDeriveType`, `apiUnsupported`), state seeded from cache on mount, `commit` helper writes cache + state, unsupported networks skip the API; `clearRecentRecipientsCache` exported for tests.
- `packages/kit-bg/src/services/ServiceHistory.ts`: `fetchTransferRecipients` gains `errored?: boolean`.
- Tests: `AddressInput/utils.test.ts` (4 cases) and `SendDataInput/useRecentRecipientsData.test.tsx` (4 cases: unsupported memo, errored retry, cached paint + background refresh, no cross-account/network reuse).

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: the recent list can briefly show a stale entry from the previous visit until the background refresh replaces it (e.g. right after a send on a chain with indexer lag); an address that was valid on the last visit but is now rejected keeps its old badges only for the duration of the re-query, and the form value / Next button are unaffected because `pending` is not toggled by the re-query.

## Test plan
- [x] `yarn jest packages/kit/src/components/AddressInput packages/kit/src/views/Send/pages/SendDataInput packages/kit-bg/src/services/ServiceAccountProfile` (48 passed)
- [ ] Android: fill a recipient that belongs to a wallet account, tap "Check for risks", return from the Prime page: the account badge stays, no spinner
- [ ] Android: return from "Add to address book" flow: the badge updates to the address-book name without blanking
- [ ] Desktop, Cosmos (unsupported chain): open Send twice in one session; second open shows the empty state / local list with no skeleton
- [ ] EVM (supported chain): open Send, pick a recent recipient, go back, re-enter: the list is visible immediately and still refreshes (new send shows up after indexer catches up)
- [ ] Switch account or network on the Send page: the Recent tab still loads fresh for the new key

[OK-63407]: https://onekeyhq.atlassian.net/browse/OK-63407?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-64dcb373-0e48-5a7f-8500-511eff1006c8)

sidmorizon:
Claude review session: https://claude.ai/code/session_016WLDocFs2tg3LTMveGkbBf

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -327,22 +387,21 @@ export function useRecentRecipientsData({
...
-        if (isStale()) return;
-        setRecentRecipients(enriched);
       } catch {
         // ignore enrichment errors
       }
     }
 
     if (isStale()) return;
-    setIsLoadingRecent(false);
+    commit({ recipients: storedRecipients, apiUnsupported });
```
sidmorizon, line 438: 🟡 P2: Failed refresh clears cached recipients

Severity: `non-severe`

When the API refresh fails and local fallback is empty, `commit` replaces a populated cache with `[]`. The recent list disappears until another refresh succeeds.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -240,75 +240,135 @@ type IUseRecentRecipientsDataParams = {
...
+      setIsLoadingRecent(true);
+      setRecentRecipients([]);
+      setLastUsedDeriveType(undefined);
+    }
+
+    const commit = (entry: IRecentRecipientsCacheEntry) => {
+      recentRecipientsCache.set(cacheKey, entry);
+      setRecentRecipients(entry.recipients);
+      setLastUsedDeriveType(entry.lastUsedDeriveType);
+      setIsLoadingRecent(false);
```
sidmorizon, line 339: 🟡 P2: Obsolete loads overwrite newer cache

Severity: `non-severe`

After an instance unmounts, its pending `load` can still write through `commit`. A slower obsolete request can overwrite fresher cached recipients from another instance.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
weatherstar: Fixed in d96949727e: a failed API refresh (errored response or thrown request) now keeps the cached list and returns early. The local store only replaces the list on a cold load or for a network the server reported as unsupported, so a populated cache is never overwritten with an empty fallback. Covered by the new test 'keeps the cached list when a background refresh fails'.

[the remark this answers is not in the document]
weatherstar: Fixed in d96949727e: a failed API refresh (errored response or thrown request) now keeps the cached list and returns early. The local store only replaces the list on a cold load or for a network the server reported as unsupported, so a populated cache is never overwritten with an empty fallback. Covered by the new test 'keeps the cached list when a background refresh fails'.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
weatherstar: Fixed in d96949727e: loads are now versioned per cache key in a module-level map, so isStale() also trips when another instance started a newer load for the same account + network. Each instance additionally bumps its own load id on unmount, so an in-flight load from a closed Send page can neither set state nor write the shared cache. Covered by the new test 'ignores a slower load from an earlier instance'.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -240,75 +240,135 @@ type IUseRecentRecipientsDataParams = {
...
+      setLastUsedDeriveType(cached.lastUsedDeriveType);
+      setIsLoadingRecent(false);
+    } else {
+      setIsLoadingRecent(true);
+      setRecentRecipients([]);
+      setLastUsedDeriveType(undefined);
+    }
+
+    const commit = (entry: IRecentRecipientsCacheEntry) => {
+      recentRecipientsCache.set(cacheKey, entry);
```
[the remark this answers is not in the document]
weatherstar: Fixed in d96949727e: loads are now versioned per cache key in a module-level map, so isStale() also trips when another instance started a newer load for the same account + network. Each instance additionally bumps its own load id on unmount, so an in-flight load from a closed Send page can neither set state nor write the shared cache. Covered by the new test 'ignores a slower load from an earlier instance'.

packages/kit/src/components/AddressInput/index.tsx
```diff
@@ -220,7 +221,7 @@ type IResolvedAddressQueryContext = {
 
 function AddressInputBadgeGroup(props: IAddressInputBadgeGroupProps) {
   const { loading, result, setResolveAddress, onRefresh } = props;
-  if (loading) {
+  if (shouldShowAddressQuerySpinner({ loading, result })) {
```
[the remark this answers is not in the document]
weatherstar: Fixed in d96949727e: during re-validation (loading with an existing result) the badge group only keeps the wallet and address-book labels. The resolved-name address and server-derived label badges are hidden and replaced by a spinner until the new query lands, so a changed ENS record is never shown as the current resolution.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -240,75 +240,159 @@ type IUseRecentRecipientsDataParams = {
...
     }
 
+    const cacheKey = getRecentRecipientsCacheKey({ accountId, networkId });
+    const loadVersion = (recentRecipientsLoadVersion.get(cacheKey) ?? 0) + 1;
+    recentRecipientsLoadVersion.set(cacheKey, loadVersion);
+    // Stale when this instance started a newer load, unmounted, or another
+    // instance started a newer load for the same account + network.
+    const isStale = () =>
+      loadIdRef.current !== currentLoadId ||
+      recentRecipientsLoadVersion.get(cacheKey) !== loadVersion;
```
sidmorizon: 🟡 P2: Abandoned load strands earlier screen

Severity: `non-severe`

When a newer `load` unmounts, its version still invalidates an older mounted instance. Both requests become stale, leaving the older screen loading indefinitely.

packages/kit/src/components/AddressInput/index.tsx
```diff
@@ -259,7 +265,8 @@ function AddressInputBadgeGroup(props: IAddressInputBadgeGroupProps) {
             </XStack>
           </Badge>
         ) : null}
-        {result.resolveAddress ? (
+        {isRevalidating ? <Spinner /> : null}
```
sidmorizon, line 276: 🟡 P2: Revalidation still shows a spinner

Severity: `non-severe`

For any existing valid result, `isRevalidating` renders a spinner beside retained badges. Focus regain still exposes a loading indicator despite preserving labels.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -240,75 +240,159 @@ type IUseRecentRecipientsDataParams = {
...
+        apiFailed = true;
       }
-    } catch {
-      // API call failed — fall through to local fallback.
+    }
+
+    // A failed background refresh keeps the cached API list on screen; the
+    // local store is only a substitute for a cold load or for a network the
+    // server reported as unsupported (OK-53284: never mix sources).
+    if (apiFailed && cached && !cached.apiUnsupported) {
```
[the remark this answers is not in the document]
weatherstar: Fixed in ec50bdeb0a: the keep-cache return now sits behind `if (isStale()) return;`, so a thrown request (or a failed `enrichAddresses` after `supported: true`) from a load this instance has already moved past can no longer flip `isLoadingRecent` for the current network. The redundant `setIsLoadingRecent(false)` was dropped since `cached` is applied at the start of `load`. Covered by the new test "keeps the skeleton when a cached refresh throws after a network switch" (cached ETH fetch rejects after switching to an uncached COSMOS load; skeleton stays until COSMOS lands).

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
weatherstar: Fixed in ec50bdeb0a: the per-key load version now gates only the shared cache write (`isLatestVersion()`), while `isStale()` is back to the instance-level load id. A mounted older instance therefore still paints its own answer even when a newer instance unmounted mid-flight, and it still cannot overwrite the cache with an older result. Covered by the new test "still paints an older mounted instance after a newer instance is abandoned".

packages/kit/src/components/AddressInput/index.tsx
weatherstar: Narrowed in ec50bdeb0a: the spinner now only appears when the previous result actually had dynamic content that gets hidden during re-validation (a resolved name address or server-derived label badges). For a plain address that only carries wallet / address-book labels — the common case on focus regain — the labels stay as-is and no spinner is rendered.

The remaining spinner for ENS / badged results is intentional: it was requested in the earlier review thread on this file (a changed ENS record must not keep showing the old resolution during re-validation). Leaving this thread open for the maintainer to decide whether that residual spinner is acceptable.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -327,29 +421,37 @@ export function useRecentRecipientsData({
...
-        if (isStale()) return;
-        setRecentRecipients(enriched);
       } catch {
         // ignore enrichment errors
       }
     }
 
     if (isStale()) return;
-    setIsLoadingRecent(false);
+    commit({ recipients: storedRecipients, apiUnsupported });
```
sidmorizon, line 438: 🟡 P2: Failed refresh erases cached recipients

Severity: `non-severe`

When an unsupported network's cached local refresh fails, `commit` replaces its recipients with an empty list. [loadStoredRecipients](packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts:186-234) converts read failures to empty data, while enrichment failures leave `storedRecipients` empty. The visible recents disappear, and the shared cache loses the last successful list.

originalix (approved):

PR #13453, opened 2026-09-15
fix: align market banner detail lists with home lists and polish mobile tabs

erikzou:
Market banner detail lists (the pages opened from the header banner cards) now follow the home lists on every platform, plus mobile list and tab polish. Everything is scoped to Market.

## Banner detail (desktop)

- The page uses the home desktop content frame (1440px max) instead of full width. The title and tab row share the home inset, the tab row marks the active tab by label weight with the default cursor, and the table sits 12px under the tabs.
- Each banner type takes the columns of its home tab, with `--` for missing values:
  - Stock banners use the Stocks tab columns.
  - Spot banners use the Trending columns without token age. The first column is titled `Name` and shows the token name, revealing the contract address on hover like Favorites. Trending itself is unchanged.
  - Perps banners use the Perps tab columns.
- Rows are 72px with a 36px header, columns hide responsively like the home tables, and sorting runs in memory.
- For stock mixed banners the spot tab reads `Stocks`.

## Banner detail (mobile)

- The lists render the home mobile rows for stocks, spot tokens and perps under a `Name / Volume` header, keeping the change sort. The stock change column reads `Change`, not 24h, matching the Stocks tab.
- The header title uses the default navigation title size (18px) everywhere except the wide web/Electron title row, which keeps 24px.

## Banner types

- `ticker_perps` banners show a spot and a perps tab, like stock perps banners.

## Mobile home lists (web and native)

- Stock rows show the company name and volume on the second line, in the Stocks tab, Favorites and banner detail.
- Perps rows use the shared row frame (72px height, 4px line gap, same padding) in the Perps tab, Favorites and banner detail. The Perps list no longer sits 4px lower than the other tabs.
- Category chips on the Favorites, Perps and Stocks (non-native) tabs match the native Stocks sub-header chips: pill shape, 44px minimum width.
- The category row has 12px above and below, and tabs without secondary controls (such as Top coins) get 12px under the tab bar.
- Native tab labels start on the same 20pt inset as the chips and column headers.

## Backend

`/utility/v2/market/banner/stock-token-list` does not return `variants`, so stock rows had no variant hover. The banner detail merges `/utility/v1/stocks/batch` by `stockId` (added to the travel mode network allowlist) and falls back to the banner rows if the batch call fails.

## Test plan

- [x] Jest: Market (143 suites)
- [ ] Desktop web: banner detail frame, tab row, Stocks / Trending / Perps columns, `Name` column hover address, sorting, `ticker_perps` tabs
- [ ] Mobile web: banner detail rows and change sort, stock and perps rows, chip shape, spacing under tabs
- [ ] iOS / Android: banner detail title size, stock and perps rows, chip shape vs native Stocks chips, tab label inset


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-bb4b058e-0308-5adf-8925-5667ca9de4cc)

sidmorizon:
Claude review session: https://claude.ai/code/session_01NinNdThFiWwxk37pr2K4tw

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/MarketTokenListBase.tsx
sidmorizon, line 1: 🟡 P2: Banner favorites report homepage source

Reported at `packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/MarketTokenListBase.tsx:546-553` (outside this PR's diff; file-level comment).

Severity: `non-severe`

When a desktop banner favorite changes, `useTrendingColumnsDesktop` receives no `watchlistFrom`. Its star hardcodes `Homepage`, so banner actions are attributed incorrectly.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
```diff
@@ -55,9 +59,44 @@ function mapStockBannerItemToToken(
...
+// renders.
+async function enrichBannerStockRows(
+  bannerRows: IMarketStockPublicItem[],
+): Promise<IMarketStockPublicItem[]> {
+  if (bannerRows.length === 0) {
+    return bannerRows;
+  }
+  try {
+    const batchRows =
+      await backgroundApiProxy.serviceMarketV2.fetchMarketStockBatch({
```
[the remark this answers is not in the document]
erikzou: Confirmed. With travel mode active, `shouldRejectTravelModeServiceCall` routes unlisted methods to the control-plane-only `commands.run`, so the call never reached the allowlisted endpoint. Fixed in 1b61e4b46c: `fetchMarketStockBatch` is now in the travel-mode `serviceMarketV2` allowlist, with the matching policy test entry. The Favorites stock rows, which call the same method, are covered too.

originalix (dismissed):

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/MarketTokenListBase.tsx
erikzou: Confirmed, and this PR introduced it: banner detail now renders ticker rows with the trending desktop columns, whose star hardcoded `EWatchlistFrom.Homepage`, while the previous banner columns used the passed `watchlistFrom`. Fixed in 264f16884e: `useTrendingColumnsDesktop` takes `watchlistFrom` (default `Homepage`, so Trending is unchanged) and `MarketTokenListBase` passes it through, so banner detail logs `BannerList`. Added tests for both the default and the banner source.

packages/kit/src/views/Market/MarketBannerDetail/useMarketBannerDetail/index.ts
```diff
@@ -55,9 +59,44 @@ function mapStockBannerItemToToken(
...
+        stockIds: bannerRows.map((row) => row.stockId),
+      });
+    const batchRowsByStockId = new Map(
+      batchRows.map((row) => [row.stockId, row] as const),
```
originalix, line 89: P2: [Case-sensitive stock batch merge drops enriched data]

When the banner endpoint returns a stock ID with different casing from `/utility/v1/stocks/batch` (for example, `aapl` versus `AAPL`), this raw-keyed map cannot find the batch row. The banner therefore keeps the un-enriched row and loses batch-only data such as `variants`.

Normalize both the map keys and the lookup key with the same canonical stock-ID form, as the watchlist batch lookup does, before merging the rows.

originalix (approved):

PR #13460, opened 2026-09-15
fix: keep iPad onboarding and wallet sheet layouts in sync with breakpoints (OK-63174 OK-62661 OK-62660 OK-62659)

huhuanming:
[OK-63174](<https://onekeyhq.atlassian.net/browse/OK-63174>) [OK-62661](<https://onekeyhq.atlassian.net/browse/OK-62661>) [OK-62660](<https://onekeyhq.atlassian.net/browse/OK-62660>) [OK-62659](<https://onekeyhq.atlassian.net/browse/OK-62659>)

----------

## Summary

Fixes four iPad layout issues caused by stale or mis-measured breakpoints.

- **OK-62661 / OK-62659: first onboarding screen kept phone breakpoints.**
  - The iPad media driver clamps tablets to phone widths for the split-view pane. Full-screen onboarding holds the real width, but screens that mounted in the same commit as the hold missed Tamagui's refresh, because native `useMedia` subscribes in a passive effect.
  - Landscape GetStarted rendered stretched phone buttons, and a portrait device picker opened as the first screen rendered a single-column list with left-aligned names.
  - `StandardOnboardingNavigator` now mounts the stack only after the hold is in place.
  - `TableSplitViewContainer` hides the main pane in the same commit via a synchronous held-state store, instead of waiting for `isOnBoardingOpenAtom`.
- **OK-62660: connect pages went stale after a root modal.**
  - Opening a root modal over onboarding (for example the QR scanner) blurs onboarding. That releases the hold and switches breakpoints to md; closing the modal switches them back.
  - Already-mounted styled components kept the md styles, leaving half-phone, half-wide pages. For example, the QR connection card was 800pt wide inside the wide layout.
  - Root cause is in Tamagui 2.7.7's native `useMedia`:
    - `createComponent` stores per-component media-listening state (`setMediaShouldUpdate`) on the component context shared by the whole subtree.
    - Once any animated component registers `mediaEmit`, every styled component takes the emit path. That path only records `pendingState` and skips the re-render.
    - Upstream `main` still has this code.
  - `patches/@tamagui+web+2.7.7.patch` now does three things, in native builds only (`useMedia.native.js`, `createComponent.native.js`, esm and cjs), and only when the media driver reports a tablet-like device:
    - keys listening state per component (its `useMedia` proxy) instead of the shared component context;
    - skips the emit shortcut so non-animated components re-render;
    - re-checks the snapshot once after subscribing.
  - `patches/@tamagui+react-native-media-driver+2.7.7.patch` exposes that `isTabletLike` flag (tablet or dual-screen) on the existing `globalThis.$$onekeyNativeMedia` control object.
  - `createComponent` writes the listening state under both keys, so each mode reads a current entry.
- **OK-63174: wallet "More" sheet changed width.**
  - The iPad ActionList sheet sizes to its content (400–480pt).
  - The referral item's loading skeleton was an in-flow `width: 100%` element, so the sheet measured 480pt and then shrank once the item loaded.
  - When the sheet is content-sized (`gtMd || platformEnv.isNativeIOSPad`), the skeleton fill is now absolutely positioned, so it no longer takes part in intrinsic width. Full-width sheets keep the in-flow placeholder.

## Scope

Every change is limited to tablet-like devices (iPad and dual-screen foldables); phones keep their current behavior.

- `Navigator.tsx`, `nativeTabletRealWidthMedia`, `TableSplitViewContainer`: the hold helpers return early unless `platformEnv.isNativeIOSPad`, the gate's initial value is already `true` on phones, and the held state is always `false` there, so the new split-view condition does not change its value.
- ActionList: the absolute placeholder applies only when the sheet is content-sized (`gtMd || platformEnv.isNativeIOSPad`, the same condition `Popover`'s sheet frame uses). A phone in portrait keeps the previous in-flow placeholder markup.
- Tamagui patch: the media driver now exposes `isTabletLike` (tablet or dual-screen), and `useMedia` applies the three fixes only when that flag is set. Phones follow upstream code paths, including the `mediaEmit` shortcut and the shared component-context state entry.

Known gap this scoping leaves open: an iPhone can rotate (`Info.plist` allows it) and, because non-tablet iOS is not width-clamped, landscape crosses the 768 breakpoint. The same upstream Tamagui bug can therefore leave stale breakpoints on an iPhone after rotation. That is out of scope here and is not a regression from this PR.

## Verification status

All simulator work was done on an iPad Air 11" (M3), iOS 26.5, through the Mobile DevSession launcher.

Re-verified on the scoped build (second commit):

- OK-63174: the wallet "More" sheet opens at 360pt and does not change width for the rest of the recording.
- OK-62661: landscape GetStarted renders three wide cards with the split-view pane hidden.
- OK-62660: landscape QR round trip (Connect with QR Code → Scan QR code → camera-denied sheet → Cancel → close the scanner). The card measured 399pt before the round trip and 399pt after it, and the page came back in the wide two-column layout with the title centered. On `x` the same round trip takes the card from 399pt to 799pt.

Verified before the scoping commit and unchanged by it, because on iPad the scoped gate is true and the code path is the same:

- OK-62659: portrait device picker opened as the first onboarding screen renders five columns with names and tags centered.

Phones:

- iPhone 17 Pro simulator (iOS 26.5) running this branch renders the phone layout (full-width stacked buttons, icon-only language button). Deeper interaction was not exercised, so this is a smoke check rather than a full pass.
- The scoping keeps phone code paths on upstream behavior, so no phone-visible change is expected from this PR.

Regression caught while scoping:

- A first attempt gated the ActionList placeholder on `gtMd` alone. iPad portrait is clamped to 767, so `gtMd` is false there and OK-63174 came back (440pt, then 360pt). Gating on the sheet's own condition restores a constant 360pt.

## Test plan

iPad Air 11" (M3) simulator, iOS 26.5, Mobile DevSession. Before = `x`, after = this branch.

- OK-62661, landscape: account selector → + Wallet → GetStarted.
  - Before: three stretched phone-style buttons across the full width, with the split-view pane still visible.
  - After: three wide cards side by side (about 250pt each), and the split-view pane hidden.
- OK-62659, portrait: More → Connect hardware wallet → Ready to go → device picker as the first screen.
  - Before: single-column phone list, names left-aligned.
  - After: 5 columns, names and tags centered.
- OK-62660, landscape: device picker → OneKey Pro series → ConnectYourDevice → Advanced → Continue with QR Code → Scan QR code → cancel → close the scanner.
  - Before: the card measured 399pt on open and 799pt after one round trip, with the header still in wide layout.
  - After: 399pt after 5 round trips across 4 app launches, the last one on the scoped build.
  - ConnectYourDevice underneath stayed 400pt, and the Bluetooth permission dialog over it rendered correctly.
  - Confirmed that the main bundle served by Metro contains the patched `useMedia`.
  - Not covered: dialogs that need a real Bluetooth or Ledger device.
- OK-63174, portrait: wallet details → More, recorded and measured per frame.
  - Before: 440pt, then 360pt about 0.55s later.
  - After: 360pt when the sheet appears, unchanged for the rest of the recording, in two runs including one after a cold app relaunch.
  - Caveat: in both runs on this branch the referral row resolved to its loaded state quickly, so the recording cannot prove a skeleton phase occurred; what it shows is that the width never changed after the sheet appeared.
  - Re-verified after scoping. A first attempt that gated the placeholder on `gtMd` alone brought the 440pt → 360pt shrink back on iPad portrait, because the width there is clamped to 767 and `gtMd` is false; gating on the sheet's own condition restores a constant 360pt.
- `yarn jest --runTestsByPath packages/components/src/hooks/nativeTabletRealWidthMediaHeld.test.tsx packages/kit/src/provider/Container/TableSplitViewContainer/TableSplitViewContainer.test.tsx`: 10 passed.
- Patch generated with patch-package: the existing `@tamagui/web` hunks are preserved, and a reverse-apply check passes against node_modules.
- `yarn agent:check --profile commit`: 7 checks passed, exit 0.
- `yarn agent:check --profile pr`: passed, exit 0 (lint-worktree skipped because there are no uncommitted JS/TS changes; GitHub checks skipped because the PR did not exist yet).

Issues:
- https://onekeyhq.atlassian.net/browse/OK-63174
- https://onekeyhq.atlassian.net/browse/OK-62661
- https://onekeyhq.atlassian.net/browse/OK-62660
- https://onekeyhq.atlassian.net/browse/OK-62659

[OK-63174]: https://onekeyhq.atlassian.net/browse/OK-63174?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62661]: https://onekeyhq.atlassian.net/browse/OK-62661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62660]: https://onekeyhq.atlassian.net/browse/OK-62660?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-62659]: https://onekeyhq.atlassian.net/browse/OK-62659?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01PL8V679bBGdwYsfCDu2koR

originalix (approved):

PR #13461, opened 2026-09-16
fix: re-attempt SPLIT_BUNDLE_NOT_FOUND once before caching it(OK-59572)

huhuanming:
[OK-59572](<https://onekeyhq.atlassian.net/browse/OK-59572>)

----------

## Problem

A user hit a blank tab right after an APK upgrade (6.3.0 → 6.5.0), recoverable only by restarting the app. The native loader reported a segment as missing:

```
[SplitBundle] SEGMENT LOAD FAILED code=SPLIT_BUNDLE_NOT_FOUND:
  seg:nm.@onekeyfe: Segment file not found: segments/nm._onekeyfe.seg.hbc
[SplitBundle] seg:nm.@onekeyfe: previously failed: ...
```

The file was on disk and complete the whole time. Android extracts builtin segments out of the APK lazily, and the main and background runtimes resolve the same segment independently; on native builds that share one `"<name>.tmp"` per segment, the thread that loses the rename reports "not found" for a file the winner just published. Full root cause and the native fix: https://github.com/OneKeyHQ/app-modules/pull/108.

`SPLIT_BUNDLE_NOT_FOUND` was classified fatal, so `loadSegmentInternal` wrote it straight into `failedSegments` and `isRetryableLazyError` refused to retry — a millisecond-wide race poisoned the route for the whole process.

## Change

`SPLIT_BUNDLE_NOT_FOUND` moves into the existing transient tier, `RETRYABLE_NATIVE_REJECT_CODES`, so it gets the same `MAX_RETRYABLE_ATTEMPTS = 3` circuit breaker as `NO_RUNTIME` and `TIMEOUT`. The first failures are flagged `retryable: true` and kept out of `failedSegments`, so `LazyRetryBoundary` re-mounts and lands on the file the winning writer published. Once the budget is spent it caches as before with `retryable` cleared, so the boundary goes fatal instead of looping.

An earlier revision of this PR gave it a dedicated two-attempt tier. Code review killed that, for two reasons worth recording:

- `retryableAttempts` is **one counter per segment**, shared by every transient code, while the tier picked its budget from the *current* failure's code. Any earlier retryable failure pre-spent it — after a `NO_RUNTIME`, the first `NOT_FOUND` on that segment arrived at attempt 2 of 2 and was cached on sight, reinstating the exact bug this PR exists to fix. The reverse held too: `NOT_FOUND` then `TIMEOUT` silently widened the budget to 3. Both codes come out of the same cold-start window from the same native methods, so the sequence is reachable. There is now a regression test for it.
- Two attempts is too tight on its own terms: one mount plus its single boundary retry (`MAX_LAZY_RETRIES = 1`) spends the whole budget inside one `RETRY_BACKOFF_MS` window, so if the race outlasted the backoff the route died with nothing left for a later navigation.

The shared budget removes both problems and the extra concept, and restores the invariant documented on `MAX_LAZY_RETRIES`.

`SPLIT_BUNDLE_NOT_FOUND` is also added to `lazyLoad`'s own `RETRYABLE_CODES`. A caller that joins an **in-flight** load receives the raw native rejection, with no `retryable` flag for the boundary to read, so that set is what decides its fate — leaving the code out made such a caller go fatal on a segment the loader still considered re-attemptable. The exhausted-budget path is unaffected: `retryable === false` stays authoritative over the code.

Cost on a genuinely missing segment (real packaging/OTA corruption, e.g. iOS bg `file-not-found`): two extra native calls before going fatal. Not free — each re-enters the loader and can queue behind in-flight extractions — but bounded, and the route is dead either way. Every later navigation is served from the cache.

This is the half of the fix that reaches devices already in the field: OTA bundles run on whatever native build is installed, so the native fix only helps after the next APK ships.

### Deliberately not included

- `SPLIT_BUNDLE_SHA256_MISMATCH` stays fatal — a retry re-reads the same file and reproduces the same mismatch unless native deletes it first.
- The dependency-propagation guard can be bypassed when a parent joins an in-flight dep load: it then receives the raw native error, wraps it under its **own** `segmentKey`, and charges its own budget. That is pre-existing (it already applied to `NO_RUNTIME` / `TIMEOUT`) and is not made worse here now that `NOT_FOUND` shares the same budget — worth its own fix, not this one.

## Verification

- `installProdBundleLoader.test.ts` — all pass, including a new regression test that an earlier `NO_RUNTIME` does not pre-spend the budget for a later `NOT_FOUND`.
- `packages/shared/src/lazyLoad` — all pass; the old assertion that `NOT_FOUND` is non-retryable is replaced by one asserting it is retryable, plus one pinning that a budget-exhausted `NOT_FOUND` still goes fatal.
- 59 tests green across both suites.
- ESLint clean on all four changed files (the pre-existing `react/destructuring-assignment` errors in `lazyLoad` are identical before and after this change); `tsc --noEmit -p tsconfig.json` reports no errors in them.

[OK-59572]: https://onekeyhq.atlassian.net/browse/OK-59572?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w

apps/mobile/src/splitBundle/installProdBundleLoader.ts
```diff
@@ -395,7 +431,8 @@ async function loadSegmentInternal(segmentKey: string): Promise<void> {
       nativeCode = getNativeRejectCode(error);
       isRetryable =
         nativeCode !== undefined &&
-        RETRYABLE_NATIVE_REJECT_CODES.has(nativeCode);
+        (RETRYABLE_NATIVE_REJECT_CODES.has(nativeCode) ||
+          RETRY_ONCE_NATIVE_REJECT_CODES.has(nativeCode));
```
sidmorizon: 🟠 P1: Concurrent missing-segment callers skip recovery

Severity: `severe`

Concurrent `SPLIT_BUNDLE_NOT_FOUND` callers do not all receive `retryable: true`. `inflightSegments` returns the raw native promise to waiters, bypassing the owner's classification. [`isRetryableLazyError`](packages/shared/src/lazyLoad/index.tsx:56-66) rejects that raw code, so those lazy boundaries fail without the recovery attempt.

apps/mobile/src/splitBundle/installProdBundleLoader.ts
```diff
@@ -411,16 +448,17 @@ async function loadSegmentInternal(segmentKey: string): Promise<void> {
           );
 
     const attempts = (retryableAttempts.get(segmentKey) ?? 0) + 1;
+    const maxAttempts = maxAttemptsForCode(nativeCode);
```
sidmorizon: # 🟠 P1 The one re-attempt is silently lost when any retryable code precedes NOT_FOUND

## Problem

`attempts` (line 450) is a single per-segment counter shared by **every** reject code, but `maxAttempts` (line 451) is derived from the **latest** code only. When a segment first fails with a code from `RETRYABLE_NATIVE_REJECT_CODES` and *then* hits `SPLIT_BUNDLE_NOT_FOUND`, the counter is already at 1, so the NOT_FOUND evaluates `attempts=2 < maxAttempts=2` → false and falls straight through to the permanent-cache branch (lines 472-474) with **zero** re-attempts.

Tracing the exact bookkeeping of lines 450-474:

```
TIMEOUT , NOT_FOUND , NOT_FOUND
  call 1: SPLIT_BUNDLE_TIMEOUT   -> retry branch, retryableAttempts=1, retryable=true
  call 2: SPLIT_BUNDLE_NOT_FOUND -> attempts=2, maxAttempts=2 -> FATAL, cached, retryable=false
  call 3: served from failedSegments, native never called again
```

`SPLIT_BUNDLE_NO_RUNTIME , NOT_FOUND` behaves identically. The homogeneous sequences the two new tests cover (`installProdBundleLoader.test.ts:457`, `:487`) both use a single code for every attempt, so neither can catch this.

## Impact

Android, `main` and `bg` runtimes alike (these module-level maps are per-runtime, so each carries the same defect independently). The precondition is the PR's own scenario, not an exotic one: on the first launch after an APK replace the install-stamp wipe forces every segment to re-extract at once, so the largest segment is also the one most likely to trip the Android watchdog (`SPLIT_BUNDLE_TIMEOUT`, per the comment at line 84) before the `<name>.tmp` rename loser reports `SPLIT_BUNDLE_NOT_FOUND`. The same shape appears when an idle preload burns attempt 1 with `SPLIT_BUNDLE_NO_RUNTIME` (`packages/kit/src/provider/preloadComponents.ts` swallows the error and never retries) and the user's later navigation lands on the NOT_FOUND. In those runs the route is permanently cached as failed for the whole process with `retryable=false`, the boundary goes fatal on first sight, and the fix delivers nothing — the exact blank-tab symptom the PR is meant to remove.

## Recommendation

Make the counter code-aware so a code change restarts its own tier's budget while an overall cap still bounds the total. Change `retryableAttempts` (line 108) to `Map<string, { code?: string; attempts: number; total: number }>` and rewrite lines 450-457 as:

```ts
const prev = retryableAttempts.get(segmentKey);
const attempts = prev && prev.code === nativeCode ? prev.attempts + 1 : 1;
const total = (prev?.total ?? 0) + 1;
const maxAttempts = maxAttemptsForCode(nativeCode);

if (
  isRetryable &&
  attempts < maxAttempts &&
  total < MAX_RETRYABLE_ATTEMPTS + RETRY_ONCE_MAX_ATTEMPTS
) {
  retryableAttempts.set(segmentKey, { code: nativeCode, attempts, total });
  ...
}
```

The existing `retryableAttempts.delete` calls at lines 396, 472 and 550 already cover reset on success / exhaustion / `retrySegment()`, so no other lifecycle change is needed. Please add a `TIMEOUT → NOT_FOUND → success` case to `installProdBundleLoader.test.ts` asserting 3 native calls and `isSegmentLoaded === true`.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

apps/mobile/src/splitBundle/installProdBundleLoader.ts
sidmorizon: # 🟠 P1 Deduped concurrent callers never see the new `retryable` flag, so the boundary still goes fatal

## Problem

The PR description argues that `lazyLoad`'s `RETRYABLE_CODES` can stay untouched because "the boundary already treats the loader's explicit `retryable` flag as authoritative". That holds only for the caller that actually runs the `catch` block — it does not hold for a caller that hits the inflight-dedup path.

`loadSegmentInternal` stores the **inner IIFE promise** in `inflightSegments` (line 391) and hands it to every concurrent caller at lines 245-248. That IIFE has no `catch` — only `finally { globalLoading.delete(...) }` (lines 386-388) — and `await loader.loadSegment(...)` at line 355 sits directly inside it. So a deduped caller rejects with the **raw** RN TurboModule error, never the `SegmentLoadError` built at lines 441-448. The raw error carries `code: 'SPLIT_BUNDLE_NOT_FOUND'` but **no `retryable` property**.

`isRetryableLazyError` (`packages/shared/src/lazyLoad/index.tsx:51-78`) therefore skips both `retryable` short-circuits at lines 59-60 and falls through to `RETRYABLE_CODES` (lines 33-36), which this PR did not update — it is still exactly `{SPLIT_BUNDLE_NO_RUNTIME, SPLIT_BUNDLE_TIMEOUT}`. Before this PR the loader and the boundary agreed that NOT_FOUND was fatal; they now disagree on the raw-error path. `packages/shared/src/lazyLoad/index.test.tsx:121-128` still asserts `isRetryableLazyError({ code: 'SPLIT_BUNDLE_NOT_FOUND' }) === false`, and the doc comment at `index.tsx:39-42` still lists NOT_FOUND as non-retryable.

## Impact

Android/iOS `main` runtime. Whenever two callers race the same segment — two lazy components from one segment mounting in the same tick, an idle preload overlapping a mount, or a segment required both directly and as a `dependsOn` (lines 344-348) — the deduped caller's `LazyRetryBoundary.componentDidCatch` (`index.tsx:117-124`) sees `isRetryableLazyError === false` and jumps straight to `phase: 'fatal'`, rethrowing to the root boundary without spending its single retry. `packages/kit/src/provider/Container/DeviceStageContainer/Lazy.tsx:175-182` does the same and cancels the stranded hardware burst. Meanwhile the loader has **not** cached the failure and would have served the segment on the very next call — so the user gets the fatal screen this PR exists to prevent, on a segment that was already healthy. Startup right after an APK replace is precisely when both the extract race and concurrent same-segment loads are most likely, so the two conditions correlate rather than cancel out.

## Recommendation

Mirror the new tier in the boundary's raw-code set: add `'SPLIT_BUNDLE_NOT_FOUND'` to `RETRYABLE_CODES` at `packages/shared/src/lazyLoad/index.tsx:33-36`, refresh the stale comment at lines 39-42, and update the expectation at `packages/shared/src/lazyLoad/index.test.tsx:121-128`. This does not widen the blast radius the way the description fears, because the authoritative `e.retryable === false` check at `index.tsx:59` still runs first — the budget-exhausted cached `SegmentLoadError` (line 473) stays fatal-on-sight.

Better still, close the leak at its source so the two paths cannot drift again: wrap the promise stored at line 391 (e.g. `inflightSegments.set(segmentKey, promise.catch(...))` resolving to the classified `SegmentLoadError`) so every caller, deduped or not, receives the same classified error object.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
```diff
@@ -448,24 +448,101 @@ describe('installProdBundleLoader', () => {
...
   });
 
-  // Fix 1 (NO-SHIP blocker): a missing bg segment file is real packaging/OTA
-  // corruption. iOS BackgroundThread now maps EBgMgrSegmentEvalErrorFileNotFound
-  // → SPLIT_BUNDLE_NOT_FOUND (fatal), instead of letting raw code 2 fall through
-  // the boundary's default to retryable NO_RUNTIME. JS MUST classify it FATAL —
-  // cached immediately, never auto re-attempted (retrying just re-misses).
-  it('treats SPLIT_BUNDLE_NOT_FOUND as FATAL (iOS bg file-not-found)', async () => {
+  // SPLIT_BUNDLE_NOT_FOUND gets exactly ONE re-attempt
+  // (RETRY_ONCE_NATIVE_REJECT_CODES) rather than being cached on sight. The
```
sidmorizon: # 🟡 P2 Test comments still describe the deleted two-attempt tier and overstate the new guarantee

## Problem

The redesign removed `RETRY_ONCE_NATIVE_REJECT_CODES` / `RETRY_ONCE_MAX_ATTEMPTS` from the loader, but two comment blocks in this file were not brought along.

**Lines 451-452 and 457.** This block still opens with `SPLIT_BUNDLE_NOT_FOUND gets exactly ONE re-attempt (RETRY_ONCE_NATIVE_REJECT_CODES) rather than being cached on sight`, and line 457 still says "the re-attempt" in the singular. A repo-wide grep for `RETRY_ONCE` now matches **only line 452** — the symbol exists nowhere else, so this is the sole surviving reference to a design that was deleted. "Exactly ONE re-attempt" is also factually wrong under the new code: `installProdBundleLoader.ts:436` grants `MAX_RETRYABLE_ATTEMPTS = 3` total, i.e. two re-attempts, which is exactly what the sibling test at line 507 asserts (`toHaveBeenCalledTimes(3)`). The two descriptions of the budget now contradict each other about 50 lines apart in the same file.

**Lines 513-518.** The new regression test's comment states that an earlier `NO_RUNTIME` "must not pre-spend the budget such that the FIRST NOT_FOUND on that segment is already past the cap and gets cached on sight". `retryableAttempts` is a single per-segment counter (`installProdBundleLoader.ts:126`, read at `:434`), so an earlier transient reject *does* pre-spend it — the property only holds for **one** prior failure. With two (`NO_RUNTIME`, `TIMEOUT`, then `NOT_FOUND`) the counter reaches 3, `3 < 3` is false at `:436`, and the first `NOT_FOUND` on that segment is cached on sight at `:455-457` with zero re-attempts. The test only exercises one prior failure, so nothing pins the stronger claim.

## Impact

Maintainability only — no runtime behaviour is affected, which is why this is P2 rather than higher. But the dangling `RETRY_ONCE_NATIVE_REJECT_CODES` sits at the top of the very test that is supposed to document the replacement design, so the next reader chasing that symbol finds a ghost and two conflicting budget numbers. The 513-518 comment is the more consequential one: it reads as a guarantee about `NOT_FOUND` specifically, which invites a later change to lean on it (dropping the boundary retry, or assuming `NOT_FOUND` always self-heals) when the real contract is the weaker, uniform one.

## Recommendation

Rewrite lines 451-459 to match `installProdBundleLoader.ts:86-105`: drop the `RETRY_ONCE_NATIVE_REJECT_CODES` mention, say that `NOT_FOUND` now shares `RETRYABLE_NATIVE_REJECT_CODES` and the `MAX_RETRYABLE_ATTEMPTS` circuit breaker, and change "the re-attempt" on line 457 to "a re-attempt".

For lines 513-518, state the contract the code actually provides — `retryableAttempts` is one per-segment circuit breaker shared by every transient code, giving a segment 3 attempts per process regardless of which codes produced them — and note that what this test pins is that the counter and the cap are no longer read from different sources. Renaming it to something like `does not let one earlier NO_RUNTIME cache a later NOT_FOUND on sight` would keep the title true to what it exercises.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

apps/mobile/src/splitBundle/installProdBundleLoader.ts
huhuanming: Fixed in `dc6c5ace5e` — this was written against `330d5db208`, before that commit.

`SPLIT_BUNDLE_NOT_FOUND` is now in `lazyLoad`'s `RETRYABLE_CODES`, so the deduped caller that receives the raw native rejection — with no `retryable` property for the boundary to read — is classified the same way as the owner. The doc comment above it is updated, and `index.test.tsx` now asserts the new behaviour plus a new case pinning that a budget-exhausted `NOT_FOUND` (`retryable === false`) still goes fatal, since that check runs first and stays authoritative.

apps/mobile/src/splitBundle/installProdBundleLoader.ts
huhuanming: Fixed in `dc6c5ace5e`, though not the way you suggested — the per-code budget is gone entirely rather than made code-aware.

Your trace was right, and this was the more serious of the two findings: the counter was shared while the cap was derived from the *current* failure's code, so one prior `NO_RUNTIME`/`TIMEOUT` put the first `NOT_FOUND` at `attempts=2` against `maxAttempts=2` and cached it on sight with zero re-attempts — the exact bug the PR exists to remove.

Rather than adding per-tier bookkeeping, `NOT_FOUND` moved into `RETRYABLE_NATIVE_REJECT_CODES` and shares `MAX_RETRYABLE_ATTEMPTS`. That removes the two-sources-of-truth problem instead of synchronising it, drops a whole concept, and restores the invariant documented on `MAX_LAZY_RETRIES` (one mount plus its single boundary retry spends 2 of 3, so it can never exhaust the budget by itself). There is now a regression test for the `NO_RUNTIME` → `NOT_FOUND` sequence.

Worth being explicit about the resulting contract, since it is weaker than "one prior failure can never hurt": it is uniform, not `NOT_FOUND`-specific — 3 attempts per segment per process regardless of which codes produced them, so **two** prior failures do exhaust it. That bound is intended. `6cb0eaba2` corrects a test comment of mine that had overstated it.

apps/mobile/src/splitBundle/installProdBundleLoader.ts
huhuanming: The primary recommendation is done in `dc6c5ace5e`: `SPLIT_BUNDLE_NOT_FOUND` added to `RETRYABLE_CODES`, the stale doc comment refreshed, and the `index.test.tsx` expectation replaced — plus a new case pinning that `retryable === false` still wins, so the budget-exhausted path stays fatal-on-sight exactly as you noted.

The "better still" — wrapping the promise stored in `inflightSegments` so every caller, deduped or not, receives the classified `SegmentLoadError` — I deliberately left out of this PR. It is the better fix and I would like it done; the reason for deferring is risk shape rather than disagreement. Settling a separate public promise correctly means covering all three throw sites in that catch block, the dep-propagation early throw included, and a missed rejection there does not fail deduped callers — it **hangs** them. Restructuring so the classification lives inside the IIFE avoids that but changes what every caller of a startup-critical path receives. Either version wants its own tests and its own review, not a ride-along on a hotfix that is holding up a release.

Happy to open that as a follow-up PR.

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
huhuanming: Fixed in `6cb0eaba2`. Both blocks were mine and both were wrong in the way you describe.

The `RETRY_ONCE_NATIVE_REJECT_CODES` reference was indeed the last one anywhere — a repo-wide grep now returns nothing. That block now says `NOT_FOUND` shares `RETRYABLE_NATIVE_REJECT_CODES` and the `MAX_RETRYABLE_ATTEMPTS` circuit breaker, and "the re-attempt" is now "a re-attempt".

The second one is the one worth catching: it read as a `NOT_FOUND`-specific guarantee the code does not provide. It now states the actual uniform contract — one shared per-segment circuit breaker, `MAX_RETRYABLE_ATTEMPTS` per process regardless of which codes produced them, so two prior failures **do** exhaust it and a third-arriving `NOT_FOUND` is cached on sight, which is the intended bound — and then says what the test actually pins: that the counter and the cap are no longer read from different sources. Renamed to `does not let one earlier NO_RUNTIME cache a later NOT_FOUND on sight`.

Comments only; 35 tests still pass.

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
```diff
@@ -448,26 +448,155 @@ describe('installProdBundleLoader', () => {
...
+        }),
+    );
     const { installProdBundleLoader, loadSegment } = getLoader();
     installProdBundleLoader(mock);
-    await expect(loadSegment('seg:test.a')).rejects.toThrow();
-    // Fatal → cached; second call must NOT hit native again.
-    await expect(loadSegment('seg:test.a')).rejects.toThrow();
+
+    // loadSegmentInternal registers the inflight entry synchronously, before
+    // it awaits anything, so the second call dedups onto the first with no
```
sidmorizon: # 🟡 P2 `dedups` in this comment fails `lint (24.x)` and is currently blocking auto-merge

## Problem

The word `dedups` on this line is not in the project dictionary, and the repo's lint gate treats a single cspell warning as a job failure. `lint (24.x)` on this PR's head is red with exactly this and nothing else:

```
[3:50:19 AM] Lint check started...
[3:53:11 AM] ##[warning]Unknown word: "dedups"
             Found 1 warning and 0 errors.
[3:53:11 AM] Oxlint check failed. (171.33s)
```

TypeScript passed cleanly (103.54s) in the same job — this is the only thing failing it.

Why this one occurrence and not the neighbouring ones: `.oxlintrc.json` configures `@cspell/spellchecker` with `checkComments: true` and `checkStrings: false`. `deduped` at line 489 and the pre-existing `dedups` at `packages/kit/src/views/Home/components/TokenListBlock/buildMergedAllNetworkSnapshot.test.ts:114` both sit inside `it(...)` title **strings**, so they are never checked. This line is a comment, so it is. `development/spellCheckerSkipWords.txt` allows `dedup`, `Dedup`, `Deduplicate`, `Deduplication` and `deduplication` — but not `dedups`. The `/* cspell:ignore dedup */` directive at `installProdBundleLoader.ts:1` is file-scoped and does not reach this test file.

## Impact

CI-blocking. Auto-merge was enabled on this PR with squash, so the PR sits unmergeable until this clears — on a change the description says is holding up a release. It is also the kind of failure that reads as noise and invites a re-run, when a re-run cannot help: the input is deterministic.

## Recommendation

Cheapest correct fix is to reword the comment into vocabulary the dictionary already allows, e.g. `so the second call is deduplicated onto the first with no tick in between`, or simply `so the second call joins the first with no tick in between`.

If you would rather keep the verb, add `dedups` (and `deduped`, which is only passing today because it happens to sit in a string) to `development/spellCheckerSkipWords.txt` next to the existing `dedup` entries at lines 312-316 — that also covers the pre-existing occurrence in `buildMergedAllNetworkSnapshot.test.ts` should it ever move out of a string.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

packages/shared/src/lazyLoad/index.tsx
```diff
@@ -30,16 +30,25 @@ import {
 // retry lands on.
 export const MAX_LAZY_RETRIES = 1;
 const RETRY_BACKOFF_MS = [150, 600];
+// Kept in sync with RETRYABLE_NATIVE_REJECT_CODES in
+// apps/mobile/src/splitBundle/installProdBundleLoader.ts. This lookup is the
+// fallback for errors that reach the boundary WITHOUT the loader's `retryable`
+// flag — notably a caller that joined an in-flight load and so receives the raw
+// native rejection rather than the wrapped SegmentLoadError. Leaving a code out
+// here makes that caller go fatal on a failure the loader still considers
+// re-attemptable.
```
sidmorizon: # 🟡 P2 This comment documents the exact coupling `d0ac6b0f` was written to delete

## Problem

`d0ac6b0f` publishes the classified outcome to in-flight joiners, and its own comment at `installProdBundleLoader.ts:482-488` says why: *"Keeping those two code sets in sync is a standing drift hazard; handing both callers the same object removes the hazard instead of managing it."*

This comment block still instructs the reader to manage it. It names the producer that no longer exists — *"notably a caller that joined an in-flight load and so receives the raw native rejection"* — and states an obligation that is now false: *"Leaving a code out here makes that caller go fatal on a failure the loader still considers re-attemptable."*

After that commit, every error leaving `loadSegmentInternal` is a `SegmentLoadError` carrying an explicit boolean `retryable`, because `inflightSegments` now holds `classifyLoadOutcome`'s promise (`installProdBundleLoader.ts:489-491`) rather than the raw IIFE promise. `isRetryableLazyError` short-circuits on that flag at `index.tsx:68-69`, before it ever reaches the `RETRYABLE_CODES` lookup at `:70`. I looked for a surviving in-tree producer of a raw `SPLIT_BUNDLE_*`-coded error at a lazy boundary and found none: `nativeBridge.ts` / `nativeBridgeBackground.ts` are only called from inside the loader, `healthCheck.ts` swallows its own errors, the post-load `require()` throws plain `Error`s with no `.code`, and web/desktop/extension never produce those codes (they are covered by `RETRYABLE_ERROR_NAMES` and the message heuristics instead).

## Impact

Maintainability, but with a specific and plausible failure mode that is the inverse of ordinary comment rot. A maintainer debugging a future "went fatal instead of retrying" report reads this block, concludes joiners still receive raw errors, and edits `RETRYABLE_CODES` to fix it — reintroducing precisely the two-sources-of-truth drift that `d0ac6b0f` removed, and leaving the real cause unexamined. The commit's stated rationale is only half-delivered while the other end of the contract still describes the old world.

## Recommendation

Retarget this block in the same PR. Say the set is defense-in-depth for errors that bypass `loadSegmentInternal` entirely, note that nothing in-tree produces one today, and point at `installProdBundleLoader.ts:489-491` for why the joiner case is gone. Drop both the "notably a caller that joined an in-flight load" clause and the "kept in sync with `RETRYABLE_NATIVE_REJECT_CODES` / leaving a code out makes that caller go fatal" obligation.

Keep `RETRYABLE_CODES` itself and `MAX_LAZY_RETRIES` as they are — the entry is harmless and the `retryable === false` check at `:68` still runs first, so a budget-exhausted failure stays fatal-on-sight either way.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
```diff
@@ -448,26 +448,155 @@ describe('installProdBundleLoader', () => {
...
+    expect(mock.loadSegment).toHaveBeenCalledTimes(2);
+  });
+
+  // A caller that joins an IN-FLIGHT load used to receive the RAW native
+  // rejection, which carries `code` but no `retryable` flag — so the lazy
+  // boundary classified it from its own code set while the owner read the
+  // loader's verdict, and the two sets could drift apart again. inflightSegments
+  // now publishes the classified outcome, so both callers observe the same
+  // object. Note the joiner does NOT get its own native call.
+  it('gives a deduped concurrent caller the same classified error as the owner', async () => {
```
sidmorizon, line 489: # 🟡 P2 The joiner test pins only the retryable outcome, leaving the in-place `retryable = false` mutation unguarded

## Problem

This test is correct and it does fail against the pre-`d0ac6b0f` code — a joiner used to receive the raw native `Error` (a `code`, no `retryable`, `name: 'Error'`), so both the identity assertion and the `retryable: true` matcher would fail. Its claim that the inflight entry is registered synchronously before any `await` is also accurate.

But it pins one of the three joiner outcomes the restructuring created, and the unpinned one is the more fragile half:

**Budget-exhausted joiner.** `installProdBundleLoader.ts:305` does `segError.retryable = false` — an **in-place mutation on the very object now shared with every joiner**. Nothing asserts a joiner observes `retryable: false`. A future change that reorders that assignment after the `throw`, or that clones/wraps the error before caching it in `failedSegments`, would leave joiners seeing `retryable: true` while the owner sees `false` — silently reinstating the exact owner/joiner disagreement this commit exists to remove. The user-visible result is the one the PR description calls out as the cost to avoid: `LazyRetryBoundary` burns a full retry round plus fallback spinner against a permanently-dead route on every re-navigation.

**Dep-propagation joiner.** `:242-246` is the only rejecting path where the published error's `segmentKey` differs from the key the joiner asked for. The existing dep tests at `:625` and `:862` exercise the owner only.

(The eager-fallback resolve path is effectively unreachable for a joiner — `loadedSegments.add` at `:386` runs synchronously before `inflightSegments.set`, so a same-tick second caller takes the already-loaded early return at `:321`. Not worth a test, but worth knowing when reading the settlement paths.)

## Impact

The most consequential half of this fix — that a joiner's `retryable` flag is authoritative and identical to the owner's — is regression-protected in one direction only. Given the flag is delivered by mutating a shared object rather than by constructing a fresh one per caller, that is the direction most likely to break silently under a later refactor.

## Recommendation

Add two tests beside this one, both asserting error **identity** (`joinerErr === ownerErr`) as well as the flag:

1. Reuse the shape of `caches SPLIT_BUNDLE_NOT_FOUND once the retry budget is exhausted` at `:532`; on the **third** attempt start a second `loadSegment('seg:test.a')` before rejecting, then assert both callers receive `retryable: false` and that a fourth call is served from `failedSegments` without hitting native.
2. Reuse the manifest from `does NOT poison the parent when a retryable DEP failure propagates` at `:625`; issue two concurrent `loadSegment('seg:test.b')` calls, reject the `seg:test.a` native load with `SPLIT_BUNDLE_NO_RUNTIME`, and assert both callers receive the same error with `segmentKey: 'seg:test.a'` and `retryable: true`, with `loadSegment` called once.

[Claude session ↗](https://claude.ai/code/session_01G8Ei3xzQh74hFy2UGKrK1w)

---

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
huhuanming: Fixed in `19e9c3ca82` — reworded to "so the second call joins the first with no tick in between".

Took the rewording rather than the dictionary entry, for the reason your analysis surfaced: the neighbouring `deduped` passes only because it sits inside an `it(...)` string, which `checkStrings: false` skips. Adding `dedups`/`deduped` to the skip list would make the gate's behaviour depend on whether a word happens to land in a string or a comment, which is exactly the distinction that made this confusing to begin with. One less word in the dictionary is the better trade here.

Verified rather than assumed, since "reworded it, should be fine" is how this comes back a second time. Putting the old word back reproduces the failure and the new wording is clean:

```
# control, old word restored
installProdBundleLoader.test.ts:502:47: warning @cspell(spellchecker): Unknown word: "dedups"
# current
(no output, exit 0)
```

packages/shared/src/lazyLoad/index.tsx
huhuanming: Fixed in `19e9c3ca82`. You are right, and this is the more interesting of the three: the comment did not just go stale, it inverted into an instruction to rebuild the thing the previous commit removed.

Retargeted as suggested — the set is now described as a backstop for a producer that bypasses `loadSegmentInternal` entirely, with the note that nothing in-tree is one today, and the "keep in sync with `RETRYABLE_NATIVE_REJECT_CODES`" obligation dropped. I added one line beyond your recommendation, because the failure mode you describe is a maintainer reaching for this list when something arrives unclassified:

> if a segment error is reaching a boundary unclassified, fix the producer rather than adding its code here.

`RETRYABLE_CODES` and `MAX_LAZY_RETRIES` left as they are, per your note.

Thanks for actually going and checking for a surviving in-tree producer rather than assuming one — that is what made "defense in depth" the honest framing instead of "dead code".

apps/mobile/src/splitBundle/__tests__/installProdBundleLoader.test.ts
huhuanming: Both tests added in `19e9c3ca82`, built on the shapes you pointed at.

`gives a joiner the same budget-exhausted error object as the owner` — spends attempts 1 and 2, then starts a second `loadSegment` before the third rejects, and asserts both callers see `retryable: false` **and** `joinerErr === ownerErr`, with a fourth call served from `failedSegments` without hitting native.

`gives a joiner the dep failure verbatim, still keyed to the dep` — two concurrent `loadSegment('seg:test.b')`, dep `seg:test.a` rejected with `SPLIT_BUNDLE_NO_RUNTIME`, asserting both receive the same object with `segmentKey: 'seg:test.a'` and `retryable: true`, native called once.

Your framing of why this mattered is the part worth keeping: the flag is delivered by mutating an object now shared with every joiner, so a refactor that clones or wraps the error before caching it would break owner/joiner agreement without breaking any existing assertion. That asymmetry is not visible from the code, only from knowing the object is shared — which is exactly what a test should hold down.

Agreed on the eager-fallback path being unreachable for a joiner (`loadedSegments.add` runs before `inflightSegments.set`, so a same-tick second caller takes the already-loaded early return). Left untested, noted here rather than in a comment so the test file does not accumulate explanations of things that cannot happen.

38 tests pass in that file.

originalix (approved):

PR #13462, opened 2026-09-16
fix: send amount caret left in token mode, recent recipients cold-start cache (OK-63413 OK-63452)

weatherstar:
[OK-63413](<https://onekeyhq.atlassian.net/browse/OK-63413>) [OK-63452](<https://onekeyhq.atlassian.net/browse/OK-63452>)

----------

## Summary
- Keep the send amount caret on the left of the `0` placeholder in token (suffix) mode, matching fiat (prefix) mode, on native and desktop/web.
- Native: map suffix mode to `textAlign: 'left'` and fold the auto-size lib's 8 dp/pt `contentAutoWidthPadding` into the suffix gap so the digits-to-symbol spacing is unchanged.
- Desktop/web: map suffix mode to left alignment, pull the suffix back over the input's trailing caret buffer, and drop the row slack compensation that only existed for the right-aligned layout.

## Intent & Context
OK-63413 (reported 2026-09-15 with an Android recording): on the send amount page, an empty amount rendered as `0| USDC` in token mode but `$ |0` after toggling to fiat. Product wants the caret parked on the left of the placeholder in both modes so the entry point looks consistent when switching between token and fiat input. The native fix was done first; the desktop/web input was then aligned to the same behavior in a second commit.

## Root Cause
`AutoSizeInput.native.tsx` mapped the inline-suffix layout to `textAlign: 'right'` (since #10507) while the prefix layout used `'left'`. On an empty field the native caret follows the text alignment, so the suffix layout parked the caret after the placeholder while the prefix layout parked it before. `AutoSizeInput.tsx` (desktop/web) had the same `right` mapping for suffix mode. This is a `textAlign` mapping issue in the app layer, not a native-lib bug.

## Design Decisions
- **Native suffix gap**: `@onekeyfe/react-native-auto-size-input` pads the `contentAutoWidth` slot by 8 (`contentAutoWidthPadding`) on both platforms. With left alignment that padding now lands between the digits and the symbol, so `suffixMarginLeft = max(inlineSuffixGapPx - 8, 0)` keeps the visual gap identical to before. The constant is named so it can be kept in sync if the lib changes.
- **Desktop/web suffix margin**: the inline input keeps a trailing caret/measurement buffer (`max(18, 0.5 x font)` px) after the digits. The suffix `marginLeft` becomes `inlineSuffixGapPx - trailingSpacePx` (negative) to overlap that buffer, so the visible gap stays at `inlineSuffixGapPx`. This is skipped when `maxWidth` shrinks the input below its natural width, because the digits then scroll inside the box and a negative margin would overlap them.
- **Row centering**: with the trailing buffer folded into the suffix margin, the row's box equals its visible content, so the `translateX` slack compensation is set to 0 for suffix mode. Prefix mode keeps its existing offset.
- Alternatives such as changing the native lib's alignment behavior were rejected: the app already owns the alignment mapping and the fix stays local to the two `AutoSizeInput` files.

## Changes Detail
- `packages/kit/src/views/Send/components/SendAutoSizeAmountInput/AutoSizeInput.native.tsx`: suffix mode now resolves to `'left'` alignment (type narrowed to `'center' | 'left'`); adds `NATIVE_CONTENT_AUTO_WIDTH_PADDING` and computes `suffixMarginLeft` from the inline suffix gap minus that padding.
- `packages/kit/src/views/Send/components/SendAutoSizeAmountInput/AutoSizeInput.tsx`: suffix mode now resolves to `'left'` alignment; computes `desktopSuffixMarginLeftPx` from the trailing input buffer with a shrunk-input guard; row offset is 0 for suffix mode and the `right`-alignment branch is removed.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS/Android) / Desktop / Web / Extension
- **Risk Areas**: The desktop negative suffix margin depends on the measured amount width and the input's trailing buffer; if `maxWidth` shrinks the input, the guard falls back to the plain gap. The native 8 dp constant must match the lib's `contentAutoWidthPadding`.

## Test plan
- [x] Android emulator and iPhone 17 Pro simulator: empty token mode shows `|0 BTC` / `|0 USDC`, typed shows `12| BTC`; fiat mode shows `$ |0` and `¥|0`; digits-to-symbol gap unchanged.
- [x] Desktop dev app via CDP: `$|0`, `$12|`, `|0 ATOM`, `5| ATOM`; DOM rects confirm the amount + symbol group is centered.
- [ ] Extension and web: open Send, toggle token/fiat with an empty amount and confirm the caret stays left of the placeholder.
- [ ] Long amount that hits `maxWidth` on desktop: confirm the suffix does not overlap the scrolling digits.

---

# OK-63452: Send address page "Recent" tab loading ~3 s

## Summary
- Persist every successful `/transfer-recipient` answer per account + API network (bounded simpleDb entity) and paint it on a cold start while the API refresh runs in the background, extending the in-session stale-while-revalidate from #13440 across app launches.
- Share one in-flight Account-table read across concurrent cold `getAllAccounts()` callers, so enriching up to 20 recipients no longer materializes the whole table once per recipient.

## Intent & Context
OK-63270 / OK-63452 (iOS recording, 2026-09-14): the Recent tab shows a skeleton for 2.5–3.2 s after entering the address page. The tester assumed the API was fast. Measurements: desktop (CDP) and iOS simulator both complete the whole client pipeline in ~0.4–0.6 s, of which the `/transfer-recipient` service call is 200–370 ms and the post-API enrichment ~40 ms. `curl` against production shows the endpoint takes 1.2–2.4 s on the first call for an account with history and 0.4–0.7 s warm, so the wait is the server round trip. The three quick-select tabs are independent and were never coupled.

## Root Cause
- The Recent tab could only paint after the API round trip on every cold start (session cache from #13440 is in-memory only).
- `LocalDbBase.getAccountNameFromAddress` falls back to a full Account-table scan for every address no local account owns; `dbAllRecordsCache` (TTL 5 s) is filled only after the first read resolves, so 20 parallel misses each ran the full read.

## Design Decisions
- Cache the raw API answer (addresses, times, badges, `lastUsedDeriveType`) rather than the enriched rows, and re-run the local enrichment on paint (~40 ms), so wallet / address-book names are never stale. Only server-confirmed lists are stored; unsupported or failed answers are not.
- The persisted entry is treated exactly like the session cache: painted at once, replaced by the API result, kept when the refresh fails (OK-53284: never mix sources).
- The in-flight read lives in the DB layer (`dbAllRecordsInflight`, flushed by `clearStoreCachedData`) so any concurrent caller benefits; joiners receive `cloneDeep` copies like a cache hit.

## Changes Detail
- `packages/kit-bg/src/dbs/simple/entity/SimpleDbEntityTransferRecipientsCache.ts` (new, registered in `SimpleDb.ts` / `SimpleDbProxy.ts` and the mobile module-id registry).
- `packages/kit-bg/src/services/ServiceHistory.ts`: `persistTransferRecipients` after a successful fetch; new `getCachedTransferRecipients`.
- `packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts`: cold-start phase `loadPersistedApiRecipients`.
- `packages/kit-bg/src/dbs/local/LocalDbBase.ts` / `LocalDbBaseContainer.ts`: shared in-flight `getAllAccounts` read.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS/Android) / Desktop / Web / Extension
- **Risk Areas**: a stale recent list for one API round trip after a send made on another device; bounded to 30 account+network entries. `getAllAccounts` callers that mutate the returned array keep the same semantics as before (first reader gets the cached array, joiners get copies).

## Test plan
- [x] Jest: `useRecentRecipientsData.test.tsx` (+3 cases, 11 total), `LocalDbBase.getAllAccountsInflight.test.ts` (4 cases), all `packages/kit-bg/src/dbs/local` suites.
- [x] iPhone 17 Pro simulator, cold app launch: persisted list painted 77 ms after the hook starts (API commit at 503 ms); skeleton only visible during the page transition (~0.2 s) versus ~0.5 s before.
- [ ] Physical device with the reporter's wallet: confirm the Recent tab paints from cache on the second launch.
- Server-side latency of `/transfer-recipient` for history-heavy accounts remains open for the backend team.

[OK-63413]: https://onekeyhq.atlassian.net/browse/OK-63413?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-63452]: https://onekeyhq.atlassian.net/browse/OK-63452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01RP8TxGNoJpHYcYN7i2XS2T

packages/kit/src/views/Send/components/SendAutoSizeAmountInput/AutoSizeInput.tsx
```diff
@@ -313,18 +313,35 @@ export const AutoSizeInput = forwardRef<IAutoSizeInputRef, IAutoSizeInputProps>(
...
+      hasPrefix || hasSuffix ? 'left' : 'center';
+
+    // The input box keeps a trailing caret/measurement buffer after the
+    // digits. With the amount left-aligned that buffer would sit between the
+    // digits and the suffix symbol, so pull the suffix back over it to keep
+    // the visual gap at inlineSuffixGapPx. Skip this when maxWidth shrinks the
+    // input below its natural width: the digits then scroll inside the box and
+    // a negative margin would overlap them.
+    const inlineInputTrailingSpacePx = Math.max(
+      inlineInputWidthPx - Math.ceil(inlineMeasuredAmountWidthPx),
```
sidmorizon, line 333: # 🟡 P2 Suffix pull-back uses the requested input width, not the width `minWidth` actually renders

## Problem

`inlineInputTrailingSpacePx` (`AutoSizeInput.tsx:328-331`) assumes the amount box is `inlineInputWidthPx` wide, and `desktopSuffixMarginLeftPx` (`:335-338`) subtracts that trailing space from the gap so the symbol lands exactly `inlineSuffixGapPx` after the last digit.

But `inlineInputWidthPx` (`:288-291`) is only the *requested* width. The same box also gets `minWidth: Math.ceil(effectiveFontSize * 1.2)` (`:407`), and both land on one `Group` via `{...containerProps}` (`packages/components/src/forms/Input/index.tsx:504`). In CSS `min-width` beats `width`, so whenever `ceil(1.2 * f) > inlineInputWidthPx` the rendered box is wider than this math assumes and the pull-back under-corrects by the difference.

That is true for any **single-character** amount — i.e. the empty `0` placeholder this PR is fixing (`inlineMeasureText = value || placeholder || '0'`, `:243`). A digit measures ~`0.58 * f` (the file's own estimator, `:22-23`) and the buffer is `max(18, round(0.5 * f))` (`:261-264`), so the requested width is ~`1.08 * f`, always below the `1.2 * f` floor.

Desktop, non-`md`, empty field (`fontSizeScale = 1.5` → `f = 84`, `index.tsx:236,373-376`):

| quantity | value |
|---|---|
| `measured("0")` | ≈ 49 |
| `inlineInputBufferPx` (`:261`) | 42 |
| `inlineInputWidthPx` (`:288`) | `max(91, 89) = 91` |
| `minWidth` (`:407`) | `ceil(84 × 1.2) = 101` ← **width actually used** |
| `inlineInputTrailingSpacePx` (`:328`) | `91 − 49 = 42` (real trailing: `101 − 49 = 52`) |
| `inlineSuffixGapPx` (`index.tsx:415-418`) | `max(4, ceil(84 × 0.28)) = 24` |
| rendered gap | `52 − 18 = 34` instead of 24 |

Same shape on `md` (`f = 67`): 27 instead of 19. The error is ~`0.12 * f` in every case.

Note `ceil(effectiveFontSize * 1.05)` at `:290` is already dead code — `minWidth` at `:407` always dominates it — which is why the two constants drifting apart was invisible until this diff started depending on the value.

Secondary, same root cause: `isInlineInputShrunk` (`:332-334`) tests against `availableInlineWidth`, which falls back to `windowWidth` until the first `onLayout` lands (`index.tsx:244,378-381`). On the first painted frame after a mount inside a narrow container, a pre-filled long value reads the guard as `false` while `maxWidth` is genuinely clamping the box, so the negative `marginLeft` draws the symbol over the digits for that frame.

## Impact

Desktop and web, token mode, empty/placeholder state — exactly the state this PR targets — render the token symbol ~10px (84px font) / ~8px (67px font) further from the `0` than `inlineSuffixGapPx` asks for: a ~40% over-wide gap. Before this change the margin was an unconditional `+inlineSuffixGapPx` against a right-aligned box, where the box width could not affect the gap, so the empty-state spacing was exact; this is a new regression. Row centering still holds (the flex row grows by the same amount), so it is cosmetic rather than a clipping bug — but it undercuts the PR's stated goal of leaving the digits-to-symbol spacing unchanged, and it is not covered by the `|0 ATOM` CDP check in the test plan, which verified caret position and centering rather than gap width.

## Recommendation

Derive the trailing space from the width the box actually renders at. The cheapest fix is one line — align the floor at `:290` with the `minWidth` at `:407` so `inlineInputWidthPx` *is* the used width. This also repairs `inlineInputSlackPx` (`:292-295`), which still feeds the prefix-mode row offset at `:346`:

```ts
const inlineInputWidthPx = Math.max(
  inlineAmountTextWidthPx,
  Math.ceil(effectiveFontSize * 1.2), // was 1.05; must match containerProps.minWidth (:407)
);
```

For the first-frame case, have `index.tsx` pass down whether the width has been measured (e.g. `layoutWidth > 0`) and treat "unmeasured" as shrunk at `:332`, so the negative margin only applies once the real container width is known.

[Claude session ↗](https://claude.ai/code/session_01RP8TxGNoJpHYcYN7i2XS2T)

---

packages/kit-bg/src/services/ServiceHistory.ts
```diff
@@ -2849,12 +2851,71 @@ class ServiceHistory extends ServiceBase {
...
+    if (!result.supported || result.errored) {
+      return;
+    }
+    void this.backgroundApi.simpleDb.transferRecipientsCache
+      .setEntry({
+        accountId,
+        networkId,
+        data: result.data,
+        lastUsedDeriveType: result.lastUsedDeriveType,
+      })
```
sidmorizon, line 2898: 🟡 P2: Cleared history retains recipient addresses

Severity: `non-severe`

Clearing transaction history leaves `transferRecipientsCache` untouched. [clearCacheOnApp](packages/kit-bg/src/services/ServiceSetting.ts:318-322) clears only local history and address data. Send displays the retained addresses after a cold start.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
```diff
@@ -9002,30 +9002,46 @@ export abstract class LocalDbBase extends LocalDbBaseContainer {
...
-      accounts = records.filter(Boolean);
-    }
-    if (!ids) {
+      return records.filter(Boolean);
+    })();
+    this.dbAllRecordsInflight.set(cacheKey, readPromise);
+    try {
+      const accounts = await readPromise;
       this.dbAllRecordsCache.set(cacheKey, accounts);
+      return { accounts };
```
sidmorizon: 🟡 P2: Account invalidation restores stale snapshots

Severity: `non-severe`

An Account write can clear caches while `readPromise` still holds the old snapshot. The promise later repopulates `dbAllRecordsCache`, so subsequent reads return pre-write accounts.

packages/kit-bg/src/services/ServiceHistory.ts
```diff
@@ -2798,7 +2798,9 @@ class ServiceHistory extends ServiceBase {
     };
 
     if (xpubEntries.length <= 1) {
-      return callOnce(xpubEntries[0]?.xpub, limit);
+      const single = await callOnce(xpubEntries[0]?.xpub, limit);
+      this.persistTransferRecipients({ accountId, networkId, result: single });
+      return single;
```
sidmorizon, line 2808: 🟡 P2: Short queries truncate recipient cache

Severity: `non-severe`

`fetchTransferRecipients` persists results under a key omitting `limit`. [The interaction lookup](packages/kit-bg/src/services/ServiceAccountProfile.ts:785-794) requests ten and can replace Send’s twenty recipients. Cold starts then show a truncated Recent list until refresh.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
sidmorizon: 🟡 P2: Cold caller mutations poison account cache

Severity: `non-severe`

The first cold caller receives the same `accounts` objects stored in `dbAllRecordsCache`. Mutating its result changes later cache hits, unlike joined or warm reads.

Suggested fix:

```
      this.dbAllRecordsCache.set(cacheKey, accounts);
      return { accounts: cloneDeep(accounts) };
```

packages/kit-bg/src/dbs/simple/base/SimpleDb.ts
```diff
@@ -668,6 +668,19 @@ export class SimpleDb {
     return value;
   }
 
+  get transferRecipientsCache() {
```
sidmorizon, line 671: # 🟠 P1 New simpleDb entity breaks two hardcoded entity-count guards — CI is red on this head

## Problem

Registering `transferRecipientsCache` here (`SimpleDb.ts:671`) takes the SimpleDb entity count from 67 to 68 and the lazy-facade count from 65 to 66. Two existing tests pin those numbers and were not updated in this PR:

- `packages/kit-bg/src/dbs/simple/base/SimpleDbEntityAsyncMethods.test.ts:105` — `expect(entityClassCount).toBe(67);`
- `packages/kit-bg/src/dbs/simple/base/SimpleDb.lazyEntities.test.ts:87` — `expect(entityNames).toHaveLength(65);`

Both enumerate the entities off `SimpleDb` itself, so adding a getter here moves the count by construction. Neither file appears in this PR's diff (`git diff --name-only 75cbad57..b3f2e5dc` lists only `SimpleDb.ts`, `SimpleDbProxy.ts` and the new entity under `dbs/simple/`).

Observed on CI run [35049816596](https://github.com/OneKeyHQ/app-monorepo/actions/runs/35049816596), head `b3f2e5dc`:

```
FAIL packages/kit-bg/src/dbs/simple/base/SimpleDbEntityAsyncMethods.test.ts
  ● SimpleDb entity async method contract › keeps every public entity instance callable explicitly async
    Expected: 67
    Received: 68
    at SimpleDbEntityAsyncMethods.test.ts:105:30

FAIL packages/kit-bg/src/dbs/simple/base/SimpleDb.lazyEntities.test.ts
  ● SimpleDb lazy entities › creates stable non-promise facades for every entity without loading them
    Expected length: 65
    Received length: 66
    at SimpleDb.lazyEntities.test.ts:87:25
```

## Impact

`Unit Tests (2/5)`, `Unit Tests (4/5)` and the `unittest (24.x)` rollup all fail on `b3f2e5dc`, so the PR cannot merge as-is. This is deterministic, not a flake — the count is derived from the registration, so every re-run fails identically. It is not a defect in the feature itself; the two guards exist precisely to force a conscious review whenever the entity set changes, and this PR changed it without acknowledging them.

## Recommendation

Update both assertions in the same commit that registers the entity:

- `SimpleDbEntityAsyncMethods.test.ts:105` → `expect(entityClassCount).toBe(68);`
- `SimpleDb.lazyEntities.test.ts:87` → `expect(entityNames).toHaveLength(66);`

Worth confirming while you are there that the new entity also satisfies what those two guards actually assert — every public instance method explicitly `async`, and a stable non-promise lazy facade — rather than only bumping the numbers to green.

[Claude session ↗](https://claude.ai/code/session_01RP8TxGNoJpHYcYN7i2XS2T)

---

packages/kit/src/views/Send/components/SendAutoSizeAmountInput/AutoSizeInput.tsx
weatherstar: Fixed in 5c57775570. Introduced `inlineInputMinWidthPx = ceil(f * 1.2)` and use it for both the `inlineInputWidthPx` floor and `containerProps.minWidth`, so `inlineInputTrailingSpacePx` (and `inlineInputSlackPx`) are derived from the width the box actually renders at. For the first frame, `index.tsx` now passes `isInlineWidthMeasured={layoutWidth > 0}` and `isInlineInputShrunk` treats "unmeasured" as shrunk, so the negative suffix margin only applies once the real container width is known.

packages/kit-bg/src/services/ServiceHistory.ts
weatherstar: Fixed in 5c57775570. `clearCacheOnApp` now calls `simpleDb.transferRecipientsCache.clear()` alongside `localHistory` / `addressInfo` when `transactionHistory` is selected, so a cleared history no longer resurfaces recipients on the next cold start.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
weatherstar: Fixed in 5c57775570. The cache is now only populated when `dbAllRecordsInflight.get(cacheKey) === readPromise` after the read resolves; a flush during the read drops our entry, so the pre-write snapshot is never written back. Added a test (`does not repopulate the cache from a read that predates a flush`). Note this race also existed before this PR, but the in-flight entry gives us a cheap way to detect it now.

packages/kit-bg/src/services/ServiceHistory.ts
weatherstar: Fixed in 5c57775570. `limit` is now part of the cache entry and key (`${accountId}__${networkId}__${limit}`); `persistTransferRecipients` / `getCachedTransferRecipients` thread it through and the Send hook reads with `limit: MAX_RECIPIENTS`. The 10-item interaction lookup writes its own entry and can no longer replace the 20-item Send list.

packages/kit-bg/src/dbs/local/LocalDbBase.ts
weatherstar: Fixed in 5c57775570. The cold caller now receives `cloneDeep(accounts)`, matching warm reads (`getAllRecordsByCache`) and joiners, so mutating a result cannot poison later cache hits. Added a test (`gives the cold caller its own copy, not the cached objects`).

packages/kit-bg/src/dbs/simple/base/SimpleDb.ts
weatherstar: Fixed in 0370c4c983: bumped the entity-count guards to 68 (`SimpleDbEntityAsyncMethods.test.ts`) and 66 (`SimpleDb.lazyEntities.test.ts`) for the new `transferRecipientsCache` entity. Both suites pass locally; CI is re-running on this head.

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -328,6 +370,19 @@ export function useRecentRecipientsData({
...
       setLastUsedDeriveType(undefined);
+      // Cold start: paint the persisted answer (if any) while the API runs.
+      const persisted = await loadPersistedApiRecipients({
+        accountId,
+        apiNetworkId,
+        networkId,
+      });
+      if (isStale()) return;
+      if (persisted) {
+        cached = persisted;
```
sidmorizon, line 395: # 🟡 P2 Persisted cold-start entry never reaches the session cache, so a failed refresh leaves it empty

## Problem

Line 381 assigns the persisted entry to the local `cached` variable, but `commit()` (`useRecentRecipientsData.ts:388-395`) is the only writer of `recentRecipientsCache` (line 390). Nothing writes the persisted entry to the shared session cache.

That matters because `cached` is now truthy with `apiUnsupported: false`, which makes the early return at line 457 reachable on a cold start:

```ts
if (apiFailed && cached && !cached.apiUnsupported) {
  return;
}
```

So a cold start whose API refresh fails returns *without* ever calling `commit()`, and the session cache stays unset for that key.

This is a behavior change introduced here. At the previous head the declaration was `const cached = recentRecipientsCache.get(cacheKey);` and was never reassigned (`git show 75cbad57:…/useRecentRecipientsData.ts`, line 322), so on a cold start `cached` stayed `undefined`, `apiFailed && cached` was false, execution fell through to Phase 2 and `commit()` at line 490 populated the cache. Changing it to `let` and reassigning at line 381 flips that early return from unreachable to taken.

## Impact

After an offline or failed cold start, `recentRecipientsCache` has no entry for that `accountId__networkId`. The next mount of the Send page therefore re-enters the cold branch: `initialCacheEntry` is `undefined` (lines 319-327), so `isLoadingRecent` initialises to `true` and `recentRecipients` to `[]` — the user sees the empty skeleton flash again, and the full persisted read plus up to `MAX_RECIPIENTS` `queryAddress` calls are replayed. This repeats on every reopen for as long as the API keeps failing, which is exactly the stale-while-revalidate behaviour the comment at lines 289-291 says this cache exists to provide.

The new test *"keeps the persisted list when the cold-start refresh fails"* (`useRecentRecipientsData.test.tsx`) covers this path but asserts only on-screen state, never cache contents, so it passes.

## Recommendation

Write the persisted entry into the session cache when it is painted, guarded the same way `commit()` guards its write:

```ts
if (persisted) {
  cached = persisted;
  if (isLatestVersion()) {
    recentRecipientsCache.set(cacheKey, persisted);
  }
  setRecentRecipients(persisted.recipients);
  setLastUsedDeriveType(persisted.lastUsedDeriveType);
  setIsLoadingRecent(false);
}
```

A later successful `commit()` overwrites it with the server answer, so the source-of-truth rules in OK-53284 are unaffected. Worth extending the existing failed-refresh test to assert the cache is populated, since on-screen state alone cannot distinguish the two behaviours.

[Claude session ↗](https://claude.ai/code/session_01RP8TxGNoJpHYcYN7i2XS2T)

---

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
```diff
@@ -328,6 +370,19 @@ export function useRecentRecipientsData({
       setIsLoadingRecent(true);
       setRecentRecipients([]);
       setLastUsedDeriveType(undefined);
+      // Cold start: paint the persisted answer (if any) while the API runs.
+      const persisted = await loadPersistedApiRecipients({
```
sidmorizon, line 388: # 🟡 P2 Cold start queues the API request behind the full persisted-enrichment batch

## Problem

This `await` (line 374) runs `loadPersistedApiRecipients` to completion before execution reaches `fetchTransferRecipients` at line 413. That is not a cheap storage read. `loadPersistedApiRecipients` (lines 255-288) does:

1. `getCachedTransferRecipients` — bridge round trip to `bg`;
2. `buildExtraMapFromApiRecipients` — one `serviceNetwork.getNetworkSafe` call per unique `networkId`;
3. `enrichAddresses` — up to `MAX_RECIPIENTS` (20) parallel `serviceAccountProfile.queryAddress` calls.

Each `queryAddress` reaches `getAccountNameFromAddress`, which is precisely the `getAllAccounts()` scan that the sibling commit in this PR added `dbAllRecordsInflight` to de-duplicate — `LocalDbBaseContainer.ts` cites "twenty recipient lookups on the Send page" as the motivation. So the cold start now runs that 20-call enrichment batch **twice**: once for the persisted list, once again for the API list at line 425.

The two are independent — the API request's arguments (`accountId`, `apiNetworkId`, `MAX_RECIPIENTS`) do not depend on the persisted result — so the serialization is incidental, not required.

## Impact

Time-to-*fresh* list on a cold start becomes `T_storage + T_networks + T_20×queryAddress + T_api` instead of `T_api`. The persisted paint still arrives early (the intended win is preserved), but the stale list now stays on screen measurably longer than the correct list used to take to appear, and total cold-start `queryAddress` work doubles. On the exact path OK-63452 targets, part of the latency the PR removes from the *first* paint is added back before the *correct* paint.

## Recommendation

Start the request before awaiting the persisted read so the round trip overlaps local enrichment. Roughly: create the promise just above line 374 and await it at line 413 instead of calling there.

```ts
// Cold start: the API request does not depend on the persisted answer, so
// let the round trip overlap the local enrichment below.
const apiPromise = backgroundApiProxy.serviceHistory.fetchTransferRecipients({
  accountId,
  networkId: apiNetworkId,
  limit: MAX_RECIPIENTS,
});
```

Nothing is wasted by issuing it eagerly: on every cold path `apiUnsupported` is `false` at line 396 (`cached` is `undefined` until line 381 assigns a `persisted` entry, which always carries `apiUnsupported: false`), so the guard at line 405 admits the request unconditionally anyway. Note this needs care on the warm path, where `cached?.apiUnsupported` can be `true` and the request must still be skipped — so hoist it inside the cold `else` branch only.

[Claude session ↗](https://claude.ai/code/session_01RP8TxGNoJpHYcYN7i2XS2T)

---

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
weatherstar: Fixed in 2a6691aa92. The persisted entry is now written to `recentRecipientsCache` when it is painted (guarded by `isLatestVersion()`, same as `commit()`), so a failed cold-start refresh leaves the next mount with an instant paint. Extended the failed-refresh test to remount and assert the cache was seeded (`isLoadingRecent` false, list painted, `getCachedTransferRecipients` called once).

packages/kit/src/views/Send/pages/SendDataInput/useRecentRecipientsData.ts
weatherstar: Fixed in 2a6691aa92. In the cold branch the request is now created before awaiting `loadPersistedApiRecipients` and awaited in Phase 1, so the round trip overlaps the local enrichment; the warm path still goes through the `apiUnsupported` guard unchanged. A no-op `.catch` keeps an early stale return from leaving the promise unhandled. Added a test that asserts the request is issued while the persisted read is still pending.

originalix (approved):

PR #13463, opened 2026-09-16
fix: preload Swap review preparation (OK-59985)

zhaono1:
[OK-59985](<https://onekeyhq.atlassian.net/browse/OK-59985>)

----------

## Summary
- Preload Swap review build, unsigned transaction, fee estimation, and encoded transaction work behind a shared preview task.
- Preserve request cancellation and stale-result protection while keeping Send fee cancellation isolated.

## Issues
- Fixes OK-59985

## Validation
- Targeted Swap/shared tests: 76 passed.
- `git diff --cached --check` passed.
- `yarn agent:check --profile commit`: lint, format, native-storage, agent-context, background API contract, and staged lint passed.
- Staged TypeScript check is blocked by pre-existing errors in `packages/components/src/composite/Tabs/context.ts` and `packages/kit/src/views/Home/components/PullToRefresh/index.tsx`.

## Risk
- Swap review preparation now spans background preview tasks and foreground lifecycle ownership; runtime UI verification and CI remain for follow-up.

[OK-59985]: https://onekeyhq.atlassian.net/browse/OK-59985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -2401,48 +2551,54 @@ export function useSwapBuildTx({
...
-              fromToken: data?.fromTokenInfo,
-              toToken: data?.toTokenInfo,
-            }),
-          });
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          const ne = new Error(e?.message ?? 'unknown error');
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          ne.name = 'buildSwapApi';
-          throw ne;
+          if (!preloading)
```
zhaono1: **[P1] 埋点口径：EVM Review 建单失败不再上报，费用成功事件被预跑放大**

这两处 analytics 门控与 `preloading` 的关系不一致：

1. **`swapCreateOrder` FAIL 丢失（EVM 全链）**：本行 `if (!preloading)` 让预跑阶段不上报 FAIL —— 这是对的。但 EVM 的 Review 建单路径恒定带 `preloading: true`：`preSwapBeforeStepActions` → `canReusePreparation`（EVM + 网络匹配，`useSwapBuiltTx.ts:4422`）→ `acquireReviewBuild` → `buildSwapAction({ preloading: true })`（`:4122`）。因此**用户在 Review 里真实遇到的建单失败也不会再上报 FAIL**，而 SUCCESS 仍由 `publishBuild` → `swapBuildFinish` 照常上报，EVM 链的建单漏斗会静默变成 100% success（非 EVM 冷路径仍上报，口径按链分裂）。只有 slippage rebuild 路径（不带 `preloading`）还会上报 FAIL。

2. **`swapEstimateFee` SUCCESS 未按 `preloading` 门控**：`:3598` / `:3780` 的 SUCCESS 只受新加的 `isCurrent()` 约束。预跑期间 `isCurrent()` 恒为 true，因此每轮 15s 报价刷新完成都会产生一条费用成功事件（自动刷新上限 5 次/次手动动作），而用户可能从未打开 Review；同一业务里的 `swapBuildFinish` 却是有门控的。

建议：给 `IBuildSwapActionOptions` 增加一个区分「推测性预跑」与「已认领的 Review 构建」的标志（例如 `reportBuildFailure`），由 `acquireReviewBuild` 传 `true`、`prepareSwapReviewBuild` 不传；费用侧用同一标志或 `!preloading` 同步门控，让三方口径一致。

_— Auto-review by Claude_

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_012e93UXTAgHoiLK7JP5Yssr

packages/kit-bg/src/services/ServiceGas.ts
```diff
@@ -53,15 +71,15 @@ class ServiceGas extends ServiceBase {
...
+        signal: controller?.signal,
         headers:
           await this.backgroundApi.serviceAccountProfile._getWalletTypeHeader({
             accountId,
           }),
       },
     );
 
-    this._estimateFeeController = null;
+    if (controller) this._estimateFeeController = null;
```
sidmorizon: 🟡 P2: Concurrent Send cancellation handle lost

Severity: `non-severe`

Overlapping Send estimates let `batchEstimateFeeInternal` clear the newer request's controller when the older request finishes. [estimateFeeInternal](packages/kit-bg/src/services/ServiceGas.ts:106-129) repeats the race. `abortEstimateFee` then cannot cancel the active estimate.

packages/shared/src/utils/swapPreviewTask.ts
```diff
@@ -0,0 +1,49 @@
+import {
```
originalix, line 1: P1: [New mobile bundle modules are not registered]

On production/UNION mobile builds, Metro resolves this newly added runtime module in strict registry mode. Because `apps/mobile/bundle-registry/module-id-registry.json` has no entry for it (or for the other three new runtime modules), the resolver throws before the bundle can be produced.

Add all four paths through the project's module-ID registry update flow and commit the resulting registry change.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3867,37 +4347,190 @@ export function useSwapBuildTx({
...
+                prepared = await preparation.promise;
+              } else {
+                preparation.cancelFee();
+              }
+            } catch (error) {
+              if (!(error instanceof SwapReviewBalanceError)) throw error;
+              // A balance failure created no order. Retry only its balance gate.
+              preparation.cancel();
+              await checkBuildBalances(data);
+            }
```
sidmorizon, line 4534: # 🟠 [P1] Rejected preload build is adopted permanently; fee-level retry replays the same failure

## Problem
A preload build that fails with a non-balance error (network/5xx from `preloadBuildTx`, rethrown at `useSwapBuiltTx.ts:2591`) leaves `preparation.build.promise` permanently rejected, but nothing ever flips its `isCurrent()`: `swapReviewPreload.ts:131-138` only clears `buildCurrent` via `cancel()`, and this `catch` (`:4451-4456`) cancels only for `SwapReviewBalanceError`. Every other error is rethrown to the outer catch (`:4565-4586`), which sets `stepBeforeActionsError: true` without cancelling the build.

Consequences on the EVM path (`canReusePreparation`, `:4422-4424`):
1. **Open**: `claimPreparation()` → `claim()` (`swapReviewPreload.ts:169-176`) returns the candidate because `candidate.isCurrent()` is still `true`, so a build that failed up to a whole quote cycle ago is adopted and the Review opens straight into the error state without one fresh `fetchBuildTx` attempt.
2. **Fee-level change** (`PreSwapDialogContent.tsx:581-597` → new `contextKey` → `preSwapBeforeStepActions` again with the same `preparation`): `preparation.build.isCurrent()` is still `true` and the key still matches (`:4426-4434`, the key excludes fee level), so `await preparation.build.promise` (`:4437`) rejects again with the *same Error instance*. `showToastOfError` is then deduped by `$$autoToastErrorTriggered` (`errorToastUtils.ts:206-212`), so the user gets no toast, no rebuild, and `stepBeforeActionsError` again.
3. The review-owned path has the same shape: `acquireReviewBuild` registers the promise before it settles (`:4131`), so a rejected `active.promise` is re-awaited and rethrown at `:4074-4079` without `active.cancel()`.

The base branch called `buildSwapAction` fresh on every fee change, so each change was a retry. The comment at `:4019-4020` ("a changed fee/input can explicitly start another preparation") holds for fee-estimate failures (`status: 'failed'`) but not for build rejections.

## Impact
One transient build-tx failure during preload, which happens with no user action and can be tens of seconds stale, locks the opened Review with Confirm disabled (`PreSwapDialogContent.tsx:374-380`). Fee-level changes appear to be ignored with no feedback. Only closing and reopening the Review (`onPreSwapClose` → `clearPreparation()`) recovers. This affects every EVM source network, i.e. the whole preload surface.

## Recommendation
- In this `catch`, for errors that are not `SwapReviewBalanceError`, call `preparation.cancel()` (or at least `preparation.build.cancel()`) before rethrowing, so the next run falls through to `acquireReviewBuild` and starts a fresh build instead of replaying the rejection.
- Mirror this in `acquireReviewBuild` (`:4076-4080`): on a rejected `active.promise` that is not a balance error, `active.cancel()` and clear `reviewPreparationStateRef.current.build` before rethrowing.
- Consider also making `claim()` skip a candidate whose build has already rejected (e.g. track a `settledWithError` flag on the build task), so opening the Review always gets one fresh attempt.
- Add a test in `useSwapBuiltTx.test.tsx`: preload build rejects → open Review → change fee level → expect `fetchBuildTx`/`preloadBuildTx` to be called a second time and the error state cleared.

[Claude session ↗](https://claude.ai/code/session_012e93UXTAgHoiLK7JP5Yssr)

---

sidmorizon:
Claude review session: https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3867,37 +4365,221 @@ export function useSwapBuildTx({
...
+                  getSwapReviewPreparationKey(data, contextKey)
+              ) {
+                publishBuild(built.buildResult);
+                prepared = await preparation.promise;
+              } else {
+                preparation.cancelFee();
+              }
+            } catch (error) {
+              if (!(error instanceof SwapReviewBalanceError)) {
+                preparation.cancel();
```
sidmorizon, line 4528: # 🟡 [P2] Claim-time build failures no longer emit `swapCreateOrder` FAIL

## Problem
`prepareSwapReviewBuild` (`useSwapBuiltTx.ts:3952-3966`) calls `buildSwapAction` with `preloading: true` and no `reportAnalytics`, so `reportAnalytics` defaults to `false` (`:2465`) and the FAIL branch at `:2558` is skipped. That is right for a speculative background build, but the same function is also the build that `claim()` starts when the user opens Review (`useSwapReviewPreload.ts:156-161` → `claim` → `preload` → `prepareBuild`, `swapReviewPreload.ts:133`), and it is the build adopted here when a preload build is still in flight at click time. When that promise rejects at `:4455`, this catch runs `preparation.cancel(); throw error;`: the outer catch shows the toast (`:4619-4620`) and sets `stepBeforeActionsError`, but no `swapCreateOrder` FAIL event is emitted, and `acquireReviewBuild` (which does pass `reportAnalytics: true`, `:4140`) is never reached because of the throw.

On the base branch every Review build failure emitted FAIL unconditionally (`useSwapBuiltTx.ts:2404` at `591ce92`). The earlier P1 thread on this topic was closed by adding `reportAnalytics: true` to `acquireReviewBuild` only; the claim-time path was left out.

## Impact
For EVM sources, order-build failures that the user actually sees (toast, Confirm disabled) are missing from the `swapCreateOrder` funnel, while SUCCESS is still reported via `publishBuild` → `swapBuildFinish`. The failure rate is under-counted exactly in the user-initiated case.

## Recommendation
Either emit the FAIL event in this catch for non-`SwapReviewBalanceError` errors (dedupe with a `reportedBuildFailure` ref like `reportedFee` at `:4481`), or thread a `reportAnalytics` flag from `claim()` → `prepareBuild` so a claim-time build reports the same way `acquireReviewBuild` does. Add a test in `useSwapBuiltTx.test.tsx`: preload build in flight → open Review → build rejects → expect exactly one `swapCreateOrder` FAIL.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3867,37 +4365,221 @@ export function useSwapBuildTx({
...
+            prepared?.status === 'estimated' &&
+            prepared.buildResult === buildResult &&
+            !isSwapPreparedFeeExpired(
+              prepared.feeResult.netWorkFee,
+              prepared.preparedAt,
+            )
+          ) {
+            if (reviewPreparationStateRef.current.reportedFee !== prepared) {
+              reviewPreparationStateRef.current.reportedFee = prepared;
+              void swapEstimateFeeEvent(
```
sidmorizon, line 4584: # 🟡 [P2] Adopting a preloaded batch-estimate fallback reports `swapEstimateFee` SUCCESS

## Problem
In `estimateNetworkFee`, when the batch estimate throws and `canFallbackToSeparateTxConfirm` is true (`:3630-3665`), FAIL is emitted only `if (reportAnalytics)` (`:3632`) and the function returns `{ fallbackToSeparateTxConfirm: true }`. During preload `reportAnalytics` is `false` (`:4020` → `:2465`), so FAIL is suppressed and `prepareSwapReview` wraps the result as `status: 'estimated'` (`:4024-4030`). On adoption the branch at `:4511-4518` accepts it (`netWorkFee` is `undefined`, so `isSwapPreparedFeeExpired` only checks age) and this line reports SUCCESS for an estimate that actually failed; the Review then correctly switches to separate steps at `:4579-4597`.

The base branch emitted FAIL unconditionally at this point (`:3407` at `591ce92`).

## Impact
Batch-estimate failures on EVM approve+swap flows are logged as SUCCESS whenever the failure happened during preload, which inverts the signal this event exists for.

## Recommendation
Carry the failure into `IEstimateNetworkFeeResult` when falling back (e.g. `fallbackError?: string`, `:215-223`), and at `:4519` emit FAIL with that message when `prepared.feeResult.fallbackToSeparateTxConfirm` is set instead of SUCCESS. Alternatively have `prepareSwapReview` classify a fallback result as `status: 'failed'` so it goes through the FAIL branch at `:4480-4497`, while still letting the fallback open separate steps.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapReviewPreload.ts
```diff
@@ -0,0 +1,165 @@
...
+      manualSelection,
+      isWaitingAutoSlippage,
+    });
+
+  useEffect(() => {
+    const owner = cache.current;
+    const snapshot = quoteRef.current;
+    if (mayPreload && key && buildKey && snapshot) {
+      const prepareSnapshot = prepareRef.current;
+      owner.preload(
```
sidmorizon, line 129: # 🟡 [P2] Preload issues a full, indistinguishable build-tx on every completed quote refresh

## Problem
This effect re-runs whenever `key` changes (`:139`). `key` embeds the whole quote minus `isBest/receivedBest/minGasCost` (`swapReviewPreload.ts:62-63`), so every quote refresh that reaches `completed` (`swapRefreshInterval` auto-refresh, up to `swapQuoteIntervalMaxCount` = 5 per cycle, plus every manual refresh or input change) starts a new `prepareBuild` → `preloadBuildTx` → `fetchBuildTx` (`ServiceSwap.ts:1427-1428`), followed by `prepareReviewUnsignedTx` and a fee estimate. The previous candidate is cancelled only client-side (`clearCandidate`, `swapReviewPreload.ts:116-119`); its HTTP requests still complete. `preloadBuildTx` sends exactly the same params as the real build, so neither the backend nor the provider can tell a speculative build from a user-confirmed one.

For bridge providers build-tx is order creation: the response carries `swftOrder` / `changellyOrder` payin addresses and `orderId` (`shared/types/swap/types.ts:977-985`), and the client itself names this step `createSwapOrder`. Unadopted preload builds never reach `swapBuildFinish` (gated on `!preloading`), so provider-side orders are created without the matching client `swapCreateOrder` SUCCESS event.

The enable condition (`SwapMainLand.tsx:895-904`) gates on `supportPreBuild`, which is a batch-transfer capability (`SwapMainLand.tsx:872-885`), not an idempotency guarantee.

## Impact
An idle user on the Swap page with a valid EVM quote now generates roughly 5-7 backend calls per refresh cycle (balance reads, build-tx, unsigned tx, estimate, native balance), most of which are discarded. For order-creating providers this leaves abandoned orders at refresh cadence and raises rate-limit / abuse-detection exposure. The call multiplication is certain from the code; the order-creation side effect is inferred from the response shape.

## Recommendation
- Mark preview builds explicitly (e.g. a `preview: true` field on the `preloadBuildTx` → `fetchBuildTx` request) so the backend can skip provider order creation or apply its own policy, or gate preload on a per-provider capability flag returned with the quote.
- Throttle re-preload: skip when the previous candidate's build is still within `SWAP_REVIEW_PRELOAD_MAX_AGE_MS` and only amounts changed within tolerance, or trigger preload on stronger intent (Review button press-in) instead of every completed refresh.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3867,37 +4365,221 @@ export function useSwapBuildTx({
...
+                String(
+                  prepared.feeError instanceof Error
+                    ? prepared.feeError.message
+                    : prepared.feeError,
+                ),
+                JSON.stringify(prepared.buildResult.encodedTx ?? ''),
+                prepared.buildResult.swapInfo,
+              );
+            }
+            throw prepared.feeError;
```
sidmorizon: # 🟡 [P2] A speculative fee-estimate failure is replayed on Review open although an order-free retry exists

## Problem
`prepareSwapReview` stores any unsigned-tx / fee failure as terminal `status: 'failed'` (`:4031-4041`). Because that promise resolves rather than rejects, the self-cancel in `swapReviewPreload.ts:167-170` never fires, the candidate stays current, and `claim()` returns it (`swapReviewPreload.ts:127`, `:178`). This line then throws the stored error → toast + `stepBeforeActionsError: true` → Confirm disabled (`PreSwapDialogContent.tsx:374-380`). The build succeeded, so the only work needed is `estimateNetworkFee` on the same `buildResult`, which the `else` branch already does for a fee-level change (`:4559-4577`) without creating a new order. On the base branch, opening Review always ran a fresh estimate.

The comment at `:4032-4033` and the test "adopts a %s failure without retrying" show this is intentional, so this is flagged as a design regression rather than an accident: a transient RPC error (e.g. `eth_estimateGas` 429 / timeout during a background preload the user never triggered) now surfaces as an error on the user's click, with no retry until they change fee level or close and reopen.

## Impact
Users see a failed Review for failures that happened without their action, up to a refresh cycle earlier, even when a retry would succeed. Preload multiplies estimate calls, which makes transient failures more likely, so this regresses exactly the case the preload creates.

## Recommendation
On `prepared?.status === 'failed'`, keep the FAIL event, then instead of throwing do `preparation.cancelFee(); prepared = undefined;` and fall through, so `acquireReviewBuild` reuses the current build (`:4087-4101`) and the `else` branch re-estimates once. If deterministic failures should not be retried, distinguish by error class (retry only network / timeout errors). Update the "adopts a %s failure without retrying" test to the new behavior.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3689,6 +3932,242 @@ export function useSwapBuildTx({
...
+        throw new CanceledError(
+          'Swap preparation is unavailable for this source network',
+        );
+      }
+      // The approval transaction only depends on the quote/account snapshot;
+      // it does not depend on the swap build response. Start it immediately so
+      // the relatively expensive account/unsigned-tx preparation overlaps the
+      // build-tx request instead of extending the preview critical path.
+      const unsignedTxPromise = getApproveUnSignedTxArr(data, true);
+      const [preparedBuildResult, unsignedTxResult] = await Promise.all([
```
sidmorizon: # 🟠 [P1] Approve unsigned-tx failure now rejects the whole preparation: order discarded, misleading error in Review

## Problem
Commit `b4c6ea5` moves `getApproveUnSignedTxArr(data, true)` out of the `try` at `:4013` and awaits it here via `Promise.all` together with `preparedBuild`. With `preloading = true`, `getApproveUnsignedTx` (`:2277-2281`) calls `serviceSwap.prepareReviewUnsignedTx` through `unwrapSwapPreviewTask`, which rethrows the RPC failure. Before this commit that failure was caught at `:4054-4063` and stored as `status: 'failed'` alongside the successful `buildResult`; now it rejects `prepareSwapReview` itself.

A rejected fee preparation triggers `task.cancel()` (`swapReviewPreload.ts:167-170`), which also runs `buildOwner.cancel()` (`:161-164`), invalidating the build's `isCurrent` predicate. Two orderings follow for quotes with `allowanceResult`:

1. **Approve prep fails before build-tx returns, user already in Review** (clicked while the preload was in flight, or claim-time build): `buildSwapAction` sees `!isCurrent()` after the response (`:2603-2604`) and rejects with `OneKeyError('Swap review changed during preparation')`. The adoption `catch` at `:4478-4482` rethrows it, so the user gets a toast with that internal message and `stepBeforeActionsError: true` (Confirm disabled). The real error (`offline`, RPC timeout) is never shown, and the created order is discarded.
2. **Approve prep fails during background preload**: the successfully built order is cancelled; on open, `preparation.build.isCurrent()` is `false` (`:4455`), so `acquireReviewBuild` issues a second `fetchBuildTx` (a second provider order) before estimating.

The test `adopts a %s failure without retrying, retaining the order for an explicit fee change` (`useSwapBuiltTx.test.tsx:624-685`) uses the base `quote` fixture without `allowanceResult` (`:76-88`), so its `prepare` rejection hits the swap unsigned tx inside `estimateNetworkFee` (`:3554`), which is still inside the `try`. The approve path this commit changed is untested, and the `status: 'failed'` contract documented at `:4054-4055` is silently broken for it.

## Impact
On every EVM approve+swap quote, one transient failure in approve unsigned-tx preparation either shows a wrong error and locks Confirm, or throws away a created order and creates another. This contradicts the retained-order design the rest of the preparation follows, and it lands exactly in the flow this PR is optimising.

## Recommendation
Keep the overlap but preserve the failure contract: start `unsignedTxPromise` here, await `preparedBuild` alone, run the `isCurrent`/`swapInfo` checks, then `await unsignedTxPromise` inside the existing `try` so a rejection becomes `status: 'failed'` with the build retained. Attach a no-op `.catch` to `unsignedTxPromise` before the first `await` so an early rejection cannot surface as unhandled while `preparedBuild` is pending. Add a variant of the `adopts a prepare failure` test with `allowanceResult` set so the approve path is covered.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
zhaono1: Fixed in `b1ac8e84`. Build failure analytics is now shared and emitted when a claimed preload build fails, while speculative background preload remains silent. Stale/cancelled runs are excluded.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
zhaono1: Fixed in `b1ac8e84`. Batch-estimate fallback now carries the original error and emits `swapEstimateFee` FAIL when the fallback result is adopted; SUCCESS is only emitted for a real successful estimate.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
zhaono1: Fixed in `b1ac8e84`. A failed speculative fee task is cancelled when Review adopts it, then the valid build is reused for a fresh fee estimate without rebuilding the order.

packages/kit/src/views/Swap/hooks/useSwapReviewPreload.ts
zhaono1: Reviewed. The repeated preload request is real, but changing it safely requires a backend/provider contract for preview builds: the current endpoint may create provider-side orders, and the client has no capability signal to distinguish safe providers. I am resolving this as a separate backend follow-up rather than introducing an unverified client-only flag in this PR.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
zhaono1: Fixed in `8c8d37db`. The unsigned approval preparation still starts in parallel, but its rejection is now awaited inside the existing fee-preparation try/catch after the build is retained. An early rejection gets a no-op handler to avoid unhandled rejections. Added an approval-path regression test to verify the built order is reused without a second build.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -2389,60 +2597,43 @@ export function useSwapBuildTx({
...
+              effectiveSlippagePercentage,
+              error: e,
+            });
+          }
+          // The proxy may still show this error after its delayed catch.
+          // Keep one instance so both paths share the Toast marker.
+          throw Object.assign(
+            e && typeof e === 'object'
+              ? e
+              : new Error(String(buildError.message ?? 'unknown error')),
```
sidmorizon, line 2630: # 🟠 [P1] `buildError` is referenced after its declaration was removed; TypeScript check fails on this head

## Problem
Commit `f16289f` extracted the FAIL analytics into `reportSwapBuildFailure` (`:2450-2508`) and deleted `const buildError = toPlainErrorObject(e);` from this `catch` (it was at `:2547` on the previous head), but the rethrow on this line still reads `buildError.message`. The only remaining `buildError` binding in the file is the local inside `reportSwapBuildFailure` (`:2462`), which is not in scope here.

CI confirms it: the `lint (24.x)` job on `8c8d37d` fails in the TypeScript step with `useSwapBuiltTx.ts: line 2629 col 34, Cannot find name 'buildError'`.

## Impact
The PR cannot pass the required lint/TypeScript check on its current head. At runtime the bundles are Babel-compiled without type checking, so if `fetchBuildTx` / `preloadBuildTx` ever rejects with a non-object value, this branch throws `ReferenceError: buildError is not defined` instead of the intended `Error` with the build message, and the `buildSwapApi` name is never attached.

## Recommendation
Restore the local before the rethrow, e.g. `const buildError = toPlainErrorObject(e);` at the top of the `catch`, or inline it on this line: `new Error(String(toPlainErrorObject(e).message ?? 'unknown error'))`. Run `yarn agent:check --profile commit` (which includes the TypeScript check) before the next push.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

zhaono1: Fixed in `8c03db5`. Restored the local `buildError` normalization before the build API error rethrow, clearing the TypeScript `Cannot find name 'buildError'` failure while preserving the intended error message and `buildSwapApi` identity.

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -2702,14 +2909,15 @@ export function useSwapBuildTx({
...
       slippageItem,
       fromUserAddress,
       toUserAddress,
       fromAccountNetworkId,
       fromAccountId,
       setSwapSteps,
-      checkLatestFromTokenBalance,
-      checkOtherFee,
+      checkBuildBalances,
+      reportSwapBuildFailure,
```
sidmorizon, line 2920: # 🟠 [P1] `buildSwapAction` keeps three dependencies it no longer reads; oxlint fails on this head

## Problem
`f16289f` moved the FAIL logging out of `buildSwapAction` into `reportSwapBuildFailure`, which is now the dependency added on this line. The removed block was the only place in `buildSwapAction` that read `swapFromAddressInfo.accountInfo?.device?.deviceType`, `isFirstTimeSwap` and `isModalPage` (see the `-` lines in that commit), but all three are still listed in this array at `:2922`, `:2925` and `:2926`. The callback body between `:2509` and `:2910` now references only `wallet?.type` (`:2595`) and `focusSwapPro` (`:2598`) from that group.

`react/exhaustive-deps` treats unused entries in a `useCallback` array as errors. CI confirms it: `lint (24.x)` on `8c03db5` passes the TypeScript step and then fails oxlint with exactly three errors, `React Hook useCallback has unnecessary dependency: isModalPage`, `... swapFromAddressInfo.accountInfo.device.deviceType`, `... isFirstTimeSwap`.

## Impact
The required lint check stays red on the current head, so the PR cannot merge. The same three values are still correctly listed on `reportSwapBuildFailure` (`:2499-2505`), so behaviour is unaffected; this is purely a stale dependency list, but it blocks CI.

## Recommendation
Delete `swapFromAddressInfo.accountInfo?.device?.deviceType` (`:2922`), `isFirstTimeSwap` (`:2925`) and `isModalPage` (`:2926`) from this dependency array. `reportSwapBuildFailure` at `:2920` already carries them transitively. Run `yarn agent:check --profile commit` before pushing; it runs oxlint on the staged files and would have caught this.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

packages/kit/src/views/Swap/hooks/useSwapBuiltTx.ts
```diff
@@ -3867,37 +4409,253 @@ export function useSwapBuildTx({
...
+                    ? prepared.feeError.message
+                    : prepared.feeError,
+                ),
+                JSON.stringify(prepared.buildResult.encodedTx ?? ''),
+                prepared.buildResult.swapInfo,
+              );
+            }
+            // The build is still valid. Drop only the failed fee task and
+            // retry estimation on the same build when Review is opened.
+            preparation?.cancelFee();
```
sidmorizon, line 4555: # 🟠 [P1] Fee-failure retry rebuilds an expired order silently; `Unit Tests (1/5)` fails on this head

## Problem
The retry branch added in `f16289f` (`:4553-4556`) drops the failed fee task and falls through to `acquireReviewBuild` on the assumption stated in the comment above this line, "The build is still valid", but nothing checks that. When the fee failed after the retained build passed `SWAP_REVIEW_PRELOAD_MAX_AGE_MS`, `acquireReviewBuild` awaits the same `preparation.build` (`:4126-4134`), sees `isSwapPreparedBuildExpired(built)` (`:4145`), cancels it and starts a fresh `buildSwapAction` → `fetchBuildTx` (`:4146`, `:4172`). That is a second provider order created without any user action, in the exact case the PR's own test forbids.

CI confirms it on `8c03db5`: `Unit Tests (1/5)` fails only in `useSwapBuiltTx.test.tsx` › `does not create a replacement order when a pending fee fails after build expiry` (`:1264-1312`): the test adopts the preparation, moves `Date.now` forward 31 s, rejects the pending fee, and expects `mockFetchBuildTx` to have been called once, one error toast, and `stepBeforeActionsError: true`; it now sees two build calls (`:1301`). The other 4074 tests in the shard pass.

## Impact
The required unit-test check is red, so the PR cannot merge. Behaviourally, a fee failure that lands after build expiry now silently creates a replacement order on the provider side and swallows the user-facing error the previous code surfaced, which also contradicts the thread on preload cadence where a client-initiated extra order was explicitly not wanted.

## Recommendation
Guard the retry with the build's age: before this line, `if (isSwapPreparedBuildExpired(prepared)) throw prepared.feeError;` (`prepared` carries `buildResult` and `builtAt`, which is what `isSwapPreparedBuildExpired` takes), so an expired build surfaces the original failure and leaves the next order to an explicit user action, while a still-valid build keeps the order-free re-estimate. Keep the test as is; it encodes the intended policy.

[Claude session ↗](https://claude.ai/code/session_01BnPFHnqCy1gjEmvUKTVz7W)

---

PR #13464, opened 2026-09-16
fix: reserve dynamic USDC fees for max withdrawals(OK-63170)

Minnzen:
[OK-63170](<https://onekeyhq.atlassian.net/browse/OK-63170>)

----------

## Summary
- Replace the fixed 0.01 USDC Max-withdrawal reserve with the account-specific Hyperliquid pre-transfer fee, keeping 0.01 USDC as the minimum buffer.
- Apply the same reserve to Max, amount validation, and withdrawal readiness for CCTP and HyperEVM routes while preserving the legacy bridge behavior.
- Keep the fee display and controls aligned with the resolved reserve, and never present an unknown reserve as a confirmed `< $0.01`.

## Intent & Context
Some users could not withdraw their full available USDC after selecting Max. The reported wallet failed on a CCTP withdrawal to Base with `Insufficient USDC or HYPE balance for token transfer gas`, while another wallet could withdraw directly to Arbitrum and retained the previous fixed 0.01 USDC buffer.

The goal is to reserve the amount returned for the selected account by Hyperliquid's `preTransferCheck`, with a 0.01 USDC minimum, so accounts for which the endpoint reports a 1 USDC fee do not submit an amount that consumes those funds. Hyperliquid's own web app withholds the same 1 USDC for these accounts, so this aligns Max with the reference behavior rather than inventing a buffer.

## Root Cause
The withdrawal form always subtracted 0.01 USDC when calculating Max for CCTP and HyperEVM routes. Hyperliquid's pre-transfer requirement is account-specific and can return a larger fee, including 1 USDC. Max, validation, and submission readiness therefore allowed an amount that left too little USDC for the transfer requirement.

The reserve must be checked for the selected account before Max or submission is enabled, and an unknown value must never be rendered as a confirmed 0.01 USDC fee.

## Design Decisions
- Resolve the source address in the background from `userAccountId` and call the existing Hyperliquid SDK `preTransferCheck` endpoint. No new dependency or external API was added.
- Use `max(preTransferCheck.fee, 0.01)` as the total reserve. The minimum is a buffer rather than an extra amount added to the returned fee.
- Bind the reserve to both the selected account ID and address, and derive a single `matchedWithdrawReserve` that feeds Max arithmetic, validation, and the fee row, so those three can never disagree or read a foreign account's value.
- Show the resolved reserve in the fee row, use the existing skeleton while it is unknown, and keep account-change recovery distinct from a retryable request failure.
- **No periodic refresh.** The reserve is `max(fee, 0.01)`, so a cached value is never below the fixed 0.01 that `x` applies today with no refresh mechanism at all. Not revalidating is therefore bounded by current behavior: Hyperliquid rejects the submission, `handleConfirm`'s catch path refetches the reserve, and the retry is primed with the fresh value. Freshness otherwise comes from the hook's account and route dependency changes, focus and reconnect revalidation, and the user-facing retry. Earlier rounds of this PR added a 30-second refresh and then a visibility gate for it; both were removed. Separately, `preTransferCheck` is called with a constant destination (`HYPEREVM_SYSTEM_ADDRESS`), so only the source wallet's own state can move the value and a rise is implausible — but that is not established from this repo, and the argument above does not rely on it.
- Keep the existing signing, routing, forwarding-fee validation, and withdrawal submission paths unchanged. `usePromiseResult` is used as-is; the shared hook is not modified by this PR.

## Changes Detail
- `ServiceHyperliquidExchange.ts`: adds the background reserve query and validates the API response.
- `DepositWithdrawModal.tsx`: uses the account-matched reserve for Max, validation, and the fee row, and separates an account change from a retryable request failure.
- `withdrawFeeDisplayUtils.ts`: keeps account-specific fee presentation in a small pure module, with its required mobile module ID registered.
- Tests cover fee values, invalid/restricted responses, wallet identity, request failures, existing route behavior, and the CCTP/HyperEVM add-vs-replace display rules.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**: Withdrawal readiness across route and account changes. A stale reserve is never below the 0.01 that ships today, so it cannot under-reserve relative to current behavior; a rejected submission refetches the reserve and the user can retry. `preTransferCheck` is not a complete transaction or gas simulation; Hyperliquid's own CCTP Max logic may also reserve a separate USDC gas estimate when spot HYPE is insufficient, so a real withdrawal is still the confirming evidence for the reported wallet.

## Test plan
- [x] Run 63 targeted Jest tests for the reserve service, withdrawal route, fee display, and existing modal helpers.
- [x] Verify changed-file lint, formatting, background API contract checks, and the native startup graph budget.
- [x] Rebase onto the latest `x` and rerun the targeted suite.
- [x] Confirm on device that the Max withdrawal amount reflects the account-specific reserve.
- [ ] Complete an end-to-end Max withdrawal under the original CCTP-to-Base failure conditions.
- [ ] Regress a direct Arbitrum withdrawal and confirm legacy bridge behavior is unchanged.

## Validation Notes
The repository-wide commit gate remains blocked by existing type errors in Tabs/PullToRefresh and AutoSizeInput dependency typings, plus the existing agent skill catalog budget violation. None of those files are changed by this PR.

The fee formatter's mobile module ID was generated with the repository tool and committed with this PR.

The unit tests mock `preTransferCheck`, so they pin the call shape and the reserve arithmetic but cannot establish which fee the failing withdrawal actually owed. The evidence for the queried direction is a live check of both reported wallets against the `/info` endpoint: the failing wallet returns `fee=1.0` for `source=wallet, user=HYPEREVM_SYSTEM_ADDRESS` and `0.0` reversed, while the working wallet returns `0.0`. The two unchecked test plan items remain the confirming evidence.

[OK-63170]: https://onekeyhq.atlassian.net/browse/OK-63170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q

packages/kit/src/hooks/usePromiseResult.ts
```diff
@@ -482,6 +482,8 @@ export function usePromiseResult<T>(
   const isLoadingRef = useRef(isLoading);
   const runWithPollingNonce = useCallback(() => {
     isDepsChangedOnBlur.current = false;
+    // Recovery must replace the old polling chain instead of starting another.
+    pollingNonceRef.current += 1;
```
sidmorizon: # 🔴 [P1] Nonce bump can permanently kill polling for `debounced` consumers

## Problem

`runWithPollingNonce` now bumps `pollingNonceRef.current` **before** dispatching (`usePromiseResult.ts:486`), then runs with the new nonce (`:487`). The bump immediately invalidates the live chain — the old chain's continuation guard in `finally` (`:375` and `:381`) compares `pollingNonceRef.current === config?.pollingNonce` and exits. Chain survival now depends entirely on the *new* dispatch reaching `runner` with its `pollingNonce` intact.

When `options.debounced` is set, `runRef.current` is not `runner` but the lodash wrapper built at `:397-413` (`{ leading: false, trailing: true }`), which passes `config` straight through. Lodash trailing-invokes with the **last** args. So if any consumer-initiated bare `run(...)` — which carries no `pollingNonce` — lands inside the debounce window after a focus/reconnect revalidation, the trailing call executes with `config = { alwaysSetState: true }`, and the guard at `:375` evaluates `N+1 === undefined` → false. No continuation is scheduled, and the old chain is already dead. **Polling stops permanently** until deps change, `setStopPolling(false)` fires, or another focus/reconnect event occurs.

Before this PR the bump did not happen, so `runWithPollingNonce`'s nonce still matched the live chain; losing the debounced dispatch was harmless because the original chain kept ticking. The change converts "a duplicate chain" into "possibly zero chains".

## Impact

Broad blast radius well outside the perps withdrawal feature, across consumers that combine `debounced` with `pollingInterval` and `revalidateOnFocus`. Confirmed consumer: `packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx:1048-1053` (`debounced: POLLING_DEBOUNCE_INTERVAL`, `pollingInterval: POLLING_INTERVAL_FOR_TOKEN`, `revalidateOnFocus: true`), whose `run` is invoked bare at `:2930`, `:2985` and `:3259`.

Concrete sequence: user returns to the Tokens tab → `isFocusedRefValue` false→true → `usePromiseResult.ts:536-541` → `runWithPollingNonce()` bumps N→N+1 and queues the debounced runner → background emits `AccountDataUpdate` within the debounce window → `run({ alwaysSetState: true })` replaces the pending args → **home token-list polling dies silently** for the rest of the session. The same shape applies to `TxHistoryContainer.tsx`, `NFTListContainer.tsx`, `DeFiListBlock.tsx` and `ApprovalListPage.tsx`.

The new test at `usePromiseResult.test.tsx:532-570` does not set `debounced`, so this path is uncovered.

## Recommendation

Make the debounce wrapper carry chain ownership forward so a bare `run()` cannot drop it. In `usePromiseResult.ts:397-413`:

```ts
let pendingPollingNonce: number | undefined;
return async (config?: IRunnerConfig) => {
  pendingPollingNonce = config?.pollingNonce ?? pendingPollingNonce;
  await runnerDebounced({ ...config, pollingNonce: pendingPollingNonce });
};
```

Add a regression test mirroring the new `'keeps one polling chain after %s revalidation'` case but with `debounced: 1000` plus a bare `run()` issued inside the debounce window, asserting polling still ticks afterwards.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -439,6 +439,41 @@ function DepositWithdrawContent({
...
+      return backgroundApiProxy.serviceHyperliquidExchange.getUsdcWithdrawReserve(
+        {
+          userAccountId: selectedAccount.accountId,
+        },
+      );
+    },
+    [needsWithdrawReserve, selectedAccount.accountId],
+    {
+      watchLoading: true,
+      undefinedResultIfReRun: true,
```
sidmorizon: # 🟠 [P2] Blanking the reserve on every 30s poll disables Max/Confirm and flickers validation

## Problem

This hook combines `undefinedResultIfReRun: true` (`:460`) with `pollingInterval: 30_000` (`:462`). On every poll tick, `usePromiseResult.ts:279-281` calls `setResult(undefined)` before re-fetching, so `withdrawReserve` is `undefined` for the entire duration of the network round-trip — once every 30 seconds, for a value that is effectively static between polls.

Three consequences follow each tick:

1. `isWithdrawReserveReady` requires `Boolean(withdrawReserve)` (`:471`), so `isWithdrawDestinationReady` (`:486`) goes false → `withdrawSubmitDisabled` (`:2395-2399`) disables Confirm on both the web button (`:2684`) and the native keypad CTA (`:2698`), plus the Max button (`:2620`).
2. `maximumWithdrawAmountBN` falls back to `withdrawReserve?.reserve ?? 0` (`:1042`), i.e. the **full** balance, so `isInsufficientBalance` (`:1713-1721`) and `isValidAmount` (`:1092-1093`) briefly stop flagging an over-max amount — the "insufficient balance" message at `:1731-1735` disappears and reappears.
3. A Max tap landing in that window is swallowed with no feedback by the new guard at `:1379` (`if (selectedAction === 'withdraw' && !isWithdrawDestinationReady) return;`).

## Impact

On the withdrawal form the primary CTA and Max blink disabled roughly every 30 seconds of user dwell time, and the validation error text flickers along with them. There is no correctness gain from the blanking: cross-account staleness is already guarded by the explicit address check at `:475-476`, which compares the reserve's `accountAddress` against `selectedAccount.accountAddress`.

## Recommendation

Drop `undefinedResultIfReRun: true` at `DepositWithdrawModal.tsx:460` so the last known reserve stays in place during revalidation. Identity safety is already covered by the address check at `:475-476` and by `selectedAccount.accountId` being a hook dep (`:457`). If a freshness bound is genuinely wanted, return a timestamp from `getUsdcWithdrawReserve` and reject a stale reserve inside `isWithdrawReserveReady`, rather than blanking the result on every tick.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1904,6 +1906,38 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+        !/^\d+(\.\d+)?$/.test(fee) ||
+        !new BigNumber(fee).isFinite() ||
+        result.isSanctioned !== false
+      ) {
+        return undefined;
+      }
+      return {
+        accountAddress,
+        // The cent is a minimum buffer, not an extra charge on top of the fee.
+        reserve: BigNumber.maximum(fee, USDC_WITHDRAW_GAS_RESERVE).toFixed(),
```
sidmorizon, line 1933: # 🟠 [P2] Dynamic reserve is withheld from Max while the fee row still discloses a hard-coded `< $0.01`

## Problem

The reserve deducted from Max is now `max(apiFee, USDC_WITHDRAW_GAS_RESERVE)` (`:1933`), but the fee quote the UI renders was not updated alongside it. Both `getUsdcWithdrawFee` and `getLiveUsdcWithdrawFee` still emit `amount: USDC_WITHDRAW_GAS_RESERVE.toString()` with `displayAsLessThan: true` for the `hyperEvmGas` component (`packages/kit-bg/src/services/ServiceHyperLiquid/cctpWithdraw.ts:158-165` and `:187-194`), and `formatWithdrawFeeComponent` renders that as `< $0.01` (`DepositWithdrawModal.tsx:140-148`, consumed at `:2389-2393`).

Before this PR both the Max deduction and the displayed fee read the same constant, so they could not disagree. After it they can.

## Impact

Whenever `preTransferCheck` returns a fee above 0.01 — which is the entire reason this PR exists, and is exercised by the PR's own test case `['1', '1']` at `ServiceHyperliquidExchange.reserve.test.ts:52` — Max quietly withholds up to that amount while the fee row still reads `< $0.01`. Funds are not lost, but the user cannot withdraw the last ~1 USDC and the UI offers no explanation for the gap, which is precisely the "Max doesn't withdraw everything" confusion the ticket set out to resolve.

This also contradicts the invariant the file documents at `DepositWithdrawModal.tsx:477-478`: *"Gated on the confirmed quote, never the preview: submitting against a fee the row never showed would take the difference out of the principal."*

## Recommendation

Feed the resolved reserve into the displayed quote for the `hyperEvmGas` component instead of the constant — either have `getUsdcWithdrawFee` / `getLiveUsdcWithdrawFee` (`cctpWithdraw.ts:150-175`, `:179-212`) accept the account and reuse `getUsdcWithdrawReserve`, or in the modal override the `hyperEvmGas` component's `amount` with `withdrawReserve.reserve` before `formatWithdrawFeeComponent`, dropping `displayAsLessThan` when the value is exact. Note that `formatWithdrawFeeComponent` also applies `.toFixed(2)` (`:143`), which would round a sub-cent dynamic fee — worth settling display precision in the same change.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -2617,6 +2656,25 @@ function DepositWithdrawContent({
           </XStack>
         </YStack>
 
+        {needsWithdrawReserve &&
+        isCheckingWithdrawReserve === false &&
+        !isWithdrawReserveReady ? (
```
sidmorizon: # 🟠 [P2] Retry banner is untranslated and misattributes an account-identity mismatch to a failed fetch

## Problem

Two distinct defects in this block:

**1. The condition is too broad.** The banner fires on any falsy `isWithdrawReserveReady` (`:2661`), but that flag folds together three independent conditions (`:469-476`): the reserve is missing, `activeAccount.accountId !== selectedAccount.accountId`, and `activeAccount.accountAddress !== selectedAccount.accountAddress`. Only the first is actually a fetch failure.

`selectedAccount` is a **one-shot snapshot** on the desktop/web entry point — `showDepositWithdrawDialog` does `const selectedAccount = await perpsActiveAccountAtom.get()` at `:3105` and bakes it into the JSX handed to `dialogInTab.show()` at `:3126`, so it never updates — while `activeAccount` at `:371` is live via `usePerpsActiveAccountAtom()`. If the perps active account changes while the dialog is open, the gate latches false for the life of the dialog: Confirm (`:2399`) and Max (`:2620`) stay disabled, and the user is shown a Retry button (`:2667-2674`) that re-runs a fetch which was never the problem. (The mobile entry at `:3012` reads the same atom, so those two comparisons are tautologies there — the gate only bites on the dialog path.)

**2. Untranslated copy.** `:2665` ships a raw English sentence with a `TODO(i18n)` at `:2663`, on the withdrawal money path, visible in every locale — while the sibling Retry button on the very next lines correctly uses `intl.formatMessage({ id: ETranslations.global_retry })`.

## Impact

A non-actionable, untranslated dead-end on a withdrawal form. The only recovery from the latched state is closing and reopening the dialog, and the copy points the user at the wrong cause — they will press Retry repeatedly with no effect. Non-English users see an English-only error string at the moment they are trying to move funds.

## Recommendation

Split the flag: make `isReserveFetchFailed = needsWithdrawReserve && !withdrawReserve` the *only* condition that renders the retry banner at `:2659-2661`, and handle the identity mismatch separately — either re-read the live atom for `selectedAccount` on the dialog path (`:3105` / `:3126`) so it tracks account switches, or render a distinct "account changed, please reopen" state with no Retry affordance. Replace the literal at `:2665` with an `ETranslations` id before merge rather than shipping the TODO.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/hooks/usePromiseResult.ts
Minnzen: Updated in c909c10332: I removed the shared usePromiseResult polling change entirely. The withdrawal dialog now owns its 30-second timer, while the existing hook handles focus and reconnect revalidation without a polling chain. This avoids changing debounced consumers elsewhere.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed in c909c10332. The reserve result is retained during periodic revalidation, so Max, validation, and Confirm stay stable while the refresh is in flight. Only a reserve matching the selected account address is used for arithmetic or the fee row.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
Minnzen: Fixed in c909c10332. The dialog fee row now includes the resolved reserve for CCTP and replaces the HyperEVM preview with it. A 0.01 minimum remains < $0.01; larger account-specific values are shown at display precision. While the reserve is unknown, the row shows its existing skeleton.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed in c909c10332. Account changes and reserve-fetch failures use separate states: account changes show the existing localized trading-account message without Retry, while a missing or mismatched reserve after a completed request shows the existing localized generic error text and Retry. The raw English TODO was removed.

packages/kit/src/views/Perp/components/TradingPanel/modals/withdrawFeeDisplayUtils.ts
```diff
@@ -0,0 +1,47 @@
+import { BigNumber } from 'bignumber.js';
```
sidmorizon, line 1: # 🔴 [P1] New module is unregistered in the mobile bundle registry, breaking the native build

## Problem

`withdrawFeeDisplayUtils.ts` is added by `cbbed69` as a new module, but `apps/mobile/bundle-registry/module-id-registry.json` was not regenerated to include it. The **Native startup graph budget (24.x)** check fails on this head:

```
packages/kit/src/views/Perp/components/TradingPanel/modals/withdrawFeeDisplayUtils.ts.
```

Verified locally at `cbbed69`: the identifier `withdrawFeeDisplayUtils` does not appear anywhere in `module-id-registry.json`, the PR diff touches no file under `apps/mobile/bundle-registry/`, and every sibling module in this same directory (`DepositWithdrawModal.tsx`, `OrderConfirmModal.tsx`, `LeverageAdjustModal.tsx`, `AccountModeModal.tsx` and the rest) *is* registered — so registration is the established requirement for this path, not an optional step.

## Impact

Hard build failure, not a flake. The union build gets through both graphs (main 14,749 modules, background 14,647) and then dies in `buildSegmentAllocation` before producing any bundle — the run uploads no allocation or budget report at all. The iOS/Android bundle cannot be produced from this commit, so the PR is not mergeable in its current state regardless of the review outcome on the other threads. Round 1's head (`b3ace18`) passed this check; it broke when `cbbed69` introduced the new file.

## Recommendation

Run the command the error names and commit the result:

```
```

Commit the regenerated `apps/mobile/bundle-registry/module-id-registry.json` as part of this PR. Regenerate it with that tooling rather than hand-editing the JSON, so the assigned module id stays consistent with the allocator. This applies to any further new modules added in later rounds of this PR.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/hooks/usePromiseResult.ts
```diff
@@ -395,11 +395,25 @@ export function usePromiseResult<T>(
...
+          optionsRef.current.debounced,
+          {
+            leading: false,
+            trailing: true,
+          },
+        );
         return async (config?: IRunnerConfig) => {
+          // A manual refresh may replace the trailing args, but the recovery
+          // request must retain ownership of the replacement polling chain.
+          pendingPollingNonce = config?.pollingNonce ?? pendingPollingNonce;
```
sidmorizon: # 🟠 [P2] The debounce rescue preserves only `pollingNonce`; `triggerByDeps` is still destroyed the same way

## Problem

The polling fix is correct — I walked the interleavings and exactly one chain survives in each (recovery + bare run in one window, two recoveries in one window, the self-scheduling continuation, a bare run after the nonce is consumed, and unmount). But the rescue is field-specific: `:416` retains only `config?.pollingNonce`, and the debounced body at `:403-405` reconstructs `{ ...config, pollingNonce }`, where `config` is still lodash's *last* args.

A deps-change dispatch is `run({ triggerByDeps: true, pollingNonce })` (`usePromiseResult.ts:473-476`). If a bare `run()` lands in the same debounce window, it replaces the trailing args; `pollingNonce` is now restored, but `triggerByDeps` is dropped — exactly the failure mode that was just fixed for the other field.

## Impact

Losing `triggerByDeps` skips the `if (config?.triggerByDeps && !isFocusedRef.current)` branch at `:252-261`, which does two things when the route is blurred:

1. It sets `isDepsChangedOnBlur.current = true`, which is what makes the refocus path at `:556` re-run for the new deps.
2. It bumps `nonceRef.current`. The comment at `:255-260` spells out why that is required: without it, the *previous* scope's in-flight request still satisfies the mount-only `shouldApplyResult` (`:228-237`) and overwrites the new scope's init/cached state.

With `pollingInterval` set, the chain's own sleep path sets `isDepsChangedOnBlur` late (`:390`), masking (1) but not (2). With no `pollingInterval`, neither recovers. Reachable on the same debounced consumers as the original finding — e.g. `TokenListBlock.tsx` with its bare `run()` calls at `:2930`, `:2985`, `:3259`.

## Recommendation

Preserve the superseded config rather than one key. Hold the pending config alongside the nonce and merge it in the debounced body, OR-ing the boolean:

```ts
let pendingConfig: IRunnerConfig | undefined;
// in the wrapper:
pendingConfig = config?.triggerByDeps ? { ...pendingConfig, ...config } : pendingConfig;
// in the debounced body:
return runner({
  ...config,
  triggerByDeps: config?.triggerByDeps || pendingConfig?.triggerByDeps,
  ...(pollingNonce === undefined ? {} : { pollingNonce }),
});
```

Worth extending the new regression test at `usePromiseResult.test.tsx:753-776` with a deps-change variant that asserts the blurred-deps bookkeeping still happens when a bare `run()` shares the window.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/withdrawFeeDisplayUtils.ts
```diff
@@ -0,0 +1,47 @@
...
+  feeQuote: IUsdcWithdrawFeeQuote | undefined;
+  reserve: string | undefined;
+  includeReserve: boolean;
+}) {
+  if (!feeQuote) return undefined;
+
+  const feeParts = feeQuote.components
+    .filter(
+      (component) =>
+        !includeReserve || !reserve || component.kind !== 'hyperEvmGas',
```
sidmorizon: # 🟠 [P2] The `!reserve` escape hatch still renders the hard-coded `< $0.01` while the reserve is unknown

## Problem

The add-vs-replace split in this module is right: `hyperEvmGas` is the same quantity as the reserve so it is filtered out and replaced (`:31-34`, `:43-45`), while `cctpForwarding` is `debitedFrom: 'withdrawAmount'` and therefore genuinely additive. But the filter predicate at `:33` has a third disjunct, `!reserve`.

When `includeReserve` is true and `reserve` is `undefined`, that disjunct keeps the `hyperEvmGas` component, `:37-38` renders it as `< $${amount}`, and `:43` skips the reserve because `includeReserve && reserve` is falsy. The `amount` being printed is the hard-coded `USDC_WITHDRAW_GAS_RESERVE` from `cctpWithdraw.ts:158-165` — the exact constant the original finding was about.

## Impact

On the hyperEVM destination the fee row shows a confident `< $0.01` in two states where the true reserve may be ~1 USDC: the entire initial reserve fetch, and permanently after a failed fetch. The row's render condition in `DepositWithdrawModal.tsx:2645` is only `withdrawFeeQuote ? ... : <Skeleton/>` — it does not consult reserve readiness — so the placeholder is shown rather than the skeleton that already exists for exactly this "not known yet" purpose.

This is display-only: `isWithdrawReserveReady` is false in both states, so Max (`:2622`) and Confirm (`:2397-2401`) are disabled and no incorrect amount can be submitted. Hence P2 rather than a repeat of the original severity.

## Recommendation

When `includeReserve` is true, drop `hyperEvmGas` unconditionally and signal "unknown" upward rather than substituting the constant:

```ts
.filter((component) => !includeReserve || component.kind !== 'hyperEvmGas')
```

and return `undefined` when `includeReserve && !reserve` leaves no parts, so `DepositWithdrawModal.tsx:2645` falls through to the existing `<Skeleton>` until the reserve resolves. That keeps the fee row honest in the one state where the number genuinely is not known yet.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -439,6 +430,48 @@ function DepositWithdrawContent({
...
+      activeAccount.accountAddress?.toLowerCase() !==
+        selectedAccount.accountAddress?.toLowerCase());
+  // The displayed balance follows the active account; its reserve must match.
+  const isWithdrawReserveReady =
+    !needsWithdrawReserve ||
+    (!hasWithdrawAccountChanged &&
+      Boolean(withdrawReserve) &&
+      withdrawReserve?.accountAddress.toLowerCase() ===
+        selectedAccount.accountAddress?.toLowerCase());
+  const hasWithdrawReserveFetchFailed =
```
sidmorizon, line 481: # 🟠 [P2] The reserve/account identity check is applied to readiness only, leaving a silent dead-end and unguarded Max arithmetic

## Problem

`isWithdrawReserveReady` (`:464-469`) is false under three conditions: the account changed, the reserve is missing, **or** `withdrawReserve.accountAddress !== selectedAccount.accountAddress`. The rewrite covers only the first two, and the third is now unguarded in two places.

**1. No banner for the address mismatch.** The previous condition was `!isWithdrawReserveReady`, which covered all three. `hasWithdrawReserveFetchFailed` (`:470-474`) now tests `!withdrawReserve` instead. So when a reserve *is* present but its address doesn't match while `hasWithdrawAccountChanged` is false, neither banner renders (`:2661`, `:2669`) yet `isWithdrawDestinationReady` is false (`:484`) — Max (`:2622`) and Confirm (`:2401`) are disabled with no message and no Retry. The query deps are `[needsWithdrawReserve, selectedAccount.accountId]` (`:448`), which exclude the address, so an address change that does not change the accountId does not even trigger a refetch to clear it.

**2. Max arithmetic ignores the address entirely.** `maximumWithdrawAmountBN` at `:1040` uses `withdrawReserve?.reserve ?? 0` with no identity check. Previously `undefinedResultIfReRun` blanked the result at dispatch (`usePromiseResult.ts:279-281`), so a foreign-account reserve was visible for at most one render. Removing that option (`:449-456` — correct, and what the other finding asked for) means the previous account's reserve now persists for the whole in-flight refetch after an account switch.

## Impact

For (1), a permanently disabled withdrawal form with no explanation — strictly worse than before the split, which at least offered a Retry. For (2), during the refetch window the new account's balance is reduced by the old account's reserve; that value feeds `isValidAmount` (`:1091`), `isInsufficientBalance` (`:1715-1719`), the error text (`:1728-1735`) and the fee row (`:2391`), so a typed amount can flip to a spurious "insufficient balance", or the ceiling can sit too high in the other direction. Submission stays blocked by `isWithdrawDestinationReady` throughout, so neither is a funds defect.

## Recommendation

Apply the identity check once and reuse it, rather than duplicating it per consumer:

```ts
const matchedWithdrawReserve =
  withdrawReserve?.accountAddress.toLowerCase() ===
  selectedAccount.accountAddress?.toLowerCase()
    ? withdrawReserve
    : undefined;
```

Use `matchedWithdrawReserve` at `:1040` and `:2391` so the arithmetic and the fee row can never see a foreign reserve, and restore full banner coverage with

```ts
const hasWithdrawReserveFetchFailed =
  needsWithdrawReserve &&
  !hasWithdrawAccountChanged &&
  isCheckingWithdrawReserve === false &&
  !isWithdrawReserveReady;
```

which keeps the account-changed/fetch-failed separation intact while making the mismatch case retryable again. Adding `selectedAccount.accountAddress` to the deps at `:448` would additionally make the mismatch self-clearing.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/withdrawFeeDisplayUtils.ts
Minnzen: Fixed in c909c10332. When the selected route needs a reserve and the account-specific value is unknown, the fee formatter returns undefined and the row renders its existing skeleton. The hard-coded HyperEVM preview is filtered out whenever the reserve is required.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Fixed in c909c10332. A single matchedWithdrawReserve is now used for readiness, Max arithmetic, and the fee row; an address mismatch shows the localized retry state rather than a silent disabled form. The selected address is also a request dependency so a changed address triggers a fresh check.

packages/kit/src/hooks/usePromiseResult.ts
Minnzen: Resolved in c909c10332 by reverting the shared usePromiseResult changes, including the debounce rescue. The withdrawal dialog now owns a local 30-second timer and retains the hook's existing focus/reconnect handling, so triggerByDeps behavior for other consumers is unchanged.

packages/kit/src/views/Perp/components/TradingPanel/modals/withdrawFeeDisplayUtils.ts
Minnzen: Fixed in c909c10332 by moving the small fee formatter back into DepositWithdrawModal.tsx and removing the new production module. The final PR adds no module that requires a mobile registry entry; the native startup check is rerunning on this head.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -439,6 +458,67 @@ function DepositWithdrawContent({
...
+      watchLoading: true,
+      undefinedResultIfError: true,
+      overrideIsFocused: (isFocused) => isFocused && needsWithdrawReserve,
+      revalidateOnFocus: true,
+      revalidateOnReconnect: true,
+    },
+  );
+  useEffect(() => {
+    if (!needsWithdrawReserve) return undefined;
+    const interval = setInterval(
```
sidmorizon: # 🟠 [P2] The hand-rolled `setInterval` drops the visibility gate `pollingInterval` had, so the dialog now polls while the app is hidden or backgrounded

## Problem

Scoping the refresh to the dialog is the right call and the timer itself is sound — cleanup is correct, `refreshWithdrawReserve` is stable so the effect doesn't churn, the callback reads `methodRef.current` so it can't capture a stale account, and stale results are dropped by the nonce check plus `matchedWithdrawReserve`. I could not find a leak, an overlap, or a wrong-account landing.

What was lost in the swap is the visibility gate. The removed `pollingInterval` scheduled its next tick through `usePromiseResult.ts:378-379`:

```ts
await timerUtils.wait(pollingInterval);
await defer.promise;
```

That `defer` is resolved and reset by `onVisibilityStateChange` (`usePromiseResult.ts:99-112`), so the polling chain **blocked entirely while the app, tab, or window was not visible**. The new `setInterval` at `:497` has no equivalent, and the only thing between a tick and a live RPC is `shouldSetState`'s focus check (`usePromiseResult.ts:212`), which reads `isFocusedRef.current` — ultimately `useRouteIsFocused`. I checked that hook (`packages/kit/src/hooks/useRouteIsFocused.ts`): it is `!isLocked && useIsFocused() && rootRoutersLength >= getRootRoutersLength()` — navigation focus, app-lock and a modal-stack guard, with **no visibility or AppState term at all**. Visibility lives in a separate module, `packages/shared/src/utils/appVisibility.ts`, which is exactly what `defer` was wired to.

So the focus gate does not compensate for the gate that was removed.

## Impact

With the withdraw dialog mounted on a still-navigation-focused route, a `preTransferCheck` RPC now fires every 30s indefinitely when the Electron window is merely blurred, when the browser tab is hidden (throttled to ~60s, not stopped), and on Android with the app backgrounded. All of these paused under the previous implementation. That is background network and battery cost on a wallet, and because it only manifests when the app is *not* in the foreground it will not show up in ordinary manual testing.

## Recommendation

Gate the tick on visibility rather than relying on the focus signal, which does not carry it. `isAppVisible` is already exported from `@onekeyhq/shared/src/utils/appVisibility` (`:33`) and importing it from `kit` is within the hierarchy:

```ts
const interval = setInterval(() => {
  if (!isAppVisible()) return;
  void refreshWithdrawReserve();
}, WITHDRAW_QUOTE_REFRESH_INTERVAL_MS);
```

Worth noting for the record: `pollingInterval` is a per-call option and never leaked into other consumers, so reverting to it would also restore this gate for free — the shared-hook changes that needed reverting were the nonce/debounce edits, not the use of `pollingInterval` itself.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -148,6 +148,24 @@ function formatWithdrawFeeComponent(
   return `${component.isEstimate ? '≈ ' : ''}$${amount}`;
 }
 
+function formatWithdrawFeeText(
```
sidmorizon: # 🟠 [P2] Inlining the fee formatter deleted its tests and left the add-vs-replace rules unverifiable

## Problem

Moving the formatter back inline resolved the registry blocker, but it also removed `withdrawFeeDisplayUtils.test.ts` (−62) without relocating its cases. That file pinned the three non-obvious rules of this logic:

- `cctpForwarding` **adds** to the reserve → `'$1.00 + $0.20'`
- `hyperEvmGas` is **replaced** by it, not shown twice → `'$1.23'`
- the floor keeps its sub-cent qualifier → `'< $0.01 + $0.20'`

`formatWithdrawFeeComponent` (`:141`) and `formatWithdrawFeeText` (`:151`) are now module-private in a ~3,200-line `.tsx` component file with no `export`, so those cases cannot be restored from the sibling `DepositWithdrawModal.test.ts` — that suite only exercises exported helpers (`getPerpsDepositMinimumCheck`, `getPerpsDepositTokenDisplayList`, and similar). The net effect is that this round *rewrote* a money-display path and *removed* the coverage of it in the same commit.

## Impact

The add-vs-replace asymmetry is the subtle part of this logic and it is easy to break silently. Dropping the `!includeReserve ||` half of the filter at `:158` would double-count HyperEVM gas in the displayed fee; dropping the `lte(USDC_WITHDRAW_GAS_RESERVE)` check at `:161-163` would print a bare `$0.01` for a value that is really a floor rather than a quote. Neither would fail any test today, and both are exactly the class of mismatch the earlier rounds of this PR were about.

## Recommendation

Export both functions from `DepositWithdrawModal.tsx` — they are pure, take no React input, and need no other change — and port the three deleted cases into the existing `DepositWithdrawModal.test.ts`, which already imports from this directory:

```ts
export function formatWithdrawFeeComponent(...)
export function formatWithdrawFeeText(...)
```

Inlining the formatter was the right fix for the registry problem; it is only the missing `export` that cost the coverage, and restoring it costs nothing.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Accepted in 01ba291e32. The dialog timer now checks getCurrentVisibilityState() before refreshing, which is the same visibility signal used by usePromiseResult. The timer remains scoped to the dialog and does not issue background preTransferCheck requests while the app is hidden.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Accepted the coverage concern in 01ba291e32, but not the suggested export from the component. I tried that path: importing DepositWithdrawModal.tsx in Jest fails on the React Navigation and native-storage ESM dependency chain before any assertion runs. The formatter is now a small pure module in the same folder, four focused cases cover the add/replace, floor, and unknown-value rules, and its single mobile module ID entry was generated with the repository tool.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1904,6 +1906,38 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+    try {
+      const wallet =
+        await this.backgroundApi.serviceHyperliquidWallet.getOnekeyWallet(
+          params,
+        );
+      const accountAddress = await wallet.getAddress();
+      const result = await hyperLiquidApiClients.infoClient.preTransferCheck({
+        source: accountAddress,
+        user: HYPEREVM_SYSTEM_ADDRESS,
+      });
```
sidmorizon, line 1920: 🟠 P1: Dynamic reserve queries wrong direction

Severity: `severe`

`preTransferCheck` receives the wallet as `source` and the always-existing system address as `user`. It returns zero instead of the wallet’s activation fee, so Max withdrawals can still fail.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -439,6 +431,68 @@ function DepositWithdrawContent({
...
+      undefinedResultIfError: true,
+      overrideIsFocused: (isFocused) => isFocused && needsWithdrawReserve,
+      revalidateOnFocus: true,
+      revalidateOnReconnect: true,
+    },
+  );
+  useEffect(() => {
+    if (!needsWithdrawReserve) return undefined;
+    const interval = setInterval(() => {
+      if (getCurrentVisibilityState()) {
```
sidmorizon: # 🟠 [P2] `getCurrentVisibilityState()` as a skip-gate isn't equivalent to the defer it replaced: no resume-on-visible, and extension blur isn't covered

## Problem

The gate works for the three cases I named — Electron blurred (`desktopApi.isFocused()`), tab hidden (`document.visibilityState`), Android backgrounded (`AppState.currentState`) — and the interval correctly no-ops rather than firing. But `getCurrentVisibilityState()` is not the signal the old path used, and the difference bites twice.

**1. The old gate deferred; this one skips.** `usePromiseResult.ts:378-379` was `await timerUtils.wait(pollingInterval); await defer.promise;` — the tick *blocked* while hidden and fired the instant `onVisibilityStateChange` resolved the defer. The new gate drops the tick and waits for the next 30s boundary. Nothing else revalidates on a visibility edge: `revalidateOnFocus` keys off `isFocusedRefValue`, which is `useRouteIsFocused` (`packages/kit/src/hooks/useRouteIsFocused.ts`) = navigation focus + app-lock + modal-stack, with no visibility term; `revalidateOnReconnect` only fires on a `false→true` network transition.

**2. On extension the gate is blind to window blur.** `getCurrentVisibilityState()` falls through to `document.visibilityState === 'visible'` (`appVisibility.ts:55-57`). The subscription that drove the defer does more: `onVisibilityStateChange` installs `window` `focus`/`blur` listeners **only** for `platformEnv.isExtension` (`appVisibility.ts:86-96`). A side-panel or expanded-tab UI that stays `document`-visible while the user switches to another application passes this gate but would have reset the old defer. `isAppVisible()` does not have this gap — its cached `_visible` is maintained by that same subscription (`appVisibility.ts:16-36`), which is exactly why the two functions differ. The file's own header documents the intended extension signal as "document.visibilitychange + window focus/blur".

## Impact

For (1): after the app is hidden for N minutes and comes back, the reserve is N minutes stale for up to 30s. `matchedWithdrawReserve?.reserve` feeds `maximumWithdrawAmountBN` and gates readiness, so a user who returns and immediately taps Max can submit against a reserve that no longer covers the transfer requirement — the precise failure this reserve exists to prevent. That window was previously zero. For (2): the 30s `preTransferCheck` RPC keeps firing for an unattended extension window, so the "backgrounded" case is still unfixed on that target.

## Recommendation

Both fall out of one change — drop the hand-rolled interval and pass the option to the existing `usePromiseResult` call:

```ts
pollingInterval: WITHDRAW_QUOTE_REFRESH_INTERVAL_MS,
```

That restores focus gating, the visibility *defer* (so it resumes on the rising edge) and correct extension behavior in one line, and is the pre-PR behavior. `pollingInterval` is a per-call option that never leaked to other consumers — the shared-hook edits that needed reverting were the nonce/debounce changes, not this.

If you prefer to keep the local timer, use `isAppVisible()` instead of `getCurrentVisibilityState()` (it is the subscription-derived state, so extension blur is covered and native/desktop/web are unchanged) and add a rising-edge refresh:

```ts
useEffect(() => {
  if (!needsWithdrawReserve) return undefined;
  return onVisibilityStateChange((visible) => {
    if (visible) void refreshWithdrawReserve();
  });
}, [needsWithdrawReserve, refreshWithdrawReserve]);
```

Separately, and **not blocking**: the sibling 30s timers in this component for `loadWithdrawRoute` and `loadWithdrawFee` are still ungated. I verified both are unchanged from base `ede1482e`, so they are pre-existing rather than this PR's regression — noting only because they cap the practical benefit of gating this one.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
sidmorizon: Corroborating this, with a correction to the diagnosis and a caution about the proposed fix.

**The effect is real.** Hyperliquid's docs define `preTransferCheck` as pre-flighting a transfer *to a destination*: `user` is the destination address, `source` the sender, and the returned `fee` is the destination's activation fee — "new HyperCore accounts require 1 quote token (e.g. 1 USDC) of fees for the first transaction which has **the new account as destination address**". This repo's own constant says `HYPEREVM_SYSTEM_ADDRESS` "receives every Core -> HyperEVM send" (`perp.constants.ts:66-67`), so that destination has certainly transacted and cannot be un-activated. Its activation fee is structurally always `0`.

That makes `BigNumber.maximum(fee, USDC_WITHDRAW_GAS_RESERVE)` at `ServiceHyperliquidExchange.ts:1933` collapse to `max(0, 0.01)` = `0.01` on every account — byte-for-byte the hard-coded constant this PR set out to replace. The dynamic reserve is a no-op, and the CCTP-to-Base failure in the issue would still reproduce.

**But the parameters are not swapped.** `user: HYPEREVM_SYSTEM_ADDRESS` / `source: accountAddress` is the correct *encoding* of "wallet sends to the system address" — that genuinely is the transfer being made. The bug is that this transfer's activation fee is not the fee that was failing. So I'd caution against the implied fix of passing the wallet as `user`: the withdrawing wallet demonstrably exists on HyperCore (it holds the USDC being withdrawn), so that query returns `0` as well. Neither direction produces the 1 USDC.

The likely conclusion is that `preTransferCheck` is the wrong source for this number and the real charge is something else — the reported error was `Insufficient USDC or HYPE balance for token transfer gas`, which points at HyperEVM gas rather than a HyperCore activation fee. Worth settling what is actually debited before re-pointing the query.

**Note on verification:** the 56 passing tests cannot catch this. `ServiceHyperliquidExchange.reserve.test.ts` mocks `preTransferCheck` and asserts `toHaveBeenCalledWith({ source, user })` (`:63`, `:93`) — it re-asserts the shape the implementation passes, so it stays green whichever address occupies which field, and the `['1','1']` fixture supplies a 1 USDC fee the live endpoint would never return here. The two unchecked items in the PR's test plan — reproducing the original CCTP-to-Base conditions and regressing the Arbitrum path — are what would actually settle this.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

Minnzen: Not adopting this change. I checked the two reported wallets against the live Hyperliquid /info preTransferCheck endpoint in both directions. For the failing wallet (0x0889…7063), source=wallet and user=HyperEVM system returns fee=1.0; reversing them returns fee=0.0. The successful wallet returns fee=0.0 with source=wallet. Reversing the parameters would therefore under-reserve the failing case. The current direction is also pinned by ServiceHyperliquidExchange.reserve.test.ts.

sidmorizon: That settles it — withdrawing my objection. Live responses from both wallets in both directions are decisive, and they beat the inference I made from the documented semantics. `fee=1.0` for the failing wallet on the current direction and `0.0` reversed is exactly the discriminating test, and it shows the reserve is not a no-op. My reasoning that the system address can never be un-activated, and therefore always returns `0`, was wrong against the real endpoint.

One narrow correction so it isn't relied on later: `ServiceHyperliquidExchange.reserve.test.ts` does not pin the direction. It mocks `preTransferCheck` and asserts `toHaveBeenCalledWith({ source, user })` (`:63`, `:93`), so it re-asserts whatever shape the implementation passes and stays green if the two addresses are swapped. Your live check is the evidence here; the test only locks the call shape against accidental edits.

Given the endpoint returns a real per-account value, it may be worth capturing the two observed responses in the test fixtures or a comment at the call site — the next reader will have the same "surely the system address always exists" reaction I did.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Accepted in 3257a62bf8. The dialog keeps its local timer, but now tracks onVisibilityStateChange events: hidden ticks are skipped, the first visible transition refreshes immediately, and extension window blur updates the same local visibility state. This avoids reintroducing the shared usePromiseResult polling-chain behavior. The pre-existing route/fee timers are outside this PR scope.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
Minnzen: Agreed on the correction: the unit test locks the current call shape, but the live endpoint comparison is the evidence that the direction yields the relevant fee. I will keep the observed wallet responses in the investigation record rather than embed user addresses or time-sensitive API values in production code.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -439,6 +430,59 @@ function DepositWithdrawContent({
...
+      overrideIsFocused: (isFocused) => isFocused && needsWithdrawReserve,
+      revalidateOnFocus: true,
+      revalidateOnReconnect: true,
+    },
+  );
+  const hasWithdrawAccountChanged =
+    needsWithdrawReserve &&
+    (activeAccount.accountId !== selectedAccount.accountId ||
+      activeAccount.accountAddress?.toLowerCase() !==
+        selectedAccount.accountAddress?.toLowerCase());
```
sidmorizon, line 471: 🟠 P1: Account switches can redirect withdrawals

Severity: `severe`

When an external wallet changes address after confirmation, `hasWithdrawAccountChanged` cannot stop the in-flight withdrawal. Funds can leave the new address using the previous address's amount and reserve.

sidmorizon: I traced this end to end and do not think it holds as described. The redirection it claims cannot occur, and the residual window is pre-existing and untouched by this PR.

**Destination is always the signer, re-resolved at submit time.** `withdraw()` receives only an `accountId` (`DepositWithdrawModal.tsx:1659-1660`), and the background resolves the address itself via `wallet.getAddress()` (`ServiceHyperliquidExchange.ts:1979-1983`). All three send branches derive the destination from that same resolved `userAddress`:

- hyperEvm — `destination: HYPEREVM_SYSTEM_ADDRESS` (`:2030`), a constant that credits the **signer's own** HyperEVM address
- cctp — `buildCctpWithdrawDestination({ ownerAddress: userAddress })` (`:2042-2044`)
- bridge — `destination: userAddress` (`:2059`)

No branch takes a user-typed address or the address the UI froze. Destination ≡ signer, structurally. So if an external wallet swapped from A to B mid-flight, the funds would leave B and land at **B's own address**, and only after B's holder approves the EIP-712 payload in their own wallet — `ServiceHyperliquidWallet.ts:164-181` re-resolves the address and calls `serviceSend.signMessage({ accountId })`, so no signature can be produced for B without B's holder confirming the exact destination and amount. The stale value that survives is the *amount*: too large and Hyperliquid rejects it, too small and the user withdraws less than max. Neither loses funds, and neither sends anywhere the user does not control.

**This PR does not touch the submit path.** I diffed against base `78cbfd90`: `ServiceHyperliquidExchange.ts` changes are two imports plus the added `getUsdcWithdrawReserve` method; `withdraw()`, the route check, the forwarding-fee validation and all three send branches are byte-identical to base. `ServiceHyperliquidWallet.ts` is untouched entirely. The `withdraw({...})` call in `handleConfirm` is character-for-character the same as base; the only change inside that function is `void refreshWithdrawReserve()` added to the catch block.

**The anchor line is a net-new guard being blamed for a pre-existing gap.** At base there was *no* account-change gate in `handleConfirm` at all. `hasWithdrawAccountChanged` and the `matchedWithdrawReserve` address-equality check (`:473-478`) are guards this PR adds, and they narrow the window rather than widen it. A readiness gate that never claimed to abort in-flight submissions is not a regression.

If the team still wants belt-and-braces on the pre-existing window, the correct repair is to pass the UI-validated address into `withdraw()` and assert it against `wallet.getAddress()` before signing (`ServiceHyperliquidExchange.ts:1983`) — a change to the submission path, and explicitly outside this PR's stated scope.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
```diff
@@ -1001,17 +1046,19 @@ function DepositWithdrawContent({
     [availableBalance.balance],
   );
 
-  // Hyperliquid may charge the sub-cent Core -> EVM gas outside the requested
-  // amount. Reserve one cent so a full-balance withdrawal does not fail.
   const maximumWithdrawAmountBN = useMemo(() => {
     const maximum = shouldReserveWithdrawGas
-      ? availableBalanceBN.minus(USDC_WITHDRAW_GAS_RESERVE)
+      ? availableBalanceBN.minus(matchedWithdrawReserve?.reserve ?? 0)
```
originalix, line 1051: P2: [Revalidate the account-specific reserve when submitting]

If `preTransferCheck` returns a higher reserve after this form refresh, Max is calculated from the older value and the user can submit an amount that no longer leaves enough USDC for the Core-to-EVM transfer. The `withdraw()` request only carries the route and CCTP fee, so the background path never detects this reserve change before signing.

Re-run the pre-transfer check in the background immediately before `sendAsset`/`sendToEvmWithData` and reject with a retryable reserve-changed error when `amount + liveReserve` exceeds the live withdrawable balance.

sidmorizon: This turns entirely on whether the reserve can ever *increase*, so it may help to separate what is established from what isn't. I've verified the surrounding mechanics but not that premise.

**Established, and it bounds the downside.** The reserve is `BigNumber.maximum(fee, USDC_WITHDRAW_GAS_RESERVE)` (`ServiceHyperliquidExchange.ts:1930-1934`), so it is always ≥ `0.01`. Base shipped `availableBalanceBN.minus(USDC_WITHDRAW_GAS_RESERVE)` with no refresh mechanism at all. A stale cached reserve is therefore always ≥ what ships today, so the worst case of not revalidating is exactly current `x` behavior — a rejected submission — and `handleConfirm`'s catch path already calls `refreshWithdrawReserve()` (`DepositWithdrawModal.tsx:1682`) before rethrowing, so the retry is primed with the fresh value. This is not a path to under-reserving below the status quo, and not a funds-loss path.

**Not established: monotonicity.** The PR's stated rationale is that the fee is monotone non-increasing per account (1 → 0 once the wallet signs anything), which is what makes the cache safe by construction. That is a claim about Hyperliquid's server-side semantics, and I could not confirm it from this repo — there is no `node_modules` here, the SDK is mocked in the tests, and this PR contains the only `preTransferCheck` call site. The live `/info` check in the PR description establishes the *direction* of the query, not that the value never rises. So your premise ("if it returns a higher reserve") is not excluded by anything in the code — it rests on an assumption that is reasonable but unverified.

Worth noting the destination is a constant (`user: HYPEREVM_SYSTEM_ADDRESS`, `perp.constants.ts:68-69`), so the only input that can move is the source wallet's own activation state. That is what makes a rise implausible, and it is also why route and destination switches cannot change the value — only an account change can, and that already re-fetches via the hook deps (`:454-458`).

So the decision is narrower than it first looks: not "is this safe" (it is no worse than today, in the worst case), but "is the monotonicity assumption solid enough to rely on". If it is, no change is needed and a pre-sign revalidation would add a round trip to every withdrawal for a window that cannot occur. If nobody can confirm it against Hyperliquid, your suggestion is the right hardening — though it is a change to the submission path, which this PR has deliberately left untouched, so it may sit better as a follow-up than as a condition on this one.

[Claude session ↗](https://claude.ai/code/session_01DSmvTGrzxLnk9AY2dHsf7Q)

---

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: No change. I verified the three send branches independently: `hyperEvm` sends to the constant `HYPEREVM_SYSTEM_ADDRESS`, `cctp` uses `buildCctpWithdrawDestination({ ownerAddress: userAddress })`, and `bridge` uses `destination: userAddress` — all derived from `wallet.getAddress()` resolved in the background at submit time, never from an address the UI is holding. The destination is therefore always the signer.

This PR's service diff is two imports plus the new `getUsdcWithdrawReserve` method; `withdraw()` and all three send branches are unchanged from base. The account-change gate this finding anchors on is net-new here and narrows the window rather than widening it.

packages/kit/src/views/Perp/components/TradingPanel/modals/DepositWithdrawModal.tsx
Minnzen: Not changing the submission path in this PR, and I have replaced the rationale in the description.

The argument that holds without any assumption about Hyperliquid's semantics: the reserve is `max(fee, 0.01)`, so a cached value is never below the fixed 0.01 that `x` applies today with no refresh mechanism at all. Not revalidating is therefore bounded by current behavior — Hyperliquid rejects the submission, `handleConfirm`'s catch path refetches the reserve, and the retry uses the fresh value. There is no path to under-reserving below the status quo.

The monotonicity claim has been dropped from the description: the response fields describe the constant `user` we pass, so they do not explain the per-source difference, and nothing in this repo establishes that the value cannot rise.

A pre-sign revalidation would add a round trip to every withdrawal for a window bounded by current behavior, on a submission path this PR deliberately leaves untouched. Tracking it as a follow-up rather than a condition on this PR.

originalix (approved):

PR #13468, opened 2026-09-16
fix: keep the device stage wall hittable on iOS (OK-63431)

franco-chan:
[OK-63431](<https://onekeyhq.atlassian.net/browse/OK-63431>)

----------

## Summary

On iOS the page behind the device stage stayed tappable during every undimmed state (BLE connecting, confirm-on-device, PIN, passphrase), so QA could re-trigger hardware calls while one was in flight (OK-63431). Android and desktop were unaffected.

## Root cause

Since #13298 (OK-62072 scrim) the touch-blocking wall in `MorphOverlay` shared one `Animated.View` with the scrim tint: `opacity = presence × scrimLevel`, which is `0` whenever `scrim` is off. React Native's iOS Fabric hit test (`RCTViewComponentView.betterHitTest`) returns `nil` for any view with `alpha < 0.01`, so the wall became invisible to touches and taps fell through to the page underneath. Android's `TouchTargetHelper` and the web ignore alpha, which is why only iOS leaked.

## Fix

- Split blocking from painting: the wall is now a plain static `View` (`pointerEvents="auto"`, `collapsable={false}`, `testID="morph-overlay-wall"`) that never carries an animated style, and the scrim tint moves to its own `Animated.View` with `pointerEvents="none"` that keeps the existing fade math untouched.
- `collapsable={false}` is required: Fabric flattens a background-less `pointerEvents="auto"` view out of the native tree, and a flattened wall takes no touches either.
- The unmount-at-hidden rule is unchanged (the web reanimated pointerEvents overwrite issue still applies).

## Tests

- New `MorphOverlay wall` cases in `index.dismiss.test.tsx`: the wall carries no `opacity`/`backgroundColor` while the tint does (both scrim grants), and the wall mounts only while `modal` and shown.
- `CI=1 yarn agent:check --profile commit` green; suite 22/22.

## Verification

- Debug build installed on a physical iPhone 17 Pro (iOS 26.5.2) for manual QA: tap the page behind the stage during the connecting capsule (should be blocked), then confirm the error / auth-failure / device-not-found cards still dim the app and stay tappable. Result to be recorded on OK-63431.

[OK-63431]: https://onekeyhq.atlassian.net/browse/OK-63431?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01SiGp23rfoxueunyZrrmqEN

originalix (dismissed):

packages/components/src/composite/MorphOverlay/index.dismiss.test.tsx
```diff
@@ -277,6 +278,40 @@ describe('MorphOverlay dismiss gesture', () => {
...
+describe('MorphOverlay wall', () => {
+  // OK-63431: RN's iOS hit test skips views whose alpha is under 0.01, so
+  // the wall must never wear the scrim's fading paint — it stays a static
+  // view of its own, and only the tint beside it carries color and opacity.
+  it.each([true, false])(
+    'keeps the blocking wall free of the scrim paint (scrim=%s)',
+    (scrim) => {
+      const { view } = setup({ modal: true, scrim });
+      const wall = view.getByTestId('morph-overlay-wall');
+      expect(wall.style.opacity).toBe('');
```
sidmorizon, line 292: # 🟡 P2 The wall's `opacity` assertion cannot fail

## Problem

`expect(wall.style.opacity).toBe('')` (line 290) is structurally vacuous under this file's own reanimated mock.

The mock at `index.dismiss.test.tsx:100` is `useAnimatedStyle: () => ({})`, so `scrimFadeStyle` (`index.tsx:1160-1167`) — the *only* source of `opacity` anywhere in this component's scrim path — always flattens to `{}` in this environment. `styles.backdrop` (`index.tsx:286-292`) is position/inset only and carries no `opacity`, and `scrimStyle` (`index.tsx:1230-1241`) adds opacity solely via `scrimFadeStyle`. No arrangement of this component can therefore produce an inline `opacity` on any element in this suite, so line 290 holds regardless of the implementation.

The sibling assertion on line 291 does have teeth: `{ backgroundColor: 'rgba(0,0,0,…)' }` is an ad-hoc object literal, and RNW's styleq leaves unregistered objects as inline style while `StyleSheet.create` entries become class names — which is also why `tint.style.backgroundColor` is non-empty on line 295.

## Impact

The test's own comment states the invariant as "the wall must never wear the scrim's *fading paint*", but only the color half is enforced. Re-attaching just the animated opacity to the wall while leaving the split and the `backgroundColor` where they are — e.g. `<View style={[styles.backdrop, scrimFadeStyle]} pointerEvents="auto" testID="morph-overlay-wall" />` — reintroduces OK-63431 on iOS (alpha back under 0.01, `betterHitTest` returns `nil`) and this suite stays green. The coarser regression of re-merging the two views back into one is still caught by lines 291/295, so this is a coverage gap on the precise alpha invariant, not a missing test.

## Recommendation

Give the mock a detectable animated style so "wears an animated style" becomes observable, without executing worklets. At `index.dismiss.test.tsx:100`:

```ts
useAnimatedStyle: () => ({ opacity: 0.42 }),
```

Then line 290 stays `expect(wall.style.opacity).toBe('')` — now genuinely failing if `scrimFadeStyle` migrates back onto the wall — and add `expect(tint.style.opacity).toBe('0.42')` next to line 295 to pin the tint as the style's only carrier. Ad-hoc objects render inline under RNW, the same mechanism that already makes line 291 meaningful.

If you would rather assert the real fade math, `useAnimatedStyle: (fn: () => unknown) => fn()` also works, but it additionally requires adding `interpolate` to the mock object (`index.dismiss.test.tsx:88-104`, where it is currently absent and only survives because the worklets are never invoked), and it starts running worklets in every other suite in this file — worth re-running all 22 cases if you take that route.

[Claude session ↗](https://claude.ai/code/session_01SiGp23rfoxueunyZrrmqEN)

---

franco-chan: Took the first recommendation in 0dc4ac9997: the reanimated mock now returns `{ opacity: 0.42 }` for every animated style, so `wall.style.opacity === ''` fails if `scrimFadeStyle` ever migrates back onto the wall, and the tint is pinned as the only carrier with `expect(tint.style.opacity).toBe('0.42')`. All 22 cases still pass.

sidmorizon (approved):

originalix (approved):

PR #13471, opened 2026-09-16
fix: sync TradingView web prices to token details OK-58635

limichange:
[OK-58635](<https://onekeyhq.atlassian.net/browse/OK-58635>)

----------

When TradingView Web loaded its latest candle, the token detail price could remain at the older API value until the next realtime tick because the app discarded the chart's initial price snapshot.

Sync that latest snapshot to token details and continue applying realtime updates. Ignore history once a valid realtime price has arrived, buffer the latest price until matching token details are ready, and use a strictly increasing local reception timestamp for the header cache. Clear pending prices and source state when the token, network, or update enablement changes. Token identity and valid-price checks remain in place. Resolves OK-58635.

Read token details directly from the market context store and replay pending prices through a store listener, so price updates do not subscribe the chart subtree to React renders. Activate each token's price-update state only after commit, preserving the current chart's updates when a concurrent token switch is interrupted.

Validation:
- 49 targeted Jest tests passed. The chart tests exercise real market actions and the header price cache, including delayed history, delayed token details, same-timestamp updates, token/network changes, disabled updates, chart render counts, suspended token switches, and unmount cleanup.

[OK-58635]: https://onekeyhq.atlassian.net/browse/OK-58635?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01U7Be5pWmmFzgGbAVjGr1zn

packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.tsx
```diff
@@ -167,10 +165,8 @@ export const MarketTradingView = memo(
         if (disableChartPriceUpdate) {
           return;
         }
-        if (data.source === 'history') {
-          return;
-        }
-
+        // History price events contain the latest bar from the initial chart
+        // load, so sync them too instead of waiting for the first realtime tick.
```
sidmorizon: 🟡 P2: Snapshot blocks live header prices

Severity: `non-severe`

When `handlePriceUpdate` applies history before a same-candle realtime tick, both carry the candle timestamp. `MarketTokenPrice` accepts only newer timestamps, so live prices cannot replace the snapshot until the next candle.

packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.tsx
```diff
@@ -161,16 +148,108 @@ export const MarketTradingView = memo(
...
+      priceUpdateStateRef.current.tokenAddress !== tokenAddress ||
+      priceUpdateStateRef.current.disabled !== disableChartPriceUpdate
+    ) {
+      priceUpdateStateRef.current = {
+        networkId,
+        tokenAddress,
+        disabled: disableChartPriceUpdate,
+        hasRealtimePrice: false,
+        lastUpdated: priceUpdateStateRef.current?.lastUpdated ?? 0,
+      };
```
sidmorizon: 🟡 P2: Interrupted renders disable price updates

Severity: `non-severe`

When React abandons an identity-changing render, `priceUpdateStateRef` retains that uncommitted identity. The committed `applyChartPrice` then rejects every tick because its captured state no longer matches the ref.

packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.tsx
```diff
@@ -161,16 +148,108 @@ export const MarketTradingView = memo(
   }: IMarketTradingViewProps) => {
     const { accountAddress } = useNetworkAccountAddress(networkId);
     const tokenDetailActions = useTokenDetailActions();
+    const [tokenDetail] = useTokenDetailAtom();
```
sidmorizon: # 📝 P2: Chart subtree re-renders on every price tick

## Problem

This line newly subscribes `MarketTradingView` to `tokenDetailAtom`. The same component also *writes* that atom on every accepted tick (`MarketTradingView.tsx:221` → `applyChartPriceUpdate`), and that action writes a new object **unconditionally** — the unchanged-price branch still does `set(tokenDetailAtom(), { ...tokenDetail, lastUpdated, chartPriceUpdatedAt })` (`packages/kit/src/states/jotai/contexts/marketV2/actions.ts:288-295`). Because the new monotonic scheme at `MarketTradingView.tsx:215-219` guarantees `lastUpdated` always advances, every tick produces a fresh object.

That closes a loop: tick → atom write → `MarketTradingView` re-render. The `memo()` at line 116 does not help, since the re-render comes from the atom subscription rather than from props.

The child absorbs the cost. `TradingViewV2` is exported un-memoized (`packages/kit/src/components/TradingView/TradingViewV2/components/tradingViewV2/TradingViewV2.tsx:168` — no `memo` anywhere in that file). The `webView` element itself sits behind a `useMemo`, so the iframe/WebView does not reload, but the rest of the subtree re-renders. Memoization further down cannot absorb it on desktop either, because `MarketTradingView.tsx:320-324` constructs a fresh `<MarketChartFullscreenHeader />` element on each render for the `!platformEnv.isNative && nativeControlsLayoutMode === 'desktop'` path. Price messages are not throttled upstream (`useTradingViewMessageHandler.ts:702` forwards each `TRADINGVIEW_PRICE_UPDATE` straight through).

Before this change `MarketTradingView` subscribed only to the stable actions ref and re-rendered zero times per tick.

## Impact

On a websocket-backed token the chart control bar and the rest of the `TradingViewV2` tree re-render at tick rate — potentially several times per second — for a state change that only the header actually consumes. Wasted render work and interaction jank on one of the most performance-sensitive screens in Market. No incorrect prices result; this is cost, not correctness, hence P2.

## Recommendation

Keep the detail subscription off the chart's render path. Either:

- Move the subscription plus the replay effect (`MarketTradingView.tsx:151-153` and `:237-241`) into a `null`-rendering sibling component — e.g. `<ChartPendingPriceSync state={priceUpdateState} onFlush={applyChartPrice} />` — so only that component re-renders per tick; or
- Read the detail imperatively from the marketV2 context store inside `applyChartPrice` instead of via `useTokenDetailAtom`, keeping `tokenDetailRef` fed without subscribing.

Separately worth considering (pre-existing, not introduced here): hoisting the fullscreen header element at `:320-324` out of the render body so the controls container can memoize at all.

[Claude session ↗](https://claude.ai/code/session_01U7Be5pWmmFzgGbAVjGr1zn)

---

packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.tsx
sidmorizon, line 1: 🟡 P2: Historical requests roll back price

Reported at `packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.tsx:272-275` (outside this PR's diff; file-level comment).

Severity: `non-severe`

Before `hasRealtimePrice` flips, a delayed older history callback replaces the current price. Panning or interval changes can leave the header showing a historical close.

packages/kit/src/views/Market/MarketDetailV2/components/MarketTradingView/MarketTradingView.test.tsx
```diff
@@ -0,0 +1,420 @@
...
+  it.each(['history', 'realtime'] as const)(
+    'does not let invalid %s prices prevent the initial snapshot',
+    (source) => {
+      const { store } = renderChart();
+      act(() => {
+        for (const price of ['', 'invalid', '0', -1, Number.NaN, Infinity]) {
+          mockOnPriceUpdate({ ...latestPrice, source, price });
+        }
+      });
+      expect(store.get(tokenDetailAtom())).toEqual(detail);
```
sidmorizon, line 424: # 📝 P2: Identity-ordering invariant lacks a regression test

## Problem

To be clear up front: the production ordering at this head is **correct**. In `MarketTradingView.tsx` the sequence is history gate (`:267`) → `isChartPriceUpdateForCurrentToken` (`:270-278`) → `normalizeChartPrice` validity (`:280-283`) → `hasAcceptedPrice = true` (`:285`) → `applyChartPrice` (`:286`). Both guards `return` before the flag is set. No defect today.

The gap is that only half of that invariant is pinned by a test. This new parameterized test proves an *invalid price* cannot set `hasAcceptedPrice` — it is load-bearing, because if any of the six invalid payloads set the flag, the trailing valid history bar would be gated out and `expectHeaderPrice('0.002930')` would fail.

Nothing makes the equivalent guarantee for a *wrong-token / wrong-network* payload. `ignores prices for another token or network` (`:393-403`) stops at `expect(store.get(tokenDetailAtom())).toEqual(detail)` and never feeds a subsequent valid bar, so it would still pass if `hasAcceptedPrice = true` were moved above the identity check at `:270`.

This increment is what makes that matter. Previously the flag was `hasRealtimePrice`, set only by realtime, so a stray payload setting it early would merely suppress history bootstrap. Now the flag latches on any accepted price and gates all history, so a single wrong-token payload setting it early would permanently prevent the chart from ever bootstrapping the header for that token.

## Impact

A future refactor that reorders `:270-285` would silently ship exactly the P1 this change is designed to avoid, with a fully green suite. The cost of the gap is only realized later, which is precisely when it is most expensive to diagnose.

## Recommendation

Extend the identity test at `:393-403` (or add the case to this parameterized block) to mirror the invalid-price pattern — feed the stray payloads, then a valid bar, and assert the header still bootstraps:

```ts
act(() => {
  mockOnPriceUpdate({ ...latestPrice, source, networkId: 'evm--1' });
  mockOnPriceUpdate({ ...latestPrice, source, tokenAddress: 'another-token' });
});
act(() => mockOnPriceUpdate(latestPrice));
expectHeaderPrice('0.002930');
```

[Claude session ↗](https://claude.ai/code/session_01U7Be5pWmmFzgGbAVjGr1zn)

---

originalix (approved):

PR #13472, opened 2026-09-16
fix: prevent duplicate hardware address verification on receive OK-63433

huhuanming:
[OK-63433](<https://onekeyhq.atlassian.net/browse/OK-63433>)

----------

https://onekeyhq.atlassian.net/browse/OK-63433

## Problem

On mobile, tapping **在设备上验证** twice in quick succession fires `verifyHWAccountAddresses` twice, so the device asks the user to confirm the address twice.

Nothing stopped it today:

- The native footer button (`ReceiveToken.tsx`) carries no `loading` / `disabled` state.
- `shouldShowAddress` stays `false` through `EAddressState.Verifying` until `hardwareUiState.action === REQUEST_BUTTON`, so the footer re-renders unchanged after the first tap.
- `useSharedPress` defaults to `onPressDebounce = 0`, which is lodash `debounce(fn, 0)` with `trailing: true` — that defers to a timer, it does not collapse taps.
- In the background, `runExclusiveOneKeyOperation` uses `Semaphore(1).runExclusive`, which **serializes** duplicates rather than dropping them, and `DeviceStageBurst`'s grace window lets the second call rejoin the stage.

The press stays live for the whole window because the device stage UI only covers the page once `prepareHardwareTransport` has picked the BLE transport — seconds after the press on native, where the `main` → `bg` bridge hop adds to it.

## Fix

Three entry points reach the same call — the footer button, the desktop/web `Page.Footer` confirm, and the QR placeholder card — so all three are guarded.

1. **Debounce the handler itself** (500ms, `leading: true`, `trailing: false`) instead of the `Button` `onPressDebounce` prop. The prop is consumed by `useSharedPress` and therefore never reaches the QR placeholder card, which is a plain `YStack`. Debouncing the handler gives both surfaces one shared window. `trailing: false` is deliberate: replaying a swallowed tap 500ms later would itself be the second hardware interaction.

   `useDebouncedCallback` keeps the function in a ref and memoizes on `[leading, maxing, wait, maxWait, trailing, useRAF]`, so the callback is stable and the window survives the re-render that `setAddressState(Verifying)` triggers — a `useMemo(() => debounce(fn), [fn])` would reset it at exactly the wrong moment.

2. **`loading` driven off `EAddressState.Verifying`** on the footer button (and `confirmButtonProps` on web). This covers the seconds-long transport window that a fixed debounce cannot, and supplies the feedback whose absence is why users tap a second time.

3. **Drop `onPress` from the QR placeholder card while verifying**, since the button's `loading` state does not reach it.

Also hoists the `currentDeriveType` / `displayAddress` guards above `setAddressState(Verifying)`. They used to return *after* it was set, which was merely cosmetic before, but would now strand the page in a permanent loading state.

## Notes

- **「我没有带设备」is intentionally left enabled.** It does not cause a duplicate hardware interaction, and it is the user's escape hatch if the call never settles.
- **Timeout / error coverage:** every rejection (SDK `DeviceMethodCallTimeout` at 60s, BLE connect timeouts, device disconnect, on-device cancel) reaches the existing `catch` and clears `Verifying`; a user-closed hardware dialog clears it via `CloseHardwareUiStateDialogManually`. The one uncovered case is a promise that never settles — queueing behind a wedged operation on the no-timeout `Semaphore(1)`, or the `bg` runtime dying. A watchdog was considered and left out: any threshold safe against the 60s SDK timeout would have to exceed it, and resetting early would re-open this very bug on slow devices. The skip button remains the exit.

## Verification

`yarn agent:check --profile pr` — all local checks PASS, including full-project `tsc --noEmit` and `react-hooks/exhaustive-deps`. The lint harness was confirmed non-vacuous by temporarily removing the new dependency entries, which produced the expected `missing dependencies` error.

Not verified on a physical device — needs QA on iOS/Android with hardware to confirm the device prompts exactly once under rapid tapping.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01KynSvPH5TGxciKqCUtSm9N

packages/kit/src/views/Receive/pages/ReceiveToken.tsx
```diff
@@ -393,6 +392,18 @@ function ReceiveToken() {
...
+
+  // Two surfaces start the same hardware call: the footer button and the QR
+  // placeholder card. On native the device stage UI only covers the page once
+  // the BLE transport is ready, seconds after the press, so the debounce
+  // collapses a rapid double tap and isVerifying holds the rest of that window.
+  const handleVerifyOnDevicePressDebounced = useDebouncedCallback(
+    handleVerifyOnDevicePress,
+    500,
+    { leading: true, trailing: false },
+  );
```
sidmorizon, line 434: # 🟡 P2 A stale attempt's settlement clears `isVerifying` while a newer verify is still in flight

## Problem

This PR makes `addressState === EAddressState.Verifying` the load-bearing guard for both entry points — `loading` on the footer button (`:719`) and on `confirmButtonProps` (`:748`), and the `onPress` removal on the QR placeholder card (`:932-933`).

But the two settlement paths write that shared state unconditionally, with no notion of *which* attempt is settling:

```ts
      setAddressState(                                    // :371-373
        isSameAddress ? EAddressState.Verified : EAddressState.Unverified,
      );
    } catch (e: any) {
      setAddressState(EAddressState.Unverified);          // :374-375
```

Meanwhile `addressState` is also reset out-of-band by the `CloseHardwareUiStateDialogManually` listener (`:407-419`), which `ServiceHardwareUI` emits on user close *independently of the caller's promise settling* — the emit at `packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts:913-916` runs after `cancelOneKeyOperation` / `serviceHardware.cancel` are kicked off with `void` (`:889-898`), not after `verifyHWAccountAddresses` rejects.

Interleaving on the cancel-then-retry path:

1. t=0 — press #1 → `Verifying`, promise P1 in flight.
2. t≈5s — user closes the device stage → event → `Unverified`. Footer un-loads, QR card regains `onPress`.
3. t≈5.1s — press #2 → `Verifying`, P2 in flight. The 500ms leading window is measured from the last call at t=0, so this tap is correctly *not* swallowed.
4. P1's rejection finally lands (BLE cancel propagation is not instant) → `:375` sets `Unverified` **while P2 is still running**.
5. Any tap ≥500ms after step 3 now starts P3 concurrently with P2 — the exact duplicate this PR set out to prevent.

The resolve branch at `:371` has the same shape: a late-resolving stale attempt can write `Verified`, which flips `shouldShowQRCode` (`:213-226`) to reveal the address and, when `exchangeSource` is set, fires the auto-navigation effect at `:273-299` — all off a superseded attempt.

## Impact

The new guard is defeatable in a narrow but reachable window, and cancel-then-retry is the most common retry path on hardware. Duplicate `verifyHWAccountAddresses` calls can still reach the device, and the page can transition to `Verified` / auto-navigate based on an attempt the user already abandoned.

## Recommendation

Make settlement attempt-scoped. Add `const verifyAttemptRef = useRef(0)` alongside the existing refs, capture `const attempt = (verifyAttemptRef.current += 1)` immediately before `setAddressState(EAddressState.Verifying)` at `:326`, and gate both writes — `:371` and `:375` — on `if (verifyAttemptRef.current !== attempt) return;` (keep the `throw e` at `:382` so `@toastIfError` behavior is unchanged). Also bump `verifyAttemptRef.current` inside the `CloseHardwareUiStateDialogManually` callback at `:408` and in the `AddressTypeSelector.onSelect` reset at `:839`, so an out-of-band reset invalidates the in-flight attempt.

A cheap re-entrancy check `if (addressState === EAddressState.Verifying) return;` at `:321` is also worth adding as belt-and-braces — the same file already uses that pattern for share at `:656` (`if (!shareData || isPreparingShare) return;`).

[Claude session ↗](https://claude.ai/code/session_01KynSvPH5TGxciKqCUtSm9N)

---

huhuanming: Confirmed and fixed in 0756f5f9.

I verified the premise before acting on it: `ServiceHardwareUI.ts:892-897` starts the device cancel with `void this.backgroundApi.serviceHardware.cancel({...})` and the `appEventBus.emit` at `:913` runs right after, so the event does land while the abandoned call is still unwinding. The interleaving is reachable as described.

What changed:

- Each press claims an id (`verifyAttemptRef`) alongside setting `Verifying`, and both settlement writes are gated on it still being current. A late rejection or a late resolve from a superseded attempt now writes nothing — so no un-gating mid-flight, and no `Verified` / `ExchangeOpenRedirect` off an abandoned attempt.
- All three out-of-band resets (stage close, `AddressTypeSelector.onSelect`, and the `btcUsedAddress` effect, which has the same shape and was not in the report) go through one `resetVerifyState` that bumps the id, so any reset invalidates what is in flight.
- Re-entrancy guard added, released in a `finally` that is likewise gated on the attempt id — a superseded attempt must not release the guard the live one holds.

One deviation from the recommendation, deliberate: I used a ref for the re-entrancy check rather than `if (addressState === EAddressState.Verifying) return;`. `handleVerifyOnDevicePress` is reached through `useDebouncedCallback`, which refreshes its inner function ref in a `useEffect` — i.e. after commit. In the only window the check matters (render delayed past the 500ms debounce), the closure's `addressState` and the UI gating are stale together, so the state-based form would read `Unverified` and wave the press through. A ref is written synchronously and does close it. The `isPreparingShare` precedent at `:656` works because it guards a plain `useCallback`, not one behind a debounce wrapper.

`yarn agent:check --profile pr` passes, including full-project tsc and `react-hooks/exhaustive-deps`. Still needs on-device QA for the cancel-then-retry path specifically.

sidmorizon (approved):

sidmorizon (approved):

originalix (approved):

PR #13473, opened 2026-09-16
fix: watchlist first paint and search-star logos OK-63468 OK-63469

zy2000-star:
[OK-63468](<https://onekeyhq.atlassian.net/browse/OK-63468>) [OK-63469](<https://onekeyhq.atlassian.net/browse/OK-63469>)

----------

## Summary

- **OK-63468**: Native Market Favorites waited on banner fetch, route focus, and quote returns before painting, so Android (and iOS) sat on a blank/skeleton for seconds. Native layout now mounts without waiting for banners, first-load quotes start without focus, spot/perps show identity rows while quotes load, and the list uses a skeleton instead of a blank screen.
- **OK-63469**: Starring a stock from search only persisted `stockId`. The watchlist icon comes from a later stock batch that is focus-gated, so AAPL first rendered `CryptoCoinOutline` until the user left and came back. Search/detail already have `logoUrl`; we remember that preview in memory until the batch quote arrives. Persistence/sync is unchanged.

Desktop/web first-paint path is unchanged. Logo preview is not written to DB/cloud.

## Test plan

- [ ] **Android**: open Discover → Favorites. Existing watchlist should appear quickly, then fill prices. Check mixed / spot-only / empty lists.
- [ ] **iOS**: same first-open Favorites path; header can still settle after banners.
- [ ] **Desktop (OK-63469)**: search `AAPL` → star the stock listing → open Favorites immediately. Apple logo should show on first paint, not the grey coin. Leave and return — still correct.
- [ ] Repeat search-star on Android. Star from token detail. Unstar/star again before quotes return.

[OK-63468]: https://onekeyhq.atlassian.net/browse/OK-63468?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63469]: https://onekeyhq.atlassian.net/browse/OK-63469?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,65 @@
...
+    watchlistItem.isNative,
+    watchlistItem.contractAddress,
+  );
+  const chainId = watchlistItem.chainId;
+  return {
+    ...PENDING_WATCHLIST_METRICS,
+    id: getMarketWatchlistKey(watchlistItem),
+    name: '',
+    symbol: '',
+    address: normalizedAddress,
```
sidmorizon: # 🟠 P1 Pending spot rows lowercase the contract address, breaking remove / move-to-top / detail nav on case-sensitive chains

## Problem

`buildPendingSpotWatchlistToken` sets `address: normalizedAddress`, and `normalizedAddress` comes from `getNativeTokenInfo`, which lowercases unconditionally:

```ts
const normalizedAddress = isNative ? '' : (address ?? '').toLowerCase();
```
(`packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/utils/tokenListHelpers.ts:28`)

But the watchlist identity is deliberately **case-preserving** for Solana/Sui/Ton/Aptos/Tron/STC:

```ts
export const caseSensitiveNetworkImpl = ['sol', 'stc', 'tron', 'aptos', 'sui', 'ton'];
...
if (caseSensitiveNetworkImpl.includes(impl)) { return contractAddress; }
return contractAddress?.toLowerCase();
```
(`packages/shared/src/utils/tokenUtils.ts:36-43`, `:844-847`, consumed by `getMarketWatchlistKey` in `packages/shared/src/utils/marketWatchlistIdentity.ts:9-14`)

Real (non-pending) rows keep the original casing — `transformApiItemToToken` returns `address: item.address` (`tokenListHelpers.ts:362`). Only the new pending row mangles it.

Downstream consumers treat `item.address` as the real identity:

- `MobileMarketNativeLists.tsx:853-855` → `removeFromWatchListV2(item.networkId, item.address, …)`, which re-normalizes and then filters with `getMarketWatchlistKey(item) !== getMarketWatchlistKey(identity)` (`packages/kit/src/states/jotai/contexts/marketV2/actions.ts:686-709`). For `sol--101` the lowercased address yields a different key than the stored favorite, so nothing matches.
- `tokenToWatchListItem(item)` → `moveToTopV2` (`MobileMarketNativeLists.tsx:783`, `:835`) has the same key mismatch.
- `toMarketDetailPage({ …item, tokenAddress: item.address })` (`MobileMarketNativeLists.tsx:905`).

This PR is what makes the window interactive: `showSkeleton = Boolean(result.isLoading) && result.data.length === 0` (`MobileMarketNativeLists.tsx:933`), and pending rows make `data.length > 0`, so the skeleton is suppressed and the rows are tappable. Previously this branch returned `undefined` and the rows did not exist.

## Impact

During first paint of Favorites on native — exactly the multi-second cold-start window this PR targets — long-press → "Remove from watchlist" and "Move to top" silently no-op for Solana/Sui/Ton/Aptos/Tron favorites (a success toast still shows, since `removeFromWatchListV2` resolves), and the backend removal is sent with a wrong address. Tapping the row navigates to a detail page for a mint that does not exist on those chains.

## Recommendation

Keep the stored address verbatim on the pending row at `watchlistPendingRows.ts:39`:

```ts
address: isNative ? '' : watchlistItem.contractAddress ?? '',
```

i.e. use `getNativeTokenInfo` only for the `isNative` flag, not for the address. Row-key stability across the pending→real swap is unaffected, since `marketTokenKey` lowercases when building its own key (`tokenListHelpers.ts:16`).

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,65 @@
...
+  const { isNative, normalizedAddress } = getNativeTokenInfo(
+    watchlistItem.isNative,
+    watchlistItem.contractAddress,
+  );
+  const chainId = watchlistItem.chainId;
+  return {
+    ...PENDING_WATCHLIST_METRICS,
+    id: getMarketWatchlistKey(watchlistItem),
+    name: '',
+    symbol: '',
```
sidmorizon: # 🟡 P2 Pending spot rows render a blank title and a duplicated testID, although the preview store already holds the symbol

## Problem

`buildPendingSpotWatchlistToken` hardcodes `name: ''` / `symbol: ''` (`watchlistPendingRows.ts:37-38`), while `buildTokenMarketRow` derives the visible title, the accessibility label and the test hook from the symbol:

```ts
testID: `market-token-item-${item.symbol}`,
accessibilityLabel: `${item.symbol}, ${price.text}, ${change.text}`,
...
title: item.symbol,
```
(`packages/kit/src/views/Market/MarketHomeV2/components/MarketNativeList/marketNativeListRows.ts:331-342`)

This contradicts the PR's stated goal of showing "identity rows while quotes load" — for spot there is no identity on the row at all.

The data is already captured: `rememberWatchlistListingPreview(watchlistItem, { logoUrl, name: tokenName, symbol: tokenSymbol })` runs for **every** star, including spot (`MarketStarV2.tsx:89-100`), where `getMarketWatchlistKey` yields `${chainId}:${addr}`. But `resolveListingWatchlistDisplay` is only called from the `assetId || stockId` branch (`useMarketWatchlistTokenList.ts:445-451`), so those spot entries are written and never read — confirmed by grepping every reference to `getWatchlistListingPreview` / `resolveListingWatchlistDisplay` across `packages/` and `apps/`.

## Impact

On the first paint the Favorites tab shows spot rows with an empty title, a fallback `CryptoCoinOutline` and `-`. Every pending spot row also collapses to the same `testID="market-token-item-"` and the same accessibility label, so e2e selectors and screen-reader output are ambiguous for the whole loading window.

## Recommendation

Read the preview store in `buildPendingSpotWatchlistToken` the same way the listing branch does:

```ts
const preview = getWatchlistListingPreview(watchlistItem);
...
name: preview?.name ?? '',
symbol: preview?.symbol ?? '',
```

and prefer `preview?.logoUrl` for `tokenImageUri` instead of the constant `''` in `PENDING_WATCHLIST_METRICS` (`watchlistPendingRows.ts:22`). This also makes the spot entries written at `MarketStarV2.tsx:96` actually useful.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/components/MarketStarV2.types.ts
```diff
@@ -9,6 +9,8 @@ export type IMarketStarV2Props = {
   contractAddress: string;
   from: EWatchlistFrom;
   tokenSymbol?: string;
+  tokenName?: string;
```
sidmorizon, line 12: # 🟡 P2 New `tokenName` / `logoUrl` props are not threaded through `MarketListingStar`, so the OK-63469 fix misses the Stocks and Top Coins tabs

## Problem

`tokenName` and `logoUrl` were added to `IMarketStarV2Props` here and wired into three call sites (`TokenDetailHeaderLeft.tsx`, `SwapProSearchTokenListItem.tsx`, `UniversalSearchV2MarketTokenItem.tsx`). But `MarketListingStar` — the wrapper used for every *listing* star — still has a closed prop list with neither prop, and forwards neither:

```tsx
export function MarketListingStar({ kind, listingId, from, tokenSymbol, testID, customIconSize = '$4' }: {
  kind: 'asset' | 'stock'; listingId: string; from: EWatchlistFrom;
  tokenSymbol?: string; testID?: string; customIconSize?: string;
}) {
  ...
  <MarketStarV2 ... tokenSymbol={tokenSymbol} ... />
```
(`packages/kit/src/views/Market/components/MarketListingStar.tsx:6-33`)

Its callers already hold the data. `MarketStockStar` receives the whole `IMarketStockPublicItem` and forwards only `stock.symbol`:

```tsx
<MarketListingStar kind="stock" listingId={stock.stockId} from={from} tokenSymbol={stock.symbol} testID={...} />
```
(`packages/kit/src/views/Market/MarketHomeV2/components/MarketStockList/MarketStockStar.tsx:14-21`)

while `IMarketStockPublicItem` declares `name: string` and `logoUrl: string` as required fields (`packages/shared/types/marketV2.ts:791-796`). `MarketTopCoinStar` and `MarketTokenSelectorRow` are in the same position.

## Impact

OK-63469's symptom — the watchlist row showing `CryptoCoinOutline` until the focus-gated batch quote arrives — is fixed only when starring from universal search, Swap Pro search, or token detail. Starring AAPL directly from the Market **Stocks** tab or Top Coins, arguably the most common path, still writes no preview, so `resolveListingWatchlistDisplay` has nothing to fall back on (`watchlistListingPreview.ts:63-70`) and the grey coin still renders.

## Recommendation

Add `tokenName?: string; logoUrl?: string;` to the `MarketListingStar` prop type (`MarketListingStar.tsx:13-20`) and forward both to `MarketStarV2` (`:23-33`); then pass `tokenName={stock.name} logoUrl={stock.logoUrl}` from `MarketStockStar.tsx:14-21`, and the equivalents from `MarketTopCoinStar` and `MarketTokenSelectorRow`.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/components/MarketStarV2.tsx
```diff
@@ -81,15 +86,19 @@ export const useStarV2Checked = ({
...
-          },
-        ]);
+        const watchlistItem = {
+          chainId: assetId || stockId ? '' : chainId,
+          contractAddress: assetId || stockId ? '' : contractAddress,
+          isNative,
+          assetId,
+          stockId,
+        };
+        rememberWatchlistListingPreview(watchlistItem, {
```
sidmorizon, line 105: # 🟡 P2 The preview map is written before the add is known to have succeeded, and has no eviction path in production

## Problem

`listingPreviews` is a module-level `Map` (`watchlistListingPreview.ts:18`). Three issues converge at this call site:

1. **Written before success.** `rememberWatchlistListingPreview(...)` runs at line 96, *before* `const added = await actions.addIntoWatchListV2([watchlistItem]);` at line 101. When `if (!added) { return; }` fires at lines 102-104 (limit reached, service failure), the entry is already stored and never rolled back.
2. **Never removed on unstar.** The `checked` branch (`MarketStarV2.tsx:73-87`) calls `removeFromWatchListV2` but does not delete the key, so unstarring leaves the entry permanently.
3. **No production eviction at all.** `clearWatchlistListingPreviews` is exported, but grepping every reference across `packages/` and `apps/` shows it is called only from the two new test files (`watchlistListingPreview.test.ts:11`, `useMarketWatchlistTokenList.listings.test.tsx:82`). Nothing in app code ever deletes an entry.

## Impact

Monotonic, unbounded growth of a module-level map for the lifetime of the `main` JS runtime — on iOS/Android that survives all navigation and only resets on app restart. Each entry is small, but it is unbounded by construction and retains display strings for favorites the user has already deleted and for adds that were rejected.

## Recommendation

- Move the `rememberWatchlistListingPreview` call from line 96 to after the `if (!added) { return; }` guard at line 104, so only successful adds are recorded.
- Add a `forgetWatchlistListingPreview(item)` export in `watchlistListingPreview.ts` (a `listingPreviews.delete(getMarketWatchlistKey(item))`) and call it from the removal branch at `MarketStarV2.tsx:79-81`, after `removed` is confirmed.
- Optionally cap the map size in `rememberWatchlistListingPreview` (`watchlistListingPreview.ts:32-41`) as a backstop.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/layouts/marketBannerLayoutReady.ts
```diff
@@ -0,0 +1,23 @@
+export function shouldWaitForNativeMarketBannerBeforeLayout(): boolean {
+  // Native watchlist quotes cannot start until Market layout mounts, so the
+  // banner request must not sit on the first-paint path. Banner header height
+  // is applied after fetch via resolveMarketBannerHeaderDecision.
+  return false;
```
sidmorizon: # 🟡 P2 `shouldWaitForNativeMarketBannerBeforeLayout()` is a constant `false`, leaving two statically dead disjuncts and a tautological test

## Problem

The helper takes no arguments and unconditionally returns `false`. It is then used as a disjunct in two first-paint conditions:

```tsx
shouldWaitForSpotCategoryReady ||
shouldWaitForNativeMarketBannerBeforeLayout()
```
(`packages/kit/src/views/Market/MarketHomeV2/MarketHomeV2.tsx:322-325` and `:402-405`)

Both right-hand sides are statically dead, so the code at the Market entry point reads as though a banner gate still exists when it cannot ever fire. The accompanying test asserts nothing about behaviour:

```ts
it('does not block native first paint so watchlist quotes can start with the banner', () => {
  expect(shouldWaitForNativeMarketBannerBeforeLayout()).toBe(false);
});
```
(`marketBannerLayoutReady.test.ts:6-8`)

It passes for any implementation that returns `false` — it restates the literal and cannot catch a regression, since the real gating now lives entirely in `shouldFetchWatchlistQuotesWhileFocused` (`marketBannerLayoutReady.ts:8-23`).

## Impact

Misleading control flow at the Market first-paint entry point: a future reader (or a future revert of the banner gate) has to prove the disjunct is dead before touching either condition. The test adds coverage without assertion value, which makes the file look better tested than it is.

## Recommendation

Delete `shouldWaitForNativeMarketBannerBeforeLayout` (`marketBannerLayoutReady.ts:1-6`) along with both `|| shouldWaitForNativeMarketBannerBeforeLayout()` disjuncts and the now-unused import (`MarketHomeV2.tsx:38`, `:322-325`, `:402-405`), keeping the explanatory comment from `marketBannerLayoutReady.ts:2-4` at one of those sites. Drop the tautological case from `marketBannerLayoutReady.test.ts:6-8` and keep the `shouldFetchWatchlistQuotesWhileFocused` cases, which do exercise real branches.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
```diff
@@ -0,0 +1,71 @@
+import { getMarketWatchlistKey } from '@onekeyhq/shared/src/utils/marketWatchlistIdentity';
```
sidmorizon, line 1: # 🟠 P1 New modules are not registered in `module-id-registry.json`, so the native union build fails

## Problem

This PR adds three new production modules but does not update `apps/mobile/bundle-registry/module-id-registry.json`. `git diff --name-only 591ce92..d10a5ce -- apps/mobile/bundle-registry/` returns nothing, and none of the new files appear in the registry:

```
watchlistListingPreview.ts    → 0 matches
watchlistPendingRows.ts       → 0 matches
marketBannerLayoutReady.ts    → 0 matches
```

`buildSegmentAllocation` in `apps/mobile/scripts/unionBuild.js:571` resolves every graph module through `moduleIdRegistry.get`, which throws on an unregistered path (`apps/mobile/plugins/moduleIdRegistry.js:391`). The **Native startup graph budget (24.x)** check fails on the first one it reaches:

```
packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts.
```
([job 104683041377](https://github.com/OneKeyHQ/app-monorepo/actions/runs/35061631547/job/104683041377), after both graphs build successfully)

The other 22 checks pass, so this is the only thing keeping the head red, and it is entirely caused by this PR's new files.

## Impact

The iOS/Android union bundle cannot be produced at this head — the build aborts after the main and background graphs are built, before segment allocation, and no allocation or budget report is emitted. Since the whole point of this PR is native first-paint behaviour, the native bundle is exactly the artifact that cannot currently be validated, and the PR is unmergeable while the check is red.

## Recommendation

Run the command the failure names and commit the result:

```
```

(script defined at `apps/mobile/package.json:34`). That should add entries for all three new modules — `watchlistListingPreview.ts`, `watchlistPendingRows.ts` and `MarketHomeV2/layouts/marketBannerLayoutReady.ts`. Do not hand-edit the registry; regenerate it. Note that if the P2 suggestion to delete `marketBannerLayoutReady.ts` is taken, regenerate afterwards so the registry matches the final file set.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -487,8 +501,13 @@ export function useMarketWatchlistTokenList({
...
         });
-        // Keep legacy chain favorites removable under their stored identity.
-        return found ? { ...found, stockId: watchlistItem.stockId } : undefined;
+        if (found) {
+          return { ...found, stockId: watchlistItem.stockId };
+        }
+        const spotQuotesPending = !apiResult && !spotResult?.failed;
+        return spotQuotesPending
+          ? buildPendingSpotWatchlistToken(watchlistItem, networkLogoUriMap)
+          : undefined;
```
sidmorizon: 🟡 P2: Desktop and web show blank favorites

Severity: `non-severe`

During initial loading, `transformedData` emits pending rows on every platform. Desktop and web suppress their skeletons and render nameless favorites.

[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -107,8 +113,17 @@ export function useMarketWatchlistTokenList({
...
+  isInitialLoadRef.current = isInitialLoad;
 
   const pageIndex = useCarouselIndex();
+  const canFetchWatchlistQuotes = (isFocused: boolean) =>
+    shouldFetchWatchlistQuotesWhileFocused({
+      isFocused,
+      pageIndex,
+      isNative: Boolean(platformEnv.isNative),
+      isInitialLoad: isInitialLoadRef.current,
+    });
```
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed

packages/kit/src/views/Market/components/MarketStarV2.types.ts
zy2000-star: fixed

packages/kit/src/views/Market/components/MarketStarV2.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/layouts/marketBannerLayoutReady.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
sidmorizon: # 🟠 P1 Registry fix is correct for the original three modules, but has regressed on the current head

## Problem

Not resolving this thread: the fix is right, but `3890cd2b` reintroduced the same failure with two new files.

Verified at `3890cd2b` — the original three modules are now registered:

```
watchlistListingPreview.ts    → registered
watchlistPendingRows.ts       → registered
marketBannerLayoutReady.ts    → registered
```

But `3890cd2b` ("split stocks out of universal search market results") adds two further production modules and does **not** touch the registry — `git diff --name-only d61ce620..3890cd2b -- apps/mobile/bundle-registry/` is empty:

```
packages/kit/src/views/UniversalSearch/components/marketSearchMetric.ts   → 0 matches
packages/kit/src/views/UniversalSearch/universalSearchTabs.ts             → 0 matches
```

Both are reachable from the production graph, so they are not test-only:

- `universalSearchTabs.ts` ← `pages/UniversalSearch.tsx:72`
- `marketSearchMetric.ts` ← `components/SearchResultItems/UniversalSearchV2MarketTokenItem.tsx:51` (value import `getMarketSearchMetricAmount`) and `components/MarketTableHeader.tsx:6` (type-only)

## Impact

`buildSegmentAllocation` (`apps/mobile/scripts/unionBuild.js:571`) will throw on `moduleIdRegistry.get` for the first unregistered path exactly as before, so **Native startup graph budget (24.x)** will fail again on this head for the same reason. The previous run's green-ness does not carry over, because the registry is keyed by module path and two new paths entered the graph.

## Recommendation

Re-run the same command after the latest commit and commit the result:

```
```

Worth making this a habit for the rest of this PR: any commit that adds a non-test module under `packages/kit/src` needs the registry regenerated in the same commit, otherwise this check goes red again on each push.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,69 @@
...
+  const { isNative } = getNativeTokenInfo(
+    watchlistItem.isNative,
+    watchlistItem.contractAddress,
+  );
+  const preview = getWatchlistListingPreview(watchlistItem);
+  const chainId = watchlistItem.chainId;
+  return {
+    ...PENDING_WATCHLIST_METRICS,
+    id: getMarketWatchlistKey(watchlistItem),
+    name: preview?.name ?? '',
```
sidmorizon, line 65: # 🟠 P1 Giving pending rows a real `name` unlocks a token-detail preview carrying placeholder `decimals: 0`, which enables the swap panel

## Problem

This line (with `address` at L41 and `symbol` at L40) is what makes a pending row pass the preview gate in `useToMarketDetailPage.buildPreviewTokenDetail`:

```ts
const previewAddress = item.address ?? item.tokenAddress;
if (
  (!previewAddress && !item.isNative) ||
  !item.name ||
  typeof item.decimals !== 'number'
) {
  return undefined;
}
```
(`useToMarketDetailPage.ts:107-115`)

At the previous head `name` was `''`, so a pending row could never build a preview. Now, whenever a preview entry exists for the token, all three conditions pass: `previewAddress` is non-empty (the correct fix from the other thread), `item.name` is truthy, and `decimals` is the hardcoded placeholder `0` from `PENDING_WATCHLIST_METRICS` (`watchlistPendingRows.ts:14`) — `typeof 0 === 'number'`.

The preview is then committed on native and consumed by `useMarketDetailDisplayData`, whose `buildPreviewTokenDetail` does **not** set `decimalsResolved`:

```ts
    decimals: tokenDetailPreview.decimals,
    price: toDisplayNumber(tokenDetailPreview.price),
```
(`useMarketDetailDisplayData.ts:33-34`)

Contrast the sibling stock builder in the same file, which sets it explicitly for precisely this reason:

```ts
    decimals: 0,
    decimalsResolved: false,
```
(`useMarketDetailDisplayData.ts:63-64`)

So the trade-readiness guard, which is written to reject unresolved precision, passes on `undefined`:

```ts
const isSwapTokenReady =
  tokenDetail?.address?.toLowerCase() === tokenAddress.toLowerCase() &&
  tokenDetail?.networkId === networkId &&
  tokenDetail?.decimalsResolved !== false &&      // undefined → passes
  typeof tokenDetail?.decimals === 'number' &&    // 0 → passes
  Number.isInteger(tokenDetail.decimals) &&       // 0 → passes
  tokenDetail.decimals >= 0;                      // 0 → passes
```
(`MarketDetailV2/layouts/MobileLayout.tsx:990-996`, and the same shape at `DesktopLayout.tsx:308-313`)

`MobileLayout.tsx:1083` then renders `<LazySwapPanel swapToken={toSwapPanelToken} …/>`.

## Impact

Repro: star a token (so a preview entry exists), open Favorites, and tap the row while the batch quote is still in flight — the exact window this PR makes interactive. The detail page opens with `decimals: 0`, `liquidity: 0`, `holders: 0`, and the trade widget mounts on that placeholder precision instead of waiting for the real token detail. The codebase already treats unresolved decimals as a correctness hazard for stocks; this path routes around that guard rather than satisfying it.

## Recommendation

Make the placeholder explicit so the existing guards fire. Either mark the row and short-circuit in `useToMarketDetailPage.buildPreviewTokenDetail` (`:102-126`) — e.g. add an `isPendingWatchlistRow?: boolean` to `IMarketToken`, set it in `buildPendingSpotWatchlistToken`, and `return undefined` on it — or carry `decimalsResolved: false` through the token preview the way `buildStockPreviewTokenDetail` already does at `useMarketDetailDisplayData.ts:64`. Dropping `decimals` from `PENDING_WATCHLIST_METRICS` would also close it, since `typeof undefined !== 'number'` fails the gate at `useToMarketDetailPage.ts:113`.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -500,20 +521,27 @@ export function useMarketWatchlistTokenList({
     spotItems,
     perpsTokenMap,
     networkLogoUriMap,
+    perpsApiResult,
+    perpsResult,
+    spotResult,
   ]);
 
   useEffect(() => {
+    const watchlistHydrated = isWatchlistMounted !== false;
```
sidmorizon: # 🟡 P2 The new hydration gate defaults to "mounted", so one native call site silently keeps the old behavior

## Problem

`isWatchlistMounted` is optional (`isWatchlistMounted?: boolean`, L68) and this line treats `undefined` as hydrated:

```ts
const watchlistHydrated = isWatchlistMounted !== false;
```

That makes the gate opt-in. Any caller that omits the prop keeps exactly the pre-fix behavior — `isInitialLoad` clears as soon as the empty-watchlist pipeline resolves, ending the native first-load bypass before the real favorites arrive.

The commit updated three of the four production call sites (`MobileMarketNativeLists.tsx:760`, `MarketWatchlistTokenList.tsx:128`, `MobileMarketWatchlistFlatList.tsx:101`) but not this one:

```tsx
const [{ data: watchListData }] = useMarketWatchListV2Atom();

const { data, isLoading } = useMarketWatchlistTokenList({
  watchlist: watchListData ?? [],
  pollingInterval: pollingInterval ?? TOKEN_SELECTOR_POLLING_INTERVAL,
});
```
(`MarketDetailV2/components/TokenSelector/MarketTokenSelectorList.tsx:160-165`)

It destructures only `data` from the atom, never `isMounted`. The component is native-reachable through `MarketTokenSelector` from `TokenDetailHeaderLeft.tsx` and `MarketChartFullscreenHeader.tsx`.

## Impact

In the token-selector Favorites tab on native, `isInitialLoad` can still clear against the empty pre-hydration atom, so the quote bypass ends early and the tab can also render its empty state before the watchlist has hydrated — the same defect class this commit was written to close, left behind at one site. Lower blast radius than the Home Favorites tab (the popover usually opens after hydration), but it is the same bug.

## Recommendation

Destructure `isMounted` at `MarketTokenSelectorList.tsx:160` and pass `isWatchlistMounted: isMounted` at `:162`. Better still, make the parameter required in the hook signature (`useMarketWatchlistTokenList.ts:68`) so the compiler catches the next call site instead of it defaulting to the old path — the three updated callers already supply it, so only this one needs touching.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.listings.test.tsx
```diff
@@ -182,6 +193,121 @@ it('retains an unavailable asset so it can still be removed', async () => {
...
+        pollingInterval: 0,
+      }),
+    {
+      initialProps: {
+        nextWatchlist: [] as typeof watchlist,
+        isWatchlistMounted: false,
+      },
+    },
+  );
+  await waitFor(() => expect(mockBatch).not.toHaveBeenCalled());
```
sidmorizon: # 🟡 P2 The new bypass test asserts two things that pass regardless of the fix

## Problem

In `it('keeps the native first-load bypass until the watchlist atom is mounted')` (L240-273), only one of the three assertions can fail if the fix is reverted.

This line is vacuous twice over:

```ts
await waitFor(() => expect(mockBatch).not.toHaveBeenCalled());
```

First, `waitFor` resolves on the first synchronous pass of a `not.toHaveBeenCalled()` callback, so it asserts nothing about the future. Second, the case is set up with an empty watchlist, and the spot method short-circuits before ever reaching the batch call:

```ts
if (!watchlist || watchlist.length === 0) { … return { list: [], failed: false } as const; }
```
(`useMarketWatchlistTokenList.ts:219-224`)

so `mockBatch` is uncalled whatever `isWatchlistMounted` is.

L272 (`await waitFor(() => expect(mockBatch).toHaveBeenCalled())` after the rerender) is also fix-independent: this file replaces `usePromiseResult` with a stub (L46-71) that ignores `overrideIsFocused`, `checkIsFocused` and polling, and simply re-runs `method()` on every deps change. So the batch fires on the rerender either way.

That mock is worth noting on its own: the mechanism the test names — `isInitialLoad` → `isInitialLoadRef` → `overrideIsFocused` → `shouldFetchWatchlistQuotesWhileFocused` — lives inside the real `usePromiseResult`, which is mocked out here. Only L270 (`expect(result.current.isLoading).toBe(true)`) is load-bearing: without the gate, `isInitialLoad` clears and `isLoading` computes `false` at L343-350.

For contrast, the sibling test at L225-239 is sound — remove the `platformEnv.isNative &&` gate and its first render immediately contains a pending row.

## Impact

The test reads as covering the hydration bypass but would stay green through a regression in the focus-gating half of it, which is the part most likely to break — it is a four-hop interaction across two files.

## Recommendation

Drop L269 and L271-272 and assert the actual state transition instead: keep `isWatchlistMounted: false` with a **non-empty** watchlist, resolve the batch, assert `isLoading` is still `true`; then rerender with `isWatchlistMounted: true` and assert it flips to `false`. Cover the focus half separately with a direct unit test on `shouldFetchWatchlistQuotesWhileFocused`, plus an assertion that the `overrideIsFocused` passed to `usePromiseResult` returns `true` when `isInitialLoad` is `true` on native — that one can be checked against the mock's captured options without the real hook.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -487,8 +503,13 @@ export function useMarketWatchlistTokenList({
...
         });
-        // Keep legacy chain favorites removable under their stored identity.
-        return found ? { ...found, stockId: watchlistItem.stockId } : undefined;
+        if (found) {
+          return { ...found, stockId: watchlistItem.stockId };
+        }
+        const spotQuotesPending = !apiResult && !spotResult?.failed;
+        return platformEnv.isNative && spotQuotesPending
+          ? buildPendingSpotWatchlistToken(watchlistItem, networkLogoUriMap)
+          : undefined;
```
sidmorizon: 🟡 P2: Cached quotes hide new favorites

Severity: `non-severe`

With cached quotes, `spotQuotesPending` and `perpsQuotesPending` reject new native favorites missing from those caches. The favorites stay absent until a successful refresh, and remain unremovable there if refresh fails.

originalix:

Review of commit `edfcaabe3fd1` finished, but some comments could not be published automatically.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.listings.test.tsx
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
```diff
@@ -0,0 +1,75 @@
...
+    IMarketListingWatchlistQuote,
+    'name' | 'symbol' | 'logoUrl' | 'variants'
+  >;
+}) {
+  const preview = getWatchlistListingPreview(watchlistItem);
+  const fallbackId = watchlistItem.assetId ?? watchlistItem.stockId ?? '';
+  return {
+    name: quote?.name || preview?.name || fallbackId,
+    symbol: quote?.symbol || preview?.symbol || fallbackId,
+    tokenImageUri: quote?.logoUrl || preview?.logoUrl || '',
```
sidmorizon: 🟡 P2: Removed listing logos stay visible

Severity: `non-severe`

Successful quotes with an empty `logoUrl` keep the remembered preview logo. `resolveListingWatchlistDisplay` uses truthiness, so server-side logo removals never reach mounted Favorites rows.

Suggested fix:

```
    tokenImageUri: quote ? quote.logoUrl || '' : preview?.logoUrl || '',
```

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,92 @@
...
+  quotesFailed,
+  hasCachedRows,
+}: {
+  isNative: boolean;
+  hasQuotePayload: boolean;
+  hasSuccessfulQuotes: boolean;
+  quotesFailed: boolean;
+  hasCachedRows: boolean;
+}): boolean {
+  if (!isNative || hasSuccessfulQuotes) {
```
sidmorizon: # 🟡 P2 The `hasSuccessfulQuotes` short-circuit means a newly starred favorite still gets no pending row while the list stays mounted

## Problem

This line returns early before the new cache-miss rule is ever consulted, and the caller derives the flag from a result that survives re-runs:

```ts
hasSuccessfulQuotes: Boolean(spotResult && !spotResult.failed),
```
(`useMarketWatchlistTokenList.ts:520`; perps equivalent at `:490`)

`usePromiseResult` only clears `result` on a deps change when `undefinedResultIfReRun` is set (`usePromiseResult.ts:279-281`), and this hook never passes it — its options are `pollingInterval`, `watchLoading`, `revalidateOnFocus`, `revalidateOnReconnect`, `overrideIsFocused`, `checkIsFocused` (`useMarketWatchlistTokenList.ts:250-257`).

So when the user stars token C on an already-mounted native Favorites list: `spotItems` changes → the batch re-runs → but `spotResult` still holds the previous successful `{ list: [A, B], failed: false }`. `hasSuccessfulQuotes` is `true`, line 40 returns `false`, and C gets no row.

The `hasCachedRows` rule at line 46 is therefore only reachable when `spotResult` is `undefined` or `failed` — i.e. on a fresh mount seeded from `dataCacheRef` (`useMarketWatchlistTokenList.ts:294-298`). That is exactly and only the path the new test covers (`useMarketWatchlistTokenList.listings.test.tsx:298`, "emits a native pending row for a new favorite missing from cached quotes"), which is a genuine regression test for the mount case but leaves the mounted-list case untested.

## Impact

On iOS/Android, starring a token from an open Favorites list still leaves it absent for the whole duration of the batch request, rather than showing an identity row — the case described as "Starring a token already missing from the in-memory quote cache left it off native Favorites". Behaviour there is unchanged from `140bd72a`. Because the fix reads as shipped and has a passing regression test, the remaining gap is easy to mistake for covered.

## Recommendation

The meaningful signal is "is a request covering this identity in flight", not "did some earlier request succeed". Replace the `hasSuccessfulQuotes` short-circuit at lines 40-42 with an explicit `hasQuotesInFlight` input, and pass `hasQuotesInFlight: apiLoading !== false` at `useMarketWatchlistTokenList.ts:517-523` (and `perpsLoading !== false` at `:487-495`), emitting a pending row whenever the identity is absent from the current payload and a request is in flight. Add a case to `watchlistPendingRows.test.ts:87` covering a successful prior payload plus a newly added identity.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,92 @@
...
+    return false;
+  }
+  if (!hasQuotePayload) {
+    return !quotesFailed;
+  }
+  return hasCachedRows;
```
sidmorizon: # 🟡 P2 `quotesFailed` is ignored on this branch, so a failed native refresh now materialises blank ghost rows

## Problem

`quotesFailed` is computed and passed by both callers (`useMarketWatchlistTokenList.ts:521` and `:491`) and honoured on the `!hasQuotePayload` branch at line 44, but this branch drops it entirely.

On native, `spotResult.failed` makes `apiResult` fall back to `lastSpotResultRef.current` (`useMarketWatchlistTokenList.ts:333-336`), which is populated from any past success (`:304-313`). That yields `hasQuotePayload: true`, `hasSuccessfulQuotes: false`, `quotesFailed: true`, `hasCachedRows: true` → this line returns `true`.

At `140bd72a` the predicate was `!apiResult && !spotResult?.failed`, which was `false` in exactly this state — so this is a behaviour change introduced by the commit, not pre-existing.

Concretely: a starred token the backend permanently omits (delisted, unsupported) renders nothing today. After the first failed poll it materialises as a placeholder with `name: ''`, `symbol: ''`, `price: NaN`, `priceChangeRaw: '-'` — and since `marketNativeListRows.ts:330,342` derive both from the symbol, it renders a blank title and `testID="market-token-item-"`. It persists for as long as refreshes keep failing, alongside the retry banner (`useMarketWatchlistTokenList.ts:352-355`), and disappears again on recovery.

The new truth table never combines `hasQuotePayload: true` with `quotesFailed: true` (`watchlistPendingRows.test.ts:87-134`), so the branch is untested.

## Impact

Blank, identity-less ghost rows appear on iOS/Android Favorites precisely when the network is failing — the moment the list is least able to explain itself. Multiple such rows also collapse to the same empty `testID`, which breaks row-targeted automation for the duration.

## Recommendation

Honour the flag that is already being passed:

```ts
return hasCachedRows && !quotesFailed;
```

and add the missing `hasQuotePayload: true` / `quotesFailed: true` case to the truth table at `watchlistPendingRows.test.ts:87`. Note this is separate from the `hasSuccessfulQuotes` ordering issue on the other thread — fixing that one does not close this, since a failed refresh leaves `hasSuccessfulQuotes` false either way.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -487,33 +511,52 @@ export function useMarketWatchlistTokenList({
...
           );
         });
-        // Keep legacy chain favorites removable under their stored identity.
-        return found ? { ...found, stockId: watchlistItem.stockId } : undefined;
+        if (found) {
+          return { ...found, stockId: watchlistItem.stockId };
+        }
+        return shouldEmitNativePendingWatchlistRow({
+          isNative: Boolean(platformEnv.isNative),
+          hasQuotesInFlight: apiLoading !== false,
```
sidmorizon, line 561: # 🟡 P2 `apiLoading !== false` is also true on every polling refresh, so a favorite the backend omits now flickers a blank row each cycle

## Problem

Thanks for taking the `hasQuotesInFlight` direction — it does close both previous threads. But `apiLoading` is not only true for the first load; it toggles on every poll.

`usePromiseResult`'s polling path re-enters the same runner:

```ts
await timerUtils.wait(pollingInterval);
await defer.promise;
...
void run({ triggerByDeps: true, pollingNonce: config.pollingNonce });
```
(`usePromiseResult.ts:374-388`)

and that runner calls `setLoadingTrue()` at `:282` (`watchLoading` is on for this hook), then `setLoadingFalse()` in the `finally` at `:369`. So `apiLoading` cycles `false → true → false` on every interval.

For a watchlist identity the backend consistently omits — delisted, unsupported, or simply never returned — `found` is always `undefined`, and the inputs alternate:

- **Between polls:** `hasQuotesInFlight: false` → line 40 returns `false` → no row. Correct.
- **During each poll:** `hasQuotesInFlight: true`, `hasQuotePayload: true` (`apiResult` still holds the previous success — `undefinedResultIfReRun` is not set), `quotesFailed: false`, `hasCachedRows: true` → line 46 returns `true` → a pending row is emitted.

So the row appears for the duration of each request and disappears when it settles. When the preview `Map` has no entry for that identity — the normal case after an app restart — it renders with `name: ''` / `symbol: ''`, i.e. a blank title and `testID="market-token-item-"` (`marketNativeListRows.ts:330,342`).

At `6be6c870` this could not happen: `hasSuccessfulQuotes` was `true` after the first success and short-circuited at line 40. The new predicate removed that short-circuit without distinguishing "not yet queried" from "queried and genuinely absent".

## Impact

On iOS/Android, any favorite missing from the batch response flickers an empty row in and out of the Favorites list once per polling interval, for as long as the list is open. Several such favorites produce several rows sharing the same empty `testID`, so row-targeted automation is ambiguous during each window.

## Recommendation

The signal that actually distinguishes the two cases is whether the identity was covered by the last *completed* payload, not whether a request is in flight. Concretely, at `useMarketWatchlistTokenList.ts:517-523`, pass something like `isIdentityUnqueried: !lastRequestedKeys.has(getMarketWatchlistKey(watchlistItem))`, where `lastRequestedKeys` is captured from the `tokenAddressList` of the most recently *settled* batch (built at `:229-240`), and gate the `hasCachedRows` branch (`watchlistPendingRows.ts:46`) on it. An identity present in the last settled request but absent from its results is known-absent and should stay hidden; one added since is genuinely pending.

Please also add a truth-table case at `watchlistPendingRows.test.ts:87` for `hasQuotesInFlight: true, hasQuotePayload: true, hasCachedRows: true` with an identity that the previous payload already covered — that is the case that currently regresses and none of the existing rows pin it.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,92 @@
...
+  quotesFailed: boolean;
+  hasCachedRows: boolean;
+}): boolean {
+  if (!isNative || quotesFailed || !hasQuotesInFlight) {
+    return false;
+  }
+  if (!hasQuotePayload) {
+    return true;
+  }
+  return hasCachedRows;
```
sidmorizon: 🟡 P2: First favorite stays hidden

Severity: `non-severe`

`shouldEmitNativePendingWatchlistRow` suppresses the first favorite when the retained quote payload is an empty successful list. Native Favorites stays skeletal until the covering request finishes.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
```diff
@@ -0,0 +1,75 @@
...
+    IMarketListingWatchlistQuote,
+    'name' | 'symbol' | 'logoUrl' | 'variants'
+  >;
+}) {
+  const preview = getWatchlistListingPreview(watchlistItem);
+  const fallbackId = watchlistItem.assetId ?? watchlistItem.stockId ?? '';
+  return {
+    name: quote?.name || preview?.name || fallbackId,
+    symbol: quote?.symbol || preview?.symbol || fallbackId,
+    tokenImageUri: quote ? quote.logoUrl || '' : preview?.logoUrl || '',
```
sidmorizon, line 105: 🟡 P2: Failed polls restore stale logos

Severity: `non-severe`

`resolveListingWatchlistDisplay` restores the remembered logo when a later stock poll fails after returning a newer logo. `listingPreviews` still holds the original search identity, so Favorites visibly regresses.

originalix:

Review of commit `4778e7cbdf32` finished, but some comments could not be published automatically.

originalix (dismissed):

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
zy2000-star: fixed — pending rows now require `isIdentityUnqueried` based on the last settled request keys, so known-absent favorites no longer flicker on poll.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
zy2000-star: fixed — empty settled payloads no longer block the first favorite; `isIdentityUnqueried` replaces `hasCachedRows`.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistListingPreview.ts
zy2000-star: fixed — successful listing quotes sync into the preview map so a later failed poll keeps the last good logo.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -223,19 +266,20 @@ export function useMarketWatchlistTokenList({
         const response = await fetchMarketTokenListBatchForPlatform({
           tokenAddressList,
         });
+        lastSettledSpotRequestKeysRef.current = requestedKeys;
```
sidmorizon: # 🟡 P2 The settled-keys ref is committed by runs whose result is discarded, so overlapping requests can resurrect the flicker

## Problem

The `isIdentityUnqueried` approach is the right signal and it does close both previous threads — I verified the three cases (newly starred identity on a mounted list, known-absent identity across polls, empty successful payload on first load) all now behave correctly.

One gap remains in *when* the ref is committed. This line runs unconditionally as soon as the fetch resolves, inside the `usePromiseResult` method body. But `usePromiseResult` does not necessarily apply that run's result:

```ts
if (
  shouldApplyResult(config) &&
  nonceRef.current === nonce &&
  capturedSwrKey === swrKeyRef.current
) {
  setResult(r);
```
(`usePromiseResult.ts:293-298`)

A superseded run — deps changed, a poll overlapping a focus revalidation — still runs to completion and still executes this assignment, even though its result is thrown away. Because the runner is re-entered on deps changes, polling and refocus, overlapping runs are reachable rather than theoretical.

Concretely, with R1 covering `{A, B}` and R2 covering `{A, B, C}` after the user stars C: if R1 resolves *after* R2, `lastSettledSpotRequestKeysRef` ends up `{A, B}` while the applied result is R2's. `C` is then treated as unqueried on subsequent polls, and if the backend omits it, the blank pending row flickers once per interval again — exactly the symptom this commit removes.

The perps path has the same shape at `:308`.

## Impact

Narrow but real: after overlapping requests, a single identity can regress to the pre-fix flicker, with the blank title and shared `testID="market-token-item-"` that comes with it. It is also the kind of state that is awkward to reproduce deliberately, so it would likely surface as an intermittent report rather than a clean repro.

## Recommendation

Let the key set travel with the result instead of side-channelling it through a ref, so it becomes visible only when that result is actually applied:

```ts
const response = await fetchMarketTokenListBatchForPlatform({ tokenAddressList });
return { ...response, failed: false, requestedKeys } as const;
```

then read `apiResult?.requestedKeys` at `:552-556` in place of `lastSettledSpotRequestKeysRef.current` (and the same for perps at `:308` / `:522-527`). That also removes the ref-mutation-inside-render-input pattern, which a `useMemo` cannot observe reliably anyway. The empty-watchlist resets at `:244` and `:249` become plain `requestedKeys: new Set()` in the returned value.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.listings.test.tsx
```diff
@@ -175,13 +217,284 @@ it('retains an unavailable asset so it can still be removed', async () => {
...
+        isNative: boolean;
+      }>;
+    }) =>
+      useMarketWatchlistTokenList({
+        watchlist,
+        isWatchlistMounted: true,
+        pollingInterval: 0,
+      }),
+    {
+      initialProps: { watchlist: [] },
```
sidmorizon: # 🟠 P1 `initialProps: { watchlist: [] }` infers `never[]`, failing the TypeScript check and turning `lint (24.x)` red

## Problem

`lint (24.x)` is failing on this head, and the tsc step names this file:

```
packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.listings.test.tsx: line 444 col 7,
Type '{ chainId: string; contractAddress: string; isNative: false; }' is not assignable to type 'never'.
[3:07:03 PM] TypeScript check failed. (153.36s)
```
([job 104849241408](https://github.com/OneKeyHQ/app-monorepo/actions/runs/35112409940/job/104849241408))

`renderHook<Result, Props>` infers `Props` from the `initialProps` object literal as well as from the callback. The empty array here is inferred as `never[]`, which wins over the explicit annotation on the callback parameter (lines 424-429), so `Props` becomes `{ watchlist: never[] }`. The `rerender` at lines 442-450 then tries to pass a real element into `never`, which is the reported error at line 444.

The sibling test in this file does not hit it only because its `initialProps` array is non-empty and therefore infers a usable element type:

```ts
initialProps: { watchlist: [cachedItem] },
```
(`:393`)

This is also the `never[]` fallback that `CLAUDE.md` calls out under "Keep TypeScript precise".

## Impact

The required `lint (24.x)` check is red on the current head, so the PR is not mergeable as it stands — and the failure is in the test added to cover the empty-settle case, so that new coverage is not actually running. It is unrelated to the shard-2 `Unit Tests (2/5)` failure, which came in from the base branch on the previous head.

## Recommendation

Give the empty array an element type so `Props` is inferred correctly. Extracting the shape is cleanest, since it is already written out twice in this test:

```ts
type IWatchlistProps = {
  watchlist: Array<{ chainId: string; contractAddress: string; isNative: boolean }>;
};
...
const { rerender, result } = renderHook(
  ({ watchlist }: IWatchlistProps) => useMarketWatchlistTokenList({ … }),
  { initialProps: { watchlist: [] } as IWatchlistProps },
);
```

A local `const emptyWatchlist: IWatchlistProps['watchlist'] = []` passed as `initialProps: { watchlist: emptyWatchlist }` works equally well. Worth running `yarn agent:check --profile commit` before the next push — it runs the same tsc pass that CI does here.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -149,6 +170,9 @@ export function useMarketWatchlistTokenList({
                   await backgroundApiProxy.serviceMarketV2.fetchMarketListingWatchlistQuote(
                     item,
                   );
+                if (quote) {
+                  syncWatchlistListingPreviewFromQuote(item, quote);
+                }
```
sidmorizon: 🟡 P2: Stale requests regress listing previews

Severity: `non-severe`

When overlapping listing fetches resolve out of order, stale `syncWatchlistListingPreviewFromQuote` calls overwrite the latest preview. A later failed poll restores an older logo or name.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
sidmorizon: 🟡 P2: Stale requests corrupt quote coverage

Severity: `non-severe`

Out-of-order requests let stale responses overwrite `lastSettledSpotRequestKeysRef` or `lastSettledPerpsRequestKeysRef`. Later polls can revive known-absent favorites as blank pending rows.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -487,33 +544,54 @@ export function useMarketWatchlistTokenList({
...
+        return shouldEmitNativePendingWatchlistRow({
+          isNative: Boolean(platformEnv.isNative),
+          hasQuotesInFlight: apiLoading !== false,
+          hasQuotePayload: Boolean(apiResult),
+          quotesFailed: Boolean(spotResult?.failed),
+          isIdentityUnqueried: !lastSettledSpotRequestKeysRef.current.has(
+            getMarketWatchlistKey(watchlistItem),
+          ),
+        })
+          ? buildPendingSpotWatchlistToken(watchlistItem, networkLogoUriMap)
```
sidmorizon, line 567: 🟡 P2: Recovery hides newly starred favorites

Severity: `non-severe`

Adding a favorite after a native quote failure leaves `quotesFailed` true during its covering retry. The new favorite stays absent until that retry settles.

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/watchlistPendingRows.ts
```diff
@@ -0,0 +1,95 @@
...
+export function shouldEmitNativePendingWatchlistRow({
+  isNative,
+  hasQuotesInFlight,
+  hasQuotePayload,
+  isIdentityUnqueried,
+}: {
+  isNative: boolean;
+  hasQuotesInFlight: boolean;
+  hasQuotePayload: boolean;
+  isIdentityUnqueried: boolean;
```
sidmorizon, line 36: # 🟠 P1 "Unqueried" is not in the spell-check word list, so oxlint fails the build — 13 warnings, 0 errors

## Problem

Both previous issues on this head are genuinely fixed: the TypeScript check now **completes** (`[4:23:03 AM] TypeScript check completed. (153.52s)`), and `requestedKeys` correctly travels with the result. `lint (24.x)` is still red, but for a different reason — the oxlint stage:

```
##[warning]Unknown word: "Unqueried"     (×13)

Found 13 warnings and 0 errors.
Finished in 229.7s on 9465 files with 259 rules using 4 threads.
[4:26:56 AM] Oxlint check failed. (231.47s)
```
([job 105073895114](https://github.com/OneKeyHQ/app-monorepo/actions/runs/35181328796/job/105073895114))

The spell checker rejects the coined word in `isIdentityUnqueried`, and CI treats warnings as failure — note `0 errors` yet a non-zero exit. There are 12 occurrences across three files (`watchlistPendingRows.ts`, `watchlistPendingRows.test.ts`, `useMarketWatchlistTokenList.ts`) plus the comment at `:40`, which matches the 13 warnings exactly.

`development/spellCheckerSkipWords.txt` exists for precisely this and already carries coined terms (`pumpfun`, `poolstrade`, `meteoradbc`, …), but has no entry for this one.

This one is on me — I suggested the `isIdentityUnqueried` name without checking it against the word list.

## Impact

`lint (24.x)` is the only red check on this head, so the PR stays unmergeable on a one-word omission, after two rounds in which the real defects were already fixed.

## Recommendation

Add the word to `development/spellCheckerSkipWords.txt`, matching the file's existing lowercase convention:

```
unqueried
```

If you would rather not extend the dictionary, renaming to something the checker already accepts works too — `isIdentityNotYetRequested` or `isIdentityOutsideLastRequest` — but that touches all 12 call sites, so the word-list entry is the smaller change. Either way `yarn agent:check --profile commit` would surface it locally, since it runs this same oxlint stage.

[Claude session ↗](https://claude.ai/code/session_0161tq4jApu9QtHDpk9ApStB)

---

packages/kit/src/views/Market/components/MarketStarV2.tsx
```diff
@@ -74,23 +82,34 @@ export const useStarV2Checked = ({
...
+          logoUrl,
+          name: tokenName,
+          symbol: tokenSymbol,
+        });
```
sidmorizon, line 109: 🟡 P2: Detail stars cache empty previews

Severity: `non-severe`

Starring through [MarketDetailHeader](packages/kit/src/views/Market/MarketDetailV2/components/MarketDetailHeader/MarketDetailHeader.tsx:237-244) or [StockDesktopLayout](packages/kit/src/views/Market/MarketDetailV2/layouts/StockDesktopLayout.tsx:204-213) calls `rememberWatchlistListingPreview` without a name or logo. The new favorite keeps its fallback icon until listing quotes arrive.

originalix:

Review of commit `55b484f934d9` finished, but some comments could not be published automatically.

sidmorizon:
Devin Review has been triggered for this PR.

[View Devin Review ↗](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13473>)

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-77715afd-82fd-5744-8192-509bc32f9594)

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/13473)

sidmorizon:
Cursor review failed.

Remote error: `cursor_request_failed: TimeoutError: The operation timed out.`

PR #7937, opened 2025-07-30
fix: resolve market homepage scrolling issue && implement lazy loading for heavy rendering page on Native && fix NFT list placeholder display && correct staking homepage background color in dark mode && fix scroll issues in Carousel component

huhuanming:
## Summary by CodeRabbit

* **New Features**
  * Added a `lazyLoad` prop to various Page components, enabling improved loading behavior on native platforms.
  * Section list components now allow section headers to scroll with content, enhancing navigation experience.

* **Bug Fixes**
  * NFT grid layouts now maintain consistent row lengths by padding incomplete rows.
  * Table scrolling and tab content rendering are now platform-aware for a smoother user experience.

* **Refactor**
  * Unified and renamed the loading control prop from `skipLoading` to `lazyLoad` across multiple components.
  * Centralized and memoized page width calculations in Carousel for consistency.
  * Improved documentation clarity for Page-related properties.

* **Style**
  * Updated header background color in Earn views for visual consistency.
  * Adjusted FAQ panel padding for better spacing.

* **Documentation**
  * Enhanced in-code documentation for Page component properties.

PR #7710, opened 2025-07-03
chore: upgrade to react-native@0.79/expo@53/react@19

huhuanming:
## Summary by CodeRabbit

* **New Features**
  * Upgraded major dependencies including React, React Native, Expo, and related libraries for improved performance and compatibility.
  * Enabled the new React Native architecture and Hermes JavaScript engine for mobile and desktop apps.
  * Introduced a new image component with enhanced loading, retry, skeleton, and fallback support.
  * Added keyboard provider integration and a custom event emitter hook for navigation events.

* **Bug Fixes**
  * Improved barcode scanning by updating camera and QR code scanning components.
  * Enhanced privacy configuration and resource handling for iOS.
  * Fixed Android back button listener cleanup and various type safety issues.

* **Refactor**
  * Transitioned iOS app entry and delegate logic from Objective-C to Swift.
  * Simplified or disabled several internal authentication and camera permission flows.
  * Replaced legacy image preloading with new expo-image based utilities.
  * Updated navigation and tab bar gesture handling on iOS.
  * Removed Flipper integration and disabled related debug features.

* **Chores**
  * Removed deprecated, unused, or legacy files and dependencies.
  * Updated build scripts, configuration files, and privacy settings.
  * Improved TypeScript typings and added explicit type assertions across components.
  * Adjusted Gradle and Podfile configurations to support new architecture and deployment targets.

* **Style**
  * Minor import and code cleanup across multiple files.
  * Improved component declarations and typing consistency.

PR #13478, opened 2026-09-16
perf: remove the Android default page loading overlay

huhuanming:
## Problem

On Android, opening most pages shows a full-screen spinner before any content. iOS does not.

## Root cause

**1. `Page lazyLoad` overlay (the main one)** — `BasicPage.native.tsx`

`LoadingScreenAndroid` covered the first ~160ms of every `lazyLoad` page with a fixed-duration spinner, on a timeline with no connection to whether the page was ready:

| t | |
|---|---|
| 0ms | spinner overlay up, children **not mounted** |
| 10ms | children start mounting, behind the overlay |
| ~160ms + idle | overlay fades out |

This hit 11 of the most frequently opened pages: TokenDetails, AccountSelectorStack, TokenSelector, PureChainSelector, EditableChainSelector, SwapMainLandModal, SwapTokenSelectModal, SwapKLineContent, MobileTabListModal, BookmarkListModal, HistoryListModal.

iOS dropped this same overlay in 06147be37d, once 3f3baf5d28 fixed the root cause with `[UIView performWithoutAnimation:]` (patched into react-native): Fabric recycles native views that retain stale frames, and UIKit was capturing the frame correction as an implicit animation inside the modal presentation block. That mechanism is UIKit-only — Android has no implicit view-frame animations. Android's copy was left behind when the shared `LoadingScreen` was split, never re-examined.

**2. `LazyLoadPage` Suspense fallback (Android only)** — `LazyLoadPage/index.tsx`

Every lazily loaded route defaults to a full-screen `Spinner` while its segment resolves. Android runs **every** navigation with `animation: 'none'` (`GlobalScreenOptions.native.ts`), so nothing masks it and it flashes on each page open.

## Changes

- **Remove the Android page overlay** — `LoadingScreenAndroid` / `Loading` / `AbsoluteContainer` / `LoadingScreen` / `AbsoluteLoadingContainer` removed; `BasicPage` renders children directly, matching the web `BasicPage.tsx` signature.
- **Drop the lazy route spinner on Android only** — `fallback ?? (platformEnv.isNativeAndroid ? null : defaultFallback)`. iOS animates the push, and web/desktop/extension resolve the route over the network where a blank page reads as a hang — `LazyRetryBoundary` also renders this node through its 150/600ms retry backoff, so those platforms keep the spinner. 16 routes pass an explicit `fallback` and are unaffected.
- **Remove the dead `lazyLoad` and `fullPage` props** from `IBasicPageProps`, both `PageContainer` layers, `Page`, 18 call sites and the developer gallery. Neither was read by anything that produces behavior: both `BasicPage` implementations destructure only `children, testID, backgroundColor`.

## Why `fullPage` was dead

It was not always. Introduced in cb4e68467e (#4784, **OK-28880**) to fix pages not filling the iOS tab container, via a min-height of window height minus tab bar height. That lived inside `LoadingScreen`, which was then **opt-out** (`skipLoading = false`), so it applied to every page.

| | commit | `fullPage` |
|---|---|---|
| introduced | cb4e68467e #4784 | ✅ applies to every page |
| RN 0.79 upgrade | db8ab67b12 #7710 — `skipLoading` flipped to `true`, branch commented out | ❌ **died here** |
| overlay returns as opt-**in** `lazyLoad` | f1848d596a #7937 | ❌ none of its 7 pages opted in |

The height fix was coupled to the loading overlay and went down with it. It has been inert for about a year with no reported height regression on any of its pages (Home, Swap, Browser ×2, PerpTrade, DeviceManagementList, EarnBlockedOverview), so it is removed rather than restored. If a page stops filling the iOS tab container, the removal commit message records the history — the fix is a min-height on `BasicPage`'s root, decoupled from any loading overlay.

`skipLoading` no longer exists on the Page API; it was renamed to `lazyLoad` in f1848d596a.

## Verification

| | |
|---|---|
| `yarn agent:check --profile pr` | all PASS |
| `yarn tsc:only` | exit 0 |
| `jest` — every touched view directory + Page / Dialog / lazyLoad | 176 suites, 1611 tests pass |

Not yet verified on a device.

## Risk

`lazyLoad` was documented as *reduce stuttering when heavy rendering Native pages*. Removing the cover means a heavy page's mount is now visible rather than hidden. Worth checking on a low-end Android: Swap token select, TokenSelector, chain selector, browser tab list — confirm no frame of unlaid-out content on open.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01P9HDmMK3AHwpcyMwYurn3M

packages/kit/src/components/LazyLoadPage/index.tsx
```diff
@@ -24,17 +20,18 @@ export function LazyLoadPage<
...
+  // spinner is never masked by a transition and flashes on every page open.
+  // Rendering nothing leaves LazyLoadPageContainer's background on screen
+  // instead; routes that want visible loading state pass `fallback` explicitly.
   const LazyLoadComponent = LazyLoad<IExtractComponentProps<T>>(
     factory as () => Promise<{
       default: ComponentType<IExtractComponentProps<T>>;
     }>,
     delayMs,
-    fallback ?? defaultFallback,
+    fallback ?? null,
```
sidmorizon: # 🔶 P1 Android-only rationale applied globally: every lazy route now loads with no indicator on web/desktop/ext/iOS

## Problem

`fallback ?? defaultFallback` became `fallback ?? null` (line 34) for **all** platforms, but the justification written directly above it (lines 23–28) is Android-specific and I confirmed it only holds on Android:

```
packages/components/src/layouts/Navigation/GlobalScreenOptions.native.ts
 36	  if (platformEnv.isNativeIOS) {
 38	      animation: 'slide_from_bottom',
 42	  if (platformEnv.isNativeAndroid) {
 45	      animation: 'none',
106	  if (platformEnv.isNativeIOS) {
109	      animation: 'slide_from_right',
```

`LazyLoadPage` is a cross-platform `@onekeyhq/kit` component. On iOS a navigation *does* run a slide transition, and on web/desktop/extension `factory` is a network/disk chunk `import()` rather than a resident RN segment — neither platform has the "unmasked flash" the comment describes, and both now lose their loading state.

This is `CLAUDE.md`'s "Use platform-specific files or `Platform.select` for platform behavior; do not apply global style fixes to platform-specific bugs."

Two things widen the blast radius:

1. **Scope.** `LazyLoadPage(` / `LazyLoadRootTabPage(` matches 461 lines under `packages/`. Only `views/Onboardingv2/router/index.tsx` and `.../index.web-only.tsx` pass an explicit `fallback`; every other route inherits the new `null`.
2. **The retry path, not just the happy path.** `null` also lands on the error boundary, which is not a fast path:

```
packages/shared/src/lazyLoad/index.tsx
 32	const RETRY_BACKOFF_MS = [150, 600];
192	    if (this.state.phase === 'retrying') return this.props.fallback ?? null;
```

and `<Suspense fallback={fallback}>` at line 251 takes the same `null`. `RETRYABLE_CODES` (lines 43–47) includes `SPLIT_BUNDLE_TIMEOUT` and `SPLIT_BUNDLE_NOT_FOUND`, so a slow/failed segment is an expected case, and the boundary renders nothing through the whole 150/600 ms backoff.

## Impact

On iOS the modal slides up (`slide_from_bottom`) over an empty `$bgApp` body. On web/extension a route on a slow connection renders an empty page for the entire chunk fetch, with no progress indication — which reads as a hang, not as loading. On native, a transient segment error now shows nothing at all rather than a spinner while it retries.

## Recommendation

Keep the change scoped to the platform whose problem it solves. `platformEnv` is already imported at line 6, so the minimal fix is at line 34:

```ts
fallback ?? (platformEnv.isNativeAndroid ? null : defaultFallback),
```

Alternatively restore the default but gate it behind a short delay so it never flashes on a fast resolve — that fixes Android without stripping the indicator elsewhere.

Separately, the comment's claim on lines 27–28 that "Rendering nothing leaves LazyLoadPageContainer's background on screen instead" is not true on the `unStyle` branch (lines 39–41), which returns `<LazyLoadComponent />` with no `Stack` wrapper. No call site sets `unStyle: true` today so that is latent, but the comment should not assert a background that branch does not render.

[Claude session ↗](https://claude.ai/code/session_01P9HDmMK3AHwpcyMwYurn3M)

---

packages/components/src/layouts/Page/type.ts
```diff
@@ -30,9 +30,11 @@ export type IBasicPageProps = PropsWithChildren<
      */
     safeAreaEnabled?: boolean;
     /** @platform native
-     * @description Lazy load. The default value is false.
-     * Mainly used to reduce stuttering when heavy rendering Native pages.
-     * If the page doesn't have much content on initial render, this doesn't need to be enabled.
+     * @deprecated No-op. This used to gate a fixed-duration spinner overlay that
```
sidmorizon: # 🔷 P2 `fullPage` loses its last consumer in this PR but is left undocumented, while `lazyLoad` beside it gets `@deprecated`

## Problem

This hunk correctly marks `lazyLoad` as a no-op. `fullPage`, seven lines above at line 26, becomes exactly as dead in the same commit and gets no marker.

At the base commit `fullPage` still had a consumer:

```
ede1482 packages/components/src/layouts/Page/BasicPage.native.tsx
127	  const minHeight = useMinHeight(fullPage);
159	    <LoadingScreenAndroid fullPage={fullPage} backgroundColor={backgroundColor}>
188	          <LoadingScreen fullPage={fullPage} backgroundColor={backgroundColor}>
```

This PR deletes all of it. At HEAD neither platform implementation reads the prop — `BasicPage.native.tsx:51-55` and `BasicPage.tsx:41-45` both destructure only `children, testID, backgroundColor`. (It was already *effectively* inert, since `useMinHeight` returned `undefined` off iOS and `LoadingScreenAndroid` was only reached on the non-iOS branch — so this is a documentation/cleanup gap, not a behaviour change.)

It is nonetheless still declared and still threaded through two layers that now discard it:

```
packages/components/src/layouts/Page/PageContainer.native.tsx:29,83,110
packages/components/src/layouts/Page/PageContainer.tsx:14,24,38
packages/components/src/layouts/Page/index.tsx:44,84
```

and eight production call sites still pass it as if it does something: `views/Home/pages/HomePageView.tsx:1263`, `views/Swap/pages/SwapPageContainer.tsx:74`, `views/Discovery/pages/Browser/Browser.native.tsx:564`, `Browser.ext.tsx:101`, `views/PerpTrade/pages/PageWebviewPerpTrade.tsx:269`, `views/DeviceManagement/pages/DeviceManagementListModal/index.tsx:497`, `views/Earn/components/EarnBlockedOverview.tsx:35`.

## Impact

The PR takes the trouble to warn "do not add new ones" on `lazyLoad`, then leaves the prop directly above it in the identical state with no warning at all. Readers will keep writing `<Page fullPage>` expecting layout behaviour that no longer exists on either platform, and the now-pointless threading through both `PageContainer` files (including as a `useMemo` dependency) is left unexplained.

## Recommendation

Handle both dead props the same way in this commit. Either give `fullPage` an `@deprecated` note at line 26 mirroring the one added here for `lazyLoad`, or delete it outright — remove it from `IBasicPageProps:26`, from `Page/index.tsx:44,84`, `PageContainer.tsx:14,24,38`, `PageContainer.native.tsx:29,83,110`, and from the eight call sites above. Leaving one of the two annotated and the other silent is the outcome to avoid.

[Claude session ↗](https://claude.ai/code/session_01P9HDmMK3AHwpcyMwYurn3M)

---

packages/kit/src/components/LazyLoadPage/index.tsx
huhuanming: Fixed in ad00fca4e1 — the platform scoping point is right and it is a direct `CLAUDE.md` violation on my side, so line 34 is now:

```ts
fallback ?? (platformEnv.isNativeAndroid ? null : defaultFallback),
```

`defaultFallback` is restored and the comment rewritten to say why only Android drops it. The retry-path argument is the one that convinced me: `LazyRetryBoundary` renders this same node for the whole 150/600ms backoff, and `SPLIT_BUNDLE_TIMEOUT` / `SPLIT_BUNDLE_NOT_FOUND` are expected cases, so a blank boundary there was strictly worse. I also removed the sentence asserting the `LazyLoadPageContainer` background, which as you note is not what the `unStyle` branch renders.

Two supporting claims do not match the tree, though neither changes the conclusion:

- *"Only `views/Onboardingv2/router/index.tsx` and `.../index.web-only.tsx` pass an explicit `fallback`"* — Onboardingv2 passes `false` as the third positional arg (`unStyle`), not a fallback. The real explicit-fallback call sites are `RootTabLoadingFallback` (Perp, PerpTrade, Earn, ReferFriends, Market), `earnLazyPageFallback` (6 Earn routes across the Discovery and Earn routers), `MarketDetailV2LoadingFallback`, and 3 dialogs in `useOneKeyAuth.tsx`.
- *"No call site sets `unStyle: true` today so that is latent"* — `useOneKeyAuth.tsx` sets it at 3 call sites. They all pass their own fallback, so the overclaim was still worth fixing, but it is reachable rather than latent.

packages/components/src/layouts/Page/type.ts
huhuanming: Agreed — fixed in ad00fca4e1 by annotating rather than deleting, since removing `fullPage` would touch both `PageContainer` files, `Page/index.tsx` and 8 call sites, which is wider than this PR should reach.

`fullPage` now carries an `@deprecated` note mirroring the one on `lazyLoad`, recording that its last consumer was the overlay min-height calculation this PR removed, that both `BasicPage` implementations ignore it, and that `Page` / `PageContainer` still thread it through. Deleting it outright is worth a follow-up.

packages/kit/src/components/LazyLoadPage/index.tsx
sidmorizon: Verified `ad00fca4` at head — line 45 now reads `fallback ?? (platformEnv.isNativeAndroid ? null : defaultFallback)`, `defaultFallback` and the `Spinner` import are back, and the background sentence is gone. Resolving.

Both of your corrections are right, and I checked each in the tree rather than taking them on trust:

- `Onboardingv2/router/index.tsx:15-48` passes `false` as the **third** positional arg, which is `unStyle`, not `fallback` — six call sites, none of them supplying a fallback. My "only two explicit-fallback call sites" claim was wrong; the real set is the one you list.
- `components/OneKeyAuth/useOneKeyAuth.tsx:43,53,63` do set `unStyle: true` (all three also pass their own `<Spinner>`), so that branch is reachable, not latent as I wrote.

Neither error affected the conclusion, but both overstated the evidence — the scope claim in particular was wrong in a way that made the blast radius sound larger than it was. Sorry for the noise.

[Claude session ↗](https://claude.ai/code/session_01P9HDmMK3AHwpcyMwYurn3M)

---

originalix:

Review of commit `ad00fca4e167` finished, but some comments could not be published automatically.

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13478>)

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-c854ac69-88a0-538b-89ae-267013b7d235)

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/13478)

sidmorizon:
Cursor review failed.

Remote error: `cursor_request_failed: HTTP 503 Service Unavailable: Server at capacity`

PR #13481, opened 2026-09-16
fix: remove dead scroll space after switching home tabs

huhuanming:
Desktop: switching from 历史记录 (History) to 现货 (Spot) leaves thousands of px of empty scroll below the content.

Two independent bugs turned up. They are on different code paths and are split into one commit each.

## 1. Shared tab scroll extent only ever grew

`Tabs.Container` pins one height on the shared pager for every tab. It measures the newly focused tab while the **previous** tab's height is still applied, and `Tabs.ScrollView` is `flex={1}`, so the focused tab is stretched to that stale height — `max(scrollHeight, clientHeight)` reads it straight back.

The extent therefore only ever grew. Leaving History (27282px) for Spot (1545px) stranded ~25700px of dead scroll.

`getTabContentHeight` now derives the height from the element's in-flow content when it is stretched, and only trusts that when it proves the box is taller than its content. The measurement is **read-only on purpose**: the obvious alternative — clearing the pinned height before measuring — collapses the scroll container, and Chrome then clamps `scrollTop` to 0. Measured: 600 → 0. Since `apply()` runs from a `ResizeObserver`, that would jump the page to the top on every content change.

Guards worth calling out, each verified against real Chromium layout:

| Case | Behaviour |
| --- | --- |
| `display: contents` child | Descend into it — Tamagui emits these wrappers, and skipping one measured 80px instead of 980px |
| Child overflowing its own border box | Use `max(offsetHeight, scrollHeight)` when `overflow-y: visible`, so content is not clipped |
| Child that scrolls its own content | Contributes only its box, never its scrolled-away content |
| Virtualized inner container taller than the union | Never undercut it |
| `flex: 1` child, or no children at all | Falls back to the previous behaviour — no regression, no improvement |

## 2. NFT grid reserved a phantom footer row

`NFTListView` passed `ListFooterComponent` as `<>{cond ? <Stack h="$16" /> : null}</>`. A fragment is always truthy, so `List.tsx` always appended a `{ type: 'footer' }` entry. In the multi-column path react-virtualized lays **every** entry out as one square cell (`clientWidth + 60`), and `NFTListView` pads its data to a whole number of columns — so the footer always landed at the start of a new row and reserved a full ~340px cell that rendered nothing.

Evidence: inner container `height: 4420px`, last row `top=4080 height=340`, lowest painted pixel `4058`.

The extension's floating-panel clearance moves into `contentContainerStyle` padding, which creates no cell. The extension now gets the 64px it actually wants instead of an empty row.

This one is **not** caused by, and was not fixed by, commit 1 — NFT renders through `Tabs.FlatList` → `List.tsx` and never takes the `Tabs.ScrollView` measurement path. Measured identically (410px) with and without commit 1.

## Verification

New scenario `scenarios/regression.mjs tabs-blank-space-desktop`: all 5 home tabs, every ordered pair (source left scrolled to its bottom), plus DeFi's inner protocol strip, its show more/less toggle, and a round trip — 33 samples, driven with **real wheel input**, not `scrollTop` assignment.

```sh
REGRESSION=1 node scenarios/regression.mjs tabs-blank-space-desktop
```

It fails on the pre-fix build (worst 26425px of dead space) and passes after. Bottom-of-tab blank space, measured with real wheel scrolling:

| Tab | before | after commit 1 | after commit 2 |
| --- | --- | --- | --- |
| Spot | 25769px | 32px | 32px |
| Perps | 620px | 32px | 32px |
| DeFi | 32px | 32px | 32px |
| NFT | 410px | 410px | 70px |
| History | 49px | 49px | 49px |

All five tabs reach their bottom and nothing is clipped. The existing `tabs-scroll-extent-desktop` scenario — which guards the **opposite** failure, an extent that stays too short — still passes, including 8 DeFi↔NFT round trips and late async growth (container correctly grows 5544→5640).

Two notes on the harness, since the first two versions of it were wrong in the same way and both produced false passes:

- Walking element rects measures the **stretched** wrappers, so "content height" comes back equal to the pager height and dead space always reads 0.
- Hit-testing counts empty wrappers as content, which reported NFT's real 410px gap as 2px.

It now requires a hit to actually paint (direct text, a replaced element, or a background/border), and cross-checks with the lowest painted leaf. Both signals were confirmed against screenshots.

`KNOWN_DEAD_SPACE_TABS` is the quarantine hook that let commit 1 land green while NFT was still broken. It is empty now; entries need a reason and a removal condition.

## Not covered

- Only Home was exercised live. The other `Tabs.Container` consumers (MarketDetail, TokenDetails, Borrow ReserveDetails, Earn, ReferFriends) were audited by hand and covered by the Chromium harness, not driven in the app.
- One window size only, so the `gtXl` Recent-activity sidebar branch is untested.
- Native is unaffected: `Container.tsx` is web-only, native uses `react-native-collapsible-tab-view`.
- A `Tabs.ScrollView` whose only child is `flex: 1` (e.g. ReferFriends' loading state) still keeps the stale height. Not a regression; fixing it needs recursion into stretched children.

## Checks

`yarn agent:check --profile pr` passes except `tsc-staged`, which reports ~5 cross-checkout type-identity errors in files this PR does not touch (`packages/shared/src/logger/*`). That is a worktree artifact, not a defect in this branch: the repo's `tsc` script writes `--tsBuildInfoFile` into the shared yarn cache folder, so a worktree inherits the main checkout's incremental state and sees every workspace type twice. Re-running with an isolated build-info file gives **zero** errors:

```sh
node ./node_modules/@typescript/native/bin/tsc -p ./tsconfig.json --noEmit --tsBuildInfoFile .tmp/ts-cache --pretty
```


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_017q7s6ygDrrXPomdtAnCJGB

originalix:

Review of commit `b2d5ef6669bb` did not finish successfully.

scenarios/regression.mjs
```diff
@@ -1037,6 +1037,419 @@ async function runTabsScrollExtentDesktop(cdpUrl) {
...
+    }
+
+    const rightArrow = page
+      .locator('[data-testid="defi-protocol-chip-scroll-btn-right"]:visible')
+      .first();
+    if ((await rightArrow.count()) > 0) {
+      await rightArrow.click({ force: true }).catch(() => {});
+      await sleep(900);
+      await sample('defi chip strip scrolled');
+    }
```
sidmorizon, line 1490: # [🟡 P2] Chip-strip scroll step targets a test id that does not exist, so the sample never runs

## Problem

`scenarios/regression.mjs:1378` looks up `[data-testid="defi-protocol-chip-scroll-btn-right"]`. No component renders that id. The only scroll affordance on the DeFi protocol strip is the `IconButton` in `packages/kit/src/views/Home/components/DeFiListBlock/ProtocolChipStrip.tsx:222`, whose id comes from `HomeTestIDs.defiProtocolChipScrollBtn(side)` in `packages/kit/src/views/Home/testIDs.ts:65-66` and resolves to `home-defi-protocol-chip-scroll-right-btn` (different prefix and different word order). A grep across `packages/` and `apps/` finds the scenario's string only in this file.

Because the step is guarded by `if ((await rightArrow.count()) > 0)` at line 1380, the mismatch is silent: the locator matches zero elements and the `defi chip strip scrolled` sample is simply skipped. Every other test id this scenario relies on (`home-tab-*`, `home-defi-tab-content`, `home-render-content-btn`, `password-input`) does exist, so this is the one dead step.

## Impact

The comment at lines 1362-1363 and the PR description both state that the DeFi inner protocol strip is driven "through its own arrow affordance" and counted among the 33 samples. In practice that sample is never taken, so the scenario reports a pass for coverage it does not have. A future regression where scrolling the strip leaves dead space at the bottom of the DeFi tab would not be caught, and nothing in the run output says the step was skipped.

## Recommendation

Use the real id so the step executes, e.g. at line 1378:

```js
.locator('[data-testid="home-defi-protocol-chip-scroll-right-btn"]:visible')
```

Optionally log when the arrow is absent instead of silently skipping, in the same style as the show more/less branch at line 1359, so a future id rename is visible in the run output. If the strip step is intentionally not part of this scenario, delete lines 1377-1384 and drop the "arrow affordance" wording at lines 1362-1363 so the scenario does not claim that coverage.

[Claude session ↗](https://claude.ai/code/session_017q7s6ygDrrXPomdtAnCJGB)

---

packages/components/src/composite/Tabs/utils.ts
```diff
@@ -10,3 +10,50 @@ export const parseCssSize = (value: string | undefined) => {
...
+        visit(child);
+        return;
+      }
+      // offsetParent is null for `display: none` and `position: fixed`, so
+      // both drop out here; an absolutely positioned child still reports one.
+      if (
+        child.offsetParent !== htmlElement ||
+        childStyle.position === 'absolute'
+      ) {
+        return;
```
sidmorizon: 🟡 P2: Positioned tab content gets clipped

Severity: `non-severe`

Mixed flow and absolute content makes `getInFlowContentBottom` omit the absolute child. The resulting positive underestimate shrinks the pager and clips that child.

scenarios/regression.mjs
```diff
@@ -1037,6 +1037,419 @@ async function runTabsScrollExtentDesktop(cdpUrl) {
...
+  );
+  const worstBlank = samples.reduce(
+    (acc, s) => (s.blankSpacePx > acc ? s.blankSpacePx : acc),
+    0,
+  );
+
+  return reportVerification('tabs-blank-space-desktop', [
+    {
+      name: `no dead space at bottom (<= ${BLANK_SPACE_TOLERANCE_PX}px)`,
+      pass: blankFailures.length === 0,
```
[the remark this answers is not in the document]
huhuanming: 已修复（a80fb21679）。新增门禁检查「every sample reached the bottom」，任何样本没滚到底都会失败。读不到滚动容器时，`measureTabExtent` 返回 null，样本记为 unreadable，由「every sample readable」直接判失败；容器存在但滚不动时，则由新检查判失败。

用变异体验证过它确实会失败：把 `scrollToBottom` 改成空操作后，只有这一项 FAIL（如 `rest DeFi stopped at 5265`），而「no dead space」依然 PASS（最差 62px）。这正是你指出的情况——中途采样时视口底部几乎总有内容，没有这条断言，测试就会假通过。

scenarios/regression.mjs
```diff
@@ -1037,6 +1037,419 @@ async function runTabsScrollExtentDesktop(cdpUrl) {
...
+      } else {
+        stuckFor = 0;
+      }
+      prev = top;
+    }
+  };
+
+  const gotoTab = async (tab, settleMs = 1600) => {
+    await page.click(`[data-testid="${tab.testId}"]`, { force: true });
+    await sleep(settleMs);
```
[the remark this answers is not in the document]
huhuanming: 已修复（a80fb21679）。phase 1 逐个进入 tab 时记录它在 pager 中的实际索引；`gotoTab` 点击后轮询，直到可见分页索引与之一致。phase 2 的每个样本都校验 `focusedIndex`，源 tab 或目标 tab 任一没切过去都记为 wrong tab，对应门禁项「every sample measured the requested tab」。

索引是运行时学出来的，而不是按 tab 栏顺序假设，因为 pager 并非每个 tab 一个子节点（web 端 Perps 不在 pager 里）。

变异体验证：phase 2 禁用点击后，只有这一项 FAIL（全部样本 `focusedIndex=4`，即 phase 1 最后停留的 History），其余检查全部 PASS。

scenarios/regression.mjs
huhuanming: Fixed in a80fb21679. The step now uses `home-defi-protocol-chip-scroll-right-btn` (`HomeTestIDs.defiProtocolChipScrollBtn('right')`), and logs `defi chip strip arrow not present; skipping that sample` when the arrow is absent instead of skipping silently, so a future rename shows up in the run output.

The sample now actually runs: the scenario reports 34 samples instead of 33, and `defi chip strip scrolled` appears in the log.

packages/components/src/composite/Tabs/utils.ts
huhuanming: Good catch — this was a real clip, and my earlier Chromium harness had even asserted the clipped value as the expected result.

Fixed in c759185227: absolutely positioned children now stay in the union. One sized by the stretched box itself (`top: 0; bottom: 0`) reports the stale height, which makes the caller fall back to the box instead of shrinking, so it can never clip. `position: fixed` children still drop out, since their `offsetParent` is null.

Measured in Chromium with the pager pinned at 5000px:

| case | before | after |
| --- | --- | --- |
| in-flow 600 + abs `top:700 h:300` | 600 (clips) | 1000 |
| in-flow 600 + abs `top:0; bottom:0` | 600 (clips) | 5000 (fallback) |
| in-flow 1400 only | 1400 | 1400 |

`tabs-blank-space-desktop` (34 samples) and `tabs-scroll-extent-desktop` both still pass.

scenarios/regression.mjs
```diff
@@ -1037,6 +1037,500 @@ async function runTabsScrollExtentDesktop(cdpUrl) {
...
+
+  // Phase 1 — each tab on its own, scrolled to the bottom. Also calibrates
+  // each tab's pager index for the switch assertions in phase 2.
+  for (const tab of available) {
+    await gotoTab(tab, 2200);
+    const m = await sample(`rest ${tab.name}`);
+    if (m && m.focusedIndex >= 0) {
+      tab.pagerIndex = m.focusedIndex;
+    }
+  }
```
sidmorizon: # [🟡 P2] Phase-1 calibration accepts a pager that did not move, so a stalled click poisons the new "requested tab" gate

## Problem

The new gate depends entirely on the `pagerIndex` learned in phase 1, but phase 1 records whatever page is on screen after the click with no check that the pager actually moved and no check that the index is unique:

```js
for (const tab of available) {
  await gotoTab(tab, 2200);
  const m = await sample(`rest ${tab.name}`);
  if (m && m.focusedIndex >= 0) {
    tab.pagerIndex = m.focusedIndex;
  }
}
```

At this point `tab.pagerIndex` is still undefined, so `gotoTab` returns at `scenarios/regression.mjs:1310-1312` before its polling loop runs. If a click "silently does nothing" here, which is the failure the comment at lines 1313-1315 describes, the tab inherits the previous tab's index. Every later `X -> thatTab` sample then polls until the *previous* tab's page is visible, passes the `m.focusedIndex !== expectedTab.pagerIndex` check at line 1343, and is recorded as a clean measurement of a page that was never shown.

There is a code path in the app that produces exactly this input. `HOME_TAB_TEST_IDS` at line 1061 includes `home-tab-perps`, and when `perpTabShowWeb` is set, `packages/kit/src/views/Home/pages/HomePageView.tsx:593-599` removes Perps from `pagerTabConfigs` while lines 626-629 route its tab-bar press to `navigation.switchTab(ETabRoutes.WebviewPerpTrade)`. The pager never moves, so Perps learns the index of whichever tab preceded it. The comment at lines 1288-1290 acknowledges this topology, but the calibration does not handle it. Whether that config is active depends on `usePerpWeb` in the remote perp config, so this is not guaranteed to fire on every desktop build.

## Impact

The gate added in this commit can report "every sample measured the requested tab" as PASS for samples that measured a different page, which is the false pass the gate was added to close. The calibration is the trust anchor for all of phase 2 and for the `defi chip strip scrolled` sample, so one stalled click in phase 1 silently invalidates the whole run's switch assertions.

## Recommendation

Make the calibration prove the switch and reject duplicates. Read the focused index before each click, then only accept the learned index when it changed from the previous tab (for every tab after the first) and no earlier tab already owns it. Otherwise log and exclude the tab from phase 2 so its samples are marked `wrongTab` rather than silently passing:

```js
let previousIndex = -1;
for (const tab of available) {
  await gotoTab(tab, 2200);
  const m = await sample(`rest ${tab.name}`);
  const idx = m?.focusedIndex ?? -1;
  const taken = available.some((t) => t !== tab && t.pagerIndex === idx);
  if (idx >= 0 && !taken && (previousIndex < 0 || idx !== previousIndex)) {
    tab.pagerIndex = idx;
    previousIndex = idx;
  } else {
    log(`${tab.name}: pager did not move or index ${idx} already taken; excluded from switch assertions`);
  }
}
```

Then filter `available` to tabs with a numeric `pagerIndex` before phase 2 at line 1368, so an excluded tab cannot appear as a source or target with a default-pass `wrongTab`.

[Claude session ↗](https://claude.ai/code/session_017q7s6ygDrrXPomdtAnCJGB)

---

scenarios/regression.mjs
```diff
@@ -1037,6 +1037,500 @@ async function runTabsScrollExtentDesktop(cdpUrl) {
...
+      log('defi chip strip arrow not present; skipping that sample');
+    }
+
+    // Round trip back into DeFi with its inner state already dirty.
+    const other = available.find((t) => t.testId !== defi.testId);
+    await gotoTab(other);
+    await scrollToBottom();
+    await sleep(600);
+    await gotoTab(defi, 2500);
+    await sample(`defi round trip via ${other.name}`);
```
sidmorizon: # [🟡 P2] Phase-3 samples bypass the switch verification this commit added, including the DeFi round trip

## Problem

The increment changes `sample` to `(label, expectedTab = null, switched = true)` at `scenarios/regression.mjs:1324` and makes `gotoTab` return whether the target page became visible, but phase 3 still calls both the old way. Only `defi chip strip scrolled` at line 1437 passes `defi`; the other six phase-3 samples pass nothing, so `wrongTab` is hard-wired to `false` for them by the defaults at lines 1338-1343:

```js
await gotoTab(defi, 3000);            // line 1383, result discarded
...
await sample('defi cold enter');      // line 1389, no expectedTab
...
await sample('defi show more (grow)');   // line 1397
await sample('defi show less (shrink)'); // line 1404
await sample(`defi inner chip ${i + 1}`); // line 1421
...
await gotoTab(other);                 // line 1444, result discarded
await scrollToBottom();
await sleep(600);
await gotoTab(defi, 2500);            // line 1447, result discarded
await sample(`defi round trip via ${other.name}`); // line 1448, no expectedTab
```

The round trip at lines 1444-1448 is an `other -> DeFi` switch with the source scrolled to its bottom, which is precisely the state phase 2 guards with `switched` and `expectedTab`. A stalled return into DeFi here is recorded as a DeFi measurement.

## Impact

The commit message states that "every later sample must match" the learned pager index. Six samples cannot, so the "every sample measured the requested tab" gate can pass while the DeFi grow/shrink, chip, and round-trip samples measured another page. Those are the samples the PR description singles out as the height transitions most likely to strand dead space.

## Recommendation

Thread the results through the same way phase 2 does:

```js
const enteredDefi = await gotoTab(defi, 3000);
...
await sample('defi cold enter', defi, enteredDefi);
...
await sample('defi show more (grow)', defi);
await sample('defi show less (shrink)', defi);
await sample(`defi inner chip ${i + 1}`, defi);
...
const left = await gotoTab(other);
await scrollToBottom();
await sleep(600);
const back = await gotoTab(defi, 2500);
await sample(`defi round trip via ${other.name}`, defi, left && back);
```

[Claude session ↗](https://claude.ai/code/session_017q7s6ygDrrXPomdtAnCJGB)

---

scenarios/regression.mjs
huhuanming: Fixed in 900acd24ee. Calibration now steps onto a different tab first, so the click on the target is an observable move regardless of which tab the app started on, and it only accepts an index that changed and that no other tab already owns. Anything else is excluded from the matrix, and the gate — now "every tab calibrated and every sample measured the requested tab" — fails naming the tab and the reason. `gotoTab` also no longer reports success for an uncalibrated tab, and phases 1–3 iterate calibrated tabs only.

I reproduced the exact input you described with a mutant that skips the Perps click during calibration:

```
calibrate Perps: pager did not move (stayed at 2); excluded from the tab matrix
FAIL every tab calibrated and every sample measured the requested tab | calibrate Perps: pager did not move (stayed at 2)
```

Index 2 is DeFi's page. Under the old calibration Perps would have learned it, and every `-> Perps` sample would have measured DeFi and passed. Every other check stayed green in that mutant, so this gate is the only thing catching it. The unmutated run calibrates all five tabs to indices 0–4 and passes.

scenarios/regression.mjs
huhuanming: Fixed in 900acd24ee, threaded the same way phase 2 does: `enteredDefi` feeds the cold-enter sample, the grow/shrink and chip samples all name `defi`, and the round trip is recorded with `left && back`, so both legs must be proven. `defi` and the round-trip `other` are drawn from calibrated tabs only.

Checked with a mutant that skips the DeFi click during phase 3:

```
defi cold enter: ... pager=4140 ... WRONG TAB (focusedIndex=3)
defi round trip via Spot: ... pager=1545 ... WRONG TAB (focusedIndex=0)
FAIL every tab calibrated and every sample measured the requested tab | defi cold enter(focusedIndex=3), defi round trip via Spot(focusedIndex=0)
```

The pager heights confirm what was really on screen: 4140 is NFT and 1545 is Spot. Before this change both would have been recorded as clean DeFi samples.

The mutant also showed that chip sampling skipped silently when DeFi wasn't on screen. It now logs, like the grow/shrink and arrow steps already did.

originalix (approved):

PR #13483, opened 2026-09-16
fix: correct firmware transfer copy

wabicai:
## Summary
- Show **Transferring data…** while firmware bytes are still moving, instead of **Installing firmware…**.
- After the app is backgrounded during a BLE transfer and comes back, fail a stalled transfer with the existing transfer-interrupted error and retry, instead of leaving the install page hung with no error.

## Jira
- https://onekeyhq.atlassian.net/browse/OK-63563

## Context
QA on TestFlight (unified install page, OK-61628):

1. During transfer the progress copy still said “正在安装固件…”.
2. Backgrounding the app mid-transfer, then returning, froze progress around 35% with no error. Device logs show the last `ui-firmware-progress` at 15:40:07, resume at 15:40:15, then a hung `EmmcFileWrite` until a manual cancel at 15:45:52.

## Test plan
- [ ] Start a Pro BLE firmware update and confirm the progress line switches to transferring copy while data is sending, then installing/verifying after transfer completes.
- [ ] During transfer, send the app to background and return: if progress does not resume within ~3s, the install page should show the transfer-interrupted error with Retry.
- [ ] Stay on the install page without backgrounding: a healthy transfer should complete as before.
- [ ] Mini / non-transfer stages (preparing, waiting for device, installing) should keep their existing copy.


originalix:

Review of commit `0fb210af9ea1` did not finish successfully.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2

packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts
```diff
@@ -0,0 +1,63 @@
...
+export function isFirmwareTransferInProgress(
+  snapshot: IFirmwareTransferUiSnapshot | undefined,
+): boolean {
+  if (!snapshot) {
+    return false;
+  }
+  if (snapshot.firmwareProgressType === 'installingFirmware') {
+    return false;
+  }
+  if (snapshot.firmwareProgressType === 'transferData') {
```
sidmorizon: # 🟠 P1 Sticky `firmwareProgressType` makes a flashing device look like a stalled transfer

## Problem
`isFirmwareTransferInProgress` treats `firmwareProgressType === 'transferData'` as proof that bytes are still moving (line 38), and `didFirmwareTransferResume` requires a *strict increase* in `firmwareProgress` (lines 58-62), so a frozen value is read as "stalled".

But `firmwareProgressType` is sticky in `hardwareUiStateAtom`. The `FIRMWARE_TIP` branch of the reducer explicitly carries the previous values forward:

```ts
// ServiceHardware.ts:1667-1681
} else if (isSameFirmwareDevice && appliedUiRequestType === EHardwareUiStateAction.FIRMWARE_TIP) {
  firmwarePayload = {
    ...appliedPayload,
    firmwareProgress: previousState?.payload?.firmwareProgress,
    firmwareProgressType: previousState?.payload?.firmwareProgressType,
    ...
```

It is only overwritten by a *new* `FIRMWARE_PROGRESS` event (`ServiceHardware.ts:1423-1424`). So once the byte transfer finishes at `firmwareProgress: 100 / 'transferData'`, the atom keeps reporting `'transferData'` at `100` for the whole device-side erase/write window — during which the SDK emits no progress events at all.

A background→foreground round-trip inside that window therefore evaluates to `isFirmwareTransferInProgress → true`, `didFirmwareTransferResume → false` (`100 > 100` is false), and the guard calls `serviceHardware.cancel({ immediate: true })` **while the device is writing flash**.

The unit test at `FirmwareTransferStallGuard.test.ts:29-42` only covers the case where a newer `installingFirmware` progress event has already landed — precisely the state the stale window does *not* have, so the suite stays green.

## Impact
Cancelling an in-flight firmware write on a completely healthy update, triggered by nothing worse than the user glancing at another app. Depending on where the device is in erase/write, this is recovery-mode / bricking risk rather than a cosmetic failure.

Note this is latent today because the guard is never started (see the separate thread on `ServiceFirmwareUpdate.ts:336`) — it activates the moment that wiring lands, so it needs fixing *before* the guard is hooked up, not after.

## Recommendation
Do not infer liveness from a carried-forward enum. Either:

1. Accept a snapshot only when the source state has `action === EHardwareUiStateAction.FIRMWARE_PROGRESS` — the UI hook already does exactly this at `useFirmwareUpdateInstallState.ts:178-185` — and drop the `hardwareUiStateCompletedAtom` fallback in `readFirmwareTransferUiSnapshot` (`ServiceFirmwareUpdate.ts:352-360`); or
2. Carry an event timestamp (or `firmwareTransferMetrics.elapsedMs`, already on the payload) into the snapshot and require a genuinely *recent* progress event before declaring the transfer live.

Separately, the `beforeProgress === undefined` path at lines 56-62 should resolve to "resumed", never to an abort.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+    }
+    if (!visible) {
+      this.backgroundedDuringTransfer = true;
+      return;
+    }
+    if (!this.backgroundedDuringTransfer) {
+      return;
+    }
+    this.backgroundedDuringTransfer = false;
+    await timerUtils.wait(FIRMWARE_TRANSFER_RESUME_GRACE_MS);
```
sidmorizon: # 🟠 P1 The 3s grace wait is uncancellable, so a stale handler can abort a later update

## Problem
`await timerUtils.wait(FIRMWARE_TRANSFER_RESUME_GRACE_MS)` (line 380) suspends an already-running handler, and `timerUtils.wait` is a plain `setTimeout` promise with no cancel token.

`stopFirmwareTransferStallGuard()` (lines 346-350) only unsubscribes *future* events:

```ts
private stopFirmwareTransferStallGuard() {
  this.transferStallUnsub?.();
  this.transferStallUnsub = undefined;
  this.backgroundedDuringTransfer = false;
}
```

It cannot cancel a handler already parked on line 380, and it does **not** reset `transferStallAborting` — that is cleared only in `start` (line 338).

There is no generation/epoch token, so after the wait resolves the handler re-validates using only `firmwareUpdateWorkflowRunningAtom.get()` (line 384) — a *global boolean* that a different, newly started update sets back to `true`. A handler left over from a completed or exited workflow therefore passes the re-check and calls `abortFirmwareTransferAfterBackground()` against the new workflow, using `this.transferStallConnectId`, which `startFirmwareTransferStallGuard` (line 340) has by then overwritten with the new device's id.

The handler is also freely re-entrant: `onVisibilityStateChange` fires on every `AppState` transition, and each foreground event spawns another pending 3s wait with nothing deduping them. A background→foreground→background sequence inside the window lets an abort run while the app is backgrounded again.

## Impact
A spurious cancel injected into a *subsequent, healthy* firmware update — the classic leaked-timer failure mode, where the damage lands on a transfer the user started later. Combined with the sticky-`progressType` issue on `FirmwareTransferStallGuard.ts:38`, the worst case is again a cancel landing mid-flash on the wrong device.

## Recommendation
Capture a monotonic epoch in `startFirmwareTransferStallGuard` (mirroring the existing `updateWorkflowSequence` / `isUpdateWorkflowCurrent` pattern at `ServiceFirmwareUpdate.ts:2149` and `:2215`), hold it in the handler's local scope, and re-check it both after the `await` at line 380 and at the top of `abortFirmwareTransferAfterBackground` (line 395). Reset `transferStallAborting = false` inside `stopFirmwareTransferStallGuard` (lines 346-350), not only in `start`. Guard against re-entrancy with a `pendingResumeCheck` flag so one visibility storm cannot queue several aborts.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -94,6 +98,12 @@ import {
 import ServiceBase from '../ServiceBase';
 import serviceHardwareUtils from '../ServiceHardware/serviceHardwareUtils';
 
+import {
+  FIRMWARE_TRANSFER_RESUME_GRACE_MS,
+  didFirmwareTransferResume,
+  getFirmwareTransferUiSnapshot,
+  isFirmwareTransferInProgress,
+} from './FirmwareTransferStallGuard';
```
sidmorizon: # 🟠 P1 New kit-bg module is missing from the mobile module-id registry, breaking the iOS union build

## Problem
This import pulls the new `FirmwareTransferStallGuard.ts` into the background bundle graph, but the module was never added to `apps/mobile/bundle-registry/module-id-registry.json`, and that registry is not among the 9 files this PR changes.

The **Native startup graph budget (24.x)** check fails deterministically as a result:

```
Error: Module is not registered in apps/mobile/bundle-registry/module-id-registry.json:
packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts.
Run `yarn workspace @onekeyhq/mobile module-id:update --map <module-id-map.json>`
and commit the registry update.
```

Verified against this head locally: the registry contains 13 entries for `packages/kit-bg/src/services/ServiceFirmwareUpdate/*` (`ServiceFirmwareUpdate.ts`, `FirmwareUpdateRuntime.ts`, `FirmwarePreparedExecution.ts`, …) and zero for `FirmwareTransferStallGuard.ts`, so this is a missed regeneration step rather than an exempt path. Test files are correctly out of scope — the registry has no `.test.` entries at all — so only the production `.ts` needs registering.

## Impact
The iOS union build fails outright after both graphs are built (main 14,752 modules, background 14,649), so no mobile bundle or bundle-architecture report is produced and the PR cannot ship to TestFlight — which is exactly where the OK-61628 regression this PR targets was reported. It is not a flake and not pre-existing on `x`: it is caused solely by adding a new kit-bg production module.

## Recommendation
Run `yarn workspace @onekeyhq/mobile module-id:update --map <module-id-map.json>` and commit the updated `apps/mobile/bundle-registry/module-id-registry.json` so `FirmwareTransferStallGuard.ts` gets a stable module id alongside its siblings. Worth doing even if the stall-guard wiring is deferred, since the file is imported at module scope here on line 106 and therefore enters the graph regardless of whether the guard is ever started.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+  private backgroundedDuringTransfer = false;
+
+  private startFirmwareTransferStallGuard(connectId: string | undefined) {
+    this.stopFirmwareTransferStallGuard();
+    this.transferStallAborting = false;
+    this.backgroundedDuringTransfer = false;
+    this.transferStallConnectId = connectId;
+    this.transferStallUnsub = onVisibilityStateChange((visible) => {
+      void this.onFirmwareUpdateVisibilityChange(visible);
+    });
```
sidmorizon: 🟠 P1: Stalled transfers remain unguarded

Severity: `severe`

After foregrounding a stalled transfer, `startFirmwareTransferStallGuard` never runs. No caller registers its visibility listener, so the update remains hung.

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+        error: toUserFacingFirmwareUpdateError(
+          toPlainErrorObject(error as any),
+        ),
+      });
+    }
+    try {
+      await this.backgroundApi.serviceHardware.cancel({
+        connectId: this.transferStallConnectId,
+        immediate: true,
+      });
```
sidmorizon: 🟡 P2: Cancellation replaces interrupted error

Severity: `non-severe`

When `cancel` rejects the active transfer, `runUpdateTask` overwrites `firmwareUpdateRetryAtom` with that cancellation. Retry then shows the cancellation instead of transfer interruption.

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
 
+  private transferStallUnsub?: () => void;
+
+  private transferStallAborting = false;
+
+  private transferStallConnectId?: string;
+
+  private backgroundedDuringTransfer = false;
+
+  private startFirmwareTransferStallGuard(connectId: string | undefined) {
```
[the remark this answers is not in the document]
wabicai: 14c6290 已接上：`setFirmwareUpdateWorkflowRunning(true/false)` 会成对 start/stop stall guard，覆盖 V1/V2 开始、失败、退出和 retry wait。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+  private async abortFirmwareTransferAfterBackground() {
+    if (this.transferStallAborting) {
+      return;
+    }
+    this.transferStallAborting = true;
+    const error = new FirmwareUpdateTransferInterruptedError();
+    const taskId = Number(Object.keys(this.updateTasks)[0]);
+    serviceHardwareUtils.hardwareLog(
+      'firmware transfer stalled after returning to foreground',
+      { connectId: this.transferStallConnectId },
```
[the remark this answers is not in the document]
wabicai: 14c6290 已改：stall 日志走 `serviceHardwareUtils.maskLogIdentifier(connectId)`。`cancel()` 仍使用原始 connectId。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,112 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+    );
+    if (Number.isFinite(taskId)) {
+      const stepInfo = await firmwareUpdateStepInfoAtom.get();
+      if (stepInfo.step === EFirmwareUpdateSteps.updateStart) {
+        await firmwareUpdateStepInfoAtom.set({
+          step: EFirmwareUpdateSteps.installing,
+          payload: {},
+        });
+      }
+      await firmwareUpdateRetryAtom.set({
```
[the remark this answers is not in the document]
wabicai: 14c6290 已改：`runUpdateTask` catch 在 `transferStallAborting` 时不再覆盖 `firmwareUpdateRetryAtom`，页面保留 `FirmwareUpdateTransferInterruptedError`。

packages/kit/src/views/FirmwareUpdate/componentsV2/firmwareUpdateInstallViewModel.ts
```diff
@@ -77,6 +81,9 @@ export function getFirmwareUpdateStage({
   if (ENTERING_UPDATE_MODE_TIPS.has(progressType)) {
     return 'enteringUpdateMode';
   }
+  if (TRANSFERRING_TIPS.has(progressType)) {
+    return 'transferring';
```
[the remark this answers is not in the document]
wabicai: 14c6290 已改：无 SDK tip/进度时的 fallback 改为 `installing`，不再合成 `StartTransferData`。真实 `StartTransferData` tip 仍映射到 transferring。

packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts
```diff
@@ -0,0 +1,63 @@
...
+}): boolean {
+  if (!isFirmwareTransferInProgress(after)) {
+    return true;
+  }
+  const beforeProgress = before.firmwareProgress;
+  const afterProgress = after.firmwareProgress;
+  return (
+    typeof afterProgress === 'number' &&
+    typeof beforeProgress === 'number' &&
+    afterProgress > beforeProgress
```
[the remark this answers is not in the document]
wabicai: 14c6290 已改：`didFirmwareTransferResume` 把 `beforeProgress === undefined` 且 `afterProgress` 为 number 视为已恢复，不会误 cancel。

packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts
wabicai: 14c6290 在接线之前先修了 liveness：只把当前 `action === FIRMWARE_PROGRESS` 且 `transferData`、<100% 当作活传输；TIP 上残留的 sticky `transferData` 和 100% 完成态都不算。同时去掉 `hardwareUiStateCompletedAtom` fallback；`beforeProgress === undefined` 视为 resumed。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
wabicai: 14c6290 已加 epoch：start 捕获代数，wait 之后和 abort 入口都会复核；`stop` 会 bump epoch 并重置 `transferStallAborting`。`transferStallResumeCheckPending` 避免 stacked wait；grace 期间再次进后台则跳过 abort。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
wabicai: 14c6290 已注册：`apps/mobile/bundle-registry/module-id-registry.json` 增加 `FirmwareTransferStallGuard.ts` → `14433`。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
wabicai: 14c6290 已接线：工作流 `running=true` 时启动 visibility listener，成功/失败/退出/retry wait 时 stop。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
wabicai: 14c6290 已改：stall abort 先写入中断错误；`runUpdateTask` catch 在 `transferStallAborting` 时不再用 cancel 错误覆盖 retry atom。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,163 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+        error: toUserFacingFirmwareUpdateError(
+          toPlainErrorObject(error as any),
+        ),
+      });
+    }
+    try {
+      await this.backgroundApi.serviceHardware.cancel({
+        connectId: this.transferStallConnectId,
+        immediate: true,
+      });
```
sidmorizon: 🟡 P2: Fast retry cancels its own workflow

Severity: `non-severe`

A fast retry can start while `cancel` remains pending because `firmwareUpdateRetryAtom` is published first. The old task then clears the new workflow after `transferStallAborting` resets.

packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts
```diff
@@ -0,0 +1,82 @@
...
+  }
+  const beforeProgress = before.firmwareProgress;
+  const afterProgress = after.firmwareProgress;
+  if (typeof afterProgress !== 'number') {
+    return false;
+  }
+  if (typeof beforeProgress !== 'number') {
+    return true;
+  }
+  return afterProgress > beforeProgress;
```
sidmorizon: 🟡 P2: Next transfer target looks stalled

Severity: `non-severe`

A target change during the grace period resets `firmwareProgress`, so `didFirmwareTransferResume` reports no progress. The healthy multi-target update is cancelled instead of continuing.

packages/kit/src/views/FirmwareUpdate/componentsV2/useFirmwareUpdateInstallState.ts
```diff
@@ -271,7 +271,7 @@ export function useFirmwareUpdateInstallState({ isDone }: { isDone: boolean }) {
   useEffect(() => {
     if (stepInfo.step === EFirmwareUpdateSteps.installing) {
       if (!lastFirmwareTipMessage && !isNumber(firmwareProgress)) {
-        updateProgressRef.current(EFirmwareUpdateTipMessages.StartTransferData);
+        updateProgressRef.current('installing');
```
sidmorizon: # 🟠 P1 V2 install bar jumps to 50% before the transfer starts, then freezes there for the whole transfer

## Problem
Switching this fallback from `StartTransferData` to `'installing'` also changes which progress *range* the bar seeds from, because the two map to different arms of `getUnifiedProgress`:

```ts
// useFirmwareUpdateInstallState.ts:106-121
case EFirmwareUpdateTipMessages.StartTransferData:
  return calculateProgressInRange({ startAt: 12, maxAt: 50, currentProgress: ctx.firmwareProgress });
case EFirmwareUpdateTipMessages.FirmwareUpdating:
case 'installing':
  return calculateProgressInRange({ startAt: 50, maxAt: 90, currentProgress: resolveFirmwareInstallProgress({...}) });
```

`calculateProgressInRange` is `startAt + (currentProgress ?? 0) * ((maxAt - startAt) / 100)` (`firmwareUpdateProgressUtils.ts:102`). At this point `firmwareProgress` is `undefined` by the guard on line 273, so the result moves from `12 + 0 = 12` to **`50 + 0 = 50`**.

This fires on a normal V2/V3/V4 run: `clearHardwareUiStateBeforeStartUpdateWorkflow` (`ServiceFirmwareUpdate.ts:2948-2955`) writes `payload: {}`, so `lastFirmwareTipMessage` is `undefined`, and no `FIRMWARE_PROGRESS` has arrived yet. The V2 paths set `installingTarget: {} as any` (`ServiceFirmwareUpdate.ts:3399`, `:3517`), so `target?.totalPhase?.length` is falsy and `legacyPhase` is `undefined` (`useFirmwareUpdateInstallState.ts:204-219`) — meaning V2 goes through `getUnifiedProgress`, not the legacy phase path.

Progress is then clamped monotonic:

```ts
// useFirmwareUpdateInstallState.ts:241-245
setProgress(() => { const value = Math.max(next, progressRef.current); ... });
```

Every subsequent `StartTransferData` update can only produce 12…50, so once the bar is at 50 they are all swallowed.

## Impact
On V2/V3/V4 the bar leaps to 50% *before the device has even rebooted to bootloader*, then sits frozen at 50% for the entire firmware transfer — the longest phase, minutes over BLE — and only moves again once `installingFirmware` events arrive. That is a regression against base `6aea968c`, which seeded 12% here. It also shows the "installing" stage word before any bytes move, which is the opposite of what this PR set out to fix. The V1 legacy path is unaffected, since `getLegacyPhaseProgress` maps `'installing'` and `StartTransferData` to the same arm.

## Recommendation
Decouple the stage word from the progress seed. Either keep `StartTransferData` on line 274 and fix the stage mapping in `firmwareUpdateInstallViewModel.ts` instead, or introduce a progress type that yields the 12…50 range while `firmwareProgress` is still `undefined`, or special-case the `'installing'` arm to fall back to `startAt: 12` when `resolveFirmwareInstallProgress` returns `undefined`.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -2812,7 +2982,7 @@ class ServiceFirmwareUpdate extends ServiceBase {
       );
     } finally {
       // Reset workflow running state at service level to prevent lock-screen bypass
-      await firmwareUpdateWorkflowRunningAtom.set(false);
+      await this.setFirmwareUpdateWorkflowRunning({ running: false });
```
sidmorizon: # 🟠 P1 An old workflow's `finally` tears down a newer workflow's stall guard (missing `isUpdateWorkflowCurrent` gate)

## Problem
This line now routes teardown through the new helper, which also stops the stall guard:

```ts
// ServiceFirmwareUpdate.ts:362-375
private async setFirmwareUpdateWorkflowRunning({ running, connectId }) {
  if (running) this.startFirmwareTransferStallGuard(connectId);
  else this.stopFirmwareTransferStallGuard();
  await firmwareUpdateWorkflowRunningAtom.set(running);
}
```

The increment converted six `firmwareUpdateWorkflowRunningAtom.set(false)` call sites to this helper. Three of them got an `isUpdateWorkflowCurrent` gate (`:3008`, `:3176`, `:3223`) — **these two did not**: `:2985` here in `runUpdateWorkflowV2`, and `:2685` in `startUpdateWorkflow` (V1). `workflowId` is in scope at both (the `runUpdateWorkflowV2` parameter, and `:2461` for V1).

The hazard is stated explicitly in this PR's own comment nine lines below:

```
// Unless a newer start has already taken the workflow over — the guard
// is shared, and dropping it here would uncover THAT workflow's page.
```

…and the gate is applied at `:3008` accordingly, but not here.

Sequence: user exits mid-update → restarts → `startUpdateWorkflowV2` bumps `workflowId` and arms guard epoch E2 → the *old* invocation is still unwinding (its `finally` awaits `setFirmwareArtifactDownloadState(false)` and `clearForceTransportType()`, both async) → this ungated `finally` runs → `stopFirmwareTransferStallGuard()` unsubscribes the new listener and bumps the epoch, and the atom is set `false`.

## Impact
The newly started update then runs with **no stall guard at all** — precisely the protection this PR exists to add — and with the lock-screen block dropped, silently, with nothing re-arming it until a retry. The atom clobbering pre-dates this PR, but attaching the guard's lifecycle to these two ungated sites is what converts it into losing the new protection. Note this is the same "old workflow clears newer workflow" family as the retry-race thread on `:476`, but the defect and the repair are here, at the ungated `finally` sites.

## Recommendation
Mirror the sibling call sites at both `:2683-2686` and `:2983-2986`:

```ts
} finally {
  if (this.isUpdateWorkflowCurrent(workflowId)) {
    await this.setFirmwareUpdateWorkflowRunning({ running: false });
  }
}
```

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.test.ts
```diff
@@ -0,0 +1,168 @@
...
+            firmwareTipData: {
+              message: EFirmwareUpdateTipMessages.StartTransferData,
+            },
+          },
+        }),
+      ),
+    ).toBe(false);
+  });
+
+  test('does not treat a sticky transferData value as a live transfer', () => {
```
sidmorizon: # 🟡 P2 The epoch/staleness fix has no test coverage, and a bug of exactly that class is in this same commit

## Problem
The added tests are genuinely good on the sticky-progress side — the cases from line 51 pin `FIRMWARE_PROGRESS + transferData@100 → false`, `FIRMWARE_TIP` with a carried-forward `transferData@100 → false`, and a missing `action → false`. That half of the fix is now regression-locked.

But the concurrency half — `transferStallEpoch` (`ServiceFirmwareUpdate.ts:334`), the post-await re-check (`:420`), the re-check at abort entry (`:443`), the `transferStallAborting` reset in `stopFirmwareTransferStallGuard` (`:358`), and the `transferStallResumeCheckPending` re-entrancy guard (`:412`, `:417`) — lives entirely in `ServiceFirmwareUpdate.ts` and has **no test at all**:

```
$ grep -rln "onFirmwareUpdateVisibilityChange\|onVisibilityStateChange\|transferStallEpoch" \
    packages/kit-bg/src --include=*.test.ts
(no matches)
```

`ServiceFirmwareUpdate.detect.test.ts` contains no stall or visibility assertions either. Every one of those guards could be removed and the suite would stay green.

## Impact
The protection against aborting a live firmware flash on a *later* workflow — a P0-class outcome — is unprotected against regression. This is not hypothetical: the ungated `finally` flagged on `:2985` is exactly this class of cross-workflow lifecycle bug, and it went in alongside the fix without any test noticing.

## Recommendation
`ServiceFirmwareUpdate.detect.test.ts` already mocks `firmwareUpdateWorkflowRunningAtom`, `hardwareUiStateAtom`, `firmwareUpdateRetryAtom` and `timerUtils`, so the fixture cost is low. Add a case that (a) captures the callback passed to `onVisibilityStateChange`, (b) drives background → foreground, (c) calls `setFirmwareUpdateWorkflowRunning({ running: true })` again during the 3s grace window to simulate a newer workflow taking over, and asserts `serviceHardware.cancel` is **not** called. A second case asserting `cancel` **is** called when the epoch is unchanged and progress is frozen would pin both directions.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,163 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+
+  private startFirmwareTransferStallGuard(connectId: string | undefined) {
+    this.stopFirmwareTransferStallGuard();
+    this.transferStallEpoch += 1;
+    const epoch = this.transferStallEpoch;
+    this.transferStallAborting = false;
+    this.transferStallResumeCheckPending = false;
+    this.backgroundedDuringTransfer = false;
+    this.transferStallConnectId = connectId;
+    this.transferStallUnsub = onVisibilityStateChange((visible) => {
```
originalix: P2: [Desktop focus loss is treated as a stalled transfer]

This subscription is also installed for Desktop, where `onVisibilityStateChange` maps Electron's ordinary window `blur` event to `visible = false`. Desktop firmware updates are explicitly USB-only, so changing window focus does not suspend the transfer.

If the window is refocused while no progress event arrives during the fixed three-second grace period, this guard cancels a healthy transfer and shows the retry error. Restrict this cancellation watchdog to native application backgrounding, or use a desktop signal that excludes normal window blur.

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,163 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+    const taskId = Number(Object.keys(this.updateTasks)[0]);
+    serviceHardwareUtils.hardwareLog(
+      'firmware transfer stalled after returning to foreground',
+      {
+        connectId: serviceHardwareUtils.maskLogIdentifier(
+          this.transferStallConnectId,
+        ),
+      },
+    );
+    if (Number.isFinite(taskId)) {
```
sidmorizon: # 🟠 P1 A stall abort with no registered task cancels the device silently and latches `transferStallAborting` forever

## Problem
This `Number.isFinite(taskId)` check concedes that `this.updateTasks` can be empty here — but the abort has already committed to aborting before it checks, and it cancels the device regardless:

```ts
446    this.transferStallAborting = true;              // latched first
447    const error = new FirmwareUpdateTransferInterruptedError();
448    const taskId = Number(Object.keys(this.updateTasks)[0]);   // {} → Number(undefined) → NaN
...
457    if (Number.isFinite(taskId)) {                  // skipped when empty
465      await firmwareUpdateRetryAtom.set({ id: taskId, error: ... });
471    }
472    try {
473      await this.backgroundApi.serviceHardware.cancel({
474        connectId: this.transferStallConnectId,
475        immediate: true,                            // runs unconditionally
476      });
```

With an empty task map the failure is self-reinforcing:

1. `firmwareUpdateRetryAtom` is never set, so `activeRetryInfo` stays undefined and `FirmwareUpdateInstallPageContent.tsx:192-199` keeps `mode = 'updating'` — a frozen bar and "Keep your device connected", **no error text and no Retry button**.
2. `serviceHardware.cancel({ immediate: true })` still fires at the device.
3. `transferStallAborting` is never cleared. Its only resets are lines 344 and 358, both inside `startFirmwareTransferStallGuard` / `stopFirmwareTransferStallGuard`, reachable only via `setFirmwareUpdateWorkflowRunning`. In `runUpdateTask`'s `finally` that call is gated on `retryInfo` existing (`:3181-3184`) — and `retryInfo` is undefined precisely because step 1 skipped writing it.
4. The next task's genuine error is then suppressed by `if (!this.transferStallAborting)` at `:3160`, so the page stays on `mode: 'updating'` indefinitely with a half-updated device.

`updateTasks` is provably empty for long stretches while the workflow is running and the guard is armed — entries are deleted on resolve (`:2365`, `:2374`), and the V1 multi-part path then sits in `waitDeviceRestart` + `waitRebootDelayForNextPhase` (~35-40s) before the next task is created, with `hardwareUiStateAtom` never cleared between phases.

## Impact
A device cancel lands mid-update with no user-visible error and no recovery path, and the latched flag then silences the real error that follows. Note this is the *consequence* branch, distinct from the false-positive entry conditions already raised on `:476` and `FirmwareTransferStallGuard.ts:81` — those ask whether the guard should fire; this is about what happens once it does fire with nothing to report.

## Recommendation
Bail out before committing, so nothing is latched and the device is not cancelled blind:

```ts
const taskKey = Object.keys(this.updateTasks)[0];
const taskId = taskKey === undefined ? Number.NaN : Number(taskKey);
if (!Number.isFinite(taskId)) {
  return; // nothing to fail: do not latch, do not cancel
}
this.transferStallAborting = true;
```

Additionally, clear `transferStallAborting` in `runUpdateTask`'s `finally` regardless of `retryInfo` (`:3175-3188`), so a suppressed error can never persist across phases.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit/src/views/FirmwareUpdate/componentsV2/useFirmwareUpdateInstallState.ts
```diff
@@ -270,9 +270,8 @@ export function useFirmwareUpdateInstallState({ isDone }: { isDone: boolean }) {
 
   useEffect(() => {
     if (stepInfo.step === EFirmwareUpdateSteps.installing) {
-      if (!lastFirmwareTipMessage && !isNumber(firmwareProgress)) {
-        updateProgressRef.current(EFirmwareUpdateTipMessages.StartTransferData);
-      }
+      // Wait for a real tip or numeric progress. A synthetic 'installing'
```
sidmorizon, line 273: # 🟡 P2 Gutting this effect left two now-dead dependencies, re-running `updateProgress('checking')` on every SDK event

## Problem
Removing the synthetic `'installing'` emission also removed the only two reads of `firmwareProgress` and `lastFirmwareTipMessage` from this effect's body, but both remain in the dependency array:

```ts
271  useEffect(() => {
272    if (stepInfo.step === EFirmwareUpdateSteps.installing) {
273      // Wait for a real tip or numeric progress. A synthetic 'installing'
274      // type maps to 50–90 and cannot decrease.
275      return;
276    }
277    if (stepInfo.step !== EFirmwareUpdateSteps.updateStart) return;
280    if (stepInfo.payload.isDownloadingArtifacts) { ...; return; }
286    updateProgressRef.current('checking');
287  }, [firmwareProgress, lastFirmwareTipMessage, stepInfo]);
```

Nothing between 272 and 286 reads either value. The effect therefore still re-fires on every numeric progress tick and every tip change, and while `stepInfo.step === updateStart` it re-issues `updateProgress('checking')`, which unconditionally calls `setStage(...)` → `'preparing'` — competing with the sibling effect at `:257-261` that derives the real stage from `lastFirmwareTipMessage`.

To be precise about severity: React bails out on identical state, so today this is redundant work rather than a visible bug. The dead dependencies are what keep that race reachable at all, and they are a direct artifact of this commit. I could not run the linter here (no `node_modules`), so I am not claiming a CI failure — only that `.oxlintrc.json:230-235` sets `react/exhaustive-deps` to `error`, which makes a stale array worth correcting rather than leaving.

## Impact
Redundant effect runs on every firmware SDK event during the `updateStart` window, and a latent stage-write race that only React's bail-out currently hides.

## Recommendation
Narrow the array to the value the body actually reads:

```ts
}, [stepInfo]);
```

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
```diff
@@ -315,6 +325,163 @@ class ServiceFirmwareUpdate extends ServiceBase {
...
+    if (!running) {
+      return;
+    }
+    const snapshot = await this.readFirmwareTransferUiSnapshot();
+    if (!isFirmwareTransferInProgress(snapshot)) {
+      this.backgroundedDuringTransfer = false;
+      return;
+    }
+    if (!visible) {
+      this.backgroundedDuringTransfer = true;
```
sidmorizon: 🟡 P2: Rapid return misses stall recovery

Severity: `non-severe`

A rapid return can let the visible callback read `backgroundedDuringTransfer` before the suspended hidden callback sets it. No later visibility event starts the grace timer, so the stalled transfer remains hung.

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
sidmorizon: 🟡 P2: Older workflow disables current guard

Severity: `non-severe`

An older background run can call `setFirmwareUpdateWorkflowRunning(false)` after a newer workflow starts. It removes the newer stall listener and clears the shared running atom during the active update.

apps/mobile/bundle-registry/module-id-registry.json
```diff
@@ -24337,5 +24337,7 @@
     "packages/shared/types/unifoldDeposit.ts": 11163,
     "packages/shared/types/wallet.ts": 1105
   },
-  "tombstones": {}
+  "tombstones": {
+    "packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts": 14433
```
sidmorizon: # 🟡 P2 Tombstone reserves an id for a module that never shipped, and booby-traps re-adding the guard

## Problem
After the revert in `8f31bcd6`, this tombstone is the *only* registry change the PR still carries against base `x`:

```json
  "tombstones": {
    "packages/kit-bg/src/services/ServiceFirmwareUpdate/FirmwareTransferStallGuard.ts": 14433
  }
```

But base `98dd334d` has `"tombstones": {}`, no `modules` key for that path, and **no module holding id 14433** at all. The module was allocated id 14433 by the intra-PR commit `14c6290f` and deleted again by `8f31bcd6`; it never existed on `x`. Since the tombstone convention exists to stop a *shipped* id from being reassigned (`moduleIdRegistry.js:338` seeds reserved ids from tombstone values), there is no released bundle here for it to protect.

## Impact
Two costs, both for an id that never shipped:

1. `14433` becomes permanently unallocatable to any other module.
2. Re-adding the stall guard at the same path — plausible, since the PR description and test plan still describe the background-stall recovery as the intended behavior — now hard-fails the tooling:

```js
// apps/mobile/scripts/module-id-registry.js:154-158
if (Object.hasOwn(tombstones, moduleKey)) {
  throw new Error(
    `Tombstoned module reappeared and must be reviewed explicitly: ${moduleKey}.`,
  );
}
```

`collectRegistryErrors` would additionally report `Module key is both active and tombstoned` (`moduleIdRegistry.js:283-285`). Whoever restores the feature has to hand-edit this tombstone out before `module-id:update` will run.

To be clear on severity: the registry validates cleanly today (0 errors from `collectRegistryErrors`, `schemaVersion`/`allocationVersion`/`registryEpoch` unchanged, module count identical to base, no duplicate ids), and `reconcileRegistries` treats a new tombstone key as an ordinary addition, so CI is not failing. This is a latent trap, not a current break — hence P2.

## Recommendation
Restore `"tombstones": {}` so the registry is byte-identical to `x`. Nothing is being removed, because the file never existed on the base branch — the tombstone is recording the deletion of something that only ever existed inside this PR's own history.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

packages/kit/src/views/FirmwareUpdate/componentsV2/useFirmwareUpdateInstallState.ts
wabicai: 已在 d4aa266419 修复：effect 依赖已收窄为 `[stepInfo]`，不会再因 firmware progress/tip 事件重复写入 checking stage。相关测试与 type-aware oxlint 已通过。

apps/mobile/bundle-registry/module-id-registry.json
wabicai: 已在 d4aa266419 修复：移除了仅在本 PR 内新增后删除模块留下的 tombstone，registry 已恢复为 0 tombstones。`yarn lint:module-id` 校验通过。

packages/kit-bg/src/services/ServiceFirmwareUpdate/ServiceFirmwareUpdate.ts
wabicai: Confirmed. The stall-guard implementation was removed from this PR in 8f31bcd6, so this callback race is no longer present in the current head. The reproduced iOS background/foreground hang is an underlying SDK transport issue: the Protocol V2 file-write call can remain pending after resume, which prevents the existing retry/recovery path from running. We will address that separately in a follow-up PR at the SDK timeout/reconnect layer instead of reintroducing app-layer cancellation in this PR.

zy2000-star (dismissed):

packages/shared/src/logger/scopes/hardware/scenes/sdk.ts
```diff
@@ -24,6 +24,12 @@ export function buildHardwareUiEventLogPayload(payload: unknown) {
   return compactLogPayload({
     eventType: event.type,
     deviceType: device?.deviceType,
+    progress: event.progress,
```
sidmorizon, line 27: # 🟡 P2 These transfer metrics never reach a production log, so they cannot diagnose a field stall

## Problem
The six new fields land in `buildHardwareUiEventLogPayload`, whose only production consumer is `uiEvent`:

```ts
// sdk.ts:75-78
@LogToConsole()
public uiEvent(type: string, payload: any) {
  return [type, devOnlyData(buildHardwareUiEventLogPayload(payload))];
}
```

Two independent gates strip this in a release build:

1. **`devOnlyData`** returns a literal placeholder, not the object, whenever `NODE_ENV === 'production'`:
   ```ts
   // packages/shared/src/utils/devModeUtils.ts:3-11
   if (process.env.NODE_ENV !== 'production') return data as unknown;
   return fallback; // '❃❃❃❃ sensitive information ❃❃❃❃'
   ```
   The builder's entire return value is discarded before it reaches the logger.

2. **`@LogToConsole()`** maps to `{ type: 'console' }`. In the single dispatch point `processEntry` (`logger/base/logFn.ts`), the `console` branch (113-123) only calls `console[level]`, and only when `platformEnv.isNative`. It never calls `getLoggerExtension('')` — that is the `local` branch at line 94, which is the persisted path.

So on desktop/web/extension release builds these fields produce **no output at all**, and on iOS/Android release builds they produce the redaction placeholder in an ephemeral console line.

Meanwhile the sibling per-tick log that *is* persisted — `updateHardwareUiStateAtom` (`sdk.ts:97-112`, `@LogToLocal`) via `buildHardwareUiStateLogPayload` — was not extended, so persisted logs still carry no transfer metrics.

## Impact
Given that this PR exists because of a TestFlight stall that had to be diagnosed from device logs, the commit as written adds developer-console-only diagnostics. It cannot contribute to diagnosing a user-reported stall from an exported log bundle or from server events — which is exactly the scenario that motivated it, and the scenario the planned SDK-layer follow-up will need evidence for. The only transfer metrics that survive into production today come from a different, pre-existing path (`defaultLogger.update.firmware.firmwareUpdateResult`, `@LogToServer() @LogToLocal()`), and those are end-of-update aggregates only — never per-tick progress, so they cannot show *where* a transfer stopped.

## Recommendation
Pick one and make it explicit:

- **If field diagnosis is the goal:** emit a throttled, identifier-free metrics record through a persisted scene. `HardwareSDKScene.serviceEvent` (`sdk.ts:119-122`) is already `@LogToLocal({ level: 'info' })` and is *not* wrapped in `devOnlyData`, so it is the natural target. Throttle it — `FIRMWARE_PROGRESS` fires per chunk, so logging every tick would flood the file.
- **If dev-console-only is the actual intent:** that is fine, but say so in the commit message; "log firmware transfer metrics" reads as a production diagnostic and this is not one.

For the record, the fields themselves are safe: `buildHardwareUiEventLogPayload` is a strict allowlist and all six additions are numeric/enum, so nothing device-linkable is exposed by this change.

[Claude session ↗](https://claude.ai/code/session_01WvaZpEB374UmFeFgThkxU2)

---

huhuanming (approved):

ByteZhang1024 (approved):

ezailWang (approved):

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13483>)

packages/kit/src/views/FirmwareUpdate/componentsV2/firmwareUpdateInstallViewModel.ts
```diff
@@ -41,11 +42,14 @@ const ENTERING_UPDATE_MODE_TIPS = new Set<string>([
   EFirmwareUpdateTipMessages.SwitchFirmwareReconnectDevice,
 ]);
 
+const TRANSFERRING_TIPS = new Set<string>([
+  EFirmwareUpdateTipMessages.StartTransferData,
+]);
```
sidmorizon, line 47: 🟡 P2: Completed transfer still shows transferring

Severity: `non-severe`

At 100% transfer progress, `getFirmwareUpdateStage` still returns transferring. [The install hook](packages/kit/src/views/FirmwareUpdate/componentsV2/useFirmwareUpdateInstallState.ts:289-307) maps every `transferData` tick to `StartTransferData`, including 100%.

PR #13484, opened 2026-09-16
fix: harden Infini payment recovery OK-62570

sidmorizon:
[OK-62570](<https://onekeyhq.atlassian.net/browse/OK-62570>)

----------

## Summary

Infini payments need to retain their invoice bindings and user identity checks through transaction acceleration, polling recovery, send failures, and purchase channel changes. This fixes [OK-62570](https://onekeyhq.atlassian.net/browse/OK-62570).

- Persist Infini invoice bindings in transaction history. Before broadcasting a Speed Up transaction, validate the invoice, subscription eligibility, payment recipient and details, original transaction payload, nonce, and expiration. Preserve bindings through transaction replacements and indexer merges.
- Atomically persist observed payment amounts using the payment cache identity so closing the waiting dialog does not discard progress already observed by polling.
- Route Gas Account send failures to payment recovery, where users can keep waiting or explicitly choose to create a new order. Handle the failure exit once so the delayed confirmation-page unmount cannot invoke cancellation and override the recovery state. Preserve ordinary send retries and explicit cancellation.
- Distinguish resuming a pending payment, cancelling its prompt, and continuing a new purchase. Dismissing a pending-payment prompt from the payment-method picker unlocks its rows without starting a payment, and users can select a method again.
- Check the authentication and checkout flow generations immediately before opening an external checkout. Invalidate the previous flow as soon as the waiting dialog starts closing.
- Open a read-only waiting monitor for existing crypto payments on Google Play, continuing polling and subscription confirmation. Preserve Apple Store-only purchase routing on iOS and macOS App Store builds, including pending-payment cancellation and starting a new store purchase after consent.

## Payment recovery behavior

- Acceptance covers new orders created after this fix.
- Allow users to create a new order after a Gas Account failure without automatically rolling back the original order's send claim.
- When the server explicitly reports an expired or failed order and the current payment amounts are zero, allow another purchase. Any late or duplicate payments are handled through the refund process.
- After explicitly accepting the duplicate-payment risk, users can archive an unfinished order and start a new purchase.

## Validation

- `yarn agent:check --profile commit` passed, including lint, formatting, TypeScript, native storage, agent context, and background API contract checks.
- All 951 tests in 35 targeted Jest suites passed after merging the latest `x`, covering SimpleDB, sending and Speed Up, history, Prime service, polling, checkout, purchase entry points, recovery, Dialog, and native store purchases.
- Added regression tests for Gas Account Refresh/Fallback failure exits with both navigation modes, ordinary send retries and cancellation, and dismissing a pending-payment prompt from both card and crypto picker rows. The new regression assertions failed before the fixes and passed afterward.
- Live on-chain payments, hardware signing, and a Google Play channel upgrade on a physical device were not tested.

[OK-62570]: https://onekeyhq.atlassian.net/browse/OK-62570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13484>)

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-9ec7d9cd-5e20-5632-8743-978457062ed6)

sidmorizon:
Claude review session: https://claude.ai/code/session_01PuN3jeSSnfBRR4KogQnHPi

packages/kit/src/views/Prime/components/PrimePurchaseDialog/PrimePurchaseDialog.tsx
```diff
@@ -370,18 +377,107 @@ export function usePrimePurchaseCallback({
...
         await beforeContinue();
+        if (platformEnv.isNativeAndroidGooglePlay) {
+          // Restore only the existing invoice monitor. The full crypto page
+          // also exposes new-payment actions that store builds must not offer.
+          showPrimeInfiniWaitingDialog({
+            context: {
+              checkoutType: 'internalWallet',
+              session: { ...pendingPaymentSession, featureName },
+            },
+          });
```
sidmorizon, line 478: 🟡 P2: Google Play monitors stale invoice

Severity: `non-severe`

When another window replaces the invoice during the prompt, `showPrimeInfiniWaitingDialog` receives the stale session. Recovery then monitors the archived invoice instead of the active payment.

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -1034,7 +1212,7 @@ class ServiceSend extends ServiceBase {
     } = params;
 
     const isMultiTxs = unsignedTxs.length > 1;
-    if (beforeBroadcastAction && isMultiTxs) {
+    if ((beforeBroadcastAction || replaceTxInfo) && isMultiTxs) {
```
sidmorizon: 🟠 P1: Ordinary replacement batches cannot send

Severity: `severe`

A multi-transaction replacement now fails whenever `replaceTxInfo` exists, even without an Infini payment. The batch throws before signing any transaction.

Suggested fix:

```
    if (beforeBroadcastAction && isMultiTxs) {
```

packages/kit-bg/src/services/ServiceSend.ts
```diff
@@ -1034,7 +1212,15 @@ class ServiceSend extends ServiceBase {
...
     const isMultiTxs = unsignedTxs.length > 1;
-    if (beforeBroadcastAction && isMultiTxs) {
+    if (
+      isMultiTxs &&
+      (beforeBroadcastAction ||
+        (await this.getPrimeInfiniReplacementContext({
+          accountId,
+          networkId,
+          replaceTxInfo,
+        })))
```
sidmorizon: 🟠 P1: Replacement batches orphan pending transactions

Severity: `severe`

When `replaceTxInfo` accompanies multiple ordinary transactions, both replace the same history entry. Each [saveSendConfirmHistoryTxs](packages/kit-bg/src/services/ServiceHistory.ts) overwrites its `replacedNextId`, leaving earlier replacements pending and disconnected.

packages/kit/src/views/Prime/components/PrimePurchaseDialog/PrimePurchaseDialog.tsx
sidmorizon: Fixed in [022fb83bf3](https://github.com/OneKeyHQ/app-monorepo/commit/022fb83bf33104c4c4460d80f38036f839c61aa3). The current version of this comment identifies an additional non-Google Play path: the earlier Google Play fix did not refresh the subscription period passed to normal crypto recovery.

After both the prompt and its closing callback finish, [all crypto recovery channels now reload the current session](https://github.com/OneKeyHQ/app-monorepo/blob/022fb83bf33104c4c4460d80f38036f839c61aa3/packages/kit/src/views/Prime/components/PrimePurchaseDialog/PrimePurchaseDialog.tsx#L469). Non-Google Play recovery takes `selectedSubscriptionPeriod` directly from that session; Google Play continues monitoring that same fresh session. A cleared session cancels recovery, and an account change, logout, or failed read cannot open stale recovery or start another purchase. Apple Store routing remains covered by the existing tests.

The payment page already preferred the restored session's plan for its content; this fix removes the stale entry-route period as well. Eight additional regression cases failed before this correction and passed afterward, covering replacement during the prompt/closing callback, refresh success/failure, session removal, account changes, logout, and read failure. All 203 tests in seven related suites and the commit gate (including TypeScript) passed. Resolving the updated finding.

packages/kit-bg/src/services/ServiceSend.ts
sidmorizon: Fixed in [5f0d07c23a](https://github.com/OneKeyHQ/app-monorepo/commit/5f0d07c23a11375bda2b41172d822aaec914757a), and rechecked on `022fb83bf3`.

[The batch guard](https://github.com/OneKeyHQ/app-monorepo/blob/022fb83bf33104c4c4460d80f38036f839c61aa3/packages/kit-bg/src/services/ServiceSend.ts#L1214) now rejects only an explicit Infini before-broadcast action or an actual Infini replacement binding resolved from history/the stored session. `replaceTxInfo` alone no longer triggers the Infini single-transaction error. This restores the pre-PR service admission behavior without dropping validation for implicitly bound Infini replacements.

The four ordinary Speed Up/Cancel cases, in send and sign-only modes, failed with the old guard and now pass. Explicit Infini batches and replacements bound through either history or a matching session still fail before signing. The 112 send, nonce-replacement, and history tests passed again during this recheck. This finding is resolved; these admission tests do not add a new multi-replacement history contract.

packages/kit-bg/src/services/ServiceSend.ts
sidmorizon: Resolved as a false positive for a regression introduced by this PR, with a pre-existing API edge case explicitly noted.

The described overwrite can occur if a direct service caller supplies multiple transactions with one shared `replaceHistoryId`. However, both relevant behaviors already exist on the PR base `2121969c8e`: [ServiceSend passes the same replacement metadata on each iteration](https://github.com/OneKeyHQ/app-monorepo/blob/2121969c8ead0608d780b20b3ebfce46c82efd10/packages/kit-bg/src/services/ServiceSend.ts#L1213), and [ServiceHistory overwrites the predecessor's `replacedNextId`](https://github.com/OneKeyHQ/app-monorepo/blob/2121969c8ead0608d780b20b3ebfce46c82efd10/packages/kit-bg/src/services/ServiceHistory.ts#L3349). This PR does not change that linkage logic.

The production callers that supply `replaceTxInfo` enforce a single transaction:

- [SendReplaceTxContainer](https://github.com/OneKeyHQ/app-monorepo/blob/022fb83bf33104c4c4460d80f38036f839c61aa3/packages/kit/src/views/Send/pages/SendReplaceTx/SendReplaceTxContainer.tsx#L741) builds one unsigned transaction and passes `[unsignedTx]`; `updateUnSignedTxBeforeSending` preserves the input count.
- [TxConfirmActions](https://github.com/OneKeyHQ/app-monorepo/blob/022fb83bf33104c4c4460d80f38036f839c61aa3/packages/kit/src/views/SignatureConfirm/components/SignatureConfirmActions/TxConfirmActions.tsx#L557) only creates replacement metadata when `newUnsignedTxs.length === 1`.
- [SendConfirmActionsContainer](https://github.com/OneKeyHQ/app-monorepo/blob/022fb83bf33104c4c4460d80f38036f839c61aa3/packages/kit/src/views/Send/pages/SendConfirm/SendConfirmActionsContainer.tsx#L179) has the same single-transaction condition. The other batch callers do not pass replacement metadata.

The previous fix restores the service's pre-PR acceptance behavior; it does not introduce a production batch-replacement flow. Generalizing or rejecting direct multi-replacement service calls would be separate API hardening. The hypothetical history-overwrite gap is not claimed to be fixed here, but the reported new user-facing regression is not reachable through the current callers.

originalix:

Review of commit `022fb83bf331` finished, but some comments could not be published automatically.

PR #13486, opened 2026-09-16
feat: refine device portfolio update flow

Shoyo1024:
Jira: [OK-63594](https://onekeyhq.atlassian.net/browse/OK-63594)

## Summary
- Move the device Portfolio update action from the token-list header to the bottom of the wallet More menu.
- Open Device Stage immediately after the action is selected and show a localized success state after the existing update completes.
- Localize the action and success copy across all 19 app locales.

## Scope
- Preserve the existing Portfolio aggregation, eligibility, authorization, and hardware update behavior.
- Keep the Device Stage success copy optional so unrelated hardware flows retain their existing behavior.
- Exclude unrelated Lokalise changes from this PR.

## Test plan
- [x] Verify the More menu closes before Device Stage opens.
- [x] Verify Device Stage starts before the token refresh.
- [x] Verify localized success copy in English and Simplified Chinese.
- [x] Run 4 focused Jest suites: 108 tests passed.
- [x] Run yarn agent:check --profile commit.

[OK-63594]: https://onekeyhq.atlassian.net/browse/OK-63594?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01BskZTyYpTfEB5scCDshwqK

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
```diff
@@ -986,11 +986,15 @@ export class DeviceStageBurstScope {
       });
       return;
     }
-    // The third-party track closes a successful burst with the ✓ done
-    // beat (doc §4.7) before leaving; OneKey bursts leave directly.
+    // An operation-authored ✓ keeps its full reading beat before leaving.
+    const current = await deviceStageAtom.get();
+    if (current?.step === 'done') {
+      this.scheduleOff(DONE_HOLD_MS);
+      return;
```
sidmorizon, line 1016: 🟡 P2: Stale success survives new bursts

Severity: `non-severe`

When a fast burst starts during an earlier `done` hold, `end` adopts that stale outcome. `begin` preserves the visible stage until its deferred opening beat. The previous success remains displayed for another full hold.

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
```diff
@@ -702,6 +702,19 @@ class ServiceHardwareUI extends ServiceBase {
...
+   * result, while the operation's hardware burst still owns the stage. */
+  @backgroundMethod()
+  async deviceStageNoteDone({
+    connectId,
+    doneI18n,
+  }: {
+    connectId?: string;
+    doneI18n?: IDeviceStageState['doneI18n'];
+  }) {
+    await this.deviceStageBurst.noteStep('done', { connectId, doneI18n });
```
sidmorizon: # 🚨 P0: Orphaned `done` beat pins an unclosable Device Stage

## Problem

The new method's own doc comment states the precondition — *"while the operation's hardware burst **still owns the stage**"* — but nothing enforces it.

`noteStep` (`DeviceStageBurst.ts:1477-1525`) has no depth/ownership guard for `done`. It checks `isEnabled()`, a `PROGRESS_WRITABLE_STEPS` ask-protection that covers only `connecting`/`processing` (`:217-220`), and `dismissSeq` **captured at entry** — so a dismissal that already happened is invisible. It then runs `this.clearOffTimer()` and `setStep('done', …)` unconditionally.

The only retirement path this PR gives an operation-authored `done` is the new early return in `end()` (`DeviceStageBurst.ts:989-993`):

```ts
const current = await deviceStageAtom.get();
if (current?.step === 'done') {
  this.scheduleOff(DONE_HOLD_MS);
  return;
}
```

But `end()` is skipped whenever the token went stale — `endExplicit` returns early on mismatch (`:733-741`). Two reachable paths:

1. **User closes the stage during the refresh.** `beginPortfolioSyncStage` now opens *before* `refreshWalletTokenList()` (`TokenListBlock.tsx:2924-2937`), so the stage rests on `connecting` through a pure network fan-out. `connecting` is `rule('stall')` (`deviceStageOwnership.ts:215`) → closable once `DEVICE_STAGE_WAIT_STALL_MS = 10_000` passes with no device activity — and with no hardware call running there *is* no activity, so the close button reliably arms on any refresh over ~10s. `userClose()` sets `depth = 0` and `explicitToken = undefined` (`:1641-1658`). The refresh keeps running and still reaches `deviceStageNoteDone` (`TokenListBlock.tsx:878`, `:2184`), which repaints the stage the user just dismissed. The following `finishPortfolioSyncRequest` → `endExplicit` then mismatches and no-ops.
2. **`TokenListBlock` unmounts mid-refresh.** `useDeviceStageBurst`'s unmount cleanup calls `endBurst()` (`useDeviceStageBurst.ts:100-102`), dropping depth to 0. The in-flight refresh still reaches `deviceStageNoteDone`, which **clears that grace timer** and paints `done`.

Once landed, `done` is `rule('never', false, false)` (`deviceStageOwnership.ts:237`) → `resolveDeviceStageExitGrant` returns `{ closable: false, exitAllowed: false }` (`:289-312`), so `DeviceStageContainer` renders with `onClose={undefined}` and back/Escape are consumed.

Note the in-file asymmetry: `noteStep` arms a hold for `authSuccess` (`:1521`) but gives `done` nothing, and the sibling `noteCardAnswered` has exactly the guard that is missing here (`:1316-1323`):

```ts
if (this.depth <= 0) {
  // Nothing holds the stage: ... the grace takes the stage off rather
  // than leave a processing capsule standing with nothing behind it.
  this.scheduleOff();
}
```

## Impact

A full-screen "Portfolio updated" capsule with no close button, no Escape, no back, and no timer to retire it. The stage's touch wall covers the app, and the only thing that would clear it is another hardware burst — which the wall prevents the user from starting. The app is soft-locked until restart. An orphaned `processing` merely stalls and stays closable; an orphaned `done` is terminal.

This is distinct from the existing `TokenListBlock.tsx:2932` finding (begin-rejects path) — fixing only the begin-failure path leaves both routes above open — and from the `DeviceStageBurst.ts:993` stale-outcome finding.

## Recommendation

- In `DeviceStageBurst.noteStep`, mirror `noteCardAnswered:1316-1323`: after writing a terminal `done`, if `this.depth <= 0` call `this.scheduleOff(DONE_HOLD_MS)` so the ✓ always owns its own exit regardless of who is still holding. Alternatively, have `ServiceHardwareUI.ts:715` skip the write and return `false` when no burst holds the stage, making the documented precondition enforced rather than assumed.
- Separately, in `TokenListBlock.tsx:2924`, consider moving `beginPortfolioSyncStage` to just before `serviceHardwarePortfolioSync.syncPortfolio` (`:846`, `:2174`). Narrating a pure network fan-out as a device `connecting` wait is what arms the stall-close in the first place, and it makes the stalled capsule tell the user to check the cable / wake the device (`stepCopy.ts:633-643`) while nothing is wrong with the device. If the immediate-open UX is required, use a step whose exit rule is not `stall`.

[Claude session ↗](https://claude.ai/code/session_01BskZTyYpTfEB5scCDshwqK)

---

packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx
```diff
@@ -0,0 +1,45 @@
+import { useCallback } from 'react';
```
sidmorizon, line 1: # ⚠️ P1: New module missing from the mobile module-id registry breaks the native build

## Problem

This PR adds `WalletActionPortfolioSync.tsx` and deletes `TokenListBlock/PortfolioSyncButton.tsx`, but never updates `apps/mobile/bundle-registry/module-id-registry.json`. The diff touches **zero** files under `apps/mobile/`:

```
$ git diff --find-renames <merge-base>..ede752e --name-only -- apps/mobile/
(no output)
```

Verified against the registry at this head:

- `WalletActionPortfolioSync` — 0 occurrences.
- The deleted `packages/kit/src/views/Home/components/TokenListBlock/PortfolioSyncButton.tsx` is still registered at line 21500 (id `16547`), now a stale entry.
- Every sibling in the same directory is registered (`WalletActionAddressList.tsx`, `WalletActionApprovals.tsx`, `WalletActionBulkSend.tsx`, … at lines 21512+), so the new file is the odd one out rather than the directory being exempt.

`Native startup graph budget (24.x)` fails hard on this, after both graphs build successfully:

```
Background graph built in 76.8s
Background graph modules: 14648
  packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx.
    at Object.get (.../apps/mobile/plugins/moduleIdRegistry.js:391:13)
    at buildSegmentAllocation (.../apps/mobile/scripts/unionBuild.js:571:62)
```

## Impact

The iOS/Android union build cannot produce a bundle — `buildSegmentAllocation` throws before segment allocation, so no `main`/`background` split is emitted and no allocation or budget report is uploaded. This blocks the PR's only red check and would block any native release cut from this branch. Because the registry pins stable module ids across the split `main`/`bg` runtimes, an unregistered module is a build-stopper by design rather than a warning.

## Recommendation

```
```

This should add an entry for `packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx` and retire the stale `PortfolioSyncButton.tsx` entry at `apps/mobile/bundle-registry/module-id-registry.json:21500`. Regenerate via the tooling rather than hand-editing the JSON, so the allocated id stays consistent with the generator.

[Claude session ↗](https://claude.ai/code/session_01BskZTyYpTfEB5scCDshwqK)

---

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -2955,13 +2924,27 @@ function TokenListBlock({
...
-    refreshWalletTokenList();
+    void beginPortfolioSyncStage({
+      connectId: device?.connectId ?? wallet?.associatedDeviceInfo?.connectId,
+      deviceName: device?.name,
+      deviceType: portfolioSyncDeviceType,
+    })
+      .catch(() => undefined)
+      .then(() => {
+        if (portfolioSyncRequestRef.current?.id === request.id) {
+          refreshWalletTokenList();
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。Portfolio 同步请求现在订阅 `DeviceStageOff`；用户关闭 Stage 时会立即清理当前 request 和 fallback timer。后续刷新即使完成，也无法再通过 request phase claim 进入硬件通信，因此不会发生后台静默同步。

packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx
```diff
@@ -0,0 +1,45 @@
...
+  onClose: () => void;
+}) {
+  const intl = useIntl();
+  const [portfolioSyncUiState] = usePortfolioSyncUiStateAtom();
+  const { disabled, request, visible } = portfolioSyncUiState;
+
+  const handlePress = useCallback(
+    (close: () => void) => {
+      close();
+      request?.();
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。入口现在复用 `runAfterActionListClose()`，等待 ActionList 关闭及 150ms 动画窗口后再发起同步，避免两个浮层的挂载竞态。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -2955,13 +2924,27 @@ function TokenListBlock({
     };
     portfolioSyncRequestRef.current = request;
     setPortfolioSyncRequestPhase(request.phase);
-    refreshWalletTokenList();
+    void beginPortfolioSyncStage({
+      connectId: device?.connectId ?? wallet?.associatedDeviceInfo?.connectId,
+      deviceName: device?.name,
+      deviceType: portfolioSyncDeviceType,
+    })
+      .catch(() => undefined)
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。`deviceStageBeginBurst` 失败后不再继续刷新；当前 request 会被清理并结束，非硬件错误按原有 Toast 路径提示。

packages/kit-bg/src/services/ServiceHardwareUI/ServiceHardwareUI.ts
Shoyo1024: Fixed. Success is now passed through token-owned `deviceStageEndBurst` and landed atomically only after `endExplicit` validates the token. A dismissed or superseded token cannot write `done`, and the separate unowned `deviceStageNoteDone` RPC was removed. Regression coverage includes dismissal before completion.

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -3349,31 +3332,38 @@ function TokenListBlock({
...
+  useEffect(() => {
+    updatePortfolioSyncUiState({
+      disabled: Boolean(
+        !hasPortfolioSyncTarget ||
+        isPortfolioSyncing ||
+        hardwareUiState ||
+        firmwareUpdateWorkflowRunning,
+      ),
+      visible: showPortfolioSyncButton,
+      request: handleSyncPortfolio,
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。More 菜单中的同步入口现在读取活跃 Home tab，仅在 `Portfolio` tab 渲染；切换到 DeFi、NFT 或 History 后，冻结子树发布的旧回调不会再暴露给用户。

packages/components/src/composite/DeviceStage/index.tsx
```diff
@@ -1392,6 +1394,7 @@ export function DeviceStage({
       errorReason,
       localizedErrorMessage,
       waitStalled ? (connectionType ?? 'usb') : undefined,
+      localizedDoneTitle,
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。成功胶囊增加了 `accessibilityLiveRegion="polite"` 和成功标题的 accessibility label，并补充了回归测试。

packages/kit/src/views/Home/components/TokenListBlock/TokenListBlock.tsx
```diff
@@ -914,14 +874,21 @@ function TokenListBlock({
...
+                  },
+                })
+                .catch(() => undefined);
             }
+            finishPortfolioSyncRequest(portfolioSyncRequest.id);
           } catch (error) {
             errorToastUtils.toastIfError(error);
             errorToastUtils.showToastOfError(error);
-            finishPortfolioSyncRequest(portfolioSyncRequest.id);
+            finishPortfolioSyncRequest(portfolioSyncRequest.id, error);
```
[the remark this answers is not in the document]
Shoyo1024: 已修复。硬件错误只交给 Device Stage 呈现；只有非硬件错误继续走 Toast，避免同一错误同时出现两套反馈。

packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx
Shoyo1024: Fixed. The new Wallet action module was added through the official `module-id:update` workflow and now has a stable native registry ID. The native startup graph budget check can resolve this module on the next CI run.

packages/kit-bg/src/services/ServiceHardwareUI/DeviceStageBurst.ts
Shoyo1024: Fixed. Starting a new burst while the previous `done` outcome is visible now replaces it with `connecting` immediately and cancels the old exit timer. A regression test covers the fast restart case.

packages/kit/src/views/Home/components/WalletActions/WalletActionPortfolioSync.tsx
```diff
@@ -0,0 +1,47 @@
...
+  const { disabled, request, visible } = portfolioSyncUiState;
+
+  const handlePress = useCallback(
+    (close: () => void) => {
+      void runAfterActionListClose(close, () => request?.());
+    },
+    [request],
+  );
+
+  if (!visible || activeTabId !== EHomeWalletTab.Portfolio) {
```
sidmorizon: # ⚠️ P1: Tab gate reads a context the More menu cannot see, hiding the action everywhere

## Problem

The new guard added in this commit reads a plain React context:

```ts
const activeTabId = useContext(HomeStickyHeaderContext)?.activeTabId;  // :19
...
if (!visible || activeTabId !== EHomeWalletTab.Portfolio) {            // :30
  return null;
}
```

`HomeStickyHeaderContext.Provider` exists only at `HomePageView.tsx:1262`, wrapping `<Page>{homePage}</Page>`. The More-menu content is not rendered in that subtree. It is handed to `Portal.Render(FULL_WINDOW_OVERLAY_PORTAL, …)`, and that portal is **not** `ReactDOM.createPortal` — which would preserve context — but a root-siblings registry (`hocs/Portal/index.tsx:52-87`):

```ts
manager.ref = portalManagers.get(container);
if (manager.ref) {
  manager.ref.update(id, guest, callback);
}
```

The guest is handed to a `RootSiblingManager` mounted elsewhere, so its React parent is that manager's `<Root/>` inside `PortalContainer` (`:200-241`), mounted at app-provider level in `FullWindowOverlayContainer/index.tsx:56`. Context does not cross that boundary. `activeTabId` is therefore permanently `undefined`, `undefined !== EHomeWalletTab.Portfolio` is always true, and the component returns `null` on **every** tab — Portfolio included.

The portal path is the normal one, not an edge case:

- **Native, always.** `RawActions.tsx:291` `onPress={handleMobilePress}` → `:241` `ActionList.show(...)` → `index.native.tsx:683` → `showActionList` → `:629` `Portal.Render`. Phones are never `$gtSm`, so this is the only path.
- **Web/desktop below `gtMd`.** `ActionListFrame` (`index.tsx:634`) returns `<BasicActionList>` inline only `if (gtMd)`; otherwise `:628 showActionList` → `:594 Portal.Render`.

That both `showActionList` implementations manually re-provide `ModalNavigatorContext` and `PageContext` — capturing them via hooks at the call site and threading them through a `contexts` argument (`index.tsx:616-624`, `index.native.tsx:628-638`) — is the codebase's own acknowledgement that context is severed here. `WalletActionMore.tsx:330-340` follows the same pattern, re-wrapping the items in `AccountSelectorProviderMirror` + `HomeTokenListProviderMirrorWrapper`. `HomeStickyHeaderContext` is given no such treatment, and `WalletActionPortfolioSync.tsx:19` is the only `useContext` call in the whole `WalletActions/` directory — there is no sibling precedent for reading context from inside this menu.

## Impact

"Update device portfolio" never appears in the More menu on iOS and Android — the primary platforms for a hardware-wallet portfolio sync — nor on any narrow web/desktop viewport. Since this PR also deletes the old `PortfolioSyncButton` from the token-list header, the feature has no remaining entry point on those targets. Only wide (`gtMd`) desktop, which renders `BasicActionList` inline, still shows it.

`WalletActionPortfolioSync.test.tsx` cannot catch this: `renderAction` (`:91-103`) hand-mounts `<HomeStickyHeaderContext.Provider value={…}>` around the component in every case. It proves the `if` expression evaluates correctly, not that production ever supplies a value. The new "stays hidden outside the Portfolio tab" case passes for precisely the reason production fails — a non-Portfolio value yields `null`, and so does the real `undefined`.

## Recommendation

Move the tab filter to the producer, which already holds the context in-tree. `TokenListBlock.tsx:3311` already reads `activeHomeTabId`; fold it into the `visible` term of the `updatePortfolioSyncUiState` effect at `TokenListBlock.tsx:3363-3373`:

```ts
visible: showPortfolioSyncButton && activeHomeTabId === EHomeWalletTab.Portfolio,
```

Add `activeHomeTabId` to that effect's dependency array, then revert `WalletActionPortfolioSync.tsx` to the plain `if (!visible)` guard and drop the `useContext`/`HomeStickyHeaderContext`/`EHomeWalletTab` imports. The consumer then needs no context at all, and this still satisfies the stale-closure concern that motivated the guard, since the producer re-publishes on tab change. Please also change the test to render `<WalletActionPortfolioSync/>` **without** a hand-made provider, so the masking cannot come back.

[Claude session ↗](https://claude.ai/code/session_01BskZTyYpTfEB5scCDshwqK)

---

Shoyo1024: Fixed in `e1fc54e6de`. The Portfolio tab gate now runs in `WalletActionMore`, which remains under `HomeStickyHeaderContext`, before the action is passed to the global ActionList renderer. `WalletActionPortfolioSync` no longer reads the context from the detached action tree. I also updated the regression coverage to assert the gate stays in the parent and the action can render without that context.

packages/shared/src/locale/enum/translations.ts
```diff
@@ -3692,6 +3692,7 @@ export enum ETranslations {
   pin_verify_reminder_dialog_verified_toast = 'pin_verify_reminder_dialog_verified_toast',
   placeholder_search_networks = 'placeholder_search_networks',
   portfolio_sync_to_device__action = 'portfolio_sync_to_device__action',
+  portfolio_updated__title = 'portfolio_updated__title',
```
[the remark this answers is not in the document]
Shoyo1024: These are generated outputs from the repository's i18n workflow, not hand-edited translations. The source entries were created/updated in Lokalise and then pulled and regenerated.

The completed workflow receipts verify both keys across all 19 locales. All 38 committed translation values also match the saved pull-verification output. The generated enum and locale files are included so the app can use the new key.

originalix (approved):

PR #13487, opened 2026-09-16
chore: add diagnostic logs for home market view more navigation

huhuanming:
## Problem

QA reports that the **View more** button under the Home market block sometimes does nothing (seen on iOS). Nothing is logged between the press and the Market page, so exported logs cannot show which hop fails.

## Changes

Adds a local-only logger scene, `defaultLogger.market.navigation`, and traces the whole chain with it. Every method is `@LogToLocal` only; `navigation.test.ts` walks the scene prototype and fails if any method routes to the analytics server.

| Step | Where | Logs |
|---|---|---|
| Press | `PopularTrading` (inside the existing `onPress` handler) | `homeViewMore` |
| Navigation request | `useNavigateToMarketTab` | `navigateToMarketTab`, `pendingNavigationCancelled` (`superseded` / `unmount`), `performNavigationStart` (`trigger`: `immediate` / `alreadyApplied` / `selectionApplied` / `timeout`), `performNavigationDispatched`, `performNavigationComplete`, `performNavigationFailed` |
| Discovery landing (native) | `Browser.native.tsx` | `discoveryDefaultTabParam`, `discoveryTabSwitchReceived`, `discoveryTabSwitchApplied` |
| Market landing | `MarketHomeV2`, `useSyncedMarketPerpsCategory`, `MobileLayout.native` | `marketHomeApplySpotCategory` / `marketHomeApplyPerpsCategory` (`applied` / `waitingForConfig` / `unknownCategory`), `marketHomeResetSpotCategory`, `marketHomePagerSync` (`jumped` / `tabNotFound`) |

The Home components only gain one log call inside the existing `onPress` handler; no button props or handlers are added.

Each hook request carries a `navigationId`, so repeated taps and the async steps (atom echo, 500ms fallback, 150ms sub-tab switch) can be correlated.

`useNavigateToMarketTab` behavior is unchanged. Its four `void performNavigation(...)` call sites now go through `startNavigation`, which logs a rejection and rethrows it, so the existing unhandled-rejection reporting still happens.

## Reading the logs

Grep `market => navigation` in the exported logs. The existing `ui => button => click` (from the button `testID`), `app => router => switchTab` / `switchTabDone` and `discovery => browser => exploreTabView` entries fill in around them.

| Pattern | Meaning |
|---|---|
| No `ui => button => click` and no `homeViewMore` | The press never fired: the touch did not reach the button, a parent scroll view took the responder, or it was held long enough to count as a long press (see Notes) |
| `ui => button => click` without `homeViewMore` | The button's deferred `onPress` never ran the handler |
| `pendingNavigationCancelled` with `reason=unmount` | The hook unmounted during the selection wait and the navigation was dropped |
| `performNavigationStart` without `performNavigationDispatched` | `switchTabAsync` hung or threw (`performNavigationFailed`) |
| `performNavigationComplete` with the route still on Home | Navigation was dispatched but did not apply |
| `performNavigationComplete` without `discoveryTabSwitchReceived` | The Discovery listener was not mounted yet |
| `waitingForConfig` with no later result | Market config never finished loading |
| `unknownCategory` or `marketHomeResetSpotCategory` | The Home category id does not exist on the Market page |

## Notes

- **Suspected cause, not changed here:** `Button` always passes an `onLongPress` handler through `useSharedPress`, and Tamagui's native press handling (`mainThreadPressEvents.native.ts`) skips `onPress` when a press with `onLongPress` is held for 500ms. Holding any `Button` for about half a second before releasing therefore does nothing on iOS and Android. The press lifecycle is intentionally not logged, since adding press handlers to business buttons is too invasive, so logs alone cannot tell this apart from a cancelled touch; reproduce with a slow tap to check it.
- Discovery tab switch logs record every header tab switch, not only Market, so a competing switch shows up next to ours.
- Desktop/web inner tab jumps (`useSyncedMarketTab`) are not logged; the category-apply logs still cover those platforms.

## Verification

- `yarn agent:check --profile commit`: all PASS, including full-project `tsc`.
- Jest, 126 tests passing: `navigation.test.ts`; `useNavigateToMarketTab.test.ts`, with new assertions for the atom echo, 500ms fallback, superseded and unmount paths; `PopularTrading.navigation.test.tsx`; `MarketCategoryTokenList.test.tsx`; and the existing Market/Discovery util suites.
- The local-only test was confirmed to catch a regression by temporarily adding `@LogToServer()` to one method; the test failed.
- A throwaway Jest run through the real `defaultLogger` confirmed entries reach the local transport (`HH:MM:SS | WARN : market => navigation => pendingNavigationCancelled : [...]`) and that `$analytics.trackEvent` is never called.
- Not run on a device or simulator.


packages/shared/src/logger/scopes/market/index.ts
```diff
@@ -1,10 +1,13 @@
 import { BaseScope } from '../../base/baseScope';
 import { EScopeName } from '../../types';
 
+import { MarketNavigationScene } from './scenes/navigation';
```
originalix, line 4: P1: [Register the new logger scene for strict mobile bundles]

On iOS/Android production and union bundle builds, loading `MarketScope` follows this new import to `scenes/navigation.ts`. That file has no entry in `apps/mobile/bundle-registry/module-id-registry.json`, and the strict module-ID mapper rejects unregistered workspace modules before the bundle is created.

Generate and commit the registry entry for `packages/shared/src/logger/scopes/market/scenes/navigation.ts` from the mobile build map, then run the strict registry check or native bundle validation.

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01BgKhf7CBdEf28zaJKZqttm

packages/shared/src/logger/scopes/market/index.ts
huhuanming: Fixed in 862aa08743: registered `packages/shared/src/logger/scopes/market/scenes/navigation.ts` as `10628` with `yarn workspace @onekeyhq/mobile module-id:update --map <map>`. It is the only new bundled module in this PR; the other added file is a Jest test.

Verified with a strict iOS union build (`UNION_BUILD=true`, same `unionBuild.js` invocation as the Native startup graph budget job), which exited 0, and `module-id:check` against the fresh build map: `OK; 24315 active modules, 0 tombstones, 1 build map(s) checked`. The fresh map includes `navigation.ts`, so the module is in the native graph and now resolves to its registered ID.

originalix (approved):

PR #13493, opened 2026-09-16
fix: sync Pro2 device name after rename OK-63565

wabicai:
[OK-63565](<https://onekeyhq.atlassian.net/browse/OK-63565>)

----------

## Summary
- After a Pro 2 / Neo device rename succeeds on the hardware, the app now updates the wallet name and cached device label immediately.
- Classic / Touch / Pro 1 are unchanged: they already returned a non-empty Success message and took the write-back path.

## Jira
- https://onekeyhq.atlassian.net/browse/OK-63565

## Context
QA: from Device Management, rename a Pro 2. The device confirms the new name, the app shows “saved”, then the header (and wallet list) still show the old name until a later reconnect / refresh.

Rename goes `WalletRenameButton` → `serviceHardware.setDeviceLabel` → Protocol V2 `DeviceSettingsSet`. The firmware `Success` payload leaves `message` empty (`optional string message = 1 [default = ""]`). The app only wrote the label back when `result.message` was truthy, so Pro 2 / Neo skipped:

1. `writeBackProtocolV2DeviceLabel` (device-state `identity.label`)
2. `handleHardwareLabelChanged` → `setWalletNameAndAvatar` (the name `DeviceBasicInfo` actually renders)

The toast still fired because the SDK call itself succeeded. This change writes the confirmed label back whenever `convertDeviceResponse` returns, instead of gating on a non-empty message.

## Test plan
- [ ] Pro 2: Device Management → pencil on the device name → save a new ASCII name. After the device confirms, the header title updates to the new name without leaving the page. Wallet list / account selector show the same name.
- [ ] Pro 2: rename again to a different name; both the details header and the wallet list follow.
- [ ] Neo: same rename path; app name updates after success.
- [ ] Classic or Touch: rename still works (non-empty Success message path).
- [ ] Rejected / cancelled on-device rename: app keeps the old name and does not toast success.
- [ ] Bluetooth subtitle (if present) is unchanged by this fix; only the primary device / wallet name should move.

## Validation
- `yarn jest packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts --testNamePattern 'label back'`
  - 3 passed: Pro2 with `Success` / empty message, Neo with missing message; write-back uses `result.label`, not the raw input
- Physical-device UI pass is still pending on this change

[OK-63565]: https://onekeyhq.atlassian.net/browse/OK-63565


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3555,24 +3555,25 @@ class ServiceHardware extends ServiceBase {
...
+    if (dbDeviceId) {
+      await this.writeBackProtocolV2DeviceLabel({
+        dbDeviceId,
+        label: p.label,
+      });
+      await this.handleHardwareLabelChanged({
         walletId: p.walletId,
+        dbDeviceId,
+        label: p.label,
+        walletName,
```
sidmorizon, line 3575: 🟡 P2: Protocol V2 label divergence

Severity: `non-severe`

With outer whitespace, `setDeviceLabel` stores the untrimmed input after hardware receives the trimmed label. The wallet and device-state cache then disagree with the hardware.

wabicai: Fixed. `DeviceSettingsManager` already sends the trimmed Protocol V2 label to hardware; write-back now uses that same trimmed value for wallet name and device-state cache.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts
```diff
@@ -2293,6 +2293,131 @@ describe('ServiceHardware SDK DeviceState synchronization', () => {
...
+      walletId: 'hw-wallet-1',
+      label: '  Renamed Pro 2  ',
+    });
+
+    expect(setWalletNameAndAvatar).toHaveBeenCalledWith({
+      walletId: 'hw-wallet-1',
+      name: 'Renamed Pro 2',
+      shouldCheckDuplicate: false,
+    });
+    expect(localDb.updateDeviceState).toHaveBeenCalledWith(
```
sidmorizon: # 🟡 P2: New trim test's device-state assertion is vacuous

## Problem

`localDb` is mocked at module scope with bare `jest.fn()`s (`:85-95`), and `localDb.updateDeviceState` is never cleared inside `describe('ServiceHardware SDK DeviceState synchronization')` (opens at `:1013`). That block declares no `beforeEach`/`afterEach`; the only hooks in the file are `:220-232`, scoped to `describe('ServiceHardware SDK debug logging')` at `:208` — and `jest.restoreAllMocks()` would not reset a manually created `jest.fn()` regardless. No `clearMocks`/`resetMocks` is enabled in the Jest config either.

Call history therefore accumulates across the whole block. The earlier `it.each([EDeviceType.Pro2, EDeviceType.Neo])` test at `:2212-2295` renames to `'Renamed Pro 2'` and already records an `updateDeviceState` call carrying `changedKeys: ['identity.label']`, `source: 'settings-write'`, and `state.identity.label === 'Renamed Pro 2'` (asserted at `:2262-2276`) — exactly the shape this new `expect.objectContaining` matches. Because `toHaveBeenCalledWith` succeeds on *any* recorded call, `:2409-2419` is satisfied by that earlier call no matter what this test's own write-back persisted.

## Impact

The half of this test that guards the device-state cache — the actual `writeBackProtocolV2DeviceLabel` payload — cannot fail. It would pass unchanged against `997c523`, where the untrimmed `'  Renamed Pro 2  '` was persisted into `identity.label`, so the regression this test exists to prevent would go undetected. The `setWalletNameAndAvatar` assertion at `:2404-2408` uses a fresh local `jest.fn()` and is genuine, so only the wallet-name half is really covered.

## Recommendation

Clear the mock immediately before the call under test, mirroring what the sibling test already does for `appEventBus.emit` at `:2255`:

```ts
jest.mocked(localDb.updateDeviceState).mockClear();
await service.setDeviceLabel({
  walletId: 'hw-wallet-1',
  label: '  Renamed Pro 2  ',
});
```

Alternatively switch `:2409` to `toHaveBeenLastCalledWith(...)`. Either change makes the assertion fail against the pre-fix commit, which is the point of the test.

[Claude session ↗](https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3555,24 +3555,26 @@ class ServiceHardware extends ServiceBase {
...
+    const dbDeviceId = wallet?.associatedDevice;
+    if (dbDeviceId) {
+      const label = p.label.trim();
+      await this.writeBackProtocolV2DeviceLabel({
+        dbDeviceId,
+        label,
+      });
+      await this.handleHardwareLabelChanged({
         walletId: p.walletId,
+        dbDeviceId,
```
sidmorizon, line 3573: # 🟡 P2: Trim rule re-derived here instead of flowing from the sender

## Problem

This does not reuse the label that was actually sent to hardware — it independently recomputes it. `DeviceSettingsManager.setDeviceLabel` derives `normalizedLabel` at `DeviceSettingsManager.ts:693-695` and sends exactly that at `:723`, while this site recomputes the same rule at `:3574-3576`. The two agree today only because both re-read the same device record and evaluate the same predicate (`_isProtocolV2Product` at `DeviceSettingsManager.ts:321` delegates to the shared `isProtocolV2ProductType`). Nothing structurally couples them, and neither site references the other.

Relatedly, `:3570` issues a `localDb.getDeviceSafe(dbDeviceId)` purely to read `deviceType`, which `writeBackProtocolV2DeviceLabel` then repeats for the same id at `:856`.

## Impact

Any future change made on the sender side alone — extending trimming to non-V2 devices, collapsing internal whitespace, or truncating at `PROTOCOL_V2_DEVICE_LABEL_MAX_LENGTH` instead of throwing (`DeviceSettingsManager.ts:696-707`) — silently reopens the precise sent-vs-persisted divergence this commit fixes, with no compile-time or test signal, because the duplicated rule lives in a different file. The structural cause of the original defect is still present; only the current values happen to line up. Minor: two DB reads per rename where one suffices.

## Recommendation

Let the normalized label flow from the one place that owns it. Preferred: have `DeviceSettingsManager.setDeviceLabel` return the label it actually sent (e.g. `{ ...result, normalizedLabel }`, applied to both the Trezor and `_withDeviceProcessing` branches so the shape is uniform) and consume `result.normalizedLabel` at `:3579` and `:3584`, dropping the `getDeviceSafe` at `:3570` entirely.

If the return shape must stay untouched, export a single `normalizeHardwareDeviceLabel(deviceType, label)` next to `isProtocolV2ProductType` in `packages/shared/src/utils/hardwareDeviceTypes.ts` and call it from both sites, so the rule has one definition.

[Claude session ↗](https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux)

---

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts
wabicai: Fixed. The test now `mockClear()`s `updateDeviceState` and asserts `toHaveBeenLastCalledWith` against the label returned by `DeviceSettingsManager`, so earlier cases cannot satisfy it.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: Fixed. `DeviceSettingsManager.setDeviceLabel` now returns `{ ...result, label }` for the value it actually sent (trimmed for every device). Write-back uses `result.label` and no longer re-derives trim or re-reads the device.

packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts
```diff
@@ -688,11 +689,12 @@ export class DeviceSettingsManager extends ServiceHardwareManagerBase {
...
-  async setDeviceLabel({ walletId, label }: ISetDeviceLabelParams) {
+  async setDeviceLabel({
+    walletId,
+    label,
+  }: ISetDeviceLabelParams): Promise<ISetDeviceLabelResult> {
     const device = await localDb.getWalletDevice({ walletId });
-    const normalizedLabel = this._isProtocolV2Product(device)
-      ? label.trim()
-      : label;
+    const normalizedLabel = label.trim();
```
sidmorizon: 🟡 P2: Non-V2 device labels lose whitespace

Severity: `non-severe`

For Classic, Touch, and Trezor names with outer spaces, `normalizedLabel` strips characters before sending them to hardware. The [rename UI](packages/kit/src/views/AccountManagerStacks/components/WalletRename/WalletRenameButton.tsx:105-115) trims only Protocol V2 labels. Those devices persist a different label than the user entered.

Suggested fix:

```
    const normalizedLabel = this._isProtocolV2Product(device)
      ? label.trim()
      : label;
```

sidmorizon: # 🔴 P1: Unconditional trim can now send an empty label to non-V2 hardware

## Problem

The trim became unconditional here, but the guard that rejects an empty result stayed gated on `_isProtocolV2Product`:

```ts
697  const normalizedLabel = label.trim();
698  if (
699    this._isProtocolV2Product(device) &&
700    (!isPrintableASCIIString(normalizedLabel) || ...
```

`isPrintableASCIIString` is `Boolean(value) && …` (`stringUtils.ts:76-78`), so on Pro2/Neo a whitespace-only label correctly throws at `:704`. On every other device there is no equivalent check — I traced `_applyTrezorSettings` → `_withTrezorDeviceProcessing` → `adapter.deviceSettings`, and `_withDeviceProcessing` → `convertDeviceResponse(() => sdk.deviceSettings(...))`; neither inspects the label.

A whitespace-only value is reachable from the UI for exactly those devices. `WalletRenameButton.tsx:105` passes `trimOuterWhitespace: isProtocolV2Product`, so for Classic/Touch/Pro1 and Trezor the dialog's own empty check at `HardwareLabelSetDialog.tsx:114` tests the **untrimmed** string — `'   '` is truthy, so it passes, and `required` passes for the same reason. `getHardwareLabelValidationError` also clears it: `isPrintableASCIIString('   ')` is `true` because space is `0x20`, inside `isPrintableASCII`'s `0x20..0x7e` range. `WalletRenameButton.tsx:114` then sends the raw `'   '`, which `:697` now normalizes to `''`.

Before this increment the same input reached hardware as `'   '`; it is the unconditional trim that turns it into `''`.

## Impact

For OneKey Classic/Touch/Pro1 and Trezor, renaming to only spaces now issues `deviceSettings(connectId, { label: '' })` — an empty label is the protocol's unset/clear value, not a no-op, so the on-device name is cleared rather than set. The same `''` is written to `preciseUpdateFields` (`:716`, `:724`) and merged verbatim into the persisted features — `LocalDbBase.ts:5830-5834` spreads `{ ...device.featuresInfo, ...preciseUpdateFields }` with no falsy filtering — so `featuresInfo.label` becomes `''` and `getDeviceLabel` returns empty afterwards.

`result.label === ''` then flows to `ServiceHardware.ts:3569`/`:3574`. The wallet name itself survives only incidentally, via `let walletName = params.name || wallet.name;` (`LocalDbBase.ts:7718`) — a fallback, not a guard on this path.

The net defect is the asymmetry this increment created: V2 devices get an explicit empty-label rejection, non-V2 devices get the empty value forwarded to hardware and into the DB.

## Recommendation

Move the empty check out of the V2-only branch so it covers every device, keeping the ASCII/length guard V2-gated:

```ts
const normalizedLabel = label.trim();
if (!normalizedLabel) {
  throw new OneKeyLocalError(
    appLocale.intl.formatMessage({
      id: ETranslations.global_hardware_label_input_error,
    }),
  );
}
```

Optionally also validate on the trimmed value for all hardware labels in `HardwareLabelSetDialog.tsx:109-136`, so the user sees the inline "empty" error instead of a background throw.

[Claude session ↗](https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux)

---

wabicai: Fixed. Trim is Protocol V2 only again, matching the rename UI. Classic / Touch / Pro 1 / Trezor keep the entered whitespace, and write-back still uses the label that was actually sent.

wabicai: Fixed by restoring Protocol V2-only trim, so a whitespace-only Classic/Touch/Trezor name is no longer collapsed to `''` before `deviceSettings`. Empty-label rejection stays on the V2 ASCII/length guard, same as before this PR.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts
```diff
@@ -2250,7 +2250,7 @@ describe('ServiceHardware SDK DeviceState synchronization', () => {
       });
       service.deviceSettingsManager.setDeviceLabel = jest
         .fn()
-        .mockResolvedValue({ message: 'Success' });
+        .mockResolvedValue({ message: 'Success', label: 'Renamed Pro 2' });
```
sidmorizon: # 🟡 P2: The dropped cases were not redundant — the fix is now uncovered

## Problem

After `52c4159a`, this mock line is the **entire** test contribution of the PR: `git diff x..HEAD` over this file is one changed line. The two tests the PR added for its own fix are gone, and this surviving `it.each([EDeviceType.Pro2, EDeviceType.Neo])` test cannot detect either regression, because both of its values defeat the check.

**`message: 'Success'` defeats the gate.** The pre-PR body at base `98dd334d` is:

```ts
const result = await this.deviceSettingsManager.setDeviceLabel(p);
if (result.message) {
  ...
  await this.writeBackProtocolV2DeviceLabel({ dbDeviceId, label: p.label });
  await this.handleHardwareLabelChanged({ ..., label: p.label, walletName });
}
```

`'Success'` is truthy, so the gate is entered and every assertion at `:2261-2293` passes against the buggy code. This test was added in `a09e0fd9`, *before* this PR — it passed on the gated implementation then and would pass on it again. The empty-`message` case that the whole PR exists to fix is exactly what the deleted `it.each([{ message: '' }, {}])` covered.

**`label: 'Renamed Pro 2'` defeats the label check.** The call at `:2256-2259` passes the same `'Renamed Pro 2'`, so `p.label`, `p.label.trim()` and `result.label` are byte-identical and nothing distinguishes them. The deleted `'writes back the label returned by DeviceSettingsManager, not the raw input'` test was built for this: it passed `'  App Input  '` while mocking the manager to return `'Hardware Label'`.

No other coverage exists. `ServiceHardware.setDeviceLabel` is exercised by this file alone; `DeviceSettingsManager.pro2.test.ts` never constructs `ServiceHardware` and asserts only the SDK call args (`:173-174`) and rejections (`:186-189`), never the returned `label` — so `return { ...result, label: normalizedLabel }` is unasserted too.

## Impact

Reverting either half of the production fix now leaves the whole suite green. The Pro2/Neo stale-name bug and the sent-vs-persisted label guarantee both have zero regression protection, while this test's name ("writes the Pro2 label back to app state after changing the device label") reads as though they are covered. The production code at HEAD is correct, so nothing is broken for users today — the risk is that "redundant" is taken at face value and the guard never comes back.

## Recommendation

Restore the dropped cases. If the intent is genuinely fewer cases rather than less coverage, keep one test and make it discriminating — parameterize over `message` and return a label that differs from the input:

```ts
it.each([
  [EDeviceType.Pro2, 'Success'],
  [EDeviceType.Pro2, ''],
  [EDeviceType.Neo, undefined],
])('writes the %s label back when Success.message is %j', async (deviceType, message) => {
  service.deviceSettingsManager.setDeviceLabel = jest
    .fn()
    .mockResolvedValue({ message, label: 'Hardware Label' });
  await service.setDeviceLabel({ walletId: 'hw-wallet-1', label: '  App Input  ' });
  // assert identity.label === 'Hardware Label' and name === 'Hardware Label'
});
```

That is smaller than the three cases it replaces and still fails on either regression.

[Claude session ↗](https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux)

---

originalix (dismissed):

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts
wabicai: Restored as one parameterized case: Pro2/`Success`, Pro2/`''`, Neo/`undefined`, with `result.label` (`Hardware Label`) different from the raw input (`  App Input  `). Empty message and sent-vs-persisted label both fail against the pre-fix gate.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
```diff
@@ -3555,24 +3555,25 @@ class ServiceHardware extends ServiceBase {
...
+      await this.writeBackProtocolV2DeviceLabel({
+        dbDeviceId,
+        label: result.label,
+      });
+      await this.handleHardwareLabelChanged({
         walletId: p.walletId,
+        dbDeviceId,
+        label: result.label,
+        walletName,
       });
```
sidmorizon, line 3576: 🟡 P2: Concurrent renames restore stale names

Severity: `non-severe`

Concurrent `setDeviceLabel` calls can persist an older label after a newer hardware rename. The [hardware operation](packages/kit-bg/src/services/ServiceHardware/DeviceSettingsManager.ts:470-518) releases its device lease before these write-backs begin. The cached label and wallet name can diverge from the hardware.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts
```diff
@@ -2248,18 +2252,21 @@ describe('ServiceHardware SDK DeviceState synchronization', () => {
           },
         } as unknown as IBackgroundApi,
       });
-      service.deviceSettingsManager.setDeviceLabel = jest
-        .fn()
-        .mockResolvedValue({ message: 'Success' });
+      service.deviceSettingsManager.setDeviceLabel = jest.fn().mockResolvedValue(
```
sidmorizon: # 🟡 P2: This line fails `oxfmt`, which is what red-lights `lint (24.x)`

## Problem

`lint (24.x)` is failing on this head, and the name is misleading — **oxlint passed**. The job log shows TypeScript ✅ (120.67s), `Oxlint check passed` ✅ (197.47s), then:

```
Checking formatting...
packages/kit-bg/src/services/ServiceHardware/ServiceHardware.pro2DeviceManagement.test.ts (5ms)
Format issues found in above 1 files. Run without `--check` to fix.
[2:43:00 PM] Oxfmt check failed. (2.48s)
```

One file, this one. This line is the cause: it is **81 characters**, and `.oxfmtrc.json` sets `"printWidth": 80`. It is the only *code* line in the file over the limit — every other over-width line is a comment, a string literal, or an `it(...)` title, none of which oxfmt reflows.

The previous revision of this exact statement was in the formatter's own output shape, split across the chain (at `52c4159a`):

```ts
service.deviceSettingsManager.setDeviceLabel = jest
  .fn()
  .mockResolvedValue({ message: 'Success', label: 'Renamed Pro 2' });
```

The rewrite collapsed it to `jest.fn().mockResolvedValue(` on one line, which pushes it one character over.

## Impact

`lint (24.x)` is red on `ba88786`, and `CLAUDE.md` states plainly: *"Do not commit code that fails lint or TypeScript checks."* The PR cannot go in on this head. Because the failing check is called "lint" while oxlint itself passes, the cause is easy to misdiagnose as a rule violation and burn a cycle — hence calling it out precisely.

## Recommendation

Restore the split chain so the statement stays within 80 columns:

```ts
      service.deviceSettingsManager.setDeviceLabel = jest
        .fn()
        .mockResolvedValue(
          message === undefined
            ? { label: 'Hardware Label' }
            : { message, label: 'Hardware Label' },
        );
```

`yarn lint:all` (or oxfmt without `--check`) will produce this automatically. Note I could not run oxfmt here to confirm the exact output shape — the diagnosis rests on CI's own report plus the 81-column measurement and the prior formatted form of the same statement.

[Claude session ↗](https://claude.ai/code/session_01UqvEZfxv4kBXk2oP6Ukfux)

---

wabicai: Fixed: ran `yarn oxfmt` on this file. `lint (24.x)` should go green on the new head.

packages/kit-bg/src/services/ServiceHardware/ServiceHardware.ts
wabicai: Out of scope for OK-63565. Device rename is a single modal; the user cannot start a second `setDeviceLabel` until the first confirm/cancel returns. Concurrent write-back would need a separate ticket if we want a device-level queue.

huhuanming (approved):

zy2000-star (approved):

PR #13495, opened 2026-09-16
fix: render the Android Market Stocks chips with the shared JS selector

erikzou:
On Android, the Market Stocks tab now draws its category chips and column header with the shared JS selector instead of the pager's native sub-header, so they match the Favorites and Perps tabs. iOS keeps the native sub-header.

## Why

- The mobile sticky header had two implementations side by side: native chips for Stocks and JS chips for Favorites and Perps. On Android the two looked different.
- The native sub-header only exposes sizes and the selected background. Its font and text colors come from the primary tab bar, and its columns are plain text, so every chip or column style change had to be made twice.
- #13350 enabled the native Stocks sub-header together with the native primary tab bar. Neither that PR nor app-modules#99 names a reason the Stocks chips have to be native.

## Gestures

- The Android pager host decides gesture ownership by what sits under the touch, not by whether it is native. A horizontal drag goes to any horizontal scroller in the sticky header, and a vertical drag goes to the list to collapse or expand the header.
- The `ACTION_UP` fix from app-modules#99 lives in that host, so it covers the JS chips too. The Favorites and Perps JS chips already run in the same region.
- The host moves the JS sticky header by the header offset in the same step as its native headers, so collapsing tracks the finger the same way.

## Behavior changes

- A selected Stocks chip now scrolls into view at the left edge, as the Favorites and Perps chips do, instead of being centered.
- Everything else is unchanged: the categories, their order, the default selection and the column labels are the same.

## Scope

- Gated on `platformEnv.isNativeAndroid`: Android phones and tablets on every release channel.
- iOS phones and iPads keep the native sub-header. Web, desktop and the extension do not load this layout.
- The native primary tab bar is unchanged on both platforms.
- JS only; no native rebuild is needed.

## Test plan

- [x] Jest: Market home layout guards and utils, native lists (5 suites)
- [x] Android emulator: Stocks chips and column header match Favorites and Perps
- [ ] Android: the chip row scrolls horizontally without paging the Market tabs
- [ ] Android: a vertical drag on the chip row collapses and expands the header
- [ ] Android: after a chip drag, chip, banner and filter taps still respond
- [ ] Android: rapid switching between Stocks and other tabs never shows a blank list
- [ ] iOS: Stocks still uses the native sub-header


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01EiyStg4YnxqaPaUh5ybUxv

originalix (approved):

PR #13438, opened 2026-09-15
fix: use system font for native prev close labels (OK-63333)

limichange:
[OK-63333](<https://onekeyhq.atlassian.net/browse/OK-63333>)

----------

## Summary

Fix the previous-close label rendering on iOS and Android by using the chart's existing system typeface at the price-axis font size. Measure the label with that same font so its background and text stay aligned. Price values and Web retain their existing font rendering.

No additional font fallback logic is introduced.

Fixes OK-63333.

## Validation

- 34 tests passed across the Skia renderer, reference-line scene, and Web renderer suites.
- `yarn agent:check --profile pr` passed all local checks, including lint, formatting, TypeScript, and repository contracts.
- iOS and Android visual verification is still pending.

[OK-63333]: https://onekeyhq.atlassian.net/browse/OK-63333?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13497, opened 2026-09-16
fix: ref the typeface returned by Skia getTypeface (OK-63558)

Minnzen:
[OK-63558](<https://onekeyhq.atlassian.net/browse/OK-63558>)

----------

## Summary

- Patch `JsiSkFont::getTypeface` in `@shopify/react-native-skia` 2.6.2 to use `refTypeface()` instead of wrapping the raw `getTypeface()` pointer in `sk_sp<SkTypeface>()`.
- Fixes an iOS `EXC_BAD_ACCESS` (SIGSEGV, wild pointer) that reproduces reliably when switching coins/symbols on any screen that renders the native Skia TradingView chart.

## Intent & Context

QA reported a reliably reproducible iOS crash on Perps when switching coins (build `6.6.0+2026091622`, iPhone 16 Pro, iOS 26.6.1, device locale `zh_CN`). A second device reproduced the same crash three times in four minutes on Swap Pro market detail (build `6.6.0+2026091625`), showing the problem is not Perps-specific.

The crash was diagnosed from the exported Sentry native payload, symbolicated against the UUID-matched binary from Sentry debug files (`9056AD40-BCEA-3854-9288-EAA03A4631EE`) plus a UUID-matched local `hermesvm`:

```
find_by_CTFontRef(SkTypeface*, void*) + 20                       <-- crash
SkTypefaceCache::findByProcAndRef(...) const + 60
SkTypeface_Mac::Make(...) + 212
SkFontMgr_Mac::onMatchFamilyStyleCharacter(...) const + 552
RNSkia::JsiSkFontMgr::matchFamilyStyleCharacter(jsi::Runtime&, ...) + 632
  [hermes interpreter]
facebook::react::Task::execute(...) -> RuntimeScheduler_Modern::runEventLoop(...)
+[RCTJSThreadManager runRunLoop] + 212                            <-- RN JS thread
```

## Root Cause

`JsiSkFont::getTypeface` built its return value as:

```cpp
sk_sp<SkTypeface>(getObject()->getTypeface())
```

`SkFont::getTypeface()` returns a raw pointer and explicitly does not alter the refcount (`SkFont.h:211`), while `explicit sk_sp(T* obj)` adopts the pointer — "No call to ref() or unref() will be made" (`SkRefCnt.h:247-250`). Every JS call therefore handed the `JsiSkTypeface` wrapper an ownership it never took. When that wrapper was collected, `~sk_sp` released a reference nobody had acquired, so the typeface was over-released by one per call.

Skia's process-global typeface cache (a function-local static in `SkTypeface_Mac::Make`, guarded by its own lock) still held a pointer to the freed `SkTypeface_Mac`. The next font lookup walked the cache and `find_by_CTFontRef` dispatched a virtual call through the dead object.

Disassembly of the crashing binary confirms this precisely:

```asm
0x101efa4ec   ldr  x8, [x0]           ; load cached->vtable      (did NOT fault)
0x101efa4f0   ldr  x8, [x8, #0x108]   ; load vtable slot         <-- FAULT
0x101efa4f4   blr  x8
```

The first load succeeded, so `cached` was a valid mapped address whose contents were garbage — the signature of a freed-and-reused object, not of a corrupted cache array. `SkTypeface_Mac::Make` was also verified to serialize `findByProcAndRef` and `add` behind `SkSemaphore::osWait`/`osSignal`, ruling out a concurrent-mutation race.

The only caller of `getTypeface()` in the whole repository is `chartSkiaRenderer.ts:303` (`referenceLineLabel` font), introduced by #13438 (OK-63333) at 2026-09-16 10:57 — the same day as the first crashing build, which matches the sudden onset. Devices in `zh_CN` fail fastest because the legend typeface is a CJK fallback returned by `matchFamilyStyleCharacter` and has fewer live owners than the system font.

## Design Decisions

- **Fix the binding, not the call site.** A JS-side change (dropping the `getTypeface()` call in `chartSkiaRenderer.ts`) would also stop the crash today, but leaves the defect armed for the next caller — and `font.getTypeface()` reads as completely harmless. Since this release has not shipped yet, there is no need for a JS-only hotfix that could ride an OTA bundle, so the native fix alone is the right scope.
- **`refTypeface()` matches the upstream fix.** Upstream react-native-skia already corrected this exact line, so this patch converges with upstream rather than inventing a OneKey-specific workaround.
- **No mutex added.** An earlier hypothesis blamed concurrent `SkFontMgr` access from the JS thread and the Reanimated UI runtime. That was refuted: Skia already serializes the cache internally (verified in the shipped binary), so any locking added here would be redundant and would not address the refcount error.

## Changes Detail

- `patches/@shopify+react-native-skia+2.6.2.patch` — adds one hunk for `cpp/api/JsiSkFont.h`, switching `getTypeface` to `getObject()->refTypeface()` with a comment explaining why the `sk_sp(T*)` form was wrong. Existing hunks are unchanged.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS primarily; Android shares the binding), Desktop/Web unaffected (different Skia path)
- **Risk Areas**:
  - Requires a native rebuild to take effect; a stale native shell will still crash.
  - Single line, single call site in-repo, and it only makes refcounting correct (no leak: the wrapper now holds a real reference and releases it on collection).
  - The patch was verified to apply cleanly to a pristine `@shopify/react-native-skia@2.6.2` tarball.

## Test plan

- [ ] Native rebuild picks up the patch (`patches/**` is in the trigger paths of `mobile-dev-shell-ios-simulator`, so the simulator dev shell rebuilds on merge to `x`).
- [ ] iOS, device language Simplified Chinese, Perps with the chart laid out at the top: switch coins 20+ times in a row — no crash. (Before this fix it crashes within a few switches; with the chart collapsed at the bottom it does not reproduce, because the collapsed panel does not follow coin switches.)
- [ ] iOS, Swap Pro market detail: loop token selector -> market detail 10+ times — no crash. This was the fastest reproduction (crash within 36 seconds).
- [ ] Market detail (MarketDetailV2) and asset detail chart: same component, spot-check for regressions.
- [ ] English locale: chart legend and reference-line labels render unchanged.

## Notes

- `yarn agent:check` currently reports two failures that are pre-existing on `x` and unrelated to this change (both reproduce on a clean `x` checkout): `tsc-staged` (`isScrollDragging` missing — a local patch-application issue) and `agent-context` (implicit skills 23 exceeds budget 22).
- This change has not yet been verified on a device; it needs a CI build first.

[OK-63558]: https://onekeyhq.atlassian.net/browse/OK-63558?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01Djqd4dQxUSUNNMw26VfMzR

originalix (approved):

PR #13498, opened 2026-09-16
fix: keep a single market detail in the stack ｜OK-63514 OK-63537

zy2000-star:
[OK-63514](<https://onekeyhq.atlassian.net/browse/OK-63514>) [OK-63537](<https://onekeyhq.atlassian.net/browse/OK-63537>)

----------

## Summary
- [OK-63514](https://onekeyhq.atlassian.net/browse/OK-63514): native Market detail header token switch pushed another `MarketDetailV2` / `MarketStockDetail` page. Top-left back then walked through the old token before reaching the list.
- [OK-63537](https://onekeyhq.atlassian.net/browse/OK-63537): Wallet Home Popular Trading used nested `navigate`, so opening another token or stock from Home stacked leftover details the same way.
- Both bugs share one owner: Market/Discovery kept multiple detail routes, and the custom back handler only `pop()`ped one screen.
- Native token switch also closed a modal first. After that, `getRootState()` often lost nested tab state, so the existing in-place update missed and fell back to a root `navigate`.
- Native empty-history back reset to `TabMarket`, which is not the native Market list host (`TabDiscovery`).

## Changes
- Collapse the Market/Discovery stack to `[list, one detail]` when opening or switching a detail (`setParams` for the same route, `reset` when details are already stacked or the route changes).
- After the token selector closes, if nested stack state is missing, update the focused detail with `setParams` / `replace` instead of pushing.
- Wire the same collapse into Home Popular Trading so a new Home entry does not leave an extra detail under the current one.
- If the previous route is still a leftover detail, back uses `popToTop` instead of a single `pop`.
- Keep Search / SwapPro / tablet split-view back behavior.
- Native empty history now resets to Discovery, not `TabMarket`.

[OK-63514]: https://onekeyhq.atlassian.net/browse/OK-63514?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63537]: https://onekeyhq.atlassian.net/browse/OK-63537?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_01D3ihLx9GWP3drKnWSfskaZ

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
```diff
@@ -0,0 +1,288 @@
...
+  const currentIndex = index ?? routes.length - 1;
+  const previousRoute = currentIndex > 0 ? routes[currentIndex - 1] : undefined;
+  if (
+    previousRoute?.name === ETabMarketRoutes.TabMarket ||
+    previousRoute?.name === ETabDiscoveryRoutes.TabDiscovery
+  ) {
+    return { type: 'pop' };
+  }
+
+  return { type: 'popToTop' };
```
sidmorizon: 🟡 P2: Banner detail back skips parent

Severity: `non-severe`

After opening a banner token, `resolveMarketDetailBackAction` returns `popToTop` instead of one pop. [MarketBannerDetailContent](packages/kit/src/views/Market/MarketBannerDetail/MarketBannerDetail.tsx:186-208) creates this valid stack, so Back jumps directly to the market list.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketDetailBackNavigation.ts
```diff
@@ -24,61 +22,35 @@ export function useMarketDetailBackNavigation() {
...
-        navigation.pop();
-        return;
-      }
-    }
-
-    navigation.pop();
+    reactNavigation.dispatch(
+      CommonActions.reset({
+        index: 0,
+        routes: [{ name: action.name }],
```
originalix: P2: [Native empty-history back opens Browser instead of Market]

When an iOS or Android detail is opened with no prior stack entry, this reset recreates `TabDiscovery` without `defaultTab`. `Browser.native.tsx` therefore uses the saved Discovery tab or its Browser default, so Back lands outside the Market list instead of returning to it.

Carry `defaultTab: ETranslations.global_market` through the native reset (and cover the one-route native stack) so this fallback preserves the Market destination.

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
```diff
@@ -0,0 +1,288 @@
...
+  ETabDiscoveryRoutes,
+  ETabMarketRoutes,
+} from '@onekeyhq/shared/src/routes';
+
+const MARKET_DETAIL_ROUTE_NAMES = new Set<string>([
+  ETabMarketRoutes.MarketDetail,
+  ETabMarketRoutes.MarketDetailV2,
+  ETabMarketRoutes.MarketStockDetail,
+  ETabMarketRoutes.MarketNativeDetail,
+  ETabMarketRoutes.MarketBannerDetail,
```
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
```diff
@@ -0,0 +1,288 @@
...
+export function openOrReplaceMarketDetailRoute({
+  routeName,
+  params,
+}: {
+  routeName: string;
+  params: Record<string, unknown>;
+}): boolean {
+  const navigation = rootNavigationRef.current as INavigationLike | undefined;
+  const stack = findMarketTabStack(navigation?.getRootState?.());
+  if (!navigation || !stack?.key || !stack.routes?.length) {
```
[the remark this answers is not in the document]
zy2000-star: fixed

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
zy2000-star: fixed

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketDetailBackNavigation.ts
zy2000-star: fixed

packages/kit/src/views/Market/utils/marketDetailNavigation.ts
```diff
@@ -0,0 +1,416 @@
...
+  const navigation = rootNavigationRef.current as INavigationLike | undefined;
+  const current = navigation?.getCurrentRoute?.();
+  if (!navigation || !isFocusedMarketDetailRouteName(current?.name)) {
+    return false;
+  }
+
+  const nextParams = buildReplacedMarketDetailParams(params);
+  if (
+    current?.name === routeName ||
+    current?.name === EModalSwapRoutes.SwapProMarketDetail
```
sidmorizon: > # 🔴 P1: SwapPro modal loses `disableTrade` on token switch

## Problem

This new branch routes `SwapProMarketDetail` into the `setParams` path, and the params it writes come from `buildReplacedMarketDetailParams` (`marketDetailNavigation.ts:96-104`), which deliberately emits **every** key of `DETAIL_ROUTE_PARAM_KEYS` — writing an explicit `undefined` for any key absent from the input:

```ts
next[key] = key in params ? params[key] : undefined;
```

`DETAIL_ROUTE_PARAM_KEYS` (`:26-44`) contains `'from'`, `'disableTrade'` and `'showFavoriteButton'` — exactly the three params the Swap Pro modal owns and always sets (`SwapMainLand.tsx:690`, `SwapHeaderRightActionContainer.tsx:588,637` all push `{ from: EEnterWay.SwapPro, disableTrade: true, showFavoriteButton: false }`).

The caller never re-supplies them: `navigateToMarketTokenDetail.ts:111-140` builds `tokenParams` with no `from` and no `disableTrade`. Since `SET_PARAMS` merges over `route.params`, one in-modal token switch clobbers both to `undefined`.

This path is new in this commit. Previously `openOrReplaceMarketDetailRoute` returned `true` here so `replaceFocusedMarketDetailRoute` was never reached, and its old guard excluded `SwapProMarketDetail` anyway.

## Impact

1. **Trading re-enabled in a modal opened with trading disabled.** `MarketDetailV2.tsx:203` reads `params.disableTrade`; it feeds the `shouldDisableTrade` OR-chain at `:282`, which now goes falsy, so the Buy/Sell panel and trade buttons become active.
2. **Back button breaks.** With `from` lost, `useMarketDetailBackNavigation` passes `from: undefined`, so `resolveMarketDetailBackAction` skips the `from === SwapPro → pop` short-circuit (`:373`). The modal stack has `routes.length === 1`, so it falls through to `:404-407` and dispatches `CommonActions.reset({ routes: [{ name: 'TabDiscovery' | 'TabMarket' }] })` onto the **SwapModal** stack, which registers neither screen.

## Recommendation

`SET_PARAMS` merges, so simply don't emit the modal-owned keys on this branch:

```ts
const isSwapProModal = current?.name === EModalSwapRoutes.SwapProMarketDetail;
const nextParams = buildReplacedMarketDetailParams(params);
if (isSwapProModal) {
  delete nextParams.from;
  delete nextParams.disableTrade;
  delete nextParams.showFavoriteButton;
}
```

Two tests currently encode the clobbering as expected and would not catch this: `marketDetailNavigation.test.ts:445-454` asserts `params: buildReplacedMarketDetailParams(...)` verbatim for `source: 'swap-detail'`, and `navigateToMarketTokenDetail.test.ts:604-609` checks only `type: 'SET_PARAMS'` without inspecting the payload.

[Claude session ↗](https://claude.ai/code/session_01D3ihLx9GWP3drKnWSfskaZ)

---

zy2000-star: fixed

originalix (approved):

PR #13499, opened 2026-09-16
fix: upgrade native modules to 3.0.141 for pager-view fonts

huhuanming:
## Summary

- Upgrade all `@onekeyfe` native modules from 3.0.140 to 3.0.141: `apps/mobile`, `packages/components`, `packages/kit`, and the root `react-native-get-random-values` dependency and resolution. `yarn.lock` and `apps/mobile/ios/Podfile.lock` are updated.
- Remove `patches/react-native-pager-view+3.0.140.patch`, added in #13476 for OK-63273. 3.0.141 ships the same fix inside the package (OneKeyHQ/app-modules#113), so the patch is no longer needed.
- No JS change. Market keeps passing `fontFamily: 'Roobert-Medium'` to the native tab bar, and no new props are involved.

## Root cause

- The pager-view native Market headers resolved `fontFamily` with `Typeface.create(fontFamily, NORMAL)`: the tab bar buttons, the category items and the column labels. On Android that only finds system font families, so the bundled `Roobert-Medium` silently fell back to the system font.
- iOS was never affected. `[UIFont fontWithName:size:]` finds Roobert because the fonts are in `UIAppFonts`, bundled as resources, and their PostScript names match `Roobert-Regular` / `-Medium` / `-SemiBold` / `-Bold`.
- 3.0.141 resolves the typeface with `ReactFontManager.getInstance().getTypeface(fontFamily, NORMAL, context.assets)`. That is the lookup React Native `Text` uses, and react-native-screens, react-native-svg and lottie-react-native make the same call for native text. It checks registered custom fonts, then `assets/fonts/<family>.ttf|otf`, and falls back to `Typeface.create` only when neither exists.

## Test plan

- **Emulator verification:** API 35 emulator on `fix/ok-63557-native-modules-3.0.140` (native modules 3.0.140, this patch removed, the 3.0.141 pager-view source placed in `node_modules`). I built once with the old `Typeface.create` file and once with the fix, then compared the native Market tab labels against CoreText's ink width for Roobert-Medium at 42px (16sp × 2.625):

  | Label | Roobert-Medium | Before | After |
  |---|---|---|---|
  | Favorites | 168.6 | 164 | 169 |
  | Trending | 164.6 | 159 | 164 |
  | Stocks | 126.8 | 124 | 126 |
  | Robinhood | 204.8 | 195 | 204 |

[...]

PR #13501, opened 2026-09-16
chore: forbid native code patches in @onekeyfe packages

huhuanming:
## Summary
- Add `yarn lint:onekeyfe-patches`. It fails when a patch-package patch changes anything other than `.js`/`.jsx`/`.ts`/`.tsx` files in an `@onekeyfe/*` package, including npm aliases such as `react-native-pager-view` → `@onekeyfe/react-native-pager-view`.
- The failure message tells developers to open a PR in https://github.com/OneKeyHQ/app-modules and publish a new release instead.
- Run the lint in `yarn lint` and `yarn agent:check --profile commit`, run its tests in the lint workflow, and document the rule for patch-package.

## Intent & Context
`@onekeyfe/*` React Native modules are built and released from OneKeyHQ/app-modules, so native changes to them should land there and ship as a new version. They have repeatedly been patched in place through patch-package instead:

- `@onekeyfe/react-native-tab-view`: Swift
- `@onekeyfe/react-native-background-thread`: Kotlin, C++, Objective-C++
- `@onekeyfe/react-native-split-bundle-loader`: Objective-C++
- `@onekeyfe/react-native-auto-size-input`: Kotlin, Swift, Gradle, Nitro bindings
- `react-native-pager-view` (an alias of `@onekeyfe/react-native-pager-view`): Kotlin. #13499 upstreamed this fix through OneKeyHQ/app-modules#113 and removed the patch.

The request was to block patches to Objective-C, Swift, Java and Kotlin code in `@onekeyfe/*` packages, keep TS/JS/TSX/JSX patches allowed, and send developers to app-modules (PR, then release) for native updates.

## Design Decisions
- **Allowlist, not a native extension denylist.** Only `.js`, `.jsx`, `.ts` and `.tsx` files may change (`.mjs`, `.cjs`, `.mts`, `.cts` and `.d.ts` count as JS/TS). Earlier `@onekeyfe` patches also changed `.cpp`/`.hpp`, `build.gradle` and Nitro-generated `.json`/`.gitattributes`, which a list of Objective-C/Swift/Java/Kotlin extensions would miss. As a side effect, other non-JS files such as `.json` or `.md` are rejected too.
- **Aliases are resolved through `yarn.lock`.** Patch file names use the install name (`react-native-pager-view+3.0.140.patch`), so the lint maps each install name to its lockfile resolutions. `bitcoinjs-lib` resolves to both `@onekeyfe/bitcoinjs-lib@7.0.1` and the third-party `bitcoinjs-lib@6.1.7`, so the version in the patch file name picks the matching entry. If no entry has that version, the patch counts as an `@onekeyfe` patch when any resolution is `@onekeyfe`.
- **Patched files come from the `diff --git` headers.** This covers sequenced (`+001+initial`), dev-only (`.dev.patch`) and nested (`a+1.0.0++b+2.0.0`) patch names, and both sides of a rename are checked.
- **No exemption list.** The only existing violation, `react-native-pager-view+3.0.140.patch`, was removed by #13499, which this branch is based on.
- **`CLAUDE.md` is unchanged** because it is at 4998 of the 5000 bytes allowed by `lint:agent-context`. The rule is documented in the `1k-patch-package-workflow` skill, which `CLAUDE.md` points to for patches, and in `patches/README.md`.

## Changes Detail
- `development/lint/onekeyfe-patches.js`: the lint. It uses only Node built-ins and takes about 70 ms on this repo.
- `development/lint/onekeyfe-patches.node-test.js`: `node:test` cases for allowed JS/TS files, rejected native and build files, aliases, version matching, third-party packages, nested/sequenced/dev-only patch names, and renames.
- `package.json`: adds the `lint:onekeyfe-patches` script and runs it from `lint:project`.
- `.github/workflows/lint.yml`: runs the new tests in the "Test developer scripts" step.
- `development/scripts/agent-check.js`: runs the lint as a local check.
- `.skillshare/skills/1k-patch-package-workflow/SKILL.md`, `patches/README.md`: document the rule and the native debugging flow. Developers may edit native code in `node_modules` to debug and rebuild with `dev-shell --shell local`, ship the change through app-modules, and delete `apps/mobile/out-dir-bundle/dev-shell/local-cache/<platform>` if they undo the edit without upgrading.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: None at runtime. This only changes lint, CI and developer tooling.
- **Risk Areas**: A false positive blocks `yarn lint`. By design, any non-JS file in an `@onekeyfe` patch fails. A third-party package with the same name and version as an `@onekeyfe` alias would also be flagged.

## Test plan
- [x] `yarn agent:check --profile commit` passes, including the new `onekeyfe-patches` check (`passed (72 patches)`).
- [x] The documented debugging flow matches the launcher. Editing a Kotlin function body (Android) or a Swift function body (iOS) in `node_modules/@onekeyfe/react-native-tab-view` leaves `shellInputKey` unchanged. `yarn.lock` is a shell input and no `node_modules` path is. `--shell local` builds are saved to `local-cache` under that key, so `--shell auto` would reuse them.
- [x] The `concurrently` part of `lint:project` passes with the new `yarn lint:onekeyfe-patches` entry. The full `yarn lint` (with `_tsc` and `_lint`) was not run locally; `agent:check` covered `tsc-staged` and oxlint/oxfmt on the changed files.
- [x] `node --test development/lint/onekeyfe-patches.node-test.js`: 8/8 pass. Breaking version matching, alias resolution, the allowlist, or rename handling makes at least one test fail.
- [x] The "Test developer scripts" command from `lint.yml`: 39/39 pass.
- [x] On the base before #13499 (`05db7b5b8c`), the lint fails on `react-native-pager-view+3.0.140.patch` (`CollapsiblePagerNativeHeaders.kt`) and points to app-modules.
- [x] Replayed against `patches/` and `yarn.lock` from earlier commits: it flags the tab-view Swift patch, the background-thread sequenced patches (`.cpp`/`.kt`/`.h`/`.mm`), split-bundle-loader (`.h`/`.mm`), and 35 native and build files in auto-size-input while allowing that patch's `.ts`/`.d.ts` files. It does not flag the third-party `react-native-pager-view+7.0.1.patch` from before the package became an alias.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_013puzLFFYLKS5FMYCaVJs3F

development/lint/onekeyfe-patches.js
```diff
@@ -0,0 +1,213 @@
...
+    `[onekeyfe-patches] Native code (Objective-C, Swift, Java, Kotlin, C/C++) and build files ship only from ${APP_MODULES_REPO_URL} releases. Once released, upgrade the dependency here and remove these changes from the patch.`,
+  ].join('\n');
+}
+
+function main() {
+  const rootDir = path.resolve(__dirname, '../..');
+  const patchesDir = path.join(rootDir, PATCHES_DIR);
+  const patches = fs
+    .readdirSync(patchesDir)
+    .filter((fileName) => fileName.endsWith('.patch'))
```
sidmorizon, line 185: 🟡 P2: Desktop runtime patches escape enforcement

Severity: `non-severe`

When a native `@onekeyfe/*` desktop runtime patch is added, `main` never scans it. The [desktop postinstall](apps/desktop/app/package.json:9) applies that separate patch directory, so the forbidden patch passes lint.

.skillshare/skills/1k-patch-package-workflow/SKILL.md
```diff
@@ -8,6 +8,16 @@ allowed-tools: Read, Grep, Glob, Bash, Edit, Write
 
 Patches live in `patches/<pkg>+<version>.patch` and auto-apply on install. Never hand-edit a `.patch` — edit the `node_modules/` source, then generate.
 
+## `@onekeyfe/*` packages
+
+In `@onekeyfe/*` packages, including npm aliases such as `react-native-pager-view` → `@onekeyfe/react-native-pager-view`, a committed patch may only change `.js`/`.jsx`/`.ts`/`.tsx` files. `yarn lint:onekeyfe-patches` enforces this.
+
+For native code (Objective-C, Swift, Java, Kotlin, C/C++) and native build files:
+
+- **Debug:** edit them directly in `node_modules/`, then rebuild with `yarn workspace @onekeyhq/mobile dev-shell --platform <android|ios> --shell local`. The default `--shell auto` reuses a cached shell because the shell key ignores edits inside `node_modules`.
```
sidmorizon, line 17: # 🟡 P2: "the shell key ignores edits inside `node_modules`" is contradicted by the repo's own test

## Problem

`SKILL.md:17` states the default `--shell auto` "reuses a cached shell because the shell key ignores edits inside `node_modules`", and `patches/README.md:13` repeats it as "does not notice edits inside `node_modules`". The cache hit requires `receipt.shellInputKey === compatibility.shellInputKey` (`apps/mobile/scripts/local-dev-shell-cache.js:71`). `shellInputKey` is derived from `{nativeBuildDigest, nativeContractKey, platform}` (`apps/mobile/scripts/native-dev-shell.js:622-628`), and `nativeContractKey = computeNativeContractKey(...)` (`:616-617`) hashes each native package's **ABI surface read out of `node_modules`** — `getNativePackageAbiDescriptor` (`apps/mobile/plugins/devVendor.js:867-898`) `readFileSync`s bridge files, codegen/Nitro specs and package manifests under `getInstalledPackageRoot` (`:419-432`, `require.resolve('<pkg>/package.json')`).

The repo asserts exactly the opposite of this line. `apps/mobile/plugins/__tests__/devVendor.test.js:679` is named *"derives native shell keys from installed ABI contracts only"*, and `:825-839` edits a file **inside `node_modules`** and requires the key to change:

```js
const codegenSpecPath = path.join(fixture.repoRoot,
  'node_modules/@react-native-community/netinfo/src/internal/NativeRNCNetInfo.ts');
fs.writeFileSync(codegenSpecPath, codegenSpec.replace('getCurrentState', 'getCurrentStateChanged'));
expect(computeNativeContractKey('android', fixture.repoRoot)).not.toBe(androidBaseline);
expect(computeNativeContractKey('ios',     fixture.repoRoot)).not.toBe(iosBaseline);
```

The claim holds only for edits that leave the ABI surface untouched (e.g. a log line inside an existing method body); the same suite shows the sensitivity is ABI-scoped rather than content-scoped, since bumping a package's `version` leaves the key unchanged (`:812-823`).

## Impact

The bullet describes debugging native code, which is precisely when a developer renames or adds an exported bridge method (`RCT_EXPORT_METHOD`, `RCT_EXTERN_METHOD`, `@ReactMethod`, expo `Function`/`AsyncFunction`, JSI `setProperty` — `devVendor.js:714-780`) or edits a `.nitro.ts`/codegen spec. For those edits `shellInputKey` changes, so `--shell auto` misses the cache; and because `getShellCompatibility` sets `requireExactInput: true` (`native-dev-shell.js:636`) the remote restore cannot match either, so it drops into the local-build fallback (`:2701-2710`) — a full CocoaPods/Xcode or Gradle rebuild, the opposite of the "reuses a cached shell" the doc promises. `nativeContractDependencies` in `apps/mobile/dev-vendor.config.js` covers the 40+ `@onekeyfe/react-native-*` packages and their aliases this section is about, so this is the common case, not a corner. A developer who trusts the stated mechanism reads that unexpected full rebuild as a broken toolchain. The same wrong mechanism is given as the reason for the "Undo without upgrading" step at `SKILL.md:19` ("cached under the unchanged key"); the recommended action there stays correct, but its justification only applies to body-only edits.

## Recommendation

Qualify the claim in both files so it matches `computeNativeContractKey`. At `.skillshare/skills/1k-patch-package-workflow/SKILL.md:17` and `patches/README.md:13`, say the shell key hashes each native package's ABI surface (exported bridge methods, codegen/Nitro specs, `package.json`) and not its implementation — so a body-only edit in `node_modules` is invisible to `--shell auto` and needs `--shell local`, while an edit that changes an exported native signature invalidates the key and forces a rebuild on its own. Since `SKILL.md:19` leans on the same mechanism, note there that the stale-cache risk is the body-only case.

[Claude session ↗](https://claude.ai/code/session_013puzLFFYLKS5FMYCaVJs3F)

---

huhuanming: 这个测试对本 PR 没有意义，PR 只需要说明 body-only native debug 修改的场景；ABI surface 变更属于另一条行为路径，不在本 PR 的验证目标内，也不影响 lint 规则。保留现有文档，resolve。

Minnzen (approved):

originalix (approved):

PR #13509, opened 2026-09-17
fix: cancel android press when dragging out of the button(OK-63573)

huhuanming:
[OK-63573](<https://onekeyhq.atlassian.net/browse/OK-63573>)

----------

## Summary
- On Android, pressing a Tamagui pressable (`Button`, `IconButton`, or a `Stack`/`XStack`/`YStack` with `onPress`), dragging the finger away and releasing still fired `onPress`. The press style also stayed on for the whole drag.
- Patch `@tamagui/web`'s native `useMainThreadPressEvents` so that on Android the press is cancelled once the touch leaves the element bounds plus `hitSlop` and React Native's default press retention offsets (20dp left/right/top, 30dp bottom), matching RN `Pressable`. Moving back inside re-activates the press.
- iOS keeps its current behavior for now (see below).

## Intent & Context
- QA report (防误触): on the mobile More page, pressing Preferences, About OneKey or the hardware wallet card, dragging far away and releasing still opened the page.
- This is a regression from the Tamagui 1.108 → 2.7.7 upgrade (1287a7147d). Tamagui 1.108 used React Native's Pressability on native, which cancels presses on drag-out.
- Upstream history:
  - tamagui/tamagui@654aa7c8ae (v2.0.0-rc.31) replaced `usePressability` with a hand-written responder implementation. Its `onResponderMove` only forwards `onPressMove`, so the press rect check was lost.
  - tamagui/tamagui@25561c4c20 (v2.4.1, fixing tamagui/tamagui#4024) routed every real press handler through that path. It also removed the RNGH press gesture, which had cancelled presses after 12pt of finger travel. That gesture only ran when the app called `setupGestureHandler`, which OneKey never does, so OneKey never had that behavior.
  - The latest release (2.7.7) and `main` still have this behavior. I found no upstream issue or discussion reporting it.
- Scroll views don't prevent it. Android `ScrollView` doesn't intercept touches when its content fits on screen (the More page is a `ScrollView`), and it only intercepts drags along its scroll axis.
- **Why Android only.** The report is from Android. iOS reports this less often, most likely because vertical `ScrollView`s bounce by default, so a vertical drag usually starts a scroll and cancels the press. The iOS path also hasn't been verified on a simulator or device. iOS still has the bug for horizontal drags, fixed header/footer buttons and dialogs. Enabling iOS later means removing one platform check.

## Design Decisions
- **Same model as RN Pressability.** At grant, measure the responder with `currentTarget.measure`. On each move, compare `pageX`/`pageY` against that region plus `hitSlop` and the retention offsets.
  - Leaving fires `onPressOut` immediately and clears any pending long press.
  - Releasing outside skips `onPress`.
  - Re-entering fires `onPressIn` again, as `Pressable` and Tamagui 1.x do.
- **Android gate.** `measureResponderRegion` returns early unless `Platform.OS === 'android'`. Without a region, the press runs exactly the previous code path. `Platform` is read at call time, not at module load: Tamagui's own code notes that `Platform` isn't initialized yet when the compiler loads the native bundle for static extraction.
- **Only trust a region that contains the start touch.** Native sheets (`@onekeyfe/react-native-native-sheet`, used by `Dialog`, `ActionList` and `Popover` with `nativeSheet`) lay their content out at `left: -100000` and dispatch touches relative to the sheet root. On device, the dialog Cancel button measured `left: -99979` while the touch was at `x: 110`. A plain Pressability port would cancel almost every real finger tap there, so when the start touch is outside the measured region the press keeps the previous behavior.
- **No usable measurement means no change.** If `measure` is unavailable or returns zeros, the press behaves as before.
- **Press out fires immediately on leave** and ignores `delayPressOut`. OneKey doesn't use `delayPressOut`, and Tamagui sets `minPressDuration: 0` on native.
- Web, desktop and the extension are unaffected: the web build of `useMainThreadPressEvents` is a no-op stub.

## Changes Detail
- `patches/@tamagui+web+2.7.7.patch`: adds hunks for `dist/cjs`, `dist/esm` and `src` `helpers/mainThreadPressEvents.native`. The same ~85 lines of logic appear once per build, and Metro loads the cjs build. Existing hunks are unchanged. The patch was regenerated with patch-package, applies cleanly to a fresh `@tamagui/web@2.7.7`, and reproduces `node_modules` byte for byte.

## Risk Assessment
- **Risk Level**: Low. It changes Tamagui presses on Android only, and only for touches that move.
- **Affected Platforms**: Android. iOS, web, desktop and the extension keep their current behavior.
- **Risk Areas**:
  - Where `measure()` and touch coordinates disagree by less than the retention offsets, the press can cancel slightly early or late. RN `Pressable` has the same limitation.
  - Components that run actions in `onPressIn` (for example `SecureQRToast`) get `onPressIn` again if the finger leaves and comes back while they are still mounted. Tamagui 1.x behaved the same way.
  - Inside native sheets, dragging out still fires `onPress` because of the coordinate mismatch above. Fixing that needs a follow-up in the sheet module.

## Test plan
- [x] Strict `tsc` on the patched source, `node --check` on both JS builds, and `yarn lint:onekeyfe-patches`
- [x] Node harness that drives the compiled cjs and esm builds with synthetic responder events. It covers 27 cases: tap, jitter, drag out, drag out and back, retention edges, `hitSlop`, long press, terminate, `delayPressIn`, missing or zero `measure`, external press ownership, composed user responder handlers, and a sheet-like coordinate offset.
  - `Platform.OS = 'android'`: drag-out no longer fires `onPress`; every other case behaves as designed.
  - `Platform.OS = 'ios'`: all 27 cases match the unpatched build exactly.
- [x] Android emulator (API 36, DevSession), More page, on the final Android-only build:
  - Dragging from Preferences and releasing far away: no navigation
  - A plain tap and a jittery tap (small moves): Preferences opens
- [x] Android emulator, earlier build with the same Android logic:
  - Dragging from About OneKey or the hardware wallet card: no navigation
  - Dragging out and back in: the page opens
  - A pressed row loses its highlight as soon as the finger leaves
  - Native sheet (About OneKey → Export state logs): a jittery tap on Cancel closed the sheet in 3 of 3 logged runs, and the logs show the mismatched region being rejected. One earlier run without logging did not close the sheet; the cause wasn't found, and on that path the patched code behaves exactly like the old code.
- [ ] iOS simulator or device verification before enabling iOS

[OK-63573]: https://onekeyhq.atlassian.net/browse/OK-63573?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

patches/@tamagui+web+2.7.7.patch
```diff
@@ -375,6 +728,210 @@ index c91b74f..f93aa0b 100644
...
++    const touch = getTouchFromPressEvent(e)
++    if (typeof touch?.pageX !== 'number' || typeof touch?.pageY !== 'number') return
++    const isOutside = !isTouchWithinPressRect(touch, region, hitSlop)
++    if (isOutside === ref.current.isOutsidePressRect) return
++    ref.current.isOutsidePressRect = isOutside
++    const isActive = ref.current.state === 'active' || ref.current.state === 'longPressed'
++    if (isOutside) {
++      if (ref.current.longPressTimer) clearTimeout(ref.current.longPressTimer)
++      ref.current.longPressTimer = null
++      if (isActive) events.onPressOut?.(e)
```
originalix, line 873: P2: [Drag-out bypasses the existing press-out delay]

When an active Android touch leaves the press rectangle, this invokes `onPressOut` immediately instead of using the existing `deactivate()` path. That bypasses `minPressDuration` and any configured press-out delay, so a quick drag-out changes the established pressed-state timing for every native control using this handler.

Route the leave transition through `deactivate(e)` (and preserve its timer/state semantics) so drag-out uses the same press-out contract as release and termination.

huhuanming: Thanks. The leave path does skip `deactivate()`, but it doesn't change pressed-state timing in practice, and routing through `deactivate()` as suggested introduces a stuck press style.

- Tamagui passes `minPressDuration: 0` to every native component without `asChild` (`createComponent.native.js`, the `events && !asChild` block), and OneKey never sets `delayPressOut`. For those controls `deactivate()` computes `remaining = 0` and calls `onPressOut` synchronously, exactly like the current code.
- Only `asChild` pressables fall back to the hook's 130ms default (a few triggers such as `Popover.Trigger`). There, a drag-out within 130ms of press-in clears the highlight up to 130ms earlier. That's intentional: leaving the rect isn't a release, so the highlight shouldn't linger after the finger has left.
- I tested the suggested change with the patch harness at `minPressDuration: 130`:
  - Quick drag-out, then release outside: `cleanup()` on release clears the pending timer, so `onPressOut` never fires and the press style stays on.
  - Quick drag-out, back in, hold: `onPressIn` fires twice, and the stale `onPressOut` fires while the finger is back inside.
  - With `minPressDuration: 0`, both versions emit the same events.

Supporting the delay correctly would need extra timer bookkeeping for a ≤130ms visual difference on a handful of `asChild` triggers, so I'd keep press-out immediate on leave.

huhuanming: Resolving, per the analysis above: press-out stays immediate when the touch leaves the press rect. For native Tamagui controls `minPressDuration` is 0 and `delayPressOut` is unused, so `deactivate()` would emit the same events. Routing through it without extra timer handling leaves `asChild` pressables with a stuck press style. Full alignment with RN Pressability's press-out delay is out of scope for this PR.

originalix (approved):

PR #13515, opened 2026-09-17
fix: keep native chart realtime subscription on 15m

limichange:
TradingView Native Market charts can stop synchronizing the chart and page-header prices when the selected interval receives no WebSocket candles. Keep a single `15m` subscription across all chart interval changes, without unsubscribing or resubscribing when the user changes the displayed interval.

History still loads at the selected interval. The `15m` chart uses complete candle updates; other intervals use live closes as price ticks, preserve historical candle values, and create new candles at their own boundaries. Subscription recovery and cleanup use the same `15m` lease.

Validation:
- 133 provider and K-line state-machine tests passed, including subscription reuse, price updates, candle boundaries, and stale-message rejection.
- `yarn agent:check --profile commit` and `yarn agent:check --profile pr` local checks passed.
- Web verification on the reported Solana token: `1m → 15m → 4H → 1m` kept chart/header prices synchronized. Captured WebSocket traffic showed one chart subscription and no unsubscribe/resubscribe during the switches.

Limitations:
- Non-15m live candles reflect observed price ticks. Volume stays at the historical value, or zero for newly created candles, until history is refreshed; a 15m feed cannot reconstruct exact OHLCV for another interval.
- The reported EVM token emitted no 15m candles during the latest verification. Subscription reuse was confirmed, but live price synchronization could not be verified for that token.


sidmorizon:
@codex review

sidmorizon:
@codex security review

zhaono1 (approved):

originalix (approved):

PR #13521, opened 2026-09-17
fix: stop mobile bundles from inlining isJest as true OK-63584

huhuanming:
[OK-63584](<https://onekeyhq.atlassian.net/browse/OK-63584>)

----------

Fixes [OK-63584](https://onekeyhq.atlassian.net/browse/OK-63584).

> [!IMPORTANT]
> Needs a security review before merge. This turns on native crypto paths for iOS/Android that have been off in shipped mobile builds (details below).

On iOS and Android, errors show their raw default message (usually the class name) instead of translated text. For example, creating DOT addresses with a software wallet that doesn't have 12 or 24 words shows `Expect24WordsMnemonicError`, while desktop shows "Only wallets with 12/24-word recovery phrases can create addresses for this network".

**Root cause:** `platformEnv.isJest` is replaced at build time by `transform-define`, using `buildTimeEnv.isJest` (`JEST_WORKER_ID !== undefined || NODE_ENV === 'test'`). The mobile Metro config merges Sentry's Expo config, which turns on `transformer.unstable_workerThreads`. jest-worker's `threadChild` then sets `JEST_WORKER_ID` in every transform thread. As a result, every Metro bundle compiled `platformEnv.isJest` to `true`: release and DevSession builds, main and background runtimes alike. `normalizeErrorProps` therefore skipped i18n. The runtime `platformEnv.isJest` property stays `false` on device, which is why checking it at runtime looked correct. rspack (web/desktop) evaluates the flag in its main process and was unaffected.

**Fix:**
- `packages/shared/src/buildTimeEnv.js`: `isJest = process.env.NODE_ENV === 'test'` (Jest sets it itself). The `JEST_WORKER_ID` check is dropped because Metro's worker threads set it. This one flag feeds the `platformEnv.isJest` define, Babel plugin selection and the runtime `platformEnv` object, so all of them follow.
- Update two comments that described the miscompile as current behavior. The earlier `errorUtils.ts` workaround is fully reverted.

**What changes on mobile (dev and prod), now matching the intended desktop behavior:**
- **Crypto paths:** `canUseQuickCryptoEC` (`publicFromPrivate` only; `sign`/`verify` stay on JS elliptic); ADA quick-crypto root keypair; TON mnemonic validation and key pair via WebEmbed; `useWebembedApi` routes (only Developer Gallery code passes `true` today).
- **Guards now enforced:** `ensureSensitiveTextEncoded(password)` in HD derivation; the simpleDb, Realm and WebEmbed main-thread guards; the firmware artifact background-only check; test-only reset helpers.
- **Behavior:** identity recovery starts `pending` until bootstrap marks it ready; market detail preload; error i18n (this bug).
- **Babel plugins:** production mobile bundles strip `console.*` (file logs use `NativeLogger` and are unaffected); `@sentry/babel-plugin-component-annotate` is enabled for mobile.

Validation:
- Compiled a probe through the real mobile Metro config: `platformEnv.isJest` is now `false` in worker-thread, child-process and in-band modes (it was `true` with worker threads).
- Reproduced `unionBuild.js`'s production Metro setup and compiled real modules. The results were: `normalizeErrorProps` guard `&& true`, `canUseQuickCryptoEC` `true && true`, simpleDb guard active, `console.log` stripped, Sentry annotation present, and the bundled runtime `isJest` folds to `false`.
- Android emulator (API 35, DevSession, zh-CN), no debugger attached to the main runtime:
  - Manage accounts → Polkadot AssetHub shows 只有 12/24 个助记词的钱包才能为此网络创建地址.
  - EVM addresses for an 18-word wallet: all 10 are identical to a screenshot taken before the fix (JS elliptic vs native quick-crypto EC).
  - Imported the public test mnemonic `test … junk`. EVM accounts 0–2 match the Hardhat defaults (`0xf39Fd6…b92266`, `0x709979…dc79C8`, `0x3C44Cd…4293BC`). Cardano accounts 0–3 match the `AdaAddressPerfTest` fixed vectors.
  - BTC (Taproot), SOL and TON addresses derive without errors. Unlock, wallet import and password verification work.
  - logcat shows none of the newly enabled guard errors.
- Jest: 25 suites and 359 tests passed. These include the suites that `jest.resetModules()` and set `NODE_ENV=production` (`perpsColdStartPerf`, `sentry/index.native`, `sesHarden`, `mobileLayoutTrace`), the isJest-gated test helpers (`ServiceIdentityExit`, identity lifecycle mutex), storage/IndexedDB, errors, the availability aggregator and dev-vendor.
- The emulator pass ran on the previous revision, which moved the same flag in `development/`. The compiled output of both revisions was checked identical with the probes above.

Not verified: iOS; a production release build on device; TON-native mnemonic import (WebEmbed); firmware update and OneKey ID recovery flows beyond startup; bundle size and performance impact of Sentry component annotation.

Notes:
- The dev-vendor and web-embed prebundle fingerprints change, so DevSessions build them locally (with `ONEKEY_USER_NOTICE`) until CI publishes them for `x`.
- Android DevSession serves WebEmbed from `http://10.0.2.2:8081`, which the `$private` origin allow-list rejects, so WebEmbed features (e.g. Cardano) hang there regardless of this change. Using `adb reverse tcp:8081 tcp:8081` with `--metro-url http://localhost:8081` works.

[OK-63584]: https://onekeyhq.atlassian.net/browse/OK-63584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63584]: https://onekeyhq.atlassian.net/browse/OK-63584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (dismissed):

wabicai (dismissed):

limichange (approved):

sidmorizon (approved):

PR #13523, opened 2026-09-17
fix: switch market tabs without the pager animation OK-63487

huhuanming:
[OK-63487](<https://onekeyhq.atlassian.net/browse/OK-63487>)

----------

## Summary

- Tapping a Market tab on iOS and Android now jumps straight to that page instead of sliding through every page in between. Swipes keep their animation.
- Upgrades the OneKey native modules to 3.0.144, which adds `nativeTabPressAnimationEnabled` to `CollapsiblePagerView` ([app-modules#114](https://github.com/OneKeyHQ/app-modules/pull/114)). The Market pager turns it off.

## Intent & Context

OK-63487: switching Market tabs on mobile flickered. The report compared a OneKey screen recording with Binance and OKX and asked for a natural transition with no elements flying across the screen. In those recordings, Binance swaps the page in one frame, and OKX slides straight to the target page without passing other pages.

## Root Cause

- #13272 moved Market home to the native `CollapsiblePagerView`, and tab presses became `setPagerIndex(index, true)`, an animated `setPage`. For native tab bar presses, the pager wrapper also queued its own animated `setPage`.
- The iOS direct pager animates its content offset over 0.28 s and Android's `ViewPager2` smooth-scrolls, so a distant tab scrolled through every page in between.
- With `pageRetentionDistance={1}`, pages more than one position away are unmounted. The slide showed blank pages and neighboring lists passing under a sticky header that had already switched to the tapped tab.
- #12842 (OK-55394) fixed the same problem for the previous tab container with `tabPressAnimationEnabled={false}`. The migration dropped it.

## Design Decisions

- Every tap jumps directly, as Binance does, instead of sliding for adjacent tabs. The filter chips and column header live in the pager's shared sticky header, not in the pages, and every adjacent tab pair has a different filter row. In the old build, an adjacent tap (Top coins → Perps) switched the chips at +0 ms, moved the old list at +39 ms, started the slide at +182 ms and settled at +320 ms, so the new chips sat over the old list for about 0.3 s.
- Delaying the sticky header update until the page command was tried and reverted. The target list rendered with the old header height and sat under the chips for 50–120 ms.
- The wrapper's own animated command is turned off with the new prop. The earlier app-side workaround sent a second, non-animated command and had to track the native page index to also cover a repeated press before the jump landed.

## Changes Detail

- `MobileLayout.native.tsx`: `handleTabPress` calls `setPagerIndex(index, false)`, and the pager gets `nativeTabPressAnimationEnabled={false}`.
- `MobileLayout.native.test.tsx` (new): a native tab press dispatches only `setPageWithoutAnimation`, and the pager receives `nativeTabPressAnimationEnabled={false}`. Each test fails when its change is reverted.
- Native modules 3.0.142 → 3.0.144: `package.json` pins, `yarn.lock` and `Podfile.lock` (pod versions and spec checksums). No native code changed between these versions. In app-modules, only `react-native-pager-view/src/CollapsiblePagerView.tsx`, its README, a test, the version files and the changelog changed.

## Risk Assessment

- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS, Android). Desktop, web and extension use other layout files.
- **Risk Areas**: Tab press → page selection sync on the native pager. Swipes are unchanged, and `syncCurrentPage` and the focus sync already used non-animated jumps.
- **Known and unchanged here**:
  - When the tapped tab's sticky header has a different height, the header updates about one frame before the page (up to ~60 ms on the Android emulator), because the pager sends the page command on the frame after React commits.
  - A distant page remounts on tap. The Stocks list can first show SWR-cached rows and replace them with the fresh response about 0.1 s later. The old slide ended the same way.

## Test plan

- [x] `yarn jest packages/kit/src/views/Market/MarketHomeV2/layouts packages/kit/src/views/Market/MarketHomeV2/components/MarketNativeList` (8 suites, 97 tests)
- [x] iOS simulator (iPhone 17 Pro, iOS 26.5), recorded with `simctl io recordVideo` before and after: Favorites → Robinhood, Perps → Stocks, Trending → Top coins, Perps → Trending, rapid double taps on distant tabs, and a swipe from Trending to Stocks. After: the tab bar, filter header and list change within one frame, with no intermediate page or blank frame, and the swipe still animates across about 13 frames.
- [x] Android emulator (Android 15), recorded with `adb shell screenrecord`: Trending → Top coins, Top coins → Perps, Perps → Trending, double taps on Favorites and Top coins, and Top coins → Stocks switch directly.
- [x] iOS simulator on the 3.0.144 build (local shell and vendor build from this branch's lockfile): a double tap from Favorites to Robinhood, Perps → Trending, and a double tap from Trending to Top coins each switch in one step with no intermediate page, and a swipe from Top coins to Perps still slides. The served main bundle contains the 3.0.144 pager logic (`nativeTabPressAnimationEnabled ?? true`) and `nativeTabPressAnimationEnabled: false` from Market, with no trace of the removed workaround.

The first two device checks ran the app-side version of the fix on native modules 3.0.141, which sent a second, non-animated command from `onNativeTabPress`. The last commit replaces it with the 3.0.144 prop, which makes the pager wrapper send the same `setPageWithoutAnimation` itself. The last check covers that build.

[OK-63487]: https://onekeyhq.atlassian.net/browse/OK-63487?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

limichange (approved):

PR #13525, opened 2026-09-17
fix: leave device stage when its help pages open on native OK-63629

franco-chan:
[OK-63629](<https://onekeyhq.atlassian.net/browse/OK-63629>)

----------

## Summary

On iOS and Android, the hardware Device Stage's **Device not connected** card offers two help exits: *Self-troubleshooting* (opens the help-center article in the in-app browser) and *Support* (opens the Support webview page). Both open **underneath** the stage — its window overlay and touch wall cover the page — so the page cannot be tapped or scrolled until the person closes the stage by hand. The authenticity-failure card's *Support* exit has the same problem.

The stage now leaves as the page opens, through the same exit path as its close button:

- Device not connected → Self-troubleshooting / Support
- Authenticity check failed → Support. The run ends with no verdict, exactly like a manual close: no wallet is created and the check can be started again.

Jira: OK-63629

## Platform scope

Native only (`platformEnv.isNative`). Desktop, web and extension are unchanged: the article leaves the app, and the Intercom messenger mounts on `<body>`, above the app root's stacking context the stage lives in, so nothing is covered there and the card stays.

## Analytics

`deviceStageClosed.via` gains two values, `troubleshoot` and `support` (native only), so a help exit is not counted as a plain close.

## Out of scope

The legacy Device-not-connected dialog (still used where the stage does not carry the call: device search, the firmware update workflow) has the same two buttons and is left as is.

## Verification

- `yarn agent:check --profile commit` and `--profile pr` pass.
- Web dev build, stage driven into the Device-not-connected step, real clicks: the article opens externally and the card stays; the Intercom messenger opens above the stage wall (hit-tested) and the card stays — confirms desktop/web needed no change and did not get one.
- iOS: verified by the author on a physical iPhone (iOS 26.5.2, Debug build) — on the Device-not-connected card both exits open their page with the stage gone, the page is fully usable, and no stage is left once the page is closed. Recorded in OK-63629.
- Android: not run.

[OK-63629]: https://onekeyhq.atlassian.net/browse/OK-63629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #13527, opened 2026-09-17
fix: prevent renderer crash when switching Perps order tabs(OK-63620)

huhuanming:
[OK-63620](<https://onekeyhq.atlassian.net/browse/OK-63620>)

----------

Fixes the frozen desktop window seen on the 6.6.0 TestFlight build: the renderer crashed twice on the Perps page (`reason: crashed`, `exitCode: 5`) and the main process never reloaded it, so the window kept its last frame and ignored input.

## Root cause

Both crash reports stop on the same Blink CHECK. A mouse release forces a lifecycle update, the compositing-inputs phase commits the text selection, `FrameSelection::SelectionHasFocus()` forces another style update, and `Document::UpdateStyleAndLayoutTreeForThisDocument()` fails `CHECK(Lifecycle().StateAllowsTreeMutations())`. `x9 = 0x1520` in both reports is the compiled bitmask of that check.

The trigger in our code is the scroll-driven header animation on the Perps desktop tables (`timeline-scope` / `scroll-timeline` / `animation-timeline` in `packages/shared/src/web/index.css`). The order info panel switches tabs by toggling `display` on panes that contain these tables. When that happens inside the click while a text range selection exists anywhere on the page, the renderer aborts. It also explains why the crash did not reproduce on an account without TWAP orders: the empty state does not render the table.

The same CHECK is reported elsewhere, including an identical mouse-up stack in openai/codex#44322 and a paint-time variant in inokawa/virtua#901.

## Reproduction

Minimal page using our real `index.css` and the same `data-testid` structure, driven by real input events in Electron 43.1.1 (Chromium 150, same engine as the TF build):

- `index.css` on `x`: the renderer crashes (`crashed`, exit code 5) within 150 tab clicks. The local crash report has the same thread, trap, `x9` fingerprint and frame offset deltas as the TF reports.
- `index.css` on this branch: 2 runs of 150 clicks, no crash.
- The crash needs the scroll-driven animation, a `display` toggle during the click, and a text range selection. Removing any one of them avoids it.

## Changes

- Remove the scroll-timeline header animation and the ResizeObserver that measured its `--perp-desktop-table-scroll-end` endpoint. `handleDesktopBodyScroll` already mirrors the body's horizontal offset onto the header, which is the path browsers without scroll-timeline support use, so header movement is now main-thread driven.
- Desktop: reload the main window when its renderer process goes away. Skipped for clean exit, quit and soft restart, and capped at 3 reloads per 5 minutes so a crash on boot cannot loop.
- Log the crashed webContents type and whether it is the main window.

## Validation

- Reproduction above, before and after this change.
- Renderer recovery: forced a real renderer crash (`Page.crash`) in Electron 43.1.1. The page reloaded in a new renderer process, and a fourth crash within five minutes was not reloaded.
- Not verified in the app with live Perps data. Please check that horizontal scrolling keeps the header aligned with the rows in the Positions, Open Orders and TWAP tables on desktop and web.

[OK-63620]: https://onekeyhq.atlassian.net/browse/OK-63620?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13527>)

sidmorizon:
Cursor review failed.

originalix:

Review of commit `0874b06562fe` finished, but some comments could not be published automatically.

sidmorizon:
Cursor review failed.

Remote error: `cursor_request_failed: TimeoutError: The operation timed out.`

zhaono1 (approved):

originalix (approved):

PR #13528, opened 2026-09-17
fix: borrow empty state drops the position metrics and leads with holdings (OK-62718)

labrinyang:
[OK-62718](<https://onekeyhq.atlassian.net/browse/OK-62718>)

----------

## What

Phone-only Borrow empty state, two changes from the ticket plus the design
follow-through they forced.

**1. No net worth / health factor / net APY when there is no position.**
All three describe a position. A market the user holds nothing in would put
a zero under each, above a list inviting them to start. They stay up while
the position state is still unsettled — reserves in flight, or a load that
failed — so the row does not blank out and come straight back on each retry.

**2. The holdings list is ordered by balance, largest first.**
It was ordered by APY. Desktop's "Assets to supply" table already defaults to
balance descending, so the two ends disagreed on the same data. Fiat value is
what makes two different tokens comparable; the sort is stable, so everything
the user holds none of ties at zero and keeps the order the server sent.

**Follow-through:** refresh used to ride the metric row. With the row gone in
the empty state it was left stranded in the gap under the market picker, so it
now rides the list heading there instead. The two rows never coexist, so the
button appears exactly once per state. The summary strip below the list also
stopped rendering cells with nothing in them — a zero-valued rewards payload
arrives whether or not there is anything to claim, and it was drawing a rule
across the page for a "$0".

## Scope

- Phones only. The wide layout is built around the net worth hero and has no
  empty state to put in its place, so the metric gate stops at the breakpoint.
- The desktop metric strip still renders its bonus and rewards cells
  unconditionally; deliberately left alone.
- The platform bonus cell is unchanged.
- The list stays capped at 5.

## Testing

`yarn jest packages/kit/src/views/Borrow` — 57 suites / 545 tests.
New coverage: the sort's tie-break (pinned so it cannot pass on a fixture that
simply lacks APY), the metric gate across empty / loading / error / retry, the
summary's per-cell gates, and the empty state's refresh press.
Every new assertion was mutation-checked: each mutation reddens only its own
test.

## Manual check

Phone width, a market with no positions: no metrics at the top, refresh sits on
the "Earns on your holding" heading, the list leads with the largest holding.
Then a market with positions: metrics back, refresh back on their row.

[OK-62718]: https://onekeyhq.atlassian.net/browse/OK-62718?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (dismissed):

originalix (approved):

PR #13529, opened 2026-09-17
fix: keep the inline token symbol intact for narrow amounts (OK-63548)

Minnzen:
[OK-63548](<https://onekeyhq.atlassian.net/browse/OK-63548>)

----------

## Problem

In the Perps withdraw dialog, typing `1` renders the token symbol as `USD…` and opens a large gap between the digit and the symbol. Typing any other digit is fine.

The reported symptom ("USDC shows as USD") is not a missing character — the symbol is ellipsized by `numberOfLines={1}`.

## Root cause

The auto-size amount row budgets the amount box at `ceil(measuredText) + max(18, fontSize * 0.5)`, but the box renders at `containerProps.minWidth` (`ceil(fontSize * 1.2)`) whenever that is larger, because `min-width` beats both `width` and `maxWidth` in CSS.

Roobert has proportional figures: `1` is `0.338em` against `0.573em` for `2` and `0.636em` for `0`. Only `1` is narrow enough to fall under the floor, so the budget under-counts the row by ~30px at 84px. The fitter then keeps a font size the row cannot hold, and two things follow:

- the trailing `SizableText` is squeezed and ellipsizes `USDC` into `USD…`
- `isInlineInputShrunk` flips true, which disables the negative-margin compensation added in #13462, exposing the box's ~72px trailing caret buffer as a visible gap

Sweeping container widths from 240px to 560px against real Roobert metrics: `1` overflows at every width from 240px to 352px, and no other digit overflows at any width. The 351px-wide withdraw dialog sits in that range.

Introduced by #13462, which added the `minWidth` floor to the render path without updating the budget path.

## Fix

Give the budget and the render one shared box-width helper, so the fitted font size reflects the width the row actually renders at.

The rendered box width is unchanged — `max(ceil(m + buffer), ceil(1.2 * fontSize))` is exactly what the two previous variables computed. The only behavioural difference is the chosen font size. As a consequence `isInlineInputShrunk` is now unreachable unless the fitter bottoms out at `minFontSize`, since the budget guarantees `boxWidth <= available - reserved - 8`, so the gap is structurally fixed too.

After the fix, the same sweep reports zero overflow for every digit at every width, and all single digits settle on one font size instead of `1` picking a larger one than its neighbours.

## Scope

Web/desktop only. Native uses `AutoSizeInputView` with its own auto-sizing and does not run this math.

Shared with the Send amount page and the DeFi position action dialog, which use the same component.

## Validation

- `packages/kit/src/views/Send/components/SendAutoSizeAmountInput/` — 17 tests passing
- `yarn agent:check --profile commit` — `tsc-staged`, `lint-staged`, `background-api-contract`, `onekeyfe-patches`, `third-party-native-storage` all pass. `agent-context` fails on `x` already (skill catalog budget: 23 skills vs 22, 3205 description chars vs 3000) and is unrelated to this change.
- Not yet verified on a packaged build.

[OK-63548]: https://onekeyhq.atlassian.net/browse/OK-63548?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #13533, opened 2026-09-17
fix: sync btc address type selector on first click(OK-63639)

huhuanming:
[OK-63639](<https://onekeyhq.atlassian.net/browse/OK-63639>)

----------

## Summary
- Fix BTC address type title clicks being consumed by the desktop tooltip trigger.
- Let AddressTypeSelector own the active derive type so the outer label updates immediately.
- Refresh the wallet address list with the account cache bypassed after switching derive type.

## Root cause
WalletAddress passed the previous account derive type and derive info as controlled props, so the selector trigger kept rendering stale outer state until the page was reopened. The BTC title was also wrapped by Tooltip without forwarding the selection handler, so title clicks did not select the row.

## Verification
- Reproduced and verified in the desktop app on the Account address page.
- Taproot to Nested SegWit and back update the outer label and address on the first click.
- `git diff --check` and targeted oxlint passed.

[OK-63639]: https://onekeyhq.atlassian.net/browse/OK-63639?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #13535, opened 2026-09-17
fix: log slow SWR cache flush and mirror passes

originalix:
Diagnostic-only change: measure the whole-store SWR cache passes that run synchronously on the calling JS runtime, so an exported log can confirm or rule out the suspected cause of the iOS "lists scroll but taps do nothing" reports. No behavior change.

### Why

Log forensics on an affected iOS export (split `main`/`bg` runtimes) show:

- The `main` (UI) JS runtime stalls for 5–10 s at a time. A bg call that executes in 0 ms has its response handled by `main` 6–10 s later, and deferred `defaultLogger` lines are stamped seconds late. The `bg` runtime (trivial calls ~1 ms) and the native UI thread (100 ms timers on time, native lists keep scrolling) are healthy.
- Queued taps are not lost: they all fire at once when the stall ends.
- Every long stall ends with one large `nativeStorage` write dispatched from `main`, and in 11 of 12 cases the queued taps / touches / pending bg responses flush within the ~80 ms that bg needs for that write. None of those writes is preceded by a `[ColdStartCache] flush` marker, which leaves the SWR patch path.

Code reading matches: on native `main`, `swrCacheUtils.flush()` (2 s trailing debounce, triggered by every successful `swrKey` run of `usePromiseResult`, polling included) parses the whole serialized store, prunes it, applies the patch against the whole serialized string in the storage mirror, and re-serializes every entry to recount sizes. `bg` persists per key. `docs/native-list-selector-performance.md` already records a 225 ms main long task on this chain. What is still unmeasured is the store size and the real flush duration on an affected device — that is what this PR adds.

### What is logged

One local-file warn line, `app => perf => swrCacheSlowOp`, only when a pass takes >= 50 ms. It carries sizes and durations only, never cache keys or data, and is not sent to the server.

| `op` | Where | Fields |
| --- | --- | --- |
| `flush` | `swrCacheUtils.flush` | `durationMs`, `readMs` / `pruneMs` / `patchMs` / `adoptMs`, `storeChars`, `entryCount`, `updatedKeyCount`, `patchChars` |
| `reload` | `swrCacheUtils.reloadFromStorage` (perps L2 snapshot path) | `durationMs`, `storeChars`, `entryCount` |
| `mirrorApply` | native storage mirror, bg acknowledgement / broadcast (replays included) | `durationMs`, `storeChars`, `source`, `mutationOp`, `replayedCount` |

The scene adds `runtime` (`main` / `background` / `standalone`) so a UI-thread pass can be told from a background one. The threshold lives in `swrCacheLimits.ts`. No new files.

### How to read the result

- Confirmed: a `runtime: "main"` line sits at the end of each long silence with `durationMs` in the same range as the silence, a multi-MB `storeChars` and a KB-sized `patchChars`.
- Partly: `flush` stays in the hundreds of ms but `mirrorApply` lines add up to seconds.
- Ruled out: multi-second stalls remain without any second-level `swrCacheSlowOp` line nearby.

### Validation

- `swrCacheUtils.test.ts`, `nativeSyncStorageMirror.test.ts`, `nativeSWRCachePersistence.test.ts`, `nativeStorageExecutor.test.ts`: 101/101 passing, including 4 new cases (slow flush with per-phase timings, fast flush stays silent, slow reload, slow mirror apply tagged with its source).
- `yarn agent:check --profile commit` and `--profile pr` pass.
- No device verification: the point of the change is to collect the numbers from a test build.


sidmorizon:
@codex review

sidmorizon:
@codex security review

huhuanming (approved):

PR #13536, opened 2026-09-17
fix: bound native storage replay and acknowledgement backlogs

sidmorizon:
A failed native storage mirror refresh left bootstrap replay recording enabled, so later successfully acknowledged SWR updates kept accumulating in main memory. Compacted offline writes also retained a separate acknowledgement promise and superseded-ID history for every overwrite.

This change ends the replay window on failure, preserves the live mirror and unacknowledged writes, and retries runtime refreshes with exponential backoff. During a snapshot, canonical SWR entries merge by key while preserving deletions and clear/replacement ordering. More than 1,000 replay keys or 10 Mi serialized characters invalidates the snapshot and releases the replay buffer; a later snapshot rebuilds from bg plus outstanding writes. Initial startup failures remain caller-handled.

Compacted writes share acknowledgement groups, including through the MMKV storage wrapper, and resolve only after bg confirms the final write. In-flight writes retain independent confirmation.

Runtime scope: iOS/Android main keeps its own deserialized JS mirror; bg owns the MMKV persistence path. The runtimes initialize independently. Desktop/web use their existing single-runtime storage path, and extension uses the non-native assembly. Storage schema and RPC payloads are unchanged.

Validation:
- Reproduced before fixing: 10,000 same-key writes retained 10,000 confirmation promises; 1,000 acknowledged SWR writes after a failed refresh left 2,000 replay records.
- 141 tests passed across 8 storage/SWR suites, including 10,000-write stress cases, wrapper confirmation sharing, retries, replay limits, stale snapshots, deletes/clears, bg acknowledgement ordering, and failures while superseding an initial snapshot.
- `yarn agent:check --profile commit` and `yarn agent:check --profile pr`.
- These are executable module regressions with a mocked native transport; no on-device iOS/Android stress run was performed.


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix:

Review of commit `5204ba46438d` did not finish successfully.

sidmorizon:
Cursor review failed.

originalix:
The replay budget can make a refresh unrecoverable, not just skip one snapshot.

`startBootstrap` re-seeds `mutationsBeforeBootstrap` from the **unacknowledged pending queue** before the snapshot call, and that seeding runs through the same `appendCompactedLocalMutation` budget. If the pending queue alone names more than `SWR_CACHE_MAX_ENTRIES` distinct keys, every attempt rejects during seeding — the snapshot never gets a chance, and the retry seeds from the same queue and fails identically.

The gap is that the two limits measure different things. The queue's own compaction guards are `SWR_PATCH_COMPACTION_MAX_PENDING` (100 patches) and `SWR_PATCH_COMPACTION_MAX_CHARS` (10 MiB); neither bounds the number of distinct keys. So a queue can sit legally under both of its thresholds and still be over the replay buffer's 1000-key budget. The 10 MiB char budget is mostly shielded because patch compaction triggers at a similar size first; the key-count budget is not shielded at all.

Reproduced against `264f9fd2f3` with a mocked transport where `bootstrap` answers normally and `syncStorage` writes are held (bg write path wedged):

- 50 pending patches x 11 keys = 550 distinct keys -> `refreshNativeSyncStorageMirrors()` resolves.
- 100 pending patches x 11 keys = 1100 distinct keys -> every `refreshNativeSyncStorageMirrors()` rejects with `snapshot replay budget exceeded`. Over six simulated minutes the automatic retries fired 17 bootstrap RPCs, all discarded. It recovers only once bg drains the pending writes.

Consequences are degradation rather than data loss: the mirror keeps serving its own data and writes keep queuing, but it never re-syncs with bg, so after a bg restart the UI runtime stays stale until the queue drains or the app restarts. Meanwhile the backoff settles at one bootstrap RPC every 30s, and each one makes bg run a full `buildBootstrapSnapshot` (walk all MMKV keys plus `readSerializedSubset`) that main throws away.

Initial startup is unaffected in practice: the pending queue is empty then, and `hasBootstrapSnapshot` is false so nothing auto-retries.

Two directions, whichever you prefer:

1. Don't apply a second, tighter budget to the seeding step. The pending queue is already bounded by its own compaction, so re-recording it doesn't need the buffer's cap.
2. On overflow, drop the buffer but keep the snapshot: prime, then replay the pending queue through the path `replayPendingLocalMutations` already uses. What the buffer uniquely protects is bg-originated broadcasts arriving during the window, not this runtime's own pending writes.

Happy to hand over the probe test if useful — it's a single file against this branch, no production changes.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-f06aab07-bbc1-5b64-8796-32e234258cd2)

packages/shared/src/storage/instance/nativeSyncStorageMirror.backlog.test.ts
```diff
@@ -0,0 +1,478 @@
...
+    mockCallNativeStorage(request) as Promise<unknown>,
+}));
+jest.mock('../nativeSWRCachePersistence', () => {
+  const actual = jest.requireActual<
+    typeof import('../nativeSWRCachePersistence')
+  >('../nativeSWRCachePersistence');
+  return {
+    ...actual,
+    applyNativeSWRCacheCanonicalEntries: (
+      ...args: Parameters<typeof actual.applyNativeSWRCacheCanonicalEntries>
```
originalix: P1: [New backlog test references a removed persistence export]

`nativeSWRCachePersistence.ts` does not export `applyNativeSWRCacheCanonicalEntries`, so this `typeof actual.applyNativeSWRCacheCanonicalEntries` reference fails the repository TypeScript check before the new backlog tests can run.

Please remove this stale mock wrapper or change it to target an export that exists, then rerun the TypeScript and focused test checks.

PR #13538, opened 2026-09-17
fix: match Market chart prices to the header quote (OK-63597)

zy2000-star:
[OK-63597](<https://onekeyhq.atlassian.net/browse/OK-63597>)

----------

## Summary

- Always append the header quote as the Simple chart's last point, instead of
  skipping the merge when the last historical bar falls outside the visible
  window.
- Give the Simple chart axis (`formatChartPrice`) the same sub-$1 rule as the
  header (`formatPrice`), and render the hover bubble with `NumberSizeableText`
  so the three labels read identically.
- Stop the Pro chart tick formatter from rounding a quote to a different number
  than the header.

## Root cause

One symptom, two unrelated defects.

**Data.** `mergeStockSimpleChartLivePrice` skipped the merge when the last
closed bar sat outside `rangeSeconds`, or when its timestamp was ahead of the
clock. A pre-market share on 1H therefore kept showing the previous session's
close while the header already carried the live quote.

**Display.** Three independent formatters exist for one concept, and only the
header was authoritative:

| | significant digits | rounding | subscript above |
| --- | --- | --- | --- |
| `formatPrice` (header) | 4 | BigNumber `ROUND_HALF_UP` on the decimal string | 4 zeros |
| `formatChartPrice` (Simple axis) | full expansion, then truncated to a character budget | truncation | 5 zeros |
| `formatTradingViewNativePriceTick` (Pro chart) | 4 | `toFixed` on the double | 3 zeros |

Both chart formatters could print a different number than the header for the
same quote: truncation dropped digits (`0.123456` vs `0.1235`), and `toFixed`
rounds the binary double, so `0.0012345` — stored as `0.00123449…` — rounded
down to `0.001234` where the header rounded up to `0.001235`.

## What changed

Both chart formatters now round the digit string returned by
`toExponential()`, which is the shortest round-trip decimal. That reproduces
BigNumber's `ROUND_HALF_UP` on the original decimal without pulling BigNumber
into a WebView-embedded function or a Reanimated worklet.

The Simple axis also adopts the header's subscript threshold. That is not
cosmetic: at five leading zeros the plain form needs 11 characters, so the
10-character budget would cut the last digit off and change the value.

## Deliberately not changed

- The Pro chart keeps its subscript threshold of 3. `0.0₄9463` and
  `$0.00009463` are the same number, and a mobile chart may compact harder than
  the header.
- The Simple axis keeps dropping trailing zeros and compacting K/M/B, so a
  round scale does not read `$100.00 / $200.00 / $300.00`.

Both are pinned by tests, so the next reader does not mistake them for bugs.

## Verification

- 1093 tests across 136 suites pass for `LightweightChart`, `StockPriceLineChart`,
  `StockSimpleChart` and `TradingViewNative`.
- New parity tests compare each chart formatter against the header over a price
  sweep. The Pro chart comparison expands subscripts and normalizes trailing
  zeros first, so it asserts the value and allows the notation to differ.
- `yarn agent:check --profile commit` and `--profile pr` pass locally.
- Not yet verified in a running app. The unit tests prove the formatters agree,
  not that the charts render as expected — the Pro chart path is a worklet.

[OK-63597]: https://onekeyhq.atlassian.net/browse/OK-63597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

originalix (approved):

PR #13540, opened 2026-09-17
fix: hold watchlist stock rows until quotes resolve (OK-63638)

zy2000-star:
[OK-63638](<https://onekeyhq.atlassian.net/browse/OK-63638>)

----------

## Summary
- Hold favorited stock/asset rows back until their first quote batch resolves, so the loading spinner covers them instead of a half-rendered row.
- Reuse the resolved quote batch across remounts, so switching tabs in the market detail token selector no longer rebuilds the favorites list from a quote-less state.
- Render an unresolved price as `--` instead of `NaN` in the token selector row.

## Root cause
Spot rows in `useMarketWatchlistTokenList` are keyed off a fetched batch, so they stay absent until data arrives and the list's `isLoading && data.length === 0` guard keeps the spinner up. Listing rows (stocks, assets) are not: the hook synthesizes them from the locally stored watchlist record, which already carries the symbol and name. They render immediately with every metric unresolved — the price column pushed `String(NaN)` into `NumberSizeableText` — and their presence made `data.length` non-zero, which suppressed the spinner for the whole list.

The market detail token selector unmounts the favorites list on every tab switch, so each switch replayed that quote-less state: stock rows flash in as a name plus placeholders, crypto rows pop in underneath a moment later, then the stock metrics fill in.

## Fix
- `useMarketWatchlistTokenList` returns `undefined` for a listing row until a quote batch has resolved. A resolved batch that carries no entry for a given listing still renders the row, which keeps a delisted favorite removable.
- The resolved batch is cached as `IMarketWatchlistDataCache.listing` and reused on remount. It is only reused while it still covers every current watchlist entry, so favoriting a new stock cannot render that stock against a stale batch.
- `MarketTokenSelector` owns the cache ref, since the shell outlives the tabs that mount and unmount the list.
- `MarketTokenSelectorRow` falls back to `MissingValue` when `item.price` is not finite.

## Scope
`useMarketWatchlistTokenList` also backs the Market home favorites list, so the withholding behavior applies there too — a favorited stock now waits for its quote instead of appearing with placeholder metrics. Affects desktop, web, extension, and mobile. No background, storage, or network-layer changes.

## Verification
- `yarn agent:check --profile commit` — 9/9 pass.
- Targeted tests — 8 suites / 65 tests pass.
- New regression tests: listing rows withheld until the first batch resolves; cached batch reused across a remount; cached batch dropped once a new listing is favorited; price column renders `--` for `NaN` and stays numeric for a real price.
- Desktop app: open a market detail page, open the token selector with stocks in favorites, switch between tabs — stock rows no longer flash in ahead of their quotes.

[OK-63638]: https://onekeyhq.atlassian.net/browse/OK-63638?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13540>)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -192,6 +198,38 @@ export function useMarketWatchlistTokenList({
...
+    // A batch fetched before a newly favorited listing would leave that row
+    // without a quote, so the cache stays usable only while it still covers
+    // every current entry.
+    const cachedKeys = new Set(cached.map((entry) => entry.key));
+    return listingItems.every((item) =>
+      cachedKeys.has(getMarketWatchlistKey(item)),
+    )
+      ? cached
+      : undefined;
+  }, [listingItems, listingQuotes]);
```
sidmorizon: 🟡 P2: New favorites bypass quote hold

Severity: `non-severe`

When a listing is favorited while the list stays mounted, `listingResult` immediately accepts the previous quote batch. [`usePromiseResult`](packages/kit/src/hooks/usePromiseResult.ts:287-290) preserves that batch during refetches, leaving the new listing unmatched. Its incomplete row suppresses the loading fallback and exposes unresolved metrics.

Suggested fix:

```
  const listingResult = useMemo(() => {
    const candidate = listingQuotes ?? lastListingQuotesRef.current;
    if (!candidate) return undefined;
    // A batch fetched before a newly favorited listing would leave that row
    // without a quote, so the result stays usable only while it still covers
    // every current entry.
    const candidateKeys = new Set(candidate.map((entry) => entry.key));
    return listingItems.every((item) =>
      candidateKeys.has(getMarketWatchlistKey(item)),
    )
      ? candidate
      : undefined;
  }, [listingItems, listingQuotes]);
```

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
```diff
@@ -192,6 +198,38 @@ export function useMarketWatchlistTokenList({
...
+    lastListingQuotesRef.current = listingQuotes;
+    if (dataCacheRef) {
+      dataCacheRef.current = {
+        ...dataCacheRef.current,
+        listing: listingQuotes,
+      };
+    }
+  }, [dataCacheRef, listingQuotes]);
+  const listingResult = useMemo(() => {
+    if (listingQuotes) return listingQuotes;
```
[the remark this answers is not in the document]
zy2000-star: Fixed in `6d03c6bd40`. Good catch — `usePromiseResult` is not called with `undefinedResultIfReRun` here, so `listingQuotes` does keep the previous batch across a refetch and the early return skipped the coverage check entirely.

I took a slightly different shape than the suggested patch: coverage is now checked **per row** instead of on the batch as a whole.

```ts
const listingResult = listingQuotes ?? lastListingQuotesRef.current;
// ...
const entry = listingResult?.find((item) => item.key === key);
if (!entry) return undefined;
```

Rejecting the whole batch would also drop the listing rows that are already on screen, so favoriting a new stock would blank the existing ones until the refetch lands. Resolving per row keeps the covered rows visible and holds back only the uncovered one, which matches how spot rows already behave.

Delisted favorites stay removable: the batch is built from the requested items rather than the response (`stockItems.map(...)`), so a stock the backend no longer returns still gets an entry with an undefined quote and renders.

Regression tests added — both fail on the previous commit:
- `withholds a newly favorited listing while the previous batch is reused` (list stays mounted, exactly the case reported here)
- `keeps cached rows and holds back only the uncovered listing` (remount with a cached batch)

packages/kit/src/views/Market/MarketHomeV2/components/MarketTokenList/hooks/useMarketWatchlistTokenList.ts
zy2000-star: Fixed in `6d03c6bd40`. Good catch — `usePromiseResult` is not called with `undefinedResultIfReRun` here, so `listingQuotes` does keep the previous batch across a refetch and the early return skipped the coverage check entirely.

I took a slightly different shape than the suggested patch: coverage is now checked **per row** instead of on the batch as a whole.

```ts
const listingResult = listingQuotes ?? lastListingQuotesRef.current;
// ...
const entry = listingResult?.find((item) => item.key === key);
if (!entry) return undefined;
```

Rejecting the whole batch would also drop the listing rows that are already on screen, so favoriting a new stock would blank the existing ones until the refetch lands. Resolving per row keeps the covered rows visible and holds back only the uncovered one, which matches how spot rows already behave.

Delisted favorites stay removable: the batch is built from the requested items rather than the response (`stockItems.map(...)`), so a stock the backend no longer returns still gets an entry with an undefined quote and renders.

Regression tests added — both fail on the previous commit:
- `withholds a newly favorited listing while the previous batch is reused` (list stays mounted, exactly the case reported here)
- `keeps cached rows and holds back only the uncovered listing` (remount with a cached batch)

originalix (approved):

PR #11798, opened 2026-05-27
perf: improve perps cold start

originalix:
## Summary
- Hydrate Perps/Pulse market and account display state from scoped cold-start caches to reduce first-frame flicker and subscription churn.
- Add market-data freshness checks so order entry is blocked only when the feed is confirmed stale or disconnected.
- Add cold-start diagnostics and targeted regression tests for subscription planning, SWR cache keys, account snapshots, and freshness guards.

## Intent & Context
This change is focused on improving Perps/Pulse cold-start performance and perceived stability. Cached data is used for display-only first-frame hydration, then replaced by live WebSocket/status data as soon as it arrives. During review, the market freshness guard was confirmed as expected trading protection rather than a pure rendering optimization.

## Root Cause
Cold-start UI state was split across several async sources: active instrument, L2 book, all-dex asset contexts, account summary, spot balances, abstraction mode, and live account status. Some UI paths treated unknown values as empty values, causing Deposit/Enable/Long-Short surfaces to jump while data was still loading. Subscription updates also missed route-only state transitions in some selector scenarios.

## Design Decisions
- Keep account display cache scoped by address/account and use it only for display; live status remains required for trading guards.
- Use exact L2 snapshot keys so option-specific order book snapshots do not fall back to default tick snapshots.
- Keep route subscription flags in a stable `subscriptionStateKey` so subscription updates rerun when route-only flags change.
- Treat confirmed stale/disconnected market data as a trading guard condition, while initial no-message/socket-pending states remain loading states.

## Changes Detail
- `ServiceHyperliquid` and `SimpleDbEntityPerp`: persist and hydrate display-only account/market snapshots with account alignment checks.
- `ServiceHyperliquidSubscription` and Perps actions: cache L2/all-dex snapshots, reduce initial subscription latency, and track market liveness.
- Perps UI components: consume cold-start snapshots for ticker/order book/account display, stabilize mobile layout, and keep trading buttons disabled while live status is pending.
- `swrCacheUtils` and tests: centralize Perps L2 cache keys and assert option-specific keys do not fall back.
- Tests: add coverage for market freshness, subscription planning, account snapshot matching, SWR key selection, and cold-start perf marks.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Mobile / Desktop / Web / Extension
- **Risk Areas**: Perps WebSocket subscription state, display-only cache freshness, account/address alignment, and order-entry freshness guard.

## Test plan
- [x] `git diff --cached --check`
[...]

PR #13543, opened 2026-09-17
fix: restore perps offline status with debounced liveness detection(OK-63588)

originalix:
[OK-63588](<https://onekeyhq.atlassian.net/browse/OK-63588>)

----------

## Summary
- Restore the Perps offline state (top banner + footer status badge). Since #11798 a dropped connection was published as "pending", which every consumer renders as online.
- Debounce the status at its source: a single 8s window covers socket failures and sockets that stay open without traffic. Blips shorter than that never change the UI, sustained outages always surface, and offline is only cleared by a real data frame.
- Fix an endless transport rebuild loop (every ~5.5s) that started after switching markets while offline and kept going after the network recovered.

## Intent & Context
OK-63588: while the network is down, Perps never shows an offline state; the expectation is to bring it back (the banner restyle is a follow-up owned by the reporter). Product requirement for the fix: short network jitter must not flip the status, so users do not see an over-sensitive indicator, but a real outage must be reflected faithfully so users do not assume the product is broken.

A first iteration only handled close events. Testing it on desktop with DevTools "Offline" showed the badge still took very long to change, which exposed the half-open case and the rebuild loop below.

## Root Cause
1. **Offline never published** (regression from #11798, f8642dec00a, shipped since v6.4.0). Socket close, forced reconnect and subscription cleanup were changed from `connected: false` to `undefined`, while the UI treats anything except `false` as online. The only remaining writer of `false` was the 60s inactivity timer (effectively 60–120s). Its value was overwritten within ≤8s by the close event that rews re-dispatches for every failed retry, and its "still alive" branch wrote `connected: true` without any evidence. Reproduced with fake timers: during a 300s outage the banner was visible for 2–3s in total.
2. **Half-open sockets.** Offline emulation, and real upstream loss (Wi-Fi up, no internet), keep established WebSockets OPEN while no frames arrive, so close-based detection never fires. Measured on desktop: the old socket stayed open with 0 frames/s. Pings cannot serve as the signal because desktop only pings while the window is focused and the native bg runtime receives no AppState events.
3. **Rebuild loop.** Switching markets while offline leaves the previous order book spec in `allSubSpecsMap`, since its unsubscribe can never complete. Every new socket then tries to unsubscribe it again. Hyperliquid answers `Already unsubscribed` but echoes normalized params (`s:null`, `m:null`; `fast:false` for `l2Book`), and the SDK rebuilds the request id from that echo and matches it exactly. The request therefore only fails at the 5s timeout. The order book transition then calls `reconnect()`, the spec survives into the next transport, and the cycle repeats forever. Verified against the live API through the SDK: `allMids` resolves `Already unsubscribed` in ~400ms, while `l2` and `l2Book` always time out at 5001ms. A local desktop log showed 70 rebuilds in 7 minutes after the network had already recovered.

## Design Decisions
- **One debounce, at the source.** The debounce lives in the bg subscription service rather than in UI components, so the banner, badge and refresh button can never disagree.
- **One window.** `HYPERLIQUID_NETWORK_OFFLINE_GRACE_MS = 8000` replaces the 60s constant.
  - 8s is one WS connection timeout (5s) plus the retry that follows it, so a network handover or a server-side reset stays invisible.
  - `allMids` alone ticks several times per second, and the service already treats 10s without data as "no data flow".
- **Transport failures** (close, forced rebuild) start a countdown where the first evidence wins: closes from later failed retries neither restart nor cancel it.
- **Data-silence watchdog.** Silence is measured from the last processed frame or from the socket open. The watchdog publishes directly, because the silence itself already spans the window.
- **Clearing offline.** Only a data frame clears `false`.
  - A socket open settles `pending` but leaves `false` untouched, because subscriptions can still fail on that socket.
  - Exception: a socket that opens while the handler is muted clears a stale `false` back to pending, since muted frames never reach the liveness path.
- **Muted handler** (blur, lock or pause): silence is never judged while muted.
- **Frozen runtimes.** A network timer that fires more than 2s late means the runtime was frozen (backgrounded app, system sleep). The timer grants a fresh window instead of publishing offline immediately on resume.
- **#11798's intent is kept.** Cold start and deliberate rebuilds still show pending (online) rather than flashing offline.
- **Loop fix.** A closed or replaced transport forgets `allSubSpecsMap`, because a new socket starts without server-side subscriptions, and destroy skips the network unsubscribe for specs the current transport never held.
  - Rejected alternative: patching the SDK matcher, which needs a patch-package change.
  - Rejected alternative: sending explicit `null` params, which depends on the server's echo format.

## Changes Detail
- `ServiceHyperliquidSubscription.ts`:
  - Offline grace timer and data-silence watchdog, both drift-aware.
  - Status helpers: pending keeps offline; socket open does not clear offline.
  - `_forgetTransportSubscriptions()` on close and forced rebuild.
  - Destroy guard for specs the current transport never held.
- `perp.constants.ts`: replace `HYPERLIQUID_NETWORK_INACTIVE_TIMEOUT_MS` (60s) with `HYPERLIQUID_NETWORK_OFFLINE_GRACE_MS` (8s).
- `ServiceHyperliquidSubscription.test.ts`: 10 behavior tests.
  - Network status: a blip is never published; a dropped socket goes offline after the window; an open socket with no frames goes offline; a reopened socket with no frames goes offline; offline is held until a frame; muted silence is ignored; a muted reopen clears stale offline; a frozen runtime gets a fresh window.
  - Transport replacement: no rebuild loop over a stale order book; held specs are still unsubscribed.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Extension / Mobile / Desktop / Web (shared bg service)
- **Risk Areas**:
  - **Status timing after suspension** (native resume, hidden web tabs), which relies on the drift detection. By static analysis, a hidden web tab may publish offline late and corrects within 8s after becoming visible.
  - **`allSubSpecsMap` reset** on close and forced rebuild, which touches subscription diffing.
  - **Semantics change**: a socket open alone no longer clears a published offline; the first frame does, a few hundred ms later. `PerpsWebSocketRecovered` (TradingView `FORCE_RECOVER_WS`) now fires only after an outage that was actually published.
  - **Known limits**:
    - The watchdog does not force-reconnect half-open sockets. Recovery still relies on rews, `navigator.onLine`/NetInfo, or pings hitting a reset.
    - In the extension, DevTools "Offline" on the expand tab does not affect the service worker socket.

## Test plan
- [x] Jest: all `ServiceHyperLiquid` suites plus `perpsMarketDataFreshness` (23 suites, 203 tests). The new tests fail on `origin/x` (10/10) and on the close-only first iteration (6/10).
- [x] Desktop dev (Electron 43), driven over CDP with `Network.emulateNetworkConditions`:
  - Sustained offline: badge and banner show Offline after 8.09s, and return to Online 0.52s after going back online.
  - 3s blip: no status change within the following 23s.
  - Market switch while offline, then recovery: Online after 0.78s, and only one new socket in the next 45s (no rebuild loop).
- [ ] iOS: 100% packet loss with Wi-Fi still connected (Network Link Conditioner) shows the banner after ~8s; background for ≥30s then foreground shows no offline flash.
- [ ] Android: airplane mode and upstream-only loss show the banner after ~8s; recovery clears it.
- [ ] Web: DevTools Offline shows the footer badge Offline after ~8s; an outage while the tab is hidden is corrected within 8s after returning.
- [ ] Extension: go offline at the OS level or in the service worker DevTools; expand-tab DevTools does not affect the service worker socket.
- [ ] Switch markets while offline, restore the network, and confirm no repeated reconnects.

[OK-63588]: https://onekeyhq.atlassian.net/browse/OK-63588?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

huhuanming (approved):

PR #13552, opened 2026-09-17
fix: re-enable import address confirm after name error clears(OK-63654)

huhuanming:
[OK-63654](<https://onekeyhq.atlassian.net/browse/OK-63654>)

----------

## Summary
- Fix the Import address (watch-only) Confirm button staying disabled after a duplicate account name is corrected. It only re-enabled once the address was re-entered.
- Derive a boolean `hasActiveFieldError` at render time. The `isEnable` memo now depends on it instead of react-hook-form's mutable `formErrors` object.
- Add a regression test that drives registered fields through: valid address → duplicate name → blur → rename.

## Intent & Context
[OK-63654](https://onekeyhq.atlassian.net/browse/OK-63654), reported on iOS with a screen recording:
1. In the Watch-Only wallet, open Add account → Import address and paste a valid address.
2. Enter an existing account name. "The name already exists" shows and Confirm is disabled, which is expected.
3. Dismiss the keyboard, then change the name to a unique one.
4. The error disappears but Confirm stays disabled. It only re-enables after the address is cut and pasted again.

## Root Cause
`isEnable` in `useImportAddressForm` is a `useMemo` that listed `formErrors` (from `useFormState`) as a dependency. react-hook-form mutates `formState.errors` in place with `set`/`unset` on the same object, so the object's identity never changes when an error is added or cleared.

What happens in the reported flow:
1. **Duplicate name:** `form.setError('accountName')` emits `isValid: false`. The memo recomputes and Confirm is disabled (correct).
2. **Keyboard dismissed:** the name input blurs. In `onChange` mode RHF skips field validation on blur but calls `_updateValid()`. That re-runs the built-in rules, none fail, and `isValid` flips back to `true` while the manual duplicate-name error is still set. The memo recomputes and Confirm is still disabled because of the error (correct).
3. **Name edited:** RHF removes the `accountName` error in place. `isValid` is already `true`, so no dependency changes and `isEnable` stays `false`.
4. **Address re-entered:** `addressValue.pending`/`resolved` change, which finally recomputes the memo.

Without the blur in step 2, the edit in step 3 flips `isValid` from `false` to `true` and the button recovers. That is why the bug needs the keyboard to be dismissed first.

## Design Decisions
- Compute the primitive `hasActiveFieldError` during render and depend on it. Every errors emission already re-renders the host through `useFormState`, so the value is always current. The enable rules themselves are unchanged.
- Rejected: re-syncing `isValid` (e.g. `trigger`) after the debounced duplicate check clears the error. That does not cover RHF clearing the error itself on change, and it adds extra validation passes.
- The hook is shared, so the bug and the fix are not iOS-specific. Dismissing the keyboard is just the most common way to blur the field. `WatchOnlyWalletContent` (WebDapp) uses the same hook.

## Changes Detail
- `packages/kit/src/views/Onboarding/pages/ImportWallet/hooks/useImportAddressForm.ts`: derive `hasActiveFieldError` outside the memo and use it in place of the `formErrors` dependency.
- `packages/kit/src/views/Onboarding/pages/ImportWallet/hooks/useImportAddressForm.test.tsx`:
  - Make `ensureAccountNameNotDuplicate` a controllable mock.
  - Add a test that registers `addressValue` and `accountName` via `useController` and replays the blur + rename sequence.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile / Desktop / Web / Extension. This covers the shared Import address form, including the WebDapp watch-only content.
- **Risk Areas**: Confirm-button gating on the Import address page. Duplicate names and invalid addresses still disable Confirm, and submission still passes `shouldCheckDuplicateName: true`.

## Test plan
- [x] `yarn jest packages/kit/src/views/Onboarding/pages/ImportWallet/hooks/useImportAddressForm.test.tsx`: 9/9 pass.
  - Without the fix, the new test fails.
  - On the old code, the same test passes if the blur step is removed.
- [x] `yarn agent:check --profile commit`: all checks pass.
- [x] iOS 26.5 simulator (DevSession), repro steps from the report with watch-only account "U" existing:
  - Before the fix: after renaming "U" to "Uppp", the error clears but Confirm stays disabled.
  - After the fix: Confirm is enabled without re-entering the address, the import succeeds, and "Uppp" is listed.
  - The duplicate name "U" still disables Confirm.
- [ ] Same flow on Android and desktop.

[OK-63654]: https://onekeyhq.atlassian.net/browse/OK-63654?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13552>)

originalix (approved):

PR #13553, opened 2026-09-17
fix: reopen the iOS image picker after denied photo access (OK-48227 OK-56123)

huhuanming:
[OK-48227](<https://onekeyhq.atlassian.net/browse/OK-48227>) [OK-56123](<https://onekeyhq.atlassian.net/browse/OK-56123>)

----------

## Summary

- On iOS, the image picker opens again after a user has denied photo library access. Denying it once used to make the OneKey ID avatar and the hardware wallpaper "+" entry dead: iOS never shows that prompt a second time, so every later tap failed silently (OK-48227).
- `react-native-image-crop-picker` 0.51.1 is replaced by `@onekeyfe/react-native-image-crop-picker` 3.0.148 ([app-modules#117](https://github.com/OneKeyHQ/app-modules/pull/117)). iOS picks with `PHPickerViewController`, which needs no photo library permission at all; Android keeps the system Photo Picker.
- The cropper is now one app-styled screen on both platforms: a "Crop image" header with a rotate button, and capsule Cancel / Confirm buttons in the app's large `Button` style. It follows the app's light and dark theme. It replaces TOCropViewController's dark screen on iOS and uCrop's toolbar screen on Android.
- Native modules go 3.0.145 → 3.0.148, which also brings the Android auto-size-input font fix (OK-57825, app-modules#116), the iOS log-roll fix and the iOS Lite card NFC cancel crash fix.

## Intent & Context

The OK-48227 recording shows the hardware wallpaper screen: tapping "+" under 自定义 does nothing, with no picker and no permission prompt. The same dead-button behavior applies to the OneKey ID avatar.

## Root Cause

- `ImageCrop.openPicker` called `react-native-image-crop-picker`'s iOS `openPicker`, which asks for **full** photo library access with `[PHPhotoLibrary requestAuthorization:]` before it shows its own `QBImagePickerController`, and rejects with `E_NO_LIBRARY_PERMISSION` for any status other than `authorized`.
- iOS presents that system prompt only once per app install. After "Don't Allow", every later call rejects immediately, and the app cannot re-request; only Settings can change it.
- Nothing surfaced the rejection: `HardwareHomeScreenModal.pressUpload` awaits the promise with no `catch`, and `ProfileEditPage` only special-cases `E_PICKER_CANCELLED`, so the user saw a button that did nothing.

## Changes Detail

- `apps/mobile/package.json`: `react-native-image-crop-picker` → `@onekeyfe/react-native-image-crop-picker@3.0.148`, imported by its own name; all other `@onekeyfe` native modules 3.0.145 → 3.0.148 (also `package.json`, `packages/kit`, `packages/components`).
- `apps/mobile/ios/Podfile`: drops the `TOCropViewController` git pod override. The new module vendors TOCropViewController 3.2.0 (up from 2.8.0) with the OK-51551 rotation fix re-applied, so the override is no longer needed. `Podfile.lock` and the Xcode resource bundle references follow from `pod install`.
- `packages/components/src/composite/ImageCrop/type.ts`: options now extend the new package's `Options` instead of re-declaring the old library's `CommonOptions`.
- `packages/components/src/composite/ImageCrop/index.native.tsx`: drops `mediaType`, `forceJpg` and `sortOrder`, which the new module does not need (results are always JPEG, single photo). Both `openPicker` and `openCropImage` now pass `cropperAppearance`: the `$bgApp`, `$text`, `$icon`, `$bgStrong(Active)`, `$bgPrimary(Active)` and `$textInverse` tokens for the current theme, `uiScale`, and Roobert SemiBold / Medium. They also pass the `global.crop_image` title that the web cropper dialog already uses.
- `packages/components/src/hocs/Provider/hooks/useAppearanceTheme`: new `getAppThemeVariant()` returns the variant the system bars follow (a foreground surface's pin, else the app theme) so the native cropper matches what is on screen; `undefined` off native.
- `apps/mobile/bundle-registry/module-id-registry.json`: one new module id for `@onekeyfe/react-native-image-crop-picker/lib/module/index.js`.
- Gallery story: adds a "crop last result" button so `openCropImage` — the path the NFT wallpaper flow uses — can be exercised without hardware.

## Risk Assessment

- **Risk Level**: Medium (native picker replacement on both platforms).
- **Affected Platforms**: iOS and Android. Web and desktop use the existing `react-mobile-cropper` implementation, untouched.
- **Affected flows**: OneKey ID avatar, hardware wallpaper upload, NFT → hardware wallpaper crop, and the Gallery story.
- **Behavior changes**: iOS shows the system photo picker instead of the library's own album grid, and no longer asks for photo library permission. The cropper on both platforms is the new app-styled screen. On Android it rotates with the button only (no two-finger rotation), shows the grid only while the image is moved, and stays in portrait like the app.
- **Unchanged**: JS API and error codes (`E_PICKER_CANCELLED` etc.), the result shape consumed by callers (`data`, `width`, `height`, `path`), and the Android system Photo Picker.

## Test plan

- [x] Reproduced on `x` on an iPhone 17 Pro simulator (iOS 26.5), Developer → Gallery → ImageCrop: first tap prompts, "Don't Allow", and every later tap does nothing (LogBox: `Uncaught (in promise): Error: User did not grant library permission.`).
- [x] This branch on the same simulator, with Photos still denied (`kTCCServicePhotos|so.onekey.wallet|0`): the system picker opens with no prompt, cropping returns an exact 500×500 `image/jpeg` with base64, five rotations keep the crop box full size, cancelling the picker or the cropper rejects with `E_PICKER_CANCELLED`, and cancelling the crop opened from the picker returns to the picker.
- [x] `openCropImage` on a `file://` path (the NFT wallpaper path) opens the cropper and returns an exact 300×200 JPEG.
- [x] Android emulator (API 36): the system Photo Picker opens with no permission dialog, uCrop crops to an exact 500×500 JPEG with base64, `openCropImage` returns an exact 300×200 JPEG, and cancelling rejects with `E_PICKER_CANCELLED`.
- [x] Both platforms re-verified after this branch's `yarn install`/`pod install`, i.e. against the published `3.0.147` package rather than a local tarball. `Podfile.lock` records `ReactNativeImageCropPicker (3.0.147)`, and the installed iOS binary contains the new module with no `RNImageCropPicker` or `QBImagePickerController` symbols.
- [x] 3.0.148 cropper screen, Gallery → ImageCrop with native shells built locally. The sources were byte-identical to the published `3.0.148` tarball (`ios/`, `android/src/`, `src/` diffed after publishing).
  - iOS (iPhone 17 Pro simulator, iOS 26.5) in light and dark: the header, crop box, dimming and Roobert buttons follow the theme. Results are exactly 500×500 and 300×200. Cancel from `openCropImage` rejects with `E_PICKER_CANCELLED`. Cancel from the picker's cropper returns to the picker, where one tap on the same photo reopens the cropper.
  - Android (API 36 emulator) in light and dark: the same layout edge to edge, with bar icons that follow the theme. Results are exactly 500×500 and 300×200, the grid shows only while dragging, and the back gesture rejects with `E_PICKER_CANCELLED`.
  - The rotate button turns the image 90° counterclockwise on both platforms; both outputs matched a counterclockwise rotation of the source photo.
- [x] `yarn agent:check --profile commit` after bumping to 3.0.148.

Not covered: physical devices, iPad, iOS below 26, Android devices without the system Photo Picker, and the OneKey ID / hardware wallpaper screens themselves (both call the same `ImageCrop.openPicker`).

[OK-48227]: https://onekeyhq.atlassian.net/browse/OK-48227?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[OK-56123]: https://onekeyhq.atlassian.net/browse/OK-56123?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

image


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13553>)

packages/components/src/composite/ImageCrop/index.native.tsx
```diff
@@ -41,8 +38,7 @@ const openPicker: IOpenPickerFunc = async (params) => {
   if (response.data) {
     response.data = `${BASE64_PREFIX}${response.data}`;
```
originalix, line 121: P2: [Preserve the cropper's actual image MIME type]

When the replacement picker returns a HEIC or PNG, this code still prepends `data:image/jpeg;base64,`. The old wrapper requested `forceJpg`, but this migration removes that conversion, so the URI declaration can no longer be assumed to match the bytes.

That incorrect declaration reaches the avatar and hardware-wallpaper consumers and can make valid selections fail to decode or process. Build the prefix from `response.mime` (with a JPEG fallback only when it is absent) in both picker and cropper result paths.

originalix (dismissed):

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

packages/components/src/composite/ImageCrop/index.native.tsx
huhuanming: Not an issue here: `@onekeyfe/react-native-image-crop-picker` never returns the picked file's own bytes. On both platforms every result, with or without cropping, goes through `makeResult`, which decodes the image (HEIC and PNG included) and re-encodes it as JPEG (`UIImage.jpegData` on iOS, `Bitmap.compress(JPEG)` on Android) and reports `mime: "image/jpeg"`. So `data` is always JPEG and the `data:image/jpeg;base64,` prefix always matches. `forceJpg` was dropped because that conversion is now unconditional in the native module.

Added a comment explaining this in 1010fe974f.

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-3bef8123-69f9-5bd3-8c50-24d3868f2230)

originalix (approved):

PR #13557, opened 2026-09-18
fix: skip SWR cache writes for unchanged results and log Hermes heap deltas

originalix:
## Summary
- `swrCacheUtils.set()` treats a deep-equal result as unchanged: it refreshes the entry timestamp in memory and leaves the key pending, but no longer schedules a flush.
- Slow `flush` / `reload` log lines (`swrCacheSlowOp`) now carry Hermes heap deltas: GC count, GC wall time, bytes allocated during the operation, heap size afterwards.
- `AppPerfScene.swrCacheSlowOp` gains the four optional fields.

## Intent & Context
iOS logs from QA and from an internal device showed the `main` JS runtime blocked for hundreds of milliseconds while the app sat idle on a market token detail screen, with taps queuing up behind the block. The instrumentation from #13535 attributed the blocks to `swrCacheUtils.flush()` and the mirror's acknowledgement pass. The idle trigger was a 6-second poller re-setting an identical 1.8 KB entry; every re-set scheduled a full flush.

This is the first of two PRs. It removes the trigger and adds the heap counters needed to explain why the same parse work runs several times slower on `main` than on `bg`. The follow-up restructures the `main` mirror to per-entry storage so that real writes stop paying whole-store passes.

## Root Cause
`set()` had no equality check: every successful `usePromiseResult` with an `swrKey` wrote its result, and any write scheduled a full flush two seconds later. On native `main` a flush re-reads and re-serializes the whole SWR store (1.4–3 MB in the captured logs) several times over — disk parse, merge, prune/stringify, mirror optimistic apply, canonical apply on acknowledgement — for 100–800 ms of main-thread time per flush. Parse throughput on `main` was also far below `bg`'s on the same data, which points at GC pressure; the heap deltas added here are meant to confirm or rule that out.

## Design Decisions
- Equality is lodash `isEqual` on the cached `d` against the new data, checked before any serialization: an unchanged result skips the single-entry stringify as well, and a changed result bails out at the first differing leaf. No serialized strings are retained, so memory is unchanged.
- The timestamp still refreshes in memory and the key stays pending (`_updatedKeys` + `_dirty`) without a timer. `isFresh` / `getWithTimestamp` consumers keep seeing a live timestamp (the market lists drop entries older than five minutes), the flush merge keeps this runtime's entry instead of the disk copy, and the refreshed timestamp is persisted by whichever flush happens next: a real change, `flushNow` on background, or a reload. An unchanged result on its own never starts a flush.
- A same-reference re-set counts as unchanged. None of the nine `set()` call sites mutates a cached object in place and re-sets it.
- Heap stats come from `HermesInternal.getInstrumentedStats` behind optional chaining and numeric guards; on runtimes without it the log payload is byte-for-byte what it was, so the existing exact-match log tests still pass. Hermes reports cumulative GC wall time in seconds; it is converted to milliseconds.

## Changes Detail
- `packages/shared/src/utils/swrCacheUtils.ts`: unchanged-result path in `set()`; `readHermesHeapStats` / `diffHermesHeapStats`; before/after sampling in `flush()` and `reloadFromStorage()`.
- `packages/shared/src/logger/scopes/app/scenes/perf.ts`: optional `heapBytes`, `allocatedBytes`, `gcCount`, `gcMs` on `swrCacheSlowOp`.
- `packages/shared/src/utils/swrCacheUtils.test.ts`: an unchanged result does not rewrite the store (whole-store backend); an unchanged result rides along with the next flush instead of starting one (native patch backend); heap deltas are attached to a slow flush.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (both runtimes), Desktop, Web, Extension. `set()` is shared; the heap fields only appear on Hermes.
- **Risk Areas**: an entry whose data stays unchanged through a long idle period keeps an older timestamp on disk until the next flush, so after a cold start it looks older than it is and is refetched sooner (or dropped by the five-minute market-list rule). `isEqual` adds a deep walk per `set()`, bounded by the entry that was already being serialized.
- **Runtime scope**: on iOS/Android each runtime applies the change to its own `swrCacheUtils` instance. Storage schema and RPC payloads are unchanged; no migration.

## Test plan
- [x] `yarn jest packages/shared/src/utils/swrCacheUtils.test.ts` — 40/40
- [ ] iOS: stay idle on a market token detail screen for two minutes; `swrCacheSlowOp` flush lines must stop appearing every ~6 s and taps must stay responsive
- [ ] iOS: `gcMs` / `allocatedBytes` / `gcCount` / `heapBytes` present on slow flush lines, with `gcMs` never above `durationMs`


sidmorizon:
Cursor review failed.

sidmorizon (approved):

PR #13566, opened 2026-09-18
fix: count Arc shared-balance USDC once in client fiat totals(OK-63633)

weatherstar:
[OK-63633](<https://onekeyhq.atlassian.net/browse/OK-63633>)

----------

## Summary
- Count the Arc (`evm--5042`) shared-balance USDC once in every client-side fiat total: home header, per-network worth, all-networks merge, Pro2 portfolio, web account panel, small-balance footer, aggregate token detail header.
- Declare the new server marker fields `sharedBalanceExcluded` / `sharedBalanceWith` on `IToken` and resolve them once in `ServiceToken.fetchAccountTokens` into a per-entry `ITokenFiat.sharedBalanceExcludedFromTotal` flag.
- Token list data is never filtered: both USDC rows stay in `data` / `keys` / `map`, the ERC-20 `0x3600…` row remains listed, swappable and approvable.

## Intent & Context
Arc's USDC is the chain's native gas token (18 decimals) and also exposes an ERC-20 interface at `0x3600000000000000000000000000000000000000` (6 decimals). Arc defines this as "two interfaces, one balance". `/wallet/v1/account/token/list` returns both rows, and every place the client re-sums `fiatValue` counted the same balance twice, inflating the home total, per-network assets and Portfolio.

Backend work (onchain OK-63529, wallet BFF OK-63604) adds a marker on the excluded ERC-20 row and skips it in server-side totals. The backend owner confirmed the client re-sums token lists itself (OK-63529 Q4) and asked the frontend to apply the same rule (Slack #prod-design, 2026-09-17). The marker is live on the test environment since 2026-09-18.

Server contract (frozen, OK-63529 tech spec §3.3):
- The excluded ERC-20 row carries `sharedBalanceExcluded: true` AND `sharedBalanceWith: <primary address>`, always together. The primary (native USDC) and every other token on every network lack both fields.
- The Arc native address is `''`, so `sharedBalanceWith` is `''`. Presence must be checked with `!== undefined`; a truthy check misclassifies it.
- Both rows keep their real `balance` / `fiatValue`.

## Design Decisions
- **Resolved flag on the fiat map entry instead of threading an exclusion set.** The three sum helpers in `tokenValueUtils.ts` (and the cache-floor / all-networks / dApp-filter re-sum paths) only see `Record<$key, ITokenFiat>` and never the token object. Marking the fiat entry once lets every map-only consumer honour the decision without new parameters, and the flag rides along through `allTokens` merge, `filterAccountTokenListByLimit`, the local token cache and the slim cold-start bundle because those all share the same entry references.
- **Rule mirrors the wallet BFF exactly (OK-63604).** Skip a marked row only when a valid primary exists in the summed set: same `networkId`, `address === sharedBalanceWith`, itself unmarked, finite positive `price`, finite `fiatValue`. Otherwise count the ERC-20 row once (never zero times) so a user holding only ERC-20 USDC, or whose native row is hidden / blocked / unpriced, still sees the right total.
- **Cross-bucket lookup over `tokens ∪ smallBalanceTokens`.** The native row always sorts first, so the ERC-20 row can fall into `smallBalanceTokens` alone; a per-bucket check would re-include it and double count. Risk-only rows are not part of any total and are ignored.
- **Match on `(networkId, address)`, never address alone.** Every EVM native token has address `''`; in all-networks flows Ethereum's native ETH must not be treated as Arc's primary.
- **Explicitly NOT the `defiMarked` filter pattern.** `tokenSelectorFilterUtils` drops rows from `data` / `keys` / `map`; applying that here would remove the ERC-20 interface from the list and token selector, and zero the USDC total when the native row is hidden. The backend owner flagged this trap in OK-63529.
- **Bucket `fiatValue` is recomputed from the flagged map only when a marker is present**, so non-Arc responses keep the server bucket total byte-for-byte. This also makes the client correct regardless of whether the BFF fix has shipped.
- **No local token cache version bump.** The only version constant in this area (`TOKEN_COLD_START_CLEANUP_VERSION`) is the one-time purge flag for the old cold-start key, not a shape version. The cached `tokenListValue` is the server total and self-heals on the next live fetch; the slim cold cache now carries the flag so cold-start small-balance sums do not double count.
- **Aggregate ROW values (`flattenAggregateTokensMap` / `sumAggregateEntry`) are intentionally untouched.** They are per-row display values, not totals, and today only `0x3600…` is a config member, so the native row is never folded into the aggregate row (pre-existing native-`''` vs `0x3600…` key mismatch, tracked separately).

## Changes Detail
- `packages/shared/types/token.ts`: `IToken.sharedBalanceExcluded` / `sharedBalanceWith` (server marker) and `ITokenFiat.sharedBalanceExcludedFromTotal` (resolved decision).
- `packages/shared/src/utils/sharedBalanceUtils.ts` (new): `isSharedBalanceMarkedToken`, `resolveSharedBalanceExcludedKeys` (pure rule over candidates) and `applySharedBalanceExclusionToTokenGroups` (flags / un-flags fiat entries in place across `tokens ∪ smallBalanceTokens`, recomputes bucket `fiatValue` only when touched).
- `packages/shared/src/utils/tokenValueUtils.ts`: `sumFiatValuesFromTokens`, `sumFiatValuesIgnoringUnavailable`, `sumTokenGroupsFiatValueIgnoringUnavailable` skip flagged entries via a shared `isCountedInTotal` guard.
- `packages/kit-bg/src/services/ServiceToken.ts`: apply the rule after USD normalization and metadata merge, before the dApp-token filter, so the flag reaches the dApp re-sum, `allTokens`, the local cache and every account-worth consumer.
- `packages/kit/src/views/Home/components/TokenListBlock/buildHomeTokenListCacheIngestRound.ts`: replace the duplicated local sum with the shared `sumFiatValuesFromTokens` so the cached small-balance total honours the flag.
- `packages/kit/src/views/AssetDetails/pages/TokenDetails/useAggregateTokenDetails.ts`: apply the same rule to the aggregate detail header fiat and balance sums (Arc native 18-dec and ERC-20 6-dec parse to the same amount).
- `packages/shared/src/utils/tokenListSlimColdCacheUtils.ts`, `packages/kit/src/states/jotai/contexts/tokenList/cells/coldStart.ts`: carry the flag through the slim cold-cache compact / widen round trip.
- `apps/mobile/bundle-registry/module-id-registry.json`: register the new shared util.
- Tests: `sharedBalanceUtils.test.ts` (new), `tokenValueUtils.test.ts`, `ServiceToken.fetchAccountTokens.test.ts`.

## Risk Assessment
- **Risk Level**: Low
- **Risk Areas**: `ServiceToken.fetchAccountTokens` runs for every token list fetch, but the new step is a no-op fast path unless a row carries the marker (Arc only today). The `tokenValueUtils` guard change is the widest-reaching edit; it only changes behaviour for entries carrying the new flag. `buildHomeTokenListCacheIngestRound` swaps a duplicated local sum for the shared helper with identical semantics for unflagged entries.

## Test plan
- [x] `yarn jest packages/shared/src/utils/sharedBalanceUtils.test.ts packages/shared/src/utils/tokenValueUtils.test.ts packages/kit-bg/src/services/ServiceToken.fetchAccountTokens.test.ts packages/shared/src/utils/tokenListSlimColdCacheUtils.test.ts packages/shared/src/utils/tokenSelectorFilterUtils.test.ts packages/kit/src/views/Home/components/TokenListBlock/buildHomeTokenListCacheIngestRound.test.ts`
- [ ] Test env, Arc account holding BOTH native USDC and ERC-20 `0x3600…` USDC: home total / single-network worth / all-networks / Portfolio / aggregate USDC detail header count the balance once.
- [ ] Test env, account holding ONLY ERC-20 `0x3600…` USDC (or native hidden / blocked): total equals that row's `fiatValue`, not 0.
- [ ] Token list still shows both rows; `0x3600…` can still be selected for swap / approve / send.
- [ ] Non-Arc networks: totals unchanged.

[OK-63633]: https://onekeyhq.atlassian.net/browse/OK-63633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Cursor review failed.

sidmorizon:
Cursor review failed.

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-64cd4d8d-a42d-5e47-812e-138af6edad47)

originalix (approved):

packages/kit-bg/src/services/ServiceToken.ts
```diff
@@ -486,6 +487,15 @@ class ServiceToken extends ServiceBase {
         };
       });
 
+    // Shared-balance groups (OK-63633, Arc native USDC vs ERC-20 0x3600…):
+    // resolve ONCE over tokens ∪ smallBalanceTokens which marked rows must be
+    // skipped by totals and flag their fiat-map entries. Rows are never
+    // dropped — the ERC-20 interface stays listed / swappable.
+    applySharedBalanceExclusionToTokenGroups({
```
originalix, line 494: P2: [Shared-balance exclusion runs before token-selector filtering]

When `withoutDappToken` or `withoutWalletToken` removes the valid primary after this call, the retained marked interface keeps `sharedBalanceExcludedFromTotal=true` even though it is now the only visible row. Downstream account-worth and selector totals therefore report zero for a balance that is still listed.

Please apply the shared-balance resolver after the token-selector filters, or rerun it on the filtered `tokens` and `smallBalanceTokens` groups before returning the response.

weatherstar: Fixed in #13571 (b16475ac4f), since this PR is already merged. `applySharedBalanceExclusionToTokenGroups` now runs after the `withoutDappToken` / `withoutWalletToken` filters, over the filtered `tokens` and `smallBalanceTokens` groups, so a marked ERC-20 row whose primary was filtered out is counted once instead of zero times. The flag still reaches `allTokens`, the local cache and the account-worth consumers because those read the same fiat-map entries. Regression test added: `counts the ERC-20 row when the selector filter drops its primary`.

PR #13511, opened 2026-09-17
fix: round desktop market detail top corners(OK-63582)

huhuanming:
[OK-63582](<https://onekeyhq.atlassian.net/browse/OK-63582>)

----------

## Summary
- Add desktop-only top corner radius and clipping to MarketDetailV2 content.
- Preserve the shared BasicPage border and rounded frame.

## Validation
- CDP desktop visual verification: passed.
- `yarn agent:check --profile commit`: passed.
- `git diff --check`: passed.
- PR profile was not rerun in the isolated worktree because dependency installation was intentionally stopped per request.

[OK-63582]: https://onekeyhq.atlassian.net/browse/OK-63582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #12316, opened 2026-07-07
perf: optimize MarketDetail lazy loading｜OK-57271

zy2000-star:
[OK-57271](https://onekeyhq.atlassian.net/browse/OK-57271)

----------

## Summary

This PR optimizes the MarketDetail loading flow by splitting the page into a lighter immediately-rendered shell and lazily loaded heavy modules. The goal is to reduce the time users spend on the full-page lazy loading fallback when navigating from the market list to the detail page, while keeping the existing fallback visual style unchanged.

## Loading Strategy

Before this change, entering MarketDetail required more of the detail page code to be loaded before the page could render. This made the navigation more likely to show the lazy loading fallback.

After this change, MarketDetail is split into:

1. **Lightweight detail shell**
   - Loads first.
   - Owns route params, providers, header, layout selection, and token detail refresh.
   - Allows the detail page structure to render earlier.

2. **Platform-specific layout layer**
   - Web/desktop uses `MarketDetailResponsiveLayout.tsx`.
   - Native uses `MarketDetailResponsiveLayout.native.tsx`.
   - Native only imports `MobileLayout`, so the iOS/Android bundle does not statically pull in desktop layout code.

3. **Heavy modules kept lazy**
   - TradingView chart remains lazy.
   - Desktop information tabs remain lazy.
   - Mobile swap dialog wrapper remains lazy.
   - Overview modules remain lazy.
   - These modules continue loading after the shell/layout has mounted.

## Mobile Optimization

Mobile first-render behavior is further improved:

- `InformationPanel` is no longer lazy inside `MobileLayout`.
- `MobileInformationTabs` is no longer lazy inside `MobileLayout`.
- These two are visible near the top of the mobile detail page, so making them eager avoids an extra local spinner after the page has already entered.
- TradingView remains lazy because it is still the heaviest part and includes chart/WebView initialization.

## Preload Changes

The preload logic is now platform-aware:

- Desktop preload targets desktop layout and desktop information tabs.
- Mobile/native preload targets mobile layout and mobile-specific modules.
- Native no longer preloads desktop-only modules.
- `SwapPanelWrap` is only preloaded for mobile, where it is used by the mobile swap dialog.

This keeps preload useful without pulling unnecessary cross-platform chunks.

## Result

The page can enter MarketDetail earlier, render the header and mobile information area sooner, and defer heavier modules to their own lazy loading path. This reduces the visible full-page lazy loading time while preserving the existing loading UI and avoiding unnecessary native bundle work.

## Verification

[...]

PR #13567, opened 2026-09-18
fix: keep page backdrop for preloaded market detail(OK-63582)

huhuanming:
[OK-63582](<https://onekeyhq.atlassian.net/browse/OK-63582>)

----------

## Summary
- Keep LazyLoadPage's `$bgSubdued` backdrop when `createMarketDetailV2Route` renders the preloaded detail shell directly, so the desktop-mode BasicPage card keeps its rounded top corners on desktop, web and extension. Native keeps its current tree.
- Remove the `Page.Body` background, radius and overflow added in #13511. They sit inside BasicPage, which already clips its children with the same radius, so they never changed the corners.

## Root cause
Opening a token from the Market list preloads the detail shell first. The route then renders the shell directly instead of going through `LazyLoadPage` (the bypass came with #12316). That also skips the `$bgSubdued` backdrop `LazyLoadPage` wraps pages in, so the square `$bgApp` stack card shows around BasicPage's rounded top corners.

Reloading on a detail URL still goes through `LazyLoadPage`, so the corners look correct on that path. #13511 was likely checked on that path.

## Validation
Desktop dev app driven over CDP. I sampled pixels just outside the card's top-left and top-right corners and compared them with the `$bgSubdued` backdrop:

| Scenario | Before (`x`, includes #13511) | After |
| --- | --- | --- |
| List → detail, light, 1197 / 900 px | ❌ `#ffffff` vs `#f9f9f9` | ✅ |
| List → detail, dark, 1197 / 900 px | ❌ `#0f0f0f` vs `#191919` | ✅ |
| Reload on detail URL, light / dark | ✅ | ✅ |

The final run used both commits, with the #13511 styles already removed.

- `yarn jest packages/kit/src/views/Market/MarketDetailV2 packages/kit/src/components/LazyLoadPage`: 86 suites / 747 tests passed. The new route test fails without the fix.
- `yarn tsc:only`: passed.
- `yarn agent:check --profile commit`: passed for both commits.
- Web and extension use the same non-native route branch but were not run separately.

[OK-63582]: https://onekeyhq.atlassian.net/browse/OK-63582?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


originalix (approved):

weatherstar (approved):

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Cursor review failed.

PR #13576, opened 2026-09-18
fix: keep stocks and robinhood tabs under weak network｜OK-63704

zy2000-star:
[OK-63704](<https://onekeyhq.atlassian.net/browse/OK-63704>)

----------

## Summary

On a weak or offline network, Market Home could render Watchlist / Trending / Top Coins / Perps while Stocks and Robinhood disappeared. The lists looked empty because those two surfaces never mounted, not because their copy was missing.

Market Home V2 tabs come from `GET /utility/v2/market/basic-config` (`spotCategories`). Watchlist and Perps are always-on local tabs. Everything else is config-driven. When that request was still pending or had failed, the client fell back to a hardcoded Trending category (plus Top Coins). Stocks (`stocks`, `/utility/v1/stocks`) and Robinhood (`robinhood_meme`) only exist when those category IDs are present, so the fallback never created their list trees. Hardcoding a label would not have helped: the missing piece was the tab identity that decides which list to mount.

A second issue made recovery worse. `useMarketBasicConfig` only persisted a successful config under SWR on native, so desktop/web cold-started from the incomplete fallback every time. `memoizee({ promise: true })` also kept rejected promises for basic-config and token-list fetches, so reconnect / Retry could reuse the failure instead of hitting the network again. Token-list Retry and `forceRemote` were similarly native-only, so desktop/web empty states had no way to bypass that cache.

## Changes

- Share a fallback category list that includes `trending`, `stocks`, and `robinhood_meme`, and use it on both Market Home and the Market token selector so Stocks / Robinhood tabs exist before config returns.
- Persist the last successful Market Home basic-config on every platform (`swrKeys.marketHomeConfig`) and ignore empty/failed payloads so a previous good config can survive a weak-network launch.
- Clear the basic-config memoizee cache after a rejected fetch in both `ServiceMarketV2` and the light API path, so a later retry is not stuck on the failed promise.
- Always drop the token-list memoizee entry on `forceRemote`, including desktop/web, instead of limiting that invalidation to Android.
- Record token-list first-page errors and treat user Retry as `forceRemote` on all platforms. Desktop and mobile empty states now show Retry when the current query failed, so a recovered network can refill Stocks / Robinhood without remounting the page.

Stocks and Robinhood list APIs can still fail while the device is offline. In that case the tabs stay visible and Retry is the recovery path; the lists are not expected to invent data without a successful response.

## Impact

Affects Market Home and the Market token selector on desktop, mobile, web, and extension. Does not change Watchlist / Perps tab construction, stock/Robinhood API contracts, or other Market surfaces that do not consume `spotCategories`.

[OK-63704]: https://onekeyhq.atlassian.net/browse/OK-63704?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-3ed352cd-1281-5d7c-8c78-967ebcaac317)

originalix (approved):

PR #13578, opened 2026-09-18
chore: add Discovery home diagnostics logs(OK-63713)

huhuanming:
[OK-63713](<https://onekeyhq.atlassian.net/browse/OK-63713>)

----------

## Summary
- Add a local-only logger scene, `discovery => homeDiagnostics`, that traces the native Discovery home (Browser dashboard) render chain end to end into `app-latest.log` through NativeLogger.
- Instrument the browser host, the outer Market/DeFi/Browser pager, the dashboard and its bookmarks section, per-tab WebView layers, and the homepage network call. Every log fires on change only and never goes to the server.
- No behavior change: the patch only adds logging hooks, `onLayout` / `onContentSizeChange` / scroll-end callbacks, and a probe inside the dashboard freeze that renders `null`.

## Intent & Context
OK-63713: on iOS and Android, the Discover > Browser dashboard sometimes shows the header with a blank body after switching back from another tab or after closing a popup. QA also reproduced it against the production API. It could not be reproduced on an iOS 26.5 simulator (dev build, test and production API) through tab switches, header taps and swipes, modals, three restored web tabs, or background/foreground.

The reporter's device log (6.6.0, build 2026091850) shows this sequence before the blank screen:
1. Tap Browser, swipe to DeFi, swipe back to Browser.
2. Wallet tab, open the account selector, remove an account and then a wallet.
3. The app sits in the background for about 90 s with the remove dialog open.
4. Back in the foreground, the modal closes, then Discover is tapped: blank.

The existing logs cannot tell which layer went blank. This PR adds diagnostics to ship in a hot-update bundle, so the next occurrence produces an explainable `app-latest.log`.

## Root Cause
Not confirmed yet; confirming it is the purpose of this PR. The main suspect is `e6b0a9778f` (2026-09-09), which wrapped the dashboard content in `DelayedFreeze freeze={!(useRouteIsFocused() && isActive)}`. A closed gate, or a failed Suspense unhide after thawing, would render exactly this: the header stays and the body is empty. Other candidates the logs also cover: the outer pager's native page drifting from the page JS considers active, a WebView layer covering the dashboard, a zero-size layout, or empty data.

## Design Decisions
- Logs use `@LogToLocal` only, so they go through NativeLogger to the device log file and never to the server. A unit test locks the scene to local output, following the market navigation diagnostics in #13487.
- Logs are change-driven: snapshots are compared by content, and layouts and content sizes are deduped. A full simulator session produced about 60 lines, far below the NativeLogger rate limit.
- The dashboard gate log mirrors every input of `useRouteIsFocused`: navigation focus, app lock, and the root route count at mount versus now, plus the root route names. A closed gate can then be attributed to a single input.
- A layout-effect probe inside `DelayedFreeze` logs `revealed` / `hidden` from what React actually committed, which separates "the gate stayed closed" from "the content did not come back".
- The pager logs the last native `onPageScroll` position alongside the JS page on `pageSelected`, `scrollState` and `setPage`, to catch native/JS page desync.
- Everything is gated to native. Dev builds only write a scene once it is enabled in the logger config; release and OTA builds write all scenes.

## Changes Detail
- `packages/shared/src/logger/scopes/discovery/scenes/homeDiagnostics.ts` (+ test, scope registration): the new local-only scene and its payload types.
- `packages/kit/src/views/Discovery/hooks/useDiscoveryHomeDiagnostics.tsx`: shared helpers — instance ids, mount/unmount logs, log-on-change, layout logger, Discovery focus and AppState logs, the focus-gate mirror, and the reveal probe.
- `Browser.native.tsx`: host lifecycle, Discovery focus, AppState, a visibility state snapshot (header tab, `displayHomePage`, dashboard/web layer visibility, tab count, header spacer), and page body / dashboard layer / web layer layouts.
- `OuterTabPagerView.tsx`: pager lifecycle and state snapshot, native `pageSelected` / `scrollState` events and programmatic `setPage` with the last native position, and pager and page layouts.
- `DashboardContent.tsx`: dashboard lifecycle, focus gate, reveal probe, render plan (which sections render and why), scroll layout / content size / scroll-end offset, and the homepage and bookmark request lifecycle, including focus-triggered refreshes skipped by the gate.
- `BookmarksSection.tsx`: its own bookmark request lifecycle and refresh triggers.
- `MobileBrowserContent.tsx`: per-tab WebView mount and visibility state.
- `ServiceDiscovery.ts`: network-level logs for `GET /utility/v1/discover/dapp/homepage` cache misses, with duration, counts, and errors.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS, Android). Desktop and web keep the shared components but log nothing.
- **Risk Areas**: The only non-logging edits are the added layout / content-size / scroll-end callbacks and `[diagId]` added to two `usePromiseResult` dependency lists. `diagId` comes from `useState` and never changes, so request timing stays as before.

## Test plan
- [x] `yarn agent:check --profile commit` passes (lint, format, tsc, test integrity).
- [x] `homeDiagnostics.test.ts` asserts every method logs locally only.
- [x] iOS 26.5 simulator DevSession with the scene enabled: `app-latest.log` shows mount, the homepage request down to the network call, pager `setPage`/`pageSelected`, the gate closing when a Market detail page is pushed (`hidden`, content size 0), reopening on return (`revealed`, content size restored), and AppState background/active. The UI is unchanged.
- [ ] Ship in a hot-update bundle, reproduce OK-63713 on a device, pull `app-latest.log`, and grep `homeDiagnostics` at the blank moment.

[OK-63713]: https://onekeyhq.atlassian.net/browse/OK-63713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5288d9f7-2b8f-5bdb-8475-a0ede5a48d8f)

PR #13579, opened 2026-09-18
fix: remove startViewTransition to stop desktop white flash (OK-63696)

huhuanming:
[OK-63696](<https://onekeyhq.atlassian.net/browse/OK-63696>)

----------

## Summary
- On desktop, toggling either switch in Settings > Wallet > Customize transaction sometimes flashed the whole window white (OK-63696). Both handlers wrapped the settings write in `startViewTransition`, which calls `document.startViewTransition` on web and desktop.
- Remove the wrapper from the call sites that don't need it. The switches that sit directly on the settings tab pages keep it, because toggling them feels janky without it.

## Root cause
- `document.startViewTransition` snapshots the whole document and cross-fades the old and new snapshots for about 250 ms. In QA's recording, 8 of about 20 toggles produced 1–3 frames where the entire window (title bar and settings modal included) was mixed with ~9–13% white. That is the `#ffffff` BrowserWindow background showing through a cross-fade frame.
- Rendering also stays paused until the callback settles, and Chromium only aborts the transition after ~4 s. The Touch ID switch awaited `promptPasswordVerify()` inside it, so when no passcode was cached the prompt stayed invisible for ~4 s.
- The wrappers were added to hide a Suspense re-suspend flicker (the react#31819 comments). #13367 (OK-53013) fixed that at the source: the async auth atoms no longer depend on atoms that change at runtime.

## Changes
- **Settings:** write directly in the Customize transaction switches, account alignment, auto-lock (duration and system idle lock) and Protection, whose `handleTransition` now runs `fn` directly. The tab page toggles (Bluetooth, menu bar tray, haptic feedback, BTC multiple addresses, Prefer Gas Account) keep `startViewTransition`.
- **`BiologyAuthSwitchContainer`:** the async handler is called directly.
- **`AccountAvatar`:** drop `displayedSource`/`displayedKey`, which only existed to defer the swap into the transition. Render `uriSource` and key the first-load fade-in on `sourceKey`.
- **`PagePrimeCloudSync` auto-lock dialog:** await `setAppLockDuration` before `onContinue()`, so the existing try/catch sees its errors.
- **Helper:** `startViewTransition` stays for the tab page toggles; its recipe is removed from the `1k-ui-recipes` skill.

## Testing
Desktop dev build (Electron 43, macOS) driven over CDP. Every setting was restored afterwards.
- Customize transaction before the fix: 6 toggles started 6 view transitions (18 `-ua-view-transition-*` animations). After the fix: 0.
- Customize transaction, account alignment and auto-lock after the fix: 0 view transitions and 0 React Suspense hides (`display: none !important` on committed content). Values persisted across a reload.
- Touch ID switch with the passcode cache cleared: the prompt painted at ~0.23 s, versus 4.1 s with the wrapper (the transition was aborted with `TimeoutError` at 4007 ms). The prompt was cancelled without entering a passcode.
- `yarn agent:check --profile commit` and `--profile pr`: passed.

Not exercised:
- The flash itself. Renderer-side frame capture never showed it and this machine has no screen-recording permission, so QA should recheck on a build.
- Protection page switches (the page always asks for the passcode) and enabling Touch ID end to end (it triggers a real Touch ID prompt).
- AccountAvatar and the Prime cloud sync dialog were changed last and were not run.
- iOS and Android. The native helper was `setTimeout(fn)`, so only a one-tick deferral is gone there.

[OK-63696]: https://onekeyhq.atlassian.net/browse/OK-63696?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5f8d392f-17ac-5217-8840-494fe3ce2d19)

originalix (approved):

PR #13580, opened 2026-09-18
fix: limit page scroll range while the keyboard is open (OK-63698)

huhuanming:
[OK-63698](<https://onekeyhq.atlassian.net/browse/OK-63698>)

----------

## Summary
- Address book → Add address and Add existing wallet → Watch-only address → Import address on iOS (OK-63698): with the keyboard open, the form could be scrolled completely out of view. Only the header, a blank body and the footer button above the keyboard were left. A downward drag in that blank area then dragged the whole sheet down instead of scrolling back, and a longer drag dismissed it.
- When `Page.Footer` lifts itself above the keyboard, the native `Page` stops adding the keyboard height as a bottom inset to its `KeyboardAwareScrollView`. The content now scrolls only to its own end, so the last field sits just above the footer.

## Root cause
- **Two keyboard adjustments at once:**
  - On native, `Page scrollEnabled` renders `KeyboardAwareScrollView` from `react-native-keyboard-controller` 1.21.9 in its default `mode="insets"`. When the keyboard opens, it sets `contentInset.bottom` to the keyboard height, which assumes the keyboard covers the bottom of the ScrollView.
  - `Page.Footer` also lifts itself by that height with `paddingBottom`. It sits below the ScrollView in the layout, so the ScrollView is squeezed above the footer and the keyboard never covers it. The keyboard height was counted twice.
- **Measurements:** on Import address (iPhone 17 Pro, iOS 26.5), a temporary `onScroll`/`onLayout` readout showed:
  - the ScrollView shrank from 628 to 317 pt;
  - `contentInset.bottom` was 335 and the content was 531 tall;
  - max offset = 531 + 335 − 317 = 549, which is more than the whole content height.
  - On Add address the same formula gives 355 + 335 − 317 = 373, also more than its 355 of content.
- **Sheet dismissal:** on RN 0.81+, touches in a ScrollView's `contentInset` area don't reach it (facebook/react-native#54123). keyboard-controller ships `applyWorkaroundForContentInsetHitTestBug` for this. With only inset area on screen, a drag never reached the UIScrollView: the offset stayed at 549, and the iOS page sheet's swipe-to-dismiss took the gesture.

## Changes
- **`BasicPageFooter`:** reports through `onKeyboardLiftChange` whether it lifts with the keyboard: a footer is set, it has no `disableKeyboardAnimation`, and the layout is not `gtMd`. It now computes that rule once and passes it to `PageFooterContainer` as `keyboardAnimationDisabled`.
- **`PageContainer.native`:** while the footer lifts, it passes `extraKeyboardSpace={-keyboardHeight}` to `KeyboardAwareScrollView`, with the height from keyboard-controller's `useKeyboardState`. The library inset becomes 0. Scrolling the focused input into view still works. `PerpKeyboardAwareDialogContent` already uses the same pattern.
- **Unchanged cases:** pages without a footer, footers with `disableKeyboardAnimation` and `gtMd` layouts keep today's inset. Web and desktop use `PageContainer.tsx` and are not affected.

## Testing
iPhone 17 Pro simulator (iOS 26.5), Mobile DevSession:
- **Import address before the fix:** dragging up with the keyboard open reached offset 549 and a blank body. A downward drag in the blank area moved the sheet while the offset stayed at 549, and a longer drag dismissed it.
- **Import address after the fix:**
  - inset 0 and max offset 214 (= 531 − 317);
  - the last line stays above Confirm;
  - a downward drag scrolls back to the top and the sheet stays open.
- **Add address after the fix:** max offset 38, and the form stays on screen.
- **Keyboard closed:** the ScrollView is 628 pt again.
- **Auto-scroll A/B:** focusing the lower "Name (optional)" field on Import address auto-scrolls to y = 23.3 both before and after, so input auto-scroll is unchanged.
- **Final code:** the numbers above came from the first version of this change. The last edit moved the `gtMd` rule with no behavior change. Afterwards I rechecked Add address on the simulator and confirmed the served Metro bundle contains the final code.
- **Checks:** `yarn agent:check --profile commit` passed, and so did the local checks of `--profile pr`. `yarn _tsc` passed.

Not covered:
- **Android:** same code path, and keyboard-controller pads the Android ScrollView the same way. It should behave the same but was not run.
- **iPad, `gtMd` layouts and real devices.**
- **First keyboard show after a page opens:** the JS keyboard height arrives after the native animation starts, so the old inset stays for that ~250 ms animation. It is reset to 0 when the animation ends; I verified that. If JS is blocked for longer than the animation, the inset stays until the next keyboard event.
- **Existing issue, not changed here:** when a lower field is focused, auto-scroll only reveals the caret line. The lower half of the field can stay under the footer, because `KEYBOARD_AWARE_SCROLL_BOTTOM_OFFSET` is 90 and the footer is 91 pt tall.

[OK-63698]: https://onekeyhq.atlassian.net/browse/OK-63698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-255fbb54-ecb7-5888-8bab-0adea86ec92c)

originalix (approved):

PR #13584, opened 2026-09-18
fix: stop iOS 26 keyboard dimming during page transitions (OK-63665)

huhuanming:
[OK-63665](<https://onekeyhq.atlassian.net/browse/OK-63665>)

----------

## Summary
- iOS 26+ (OK-63665): the keyboard turns darker gray for a moment when it comes up during a page transition. QA hit it when returning from History to the DeFi Pendle buy page (every time), on OneKey ID → Edit profile, and on DeFi → Tokens after coming back from a token's protocol list. It is visible in light mode.
- Focus requests that would bring the keyboard up in the middle of a push/pop now wait until the transition ends, so the real keyboard slides up once the page is in place. It is one native patch that covers every React Native text input; no page code changes.

## Root cause
- When a text input becomes first responder while a `UINavigationController` push/pop is running, UIKit animates the keyboard in with the incoming screen. During that animation it draws a stand-in of the keyboard (no candidate bar or predictions). On iOS 26 this Liquid Glass stand-in renders dark gray, and it is swapped for the real keyboard when the transition finishes.
- Two triggers:
  - **`autoFocus` on a pushed screen:** `RCTTextInputComponentView` calls `becomeFirstResponder` in `didMoveToWindow`, while the push is still running.
  - **Pop back to a screen whose input had focus before the push:** UIKit restores that first responder by itself during the pop: `-[UINavigationController transitionConductor:beginPinningInputViewsForTransitionFromViewController:…]` → `-[UIKeyboardSceneDelegate _restoreInputViewsWithId:animated:]` → `-[UITextField _restoreFirstResponder]`. JS never sees this call, so JS-side delays cannot avoid it.
- Reproduced in a bare UIKit app (no React Native) on the iOS 26.5 and 27.0 simulators, so it is system behavior and still present in 27.0. It is intermittent on the simulator (6 of 8 plain push runs). Related upstream reports: react/react-native#56595 (fix PRs unmerged), kirillzyusko/react-native-keyboard-controller#1215, software-mansion/react-native-screens#1637.
- Workarounds that do not help here:
  - **`keyboardAppearance="light"`:** the stand-in was still dark in 7 of 8 runs.
  - **`animation: 'simple_push'`:** our modal stacks already use `slide_from_right`, which react-native-screens 4.26 runs through its custom animator, and QA's recordings still show the dark keyboard there.
  - **Delayed JS focus:** `SearchBar` already waits 350 ms, and a delay cannot cover UIKit's own restore on pop.

## Changes
New `patches/react-native+0.86.2+004+defer-ios26-nav-transition-focus.patch`:
- **`RCTUITextField` / `RCTUITextView`:** override `becomeFirstResponder`. On iOS 26+, if the input's view controller is in a push/pop (`transitionCoordinator.presentationStyle == UIModalPresentationNone`) and the keyboard is not already on screen, the request is queued on the coordinator's completion and replayed on the next main-queue turn. This covers `autoFocus`, `focus()` and UIKit's restore on pop.
- **Not deferred:**
  - modal presentations and dismissals, which keep the live keyboard and showed no stand-in in tests;
  - keyboard handoffs while the keyboard is already fully shown, which keep today's in-place behavior (tracked with `UIKeyboardDidShow` / `WillShow` / `WillHide` / `DidHide`);
  - iOS < 26.
- **Cancellation:** a blur (`resignFirstResponder`) or a cancelled interactive pop drops the pending focus.
- **Placement:** the helpers live in `RCTBackedTextInputDelegateAdapter.{h,mm}` because both text views already import it. There are no new files, so `pod install` is not needed.

UX change: on push/pop the keyboard no longer slides in sideways with the screen. It slides up from the bottom right after the screen settles, like UIKit apps that focus in `viewDidAppear`.

## Testing
Bare UIKit repro on the iOS 26.5 simulator, with recordings measured frame by frame:
- **No fix:** dark stand-in on push with autoFocus and on pop with restore.
- **Deferring to the transition end:** push 5/5 and pop 5/5 clean; iOS 27.0 push and pop clean. A keyboard-to-keyboard push keeps the keyboard in place.
- **Modal pageSheet / fullScreen with autoFocus:** no stand-in even without a fix (6/6), which is why presentations are excluded.

OneKey app on an iPhone 17 Pro simulator (iOS 26.5), `dev-shell --shell local` with the patch (symbol confirmed in `OneKeyWallet.debug.dylib`):
- **DeFi → Tokens, focus search → open a token → back button:** before, the keyboard was dark for about 0.45 s during the pop. After, the page slides in without a keyboard and the real keyboard slides up once it settles.
- **Push with `autoFocus`** (temporarily set on the Tokens `SearchBar`, reverted): no dark keyboard.
- **Interactive swipe back, completed:** no dark keyboard, and the keyboard plays its normal show animation.
- **Interactive swipe back, cancelled:** no keyboard pops up on the screen that stays.
- **Typing after the restored focus:** works ("WB" filters the list to WBTC).
- **Universal Search modal (`autoFocus`):** the keyboard shows, with no dark keyboard.
- **Checks:** `yarn agent:check --profile commit` and the local checks of `--profile pr` passed.

## Not covered
- **Real devices and the exact QA flows that need a wallet or login** (Pendle buy page ← History, OneKey ID → Edit profile). They use the same text input and push/pop path; please verify them on a device.
- **Release:** this is a native change, so it ships with an app release, not OTA.
- **Swipe back started, cancelled, then the back button:** the input is not refocused afterwards, because UIKit's restore already ran during the cancelled swipe. There is no visual glitch. I could not compare this case with stock UIKit on the simulator.
- **Universal Search in the dev build:** the keyboard appeared about 3.7 s after opening, about 1.2 s after its list rendered. That matches `SearchBar`'s 350 ms `setTimeout` running late on a busy dev JS thread, and the patch skips modal presentations. An A/B with the unpatched remote shell was not possible because its download failed.
- **Other platforms:** Android is unaffected, and iOS < 26 is unchanged.

[OK-63665]: https://onekeyhq.atlassian.net/browse/OK-63665?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-6c1a2118-af69-52e2-8b0c-bd27a6a4fc45)

originalix:

Review of commit `06c21d435a2a` finished, but some comments could not be published automatically.

originalix (approved):

PR #13585, opened 2026-09-18
fix: remove the white band behind the home offline banner in dark mode (OK-63706)

huhuanming:
[OK-63706](<https://onekeyhq.atlassian.net/browse/OK-63706>)

----------

## Summary
- On the native Home tab in dark mode, the offline banner ("You are offline. Please check your network.") sat on a full-width white band. The banner itself turned pale pink, and its text was almost unreadable (OK-63706).
- Paint the Home `Tabs.Container` header container with `$bgApp`, as `EarnMainTabs` already does.

## Root cause
- #13270 (OK-62183) moved the native Home alerts into the collapsible header (`renderHeader`). `react-native-collapsible-tab-view` gives that header container `backgroundColor: 'white'` (its `topContainer` style). Our native `Tabs.Container` wrapper only overrides the shadow and elevation, and Home never set a background.
- `WatchOnlyAlert` and `RiskApprovalAlert` wrap their `Alert` in a `Stack` with `bg="$bgApp"`, so they cover the white. `NetworkAlert` places its `Alert` with bare margins (`mt="$2"`, `mx="$pagePadding"`), so the white shows around it.
- In dark mode the critical Alert colors are translucent. `$bgCriticalSubdued` is `redA2` (`#f22f3e11`, about 7% alpha), which composites to `#fef1f2` over white. That is the pale pink card, and its light text has almost no contrast against it.
- In light mode `$bgApp` is `#FFFFFF`, so the white is invisible.
- The same white also showed in dark mode as a 1px full-width line just above the tab bar, even online with no banner. It comes through a layout seam between the header content and the tab bar.

## Changes
- `HomePageView`: build `headerContainerStyle` (`{ backgroundColor: theme.bgApp.val }`) from `useTheme()` and pass it to `Tabs.Container`. It is added to the `tabs` memo dependencies, so it follows live theme switches.
- The fix is scoped to Home instead of changing the shared native `Tabs` default, so other `Tabs.Container` screens are unaffected. The web `Tabs.Container` ignores `headerContainerStyle`, so web and desktop are unchanged.

## Testing
Android 16 emulator (API 36, DevSession), with a wallet on the Home tab and airplane mode on to trigger the banner. Colors were sampled from screenshots. The before/after runs toggled only this change and reloaded in the same state.
- Dark, offline:
  - Before: the band around the banner is `#ffffff`, and the card is `#fef1f2`.
  - After: the band is `#0f0f0f` (`$bgApp`), and the card is `#1e1112` (`redA2` over `#0f0f0f`). The text is readable.
- Dark, online, no banner: the 1px row above the tab bar was `#ffffff` before and is `#0f0f0f` after.
- Light mode is unchanged (`#ffffff`). Switching the system theme live in both directions repaints the header without a reload.
- With the header partly or fully collapsed while offline, no white shows.

## Not covered
- iOS was not run. It uses the same header container, so the white band should show there in dark mode too, and this change also applies to iOS.
- A faint box appears behind the Home balance after the app cold-starts offline. It shows with and without this change and is not addressed here.

[OK-63706]: https://onekeyhq.atlassian.net/browse/OK-63706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Cursor review failed.

Remote error: `cursor_write_outcome_unknown: TimeoutError: The operation timed out.`

weatherstar (approved):

originalix (approved):

PR #13262, opened 2026-09-09
fix: native chart fullscreen (OK-61030, OK-61418, OK-61424, OK-61427, OK-62076)

limichange:
[OK-61030](https://onekeyhq.atlassian.net/browse/OK-61030) [OK-61418](https://onekeyhq.atlassian.net/browse/OK-61418) [OK-61424](https://onekeyhq.atlassian.net/browse/OK-61424) [OK-61427](https://onekeyhq.atlassian.net/browse/OK-61427) [OK-62076](https://onekeyhq.atlassian.net/browse/OK-62076)

----------

Switching the native Market chart into fullscreen currently resizes it inside the scrolling page and sticky information tabs. This can clip the chart, expose overview or trading content during rotation, and delay both entry and exit.

This change renders fullscreen charts in a dedicated synchronous window layer with its own safe-area provider. The original page remains mounted but hidden, with its information and trading content frozen. The data controller stays mounted; the canvas presentation remounts while reusing its viewport and price-scale runtime. Canvas resizing runs on the UI runtime without retransferring candle and indicator arrays for size changes.

The exit action is now the rightmost toolbar control. Orientation locks remain stable when callbacks change, and inactive Discovery dashboard content is frozen to avoid unrelated work during rotation. The fullscreen layer stays below existing dialogs and menus. Its native entry is registered for strict production bundle builds.

Related Jira issues:

| Issue | Behavior addressed |
| --- | --- |
| OK-61030 | Trading-page flash when leaving fullscreen |
| OK-61418 | Transaction content appearing in fullscreen after scrolling |
| OK-61424 | Fullscreen exit action visibility |
| OK-61427 | Overview content flashing during fullscreen entry |
| OK-62076 | Slow fullscreen entry and exit |

Validation:

- `yarn agent:check --profile commit` — passed.
- `yarn agent:check --profile pr --pr 13262` — local checks passed; GitHub CI is rerunning on the latest commit.
- Production iOS union build, native module ID map validation, main/background startup budgets, and bundle architecture checks — passed locally.
- Targeted regression tests for orientation, fullscreen host lifecycle, runtime reuse and resizing, price-scale restoration, controls, and information tabs.
- Android ADB smoke checks for fullscreen bounds, menu interaction, and return to the inline chart.
- Final device smoothness and iOS visual acceptance remain with the author.

[...]

PR #13586, opened 2026-09-18
fix: stop Discover dashboard blanking during transitions (OK-63713)

huhuanming:
[OK-63713](<https://onekeyhq.atlassian.net/browse/OK-63713>)

----------

## Problem
On iOS, tapping the tab count in Discover > Browser blanks the dashboard behind the tab list while the sheet slides up. QA reads it as a flicker. The same blank also shows:
- when returning to Discover from another bottom tab (about 100 ms of header-only frames on the simulator);
- during the whole dismiss animation on slower devices;
- during a back swipe from a page pushed over Discover, such as Market detail opened from search.

## Cause
#13262 wrapped the native `DashboardContent` in `DelayedFreeze freeze={!(useRouteIsFocused() && isActive)}`. Opening a root modal, switching bottom tabs or pushing a page blurs the route while the dashboard is still on screen. react-freeze then hides it with `display: none` until JS unfreezes it after the transition. The navigators avoid this already: native tabs use `freezeOnBlur: false`, and native-stack does not freeze the screen right below the focused one.

## Change
- Remove the freeze from `DashboardContent`. This restores the behavior from before #13262, and a comment records why the content stays unfrozen.
- The native content is light: the banner, up to 14 bookmarks and the trending grid. Banner autoplay already stops when the route is not focused, and the animated welcome items render on web only.
- `isContentActive` still gates banner autoplay and the scroll handler.

## Validation
- `yarn agent:check --profile commit` and `--profile pr` passed their local checks.
- iOS 26.5 simulator (DevSession), `simctl` recordings checked frame by frame before and after the change:
  - Opening and closing the tab list keeps Trending visible behind the sheet in every frame. Before, it disappeared on the first frame after the tap.
  - Wallet → Discover shows the content on the first frame. Before, about 100 ms showed only the header.
  - A slow back swipe from Market detail reveals the dashboard content. Before, only the header was visible.
  - Switching the header from Market back to Browser still shows the dashboard.
- Not run on Android.

Fixes: https://onekeyhq.atlassian.net/browse/OK-63713

[OK-63713]: https://onekeyhq.atlassian.net/browse/OK-63713?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-93028460-fbca-50a1-8944-6262945ccf9c)

originalix (approved):

PR #13590, opened 2026-09-19
fix: keep home header actions and banner after an off-tab account switch on mobile(OK-63721)

weatherstar:
[OK-63721](<https://onekeyhq.atlassian.net/browse/OK-63721>)

----------

## Summary
- Native home: the wallet (spot) tab pane now mounts together with `Tabs.Container` even when another home tab is active, so an account switch made from the DeFi/NFT/History tab still fetches the new owner's tokens and the header (worth, WalletActions row, WalletBanner) renders instead of staying blank.
- `react-native-collapsible-tab-view` patch: new per-tab `startMounted` prop (`Tabs.Tab`), plumbed through `Container` into `Lazy` as the initial `canMount` value. Default behavior of every other `Tabs.Container` in the app is unchanged.

## Intent & Context
QA report OK-63721 (App 6.6.0, iOS): on mobile, switching accounts while the DeFi tab is active leaves the home header with only the balance number. The Send/Receive/Buy/More row and the wallet banner are missing, a blank band is left in their place and the tab row sits lower. Tapping the spot tab brings everything back.

Reproduced 1:1 on an iPhone 17 Pro simulator with a freshly imported, funded watch-only address (Binance hot wallet) and with an HD account never viewed in All Networks mode. Accounts that already have a persisted `lastConfirmedOverviewBalance.byOwner` entry are masked by that cache, which is why QA only hits it on new / never-viewed accounts.

## Root Cause
1. An account switch changes the `Tabs.Container` key (wallet + indexedAccountId, or `account.id` for Others wallets), so the whole container and header remount with `initialTabName` = the active tab (DeFi).
2. In `react-native-collapsible-tab-view` 8.0.1 `Lazy` starts with `canMount = false` regardless of `startMounted` / `lazy` (`startMounted` only seeds the fade-in opacity). A pane mounts its children only after it has been focused once. With DeFi active, the wallet pane renders an empty `ScrollView`; `TokenListBlock` (and even its `FreezeInactiveHomeTab` wrapper) never renders. Confirmed with temporary mount/render logs on the simulator.
3. Nothing else fetches the new owner's tokens: the OK-62678 off-tab refresh (`resolveOffTabTokenListRefreshOnMount`) lives inside `TokenListBlock`'s mount effect, and the `RefreshTokenList` emit in `HomePageView` only covers network switches. `accountWorthAtom` and the token-list structure keep the previous owner's data.
4. `useHomeBalanceState` therefore has no signal (no session latch, no byOwner cache, live worth belongs to the old owner) and returns `unknown` (its sticky ref was reset by the header remount). `WalletActions` returns `null` on `unknown`, `WalletBanner` is hidden unless `positive`, and `HomeHeaderContainer` keeps the 182pt native `minHeight`, producing the blank band.

Not a fresh regression by git history: the `unknown` gating dates from 6.3.0, the tab library from the TabsV2 upgrade, the account-keyed remount from #11386. It was latent and only surfaces for accounts without a cached confirmed balance.

## Design Decisions
- **Mount the wallet pane, not re-implement the fetch.** The wallet pane owns the header's data (single-network explicit refresh, All Networks fan-out, token-list structure). Lifting that logic into `HomePageView` would duplicate ~130 lines of fetch/merge-derive handling and still miss the All Networks path. Mounting the pane reuses the existing, already-verified off-tab refresh (#13300) and the freeze timing from OK-62565.
- **Opt-in per tab instead of fixing `Lazy` globally.** Making `Lazy` honor `startMounted` for every tab would mount all panes of every `Tabs.Container` in the app (Market detail, Discovery, Earn, and the four other home panes) on each mount, with a cold-start first-paint cost (OK-61505 work). The new `Tabs.Tab startMounted` prop is only set on the Portfolio tab; other home panes keep mounting lazily on first focus.
- The eagerly mounted inactive pane is frozen by `FreezeInactiveHomeTab` after its first committed render, which is the same state the pane is in today after the user has visited it once and moved to another tab, so no new runtime state is introduced.
- Web is unaffected: the web `Tab` component ignores the prop and `HomePageView` already gates pane content by `mountedHomeTabIds` there.

## Changes Detail
- `patches/react-native-collapsible-tab-view+8.0.1.patch`: `TabProps.startMounted?: boolean` (src + d.ts); `Container` passes `forceStartMounted={tabProps.get(tabName)?.startMounted}` to `Lazy`; `Lazy` initializes `canMount` from `forceStartMounted`. Regenerated with patch-package; applies cleanly to the pristine tarball.
- `packages/kit/src/views/Home/pages/HomePageView.tsx`: `startMounted={tab.id === EHomeWalletTab.Portfolio}` on the home `Tabs.Tab`, with a comment explaining why.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (iOS / Android)
- **Risk Areas**: one extra pane (TokenListBlock) mounts on every home container remount while another tab is active; its list fetch stays focus-gated, only the owner refresh / All Networks fan-out runs, which is exactly the data the header needs. Cold start on the wallet tab is unchanged (the focused pane mounted already).

## Test plan
- [x] iPhone 17 Pro simulator, single network: on the DeFi tab import a funded watch-only address; header shows worth, Send/Receive/Buy/More and the banner after the token fetch without touching the spot tab.
- [x] Simulator, All Networks: on the DeFi tab switch to an HD account never viewed in that mode; worth, actions and banner appear.
- [x] Simulator: spot / history tab switches after the fix render normally; cold start unchanged.
- [x] `git apply --check` of the regenerated patch against the pristine 8.0.1 package; `yarn agent:check --profile commit` passes.
- [ ] Device QA (iOS + Android): OK-63721 repro steps on a real device, including the reporter's watch-only All Networks accounts.

[OK-63721]: https://onekeyhq.atlassian.net/browse/OK-63721?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-a5b99f8e-2050-5c26-864a-355f419be4fe)

originalix (approved):

originalix:

Review of commit `aa5b0f346c3c` did not finish successfully.

PR #13591, opened 2026-09-19
fix: wrap account selector empty-state text on mobile (OK-63720)

huhuanming:
[OK-63720](<https://onekeyhq.atlassian.net/browse/OK-63720>)

----------

## Summary
- In the V2 account selector on mobile, a search with no matches in the watch-only or private-key wallet shows the section's empty-state message (e.g. "暂无观察账户。导入地址即可开始监控。" / "No watch-only accounts. Import an address to start monitoring."). The message was cut off at the right edge of the pane (OK-63720).
- Empty-section messages are now rendered with the existing `EmptyNoAccountsView` above the NativeList, so they wrap.

## Root cause
- `WalletDetailsV2` added each empty section's `emptyText` to the NativeList snapshot as an `action` row (`presentation: 'accountSelector'`, `height: 56`).
- NativeList action rows are single-line with a fixed height: iOS `bindAction` shows the title with `lines: 1`, and Android calls `showText(title, …, 1)`. On iOS the attributed title keeps the default word-wrapping line break mode, so the overflow is clipped mid-glyph with no ellipsis.
- On a 402pt-wide iPhone the text area of the pane is about 262pt wide. The en and zh-CN messages need two lines there, and de needs three. V1 rendered the same message with `EmptyNoAccountsView` (a `ListItem`), which wraps.

## Changes
- `WalletDetailsV2.tsx`:
  - Compute `emptySections` (sections with no data and an `emptyText`) and stop emitting the `empty:*` action rows. `emptySections` travels in `candidateList`/`presentedList` with the snapshot, so the message and the rows switch together.
  - Render `EmptyNoAccountsView` for each empty section above the NativeList container. It is hidden for mocked standard HW wallets, like the list. Its `ListItem` geometry (`mx=$2 px=$3 py=$2`, title `$bodyLg` with `py=$2`) matches the old native row: 20pt side insets, 16pt above the first line, 56pt for one line.
  - The `DeprecatedWalletBanner` from #13593 renders before this block, so a reset-device warning stays above the message.

## Design decisions
- Supporting multi-line accountSelector action rows natively (measured height on iOS, Android and web) would need an app-modules release and a dependency bump, and could not ship in a bundle update. `NetworkSectionListV2` already renders its empty state in React outside the NativeList, so this follows the same pattern.
- The message text is unchanged. In the search case it still says there are no accounts, as in V1.

## Testing
iPhone 17 Pro simulator, iOS 26.5, DevSession. Watch-only wallet with one account, searched for `ajsksksj` (no match).
- Before: "No watch-only accounts. Import an a" (clipped).
- After: the whole message wraps, 2 lines in en and zh-CN and 3 lines in de. The first line is pixel-identical to the old native row (same position, font and `#202020` color). The next line is 24pt below.
- Clearing the search brings back the account row and "Add account" with no leftover message. In dark mode the message uses the light `$text` color. No new LogBox warnings.
- Checked after a cold relaunch, with the change confirmed in the served Metro bundle.
- `jest packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack`: 8 suites, 31 tests passed.

## Not covered
- The private-key wallet was not created on the simulator, because that needs a private key and an app password. It uses the same `emptyText` → `EmptyNoAccountsView` path with a different string.
- Android, desktop/web and a deprecated hardware wallet were not run. The change is shared JS, so it applies to them too. On desktop/web the message now renders as the V1 `ListItem` instead of the web NativeList action row.
- A legacy `$$others` focused wallet (three sections, no entry in the V2 sidebar) now lists its empty-section messages above the list instead of between the sections.

[OK-63720]: https://onekeyhq.atlassian.net/browse/OK-63720?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-7a87e04d-5914-556a-84d5-209d6601dc7f)

packages/kit/src/views/AccountManagerStacks/pages/AccountSelectorStack/WalletDetails/WalletDetailsV2.tsx
```diff
@@ -927,6 +928,15 @@ function WalletDetailsViewV2({ num }: IWalletDetailsProps) {
             ) : null}
           </Stack>
         ) : null}
+        {!presentedList.isMockedStandardHwWallet &&
+        presentedList.emptySections.length ? (
+          <>
+            {deprecatedAlert}
+            {presentedList.emptySections.map((section) => (
```
originalix: P2: [Empty-section messages lose their section order]

When the `$$others` wallet has an empty middle section (for example, Watching) but populated sections before or after it, this map renders that message before the entire NativeList. The message therefore appears above earlier sections instead of where that section belongs, changing the account-category order users see. Keep the wrapped empty view attached to its section position, such as by rendering an ordered section/header structure around the list.

huhuanming: Thanks. This can't happen in the current app. Nothing can focus `$$others`: the Others entry is commented out in the V2 sidebar, the V1 sidebar isn't mounted (the router loads `AccountSelectorStackV2`), and `showAccountSelector` refocuses the active account's wallet every time the selector opens. `$$others` can only come back as leftover state with no active account. In that case every section is empty, so the messages keep their section order. If the Others view returns, the empty rows need to live inside the list, for example as multi-line NativeList action rows.

originalix (approved):

PR #13596, opened 2026-09-19
fix: iPad landscape wallet home glass flash and narrow account list (OK-63486)

huhuanming:
[OK-63486](<https://onekeyhq.atlassian.net/browse/OK-63486>)

----------

## Summary

OK-63486, iPad in landscape:

1. After going back from Create Wallet to Wallet home, the header's search bar and bell/apps capsule flashed dark before turning light.
2. Opening the account selector again sometimes showed the account list as a narrow centered column. In the QA video this starts at 0:28.

Scope: iOS only, `main` (UI) runtime only. The glass fix applies to iOS 26+ with Liquid Glass. The account list fix is native, in `@onekeyfe/react-native-native-list` 3.0.149.

## Changes

- **Glass flash:** the iOS `GlassView` wrapper defaults `colorScheme` to the app theme (`light`/`dark`) instead of expo-glass-effect's `'auto'`, the same way `BlurView` passes `tint={themeName}`. An explicit `colorScheme` still wins. Every glass in the app goes through this wrapper, so callers are unchanged.
- **Narrow account list:** upgrade all 41 `@onekeyfe/*` native modules from 3.0.148 to 3.0.149. The release carries two fixes:
  - OneKeyHQ/app-modules#119, **native-list (iOS)**: re-measure rows when the list's width changes.
  - OneKeyHQ/app-modules#118, **pager-view (Android)**: stop Market native tabs from truncating their labels at densities such as 450 dpi (OK-63273). It merged after 3.0.148 was published, so this is its first release.

## Root cause

**1. Glass flash.** With `colorScheme: 'auto'`, UIKit picks the glass variant from the content behind it, and every time the view is shown again it starts from the dark variant and fades to light about 0.5 s later. While full-screen onboarding is open, the split view's main pane is `display: none` (`UIView.hidden` on Fabric). When onboarding closes, the pane shows again and the header glass goes through that dark-to-light fade. Switching back to the Wallet tab did the same with a shorter dark hold.

**2. Narrow account list.** iOS NativeList sizes rows in `sizeForItemAt` from the list's width. On a bounds change, UIKit's flow-layout invalidation context does not ask the delegate for sizes again, so rows keep their first width, and the flow layout centers a narrower row. The account selector sheet's content is first laid out at the parent stack's width, 494 pt (the 590 pt pane minus the 96 pt wallet sidebar), before the sheet frame of 840 pt arrives. When the list's first layout landed in that window, rows stuck at about 478 pt. A later data update, such as a fiat value refresh, re-measured the rows, so the bug came and went. Opening the selector in portrait and rotating to landscape reproduces it every time.

A native `@onekeyfe/*` fix cannot be patched here (`lint:onekeyfe-patches`), so it shipped through app-modules.

## Verification

iPad Air 11" (M4) simulator, iPadOS 26.5, landscape, Mobile DevSession.

**Glass flash**, measured from `simctl` screen recordings (header luminance):

- Before: after onboarding closes, the header holds at about 80 (dark) for about 300 ms and reaches its normal value of about 245 only after about 600 ms. A tab switch holds dark for about 90 ms.
- After: in the light theme, the first frame is already light; in the dark theme, the first frame is dark with no flash. Tab switches don't flash either.

**Account list**, with the app-modules#119 Swift change built into a `dev-shell --shell local` shell:

- Open in portrait, then rotate to landscape: before, the rows are narrow and centered; after, they are full width.
- Open directly in landscape (494 pt, then 840 pt): full-width rows.
- The network selector's sections, sticky headers and scrolling behave as before.
- Android: on a Pixel Fold emulator (Android 15), folding and unfolding with the selector open re-lays the rows out at the new width with the unchanged code, so Android needed no change.

**This branch** installs the published 3.0.149 packages ([publish run](https://github.com/OneKeyHQ/app-modules/actions/runs/35429074002)). The published `RNCNativeListView.swift` is byte-identical to the file verified above; the app was not rebuilt from the published packages. `yarn install` changes only the 41 `@onekeyfe` lockfile entries, and `pod install` changes only their pod versions and checksums. The dev-vendor registry update found no module changes. `yarn agent:check --profile pr` passes.

[OK-63486]: https://onekeyhq.atlassian.net/browse/OK-63486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-aa68afe1-d24e-510b-8583-dbe185d92686)

zhaono1 (approved):

originalix (approved):

PR #13601, opened 2026-09-19
fix: prevent OneKey ID alternate sign-in card reload (OK-63716)

yikZero:
[OK-63716](<https://onekeyhq.atlassian.net/browse/OK-63716>)

----------

Keep the logged-in OneKey ID add-other-sign-in-methods card mounted across focus changes and incomplete profile refreshes. Reuse the same user's successful local Keyless preparation on remount to avoid disabled-button flashes, and show the linked status when OAuth identities arrive.

The final diff is limited to 7 files for OK-63716. Last-used sign-in method changes (OK-63287) inherited from this branch have been removed from the final diff.

Validation: `yarn agent:check --profile commit` and `yarn agent:check --profile pr` passed locally. Six targeted Jest suites passed (316 tests), including bind prompt, cache, identity classification, login dialog and ServicePrime coverage; existing act(...) warnings remain. GitHub CI was skipped before PR creation because no PR existed. No new device walkthrough was performed after the scope cleanup.

[OK-63716]: https://onekeyhq.atlassian.net/browse/OK-63716?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #13606, opened 2026-09-19
fix: register the OneKey ID bind modules in the module id registry

originalix:
## Problem

`Startup Graph Budget Check / Native startup graph budget` fails on every pull request opened or updated since #13601 was merged (the first red runs start at 14:31 UTC on 2026-09-19, three minutes after the merge), including Dependabot PRs that touch nothing in the app:

```
Error: Module is not registered in apps/mobile/bundle-registry/module-id-registry.json:
packages/kit/src/views/Prime/components/OneKeyIdLegacyOAuthBind/oneKeyIdLocalKeylessPrepareCache.ts
```

#13601 added two modules that are imported by `OneKeyIdLegacyOAuthBind.tsx`, which is part of the native graph, without registering them. The union build stops at the first one, so only one of them shows up in the log. The same strict lookup is used by the release bundle build, so bundles cannot be built from `x` either.

## Change

Register both modules. Nothing else in the registry changes.

- `oneKeyIdLocalKeylessPrepareCache.ts`
- `shouldShowOneKeyIdLegacyOAuthBindPrompt.ts`

The ids come from the repository's own allocator (`updateRegistryFromModulePaths` + `writeRegistry` in `apps/mobile/scripts/module-id-registry.js`), so they are the ids `module-id:update` would assign for the same registry state. The registry validates with 24,348 modules and 24,348 unique ids.

## Scope check

Six source files added to `x` recently are not in the registry. Only the two above are reachable from the native graph; the other four are web / desktop only and are left alone:

| File | Why it is not in the native graph |
| --- | --- |
| `components/.../Image/preloadedImageUris.ts` | imported by `ImageV2.tsx` and `preload.ts`, which have `.native` variants |
| `kit/.../ScanCamera/zxingReader.ts` | imported by `ScanCamera/index.tsx` (has `index.native.tsx`) and `scanFromURLAsync/index.web.ts` |
| `shared/src/utils/tradingViewEmbedRelease.ts`, `tradingViewEmbedPinnedReleases.ts` | imported by `tradingViewEmbedLoader.web.ts` and the web service worker |

## Verification

- This PR's own `Native startup graph budget` run is the real test: it builds both graphs in strict mode.
- `yarn agent:check --profile commit`: every check passes except `tsc-staged`, which fails locally on `Cannot find module '@onekeyfe/react-native-image-crop-picker'` (three errors, all in `components/src/composite/ImageCrop`). That dependency was added by #13553 and is missing from my local install; this PR changes two lines of JSON and no TypeScript.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-c444a9d1-6956-5283-867e-6a04690efa9f)

huhuanming (approved):

PR #13611, opened 2026-09-20
fix: stop the desktop white screen when reloading a route url

huhuanming:
## Summary

- QA refreshed the packaged desktop app while a market stock detail page (JNJ) was open and got a blank window that never came back. The Dock window list showed the stock symbol instead of `OneKey` until the app was force quit. Both symptoms are the same bug.
- A window reload from any route url now falls back to the app shell instead of sitting on a blank error page.

## Root cause

- The packaged renderer is served from `file://`, and react-navigation's web linking keeps the current route in the address bar (`file:///market/stock/JNJ?...`). Nothing on disk backs those urls, so every reload of the window fails with `ERR_FILE_NOT_FOUND`. That is not only View > Reload: `appRestart()` (language / currency change) calls `browserWindow.reload()` too, and the app-lock watchdog calls `location.reload()`.
- The `did-fail-load` handler turned that failure back into `index.html` only when the failed url contained no `"."`. A dot is exactly what a market route carries: `stockPreviewLogoUrl=https%3A%2F%2F...png`, `legacyTokenPreview={"price":"1.23"}`, or a symbol such as `BRK.B`. The recovery was skipped and the window stayed on the error page.
- The error page has no `<title>`, so the window title falls back to the last path segment. That is why the Dock listed `JNJ` / `INJ` rather than `OneKey`.
- From the QA log: the 11:13:22 reload produced a single `did-finish-load` and no renderer log at all, and the renderer process was replaced by one that burned 0.21 s of CPU over the next 10 s. The reloads that did recover (11:14:08, 11:14:11, 11:14:46) each produced two `did-finish-load` — the failed load plus the recovery reload.

## Changes

- `apps/desktop/app/libs/rendererLoadRecovery.ts` (new): `shouldReloadAppShellAfterFailedLoad()` decides on the frame and the error code instead of guessing from the url — main frame, not `ERR_ABORTED`, `file://`, and not the shell itself (which would loop).
- `apps/desktop/app/app.ts`: use it, and log only the error code, since route urls carry account and token addresses. This also stops a failing sub-frame (js-sdk iframe, dapp webview) from reloading the whole window, which the old check allowed.
- `packages/shared/src/utils/documentTitleUtils.ts` (new): `setDocumentTitle()` / `getDocumentTitle()` no-op unless `platformEnv.isWeb`. The perps title sync and react-navigation's own `documentTitle` option go through it, so desktop and the extension keep the static title from the html shell.

## Testing

- A throwaway Electron harness (not committed) that mirrors the packaged path — the `file://` interceptor, the shell url, the `did-fail-load` handler — and then does what QA did: `history.replaceState` to a route, then reload.
  - Before: white screen on `/market/stock/JNJ?...LogoUrl=...png`, `/market/stock/BRK.B` and `/market/0xabc?...legacyTokenPreview=...`, with window titles `JNJ` / `INJ`. One `did-finish-load`, matching the QA log.
  - After: 7/7 routes recover, window title back to `OneKey`, two `did-finish-load`.
- `rendererLoadRecovery.test.ts` (5 cases) and `documentTitleUtils.test.ts` (2 cases).
- `yarn agent:check --profile pr` passes except for one pre-existing error unrelated to this PR: `HomePageView.tsx:957` passes a `startMounted` prop that no tab component declares, from #13590, already on `x`.

## Not covered

- Recovery lands on the app's default route; the route open before the reload is not restored. Restoring it needs the shell to carry route state, which is a larger change.
- Dev builds serve the renderer over `http://localhost:3001`, so they never hit this path — only packaged `file://` builds do.
- No Jira id yet; add one if QA filed a ticket for this.


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-d43b8066-fe9f-5fd8-8d3b-0ca357aa0ae1)

originalix (approved):

PR #13615, opened 2026-09-20
fix: keep the market simple chart price live via the kline poll (OK-63597)

zy2000-star:
[OK-63597](<https://onekeyhq.atlassian.net/browse/OK-63597>)

----------

## Problem

On the market detail page, Simple mode showed a price and a K-line that sat
still for minutes, while Pro mode on the same token stayed live.

Simple mode had no live price source at all:

- `tokenDetail.price`, which the chart pins its last point to, comes from a 6s
  poll of `/utility/v2/market/token/detail`, and that endpoint answers from a
  snapshot that holds one price for minutes.
- The `StockSimpleChart` series itself was fetched once with no
  `pollingInterval`, so the drawn history stopped at mount time.

Pro mode looked live for an unrelated reason: TradingView runs its own K-line
feed inside the embed and reports the price back through
`TRADINGVIEW_PRICE_UPDATE`, a path Simple mode never mounts.

The earlier OK-63597 fix aligned the chart with the header, so the staleness
stopped looking like a disagreement between two numbers and started looking
like both of them freezing together.

## Why not the ohlcv websocket

That was the first attempt and it was measurably inert. The ohlcv stream only
emits on a trade: over a 30s capture it delivered frames for 8 other addresses
(including other tokens on the same chain) and none for the token being viewed.
Thin markets are exactly the ones whose snapshot sits still, so the stream
cannot correct them. The K-line endpoint, by contrast, had a fresh bucket every
minute for that same token.

## Changes

- `useMarketKlineLivePrice` polls a 10-minute 1m K-line window every 6s, takes
  the newest bucket's close and writes it back through `applyChartPriceUpdate`.
  The header and the chart tail read that same quote, so both go live at once
  and stay in sync by construction.
- Range-aware polling for the drawn series (1H 30s → 1Y/All 10min), so the
  history advances instead of becoming one straight segment to a moving tip

[OK-63597]: https://onekeyhq.atlassian.net/browse/OK-63597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-6642b678-9075-508d-86b8-0244290d9733)

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13615>)

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
```diff
@@ -158,7 +207,12 @@ export function StockSimpleChart({
   );
 
   let chartContent;
-  if (isLoading || chartState.status === 'pending') {
+  // Every polling tick re-enters the loading state, so the skeleton belongs to
+  // the first load only — a refresh keeps the current line on screen.
+  if (
+    (isLoading && !chartState.data.length) ||
+    chartState.status === 'pending'
```
sidmorizon: 🟡 P2: Token switches retain the previous chart

Severity: `non-severe`

When the token or range changes, `chartState` retains the previous series while the new request loads. The refresh loading guard treats that data as current and displays the wrong chart.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,98 @@
...
+      const timeTo = Math.floor(Date.now() / 1000);
+      const response =
+        await backgroundApiProxy.serviceMarketV2.fetchMarketTokenKline({
+          interval: MARKET_KLINE_LIVE_PRICE_INTERVAL,
+          networkId,
+          tokenAddress,
+          timeFrom: timeTo - MARKET_KLINE_LIVE_PRICE_WINDOW_SECONDS,
+          timeTo,
+          autoHandleError: false,
+        });
```
sidmorizon: 🟠 P1: Failed price polls reject unhandled

Severity: `severe`

When the K-line endpoint fails, the polling method lets the rejection escape. `usePromiseResult` launches automatic runs with `void`, so each failure becomes an unhandled rejection.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,98 @@
...
+        Number.isFinite(detailLastUpdated) ? detailLastUpdated + 1 : 0,
+      );
+      lastWrittenAtRef.current = lastUpdated;
+
+      actions.current.applyChartPriceUpdate({
+        networkId,
+        tokenAddress,
+        price: formatMarketKlineLivePrice(price),
+        lastUpdated,
+      });
```
sidmorizon, line 124: 🟡 P2: Currency switches accept stale USD quotes

Severity: `non-severe`

If currency changes during a poll, the previous `enabled` invocation still applies its USD response. The non-USD header then displays that unconverted amount until another detail refresh replaces it.

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
```diff
@@ -135,11 +177,18 @@ export function StockSimpleChart({
...
       requestTokenAddress,
+      scopeKey,
     ],
     {
       initResult: { data: [], status: 'pending' },
       watchLoading: true,
       checkIsFocused: false,
+      // Without this the series stops at mount time and only its pinned tail
+      // moves, drawing the time since as one straight segment.
+      pollingInterval: resolveStockSimpleChartPollingInterval({
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The polling interval is now a single constant (`STOCK_SIMPLE_CHART_POLLING_MS`, 30s) for every range, so a range switch no longer retunes the timer and the dependency-triggered run fires immediately. Per-range pacing moved inside the request: `resolveStockSimpleChartMinRefreshMs` lets the callback return the cached series without a network call until that range floor has elapsed, so a year of daily buckets is still not reloaded every 30s.

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
```diff
@@ -135,11 +177,18 @@ export function StockSimpleChart({
...
+        range: requestRange,
+      }),
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The polling interval is now a single constant (`STOCK_SIMPLE_CHART_POLLING_MS`, 30s) for every range, so a range switch no longer retunes the timer and the dependency-triggered run fires immediately. Per-range pacing moved inside the request: `resolveStockSimpleChartMinRefreshMs` lets the callback return the cached series without a network call until that range floor has elapsed, so a year of daily buckets is still not reloaded every 30s.

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
```diff
@@ -118,11 +153,18 @@ export function StockSimpleChart({
           tokenAddress: requestTokenAddress,
         });
 
+        lastLoadedRef.current = { key: scopeKey, data };
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. `lastLoadedRef` is written only when `scopeKeyRef.current` still equals the scope the request started for, so a late success from an abandoned scope can no longer overwrite the fallback belonging to the current one. The ref also carries `loadedAt`, which now drives the per-range refresh floor.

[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. `lastLoadedRef` is written only when `scopeKeyRef.current` still equals the scope the request started for, so a late success from an abandoned scope can no longer overwrite the fallback belonging to the current one. The ref also carries `loadedAt`, which now drives the per-range refresh floor.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,98 @@
...
+        store?.get(tokenDetailAtom())?.lastUpdated ?? 0,
+      );
+      const lastUpdated = Math.max(
+        Date.now(),
+        lastWrittenAtRef.current + 1,
+        Number.isFinite(detailLastUpdated) ? detailLastUpdated + 1 : 0,
+      );
+      lastWrittenAtRef.current = lastUpdated;
+
+      actions.current.applyChartPriceUpdate({
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The hook builds a `requestScope` from `enabled`/`networkId`/`tokenAddress`, keeps it in a ref, and re-checks it together with `useIsMounted()` after the fetch resolves. A response started while the overlay was enabled in USD is dropped if the currency, price mode, token or mount state changed in the meantime, so a stale USD close can no longer land on a converted quote.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The hook builds a `requestScope` from `enabled`/`networkId`/`tokenAddress`, keeps it in a ref, and re-checks it together with `useIsMounted()` after the fetch resolves. A response started while the overlay was enabled in USD is dropped if the currency, price mode, token or mount state changed in the meantime, so a stale USD close can no longer land on a converted quote.

zy2000-star: Fixed in 83530325b2. The hook builds a `requestScope` from `enabled`/`networkId`/`tokenAddress`, keeps it in a ref, and re-checks it together with `useIsMounted()` after the fetch resolves. A response started while the overlay was enabled in USD is dropped if the currency, price mode, token or mount state changed in the meantime, so a stale USD close can no longer land on a converted quote.

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The result carries the `scopeKey` it was loaded for, and the chart renders the skeleton whenever `chartState.scopeKey !== scopeKey`. A refresh of the current scope still keeps its line on screen, but a result from a scope the user has left is never drawn under the new range axis and quote.

zy2000-star: Fixed in 83530325b2. The result carries the `scopeKey` it was loaded for, and the chart renders the skeleton whenever `chartState.scopeKey !== scopeKey`. A refresh of the current scope still keeps its line on screen, but a result from a scope the user has left is never drawn under the new range axis and quote.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. The K-line request is wrapped in try/catch and a failure returns early, keeping the price already on screen. With `autoHandleError: false` the rejection would otherwise surface as an unhandled rejection on every polling tick, since the polling chain discards this promise.

zy2000-star: Fixed in 83530325b2. The K-line request is wrapped in try/catch and a failure returns early, keeping the price already on screen. With `autoHandleError: false` the rejection would otherwise surface as an unhandled rejection on every polling tick, since the polling chain discards this promise.

packages/kit/src/views/Market/MarketDetailV2/utils/marketKlineLivePrice.ts
```diff
@@ -0,0 +1,90 @@
...
+  networkId: string;
+  priceMode: IMarketPriceSource;
+  tokenAddress: string;
+}): boolean {
+  return Boolean(
+    priceMode === 'token' &&
+    !marketAssetId &&
+    networkId &&
+    tokenAddress &&
+    currencyId?.toLowerCase() === MARKET_KLINE_LIVE_PRICE_CURRENCY,
```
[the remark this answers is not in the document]
zy2000-star: Fixed in 83530325b2. `resolveMarketKlineLivePriceEnabled` now takes `isNative` and accepts an empty `tokenAddress` for a native coin, matching how the historical series on the same chart already requests that identity. Two new cases in `marketKlineLivePrice.test.ts` cover it.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,98 @@
...
+    },
+    [actions, enabled, networkId, store, tokenAddress],
+    {
+      pollingInterval: MARKET_KLINE_LIVE_PRICE_POLLING_MS,
+      revalidateOnFocus: true,
+      revalidateOnReconnect: true,
+      // A modal left in the navigation stack makes this route report itself as
+      // unfocused, which gates the runner into a no-op. The detail poll and the
+      // chart series both opt out of that check for the same reason.
+      checkIsFocused: false,
```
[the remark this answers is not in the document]
zy2000-star: Confirmed, and tracked as a follow-up rather than fixed in this PR. The correct gate is `shouldOwnSharedMarketDetailState` from `MarketDetailV2`, which today is neither exposed on the marketV2 context nor passed down, while this chart sits six levels below it (`MarketDetailV2 -> MarketDetailResponsiveLayout -> DesktopLayout -> TokenDesktopLayout/TopCoinsDesktopLayout -> TokenDetailChart -> StockSimpleChart`). Threading an `active` prop through that chain or introducing an `active` atom is an architectural choice we are keeping out of this change. Cross-token corruption from a retained route is already prevented by the token-identity guard in `applyChartPriceUpdate`; what remains is the redundant request volume, so this thread stays open.

packages/kit/src/views/Market/MarketDetailV2/components/StockSimpleChart/StockSimpleChart.tsx
```diff
@@ -118,12 +174,22 @@ export function StockSimpleChart({
           tokenAddress: requestTokenAddress,
         });
 
+        if (scopeKeyRef.current === scopeKey) {
+          lastLoadedRef.current = { key: scopeKey, data, loadedAt: Date.now() };
+        }
```
sidmorizon: 🟡 P2: Older refresh restores stale series

Severity: `non-severe`

Overlapping same-scope refreshes let an older request overwrite `lastLoadedRef` after a newer response wins. The next paced poll restores the older series until another network fetch completes.

zy2000-star: Fixed in caf6d5b792. The fallback write is now ordered as well as scoped: each run takes a monotonic sequence number, and `shouldStoreStockSimpleChartSeries` only lets a response publish its series when it is still the current scope **and** its sequence is higher than the stored one. An overlapping refresh that answers late is dropped, so the paced polls can no longer restore an older line. The decision is a pure function with four cases covered in `StockSimpleChart.test.ts`, including the overtaken-response one.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,129 @@
...
+      tokenAddress,
+    ],
+    {
+      pollingInterval: MARKET_KLINE_LIVE_PRICE_POLLING_MS,
+      revalidateOnFocus: true,
+      revalidateOnReconnect: true,
+      // A modal left in the navigation stack makes this route report itself as
+      // unfocused, which gates the runner into a no-op. The detail poll and the
+      // chart series both opt out of that check for the same reason.
+      checkIsFocused: false,
```
sidmorizon, line 142: 🟠 P1: Retained detail routes poll forever

Severity: `severe`

On Desktop/Web, each retained Simple detail keeps `useMarketKlineLivePrice` polling after another route takes focus. `StockSimpleChart` also disables focus checks, adding a second poll from every hidden route. Repeated navigation accumulates background API traffic without bound.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
```diff
@@ -0,0 +1,129 @@
...
+        Number.isFinite(detailLastUpdated) ? detailLastUpdated + 1 : 0,
+      );
+      lastWrittenAtRef.current = lastUpdated;
+
+      actions.current.applyChartPriceUpdate({
+        networkId,
+        tokenAddress,
+        price: formatMarketKlineLivePrice(price),
+        lastUpdated,
+      });
```
sidmorizon: 🟡 P2: Older K-line response restores stale price

Severity: `non-severe`

Reconnect revalidation can overlap `useMarketKlineLivePrice` requests, letting an older response finish last. The side effect checks scope only, then gives the older close a newer timestamp. The header and chart regress until another poll.

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/13615)

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
zy2000-star: Fixed in eb25f44fc7. Route ownership now reaches the chart: `MarketDetailV2` passes `shouldOwnSharedMarketDetailState` down as `active` (`MarketDetailResponsiveLayout` -> `DesktopLayout` -> `TokenDesktopLayout`/`TopCoinsDesktopLayout`/`StockDesktopLayout`/`StockChart` -> `TokenDetailChart` -> `StockSimpleChart`), and both polls are gated on it: `useMarketKlineLivePrice` is disabled, and the series request returns its cached scope without reaching the network. Following `useAutoRefreshTokenDetail`, the focus check stays off and the polling interval keeps its identity, so a route that becomes active again refetches immediately instead of waiting out an interval. The context was not an option here: `MarketWatchListProviderMirrorV2` mirrors one store per `storeName`, so a retained route would read the active route value. Covered by a new case in `TokenDetailChart.test.tsx`.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
zy2000-star: Fixed in eb25f44fc7. The hook now orders overlapping polls: each run takes a sequence number, and `shouldApplyMarketKlineLivePrice` requires both the scope to still be current and the sequence to be ahead of the applied one, so an older response that finishes last is dropped rather than stamped as the newest price. Same shape as `shouldStoreStockSimpleChartSeries` for the series fallback. Four new cases in `marketKlineLivePrice.test.ts` cover the first write, a newer write, an out-of-order write and a scope change.

packages/kit/src/views/Market/MarketDetailV2/hooks/useMarketKlineLivePrice.ts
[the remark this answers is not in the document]
zy2000-star: Fixed in eb25f44fc7. Route ownership now reaches the chart: `MarketDetailV2` passes `shouldOwnSharedMarketDetailState` down as `active` (`MarketDetailResponsiveLayout` -> `DesktopLayout` -> `TokenDesktopLayout`/`TopCoinsDesktopLayout`/`StockDesktopLayout`/`StockChart` -> `TokenDetailChart` -> `StockSimpleChart`), and both polls are gated on it: `useMarketKlineLivePrice` is disabled, and the series request returns its cached scope without reaching the network. Following `useAutoRefreshTokenDetail`, the focus check stays off and the polling interval keeps its identity, so a route that becomes active again refetches immediately instead of waiting out an interval. The context was not an option here: `MarketWatchListProviderMirrorV2` mirrors one store per `storeName`, so a retained route would read the active route value. Covered by a new case in `TokenDetailChart.test.tsx`.

originalix (approved):

PR #13619, opened 2026-09-20
fix: scope signature records to selected network (OK-63692)

huhuanming:
[OK-63692](<https://onekeyhq.atlassian.net/browse/OK-63692>)

----------

## Summary

- Show **All networks** in the signature record network selector.
- Render transaction, signed text, and connected site results only when they match the currently selected network and address filter. This removes records from the previous filter while the next query is pending.

## Root cause

The chain selector excludes the All networks item by default, and the signature record page did not override that default. The signature record hook also rendered the previous `usePromiseResult` value until the new asynchronous query completed. On a slow query, the selected network icon changed while cards from the previous network remained visible.

## Reproduction and verification

- Reproduced in an isolated Desktop dev profile with Ethereum, BNB Chain, and Polygon test records. With an intentionally delayed Arbitrum query, the Polygon card stayed visible for about five seconds before the fix.
- After the fix, the Polygon card disappears immediately when selecting Arbitrum; the empty result remains after the query completes.
- Confirmed BNB Chain and Polygon show only their own transaction records, and All networks shows all three test records.
- Desktop main build and renderer compilation succeeded. `yarn agent:check --profile commit` and `yarn agent:check --profile pr` passed.

A separate package-wide TypeScript invocation reports errors in files outside this change; the staged TypeScript check in `agent:check` passed.

[Jira: OK-63692](https://onekeyhq.atlassian.net/browse/OK-63692)

[OK-63692]: https://onekeyhq.atlassian.net/browse/OK-63692?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-5311500c-214b-59bb-8d5e-116bdb282473)

originalix (dismissed):

sidmorizon (dismissed):

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13619>)

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
```diff
@@ -82,18 +80,30 @@ export const useGetSignatureSections = <T extends { createdAt: number }>(
...
     },
     [networkId, query.limit, query.offset, address],
-    { initResult: { sections: [], ending: false } },
+    {
+      initResult: { sections: [], ending: false, networkId: '', address: '' },
+    },
+  );
+
+  const isCurrentFilter =
+    result.networkId === networkId && result.address === address;
```
sidmorizon, line 133: 🟡 P2: All-network pagination mixes record pages

Severity: `non-severe`

Returning to All networks after filtered pagination can let the pre-reset offset request mutate `ref` after page zero loads. Later pages contain skipped or misordered records.

huhuanming: Fixed in c5283eb915. Pagination state now records its network and address scope, so a filter change uses offset 0 in the same render instead of dispatching the previous offset before the reset effect runs. The focused regression test reproduces All networks offsets [10, 0] on the old code and [0] after the fix.

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
```diff
@@ -41,59 +41,88 @@ export const useGetSignatureSections = <T extends { createdAt: number }>(
...
+    if (ending || !isCurrentFilter || !hasLoadedFirstPageRef.current) {
       return;
     }
-    setQuery((prev) => ({ ...prev, offset: prev.offset + prev.limit }));
-  }, [ending]);
+    setQuery((prev) => ({
+      ...prev,
+      networkId,
+      address,
+      offset:
```
originalix, line 154: P1: [Stale discarded fetches can blank the current filter]

When a filter change bumps `resetGenRef` while an older request is still in flight, that older request can still settle after a newer current-filter success.

Because the discarded branch returns `{ sections: [], ending: false, networkId, address }` and `usePromiseResult` may still write that value, `isCurrentFilter` becomes false for the active filter and the list stays empty until something remounts or refetch deps change again.

Please avoid publishing discarded settlements into `result`—for example by returning the previous matching current-filter snapshot or otherwise ignoring stale settles so a late outdated response cannot blank a newer successful page.

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/13619)

originalix (dismissed):

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
huhuanming: Checked on current head 0a2f204: networkId and address are dependencies of usePromiseResult. A filter change starts a new run and advances its nonce. The older run may return the discarded empty value, but usePromiseResult calls setResult only when that run still owns the current nonce. A focus-gated dependency change invalidates the old nonce too. Thus a late old response cannot overwrite a successful current-filter result. resetGenRef prevents that old response from mutating the accumulated ref. The focused stale-request tests in usePromiseResult.test.tsx pass locally (2 passed). No additional result state is needed.

sidmorizon:
Claude review session: https://claude.ai/code/session_01CZNED8WNywhBqRaKkYAWRW

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
```diff
@@ -41,59 +41,88 @@ export const useGetSignatureSections = <T extends { createdAt: number }>(
...
+    },
+  );
+
+  const isCurrentFilter =
+    result.networkId === networkId && result.address === address;
+  const sections = useMemo(
+    () => (isCurrentFilter ? result.sections : []),
+    [isCurrentFilter, result.sections],
   );
+  const ending = isCurrentFilter ? result.ending : false;
```
sidmorizon, line 138: > # 🟠 P1: Filter switch renders a false "no records" empty state

## Problem

`sections` is forced to `[]` **synchronously in the render where the filter changes**, and the hook's public surface (`hooks.ts:127`) is still only `{ sections, onEndReached }` — it exposes no loading or stale signal:

```ts
const isCurrentFilter =
  result.networkId === networkId && result.address === address;
const sections = useMemo(
  () => (isCurrentFilter ? result.sections : []),
  [isCurrentFilter, result.sections],
);
```

All three consumers pass `ListEmptyComponent` to `Tabs.SectionList` (`Transactions.tsx:377`, `SignText.tsx:540`, `ConnectedSites.tsx:673`), and `packages/components/src/composite/Tabs/List.tsx:674-682` renders it via `noContentRenderer` whenever there are no rows. There is no skeleton or loading branch anywhere in this page. So an empty `sections` is indistinguishable from "this filter has no records", and the page immediately asserts *"You haven't signed any transactions yet"*.

The gap is not a single frame. `useConfigurableChainSelector` pushes a modal (`index.tsx:83`), and `ChainSelector.tsx:175-177` calls `onSelect(network)` **before** `navigation.goBack()` — so `setNetworkId` lands while the modal is still stacked. `useRouteIsFocused` (`useRouteIsFocused.ts:24-29`) is false in that state, and `usePromiseResult` defaults `checkIsFocused: true` (`usePromiseResult.ts:177`), so the deps-change run hits the focus gate at `usePromiseResult.ts:260` and **starts no request at all** — it only sets `isDepsChangedOnBlur`. The fetch is deferred to the refocus effect (`usePromiseResult.ts:583-590`), and on native further through `requestIdleCallback`. Only then does the round trip begin, and `ServiceSignature.getSignedTransactions` does `getNetwork` + `getVaultSettings` *per record*.

The same blanking also fires on **every search keystroke**, since `searchContent` feeds `address` with no debounce (`index.tsx:24-29`, `hooks.ts:44`).

Before this PR, `sections` came straight off the retained `usePromiseResult` value (base `hooks.ts:58-59`), so the previous list stayed until the new one landed. Clearing the stale list is the intended fix; clearing it into a *definitive* empty state is the new defect.

## Impact

Every network switch — the exact interaction this PR enables via `excludeAllNetworkItem: false` — and every search keystroke flashes the illustrated "no signed transactions / no signed text / no connected sites" empty state for the modal-dismiss window plus a full background round trip. A user switching to a network that *does* have records is told, definitively, that it has none, and on a wallet with many records may navigate away before the real data arrives.

## Recommendation

Give the hook a stale/loading signal rather than silently emitting `[]`. Pass `watchLoading: true` to `usePromiseResult` at `hooks.ts:99-101`, and return `isLoading: isLoading || !isCurrentFilter` from the `useMemo` at `hooks.ts:127`. Then in `Transactions.tsx:377`, `SignText.tsx:540` and `ConnectedSites.tsx:673`, render a skeleton instead of `ListEmptyComponent` while that flag is set, so the empty state is only asserted once a matching query has actually settled. Debouncing `address` before it reaches `query` would additionally remove the per-keystroke churn, but the loading signal is the minimal fix.

[Claude session ↗](https://claude.ai/code/session_01CZNED8WNywhBqRaKkYAWRW)

---

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
```diff
@@ -41,59 +41,93 @@ export const useGetSignatureSections = <T extends { createdAt: number }>(
...
+  const isSectionsLoading = Boolean(isLoading || !isCurrentFilter);
```
sidmorizon, line 139: 🟡 P2: Failed filters leave permanent skeletons

Severity: `non-severe`

When a filter's first `methodRef.current` call rejects, `result` retains another filter's identity. The list shows permanent skeletons with no retry path.

huhuanming: Fixed in 4105f0a8f2. A failed first-page query now settles as a current-filter error result, so the skeleton clears and each list shows an error with Retry. Retry reruns the query and restores records after success. Added a focused failure/retry regression test; the three SignatureRecord hook tests pass.

packages/kit/src/views/Setting/pages/SignatureRecord/hooks.ts
huhuanming: Addressed in 7ee536b3d2. The hook exposes loading for stale or in-flight filters, and Transactions, SignText, and ConnectedSites show skeletons until matching results settle. The focused filter-switch regression test passes.

originalix (approved):

PR #12727, opened 2026-08-03
chore: upgrade React Native stack and harden SNI requests

huhuanming:
## Summary
- upgrade the mobile UI/runtime stack to React Native 0.86.2, Expo 57.0.14, Tamagui 2.7.7, NitroModules 0.36.5, MMKV 4.3.2, and the OneKey React Native module set to 3.0.90
- add shared, bounded SNI admission and AbortSignal cancellation to Desktop and React Native requests
- cancel queued and active native work by request ID, with process-shared Android/iOS capacity across the isolated main and background JS runtimes
- harden background runtime startup, reload coalescing, Hermes background-bundle compilation, and native paste/image behavior
- align Web, React Native, and Electron on a single @sentry/core@10.39.0 dependency graph
- restore cross-platform UI behavior affected by the stack migration, including popovers, modals, tabs, selects, overlays, safe areas, and Web layout
- preserve Market idle preloading without restoring startup-critical webpackPrefetch edges

## Why
The React Native, Expo, and Tamagui migration exposed compatibility issues across native and Web UI paths. In parallel, SNI requests had no consistent cancellation contract and saturated native requests had no shared, observable queue. Queued or in-flight work could therefore continue after JavaScript cancellation, and cleanup could not be proven on actual builds.

## Runtime and request model
- Android and iOS run main and background as isolated JS runtimes with independent JS heaps and initialization order.
- Native SNI admission is process-shared, so both runtimes consume the same bounded native capacity.
- Queued and active work can be cancelled by request ID; the QA panel exposes cancellation acknowledgements, raw transport settlement, limiter state, drain, and recovery diagnostics.
- Desktop remains a single JS runtime while using the same AbortSignal-facing request contract.
- POST health checks are never replayed through fetch after an ambiguous SNI null response or transport error. GET health checks retain the existing fallback behavior.

## Supported OS
This migration intentionally raises the supported mobile baseline:

- Android: API 26 / Android 8.0 or newer
- iOS: 16.4 or newer

Android API 24-25 and iOS 15.5-16.3 devices will no longer be eligible for this update. Repository documentation and native build settings use the same values.

## Review follow-up
- replaced ambiguous StyleSheet.absoluteFill object spreads with explicit absolute edges in the affected OTP, browser, DotMap, and Prime media styles
- resolved non-union background Hermes compilation through hermes-compiler, matching the union builder
- added Market after Swap in the Web full idle-preload queue
- documented Android API 26 and iOS 16.4 in the platform requirements reference
[...]

PR #10080, opened 2026-02-09
feat: redesign wallet home layout with improved spacing and actions

franco-chan:
## Summary
- Redesign wallet action buttons with mobile card style and desktop pill buttons
- Flatten token list, DeFi, and PopularTrading sections with consistent padding (`$pagePadding`)
- Increase section heading sizes to `$headingXl` for better visual hierarchy
- Unify token list text sizes to `$bodyLgMedium` across mobile and desktop
- Redesign DeFi protocol accordion with interactive trigger and inline net worth
- Move Swap/Bridge/Earn/Sell into "More" menu on token details page
- Add Buy as a main wallet action, replacing Swap on the home page
- Improve banner items with custom widths, i18n titles, and proper navigation handlers
- Improve NumberSizeableText to use consistent font weight and line height
- Add skeleton loading for async ActionList items
- Widen right sidebar from 328px to 400px
- Add responsive gap to portfolio container (`$6` default, `$8` on gtMd)

## Test plan
- [ ] Verify wallet home page layout on mobile (< 640px) and desktop (> 768px)
- [ ] Check token list item padding on mobile — items should not touch screen edges
- [ ] Verify DeFi protocol accordion expand/collapse with new trigger styling
- [ ] Test PopularTrading table row alignment and "View more" button
- [ ] Verify wallet action buttons: Send, Receive, Buy on mobile (cards) and desktop (pills)
- [ ] Test "More" menu on token details page: Swap, Bridge, Earn, Sell actions
- [ ] Check banner carousel scrolling and navigation for static banners
- [ ] Verify balance display with long numbers (no overflow)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

<a href="https://app.devin.ai/review/onekeyhq/app-monorepo/pull/10080" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

PR #13621, opened 2026-09-20
fix: bulk send caret and back-to-home pop, group border under opaque items, disabled mobile action cards (OK-63694, OK-63708, OK-63744, OK-63776)

weatherstar:
[OK-63694](<https://onekeyhq.atlassian.net/browse/OK-63694>) [OK-63708](<https://onekeyhq.atlassian.net/browse/OK-63708>) [OK-63744](<https://onekeyhq.atlassian.net/browse/OK-63744>) [OK-63776](<https://onekeyhq.atlassian.net/browse/OK-63776>)

----------

## Summary
- Bulk send (desktop/web/ext): address textarea caret and selection now land on the right character (OK-63694).
- Bulk send (desktop/ext): "Back to home" pops the Home tab stack in one transition instead of replaying every page (OK-63708).
- dApp connect request (native): the Accounts card keeps its bottom hairline border and rounded bottom corners under the account row (OK-63744).
- Wallet actions (iOS/Android): a greyed-out Receive/Send/Swap card on the home page and token detail no longer opens the action when tapped, e.g. Receive on a watch-only account (OK-63776).

## Intent & Context
Four QA tickets from the current wallet bug batch. OK-63694 and OK-63708 were reported on desktop bulk send flows; OK-63744 was reported on Android when connecting to Uniswap (the account selector card visibly lost its bottom border). OK-63776 was reported on mobile: a watch-only account's greyed Receive button still opened the receive flow. All four are pure UI/navigation fixes with no data or signing changes.

## Root Cause
- **OK-63694**: `LineNumberedTextArea` overlays a transparent RN `TextInput` on a `SizableText` display layer. Since app-wide tabular numerals (#12391), web applies `font-variant-numeric: tabular-nums` on `body`, which the display text inherits, but react-native-web's `TextInput` base style uses the `font` shorthand, which resets `font-variant-numeric`. Hex addresses are roughly half digits, so the invisible textarea text was ~8% narrower than the display layer and the caret landed a couple of characters off.
- **OK-63708**: `popToTabRootScreen` in `NavigationContainer.tsx` unwound the nested Home stack with a recursive `goBack()` + 150 ms wait, so returning from the bulk-send Process page visibly replayed Amounts → Addresses → Home.
- **OK-63744**: measured with `onLayout` on an Android emulator. `Divider` extends Tamagui `Separator`, which ships `height: 0, maxHeight: 0` and relies on its border for a visible size. Yoga sums a child's flex basis clamped to min/max (0) when measuring the parent, but lays the child out no smaller than padding + border (one hairline), so the auto-height `YGroup` card came up one hairline short (144.76 dp instead of 145.14 dp) and its last child (account row with `bg="$bgApp"`) sat exactly on the parent's bottom border. A second regression from the Tamagui 2 upgrade (#12727): `Group.Item` no longer passes the group's radius to the first/last item, so the opaque account row also painted square corners over the card's rounded bottom corners.
- **OK-63776**: `ActionItem` in `RawActions.tsx` renders a mobile card `Stack` and a desktop pill `Button`. The home redesign (#10080) replaced the mobile `IconButton {...rest}` with the card and dropped `disabled` from it; the later bot-wallet work (#10990) only added `opacity: 0.4`. Tamagui only skips attaching press handlers when `disabled` is set, so the card looked disabled but still fired `onPress`. The desktop `Button` always received `disabled`, which is why desktop was unaffected.

## Design Decisions
- OK-63744 is fixed in the shared components rather than in the dApp page: `Divider` keeps `maxHeight`/`maxWidth` equal to its hairline border so measure and layout agree everywhere (any `YGroup` + `Divider` + opaque last row had the same bug, e.g. the Discovery tab list item), and the OneKey `Group` wrapper restores the Tamagui 1 behaviour of passing the group radius (minus its border width) to edge items. An item's explicit corner radius still wins because Tamagui's slot merge lets child props override.
- `overflow: hidden` on the group was rejected as it would clip focus rings and shadows on web.
- OK-63708 dispatches a single `POP_TO_TOP` targeted at the focused tab's nested stack key (same pattern as `earnUtils.dispatchToTargetStack`); the mobile modal path (`popStack`) is untouched.
- OK-63694 sets `fontVariantNumeric` explicitly on the web text input style only, since `getFontVariantStyle` intentionally returns nothing on web.
- OK-63776 passes the existing `effectiveDisabled` to the mobile card, so `allowPressWhenDisabled` (bot-wallet deactivated toast) keeps working and no consumer changes are needed.

## Changes Detail
- `packages/kit/src/views/BulkSend/components/AddressesInput/LineNumberedTextArea.tsx`: web textarea style adds tabular-nums so the invisible input matches the display text width.
- `packages/components/src/layouts/Navigation/Navigator/NavigationContainer.tsx`: `popToTabRootScreen` dismisses overlays, resolves the focused tab's nested stack and pops it to top once.
- `packages/components/src/content/Divider/index.tsx`: `maxHeight` (horizontal) / `maxWidth` (vertical) set to `StyleSheet.hairlineWidth`.
- `packages/components/src/primitives/Group/index.tsx`: resolve the group's radius token and clone the matching corner radii onto the first/last `Group.Item`; separator interleaving unchanged.
- `packages/kit/src/views/Home/components/WalletActions/RawActions.tsx`: mobile card `Stack` receives `disabled={effectiveDisabled}`.
- `packages/kit/src/views/Home/components/WalletActions/RawActions.test.tsx`: new unit test covering disabled, `allowPressWhenDisabled` and enabled presses on both render paths.

## Risk Assessment
- **Risk Level**: Low
- **Affected Platforms**: Mobile (Divider/Group layout; wallet action cards), Desktop / Web / Extension (bulk send caret and navigation; Divider/Group also render there)
- **Risk Areas**: `Divider` is used in ~200 places; auto-height parents grow by one hairline (the size they should have had). `Group` edge items now receive corner radii by default (matches pre-upgrade behaviour); only items that use `Group.Item` are touched. Every greyed wallet action card on mobile (Send/Receive/Swap/Buy/Earn) now blocks presses like desktop already did.

## Test plan
- [ ] Android: Discover → Uniswap → Connect → OneKey: Accounts card shows the bottom border and rounded bottom corners under the account row (skeleton and loaded states).
- [ ] Android/iOS: Discovery tab list (Browser → tabs): card top corners under the favicon row are rounded.
- [ ] Desktop/web: Bulk send → paste a list of hex addresses → click mid-address: caret lands on the clicked character.
- [ ] Desktop: Bulk send many-to-one → finish → "Back to home": single transition to Home, no intermediate pages flash.
- [ ] Extension/desktop dApp sign message and derive-context modals: card borders and corners unchanged.
- [ ] iOS/Android: watch-only account → home Receive (greyed) and ETH token detail Receive (greyed): tap does nothing; bot-wallet deactivated Receive still shows the toast.

[OK-63694]: https://onekeyhq.atlassian.net/browse/OK-63694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63708]: https://onekeyhq.atlassian.net/browse/OK-63708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63744]: https://onekeyhq.atlassian.net/browse/OK-63744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[OK-63776]: https://onekeyhq.atlassian.net/browse/OK-63776


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-ac19ac04-cabb-5c90-8225-f5928a0c71f7)

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13621>)

packages/components/src/primitives/Group/index.tsx
```diff
@@ -1,44 +1,148 @@
...
+            (!isFirst && !isLast) ||
+            !isValidElement(child) ||
+            child.type !== Item
+          ) {
+            return child;
+          }
+          return cloneElement(
+            child,
+            getEdgeRadiusProps({ isFirst, isLast, vertical, radius }),
+          );
```
sidmorizon: 🟡 P2: Explicit item radii are overwritten

Severity: `non-severe`

When an edge `Group.Item` sets corner radii, `cloneElement` replaces them with the group radius. Custom edge shapes render incorrectly.

Suggested fix:

```
          return cloneElement(child, {
            ...getEdgeRadiusProps({ isFirst, isLast, vertical, radius }),
            ...child.props,
          });
```

weatherstar: Fixed in 7955e587e0: the edge radius is now only applied to corner props the item has not set itself (filtered by `child.props[key] === undefined`), so an explicit per-item radius keeps winning over the group default. Went with a filter instead of spreading `child.props` back in so `cloneElement` only receives the props that actually change.

originalix:

Review of commit `7955e587e01e` finished, but some comments could not be published automatically.

originalix (approved):

PR #13630, opened 2026-09-20
fix: fall back when Perps withdrawal precheck is unavailable(OK-63810)

Minnzen:
[OK-63170](<https://onekeyhq.atlassian.net/browse/OK-63170>)

----------

## Summary
- Keep CCTP and HyperEVM withdrawals available when the pre-transfer fee lookup times out, loses connectivity, or returns HTTP 5xx by reserving an estimated 1.01 USDC.
- Mark the fee display as estimated and silently report lookup failures through the existing server logger.

## Intent & Context
The precheck introduced in #13464 (OK-63170) made withdrawal readiness depend on a successful fee lookup. When Hyperliquid's preTransferCheck endpoint returns HTTP 500, the form shows an unknown-error/retry row and disables Max and withdrawal, even though the withdrawal action does not require a precheck receipt. This change degrades the fee estimate while preserving the selected destination and withdrawal route.

## Root Cause
Request failures returned undefined, so the form could not resolve an account-matched reserve. The existing API failure logger writes locally; it does not send a server event.

## Design Decisions
- On SDK HTTP request errors without a response, or HTTP 5xx, return an account-bound 1.01 USDC reserve with isEstimate=true. Exclude wrapped JSON parsing errors. Keep successful dynamic reserve calculation unchanged.
- Preserve blocking for explicit restrictions, malformed responses, other HTTP statuses, and wallet-resolution failures. Resolve the wallet outside the fallback catch so unrelated account errors cannot enable withdrawal.
- Max, manual amount validation, and submission readiness already consume the same reserve. A usable fallback naturally hides the precheck error/retry row; the fee gets an estimate indicator. CCTP forwarding fees and live submission checks remain unchanged.
- Report preTransferCheckFailure with reason, optional HTTP status, and whether fallback was applied. Do not send wallet addresses, raw responses, or signatures. Keep the existing refresh lifecycle without adding a cache, timer, or automatic transaction retry.

## Changes Detail
- ServiceHyperliquidExchange: classify precheck failures and return the account-bound fallback.
- DepositWithdrawModal and withdrawFeeDisplayUtils: propagate and display the estimate flag.
- HyperLiquidScene: add the server event; focused tests cover failure classification, recovery, account identity, fee presentation, and logger routing.

## Risk Assessment
- **Risk Level**: Medium
- **Risk Areas**: 1.01 USDC is an emergency reserve, not a protocol-guaranteed fee ceiling. An actual withdrawal can still fail if fees exceed the estimate or the withdrawal service is also unavailable. Unused reserve stays in the source account.

## Test plan
- [x] 68 targeted Jest tests across reserve handling, fee display, server logger routing, and existing withdrawal routes.
- [x] Changed-file lint, formatting, full TypeScript check, background API contract, test integrity, native storage, and dependency patch checks.
- [ ] Full local agent check green: only agent-context fails because the locally installed, Git-ignored 1k-jira-batch-fix skill exceeds the workspace budget (23 implicit skills vs 22; 3205 description characters vs 3000). The tracked origin/x skill catalog passes a separate audit. No skill or budget changes are included in this PR.
- [ ] Verify the dialog under a simulated precheck outage on a running app, including Max and the hidden error/retry row.
- [ ] Confirm a real CCTP withdrawal and server-side receipt of the failure event.

[OK-63170]: https://onekeyhq.atlassian.net/browse/OK-63170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13630>)

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-e07b2cdd-6fff-55b1-8e23-eb946921e79c)

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://github.com/OneKeyHQ/app-monorepo/pull/13630)

sidmorizon:
Claude review session: https://claude.ai/code/session_014MZtfHAXUABhhBrPJXSPcS

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1985,24 +1989,53 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+          (httpStatus === undefined || (httpStatus >= 500 && httpStatus < 600));
+        defaultLogger.perp.hyperliquid.preTransferCheckFailure({
+          reason: 'requestFailed',
+          httpStatus,
+          fallbackApplied,
+        });
+        if (fallbackApplied) {
+          // An unavailable fee estimate must not disable the withdrawal rail.
+          // Leave the activation fee plus a gas buffer in the source account.
+          return { accountAddress, reserve: '1.01', isEstimate: true };
```
sidmorizon, line 2035: # 🟠 P1 The `1.01` fallback hides the retry affordance and still dead-ends balances in [2.01, 3.01) USDC

## Problem
The fallback reserve is ~100x the normal one, and returning it simultaneously removes the only in-modal recovery path.

`ServiceHyperliquidExchange.ts:2032-2036` returns a flat `'1.01'`, while the success path at `:2014` computes `BigNumber.maximum(fee, USDC_WITHDRAW_GAS_RESERVE).toFixed()` with `USDC_WITHDRAW_GAS_RESERVE = 0.01` (`packages/shared/types/hyperliquid/perp.constants.ts:70`). Because the precheck is issued with `user: HYPEREVM_SYSTEM_ADDRESS` (`:1995`) — a system address every Core→EVM send targets — the activation `fee` is structurally the "already exists" value, so the reserve in normal operation is `0.01`.

The reserve feeds the max directly (`DepositWithdrawModal.tsx:1049-1056`):

```ts
const maximum = shouldReserveWithdrawGas
  ? availableBalanceBN.minus(matchedWithdrawReserve?.reserve ?? 0)
  : availableBalanceBN;
```

and the floor is `MIN_WITHDRAW_AMOUNT = 2` for a HyperEVM destination (`perp.constants.ts:172`; no `cctpForwarding` component exists on that rail, `:1076-1085`).

At the same time, the now-truthy `matchedWithdrawReserve` makes `DepositWithdrawModal.tsx:481-485` evaluate **false**:

```ts
const hasWithdrawReserveFetchFailed =
  needsWithdrawReserve &&
  !hasWithdrawAccountChanged &&
  isCheckingWithdrawReserve === false &&
  !matchedWithdrawReserve;
```

so the critical-text row and its `perp-withdraw-reserve-retry` button (`:2690-2706`) are hidden. That button is the only user-reachable caller of `refreshWithdrawReserve`; the other call site (`:1682`) runs in the submit failure handler, which is unreachable while the amount is invalid. The reserve is not on the fee's `setInterval` refresh (`:2385-2388`).

## Impact
On a transient HyperLiquid 5xx or network blip, withdrawable max silently drops by 1.00 USDC for every user on the HyperEVM/CCTP rail. For a Core balance in `[2.01, 3.01)` USDC, `max = balance − 1.01 < 2 = min`, so no valid amount exists and withdrawal stays fully blocked — the exact failure mode this PR sets out to fix, now relocated onto small accounts. In that state the user sees only "minimum withdraw is 2 USDC" (`:1140-1147`) and a `≈` on the fee line: no error text, no Retry, and no way to recover the accurate reserve without closing or refocusing the modal. Before this PR the same outage produced an explicit, retryable error row.

## Recommendation
Two small, independent fixes:

1. Keep a recovery affordance whenever the reserve is estimated. Extend the render condition at `DepositWithdrawModal.tsx:2690` to also cover `matchedWithdrawReserve?.isEstimate` with softer copy (e.g. "fee estimate unavailable"), so `refreshWithdrawReserve` stays reachable.
2. Anchor the fallback to a named constant instead of the bare `'1.01'` literal at `:2035`. Note the literal also contradicts the formula it stands in for — `:2014` is a `maximum(fee, gasReserve)`, whereas the comment at `:2033-2034` describes `1.01` as a *sum* of activation fee plus buffer, and neither `USDC_WITHDRAW_GAS_RESERVE` nor `WITHDRAW_FEE` (`perp.constants.ts:64`, explicitly the legacy Arbitrum bridge fee) pins it. Add `USDC_WITHDRAW_FALLBACK_RESERVE` next to `USDC_WITHDRAW_GAS_RESERVE`, derive it from the named constants, use `USDC_WITHDRAW_GAS_RESERVE` as the floor, and escalate to the activation-fee-inclusive value only when a prior successful precheck for the same `accountAddress` reported a non-zero `fee`. Reference the constant (not `1.01`) in `ServiceHyperliquidExchange.reserve.test.ts:144` and `:253` so a constant change fails loudly.

[Claude session ↗](https://claude.ai/code/session_014MZtfHAXUABhhBrPJXSPcS)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1985,24 +1989,53 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+          error instanceof HttpRequestError
+            ? error.response?.status
+            : undefined;
+        const fallbackApplied =
+          error instanceof HttpRequestError &&
+          // The SDK also wraps JSON parsing errors without a response.
+          !(error.cause instanceof SyntaxError) &&
+          (httpStatus === undefined || (httpStatus >= 500 && httpStatus < 600));
+        defaultLogger.perp.hyperliquid.preTransferCheckFailure({
+          reason: 'requestFailed',
```
sidmorizon, line 2028: # 🟡 P2 `preTransferCheckFailure` collapses parse failures and pre-flight validation errors into one `requestFailed` reason

## Problem
Every throw that is not a 5xx/no-response `HttpRequestError` is reported with the same hard-coded reason at `:2027-2031`:

```ts
defaultLogger.perp.hyperliquid.preTransferCheckFailure({
  reason: 'requestFailed',
  httpStatus,
  fallbackApplied,
});
```

`httpStatus` is `undefined` for all of them (`:2018-2021`), so these three distinct causes become byte-identical `{reason: 'requestFailed', httpStatus: undefined, fallbackApplied: false}`:

- an SDK-wrapped `SyntaxError` — HyperLiquid returned unparseable JSON;
- an SDK `ValidationError` — our own `source`/`user` address failed schema validation **before any request was sent**;
- any other unexpected error thrown inside the `try`.

This event is the only signal the PR adds for these failures, since they are swallowed into `return undefined`. The new test asserts the collision rather than flagging it — `ServiceHyperliquidExchange.reserve.test.ts:199-215` feeds a `SyntaxError`-caused `HttpRequestError` and a plain `new Error('Invalid request parameters')` through one `it.each` and expects the same payload.

## Impact
When OK-63170 recurs, the server event cannot answer the first triage question: is HyperLiquid returning garbage, or are we sending a malformed request? `reason: 'requestFailed'` is also factually wrong for the `ValidationError` case, where no request ever left the client.

## Recommendation
Widen the union at `packages/shared/src/logger/scopes/perp/scenes/hyperliquid.ts:132` to include `'parseFailure' | 'unexpectedError'`, and derive the reason at `:2028`:

```ts
let reason: 'requestFailed' | 'parseFailure' | 'unexpectedError';
if (!(error instanceof HttpRequestError)) reason = 'unexpectedError';
else if (error.cause instanceof SyntaxError) reason = 'parseFailure';
else reason = 'requestFailed';
```

Then split `ServiceHyperliquidExchange.reserve.test.ts:199-215` into two cases asserting the distinct reasons.

[Claude session ↗](https://claude.ai/code/session_014MZtfHAXUABhhBrPJXSPcS)

---

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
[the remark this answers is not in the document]
Minnzen: Reviewed the call chain and confirmed the intended behavior with the PR author. OK-63810 deliberately allows the estimated reserve when preTransferCheck is unavailable due to a transport failure or HTTP 5xx. Requiring another successful call to that same endpoint before submission would restore the outage dependency this change is intended to remove.

The request uses the fixed HYPEREVM_SYSTEM_ADDRESS as user and the wallet address as source. Explicit isSanctioned=true responses still block, as do malformed responses, JSON parsing errors, and non-5xx HTTP errors. The fallback does allow submission without a fresh negative sanction result; this is an acknowledged behavior decision, not a claim that downstream sanction enforcement has been independently verified.

Keeping the agreed fallback behavior and resolving without a code change.

Minnzen: Confirmed the small-balance calculation. With the 1.01 USDC estimated reserve, some balances cannot satisfy the 2 USDC minimum. This is an accepted limitation of the conservative outage fallback; the reserve stays in the source account and is not necessarily an actual fee.

Product explicitly confirmed that the error/retry row should be hidden when an estimated reserve is available. Closing and reopening the withdrawal modal runs preTransferCheck again: a successful response replaces the fallback with the current reserve and removes the estimate marker. Focus recovery and network reconnection also trigger revalidation; simply leaving the modal open does not periodically refresh this reserve.

Keeping the agreed UI and fallback amount. A historical-fee cache or alternative reserve policy is outside this fix; extracting a named constant alone would not change the behavior. Resolving without a code change.

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
Minnzen: Confirmed that these exception categories currently share requestFailed. The classification is coarse, but it does not change fallback eligibility: SDK-wrapped JSON parsing failures and validation/unexpected errors remain blocking and report fallbackApplied=false.

For OK-63810, the event records whether the outage fallback was applied and the HTTP status when available. More granular parse/validation reasons would improve diagnostics, but are not required for the withdrawal fallback to work correctly. Treating this as an optional observability improvement and resolving without expanding this PR.

packages/shared/src/logger/scopes/perp/scenes/hyperliquid.ts
```diff
@@ -127,6 +127,15 @@ export class HyperLiquidScene extends BaseScene {
     return params;
   }
 
+  @LogToServer()
```
originalix, line 130: add LogToLocal

originalix (changes requested):

packages/shared/src/logger/scopes/perp/scenes/hyperliquid.ts
Minnzen: Added @LogToLocal({ level: 'error' }) alongside @LogToServer() in 05633fcaa3, matching the existing apiRequestFailure local log level. Updated the existing test to verify both log destinations are emitted in one call. All 4 tests in hyperliquid.test.ts and yarn agent:check --profile commit passed.

originalix (approved):

packages/kit-bg/src/services/ServiceHyperLiquid/ServiceHyperliquidExchange.ts
```diff
@@ -1985,24 +1989,53 @@ export default class ServiceHyperliquidExchange extends ServiceBase {
...
+        const result = await hyperLiquidApiClients.infoClient.preTransferCheck({
+          source: accountAddress,
+          user: HYPEREVM_SYSTEM_ADDRESS,
+        });
+        const fee = result?.fee;
+        if (
+          typeof fee !== 'string' ||
+          !/^\d+(\.\d+)?$/.test(fee) ||
+          !new BigNumber(fee).isFinite() ||
+          result.isSanctioned !== false
```
originalix, line 2002: P1: [Null precheck responses are misclassified as request failures]

When preTransferCheck resolves null or undefined, this direct `result.isSanctioned` access throws before the invalid-response branch runs. The outer catch then records `requestFailed`, so the new null/undefined regression cases fail and diagnostics are misleading.

Please guard the response object before reading `isSanctioned` and classify nullish or structurally invalid payloads as `invalidResponse`.

PR #13632, opened 2026-09-20
fix: unblock Android hardware rename confirmation

wabicai:
## Summary
- Show the confirm button loading state as soon as Pro2/Neo hardware rename submission starts.
- On Android, dismiss the keyboard without awaiting a keyboard hide event that may never arrive, then continue submission after a bounded delay. Keep the existing behavior on other platforms.

## Intent & Context
The Android `prodRelease` build intermittently left the Pro2/Neo rename dialog open after tapping Confirm, with no loading feedback. The user reproduced this on a build based on `4ed36ead6fe7` and confirmed that the loading state appears in the patched Release build.

## Root Cause
The previous handler awaited `Keyboard.dismissWithDelay(350)` before setting `isLoading`. That helper awaits `react-native-keyboard-controller`'s `keyboardDidHide` event without a timeout. The reproduction log showed the IME hiding while the dialog stayed open, consistent with a missed event leaving the handler pending. The logs do not prove the exact event delivery failure.

## Design Decisions
- Scope the behavior to Android in the existing hardware label dialog; iOS keeps its current path.
- Use an 800 ms delay to cover the observed keyboard hide animation and the prior post-hide delay, while ensuring the submit path advances.
- Leave the rename input cursor behavior for separate work.

## Changes Detail
- `HardwareLabelSetDialog.tsx`: move `isLoading` before keyboard dismissal and use a bounded Android keyboard transition before reading the form and submitting the name.

## Risk Assessment
- **Risk Level**: Medium
- **Affected Platforms**: Android mobile; iOS follows the existing path.
- **Risk Areas**: Android keyboard timing and the device label submission flow. The patched Release build was manually verified to update the device label and close the dialog.

## Test plan
- [x] Focused hardware label dialog Jest tests (15 passed).
- [x] `yarn agent:check --profile commit`.
- [x] Build and install Android `prodRelease`; verify the APK contains the rebuilt main JS bundle.
- [x] Manual Release check: confirm button loading state appears.
- [x] Manually verify the device label updates and the dialog closes after Confirm.


sidmorizon:
[View Devin Review :arrow_upper_right:](<https://app.devin.ai/review/OneKeyHQ/app-monorepo/pull/13632>)

sidmorizon:
[Cursor Cloud Agent review session ↗](https://cursor.com/agents/bc-ba78960d-f4b8-541a-8c2c-865a6a5cdd52)

sidmorizon:
@codex review

sidmorizon:
@codex security review

sidmorizon:
Claude review session: https://claude.ai/code/session_014h2XKhepWv8GhwETngc4kh

sidmorizon:
## Local Codex CLI review

[Open the review session in Slack](https://onekeygroup.slack.com/archives/C0BACUCK9H9/p1789956545567159?thread_ts=1789956544.285529&cid=C0BACUCK9H9)

huhuanming (approved):

originalix (dismissed):

patches/react-native-keyboard-controller+1.21.9.patch
```diff
@@ -0,0 +1,54 @@
...
+@@ -32,17 +34,41 @@ const dismiss = async (options?: Partial<DismissOptions>): Promise<void> => {
+ 
+   return new Promise((resolve) => {
+     if (isClosed) {
++      // Android keyboards can live in a separate window (for example a
++      // BottomSheetDialog). The activity listener does not observe that
++      // window, so its cached visibility can be stale even while an input is
++      // focused. Always forward the native dismissal on Android.
++      if (Platform.OS === "android") {
++        KeyboardControllerNative.dismiss(keepFocus, animated);
```
sidmorizon: # 🔴 P1 Android `isClosed` branch now blurs the globally-focused input on every dismiss

## Problem

`keepFocus` defaults to `false` and `animated` to `true` (pristine `src/module.ts:30-31`, unchanged by this patch). In the library's Android implementation, that combination does more than hide the IME — it also clears focus:

```kotlin
// android/.../modules/KeyboardControllerModuleImpl.kt:43-73 (1.21.9, verified against the npm tarball)
fun dismiss(keepFocus: Boolean, animated: Boolean) {
  val view: View? = FocusedInputHolder.get()
  if (view != null) {
    UiThreadUtil.runOnUiThread {
      fun maybeClearFocus() { if (!keepFocus) { view.clearFocus() } }
      if (Build.VERSION.SDK_INT >= R && !animated) { ... } else {
        imm?.hideSoftInputFromWindow(view.windowToken, 0)
        maybeClearFocus()   // ← runs, because keepFocus == false
      }
    }
  }
}
```

`view` is `FocusedInputHolder.get()` — a process-global `WeakReference<EditText>` of the last focused RN input, not anything scoped to the caller.

`isClosed` is set `false` only by `keyboardWillShow` and back to `true` by `keyboardDidHide` (`src/module.ts:19-27`). So "IME hidden **but** an RN input still focused" is a genuinely reachable state, and before this patch `dismiss()` was a complete no-op there. Two ways the app reaches it:

- **Deterministic:** Android `keyboardDismissMode="on-drag"` hides the IME via `hideSoftInputFromWindow` *without* blurring the `ReactEditText`. There are 20 such call sites in `packages/`, including `components/src/layouts/ListView/list.native.tsx:101` and `components/src/layouts/SortableListView/index.native.tsx:130`. Type in a search field → scroll the list → open any overlay, and the field is silently blurred.
- **Race:** an `autoFocus` input that has taken focus but whose IME has not yet emitted `keyboardWillShow`.

The callers that now trigger this mostly never asked to blur anything: `components/src/actions/Toast/index.tsx:453` (`Toast.show`), `components/src/actions/ActionList/index.native.tsx:520`, `components/src/actions/Popover/popoverSideEffects.ts:11,20`, and `components/src/composite/Dialog/index.tsx:780,929,960`. Line 960 is the sharpest — inside the close path, a `setTimeout(..., 300)` fires `void Keyboard.dismissWithDelay(50)` **300 ms after a dialog starts closing**, landing on whatever surface just autoFocused an input.

## Impact

Android only. Best case the user loses focus and caret in a search or amount field and has to tap again. Worst case a freshly mounted `autoFocus` input is blurred before its IME appears and **the keyboard never shows**. That is exactly the failure this repo already documented for this code path — `packages/shared/src/keyboard/index.native.ts:21-23`: *"Prefer this when the next screen still autoFocuses an input: Android `hideSoftInputFromWindow` (used by `dismissKeyboard`) can block that programmatic focus from showing the IME."* The patch makes that path fire app-wide in strictly more situations than before, to fix one native-sheet case.

## Recommendation

Pass `keepFocus: true` in this new branch only:

```ts
if (Platform.OS === "android") {
  KeyboardControllerNative.dismiss(true, animated);
}
```

`hideSoftInputFromWindow` still runs, so the stale-visibility native-sheet case described in the comment above is still fixed, but nothing gets blurred. The `isClosed === false` path keeps `keepFocus` as-is, so no currently working behavior changes. Alternatively, gate this branch on a native sheet actually being on screen rather than on `Platform.OS === "android"` for every closed-keyboard dismiss.

[Claude session ↗](https://claude.ai/code/session_014h2XKhepWv8GhwETngc4kh)

---

patches/react-native-keyboard-controller+1.21.9.patch
```diff
@@ -0,0 +1,54 @@
...
++      }
+       resolve();
```
sidmorizon, line 21: # 🟡 P2 Stale-visibility path resolves with zero wait, so awaiting callers run while the IME is still on screen

## Problem

`KeyboardControllerNative.dismiss(...)` is fire-and-forget — its entire body is wrapped in `UiThreadUtil.runOnUiThread { ... }` (`KeyboardControllerModuleImpl.kt:51`). This branch calls it and then `resolve()`s synchronously on the same tick.

So in precisely the scenario the comment above targets — *"its cached visibility can be stale even while an input is focused"* — `await KeyboardController.dismiss()` returns in ~0 ms while the Android IME hide animation (~200–300 ms) has not even started. The patch builds a perfectly good listener plus 500 ms fallback for the other branch (lines 43, 46–51) and then doesn't use it here.

## Impact

`dismissKeyboardWithDelay` (`packages/shared/src/keyboard/index.native.ts:28-31`) then waits only its own `delayMs`, which is far too short to cover a hide that hasn't begun. Awaiting call sites: `views/Send/pages/SendAmountInput/SendAmountInputContainer.tsx:2498` (80 ms), `components/DeFi/ProtocolLendingActionDialogContent.tsx:775,1585` (80 ms), `components/DeFi/ProtocolPositionActionDialog.tsx:2035` (80 ms), `components/Password/components/PasswordSetup.tsx:117` (150 ms), `views/AccountManagerStacks/components/WalletRename/HardwareLabelSetDialog.tsx:233` (350 ms).

The documented reason those awaits exist is `packages/shared/src/keyboard/index.native.ts:4-16` — OK-42598, *"`Dialog.show` ends up rendering its Sheet behind the still-visible system keyboard."* In the stale-visibility case this patch leaves that guarantee unmet.

To be clear on severity: before this patch that path hung forever, so this is an **incomplete fix rather than a regression** — hence P2.

## Recommendation

On Android, skip the `isClosed` shortcut entirely and fall through to the path this patch already builds (listener + 500 ms fallback):

```ts
if (isClosed && Platform.OS !== "android") {
  resolve();
  return;
}
```

The caller then waits for a real `keyboardDidHide`, bounded at 500 ms, instead of not waiting at all. This is independent of, and compatible with, the `keepFocus` repair suggested on the line above.

[Claude session ↗](https://claude.ai/code/session_014h2XKhepWv8GhwETngc4kh)

---

patches/react-native-keyboard-controller+1.21.9.patch
```diff
@@ -0,0 +1,54 @@
...
+-    });
++      resolve(undefined);
++    };
++    subscription = KeyboardEvents.addListener("keyboardDidHide", finish);
+ 
+     KeyboardControllerNative.dismiss(keepFocus, animated);
++    if (Platform.OS === "android") {
++      // WindowInsets events from custom Dialog windows are not delivered to
++      // the controller attached to the activity root. Do not leave callers
++      // and this event subscription pending forever when that happens.
```
sidmorizon, line 49: 🟡 P2: Rename loading feedback remains delayed

Severity: `non-severe`

A missed `keyboardDidHide` still leaves rename confirmation without loading feedback for 850 ms. [`dismissKeyboardWithDelay`](packages/shared/src/keyboard/index.native.ts:28-30) adds 350 ms before the handler sets `isLoading`.

huhuanming (dismissed):

patches/react-native-keyboard-controller+1.21.9.patch
wabicai: Fixed in d87cf2ff444: the new Android `isClosed` fallback now calls `KeyboardControllerNative.dismiss(true, animated)`, so it still forwards the IME dismissal for stale visibility without clearing the globally tracked input focus. `git apply --check` and `yarn agent:check --profile commit` pass.

patches/react-native-keyboard-controller+1.21.9.patch
```diff
@@ -0,0 +1,54 @@
...
+   return new Promise((resolve) => {
+     if (isClosed) {
++      // Android keyboards can live in a separate window (for example a
++      // BottomSheetDialog). The activity listener does not observe that
++      // window, so its cached visibility can be stale even while an input is
++      // focused. Always forward the native dismissal on Android.
++      if (Platform.OS === "android") {
++        KeyboardControllerNative.dismiss(true, animated);
++      }
+       resolve();
```
sidmorizon, line 21: 🟡 P2: Stale state resolves dismissal early

Severity: `non-severe`

When `isClosed` is stale on Android, `dismiss()` resolves immediately after starting an asynchronous keyboard hide. Awaiting callers can open the next surface while the keyboard still covers it.

zy2000-star (approved):

huhuanming (approved):

patches/react-native-keyboard-controller+1.21.9.patch
```diff
@@ -0,0 +1,54 @@
...
+@@ -32,17 +34,41 @@ const dismiss = async (options?: Partial<DismissOptions>): Promise<void> => {
+ 
+   return new Promise((resolve) => {
+     if (isClosed) {
++      // Android keyboards can live in a separate window (for example a
++      // BottomSheetDialog). The activity listener does not observe that
++      // window, so its cached visibility can be stale even while an input is
++      // focused. Always forward the native dismissal on Android.
++      if (Platform.OS === "android") {
++        KeyboardControllerNative.dismiss(true, animated);
```
sidmorizon, line 19: 🟡 P2: Default dismissal retains input focus

Severity: `non-severe`

When Android’s visibility cache is stale, `dismiss()` retains focus even when callers omit `keepFocus`. Default calls violate the library contract and leave the input focused after hiding the keyboard.

Suggested fix:

```
+        KeyboardControllerNative.dismiss(keepFocus, animated);
```

